Issue #2580, opened 2018-05-11
unable to setup trustmanager exception when TLS is enabled


hldnova:
**Problem description**

Enable TLS in config/standalone-config.properties and start standalone pravega.
The follow exception shows repeatedly in pravega log. Note that the pravega certificate was added to local java truststore.
```
2018-05-11 11:09:49,850 34706 [clientInternal-4] WARN  i.p.c.s.i.SegmentOutputStreamImpl - 98a4b7e5-56f8-4e04-a48b-fefb159615bf Failed to connect: 
java.util.concurrent.CompletionException: java.lang.RuntimeException: javax.net.ssl.SSLException: unable to setup trustmanager
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
        at java.util.concurrent.CompletableFuture.uniCompose(CompletableFuture.java:961)
        at java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:926)
        at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:442)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.RuntimeException: javax.net.ssl.SSLException: unable to setup trustmanager
        at io.pravega.client.netty.impl.ConnectionFactoryImpl.establishConnection(ConnectionFactoryImpl.java:117)
        at io.pravega.client.segment.impl.SegmentOutputStreamImpl.lambda$null$1(SegmentOutputStreamImpl.java:522)
        at java.util.concurrent.CompletableFuture.uniCompose(CompletableFuture.java:952)
        ... 9 common frames omitted
[...]

PR #3343, opened 2019-02-05
Issue 3342: Increase the read timeout to avoid test failure 

### LINKED ISSUE #3342: Sporadic build failure: TlsEnabledInProcPravegaClusterTest
**Problem description**
Encountered following test failure during travis build:
```
io.pravega.local.TlsEnabledInProcPravegaClusterTest > testWriteAndReadEventWhenConfigurationIsProper FAILED
    java.lang.AssertionError: expected:<Test event over TLS channel> but was:<null>
        at io.pravega.local.TlsEnabledInProcPravegaClusterTest.testWriteAndReadEventWhenConfigurationIsProper(TlsEnabledInProcPravegaClusterTest.java:121)
```

**Problem location**
TLS enabled cluster test

Fix the test

ravisharda:
Signed-off-by: Ravi Sharda <ravi.sharda@emc.com>

**Change log description**  
Increase read timeouts (tripled) to avoid test failures under abnormal delays in build/test environments. 

**Purpose of the change**  
Fixes #3342. 

**What the code does**  
Increases the read timeout. 

**How to verify it**  
Tests in TlsEnabledInProcPravegaClusterTest pass consistently.

### DIFF
--- a/standalone/src/test/java/io/pravega/local/TlsEnabledInProcPravegaClusterTest.java
+++ b/standalone/src/test/java/io/pravega/local/TlsEnabledInProcPravegaClusterTest.java
@@ -57,7 +57,7 @@ public class TlsEnabledInProcPravegaClusterTest extends InProcPravegaClusterTest
     // Note: Strictly speaking, this test is really an "integration test" and is a little
     // time consuming. For now, its intended to run as a unit test, but it could be moved
     // to an integration test suite if and when necessary.
-    @Test(timeout = 50000)
+    @Test(timeout = 200000)
     public void testWriteAndReadEventWhenConfigurationIsProper() throws ExecutionException,
             InterruptedException, ReinitializationRequiredException {
 
--- a/standalone/src/test/java/io/pravega/local/TlsEnabledInProcPravegaClusterTest.java
+++ b/standalone/src/test/java/io/pravega/local/TlsEnabledInProcPravegaClusterTest.java
@@ -115,7 +115,7 @@ public class TlsEnabledInProcPravegaClusterTest extends InProcPravegaClusterTest
 
         // Keeping the read timeout large so that there is ample time for reading the event even in
         // case of abnormal delays in test environments.
-        EventRead<String> event = reader.readNextEvent(50000);
+        EventRead<String> event = reader.readNextEvent(150000);
         String readMessage = event.getEvent();
         log.debug("Read event '{}", readMessage);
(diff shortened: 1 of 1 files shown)


standalone/src/test/java/io/pravega/local/TlsEnabledInProcPravegaClusterTest.java
```diff
@@ -115,7 +115,7 @@ public void testWriteAndReadEventWhenConfigurationIsProper() throws ExecutionExc
 
         // Keeping the read timeout large so that there is ample time for reading the event even in
         // case of abnormal delays in test environments.
-        EventRead<String> event = reader.readNextEvent(50000);
+        EventRead<String> event = reader.readNextEvent(150000);
```
fpj: why do we need to increase so much the read timeout? it is kind of harmless to do it, but it does not look like the test case requires a timeout of over 2 minutes.

ravisharda: It is not expected that such a large timeout would actually be needed in the build/test environments, but I used it just to reduce the chances of test failures due to unexpected and excessive delays. Should I reduce it to something like 100,000 ms, instead?

andreipaduroiu:
@ravisharda @fpj is there any progress on this? Virtually every build is failing with this and it's really starting to affect our productivity. 

Does this point to an underlying issue with the application or is it related to TLS functionality?

ravisharda:
@andreipaduroiu 
The investigation and analysis are still on. More on this below.

When we looked at the logs from a previous system test build, we noticed a bunch of errors in the log, including a message ``Rescheduling ControllerNameResolver task for after 120000 ms``, which seemed to indicate that the name resolver was sleeping for 2 minutes. There was also an SSL/TLS certificate issue visible in the logs.
 
Today, I created server certificates/keys and other PKI material to setup SSL/TLS afresh: this was to ensure that these materials were created and installed properly. The standalone mode seems to work w.r.t. reader/writer apps and controller REST endpoints using these new material, but I still see a different error (#2580) in the logs that I was trying to resolve. This is happening in SegmentOutputStreamImpl when it invokes connectionFactory.establishConnection(uri, responseProcessor). 

In the meantime, a workaround is to trigger the tests again.

fpj:
Approving this PR as a temporary change, until we fix the root cause in #3372.

PR #3349, opened 2019-02-06
Issue 3332: Unified AutoScale Stats and Metrics reporting in Segment Store

### LINKED ISSUE #3332: Unify segment stats recording in PravegaRequestProcessor and AppendProcessor
**Problem description**
Segment stats recording is done in two different places: SegmentStatsRecorder and ad-hoc, throughout AppendProcessor and PravegaRequestProcessor.

**Problem location**
AppendProcessor and PravegaRequestProcessor.

Unify.

andreipaduroiu:
**Change log description**  
- Moved the metric reporting logic from `PravegaRequestProcessor` and `AppendProcessor` into `SegmentStatsRecorderImpl`
- Reordered the sequence of calls in `AppendProcessor.handleAppendResult` to report metrics after sending the ack back to the client.

**Purpose of the change**  
Fixes #3332.

**What the code does**  
- `PravegaRequestProcessor` and `AppendProcessor` were reporting metrics at the same time as updating Segment Stats (for auto-scaling purposes). Hence, two divergent paths of reporting the same information were slowly evolving, requiring extra constructor arguments and private fields in both these classes.
- The metric reporting logic has been moved out of these two classes into `SegmentStatsRecorderImpl`, under appropriate methods. New methods have been added to it to support new types of information.
    - This also allowed the reduction of duplicated checks (i.e., check if a segment is a transaction), thus further de-cluttering the code.
- Metrics and stats are now reported after sending an ack back to the client. As such, they will not add to latency (internal calculations, metric reporting, auto-scale events, etc.) and will not fail an append if any exceptions occur in this code.
- Those operations that are expected to take a longer time to execute have been moved to an async mode (i.e., run in a new thread and not block synchronously). This enables us not to block the core thread pool should auto-scale reporting fall behind or get deadlocked. 

**How to verify it**  
Unit tests must pass.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -290,7 +254,8 @@ private Append getNextAppend() {
         }
     }
 
-    private void handleAppendResult(final Append append, Throwable exception) {
+    private void handleAppendResult(final Append append, Throwable exception, Timer elapsedTimer) {
+        boolean success = exception == null;
         try {
             boolean conditionalFailed = exception != null && (Exceptions.unwrap(exception) instanceof BadOffsetException);
```
tkaitchuck: !success && ...

andreipaduroiu: Fixed

PR #2302, opened 2018-02-08
Issue 2034: Enable TLS for segmentstore to bookkeeper interactions

arvindkandhare:
**Change log description**
* Enable TLS settings for docker deployment.
* Add config settings to enable TLS.
* Add sample key and truststores for simple deployment.

**Purpose of the change**
This fixes #2034. 

**What the code does**

**How to verify it**
Unit tests and system tests.

PR #3357, opened 2019-02-08
Issue 3311: Update Bookkeeper image to use official image as base

### LINKED ISSUE #3311: Update Bookkeeper image to use official image as base
**Problem description**
We want to enable Bookkeeper metrics in Pravega but our current Bookkeeper image cannot support this feature. The related issues are [Pravega operator Issue 63](https://github.com/pravega/pravega-operator/issues/63) and [Pravega Issue 3101](https://github.com/pravega/pravega/issues/3101).

To enable Bookkeeper metrics we need to configure Bookkeeper. The way to configure Bookkeeper is to override configurations in `bk_server`, which is already done [here](https://github.com/pravega/pravega/blob/master/docker/bookkeeper/entrypoint.sh#L40). This works for now but is not enough for enabling Bookkeeper metrics as there are too many configurations that are available. 

**Problem location**
In Pravega Bookkeeper [Dockerfile](url) and [Entrypoint.sh](url)

The official Bookkeeper image has a nice way to easily modify configurations by passing environment variables with a prefix `BK_` or `ZK_`. It has a script that can parse env variables with those prefixes and override the configuration if their names match. Details can be found [here](https://github.com/apache/bookkeeper/blob/master/docker/README.md#environment-variable-names-that-mostly-used-for-your-configuration)

It would be nice if we can use the official Bookkeeper image as our base image and customize it from that.

Tristan1900:
**Change log description**  
This PR changes the Pravega bookkeeper base image to official bookkeeper image. The advantage is that we can pass parameters with prefix `BK_` and `ZK_` as environment variables so that the scripts in official image can help us to parse those parameters and override the corresponding ones in the config file.

In addition, the official bookkeeper docker image only has the Prometheus stats provider. This PR uses bookkeeper `all` package to override the official docker image so that other stats provider such as Codahale can also be enabled.

**Purpose of the change**  
Fix #3311 

**What the code does**  
- Change the base image to the bookkeeper official image in Dockerfile.
- Use `all` package to include all stats provider.

**How to verify it**  
Test should pass


adrianmo:
PR is still in progress. I'm working on it.

docker/bookkeeper/entrypoint.sh
```diff
@@ -8,68 +8,38 @@
...
-${ZK_HOME}/bin/zkCli.sh -server $ZK_URL create /${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME} || :
-${ZK_HOME}/bin/zkCli.sh -server $ZK_URL create /${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME}/${BK_CLUSTER_NAME} || :
+until zk-shell --run-once "ls /" ${ZK_URL}; do sleep 5; done
 
-echo "format the bookie"
-# Silence exit codes with "|| :" as the command can safely fail if another instance has already formatted the bookie
-BOOKIE_CONF=${BK_HOME}/conf/bk_server.conf ${BK_HOME}/bin/bookkeeper shell metaformat -nonInteractive || :
+echo "delete bookie cookie"
+sed -i "s|metadataServiceUri=.*\$|metadataServiceUri=${BK_metadataServiceUri}|" /opt/bookkeeper/conf/bk_server.conf
+/opt/bookkeeper/bin/bookkeeper shell bookieformat -nonInteractive -force -deleteCookie
```
eolivelli: I am not sure about this step, mayve I missing bit of context
why are you always clearing the bookie ?

second point: where is the 'metaformat' ? which is now called 'initnewcluster' ?

adrianmo: @eolivelli correct me if I'm wrong, but I don't think we need to init a new cluster every time an instance of BK is started. I've added the `bookieformat` to delete the cookie related to that Bookie instance ID. When an instance running on a node was deleted and then recreated, we were getting an invalid cookie exception. The `bookieformat` attempts to fix that.

Let me know if we are following the right approach.

adrianmo:
@eolivelli one of the things we want to achieve in this PR is the ability to set up the multiple metrics providers supported by BK (https://bookkeeper.apache.org/docs/4.7.3/admin/metrics/). 

However, it looks like the `bookkeeper-server` distribution embedded in the BK Docker image only contains the Prometheus Java classes, not the rest. What do you recommend us to do to enable all metrics providers?

eolivelli:
@adrianmo which provider do we need in Pravega?
I did not think to this problem because I am using Prometheus (and not with the docker package).

If it is only a matter of adding one provider we can propose to add that provider as well.
If we need more than one we can think about proposing more variants for the image.
Currently the binaries are distributed with two packages: simple server standalone and full package with all deps and the additional 'stream storage module'
So we can think about having a docker image for the latter ad well

adrianmo:
@eolivelli we currently need Codahale, which is not included in the `bookkeeper-server` distribution. I like the idea of releasing two types of Docker images (e.g. `4.9.0` and `4.9.0-all`), one based on the `bookkeeper-server` (as of today) and another one based on the `bookkeeper-all`. I'll create an issue in the BK repository to propose that...

In the meantime, I'll update our Dockerfile to download the `bookkeeper-all` distribution and override the contents on the BK base image.

adrianmo:
FYI: https://github.com/apache/bookkeeper/issues/1943

Tristan1900:
@adrianmo Sorry that I force-pushed your commits, I have some previous commits that are not signed...

adrianmo:
@Tristan1900 is there a reason why you removed the `HEALTHCHECK` step from the `Dockerfile`?

adrianmo:
@fpj @eolivelli any comments about the changes?

docker/bookkeeper/entrypoint.sh
fpj: To confirm, is this creating a new bookie from scratch? if so, it is correct to reformat the disk, and otherwise it isn't. See the documentation:

https://bookkeeper.apache.org/docs/4.7.3/admin/bookies/

But assuming that we are starting a bookie from scratch, then why are we finding these cookies in the first place?

docker/bookkeeper/entrypoint.sh
```diff
@@ -8,68 +8,45 @@
...
-# Silence exit codes with "|| :" as the commands can safely fail with a "Node already exists" error
-${ZK_HOME}/bin/zkCli.sh -server $ZK_URL create /${PRAVEGA_PATH} || :
-${ZK_HOME}/bin/zkCli.sh -server $ZK_URL create /${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME} || :
-${ZK_HOME}/bin/zkCli.sh -server $ZK_URL create /${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME}/${BK_CLUSTER_NAME} || :
-
-echo "format the bookie"
-# Silence exit codes with "|| :" as the command can safely fail if another instance has already formatted the bookie
-BOOKIE_CONF=${BK_HOME}/conf/bk_server.conf ${BK_HOME}/bin/bookkeeper shell metaformat -nonInteractive || :
+echo "delete bookie cookie"
+# We need to update the metadata endpoint and Bookie ID before attempting delete the cookie
```
fpj: `... attempting to delete the cookie.`

docker/bookkeeper/entrypoint.sh
```diff
@@ -8,68 +8,45 @@
...
+
+export BOOKIE_PORT=${BOOKIE_PORT}
+export BK_zkServers=${BK_zkServers}
+export BK_metadataServiceUri=zk://${ZK_URL}${BK_LEDGERS_PATH}
+export BK_journalDirectory=${BK_DIR}/journal
+export BK_ledgerDirectories=${BK_DIR}/ledgers
+export BK_indexDirectories=${BK_DIR}/index
+export BK_CLUSTER_ROOT_PATH=/${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME}/${BK_CLUSTER_NAME}
+
+export BK_tlsProvider=OpenSSL
```
fpj: Why are we setting these security options like this? As far as I can tell, we are hardwiring the security configuration here, which is possibly not a good idea. cc/ @ravisharda

ravisharda: In the code below, keystores, truststores and their passwords are being dealt with (except for in line 35). 

https://github.com/pravega/pravega/blob/5af078f0f353fe7b74ba63f5d1336ff55ec6167f/docker/bookkeeper/entrypoint.sh#L33-L37

Questions:

1. I haven't got a chance to spend  enough time looking at it, but does it also imply that these files are baked in the image? Does that mean that if these need to be replaced by user-supplied material, the image would need to change? 

2. Should we consider other options for passing these sensitive material? Some of the popular options these days are: 
   * As secrets in files via a volume mounted on the container, 
   * Using secrets management facilities like [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) or  [Docker Swarm](https://docs.docker.com/engine/swarm/secrets/)

docker/bookkeeper/entrypoint.sh
adrianmo: > I am not sure about this step, mayve I missing bit of context
> why are you always clearing the bookie ?

@eolivelli correct me if I'm misunderstanding something... I'm clearing the bookie because of the following use case: say that we have a bookie running in a node, then we scale down the BK cluster and that bookie and all its associated volumes are deleted, then we scale up the BK again and a new bookie is created in the same node, having the same instance ID. If we don't format the bookie metadata, the bookie will find existing metadata from the previous bookie and it will fail to start and throw a `InvalidCookieException` exception.

> second point: where is the 'metaformat' ? which is now called 'initnewcluster' ?

`initnewcluster` is handled by the official BK Docker scripts here: https://github.com/apache/bookkeeper/blob/release-4.7.3/docker/scripts/init_bookie.sh#L47. So that script will take care of initializing a new cluster if there isn't any available. 

> To confirm, is this creating a new bookie from scratch? if so, it is correct to reformat the disk, and otherwise it isn't.

@fpj yes, every time a Pravega BK container is started, it's assuming it's a new bookie.

> But assuming that we are starting a bookie from scratch, then why are we finding these cookies in the first place?

Because a previous Bookie with the same instance ID could have been created and destroyed before, leaving conflicting metadata behind that would cause the Bookie to fail when starting.

docker/bookkeeper/entrypoint.sh
adrianmo: Those TLS settings were hardwired in https://github.com/pravega/pravega/pull/2302 in an attempt to secure segmentstore-bookie communication, the only thing that changed in this PR is how it's done: before, they were appended to `bk_server.conf`; now, via env vars. The result should be the same.

I also think that it's not a good idea to hardwire them, however, I'm not sure if we should just use that configuration as the default, so that users can override it, or do something else. What do you think @fpj @ravisharda ?

ravisharda:
A separate issue (#3390) has been created to assess, evaluate and decide on the approach for passing secrets to Pravega containers.

eolivelli:
@adrianmo 
Can it happen that the bookie is restarted without losing bookie contents and identity?
In that case we are not deleting the cookie.
Can you confirm?

adrianmo:
@eolivelli not that I know. In our current environment (i.e. Kubernetes), containers are handled as ephemeral objects that are created and disposed when they are needed (e.g. scaling down, node evacuation), but they are not restarted or reused. 

However, if we also want to cover such scenarios, what would you recommend doing to handle restarts without losing bookie contents (local and remote)?  Maybe run a sanity check (`bookiesanity` command) before starting the bookie?

docker/bookkeeper/entrypoint.sh
adrianmo: Issue https://github.com/pravega/pravega/issues/3390 has been create to discuss this.

adrianmo:
@eolivelli I tried to run the sanity check or obtain information from the bookie with `bookieinfo` before deleting the cookie, but the command errors out and does not return the control to the main process. It gets stuck, so we can't check the exit code and react to that. 

```
10:09:36,698 ERROR Failed to get bookie list :
org.apache.bookkeeper.client.BKException$ZKException: Error while using ZooKeeper                                                                              
        at org.apache.bookkeeper.discover.ZKRegistrationClient.lambda$getChildren$0(ZKRegistrationClient.java:212)                                             
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient$25$1.processResult(ZooKeeperClient.java:1174)                                                       
        at org.apache.zookeeper.ClientCnxn$EventThread.processEvent(ClientCnxn.java:604)                                                                       
        at org.apache.zookeeper.ClientCnxn$EventThread.run(ClientCnxn.java:508)
Exception in thread "main" org.apache.bookkeeper.client.BKException$ZKException: Error while using ZooKeeper                                                   
        at org.apache.bookkeeper.discover.ZKRegistrationClient.lambda$getChildren$0(ZKRegistrationClient.java:212)                                             
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient$25$1.processResult(ZooKeeperClient.java:1174)                                                       
        at org.apache.zookeeper.ClientCnxn$EventThread.processEvent(ClientCnxn.java:604)                                                                       
        at org.apache.zookeeper.ClientCnxn$EventThread.run(ClientCnxn.java:508)
```

eolivelli:
> @eolivelli not that I know. In our current environment (i.e. Kubernetes), containers are handled as ephemeral objects that are created and disposed when they are not needed (e.g. scaling down, node evacuation), but they are not restarted or reused.

I am now a Kubernetes user, but if you don't keep Bookies data, in case of a lost bookie the replicator needs to copy all of the data from one bookie (which must be up-and-running) to a new bookie.
It is a terrible expense of resources.
How many Bookie are you expected to run in a production environment ? and which volume of data ?

I expect that Bookies are using persistent "volumes", and on the disk you are keeping the identity of the bookie

btw it this is the expected workload I am fine with it, I am used to very long-lived bookies, not to ephemeral ones.
If this is how it works you can evaluate to enable the 'journalSyncData=true' flag, this way you won't fsync the disks

eolivelli:
@adrianmo *bookieinfo* does not report about the local bookie but about all of the bookies in the cluster

adrianmo:
Thanks for your input, @eolivelli

> How many Bookie are you expected to run in a production environment ? and which volume of data ?

We have a hard minimum requirement of 3 bookies, and we try to protect bookies from planned disruptions with a`PodDisruptionBudget`. However, for production environments, we will surely have more than 3. The number of bookies will depend on the use case requirements and workload to be run. We haven't gotten to this point yet.

> btw it this is the expected workload I am fine with it, I am used to very long-lived bookies, not to ephemeral ones.

Don't get me wrong, Bookie pods are not expected to be terminated and recreated often. They are supposed to live long and have prosperous life :) but Kubernetes' `Pod` objects are considered to be relatively ephemeral (see [here](https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods)), similar to individual containers. 

> I expect that Bookies are using persistent "volumes", and on the disk you are keeping the identity of the bookie

Yes, they do use persistent volumes for the ledger and journal. That's why I wanted to find a way to check whether or not the bookie has a working environment (i.e. valid cookie, valid ledger...) when starting the bookie. This can happen when a Pod is evacuated from a node and recreated in a new node, the Pod will be brand new, but the volume will be reused.

> If this is how it works you can evaluate to enable the 'journalSyncData=true' flag, this way you won't fsync the disks

Will check that option, thanks!

adrianmo:
@eolivelli do you have any idea how to check whether or not a new container bookie has the proper conditions to resume (metadata in ZK, local data, etc)?

adrianmo:
@eolivelli I've been experimenting with your suggestions and this is what I found out.

> This way the bookie is preserved in case of presence of data , and it is automatically "formatted" if all is empty.

There's a use case where this does not work: scale down -> up. Say that we have 5 bookies (from `1` to `5`). We scale down the cluster to 4 bookies, that automatically deletes the `bookie-5` container. After a while, we scale up the cluster back to 5 bookies, creating a `bookie-5` again. When `bookie-5` starts, it will find metadata from the previous `bookie-5` container, but there will be no cookie in the directories as they will be empty, so it will end up here: https://github.com/apache/bookkeeper/blob/release-4.7.3/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java#L475

I think it all boils down to identifying when a container is started for the first time, or it's being restarted. I we identify that, we will be able to format bookie data and metadata when a new container is created, and reuse existing data and metadata when it is restarted.

I'll make the changes and push them for review.

docker/bookkeeper/Dockerfile
```diff
@@ -7,52 +7,37 @@
...
+RUN chmod +x -R /opt/bookkeeper/scripts/
+
+ENTRYPOINT [ "/bin/bash", "/opt/bookkeeper/scripts/pravega_entrypoint.sh" ]
+CMD ["bookie"]
 
-ENTRYPOINT [ "/bin/bash", "/opt/bookkeeper/entrypoint.sh" ]
+# BookKeeper healthcheck was broken in 4.7.3 and was not fixed until 4.9.0
+# This overrides the healthcheck to the default BK healthcheck method
+# https://github.com/apache/bookkeeper/issues/1687
+HEALTHCHECK --interval=10s --timeout=60s CMD /opt/bookkeeper/bin/bookkeeper shell bookiesanity
```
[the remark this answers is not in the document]
adrianmo: @eolivelli this is the default healthcheck command in BK: https://github.com/apache/bookkeeper/blob/master/docker/scripts/healthcheck.sh#L30

Should we use something different?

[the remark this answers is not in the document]
eolivelli: I think it is okay, I didn't know
Sorry for late reply @adrianmo

docker/bookkeeper/entrypoint.sh
```diff
@@ -19,6 +19,7 @@ PRAVEGA_CLUSTER_NAME=${PRAVEGA_CLUSTER_NAME:-"pravega-cluster"}
 BK_CLUSTER_NAME=${BK_CLUSTER_NAME:-"bookkeeper"}
 BK_LEDGERS_PATH="/${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME}/${BK_CLUSTER_NAME}/ledgers"
 BK_DIR="/bk"
+CONTAINER_INITIALIZED="/container_initialized"
```
eolivelli: nice trick !

docker/bookkeeper/entrypoint.sh
```diff
@@ -39,13 +40,24 @@ export BK_tlsTrustStorePasswordPath=/var/private/tls/bookie.truststore.passwd
...
-/opt/bookkeeper/bin/bookkeeper shell bookieformat -nonInteractive -force -deleteCookie
+
+if [ ! -e $CONTAINER_INITIALIZED ]; then
+  # This is the first time the container is started, let's format the bookie
+  # to delete any preexistent data and metadata that can cause conflicts.
+  echo "format bookie data and metadata"
+  /opt/bookkeeper/bin/bookkeeper shell bookieformat -nonInteractive -force -deleteCookie
+  touch $CONTAINER_INITIALIZED
+else
+  # Rhe container has been restarted and there is no need to format
```
eolivelli: typo: 'Rhe'

Tristan1900:
FYI, the current code passes the e2e tests in Pravega operator.

adrianmo:
> FYI, the current code passes the e2e tests in Pravega operator.

Thanks for checking that, @Tristan1900. 

I just triggered the system tests: https://asd-nautilus-jenkins.isus.emc.com/view/Pravega/job/pravega-branch-builder/911/

adrianmo:
@fpj @ravisharda @eolivelli can you please review the PR? Thank you.

adrianmo:
The Jenkins job failed because BK pods were not able to start.

```
$ kubectl.exe logs pravega-bookie-0
/bin/bash: /opt/bookkeeper/entrypoint.sh: No such file or directory
```

In the PR, we have modified the entrypoint script of the Docker image. The entrypoint has been moved from `/opt/bookkeeper/entrypoint.sh` to `/opt/bookkeeper/scripts/pravega_entrypoint.sh`. However, the current (and old) operator version is overriding the entrypoint in https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravega/bookie.go#L93. This will be fixed in https://github.com/pravega/pravega-operator/pull/122, however, we need to make the new BK image backwards compatible with old operator versions.

I'll make the changes and push them shortly.

adrianmo:
Backwards compatibility changes made. Jenkins job in progress: https://asd-nautilus-jenkins.isus.emc.com/view/Pravega/job/pravega-branch-builder/913/

Tristan1900:
Merged master. 
The last Jenkins job failed because we are using an outdated master branch. The outdated master has a procedure to clean up BK PVC when scaling down, but the latest Operator has already addressed that issue so there is an exception `ApiException: not found` in the Pravega test. 
After merging the master, it should pass all the tests.

adrianmo:
I tested the changes with the 'update` use case in Kubernetes and it didn't work. 

When BK is updated in Kubernetes, pods will be terminated and new ones will be created, therefore, it will spawn a brand new container, but ledger and journal volumes will be reused from the previous pod. This use case was not considered before, so instead of checking if the container is started for the first time, I've updated the entrypoint to check whether or not the BK data volumes contain any data. 

If BK data is present, it means that the bookie has been migrated or updated without losing its volumes, so the bookie should just try to resume its work. Otherwise, if there's no BK data in the volumes, it will format all metadata to avoid potential conflicts and start as a new bookie.

adrianmo:
@eolivelli sorry to bother you again, but can you please have another look at the PR after the latest changes?

adrianmo:
@eolivelli @fpj any final comments before we merge the PR?

PR #3359, opened 2019-02-09
Issue 3341: Remove many Javadoc warning

### LINKED ISSUE #3341: Clean up warnings in build for pravega
Problem description
    Modules in the pravega build have a large number of warnings. 

  Pravega/pravega modules

Make modifications to eliminate the warnings - hopefully to improve build time and clarity. In a couple of cases the code examples in Javadoc were truncated.

tkaitchuck:
**Change log description**  
* Fix some issues with Javadocs
* Import Lombok when generating Javadocs to reduce errors
* Suppress minor warnings.

**Purpose of the change**  
Fixes #3341 

**What the code does**  
Removes 95% of out warnings from the `generateJavadoc` target and 100% for the `javadocs` target.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ project('common') {
             project(':shared:protocol').sourceSets.main.java
         }
         dependencies {
-            compile 'org.projectlombok:lombok:' + lombokVersion   
+            compileOnly 'org.projectlombok:lombok:' + lombokVersion   
         }
         options.addBooleanOption('html5', true)
         options.addBooleanOption("Xdoclint:none", true)
@@ -138,7 +138,7 @@ project('shared:authplugin') {
         title = "Pravega Auth API"
         failOnError = false
         dependencies {
-            compile 'org.projectlombok:lombok:' + lombokVersion   
+            compileOnly 'org.projectlombok:lombok:' + lombokVersion   
         }
         source {
             sourceSets.main.java
@@ -161,7 +161,7 @@ project ('shared:cluster') {
     javadoc {
         exclude 'io/pravega/shared/*'
         dependencies {
-            compile 'org.projectlombok:lombok:' + lombokVersion   
+            compileOnly 'org.projectlombok:lombok:' + lombokVersion   
         }
         failOnError = false
         options.addBooleanOption('html5', true)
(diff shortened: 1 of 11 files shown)


elizabethbain:
That looks so much better! Saw that the build shows as failed when javadocs fails (though it looks like it compiled and runs ok) I like the concept of being sure javadocs are correct on code checkin, but there are some more warnings that need to be fixed/ merged for clean Javadoc generation. Should we wait until we have a totally clean java doc generation before setting that Boolean for failure? Thanks Beth

elizabethbain:
Noticed I was getting errors when compiling with Java 8 in javadocs with the new gradle. I've been on Java 8 since Andrei advised earlier that pravega had problems with Java 10 and our documentation all references Java 8. The errors are in reference to the html5 option which first showed up in javadocs for Java 9.  If 10 is our official build and is what we advise using, (or even if 9 is), I'll open a PR to fix the docs references.

build.gradle
```diff
@@ -119,6 +119,11 @@ project('common') {
             project(':common').sourceSets.main.java
             project(':shared:protocol').sourceSets.main.java
         }
+        dependencies {
```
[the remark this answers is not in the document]
tkaitchuck: This dependency is being added to the Javadoc target only. It should not affect the main build or our runtime dependency requirements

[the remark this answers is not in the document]
huide9: @andreipaduroiu  #3362 is based on this one. please ignore the changes made on #3362. they're same.  sorry for the confuse.

build.gradle
```diff
@@ -888,7 +966,7 @@ task javadocs(type: Javadoc) {
 
     // Include names of any project that is to be included in the javadoc distribution
     ext.projects = [':client']
-    options.links("http://docs.oracle.com/javase/8/docs/api/");
+    options.links("http://docs.oracle.com/javase/10/docs/api/");
```
fpj: We haven't made the move to Java 10 quite yet, right? Should we update this to point to the Java 10 page?

tkaitchuck: This is just the page that will get linked to if someone clicks on the class for a return value or a parameter that is defined in the standard library (like a String). I think it is OK to link them to more current docs.

fpj: there could be new or different behavior in the newer version, no?

tkaitchuck: Yes, but we have no idea what version they are going to run their app with. If they run with 11 or 9 then either link is wrong. I am simply picking 10 because it is newer, common, and not past end of life.

PR #3364, opened 2019-02-12
Issue #3333, #2572: Keep completed segments in position objects.

### LINKED ISSUE #2572: StreamCuts generated from ReaderGroups can have missing ranges
**Problem description**
Take the following example:
epoch 0: S0 [0.0, 0.5), S1 [0.5, 1.0)
epoch 1: S2 [0.0, 0.6) , S3 [0.6, 1.0)

Readers in reader group may have read S1 completely while they may still be reading S0. 
In such cases, readers can continue reading from S3, while S2 will wait until S0 has been read completely. 

If a streamcut is generated at this point, then the stream cut will contain `(S0/offset, S3/offset)`. 
Which essentially means the key range covered in the StreamCut will be [0.0, 0.5), [0.6, 1.0). 

Such a stream cut will have a missing range of [0.5, 0.6). And it becomes hard to determine casual precedence wrt the stream cut for the missing range.
Hence it makes the stream cut inconsistent with the definition which states: 
```
 * A set of segment/offset pairs for a single stream that represent a consistent position in the
 * stream. (IE: Segment 1 and 2 will not both appear in the set if 2 succeeds 1, and if 0 appears
 * and is responsible for keyspace 0-0.5 then other segments covering the range 0.5-1.0 will also be
 * included.)
```
Note: what makes it harder here is that we do not have a segment that matches the missing keyspace. 
And current implementation of Stream Cut does not allow such partial range to be captured.
The missing range is implicit and readergroup can handle this cleanly. 
However, validation checks on partial streamcuts (those missing intermediate ranges) become harder. 

Also, with current implementation of streamcut which does not provide entire range, operations like apply streamcut on head of stream will yield segments without encompassing complete keyspace. 
**Problem location**
StreamCut

Improve streamcut implementation such that we are explicitly capturing the entire keyspace.

### LINKED ISSUE #3017: EndToEndTruncationTest.testDeleteStreamWhileReading is too slow
**Problem description**
The test EndToEndTruncationTest.testDeleteStreamWhileReading() slows down Travis builds by over two minutes. 

**Problem location**
EndToEndTruncationTest.testDeleteStreamWhileReading()

Find a way to optimize this test.

### LINKED ISSUE #3333: Problem in reader positions used by Event Processor (readerOffline)
**Problem description**
We are consistently facing test failures on branch `r0.4` (Docker Swarm). The failures normally impact multiple tests. The problem seems to be related to the Controller and how it manages a failover event within the Event Processor framework. The visible consequence of this problem is that the Controller service cannot continue reading from internal streams (`_requeststream`), so requests such as `scaleStream` or `sealStream` are not processed anymore.

Let us describe an instance of this problem recently reproduced (`ReadWriteAndAutoScaleWithFailover`):
1) In a system test build, everything works fine until `2019-01-30 16:25`. In particular, we can observe that Controller instances read from `_requestStream` as expected to process Stream operations (recall that the problem we are facing manifests as Stream operations being stalled, such as `sealStream` or `scaleStream`):
```
(Controller A)
2019-01-30 16:20:58,852 2648 [epollEventLoopGroup-2-47] DEBUG i.p.c.n.i.ClientConnectionInboundHandler - xxx processing reply: WireCommands.SegmentRead(type=SEGMENT_READ, segment=_system/_requeststream/0.#epoch.0, offset=16319, atTail=true, endOfSegment=false, data=java.nio.HeapByteBuffer[pos=0 lim=570 cap=570])
2019-01-30 16:20:58,857 2653 [epollEventLoopGroup-2-47] DEBUG i.p.c.n.i.ClientConnectionInboundHandler - xxx processing reply: WireCommands.SegmentRead(type=SEGMENT_READ, segment=_system/_requeststream/0.#epoch.0, offset=0, atTail=false, endOfSegment=false, data=java.nio.HeapByteBuffer[pos=0 lim=16319 cap=16319])
```
```
(Controller B)
2019-01-30 16:24:16,332 200128 [epollEventLoopGroup-2-66] DEBUG i.p.c.n.i.ClientConnectionInboundHandler - yyy processing reply: WireCommands.SegmentRead(type=SEGMENT_READ, segment=_system/_requeststream/0.#epoch.0, offset=21241, atTail=false, endOfSegment=false, data=java.nio.HeapByteBuffer[pos=0 lim=298 cap=298])
```
2) But around `2019-01-30 16:25`, `ReadWriteAndAutoScaleWithFailover` performs a scaledown:
```
2019-01-30 16:25:25,908 434639 [Time-limited test] INFO  i.p.t.system.AbstractFailoverTests - Scaling down to 1 controller, 1 Segment Store instance
2019-01-30 16:25:25,909 434640 [Time-limited test] INFO  i.p.t.system.AbstractFailoverTests - Stop write flag status: false, stop read flag status: false 
2019-01-30 16:25:25,909 434640 [Time-limited test] INFO  i.p.t.system.AbstractFailoverTests - Read count: 9614, write count: 10709 with Segment Store  and controller failover after sleep
```
3) From this point onwards, we start observing exceptions coming from `EventProcessorGroupImpl` trying to remove a non-existent reader (probably related to the killed controller in the scale down) from the metadata stored through `CheckpointStore`:
```
2019-01-30 16:25:29,878 2390 [controllerpool-26] ERROR i.p.c.s.e.ControllerEventProcessors - Error notifying failure of process=controller-71d8dd4b-6274-46aa-ae94-11212fb8e929 in event processor group EventProcessorGroup[abortStreamReaders]
io.pravega.controller.store.checkpoint.CheckpointStoreException: org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /pravega/pravega-cluster/eventProcessors/controller-71d8dd4b-6274-46aa-ae94-11212fb8e929/abortStreamReaders  
```
4) We also observe problems when closing a reader of the reader group of internal streams:
```
2019-01-30 16:25:28,997 2404 [controllerpool-54] WARN  i.p.c.f.ControllerClusterListener - java.lang.IllegalArgumentException: When shutting down a reader: Given position does not match the segments it was assigned: 
[_system/_requeststream/0.#epoch.0, _system/_requeststream/1.#epoch.0] 
 vs 
 [_system/_requeststream/0.#epoch.0]
```
5) From that point onwards, we do not see any further reads on `_requeststream`.
6) This is leaves the Controller service unable to further serve `sealStream` and `scaleStream` requests, which makes `ReadWriteAndAutoScaleWithFailover` and subsequent tests to fail due to timeouts.

The current status of the investigation seems to point to a mismatch in the way _reader positions_ are interpreted in Event Processor framework and the ReaderGroup. That is, Event Processor may store a position for a reader that is "before" the actual reader position that the ReaderGroup is seeing. When this position is used in `readerOffline` method, it could break a safety check for which the position provided should contain the segments for a reader that the ReaderGroup is seeing. Otherwise, the previous `IllegalArgumentException` is thrown.

**Problem location**
Event Processor reader group management.

Fix the management of reader positions in Event Processor framework to be aligned with ReaderGroup semantics.

### LINKED ISSUE #3539: EndToEndStatsTest fail occasionally
**Problem description**
Example:
```
io.pravega.test.integration.endtoendtest.EndToEndStatsTest > testStatsCount FAILED
    java.lang.AssertionError: expected:<9> but was:<10>
        at org.junit.Assert.assertEquals(Assert.java:631)
        at io.pravega.test.integration.endtoendtest.EndToEndStatsTest.testStatsCount(EndToEndStatsTest.java:117)
```

**Problem location**
EndToEndStatsTest

Use assertEvnetuallyEquals to improve robustness.

tkaitchuck:
**Change log description**  
* Keeps completed segments in position objects.
* Delays release of successors following a completed segment until a checkpoint.

**Purpose of the change**  
Fixes #3333 and fixes #2572
Currently the segments returned in the Position objects are inconsistent and following a scaling event can be rejected by the `readerOffline` method. This change resolves that by making sure that sealed segments are included in the position object and held until a checkpoint.

Also Incidentally fixes #3017 and fixes #3539

**What the code does**  
Per the comment in #3333, changes the way that reader releases segments. Instead of handling a completed segment right away, it is kept in a map, (whose entries are included in subsequent getPosition() calls) and then only invokes updates the reader group state to reflect the fact that the segment is completed upon checkpoint.

**How to verify it**  
In addition to all the usual tests, careful review of the state machine should be performed, as this has been a difficult area.


client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -168,24 +170,39 @@ private String updateGroupStateIfNeeded() throws ReaderNotInReaderGroupException
...
+            acquireSegmentsIfNeeded();
+            return null;
         }
-        acquireSegmentsIfNeeded();
-        return null;
     }
 
     @GuardedBy("readers")
     private void releaseSegmentsIfNeeded() throws ReaderNotInReaderGroupException {
+        for (Iterator<Segment> iterator = sealedSegments.iterator(); iterator.hasNext();) {
```
fpj: This is enforcing this invariant, yes?

```
No portion of keyspace that is owned by one reader will
be read by another without the first reader checkpointing
first.
```

client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
```diff
@@ -16,15 +16,10 @@
 import java.util.Set;
 
 /**
- * A position has two components
- * 1. ownedSegments -- segments that can be read currently. Each ownedSegment also has an offset indicating the
+ * A position has ownedSegments -- segments that can be read currently. Each ownedSegment also has an offset indicating the
```
fpj: We need to qualify owned segment with respect to completed segments, the segments for which this reader has reached the end but not released yet. Concretely, does `ownedSegments` include completed segments?

client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
```diff
@@ -16,15 +16,10 @@
...
  * point until which events have been read from that segment. Completely read segments have offset of -1.
- * 2. futureOwnedSegments -- segments that can be read after one of the currently read segment is completely read. Each
- * segment in this set has exactly one previous segment that belongs to the set ownedSegments.
  * <p>
- * Well-formed position object. A position is called well-formed iff the following hold.
- * 1. for each segment s in futureOwnedSegment, s.previous belongs to ownedSegments and s.previous.offset != -1
- * 2. for each segment s in ownedSegment, s.previous does not belongs to ownedSegments
+ * Well-formed position object. A position is called well-formed iff each segment s in ownedSegment, s.previous does not belongs to ownedSegments
```
fpj: This is trying to express this invariant, yes?

```
Obtaining a Position at any given point will return
a segment for all of the keyspace owned by the reader.
```

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -151,31 +147,40 @@ void close() {
...
         AtomicBoolean reinitRequired = new AtomicBoolean(false);
-        sync.updateState((state, updates) -> {
+        boolean result = sync.updateState((state, updates) -> {
             if (!state.isReaderOnline(readerId)) {
                 reinitRequired.set(true);
             } else {
                 log.debug("Marking segment {} as completed in reader group. CurrentState is: {}", segmentCompleted, state);
                 reinitRequired.set(false);
-                updates.add(new SegmentCompleted(readerId, segmentCompleted, segmentToPredecessor));
+                if (state.getCheckpointForReader(readerId) == null) {
```
fpj: I'm not sure I understand this predicate. Is it checking whether there is an outstanding checkpoint for `readerId`? But, aren't we supposed to release completed segments (we are in the `handleEndOfSegment` method of the reader groups state manager) strictly after completing a checkpoint?

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
tkaitchuck: There are two parts to that invariant:
The `handleEndOfSegment` method 2 lines down enforces that no checkpoints are pending. 
The fact that a checkpoint was just performed is enforced by this method's one and only caller.

client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
tkaitchuck: Yes. Added a comment.

client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
tkaitchuck: yes. And the invariant that it will not contain multiple segments for the same portion of keyspace.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
tkaitchuck: This is guarding against the possible race condition of a second checkpoint being created at the last moment.

tkaitchuck:
>I have a few questions. Also, I don't understand this invariant:
> ```
> If a reader dies and a position is specified from any point after
> the last checkpoint, the checkpoint will contain a subset of
> the keyspace in the group state
> ```

This is enforced by not allowing any segments to be added or removed from the reader's position until a checkpoint, and only updating the group state upon checkpoint boundaries. So all position objects must contain exactly the same segments that were held by the reader at the time the checkpoint was initiated.

client/src/main/java/io/pravega/client/stream/notifications/notifier/EndOfDataNotifier.java
```diff
@@ -29,6 +29,11 @@ public EndOfDataNotifier(final NotificationSystem notifySystem,
                              final ScheduledExecutorService executor) {
         super(notifySystem, executor, synchronizer);
     }
+    
+    @VisibleForTesting
+    public void pollNow() {
```
fpj: Why do you need this in the tests? Is this related to the checkpoint guarantees somehow?

fpj:
> > I have a few questions. Also, I don't understand this invariant:
> > ```
> > If a reader dies and a position is specified from any point after
> > the last checkpoint, the checkpoint will contain a subset of
> > the keyspace in the group state
> > ```
>
> This is enforced by not allowing any segments to be added or removed from the reader's position until a checkpoint, and only updating the group state upon checkpoint boundaries. So all position objects must contain exactly the same segments that were held by the reader at the time the checkpoint was initiated.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupState.java
```diff
@@ -536,13 +536,9 @@ void update(ReaderGroupState state) {
                 while (iter.hasNext()) {
                     Entry<Segment, Long> entry = iter.next();
                     Segment segment = entry.getKey();
-                    Long offset;
-                    if (ownedSegments == null || ownedSegments.isEmpty()) {
+                    Long offset = ownedSegments.get(segment);
```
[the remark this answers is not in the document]
tkaitchuck: No. I removed the code path that was passing null, it is now passing an empty map.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -92,8 +92,8 @@
             nanoClock = System::nanoTime;
         }
         releaseTimer = new TimeoutTimer(TIME_UNIT, nanoClock);
-        acquireTimer = new TimeoutTimer(TIME_UNIT, nanoClock);
-        fetchStateTimer = new TimeoutTimer(TIME_UNIT, nanoClock);
+        acquireTimer = new TimeoutTimer(Duration.ZERO, nanoClock);
```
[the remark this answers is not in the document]
tkaitchuck: Yes it is for initialization. It doesn't matter for production, but it slows down the unit tests a lot if they have to wait 1s for the first possible checkpoint.

client/src/test/java/io/pravega/client/batch/impl/BatchClientImplTest.java
```diff
@@ -171,16 +171,6 @@ private Stream createStream(String scope, String streamName, int numSegments, Mo
     private MockConnectionFactoryImpl getMockConnectionFactory(PravegaNodeUri location) throws ConnectionFailedException {
         MockConnectionFactoryImpl connectionFactory = new MockConnectionFactoryImpl();
         ClientConnection connection = mock(ClientConnection.class);
-        Mockito.doAnswer(new Answer<Void>() {
```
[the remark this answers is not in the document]
tkaitchuck: This was a common pattern so I embedded this into the mock. This is related to the 'false' flag passed above.

client/src/test/java/io/pravega/client/byteStream/ByteStreamReaderTest.java
```diff
@@ -44,19 +38,7 @@
     public void setup() throws ConnectionFailedException {
         PravegaNodeUri endpoint = new PravegaNodeUri("localhost", 0);
         connectionFactory = new MockConnectionFactoryImpl();
-        ClientConnection connection = Mockito.mock(ClientConnection.class);
```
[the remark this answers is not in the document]
tkaitchuck: See above

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -151,31 +147,41 @@ void close() {
...
+     * completed segment. To ensure consistent checkpoints, a segment cannot be released while a
+     * checkpoint for the reader is pending, so it may or may not succeed.
+     * 
+     * @return true if the completed segment was released successfully.
      */
-    void handleEndOfSegment(Segment segmentCompleted, boolean fetchSuccesors) throws ReaderNotInReaderGroupException {
+    boolean handleEndOfSegment(Segment segmentCompleted) throws ReaderNotInReaderGroupException {
         final Map<Segment, List<Long>> segmentToPredecessor;
-        if (fetchSuccesors) {
+        if (sync.getState().getEndSegments().containsKey(segmentCompleted)) {
```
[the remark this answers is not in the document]
tkaitchuck: This is the case where the reader has reached the endStreamCut. In this case there are no successors to add, as it is not going to read further.

client/src/test/java/io/pravega/client/stream/mock/MockController.java
```diff
@@ -71,6 +71,7 @@
     @GuardedBy("$lock")
     private final Map<Stream, StreamConfiguration> createdStreams = new HashMap<>();
     private final Supplier<Long> idGenerator = new AtomicLong(0)::incrementAndGet;
+    private final boolean callServer;
```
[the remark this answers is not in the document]
tkaitchuck: Before it defaulted to true, but most of the tests calling it needed to have mocks to stub out this behavior even though they weren't testing it. See my replies above.

client/src/main/java/io/pravega/client/stream/notifications/notifier/EndOfDataNotifier.java
tkaitchuck: This allows the tests to avoid depending on timing. Before they were sleeping and simply assuming the background thread would run if they waited some number of seconds. By invoking the method in the test's thread it's much faster, and no longer prone to flakes.

tkaitchuck:
@fpj I added some more details in the invariants listed in the issue comment.

client/src/test/java/io/pravega/client/stream/notifications/SegmentNotifierTest.java
```diff
@@ -47,7 +44,7 @@
     @Spy
     private NotificationSystem system = new NotificationSystem();
     @Spy
-    private ScheduledExecutorService executor = new InlineExecutor();
+    private ScheduledExecutorService executor = ExecutorServiceHelpers.newScheduledThreadPool(1, "SegmentNotifierTest");
```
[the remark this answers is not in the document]
tkaitchuck: Undone

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -114,8 +116,7 @@
                 try {
                     buffer = segmentReader.read(waitTime);
                 } catch (EndOfSegmentException e) {
-                    boolean fetchSuccessors = e.getErrorType().equals(ErrorType.END_OF_SEGMENT_REACHED);
-                    handleEndOfSegment(segmentReader, fetchSuccessors);
+                    handleEndOfSegment(segmentReader);
```
shrids: The default automaticCheckpoint interval is 120 seconds (`ReaderGroupConfig#automaticCheckpointIntervalMillis`), this would imply that reader would not read successor segments until the checkpoint has completed. (i.e. the reader can only acquire newer unassigned segments, that belong to a different keyspace until the checkpoint completes) . This would cause a slow down of the reader during a scale up of a stream. Is my understanding correct here?

tkaitchuck: yes. Good point. I will lower the default interval.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -168,24 +170,35 @@ private String updateGroupStateIfNeeded() throws ReaderNotInReaderGroupException
...
+        if (checkpoint != null) {
+            log.info("{} at checkpoint {}", this, checkpoint);
+            if (groupState.isCheckpointSilent(checkpoint)) {
                 // Checkpoint the reader immediately with the current position. Checkpoint Event is not generated.
-                groupState.checkpoint(atCheckpoint, getPosition());
-                atCheckpoint = null;
+                groupState.checkpoint(checkpoint, getPosition());
+                if (atCheckpoint != null) {
+                    //In case the silent checkpoint held up releasing segments
+                    releaseSegmentsIfNeeded();
```
shrids: if `atCheckpoint` is not null then releaseSegmentIfNeeded() will be invoked at line 171.  Did not understand the need to invoke it again.

tkaitchuck: The releaseSegmentIfNeeded() at 171 will not release any segments if a checkpoint is pending, so that the checkpoint can be performed first. In the case that this is a normal checkpoint this is fine, but in the case that it is a silent checkpoint this is not desirable. 

I understand that that is confusing. So I am changing this to handled internally to relaeasSegment(). That way these sort of operations are not affected by silent checkpoints.

fpj:
I'm not sure I understand the code coverage report. There are lots of files that it says that it doesn't track, and I don't have it clear whether the diff hit takes these files into account or not. Why doesn't it track all those files in the first place?

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndTruncationTest.java
```diff
@@ -547,22 +573,18 @@ public void testSegmentTruncationWhileReading() throws InterruptedException {
...
      * client is reading events (Segment Store) the test deletes the Stream (Controller and metadata). Once the client
      * reads all the events and reaches the end of segment, it contacts the Controller to retrieve subsequent segments
-     * (if any). However, the Stream-related metadata to answer this request has been previously deleted. The current
-     * behavior is that the Controller keeps looking for the Stream metadata for a non-negligible period of time (+100
-     * sec.) while the client waits for a response. After this period, a RetriesExhaustedException is thrown to the
-     * client, which is expected in this situation.
+     * (if any). However, the Stream-related metadata to answer this request has been previously deleted.
      */
-    @Test(timeout = 200000)
+    @Test(timeout = 20000)
```
RaulGracia: Is this addressing #3017, right?

test/integration/src/test/java/io/pravega/test/integration/StreamCutsTest.java
```diff
@@ -125,18 +130,21 @@ public void testReaderGroupCuts() throws Exception {
         ReaderGroupManager groupManager = new ReaderGroupManagerImpl("test", controller, clientFactory,
                 connectionFactory);
         groupManager.createReaderGroup("cuts", ReaderGroupConfig
-                .builder().disableAutomaticCheckpoints().stream("test/test").build());
+                .builder().disableAutomaticCheckpoints().stream("test/test").groupRefreshTimeMillis(0).build());
```
RaulGracia: Specifying `groupRefreshTimeMillis(0)` when using `disableAutomaticCheckpoints()` is done to speed up the test, right?

tkaitchuck: yes. It also makes it deterministic. Instead of relying on a clock and sleeping, it will check for updates to the state every time.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -51,7 +51,7 @@
 
    public static class ReaderGroupConfigBuilder implements ObjectBuilder<ReaderGroupConfig> {
        private long groupRefreshTimeMillis = 3000; //default value
-       private long automaticCheckpointIntervalMillis = 120000; //default value
+       private long automaticCheckpointIntervalMillis = 30000; //default value
```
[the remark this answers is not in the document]
tkaitchuck: The reader group state won't expand, rather the readers will simply stall at the first scaling event until a checkpoint is performed manually.

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndTruncationTest.java
tkaitchuck: Yes, added to description.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
tkaitchuck: There is a bound on the number of concurrent checkpoints, so checkpointing that frequently will almost certainly throw an exception. In general a reader is not going to notice a checkpoint any sooner than the refresh interval which is by default 3 seconds. This of course can also be adjusted, but it seems unlikely a checkpoint would be able to reach all readers and get the results in under 100ms. However even attempting to achieve this is probably not a good idea as it would impose significant overhead, which would impact throughput and waste resources.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -121,7 +117,7 @@ public void tearDown() throws Exception {
     public void testSegmentNotifications() throws Exception {
         final String streamName = "stream1";
         StreamConfiguration config = StreamConfiguration.builder()
-                                                        .scalingPolicy(ScalingPolicy.byEventRate(10, 2, 1))
+                                                        .scalingPolicy(ScalingPolicy.fixed(1))
```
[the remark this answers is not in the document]
tkaitchuck: This test is manually scaling the stream and asserting that the notifications for the manual scaling are delivered as expected. Enabling automatic scaling on the same stream, just leaves open an opportunity for the controller to scale the stream and cause the test to fail.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -198,17 +211,35 @@ private void releaseSegmentsIfNeeded() throws ReaderNotInReaderGroupException {
...
     private void acquireSegmentsIfNeeded() throws ReaderNotInReaderGroupException {
         Map<Segment, Long> newSegments = groupState.acquireNewSegmentsIfNeeded(getLag());
         if (!newSegments.isEmpty()) {
             log.info("{} acquiring segments {}", this, newSegments);
             for (Entry<Segment, Long> newSegment : newSegments.entrySet()) {
-                final EventSegmentReader in = inputStreamFactory.createEventReaderForSegment(newSegment.getKey(),
-                        groupState.getEndOffsetForSegment(newSegment.getKey()));
-                in.setOffset(newSegment.getValue());
-                readers.add(in);
+                if (newSegment.getValue() < 0) {
```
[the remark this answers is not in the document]
tkaitchuck: Yes

client/src/test/java/io/pravega/client/stream/impl/EventStreamReaderTest.java
```diff
@@ -414,12 +440,235 @@ public void testTruncatedSegmentDeleted() throws Exception {
...
+        InOrder inOrder = Mockito.inOrder(groupState, segmentInputStream);
         // Validate that TruncatedDataException is thrown.
         AssertExtensions.assertThrows(TruncatedDataException.class, () -> reader.readNextEvent(0));
+        inOrder.verify(groupState).getCheckpoint();
         // Ensure this segment is closed.
-        Mockito.verify(segmentInputStream, Mockito.times(1)).close();
-        // Ensure groupstate is updated to handle end of segment.
-        Mockito.verify(groupState, Mockito.times(1)).handleEndOfSegment(segment, true);
+        inOrder.verify(segmentInputStream, Mockito.times(1)).close();      
+        // Ensure groupstate is updated not updated before the checkpoint.
```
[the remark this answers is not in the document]
tkaitchuck: Fixed

client/src/test/java/io/pravega/client/stream/impl/EventStreamReaderTest.java
```diff
@@ -414,12 +440,235 @@ public void testTruncatedSegmentDeleted() throws Exception {
...
+        inOrder.verify(groupState, Mockito.times(0)).handleEndOfSegment(segment);
+        Mockito.when(groupState.getCheckpoint()).thenReturn("Foo").thenReturn(null);
+        EventRead<byte[]> event = reader.readNextEvent(0);
+        assertTrue(event.isCheckpoint());
+        assertEquals("Foo", event.getCheckpointName());
+        inOrder.verify(groupState).getCheckpoint();
+        // Verify groupstate is updated not updated before the checkpoint, but is after.
```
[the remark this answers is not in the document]
tkaitchuck: fixed

fpj:
@tkaitchuck I'm not comfortable merging this PR. It is causing a lot of intermittent failures in the travis builds (integration tests). I can't directly point to the changes here, but the observation is that I'm seeing a lot more failures here than with other PRs. Can you please debug this locally?

PR #3375, opened 2019-02-13
Issue 3319: Micrometer metrics migration

### LINKED ISSUE #3319: Migrate Pravega metrics into Micrometer based to support dimensional metric data model
**Problem description**
Pravega application metrics needs to support dimensional data model for metrics back end to take full advantage of rich queries crossing dimensions and measurements.
With the support of dimensional data model, Pravega will be able to talk to time series database directly without local agent such as Telegraf.

**Problem location**
Replace the existing readytalk statsD and codahale libraries, and the application codes based on them.
Replace all the application codes calling method to flat metric name with dimensions, to use the new API with tag support.

Replace existing readytalk and codahale with Micrometer to support dimensional data model;

This umbrella issue covers issues:

- [ ] Update statsd library to send the correct metric type (#3004) 
- [ ] Use Micrometer library for Pravega metrics(#2977) 
- [ ] Enable Configurable Stats Providers in Bookkeper docker container (#3101) 
- [ ] #3405 Add tags to those metrics defined by the Platform team

kevinhan88:

**Change log description**  
Changed to micrometer based implementation to support dimensional data model.

**Purpose of the change**  
Fixes #3319 

**What the code does**  
Replaced the entire underlying metric libraries to micrometer, and refactored all the metric classes and test codes.
. Added Micrometer core, statsd and influx libraries.
. Added varArgument "String... tags" to all the metric lookup APIs
. Added InfluxDB config so that Pravega can talk to InfluxDB directly.
. Added "nameFromTags" utility method to form metric lookup key using tags
. Removed dec() from Counter API since it is obviously anti-principal anti-pattern.
. Replaced codahale system metrics with Micrometer's implementation
. Replaced getName() of Metric base class with getId(), because Codahale lookup is String based but Micrometer lookup is Id based.
. Align OpStatsDataPERCENTILESET with Micrometer's percentile set.
. Refactored CounterImpl, GaugeImpl, MeterImpl and TimerImpl to be Micrometer based.
. Fixed FileSystemStorageTest as it didn't bind the metric register to a concrete storage. 

**What the code does not do**
All the essential API changes have been baked in this PR, but no tags have been added into metrics.
Following this PR, there is going to be another PR to make all the metrics changes, initially based on the telegraf.conf provided by the Platform team.
For example, for metric "container.failover" with dimension "container", today we call nameFromContainer("container.failover", 6) to construct a flat metric name "container.failover.6", and then call incCounterValue("container.failover.6", 1) - with this PR everything works as before. 

With the next PR, we no longer construct the flat name, instead we call incCounterValue("container.failover", 1, "containerId", "6") - note the tag pair(s) added.

**How to verify it**  
This is the first step of the refactoring, which is to replace the underlying libraries.
Because the change is backwards compatible, all the existing hierarchical metrics should still work.
After step one, telegraf agent should work the same way as before. However it is not ready to talk to InfluxDB directly.

The second step is to add tags into the required metrics (according to the Platform team's mapping file). 
After this step, the metric should be able to talk directly to InfluxDB and Telegraf agent.


shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +229,27 @@ public static String failMetricName(String metricName) {
...
+    /**
+     * Generate a dot separated combination of metric name and tag value[s].
+     * e.g. metric "pravega.segmentstore.container.append_count" with tag "container=6" will get:
+     * "pravega.segmentstore.container.append_count.6".
+     *
+     * @param metric the metric name.
+     * @param tags the tag(s) associated with the metric.
+     * @return the dot separated combination of metric name and tag value(s).
+     */
+    public static String nameFromTags(String metric, String... tags) {
```
EronWright: II believe that Micrometer has this same logic, why would you need to do this here?  See [Concepts -> Naming meters](http://micrometer.io/docs/concepts#_naming_meters).

kevinhan88:
Yes, this PR is the make the API ready, and the change is fully backwards compatible.
The following PR will add tags(dimensions) into metrics based on the telegraf.conf file from the Platform team.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: This is to keep the backwards compatibility. 
Micrometer does generate id which is a combination of metric name and tags, something like "pravega.segmentstore.container.append_count.containerid=6" (note both tag key and tag value are showing on id); however, our existing name convention is to append tag value only to metric name, something like: "pravega.segmentstore.container.append_count.6".

EronWright: Simply put, `nameFromTags` should not be used for InfluxDB; the metric name should be `pravega.segmentstore.container.append_count` in the InfluxDB scenario, because that is the effective name that the (obsolete) telegraf script would have produced.

kevinhan88: Additionally, in order not to keep strong reference to metric objects, Pravega always use String key to lookup a metric object; so we have to always find a way to form a unique key for a metric object.

Micrometer holds weak reference to metric object, and encourage application code to hold direct reference to metric object, which gives two benefits:
By application holding direct strong reference to metric objects, there is no need to cache metric object;
By metrics registry holding weak reference to metric objects, there is no need to manage the life-cycle of metric objects separately (like the eviction we do today), because metric objects stay with application objects, their life-cycle should be managed naturally together.

We'll probably re-factoring the metrics further in the future to make full advantage of Micrometer.

kevinhan88: Yes Eron you're right, when we still use flat name (which needs telegraf agent to map it back to dimensional metrics", the flat name is like this: "pravega.segmentstore.container.append_count.6.Counter"

With next PR to convert flat metric to dimensional, it will be like this:
"pravega.segmentstore.container.append_count", with a tag "containerid=6".

There are two separate keys: one is lookup key used by application to locate a metric object; another is the register key ("effective key" in your term) used by metric registry.
We always need nameFromTags() to form application lookup key; however, when there's tag available, we only use the original key (effective key) to register metric, lookup key is not used.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Mon Feb 04 17:21:02 PST 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
```
fpj: I don't understand this change, seems gratuitous.

shared/metrics/src/main/java/io/pravega/shared/metrics/Counter.java
```diff
@@ -25,11 +24,6 @@
      */
     void inc();
 
-    /**
-     * Decrement the value associated with this stat.
-     */
-    void dec();
```
fpj: Any particular reason why we are removing decrements?

shared/metrics/src/main/java/io/pravega/shared/metrics/Metric.java
```diff
@@ -5,21 +5,23 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
```
fpj: Do not reformat the license header.

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
```diff
@@ -27,18 +29,18 @@
...
 
-    OpStatsLoggerImpl(MetricRegistry metricRegistry, String baseName, String statName) {
+    OpStatsLoggerImpl(MeterRegistry metricRegistry, String baseName, String statName, String... tags) {
         this.metricRegistry = Preconditions.checkNotNull(metricRegistry, "metrics");
-        this.successName = name(baseName, statName);
-        this.failName = name(baseName, failMetricName(statName));
-        this.success = this.metricRegistry.timer(this.successName);
-        this.fail = this.metricRegistry.timer(this.failName);
+        this.successName = baseName + "." + statName;
+        this.failName = baseName + "." + failMetricName(statName); //TODO: replace the original name() method
```
fpj: Make sure to incorporate the guidelines in your code changes:

https://github.com/pravega/pravega/wiki/Contributing#todos

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
```diff
@@ -90,15 +92,22 @@ public void clear() {
...
+        long numSuccess = success.count();
+        HistogramSnapshot snapshot = success.takeSnapshot();
+        double avgLatencyMillis = success.mean(TimeUnit.SECONDS);
 
         EnumMap<OpStatsData.Percentile, Long> percentileLongMap  =
-                new EnumMap<OpStatsData.Percentile, Long>(OpStatsData.Percentile.class);
+                new EnumMap<>(OpStatsData.Percentile.class);
+        ValueAtPercentile[] values = snapshot.percentileValues();
+
+        //TODO: refactor OpStatsData.Percentile based on micrometer implementation
```
fpj: Another TODO instance:

https://github.com/pravega/pravega/wiki/Contributing#todos

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
```diff
@@ -90,15 +92,22 @@ public void clear() {
...
         for (OpStatsData.Percentile percent : OpStatsData.PERCENTILESET) {
-            percentileLongMap.put(percent, (long) s.getValue(percent.getValue() / 100));
+            for (ValueAtPercentile vp : snapshot.percentileValues()) {
+                if (vp.percentile() == percent.getValue() / 100) {
+                    percentileLongMap.put(percent, (long) vp.value()); //TODO: change O(N*N) to O(N) after refactoring
```
fpj: Another TODO instance:

https://github.com/pravega/pravega/wiki/Contributing#todos

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -5,17 +5,19 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
```
fpj: Do not reformat the license header.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -77,100 +79,104 @@ public StatsLogger createScopeLogger(String scope) {
         if (0 == basename.length()) {
             scopeName = scope;
         } else {
-            scopeName = name(basename, scope);
+            scopeName = basename + "." + scope;
```
fpj: Encapsulating the logic in a function is better as we don't have to go looking for instances to change in the case the format changes. I'm not sure why you are making this change here from calling `name` to formatting directly, but I'd suggest that you encapsulate the logic in a function. I'm also wondering whether that should be in `NameUtils`:

https://github.com/pravega/pravega/blob/679c7b7c35dd112977815de5abebc6e077ecc044/shared/protocol/src/main/java/io/pravega/shared/NameUtils.java

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -77,100 +79,104 @@ public StatsLogger createScopeLogger(String scope) {
...
         }
         return new StatsLoggerImpl(metrics, scopeName);
     }
 
     private class CounterImpl implements Counter {
-        private final com.codahale.metrics.Counter counter;
+        private io.micrometer.core.instrument.Counter counter; //TODO: mutability issue to be discussed
```
fpj: Another `TODO` instance.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +229,27 @@ public static String failMetricName(String metricName) {
...
+        if (tags == null || tags.length == 0) {
+            return metric;
+        }
+        if (tags.length % 2 == 1) {
```
andreipaduroiu: Use `Preconditions.checkArgument`

shared/metrics/src/main/java/io/pravega/shared/metrics/Counter.java
andreipaduroiu: I don't think we ever use them

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +229,27 @@ public static String failMetricName(String metricName) {
...
+            throw new IllegalArgumentException("tags size must be even, it is a set of key=value pairs");
+        }
+        StringBuilder sb = new StringBuilder(metric);
```
andreipaduroiu: From what I can tell below, this method is used extensively throughout the code, everytime we report something, and sometimes on the hot path. As such, we need to make this as efficient as possible. 

In this method you inefficiently use StringBuilder to append a number of different strings, then run the result through the extremely inefficient `escapeSpecialChar` method (just open it up to see what it does).

I recomment we make this O(n) with a single pass through.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -9,7 +9,7 @@
  */
 package io.pravega.shared.metrics;
 
-import com.codahale.metrics.MetricRegistry;
+import io.micrometer.core.instrument.MeterRegistry;
```
andreipaduroiu: Did we completely remove `codehale`? If so, shall we remove the gradle dependency?

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -80,25 +80,25 @@ public void onRemoval(RemovalNotification<String, Gauge> removal) {
...
             build();
     }
 
     @Override
-    public void incCounterValue(String name, long delta) {
+    public void incCounterValue(String name, long delta, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
         Preconditions.checkNotNull(delta);
-        String counterName = name + ".Counter";
+        String counterName = nameFromTags(name, tags) + ".Counter";
```
andreipaduroiu: This is yet another example of inefficient code on the hot path. You are concatenating two strings, which requires a new allocation and memory copy. Consider updating `nameFromTags` to accept additional args to format the string from the first pass.

andreipaduroiu: Also you only need this variable in one place, and subject to a condition. Please inline this in the `createCounter` call below (line 101).

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -108,72 +108,77 @@ public Counter call() throws Exception {
     }
 
     @Override
-    public void updateCounterValue(String name, long value) {
+    public void updateCounterValue(String name, long value, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
-        String counterName = name + ".Counter";
+        String counterName = nameFromTags(name, tags) + ".Counter";
```
andreipaduroiu: Same comments as in the method above.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -108,72 +108,77 @@ public Counter call() throws Exception {
...
         countersCache.put(name, counter);
     }
 
     @Override
-    public void freezeCounter(String name) {
-        String counterName = name + ".Counter";
+    public void freezeCounter(String name, String... tags) {
+        String counterName = nameFromTags(name, tags) + ".Counter";
+        Counter counter = countersCache.getIfPresent(counterName);
+        metrics.remove(counter.getId());
```
andreipaduroiu: What happens is counter is not present? This will be null, right?

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -108,72 +108,77 @@ public Counter call() throws Exception {
...
 
     @Override
-    public <T extends Number> void reportGaugeValue(String name, T value) {
+    public <T extends Number> void reportGaugeValue(String name, T value, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
         Preconditions.checkNotNull(value);
         Gauge newGauge = null;
-        String gaugeName = name + ".Gauge";
+        String gaugeCacheKey = nameFromTags(name, tags) + ".Gauge";
+        String gaugeRegisterKey = (tags == null || tags.length == 0) ? gaugeCacheKey : name;
```
andreipaduroiu: This conditional pattern is used throughout this file. Can you abstract it out?

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -108,72 +108,77 @@ public Counter call() throws Exception {
...
         }
     }
 
     @Override
-    public void freezeGaugeValue(String name) {
-        String gaugeName = name + ".Gauge";
+    public void freezeGaugeValue(String name, String... tags) {
+        String gaugeName = nameFromTags(name, tags) + ".Gauge";
+        Gauge gauge = gaugesCache.getIfPresent(gaugeName);
+        metrics.remove(gauge.getId());
```
andreipaduroiu: null?

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -108,72 +108,77 @@ public Counter call() throws Exception {
...
-        metrics.remove(gaugeName);
     }
 
     @Override
-    public void recordMeterEvents(String name, long number) {
+    public void recordMeterEvents(String name, long number, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
         Preconditions.checkNotNull(number);
-        String meterName = name + ".Meter";
+        String meterName = nameFromTags(name, tags) + ".Meter";
```
andreipaduroiu: Inline this below.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
andreipaduroiu: On third thought, doesn't `nameFromTags(name, tags)` return `name` if `tags` is null or empty? So your conditional at line 101 will always return `name`.

shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsProvider.java
```diff
@@ -9,21 +9,15 @@
...
-import com.codahale.metrics.Metric;
-import com.codahale.metrics.MetricRegistry;
-import com.google.common.annotations.VisibleForTesting;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
 
 public class MetricsProvider {
 
-    static final MetricRegistry METRIC_REGISTRY = new MetricRegistry();
+    static final MeterRegistry METRIC_REGISTRY = Metrics.globalRegistry;
```
andreipaduroiu: Why do you have a static final member that is just an instance of another static final member of another class?

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
```diff
@@ -27,18 +29,18 @@
     private final String successName;
     private final Timer fail;
     private final String failName;
-    private final MetricRegistry metricRegistry;
+    private final MeterRegistry metricRegistry;
```
andreipaduroiu: `metricRegistry` -> `meterRegistry`

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
```diff
@@ -90,15 +92,22 @@ public void clear() {
...
         EnumMap<OpStatsData.Percentile, Long> percentileLongMap  =
-                new EnumMap<OpStatsData.Percentile, Long>(OpStatsData.Percentile.class);
+                new EnumMap<>(OpStatsData.Percentile.class);
+        ValueAtPercentile[] values = snapshot.percentileValues();
+
+        //TODO: refactor OpStatsData.Percentile based on micrometer implementation
         for (OpStatsData.Percentile percent : OpStatsData.PERCENTILESET) {
-            percentileLongMap.put(percent, (long) s.getValue(percent.getValue() / 100));
+            for (ValueAtPercentile vp : snapshot.percentileValues()) {
+                if (vp.percentile() == percent.getValue() / 100) {
```
andreipaduroiu: store `percent.getValue()/100` into a variable and use it here. Otherwise you'll end up calculating this over and over again in every loop iteration.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
andreipaduroiu: Also make this `final`

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -77,100 +79,104 @@ public StatsLogger createScopeLogger(String scope) {
...
         public void close() {
-            metrics.remove(this.name);
+            metrics.remove(this.id);
         }
 
     }
 
     private class MeterImpl implements Meter {
-        private final com.codahale.metrics.Meter meter;
+        private io.micrometer.core.instrument.DistributionSummary summary; //TODO: mutable, strong reference
```
andreipaduroiu: 1. `final`
2. Remove `TODO` or file an issue and log it here.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
```diff
@@ -47,17 +40,10 @@
...
-        // and re-insert.
-        MemoryUsageGaugeSet memoryGaugeNames = new MemoryUsageGaugeSet();
-        GarbageCollectorMetricSet gcMetricSet = new GarbageCollectorMetricSet();
-
-        memoryGaugeNames.getMetrics().forEach((key, value) -> metrics.remove(key));
-        gcMetricSet.getMetrics().forEach((key, value) -> metrics.remove(key));
-
-        metrics.registerAll(new MemoryUsageGaugeSet());
-        metrics.registerAll(new GarbageCollectorMetricSet());
+        new JvmMemoryMetrics().bindTo(metrics);
```
andreipaduroiu: What are this dangling objects? Do they require cleanup?

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
```diff
@@ -66,84 +52,93 @@ public void start() {
...
-                log.warn("ganglia create failure: {}", e);
-            }
-        }
-        if (conf.isEnableConsoleReporter()) {
-            log.info("Configuring console reporter");
-            reporters.add(ConsoleReporter.forRegistry(getMetrics())
-                .convertRatesTo(TimeUnit.SECONDS)
-                .convertDurationsTo(TimeUnit.MILLISECONDS)
-                .build());
+            StatsdConfig config = new StatsdConfig() {
```
andreipaduroiu: Please make this a separate, private static class. You've inlined 2 such classes in one method, making it much harder to read. 


Same below.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
```diff
@@ -66,84 +52,93 @@ public void start() {
...
+                @Override
+                public String get(String k) {
+                    return null;
+                }
+            };
+
+            MeterRegistry registry = new InfluxMeterRegistry(config, Clock.SYSTEM);
+            Metrics.addRegistry(registry);
         }
+        //TODO: add more registries
```
andreipaduroiu: Dangling TODO.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
```diff
@@ -66,84 +52,93 @@ public void start() {
...
-            try {
-                r.report();
-                r.stop();
-            } catch (Exception e) {
-                log.error("Exception report or stop reporter", e);
-            }
-        }
-
-        metrics.removeMatching(MetricFilter.ALL);
+        metrics.close();
```
andreipaduroiu: Why do you close `metrics`? You do no own this instance; it's a reference to an external, static field!

In `close` you only need to clean up your own resources, not external resources. For example, you should remove the recorded registries that you added.

shared/metrics/src/test/java/io/pravega/shared/metrics/MetricRegistryUtils.java
```diff
@@ -9,38 +9,30 @@
...
-    public static Meter getMeter(String metricsName) {
-        return (Meter) getMetric(metricsName);
+    public static Counter getCounter(String metricsName, String... tags) {
+        return Metrics.globalRegistry.find(metricsName).tags(tags).counter();
     }
 
-    public static Gauge getGauge(String metricsName) {
-        return (Gauge) getMetric(metricsName);
+    public static DistributionSummary getMeter(String metricsName, String... tags) {
+        return Metrics.globalRegistry.find(metricsName).tags(tags).summary();
```
andreipaduroiu: Any chance some of this calls will return `null` along the way?

andreipaduroiu: Same everywhere below.

shared/metrics/src/test/java/io/pravega/shared/metrics/StatsLoggerProxyTest.java
```diff
@@ -139,14 +139,18 @@ private TestMetric create(String name) {
         }
     }
 
+    //TODO: get rid of this class.
```
andreipaduroiu: So, are you going to get rid of it or not?

bindings/src/test/java/io/pravega/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -45,6 +46,8 @@ public void setUp() throws Exception {
         this.baseDir = Files.createTempDirectory("test_nfs").toFile().getAbsoluteFile();
         MetricsConfig metricsConfig = MetricsConfig.builder().with(MetricsConfig.ENABLE_STATISTICS, true).build();
         MetricsProvider.initialize(metricsConfig);
+        StatsProvider statsProvider = MetricsProvider.getMetricsProvider();
```
andreipaduroiu: Why modifying this test class? Or better said, why modify only this class and not others?

shared/metrics/src/main/java/io/pravega/shared/metrics/Counter.java
kevinhan88: One principal of Counter is that it never goes down. Our existing implementation is anti-pattern, it decreases counter by resetting a new counter with a lower value.
Hence the dec() is removed from API to tight the discipline .

bindings/src/test/java/io/pravega/storage/filesystem/FileSystemStorageTest.java
kevinhan88: Sharp!
With the implementation of Dropwizard, metric registry is the storage of metric objects, and exporters are around the registry to send metric object value to outside.
In this test, which only concerns the storage of metrics, it is ok to not start (binding exporter to register).

With the implementation of Micrometer, there's no concept of exporter. Registry is for both storage and exporting. To solve the multiple exporting problem, Micrometer introduced a compositeRegistry, which is a API wrapper. CompositeRegistry must bind a concrete registry, such as Telegraf, InfluxDB, to land metric object into a concrete place.
Without binding to concrete registry, the operation to metric object is just an no-op.
Seems only this test class didn't call start() (where the binding happens), so I have to add it; other test classes already do this.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
EronWright: Micrometer docs make mention of how to use its API efficiently, in the bullet points here:
https://micrometer.io/docs/concepts#_global_registry

I think it is counter-productive to compose metric names on the fly.   The metric name may well be a constant; the tags provide the dynamic context such as `container_id`.  In other words, eliminate this method altogether.

@kevinhan88 mentioned that there's a back-compat concern, and I would suggest solving that in a different way, e.g. with a custom metric registry that encodes the metric name in the old way.

shared/metrics/src/test/java/io/pravega/shared/metrics/MetricRegistryUtils.java
kevinhan88: No, this is of builder pattern, along the way it always return builder object (in this case Search).
Finally it may return null metric object, which is expected.

shared/metrics/src/test/java/io/pravega/shared/metrics/StatsLoggerProxyTest.java
kevinhan88: I'm actively working on the @TODOs, just feel it's easier for the team to review with the minimum essential changes first, and then gradually go with further changes.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
kevinhan88: This is to close registry so it stops accepting new metric object, also stops sending more metrics.
Micrometer registry holds weak references to metric object, so there's no need to explicitly remove all metric objects inside.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: @EronWright yes our existing implementation is very heavy to look up metric object always starting from forming search key string from scratch. 
However seems it's not possible to directly store strong reference to metric inside Pravega particularly for those "dynamic" (not global) metrics, such as segment level metrics, because Pravega should remain state-less in the sense, but low level metric objects are obviously state objects.
Having said that, the life-cycle management of metric objects can still be significantly simplified by taking advantage of Micrometer's weak referencing to metric objects.

@andreipaduroiu - yes the lookup of metric object is costly, however this refactoring doesn't introduce any additional burden to it:
With the old implementation, to look up metric with dimension, we have to call "nameFromTransaction()", "nameFromSegment", "nameFromContainer" or "nameFromHost" to form lookup key from scratch. (we do this to avoid direct strong reference to metric object, which is still valid). 
For example, to update counter for metric "append_counter" with container id 6, we call nameFromContainer("append_count", 6) to form the lookup key, which is "append_count.6", and pass this key to metric register to locate the metric object.
With the introduction of tags, we no longer call "nameFromContainer", instead we call nameFromTags("append_count", "containerId", "6") to form lookup key, which is "append_count.6". We still need this lookup key to manage lifecycle of this particular metric object (e.g. when container is closed, all the related metric objects should be evicted)
But when we call metric register, we do no use the lookup key above, we call register this way: register.search("append_count").tags("containerId", "6").counter()...

The point I'm trying to make, "nameFromTags()" is doing similar things as the old implementation does. There is a big space to improve this lookup mechanism, particularly considering the weak reference of Micrometer registry, however, it is in addition to the refactoring here, which is to replace Dropwizard with Micrometer to support dimensional data model. I'll create new issue to refactor the code further. Before that, we need to see Pravega works well with Telegraf, InfluxDB etc.

gradle/wrapper/gradle-wrapper.properties
kevinhan88: sorry this is by accident (executed distribution command locally which changed this value)

shared/metrics/src/main/java/io/pravega/shared/metrics/Metric.java
kevinhan88: The formatting tool removed the leading spaces - please confirm indeed we need to add the leading two spaces back?

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
kevinhan88: Yes, I'm actively working on the TODOs.
I believe that you encouraged me to have the minimum essential changes out asap, so the team could have sufficient time to review it.
This TODOs is also my approach to highlight the critical places for discussion.
I'll ensure the final commit will stick with the guidelines strictly.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
kevinhan88: The static "name" method is from Dropwizard library. 
Yes I'll clone this method into NameUtils.java.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: Yes, it will be removed eventually.
I need the old libraries for the moment to compare Micrometer with Dropwizard.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: Fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
EronWright: My understanding of a dynamic logger is that it is a logger that includes instance data.   In the past, the dimensional data was packed into the metric name before the metric was looked up from the registry.  The reason for packing the data into the metric name is because codahale does not natively understand tags, and neither do many of the protocols.

Micrometer is different; it handles rewriting of the metric name as needed on a per-protocol basis.  For example, it transforms periods `.` to underscores `_` for some protocols.   For the statsd protocol, micrometer transforms the metric name in a manner similar to what `nameForX` does.   For the InfluxDB protocol, the metric name should not be transformed because InfluxDB protocol natively understands tags.  See [dimensionality](https://micrometer.io/docs/concepts#_supported_monitoring_systems).

I believe that the metric names may now be a constant, even when the metric is multi-dimensional.

kevinhan88: Agree, with the support of tag, the more general metric name should be a constant. 
I'm thinking to store tags into thread-local, or something similar. Hence lookup will be efficient as forming lookup key with metric name and tags is efficient, also the life-cycle mgmt is handled naturally with thread. No concrete solution yet.

Indeed it seems the teams are inspired by the Micrometer's programming model, makes it even more urgent to revisit our existing metric caching/lookup and life-cycle management mechanisms. Great!

EronWright: I see no need for a cache of metric objects.  Is it not as simple as:
```
import io.micrometer.Metrics;
Metrics.counter(
        Constants.PRAVEGA_SEGMENTSTORE_CONTAINER_APPENDCOUNT, 
        Constants.CONTAINER_ID, containerId).
    append();
```

kevinhan88: Yes got your point. We also need to manage the life-cycle of these metric objects. E.g the containers keep going up and down, end up with many objects inside register(s), but when should register clean them up? 
Micrometer register holds weak reference to these objects, so it depends on the strong reference from application. Seems Micrometer expects application to manage application objects life-cycle in some way, and metric objects just stay with application objects. If application object is gone, so does the strong reference to the metric object, then the metric objects is eligible for garbage collection, which marks the end of its life cycle.

kevinhan88: What Micrometer promotes is even further, it encourages application to store object directly, in the above case the APPEND_COUNTER_CONTAINER_6, and application directly updates this value, register directly export this value. No lookup at all end to end!
I'm sure we can come up with solid solution eventually.

EronWright: Why do you feel that Pravega need maintain a cache of meters?  My understanding is that the application need not keep references to the meters.  [Section 3.2](https://micrometer.io/docs/concepts#_global_registry) says:
> When tags need to be determined from local context, you have no choice but to construct/lookup the Meter inside your method body. The lookup cost is just a single hash lookup, so it will be acceptable for most uses.

kevinhan88: Metric registry must know when to terminate metric objects, otherwise it will run out of memory.
Micrometer terminates metric objects when:
1. all the strong reference from application ended;
2. registry.remove(metric) is explicitly called.

Pravega cache calls registry.remove() when time's up.
I'm not defending Pravega metric cache :-), I'm also thinking the way to get rid of it.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -80,25 +80,25 @@ public void onRemoval(RemovalNotification<String, Gauge> removal) {
...
 
     @Override
-    public void incCounterValue(String name, long delta) {
+    public void incCounterValue(String name, long delta, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
         Preconditions.checkNotNull(delta);
-        String counterName = name + ".Counter";
+        String counterName = nameFromTags(name, tags) + ".Counter";
         try {
             Counter counter = countersCache.get(counterName, new Callable<Counter>() {
```
tkaitchuck: This countersCache interface is very strange taking a callable like this. Perhaps we should look into ways to optimize this.

shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java
```diff
@@ -96,6 +101,30 @@
...
+     * The username to access InfluxDB.
+     */
+    @Getter
+    private final String influxDBUserName;
+
+    /**
+     * The password of user account accessing InfluxDB.
+     */
+    @Getter
+    private final String influxDBPassword;
```
tkaitchuck: Why would this be here?

shared/metrics/src/test/java/io/pravega/shared/metrics/StatsLoggerProxyTest.java
```diff
@@ -139,14 +139,18 @@ private TestMetric create(String name) {
...
     private static class TestMetric implements OpStatsLogger, Counter, Meter, Gauge {
         @Getter
-        private final String name;
+        private final io.micrometer.core.instrument.Meter.Id id;
```
tkaitchuck: Import Meter?

shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java
kevinhan88: In case Pravega sends metric directly to InfluxDB, username and password are needed to access the database.

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
kevinhan88: Yes.
I'm changing OpStatsData.PercentileSet to be based on Micrometer's set, the code will be optimized along the way.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
kevinhan88: This is the way Micrometer binds system metrics, per the documents.

shared/metrics/src/main/java/io/pravega/shared/metrics/Metric.java
andreipaduroiu: Please do not change it if it's already formatted. You can change your IDE to ignore this.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
andreipaduroiu: @kevinhan88 Just because a window is broken, you don't need to replace it with yet another broken window.

Similarly, just because the old code was inefficient is not an excuse to write inefficient code. I pointed out something that use a lot of CPU for the value that it produces and which is used on the hot path. Everything that we use on the hot path adds to latency and decreases our performance. 

So please try to figure out a better way of redoing this code.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
andreipaduroiu: Please remove this from gradle. You can run your benchmark against this branch but both of these should not make it to master.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
andreipaduroiu: The point is that you are closing an object that you do not own. This is a rather bad design and should be avoided as it may have unintended side effects.

Either create a new such instance with your class or don't close that object to which you have a static reference.

build.gradle
```diff
@@ -155,6 +155,12 @@ project ('shared:cluster') {
 
 project ('shared:metrics') {
     dependencies {
+        compile group: 'io.micrometer', name: 'micrometer-core', version: micrometerVersion
```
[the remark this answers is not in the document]
kevinhan88: Yes, removed. (They were kept there temporarily as I need to refer back from time to time)

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: Tags could be null or empty, it is not illegalArgument.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: Now codehal (and readytalk) are removed completely.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: This is the callback to be called when the counter is expired. Yes overall it is more feasible to revisit the caching mechanism with the new properties provided by Micrometer. The effort will probably be after this migration (as we need to ensure the migration itself runs smoothly)

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: sure

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: fixed.

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/test/java/io/pravega/shared/metrics/StatsLoggerProxyTest.java
kevinhan88: This is inner static class, seems cannot import to the parent class.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -21,20 +21,27 @@
...
+
 @Slf4j
 public class DynamicLoggerImpl implements DynamicLogger {
+
     private final long cacheSize;
     private final long cacheEvictionDuration;
 
-    private final MetricRegistry metrics;
+    private final MeterRegistry metrics;
     private final StatsLogger underlying;
```
RaulGracia: I feel that `underlying` is lacking something, like `underlyingStatsLogger` or `backendLogger`. Otherwise is not clear when it is use throughout the code.

shared/metrics/src/main/java/io/pravega/shared/metrics/NullStatsLogger.java
```diff
@@ -20,22 +21,22 @@
     static final NullGauge NULLGAUGE = new NullGauge();
     static final NullMeter NULLMETER = new NullMeter();
     static final NullOpStatsLogger NULLOPSTATSLOGGER = new NullOpStatsLogger();
-    private static final String NULLNAME = "";
+    private static final Id NULLID = null;
```
RaulGracia: Minor: `NULLID` -> `NULL_ID`

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsData.java
```diff
@@ -19,6 +19,8 @@
  */
 public class OpStatsData {
     static final EnumSet<Percentile> PERCENTILESET = EnumSet.allOf(Percentile.class);
+    // Added for Micrometer's implementation, PERCENTILEARRAY MUST matches PERCENTILESET!
+    static final double[] PERCENTILEARRAY = {0.1, 0.5, 0.9, 0.99, 0.999, 0.9999};
```
RaulGracia: Minor: `PERCENTILEARRAY` -> `PERCENTILE_ARRAY` (same for `PERCENTILE_SET`)

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsData.java
```diff
@@ -19,6 +19,8 @@
  */
 public class OpStatsData {
     static final EnumSet<Percentile> PERCENTILESET = EnumSet.allOf(Percentile.class);
+    // Added for Micrometer's implementation, PERCENTILEARRAY MUST matches PERCENTILESET!
```
RaulGracia: Minor: `matches` -> `match`

standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
```diff
@@ -53,7 +53,7 @@ public LocalPravegaEmulator build() {
                     .containerCount(4)
                     .restServerPort(restServerPort)
                     .enableRestServer(enableRestServer)
-                    .enableMetrics(false)
+                    .enableMetrics(true)
```
RaulGracia: If we enable metrics in Pravega standalone, maybe we will need a test to verify that metrics are working on standalone as well.

shared/metrics/src/test/java/io/pravega/shared/MetricsNamesTest.java
```diff
@@ -17,19 +17,42 @@
...
-    assertEquals("tag_fail", MetricsNames.failMetricName("tag"));
-    assertEquals("0_fail", MetricsNames.failMetricName("0"));
-    assertEquals("tag1_fail", MetricsNames.failMetricName("tag1"));
-    assertEquals("tag.tag_fail", MetricsNames.failMetricName("tag.tag"));
-    assertEquals("tag_fail.1", MetricsNames.failMetricName("tag.1"));
-    assertEquals("tag.tag1_fail", MetricsNames.failMetricName("tag.tag1"));
-    assertEquals("tag1.tag2.tag3_fail", MetricsNames.failMetricName("tag1.tag2.tag3"));
-    assertEquals("tag1.tag2_fail.3", MetricsNames.failMetricName("tag1.tag2.3"));
-  }
+    @Test
```
RaulGracia: Testing of metrics is somewhat limited in the current codebase of Pravega. Given the amount of changes that this PR involves, it would be interesting to think of how to extend the number of tests related to metrics (names, providers, etc.).

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/Metric.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: @andreipaduroiu I created a new class MetricKey to hold both cache key and registry key, in order to make things backwards compatible, and as efficient as possible.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: Please have a look at the MetricKey implementation.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
kevinhan88: Fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
kevinhan88: Fixed by removing all concrete registry inside the composite registry.
I think the registry.close() is the api for application to manage the lifecycle of registry. For example, when we start and close segmentstore, we start and close metric registry accordingly.

shared/metrics/src/test/java/io/pravega/shared/metrics/StatsLoggerProxyTest.java
kevinhan88: Fixed.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -50,9 +57,8 @@ public DynamicLoggerImpl(MetricsConfig metricsConfig, MetricRegistry metrics, St
                     public void onRemoval(RemovalNotification<String, Counter> removal) {
                         Counter counter = removal.getValue();
                         if (removal.getCause() != RemovalCause.REPLACED) {
-                            Exceptions.checkNotNullOrEmpty(counter.getName(), "counter");
-                            metrics.remove(counter.getName());
-                            log.debug("Removed Counter: {}.", counter.getName());
+                            metrics.remove(counter.getId());
```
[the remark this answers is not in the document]
kevinhan88: Yes in this implementation, Id is guaranteed available (even if it's not registered) so there's no need to null check.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: Are you talking about enhancement beyond this migration? Not quite getting your point.

shared/metrics/src/main/java/io/pravega/shared/metrics/NullStatsLogger.java
kevinhan88: Fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsData.java
kevinhan88: fixed

standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
kevinhan88: Thank you, I enabled it in local test but forgot to revert it back. Fixed.
(Metrics works in standalone mode partially, it can talk to InfluxDB in standalone mode, but it cannot talk to Telegraf, because some shaded libraries cannot be loaded in standalone mode)

shared/metrics/src/test/java/io/pravega/shared/MetricsNamesTest.java
kevinhan88: Yes, adding more tests to meet the diff coverage requirements.

shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java
```diff
@@ -30,13 +30,18 @@
...
     public final static Property<Integer> GRAPHITE_PORT = Property.named("graphitePort", 2003);
     public final static Property<String> JMX_DOMAIN = Property.named("jmxDomain", "domain");
     public final static Property<String> GANGLIA_HOST = Property.named("gangliaHost", "localhost");
     public final static Property<Integer> GANGLIA_PORT = Property.named("gangliaPort", 8649);
     public final static Property<Boolean> ENABLE_CSV_REPORTER = Property.named("enableCSVReporter", true);
     public final static Property<Boolean> ENABLE_STATSD_REPORTER = Property.named("enableStatsdReporter", true);
+    public final static Property<Boolean> ENABLE_INFLUXDB_REPORTER = Property.named("enableInfluxDBReporter", false);
     public final static Property<Boolean> ENABLE_GRAPHITE_REPORTER = Property.named("enableGraphiteReporter", false);
     public final static Property<Boolean> ENABLE_JMX_REPORTER = Property.named("enableJMXReporter", false);
     public final static Property<Boolean> ENABLE_GANGLIA_REPORTER = Property.named("enableGangliaReporter", false);
```
andreipaduroiu: Haven't we just removed all these other ones? Let's get rid of them here too.

Also find anywhere else in the code these might still be showing up and clean them up too.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -77,100 +79,102 @@ public StatsLogger createScopeLogger(String scope) {
...
         } else {
-            scopeName = name(basename, scope);
+            scopeName = basename + "." + scope;
         }
         return new StatsLoggerImpl(metrics, scopeName);
     }
 
     private class CounterImpl implements Counter {
-        private final com.codahale.metrics.Counter counter;
+        private io.micrometer.core.instrument.Counter counter;
```
andreipaduroiu: `GuardedBy(this)`

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -77,100 +79,102 @@ public StatsLogger createScopeLogger(String scope) {
...
-            long cur = counter.getCount();
-            counter.dec(cur);
+            metrics.remove(counter.getId());
+            this.counter = metrics.counter(name, tags);
         }
 
         @Override
         public long get() {
-            return counter.getCount();
+            return (long) counter.count();
```
andreipaduroiu: This method needs to be `synchronized`

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -77,100 +79,102 @@ public StatsLogger createScopeLogger(String scope) {
...
 
         @Override
         public void inc() {
-            counter.inc();
-        }
-
-        @Override
-        public void dec() {
-            counter.dec();
+            counter.increment();
```
andreipaduroiu: this one too

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -77,100 +79,102 @@ public StatsLogger createScopeLogger(String scope) {
...
         }
 
         @Override
         public void add(long delta) {
-            counter.inc(delta);
+            counter.increment(delta);
```
andreipaduroiu: and this one

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProvider.java
```diff
@@ -19,6 +19,12 @@
      */
     void start();
 
+    /**
+     * Initialize the stats provider without exporting functionality.
+     * In other words, starts with Micrometer SimpleMeterRegistry only.
```
andreipaduroiu: Can you please combine these two sentences into a single one that clearly explains what this is doing. `In other words` really doesn't belong in a Javadoc comment.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
```diff
@@ -66,84 +49,35 @@ public void start() {
...
         }
-        if (conf.isEnableConsoleReporter()) {
-            log.info("Configuring console reporter");
-            reporters.add(ConsoleReporter.forRegistry(getMetrics())
-                .convertRatesTo(TimeUnit.SECONDS)
-                .convertDurationsTo(TimeUnit.MILLISECONDS)
-                .build());
+
+        if (Metrics.globalRegistry.getRegistries().size() == 0) {
+            log.error("Error! No concrete metrics register bound, the composite registry runs as no-op!");
```
andreipaduroiu: You're already logging at error level; no need to include `Error!` in your message.

test/integration/src/test/java/io/pravega/test/integration/ControllerMetricsTest.java
```diff
@@ -162,7 +162,7 @@ public void streamMetricsTest() {
             // Check that the number of streams in metrics has been incremented.
             streamManager.createStream(scope, iterationStreamName, streamConfiguration);
             Counter createdStreamsCounter = MetricRegistryUtils.getCounter(getCounterMetricName(CREATE_STREAM));
-            Assert.assertTrue(i <= createdStreamsCounter.getCount());
+            Assert.assertTrue(i <= createdStreamsCounter.count());
```
andreipaduroiu: You can use `AssertExtensions.assertGreaterThanOrEqualTo`. In that case, your failure message will also include both values.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -185,7 +191,7 @@ public static String nameFromSegment(String metric, String segmentName) {
         return escapeSpecialChar(name);
     }
 
-    public static String nameFromHost(String metric, String hostId) {
+    public static String  nameFromHost(String metric, String hostId) {
```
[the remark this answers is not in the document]
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +235,56 @@ public static String failMetricName(String metricName) {
             return metricName + "_fail";
         }
     }
+
+    /**
+     * Generate MetricKey holds cache lookup key and metric registry key.
```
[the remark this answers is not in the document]
kevinhan88: Reworded.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +235,56 @@ public static String failMetricName(String metricName) {
...
+    public static MetricKey metricKey(String metric, String typeSuffix, String... tags) {
+
+        StringBuilder sb = new StringBuilder(metric);
+        if (tags == null || tags.length == 0) {  //no tags: append suffix to form cache & registry key
+            sb.append(typeSuffix);
+            String key = sb.toString();
+            return new MetricKey(key, key);
+        } else { //tags: append tag values and suffix to form cache key, original name is registry key
+            if (tags.length % 2 == 1) {
+                throw new IllegalArgumentException("Tags size must be even, it is a set of key=value pairs");
```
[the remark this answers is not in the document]
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +235,56 @@ public static String failMetricName(String metricName) {
...
+            }
+            for (int i = 0; i < tags.length; i += 2) {
+                if (Strings.isNullOrEmpty(tags[i]) || Strings.isNullOrEmpty(tags[i + 1])) {
```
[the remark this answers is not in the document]
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +235,56 @@ public static String failMetricName(String metricName) {
...
+            return new MetricKey(sb.toString(), metric);
+        }
+    }
+
+    /**
+     * MetricKey is a place holder to hold metric cache lookup key and registry key.
+     * This class is provided to keep the metric name convention backwards compatible.
+     *
+     */
+    public static class MetricKey {
```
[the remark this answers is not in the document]
kevinhan88: Fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java
kevinhan88: Fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProvider.java
kevinhan88: Reworded

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
kevinhan88: Fixed

test/integration/src/test/java/io/pravega/test/integration/ControllerMetricsTest.java
kevinhan88: Fixed

shared/metrics/src/test/java/io/pravega/shared/metrics/StatsLoggerProxyTest.java
tkaitchuck: You should be able to import the parent class and refer to it as `Meter.Id` assuming there is not another class named `Meter` being imported.

kevinhan88: In Pravega Meter is a concrete metric type (Rate in more general sense) on top of Metric, which is the base class of all metric types; with Micrometer Meter is the the base class of all metrics type.
So there is indeed a class name collision.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
andreipaduroiu: `@Synchronized` is not the same as adding the `synchronized` modifier to the method. Please be consistent.

In addition, mark the `counter` field as `@GuardedBy("this")` (assuming you go with the `synchronized` modifier).

EronWright:
@kevinhan88 I would like to verify the metrics data as it appears in InfluxDB.   Could you try it out and share with reviewers a backup of an influxdb database, or something similarly convenient?

kevinhan88:
@EronWright I did some simple manual test and verified Pravega now can talk to InfluxDB directly, or go through Telegraf, with or without tags. 
However I didn't run a big test to have massive metrics generated. I'm doing the following-up changes to add tags into those metrics per the Platform team's mapping file. It may make better sense to have a large scale test after both changes are in place - thoughts?

fpj:
@kevinhan88 I don't think @EronWright is asking for a "big test". he is asking for a way of validating that the data in influxdb is correct. as such, if you do a local test with an instance of influxdb and dump a backup, then you'll have what he is asking for.

also, the travis is not passing, please fix it.

kevinhan88:
metricswithtags png

Please find the attached screenshot for the final look of the metric "container.active_segments".
Before adding tags into this metric, all the flat metric names go directly to influxdb as showing as "pravega_segmentstore_container_active_segments_***_Gauge".
After adding tags into the metric (which the following PR will do), the metric name is always "pravega_segmentstore_container_active_segments", and by expanding it, we see tag "container" with value "1", "2", "3", "4".

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
RaulGracia: Sorry, I just referred to the name of the field `underlying` is ambiguous and could be renamed to make it clearer.

standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
RaulGracia: One question: Currently, in Pravega `standalone` we have a very useful `csvReporter` that allows us to see the output of the metrics locally. This is great for development purposes. Are we going to be able of using the `csvReporter` (or something similar) after this change?

gradle.properties
```diff
@@ -38,9 +38,7 @@ jerseyVersion=2.27
 junitVersion=4.12
 lombokVersion=1.18.4
 marathonClientVersion=0.6.0
-metrics3StatsdVersion=4.2.0
-metricsVersion=3.2.5
-metricsGangliaVersion=1.0.10
+micrometerVersion=1.1.2
```
RaulGracia: Just noticed that `1.1.3` has been released. Maybe it is worth to consider before merging the PR, in case that it does not involve code changes.

standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
kevinhan88: Sorry Micrometer doesn't support CSV yet, which is probably the biggest downside of the migration unfortunately.
This could be definitely an enhancement after the migration. Micrometer currently support console output, but it's less helpful as file output.

kevinhan88: I'm sure after the migration is in place, teams will raise issues for improvements.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -229,4 +238,58 @@ public static String failMetricName(String metricName) {
...
     }
+
+    /**
+     * Convenient method to concastenate two Strings using dot.
+     *
+     * @param element1 the first String object.
+     * @param element2 the second String object.
+     * @return String object with element1 and element2 connected with dot.
+     */
+    public static String name(String element1, String element2) {
```
RaulGracia: Some comments here: i) the method name `name` is confusing and the parameter names are ambiguous. As a method, you may want to describe the action that is happening inside, like `buildMetricName(String prefix, String metricName)` or something of the sort. ii) The description of the method looks as it was a general string concatenation, but this is using a dot between two strings for a reason: to respect the naming structure of metrics. So, maybe the description of the method should refer what is doing and why. iii) Typo: `concastenate` -> `concatenate`.

shared/metrics/src/main/java/io/pravega/shared/metrics/RegistryConfigUtil.java
```diff
@@ -0,0 +1,108 @@
...
+@Slf4j
+public class RegistryConfigUtil {
+
+    /**
+     * Create StatsdConfig for Statsd Register.
+     *
+     * @param conf the metric config from Pravega.
+     * @return instance of StatsdConfig to be used by Statsd Register.
+     */
+    public static StatsdConfig createStatsdConfig(MetricsConfig conf) {
```
RaulGracia: I have seen it written `Statsd` and `StatsD`; maybe we can pick just one way to refer to it.

shared/metrics/src/main/java/io/pravega/shared/metrics/RegistryConfigUtil.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            @Override
+            public String password() {
+                return conf.getInfluxDBPassword();
+            }
+
+            @Override
+            public String get(String k) {
+                return null;
```
RaulGracia: Is this the expected behavior?

kevinhan88: Yes, this is micrometer's way to provide default value.

shared/metrics/src/main/java/io/pravega/shared/metrics/RegistryConfigUtil.java
kevinhan88: Yes, changed to "Statsd"

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: fixed

gradle.properties
kevinhan88: fixed

config/config.properties
```diff
@@ -183,16 +183,16 @@ metrics.enableStatistics=false
...
 
-# Whether to enable StatsD reporting.
+# Whether to enable Statsd reporting.
 # Valid values: 'true' or 'false'.
 #metrics.enableStatsdReporter=true
 
-# StatsD endpoint host name.
-#metrics.statsDHost=localhost
+# Statsd endpoint host name.
+#metrics.statsdHost=localhost
```
RaulGracia: I think that changing this may have consequences in any application passing parameters to Pravega, right? Please, check with platform team if this is working after the change.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -240,13 +240,13 @@ public static String failMetricName(String metricName) {
...
-     * Convenient method to concastenate two Strings using dot.
+     * Convenient method to join two Strings with dot.
      *
      * @param element1 the first String object.
      * @param element2 the second String object.
-     * @return String object with element1 and element2 connected with dot.
+     * @return String object with element1 and element2 joined with dot.
      */
-    public static String name(String element1, String element2) {
+    public static String joinWithDot(String element1, String element2) {
```
RaulGracia: In all the public methods of this class we are using names that refer to the metrics, right? Instead, the name used for this method exposes why it does. Similarly to the rest of public methods, I think that we need to change it to something like `buildMetricName`, as it describe "what" it does (how it does it (using a `.` or a `-`), is an implementation detail).

config/config.properties
EronWright: Co
uld you elaborate on why you think that is?

kevinhan88: Not answering on Raul's behalf - The Platform team will make certain config changes to adopt this migration, e.g. start using InfluxDB directly, and provide influxdb connection details through config. 
(Not sure if the original question is about backwards compatibility, but config isn't within the scope I think)

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: This method is there simply because one static method "name" was gone with Dropwizard library.
"buildMetricName" is not a good name either as this method doesn't always return metric name. (metric name is quite complex so we even have MetricKey class and metricKey method to handle it). It indeed just joins two string with dot.

If you still insist a name change, could you advise one here and let's agree on it before I commit code change.

kevinhan88:
@RaulGracia to your question regarding system test: I've done end-to-end integration test, and attached screenshot onto this PR.

config/config.properties
RaulGracia: Just thinking in scripts, applications, etc. already using the previous name (`statsDHost`), they may need to be aware of this change for configuring the StatsD reporter correctly. If this situation is already controller, that's great.

RaulGracia:
Thanks @kevinhan88, but apart from integration test is necessary to check that system tests are passing as well. I started a full run on this PR for you.

shared/metrics/src/main/java/io/pravega/shared/metrics/RegistryConfigUtil.java
fpj: Can we please add the documentation reference here (because this is rather odd):

https://micrometer.io/docs/registry/influx

I'd also suggest adding the same comment in that page:

```
// accept the rest of the defaults
```

kevinhan88: Fixed.

shared/metrics/src/main/java/io/pravega/shared/metrics/OpStatsData.java
kevinhan88: fixed

config/config.properties
kevinhan88: Yes the Platform team uses statsDXXX.

PR #3383, opened 2019-02-15
Issue 2668: Enhancing StreamCut Validation on the client side

### LINKED ISSUE #2668: Incomplete verification in ReaderGroupConfig.verifyStartAndEndStreamCuts()
**Problem description**
When start and end stream cuts are passed to reader group, it verifies that the end cut's highest segment is below the begin cut's and the reverse for the lowest segment. However this is not sufficient to catch all invalid pairs of stream cuts. It could be possible for streamcuts to be overlapping and not ordered with respect to one another, but still have the highest and lowest segments be consistent. 

**Problem location**
ReaderGroupConfig.verifyStartAndEndStreamCuts()

More thoroughly verify that the ranges are consistent. 
This could be generalized by having streamCuts implement comparable.

huide9:
**Change log description**  
enhancement implementation for stream cut validation

**Purpose of the change**  
Fixes #2668 
Also fixes javadoc build warnings

**What the code does**  

- isStreamCutValid API added for utilize the controller implementation to verify key space integrity of the steram cut and make sure there is no key space overlapping at the same time.
- verifies: for the same segment in different cuts, the position (offset) sequence should be consistent 

**How to verify it**  
CleintFactoryTest - to be done

### DIFF
--- a/client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
+++ b/client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
@@ -51,10 +51,14 @@ import io.pravega.client.stream.InvalidStreamException;
 import io.pravega.client.stream.ReaderConfig;
 import io.pravega.client.stream.Serializer;
 import io.pravega.client.stream.Stream;
+import io.pravega.client.stream.StreamCut;
 import io.pravega.client.stream.TransactionalEventStreamWriter;
+import io.pravega.client.stream.ReaderGroupConfig;
 import io.pravega.common.concurrent.ExecutorServiceHelpers;
 import io.pravega.common.concurrent.Futures;
 import io.pravega.shared.NameUtils;
+
+import java.util.Map;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
@@ -237,6 +241,33 @@ public class ClientFactoryImpl implements ClientFactory, EventStreamClientFactor
     public void close() {
         controller.close();
         connectionFactory.close();
+    }
+
+    public void validateStartAndEndStreamCuts(final ReaderGroupConfig readerGroupConfig) {
+        readerGroupConfig.getEndingStreamCuts().entrySet().stream().filter(e -> !e.getValue().equals(StreamCut.UNBOUNDED))
+                .forEach(e -> {
+                    if (readerGroupConfig.getStartingStreamCuts().get(e.getKey()) != StreamCut.UNBOUNDED) {
+                        verifyStartAndEndStreamCuts(readerGroupConfig.getStartingStreamCuts().get(e.getKey()), e.getValue());
+                    }
+                });
+    }
+
+    private void verifyStartAndEndStreamCuts(final StreamCut startStreamCut, final StreamCut endStreamCut) {
+        final Map<Segment, Long> startPositions = startStreamCut.asImpl().getPositions();
+        final Map<Segment, Long> endPositions = endStreamCut.asImpl().getPositions();
+
+        //* verify both stream cut satisfy the "cover full key space" and "no key space overlap" in each of the streamcuts
+        Preconditions.checkArgument(controller.isStreamCutValid(startStreamCut).equals(true),
+                "startStreamCut key space violation.");
+        Preconditions.checkArgument(controller.isStreamCutValid(endStreamCut).equals(true),
+                "endStreamCut key space violation.");
+
+        //* check if the offsets are legal
+        Preconditions.checkArgument(startPositions.entrySet().stream().allMatch(s ->
+                endPositions.entrySet().stream().allMatch(e ->
+                        e.getKey().getSegmentId() == s.getKey().getSegmentId() &&
+                                startPositions.get(s.getKey()) < endPositions.get(e.getKey()))),
+                "Streamcut segment offset illegal.");
     }
 
 }
--- a/client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
+++ b/client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
@@ -59,6 +59,7 @@ import io.pravega.controller.stream.api.grpc.v1.Controller.SegmentValidityRespon
 import io.pravega.controller.stream.api.grpc.v1.Controller.SegmentsAtTime;
 import io.pravega.controller.stream.api.grpc.v1.Controller.StreamConfig;
 import io.pravega.controller.stream.api.grpc.v1.Controller.StreamInfo;
+import io.pravega.controller.stream.api.grpc.v1.Controller.StreamCutValidityResponse;
 import io.pravega.controller.stream.api.grpc.v1.Controller.SuccessorResponse;
 import io.pravega.controller.stream.api.grpc.v1.Controller.TxnRequest;
 import io.pravega.controller.stream.api.grpc.v1.Controller.TxnState;
@@ -661,6 +662,26 @@ public class ControllerImpl implements Controller {
             }
             LoggerHelpers.traceLeave(log, "getSuccessors", traceId);
         });
+    }
+
+    @Override
+    public CompletableFuture<Boolean> isStreamCutValid(final StreamCut streamCut) {
+        Exceptions.checkNotClosed(closed.get(), this);
+        long traceId = LoggerHelpers.traceEnter(log, "isStreamCutValid", streamCut);
+
+        final CompletableFuture<StreamCutValidityResponse> result = this.retryConfig.runAsync(() -> {
+            RPCAsyncCallback<StreamCutValidityResponse> callback = new RPCAsyncCallback<>(traceId, "isStreamCutValid");
+            client.isStreamCutValid(io.pravega.controller.stream.api.grpc.v1.Controller.StreamCut.newBuilder().putAllCut(getStreamCutMap(streamCut)).build(),
+                    callback);
+            return callback.getFuture();
+        }, this.executor);
+        return result.thenApply(StreamCutValidityResponse::getResponse)
+                .whenComplete((x, e) -> {
+                    if (e != null) {
+                        log.warn("isStreamCutValid failed: ", e);
+                    }
+                    LoggerHelpers.traceLeave(log, "isStreamCutValid", traceId);
+                });
     }
 
     @Override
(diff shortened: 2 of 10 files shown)


huide9:
Need comments from Tom and Shivesh. 
1. code location - moved from ReaderGroupConfig to ClientFactoryImpl
2. the timing to run the validation - need to be explicitly called by others VS. user object creation (in ReaderGroupConfig builder() )
3. if the API added fit for our convention

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
...
+        final Map<Segment, Long> endPositions = endStreamCut.asImpl().getPositions();
+
+        //* verify both stream cut satisfy the "cover full key space" and "no key space overlap" in each of the streamcuts
+        Preconditions.checkArgument(controller.isStreamCutValid(startStreamCut).equals(true),
+                "startStreamCut key space violation.");
+        Preconditions.checkArgument(controller.isStreamCutValid(endStreamCut).equals(true),
+                "endStreamCut key space violation.");
+
+        //* check if the offsets are legal
+        Preconditions.checkArgument(startPositions.entrySet().stream().allMatch(s ->
```
shiveshr: this may not be sufficient. as suggested earlier, add a new grpc api on controller that takes two streamcuts and responds if start is strictly less than end.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
...
+                    }
+                });
+    }
+
+    private void verifyStartAndEndStreamCuts(final StreamCut startStreamCut, final StreamCut endStreamCut) {
+        final Map<Segment, Long> startPositions = startStreamCut.asImpl().getPositions();
+        final Map<Segment, Long> endPositions = endStreamCut.asImpl().getPositions();
+
+        //* verify both stream cut satisfy the "cover full key space" and "no key space overlap" in each of the streamcuts
+        Preconditions.checkArgument(controller.isStreamCutValid(startStreamCut).equals(true),
```
shiveshr: these are remove calls being made to controller. you may want to chain these futures for optimization.
secondly, if a new api is introduced that takes two stream cuts, then the validation of individual cuts can happen as part of that single api call and you dont need to make two separate calls into controller.

shiveshr: also, `controller.isStreamCutValid(endStreamCut)` returns a future.. you cannot compare it with a boolean

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
         connectionFactory.close();
     }
 
+    public void validateStartAndEndStreamCuts(final ReaderGroupConfig readerGroupConfig) {
+        readerGroupConfig.getEndingStreamCuts().entrySet().stream().filter(e -> !e.getValue().equals(StreamCut.UNBOUNDED))
+                .forEach(e -> {
+                    if (readerGroupConfig.getStartingStreamCuts().get(e.getKey()) != StreamCut.UNBOUNDED) {
```
andreipaduroiu: Cache this value into a local variable. You are essentially fetching it twice.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
andreipaduroiu: Correct. There is no need to use `equals(boolean)`; just use ` == true` and the compiler would have caught this problem.

Also, if it's a Future, then this must be making a call over the wire. This will make this method very slow.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
...
+                "startStreamCut key space violation.");
+        Preconditions.checkArgument(controller.isStreamCutValid(endStreamCut).equals(true),
+                "endStreamCut key space violation.");
+
+        //* check if the offsets are legal
+        Preconditions.checkArgument(startPositions.entrySet().stream().allMatch(s ->
+                endPositions.entrySet().stream().allMatch(e ->
+                        e.getKey().getSegmentId() == s.getKey().getSegmentId() &&
+                                startPositions.get(s.getKey()) < endPositions.get(e.getKey()))),
+                "Streamcut segment offset illegal.");
```
andreipaduroiu: It may help to include more details in the error message, such as an illegal segment id or offending offsets (including segment id).

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -663,6 +664,26 @@ private Boolean handleScaleResponse(Stream stream, ScaleResponse response, long
...
+        final CompletableFuture<StreamCutValidityResponse> result = this.retryConfig.runAsync(() -> {
+            RPCAsyncCallback<StreamCutValidityResponse> callback = new RPCAsyncCallback<>(traceId, "isStreamCutValid");
+            client.isStreamCutValid(io.pravega.controller.stream.api.grpc.v1.Controller.StreamCut.newBuilder().putAllCut(getStreamCutMap(streamCut)).build(),
+                    callback);
+            return callback.getFuture();
+        }, this.executor);
+        return result.thenApply(StreamCutValidityResponse::getResponse)
+                .whenComplete((x, e) -> {
+                    if (e != null) {
+                        log.warn("isStreamCutValid failed: ", e);
```
andreipaduroiu: Why bother to log this exception here when it would bubble up to the caller anyway?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupState.java
```diff
@@ -211,6 +211,7 @@ boolean isReaderOnline(String reader) {
 
     /**
      * Returns the number of segments currently being read from and that are unassigned within the reader group.
+     * @return The number of segments currently being read from and that are unassigned within the reader group.
```
andreipaduroiu: These Javadoc changes are unrelated to the change at hand. Please undo them and submit them using a separate PR.

controller/src/main/java/io/pravega/controller/server/eventProcessor/LocalController.java
```diff
@@ -373,6 +373,12 @@ private StreamSegments getStreamSegments(List<SegmentRange> ranges) {
         return controller.isSegmentValid(segment.getScope(), segment.getStreamName(), segment.getSegmentId());
     }
 
+    @Override
+    public CompletableFuture<StreamSegments> isStreamCutValid(final StreamCut streamCut) {
+        return controller.isStreamCutValid(streamCut)
+                .thenApply(this::isStreamCutValid);
```
andreipaduroiu: This smells like an infinite loop.

andreipaduroiu:
@huide9 Please check your PR description and title for typos.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
         connectionFactory.close();
     }
 
+    public void validateStartAndEndStreamCuts(final ReaderGroupConfig readerGroupConfig) {
```
tkaitchuck: Does this need to be public? Is default sufficient?

tkaitchuck: I don't see where this is called.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
tkaitchuck: Don't we also need to check that the streams with just a start or just an end have a stream cut which is valid in that it covers all keyspace?

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
tkaitchuck: Please write a test to cover this case.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
...
+    }
+
+    private void verifyStartAndEndStreamCuts(final StreamCut startStreamCut, final StreamCut endStreamCut) {
+        final Map<Segment, Long> startPositions = startStreamCut.asImpl().getPositions();
+        final Map<Segment, Long> endPositions = endStreamCut.asImpl().getPositions();
+
+        //* verify both stream cut satisfy the "cover full key space" and "no key space overlap" in each of the streamcuts
+        Preconditions.checkArgument(controller.isStreamCutValid(startStreamCut).equals(true),
+                "startStreamCut key space violation.");
+        Preconditions.checkArgument(controller.isStreamCutValid(endStreamCut).equals(true),
```
tkaitchuck: Please write a test to cover this case.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
...
+                "endStreamCut key space violation.");
+
+        //* check if the offsets are legal
```
tkaitchuck: no *

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
...
+        Preconditions.checkArgument(startPositions.entrySet().stream().allMatch(s ->
+                endPositions.entrySet().stream().allMatch(e ->
```
tkaitchuck: This looks like it is going to iterate over every end position for each start position. That is going to make this an n^2 computation needlessly.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -239,4 +243,31 @@ public void close() {
...
+                        e.getKey().getSegmentId() == s.getKey().getSegmentId() &&
```
tkaitchuck: This does not cover the case where the segment IDs don't match.

andreipaduroiu:
@huide9 What is the status of this PR? I see no activity since February. If we aren't planning on merging it, I suggest we close it.

andreipaduroiu:
Closing for inactivity.

PR #3384, opened 2019-02-16
Issue 3385: Use Pravega Tables for storing controller metadata

shiveshr:
**Change log description**  
Note: All unit tests do not pass as we need `getAllKeys` api before all scenarios can be tested. Those that fail should fail with UnsupportedException. 

From this change onwards, Controller uses `Pravega Tables` for storing all stream metadata. 

**Purpose of the change**  
Fixex #3385

**What the code does**  
We have added a new Store type called PravegaTables for all controller metadata related to "streams". 
New files with the prefix `PravegaTables`* (like `PravegaTableScope, PravegaTableStream, PravegaTableStreamMetadataStore` etc) is where all the store specific implementation logic resides. 
These classes make use of `PravegaTableStoreHelper` which is a wrapper over segmentHelper and ensures it calls appropriate Wire command into segment store for all table store related activities and handles exceptions and translates them into store specific exceptions. 

Following is how the stream metadata is structured in tables:
Tables:
```
1. System/scopes table
key: scope-name, value: empty

2. scope/streams-in-scope table
key: stream name, value: empty

3. stream metadata tables:

    3.1 `<scoped-stream-name>_streamMetadata`
     1. Key: creation time  // never deleted
     2. Key: configuration record // never deleted
     3. key: truncation record // never deleted
     4. key: state record // never deleted
     5. key: epoch transition record // never deleted
     6. key: committing-transaction-record // never deleted
     7. key: current epoch record // never deleted
     8. key: retention set record // never deleted
     9. key: epoch record <epoch number>  // never deleted
     10. key: history time series <chunk> // never deleted O(number of epochs / 1000)
     11. key: segment sealed size map <shard-number> // never deleted O(number of epochs / 1000)
     12. key: <segment number> // never deleted O(number of segments)
     13. key: markers_<segment-number> // deletable entries
     14. key: stream-cut-<reference> // deletable entries
     15. key: waitingRequestProcessor // deletable entry

     3.2 <scoped-stream-name>_transactions_in_epochs
     1. key: epoch Value: null
    
     3.3 <scoped-stream-name>_transactions_<epoch>
     key: txnId value: active txn record
    
     3.4 system/transaction batches
     key: batch, value: empty

     3.5 completed_transactions_<batch>
     key: <scoped-stream-name-txnId>, value: completed txn record 
    
        create transaction: 
        0. add transaction key in `transactions-epochs-table`
        if (table does not exist) --> 
        1. add epoch key in transactions-epochs-table --> create epoch specific table
        try step 0 again
        
        complete txn:
        1. write completed txn record to new batch.
        2. remove transaction key from the table
        3. delete empty table --> delete epoch from transactions-epochs-table
        
        list transactions in epoch:
        1. list all entries in epoch specific transaction table. (this can also throw table not exists which would be transalated to DataNotFoundException)
        
        list all transactions:
        1. get all epochs with transactions from `transactions-epochs-table`
        2. call list transactions in epoch
```
Additional related changes include refactoring segment helper to take `hoststore`, `connectionfactory` and `authhelper` in constructor as opposed to each method call as params. 
This is done because SegmentHelper is used by PravegaTablesStoreHelper and we dont need to pass these objects with every call.
This results in changes to streamMetadataTasks and StreamTransactionMetadataTasks in their requirement for aforesaid objects. They simply need the segment helper and no other object. 

The StoreCLientConfig has a new store type called PravegaTable. The PravegaStoreClient has handle to zookeeper client and segment helper. 
The controller has a new configuration to disable using pravega tables for metadata store. It reverts to using zookeeper as metadata store if the use-pravega-tables configuration is set to false. 
The other stores (namely taskStore, checkpoint store and hoststore) continue to be on zookeeper. 
The stores that use Pravega Tables all related to stream metadata. 

The store is also made closeable because the ZK garbage collector that is passed to both zookeeper based implementation and pravega-tables based implementation has to be closed otherwise it will continue to run in the background. 

ZkInt96 counter is also abstracted out of ZkStreamMetadataStore into a utility class that is used in both 
ZkStreamMetadataStore and PravegaTablesStreamMetadataStore. 

All unit tests that used zk store have also been abstracted to have both zkstore and pravegaTables store implementations. So all scenarios that were tested with zkStore are tested with PravegaTableStore. 
The segmentHelper is mocked using `SegmentHelperMock` which provides an in memory implementation of pravegaTables which is used in all these unit tests. 

**How to verify it**  
New unit Tests added.


controller/src/main/java/io/pravega/controller/store/index/PravegaTablesHostIndex.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.Executor;
+
+/**
+ * Zookeeper based host index.
```
tkaitchuck: I don't think this comment is correct.

controller/src/main/java/io/pravega/controller/store/index/PravegaTablesHostIndex.java
```diff
@@ -0,0 +1,143 @@
...
+                        return CompletableFuture.completedFuture(null);
+                    }
+                });
+    }
+
+    @Override
+    public CompletableFuture<List<String>> getEntities(final String hostId) {
+        Preconditions.checkNotNull(hostId);
+        String table = getHostEntityTableName(hostId);
+        List<String> entries = new LinkedList<>();
```
tkaitchuck: Never use a linked list. Iterating over one is over 500x slower than over an array list.

controller/src/main/java/io/pravega/controller/store/index/PravegaTablesHostIndex.java
```diff
@@ -0,0 +1,143 @@
...
+        return storeHelper.getAllKeys(SYSTEM_SCOPE, table)
+                .forEachRemaining(entries::add, executor)
```
tkaitchuck: This involves multiple threads and needs to be made thread safe. I think a better solution would be to implement a 'collect' method on async iterator.

controller/src/main/java/io/pravega/controller/store/index/PravegaTablesHostIndex.java
```diff
@@ -0,0 +1,143 @@
...
+                .thenApply(x -> entries);
+    }
+
+    @Override
+    public CompletableFuture<Set<String>> getHosts() {
+        Set<String> hosts = new ConcurrentSkipListSet<>();
+        return storeHelper.getAllKeys(SYSTEM_SCOPE, hostsTable)
+                          .forEachRemaining(hosts::add, executor)
+                          .thenApply(v -> hosts);
```
tkaitchuck: Collect would make this clearer.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public CompletableFuture<Void> deleteScope() {
+        return Futures.toVoid(storeHelper.deleteTable(scopeName, streamsInScopeTable, true))
+                .thenCompose(deleted -> storeHelper.removeEntry(SYSTEM_SCOPE, SCOPES_TABLE, scopeName));
+    }
+
+    @Override
+    public CompletableFuture<List<String>> listStreamsInScope() {
+        List<String> result = new LinkedList<>();
```
tkaitchuck: Not linked list.

Issue #3277, opened 2019-01-13
Revisit authorization logic


EronWright:
**Problem description**
1. The authorization subsystem doesn't give adequate information to the authorization handler, to disambiguate between scope creation and scope use.   One is unable to grant to the administrator the ability to create scopes, and grant to a user the ability to create streams within a given scope (but not to create scopes).   Similar situation for scope deletion.

2. Similar situation for streams with respect to scopes; e.g. the ability to write to a stream without the ability to create streams.

3. Some of the calls are marked as requiring `READ_UPDATE` whereas `READ` may be sufficient:
- `getCurrentSegments`
- `getSegments`
- `getURI`
- `isStreamCutValid`
- `getDelegationToken`  (varies by situation?)

**Problem location**
- `io.pravega.controller.server.rpc.grpc.v1.ControllerServiceImpl`
- `io.pravega.controller.server.rest.resources.StreamMetadataResourceImpl`

**Suggestions for an improvement**
Adjust the argument passed to the `authorize` call on the auth handler, such that a call to create or delete a sub-resource has the parent resource as the argument.   For example:

|action|logic|
|-----------|-----------------------|
|list scopes | `authorize("/", READ)` |
|create scope | `authorize("/", READ_UPDATE)` |
|delete scope | `authorize("/", READ_UPDATE)` |
|list streams | `authorize(scopeName, READ)` |
|create stream | `authorize(scopeName, READ_UPDATE)` |
|delete stream | `authorize(scopeName, READ_UPDATE)` |
|read stream | `authorize(scopeName + "/" + streamName, READ)` |
|write stream | `authorize(scopeName + "/" + streamName, READ_UPDATE)` |

Be sure to update `ControllerServiceImpl` and `StreamMetadataResourceImpl` since the logic is duplicated.

sarlaccpit:
I think the table captures a state of things that makes sense to me.

One nuance that is missing from the list though is the desire to be able to list all scopes on one hand, but also filtered by what the user has access to on the other hand.

In other words, if a user has READ on 3 out of the 10 scopes available, list scopes should return those 3.   Additionally, it would be best if that didn't require the user to have READ on /, since this is what the list all scopes requires.

Perhaps what we need is 2 calls.

- An admin call:  "list all scopes" (/v1/scopes)  --> requires READ on /.  You have it, you see everything, you don't have it, you see nothing.
[...]

PR #3110, opened 2018-11-15
Issue 3108: List all the scopes during listScopes call

arvindkandhare:
Signed-off-by: arvindkandhare <arvind.kandhare@emc.com>

**Change log description**  
Removes auth for each scope during `listScopes` call.

**Purpose of the change**  
Fixes #3108.

**What the code does**  

Implementation of auth*n for `listScopes` is as follows:
1. Check whether the principal has access to `/`
2. Check whether the prinicipal has access to individual scopes
3. Return only the scopes that the principal has access to.

This gives wrong info through the UI.  This is also not in sync with other `list*` operations (e.g. `listStreams`). The generic behavior is that if a principal has appropriate access permissions for a container (e.g. `scope` in case of `listStreams`, `\` in case of `listScopes`), listing is allowed.

Updating the behavior of `listScopes` to match other implementations.

**How to verify it**  
Unit tests pass.

Issue #3108, opened 2018-11-15
listScopes returns 500 when auth is enabled


sarlaccpit:
**Problem description**

When auth is enabled on the controller and segment store, listScopes returns 500.  Create scope and get one scope behave normally.

This is using the built-in basic auth handler with password file.

```
Get scopes without credentials, 401 as expected

usxxcbalczunm1:pravega-auth cbalczun$ curl http://...:10080/v1/scopes -H "Content-Type: application/json" -i
HTTP/1.1 401 Unauthorized
Content-Length: 0

Now passing the right credentials:

usxxcbalczunm1:pravega-auth cbalczun$ curl http://...:10080/v1/scopes -H "Authorization: Basic YWRtaW46MTExMV9hYWFh" -H "Content-Type: application/json" -i
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Length: 0

Create a scope

usxxcbalczunm1:pravega-auth cbalczun$ curl http://...:10080/v1/scopes -H "Authorization: Basic YWRtaW46MTExMV9hYWFh" -H "Content-Type: application/json" -i -d "{ \"scopeName\": \"myscope\"}"
HTTP/1.1 201 Created
Content-Type: application/json
Content-Length: 23

{"scopeName":"myscope"}

Get one scope:

usxxcbalczunm1:pravega-auth cbalczun$ curl http://...:10080/v1/scopes/myscope -H "Authorization: Basic YWRtaW46MTExMV9hYWFh" -H "Content-Type: application/json" -i
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 23

{"scopeName":"myscope"}

But list all scopes:

usxxcbalczunm1:pravega-auth cbalczun$ curl http://...:10080/v1/scopes -H "Authorization: Basic YWRtaW46MTExMV9hYWFh" -H "Content-Type: application/json" -i
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Length: 0
```

When the 500 occurs, this can be observed in the controller logs:

```
2018-11-15 02:13:02,124 77882 [controllerpool-10] WARN  i.p.c.s.r.r.StreamMetadataResourceImpl - listScopes failed with exception:
java.util.concurrent.CompletionException: java.lang.IllegalStateException: Not inside a request scope.
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
[...]

PR #3386, opened 2019-02-18
Issue 3277: Authorization model enhancements

ravisharda:
**Problem description**  

This PR addresses 2 authorization-related problems described below. (For more information, refer to the description section and comments in the corresponding issue #3277.) 

_Problem 1:_

The current authorization model has some gaps that make it difficult to implement fine-grained authorization policies such as that illustrated by the following scenario: 

1. Admins can create (as well as read and delete) scopes. They have `READ_UPDATE` permission on the resource `/`.  
2. Everyone else can list scopes, but cannot create or delete them. They have `READ` permission on `/`. 
3. A given analytics project has a project-specific scope. Admin creates a dedicated scope for that project. She can do so, as she has permission to create a scope (based on 1 above).  
4. Users/client applications of the analytics project create streams within the project's scope. To be able to do so, they are assigned `READ_UPDATE` permission on the scope created for the project. However, they are not allowed to delete the scope. Deleting the scope requires READ_UPDATE on `/`.
5. There are other users/client applications that need to be able to list the streams within that project's scope. They are assigned `READ` permission on the scope. 
6. There are users and client applications that will write to a specific stream within a given scope. They are assigned `READ_UPDATE` permission on the stream. There are others that need to only read from the stream; they are assigned `READ` permission on the stream.

_Problem 2:_

Another problem is that if a user has permission to list scopes, she sees all scopes in the system. There is no mechanism to limit visibility of individual scopes, such that I see scope A and B, and you see scopes A, C, and D. The functionality to see only the "authorized-to-see" scopes was actually present earlier but was taken out as part of [PR 3110](https://github.com/pravega/pravega/pull/3110/) in response to issue #3108. (The root cause behind #3108 has been fixed in this PR.)

**Purpose of the change**  
Enhances Pravega authorization logic as discussed in #3410 .

**What the code does**  
The main changes are as follows: 

1. The controller gRPC and REST API implementation classes have modified resource strings passed to AuthHandler, in order to solve the problems modify the authorization model. 
    * ``io.pravega.controller.server.rpc.grpc.v1.ControllerServiceImpl`` 
    * `` io.pravega.controller.server.rest.resources.StreamMetadataResourceImpl``

Specifically, the resource strings now follow the following rules (See https://github.com/pravega/pravega/issues/3277#issuecomment-466900097 and the discussions following it for the rationale):

* Create and list scopes: `/` (parent resource)
* Create and list streams: `<scope-name>` (parent resource)
* Delete scope: `/` (parent resource) 
* Update/delete stream: `<scope-name>/<stream-name>` (leaf resource)
* List scopes filters results based on authorized-to-see objects: `<scope-name>` (leaf resource)
* List streams filters results based on authorized-to-see objects: `<scope-name>/<stream-name>` (leaf resource)

2. Controller REST API ``list scopes`` now filters scopes based on whether the user has access to the given scope. The same applies to ``list streams`` as well.

3. Creation of resource strings for auth was scattered and duplicated in the gRPC and REST implementation classes. That logic has now moved to its own ``AuthResourceRepresentation`` class. This was done in order to make it easy to inspect, modify and unit test auth resource strings. This PR required changes to some of the resource strings.  

4. Authorization methods were refactored out of ``StreamMetadataResourceImpl`` into a dedicated ``RESTAuthHelper``, which opened up opportunities for removing duplication and unit test this functionality. The methods also needed a change to avoid having to fetch request headers directly, which was necessary in order to be able to avoid the issue mentioned in #3108 (`` java.lang.IllegalStateException: Not inside a request scope.``). Also, new functionality required for authorizing list scope filtering based on access has been added in this new class.  

5. Unit testing related changes: 
   * Unit test classes were added for all new classes. 
   * New ``Fake`` (not Mock) classes to aid unit testing. 
   * ``PravegaAuthManager`` was modified to enable registering of FakeAuthHandler, to aid unit testing. 
   * ``PartialSecurePravegaClusterTest`` was renamed to ``AuthEnabledInProcPravegaClusterTest``, in order to make the class name more precise. This class as touched in order to add a more comprehensive test for auth. 
   * A new test class focused on testing authorization logic was added: `StreamMetaDataAuthFocusedTests`. It uses the controller REST API to drive the tests. 
  
**How to verify it**  
All tests should pass.


controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,47 @@
...
+package io.pravega.controller.server;
+
+import io.pravega.common.Exceptions;
+
+/**
+ * A utility class with methods for preparing string representations of auth-protected resources.
+ */
+public class AuthResourceRepresentation {
+
+    public static String ofScopes() {
```
andreipaduroiu: Please add Javadoc for each of these methods. I cannot tell what they're supposed to do just from their names.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,47 @@
...
+        return "/";
+    }
+
+    public static String ofAScope(String scopeName) {
+        Exceptions.checkNotNullOrEmpty(scopeName, "scopeName");
+        return scopeName;
+    }
+
+    public static String ofStreams(String scopeName) {
```
andreipaduroiu: Did you mean to name the argument `streams`? Are you sure it's supposed to be plural too?

andreipaduroiu: Otherwise what's the difference between this and `ofAScope`?

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
```diff
@@ -132,54 +131,9 @@ public void createScope(final CreateScopeRequest createScopeRequest, final Secur
...
-            }
-
-            String credentials = authParams.get(0);
-            Preconditions.checkNotNull(credentials, "Credentials not specified in the parameters.");
-
-            if (!pravegaAuthManager.authorize(resourceName, principal, credentials, level)) {
-                throw new AuthorizationException("Auth failed for " + resourceName, Status.FORBIDDEN.getStatusCode());
-            }
-        }
+    private List<String> fetchAuthorizationHeader() {
```
andreipaduroiu: `getAuthorizationHeader` sounds more appropriate

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
```diff
@@ -127,6 +127,11 @@ public boolean authorize(String resource, Principal principal, String credential
         return handler.authorize(resource, principal).ordinal() >= level.ordinal();
     }
 
+    // For enabling unit testing
```
andreipaduroiu: Replace this comment with `@VisibleForTesting`

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
```diff
@@ -127,6 +127,11 @@ public boolean authorize(String resource, Principal principal, String credential
...
+    public void registerHandler(AuthHandler authHandler) {
+        this.handlerMap.put(authHandler.getHandlerName(), authHandler);
```
andreipaduroiu: `synchronized`

controller/src/main/java/io/pravega/controller/server/rpc/auth/RESTAuthHelper.java
```diff
@@ -0,0 +1,85 @@
...
+        return null;
+    }
+
+    public void authenticateAuthorize(List<String> authHeader, String resourceName, AuthHandler.Permissions level)
+            throws AuthException {
+        if (pravegaAuthManager != null) {
+            String credentials = parseCredentials(authHeader);
+
+            if (!pravegaAuthManager.authenticateAndAuthorize(resourceName, credentials, level)) {
+                throw new AuthorizationException("Auth failed for " + resourceName,
```
andreipaduroiu: please be consistent in your messages. I prefer `Authorization` instead of `Auth`.

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    @Override
+    public void setUp() throws Exception {
+        this.authEnabled = true;
+        this.tlsEnabled = false;
+        super.setUp();
+    }
+
+    @Override
+    public void createTestStream() {
+        // his test from the base is unnecessary in this class as that and more is already being
```
andreipaduroiu: `his test`?

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        // tested in another test method.
+    }
+
+    // Note: Strictly speaking, this test is really an "integration test" and is a little
```
andreipaduroiu: Why don't you use proper Javadoc for methods?

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        @Cleanup
+        EventStreamReader<String> reader = clientFactory.createReader(
+                "readerId", readerGroup,
+                new JavaSerializer<String>(), ReaderConfig.builder().build());
+
+        // Keeping the read timeout large so that there is ample time for reading the event even in
+        // case of abnormal delays in test environments.
+        String readEvent1 = reader.readNextEvent(20000).getEvent();
+        log.info("Done reading event [{}]", readEvent1);
```
andreipaduroiu: This is a test, Why do you log?

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
tkaitchuck: Given that these method's signatures don't differentiate by type streams from scopes. It might be helpful to make their names even more explicit. IE: "ofStreamsInScope" / "ofReaderGroupsInScope"

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,47 @@
...
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.controller.server;
+
+import io.pravega.common.Exceptions;
+
+/**
+ * A utility class with methods for preparing string representations of auth-protected resources.
+ */
+public class AuthResourceRepresentation {
```
tkaitchuck: final

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
```diff
@@ -512,7 +477,17 @@ public void listScopes(final SecurityContext securityContext, final AsyncRespons
...
                              ScopesList scopes = new ScopesList();
                              scopesList.forEach(scope -> {
-                                     scopes.addScopesItem(new ScopeProperty().scopeName(scope));
+                                 try {
+                                     if (restAuthHelper.isAuthorized(authHeader,
+                                             AuthResourceRepresentation.ofAScope(scope),
+                                             principal, READ)) {
+                                         scopes.addScopesItem(new ScopeProperty().scopeName(scope));
+                                     }
+                                 } catch (AuthException e) {
```
tkaitchuck: Let's not use exceptions as flow control. This is not only ugly, it about 3 orders of magnitude slower. Is there a way to simply check if the principal has access?

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
ravisharda: Sure. Have done so now.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
ravisharda: Yes, I did intend the plural ``streams``. I've now added JavaDocs comments to this class and its methods now. Hopefully, the comments clarify the reason for the name and also clarify the difference between the two: ``ofStreams`` and ``ofAScope``.

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
ravisharda: Yeah, I initially thought the logic it contains is expensive. I avoided calling it ``get..()`` as one usually expects that it is cheap to call get...() as the return value is say cached/precomputed, and I thought that is not the case here. On a deeper look, it seems like the method really meets the criteria. Have renamed it to get..().

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
ravisharda: Fixed/

controller/src/main/java/io/pravega/controller/server/rpc/auth/RESTAuthHelper.java
ravisharda: Actually, in this case, the exception AuthException (representing authentication and/ authorization) should have been thrown, so I have modified it accordingly. We use three related terms: authentication, authorization and just auth: the last one represents the first two collectively. 
Have also modified the message to make it more explicit.

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
ravisharda: Yeah, you are right. Will try to use them more frequently.

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
ravisharda: I do want to keep the log statements as the test brings up an in-memory cluster and is not totally unlikely to fail due to environmental factors. When a failure happens, the log statements might come in handy - especially when the build is tested after the merge in the Kubernetes cluster. 
What do you say?

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
ravisharda: Fixed.

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
ravisharda: Fixed. :)

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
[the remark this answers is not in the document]
ravisharda: Yeah, I initially thought the logic it contains is expensive. I avoided calling it ``get..()`` as one usually expects a ``get`` method to be very cheap (as the return value is say cached/precomputed), and I thought that is not the case here. On a deeper look, it seems like the method really meets the criteria. Have renamed it to ``get..(``).

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
ravisharda: Sure, that sounds great to me. Have pushed a commit with this change.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
ravisharda: Done.

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
ravisharda: To clarify, we are not using the exception for the purpose of flow control here. I introduced a new method ``isAuthorized(..)`` precisely to avoid having to use exception as flow control. Under normal circumstances the exception is not expected to be encountered during authorization. 

But, ``RESTAuthHelper's isAuthorized(..)`` delegates authorization decision internally to an existing method: ``PravegaAuthManager's authorize()``, which returns true/false usually. In one specific scenario, it returns an exception. The catch block in the highlighted code in there is to handle that.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,111 @@
...
+     * Creates a resource representation for use in authorization of actions pertaining to the collection of streams
+     * within the specified scope.
+     *
+     * @param scopeName the name of the scope
+     * @return a string representing the collection of streams within the scope with the specified name
+     * @throws NullPointerException if {@code scopeName} is null
+     * @throws IllegalArgumentException if {@code scopeName} is empty
+     */
+    public static String ofStreamsInScope(String scopeName) {
+        Exceptions.checkNotNullOrEmpty(scopeName, "scopeName");
```
andreipaduroiu: FYI, you can just return the result of this method. It returns the argument you passed.

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
```diff
@@ -127,6 +130,12 @@ public boolean authorize(String resource, Principal principal, String credential
         return handler.authorize(resource, principal).ordinal() >= level.ordinal();
     }
 
+    @Synchronized
```
andreipaduroiu: No. `@Synchronized` will create a field named `$lock` and synchronize the method on that one.

If you look at the definition of `handlerMap`, it is `GuardedBy(this)`. So you either need to add the `synchronized` modifier (not annotation) to the method or use a `synchronized(this)` block inside it.

Doesn't your IDE highlight this problem? You can configure IntelliJ to highlight concurrency issues. You can find that in Settings->Editor.

controller/src/main/java/io/pravega/controller/server/rpc/auth/RESTAuthHelper.java
```diff
@@ -0,0 +1,92 @@
...
+    public void authorize(List<String> authHeader, String resourceName, Principal principal,
+                          AuthHandler.Permissions permissionLevel) throws AuthException {
+        if (!isAuthorized(authHeader, resourceName, principal, permissionLevel)) {
+            throw new AuthorizationException(
+                    String.format("Failed to authorize for resource [%s]", resourceName),
+                    Response.Status.FORBIDDEN.getStatusCode());
+        }
+    }
+
+    public  Principal authenticate(List<String> authHeader) throws AuthException {
```
andreipaduroiu: unnecessary extra space after `public`

controller/src/main/java/io/pravega/controller/server/rpc/auth/RESTAuthHelper.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public class RESTAuthHelper {
+
+    private final PravegaAuthManager pravegaAuthManager;
+
+    public RESTAuthHelper(PravegaAuthManager pravegaAuthManager) {
+        this.pravegaAuthManager = pravegaAuthManager;
+    }
+
+    public boolean isAuthorized(List<String> authHeader, String resourceName, Principal principal,
```
andreipaduroiu: Javadoc for these methods please.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
ravisharda: Ah! I didn't realize that. Will fix it.

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
ravisharda: Thanks for catching it. Will fix it in the next commit. And, thanks for the suggestion about enabling it in IntelliJ - I didn't know it had something like that (mostly had been Eclipse and Visual Studio user).

controller/src/main/java/io/pravega/controller/server/rpc/auth/RESTAuthHelper.java
ravisharda: Hmm.. you certainly have a keen eye! :) Will fix it.

controller/src/main/java/io/pravega/controller/server/rpc/auth/RESTAuthHelper.java
ravisharda: I'm not a big fan of documenting internal methods extensively - I'd rather have the code structure explain it. I also understand the opposite school In this case, the structure certainly doesn't explain it well enough, so I'll document the methods. Curious, what's your take on it in general?

fpj:
@ravisharda I'd like to discuss the authorization model a bit. Here are two points that call my attention:

1- Root is not an independent resource. If I have `READ` access to root, then I can't read root entirely under the described model, I can only read the other resources that I have access to. Consequently, `READ` permission for root is really a combination of the access permission to root plus access permission to the other resources (scopes).
2- `UPDATE` permission on a scope does not authorize the corresponding user to delete the scope. In the model described, to be able to delete a scope, the user needs update access to root. This is another instance in which the resources are not treated independently.

My preference is that we respect the independence of resources. I understand that there is the issue that we only want to expose to users the scopes they can access. To achieve it, we can do one of thwo things:

a. Filter at the API level and not in the authorization model, which is not ideal because it still enables in theory that any user lists all scopes, which could be consider a security leak.

b. Remote the notion of root and directly provide a mapping of principal to scope resources. The call to list scopes for a given user returns the scopes associated to the principal.

What do you think?

ravisharda:
@fpj 
I think I see where you are coming from. 

Thinking out loud, another way of thinking about our authorization model is that there are two different authorization levels in the order of granularity:

1. **Action-level:** Can the user perform this "action" on an existing object collection? The following are examples of this type: 
   a) List, create and delete scopes (on the collection of scopes represented as ``/``).
   b) List, create and delete streams (on the collection of streams of a scope represented as ``<scope-name>``)

2. **Instance-level:**  Can the subject perform an action to "this particular instance" of an object? Examples of this type include:
   a) List a particular scope (represented as ``<scope-name>``)
   b) Delete a particular scope (could be represented as ``<scope-name>``)
   c) Get or update a particular stream (represented as ``<scope-name>/<stream-name>``)
   d) Delete a particular stream (could be represented as ``<scope-name>/<stream-name>``)

(I've left out rest of the resource types for brevity)

Now, what we have in the code of this PR are: 1.a, 1.b, 2.a, 2.c. What we don't have are 2.b, and 2.d. 

My reading of your comments is that there is a mixup of the two types of authorization levels described above. The fix would be straightforward: Enhance the API's to check for instance-level authorization when service requests such as 2.b, and 2.d. There are alternatives as well, such as pasing a hierarchical resource argument format such as ``//scopeName:streamName/eventId`` and letting the auth handler do multiple levels of authorizations in a single call (this was discussed in the issue comment [here](https://github.com/pravega/pravega/issues/3277#issuecomment-460157931). 

Hope I fully understood your concern.

fpj:
@ravisharda thanks for bearing with me. I'm after some principled way to explain the model we are using, right now it feels ok but a bit ad-hoc. I'm not sure I agree with your separation of *action* and *instance* because it is hard to separate the two concepts when talking about. It is better to talk about resource and rights, just like you did in the description. An authorization domain for us is a collection of <resource, rights> pairs. The domain for a given user is established by his/her credentials. The resources are three at the moment:
- Root
- Scope
- Stream

and the rights are:
- `READ`
- `READ_UPDATE`

Next we need to define what `READ` and `READ_UPDATE` means in each case. 
- Root `READ`: read the state of root, which is a list of scopes. It is odd to make it conditional on the rights of the scopes.
- Root `READ_UPDATE`: in addition to `READ`, it enables the creation and deletion of scopes.
- Scope `READ`: list the streams in the scope. Is listing streams here also conditional on the rights of the streams?
- Scope `READ_UPDATE`: in addition to `READ`, it enables the creation and deletion of streams.
- Stream `READ`: ability to read events/bytes of a stream
- Stream `READ_UPDATE`: ability to read and write events/bytes to a stream

If I look at the authorization model this way, then the only issue standing out for me is the conditional listing. `READ_UPDATE` does not give the ability of changing the resource itself, like deleting it, which makes sense, otherwise we would be enabling an admin user to delete root (I assume that only admins will have `READ_UPDATE` on root).

I could justify the reduced visibility of root and scopes when listing their corresponding elements in the case `READ` also implies that the element can be listed.

- Scope `READ`: list the streams in the scope and can itself be listed.
- Stream `READ`: ability to read events/bytes of a stream and can be listed.

If that's the case, then we are really conflating two different aspects into one type of rights, which is fine in the case we never envision them going separately.

One other confusion I have is whether a user needs root `READ` access to be able to work with a stream. I'm basically trying to assess whether all users need at least `READ` access for root.

ravisharda:
@fpj 

In general, 

* **Permissions** define the type of access a subject has on a securable object/resource. For instance, in Windows you can grant a user read and modify permissions to a file.
* On the other hand, **rights** (which is another vehicle of assigning privileges) defines the type of operations a subject can perform. For example, in Windows, user rights authorize users to perform per-computer operations such as signing in to a computer interactively, backing up files, auditing user's successful/failed access to objects, etc. - all actions that affect the entire computer, rather than a particular object. 

(An interesting trivia: Typically, permissions are attached to the object, while rights are attached to user accounts.)

IMO, these concepts map very well to the two authorization levels I described earlier: **Instance-level** and **action-level** (for lack of better terms):
* Instance-level: ``READ on myscope`` that allows a subject to see a particular scope (i.e. an object) ``myscope`` can be viewed as a "permission". 
* Action-Level: ``READ on /`` that allows a subject to perform a scope listing operation can be seen as a "user right". It is really an abstraction of an "action": a right to list scopes. 

To answer your question: 
> One other confusion I have is whether a user needs root `READ` access to be able to work with a stream. I'm basically trying to assess whether all users need at least `READ` access for root.

No, they do not. For example, to list a stream, the user needs ``READ on <scope-name>``. To create or delete a stream, the user needs ``READ_UPDATE on <scope-name>``. (The permission string ``READ_UPDATE`` is indeed confusing: it could be easier to understand if it were something like ``READ_WRITE``.)

fpj:
@ravisharda I might have used permission indistinctly, which is not appropriate. I follow the POSIX terminology:

https://www.usenix.org/legacy/publications/library/proceedings/usenix03/tech/freenix03/full_papers/gruenbacher/gruenbacher_html/main.html

I don't think it is necessary to talk about instance and action. I prefer to stick to the old terminology, which is objects, permission and rights.

To summarize what I think you are saying:
- The restricted visibility is explained by the fact that list-ability is a right associate to the `READ` permission.
- We are consistently using that right with `READ`, except for streams, which do not have anything under it to list.

Now let me go over a few scenarios:

1- Principal has: 
   - Root: No permission
   - Scope _X_: read permission
   - Stream _X/A_ : read permission
   - Scope _Y_ : read permission
   - Stream _Y/B_: read permission

The principal has the ability of reading streams _X/A_ and _Y/B_, but it can't list root to get _X_ and _Y_. Is it right?

2- Principal has: 
   - Root: No permission
   - Scope _X_: No permission
   - Stream _X/A_ : read permission
 
The principal has the ability of reading the stream _X/A_, but not list scope _X_. Is it right?

Let me know if these scenarios are valid and confirm whether my understanding above is correct, pls.

ravisharda:
@fpj yes, the two scenarios are valid and the behaviors you mentioned about them are correct.

ravisharda:
The commits I just made were for rebasing purposes. It doesn't have the latest changes as discussed in issue #3277 yet, so it's not ready for further review yet.

ravisharda:
The latest commit has the changes discussed in the latest comments under the corresponding [issue 3227](https://github.com/pravega/pravega/issues/3277). Requesting the reviewers to initiate reviews.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Creates a resource representation for use in authorization of actions pertaining to the collection of streams
+     * within the specified scope.
+     *
+     * @param scopeName the name of the scope
+     * @return a string representing the collection of streams within the scope with the specified name
+     * @throws NullPointerException if {@code scopeName} is null
+     * @throws IllegalArgumentException if {@code scopeName} is empty
+     */
+    public static String ofStreamsInScope(String scopeName) {
```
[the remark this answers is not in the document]
ravisharda: These methods do the same thing. I created three separate methods so that the caller calls a method with a role-suggesting name.

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
```diff
@@ -206,7 +166,8 @@ public void createStream(final String scopeName, final CreateStreamRequest creat
         }
 
         try {
-            authenticateAuthorize(scopeName + "/" + streamName, READ_UPDATE);
+            restAuthHelper.authenticateAuthorize(getAuthorizationHeader(),
+                    AuthResourceRepresentation.ofStreamsInScope(scopeName), READ_UPDATE);
```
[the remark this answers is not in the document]
ravisharda: IMHO, the current way seems simpler and sufficient. We don't need to vary the object's internal representation or prepare it in a step-by-step fashion. What do you say?

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
```diff
@@ -127,6 +130,11 @@ public boolean authorize(String resource, Principal principal, String credential
         return handler.authorize(resource, principal).ordinal() >= level.ordinal();
     }
 
+    @VisibleForTesting
+    public synchronized void registerHandler(AuthHandler authHandler) {
+        this.handlerMap.put(authHandler.getHandlerName(), authHandler);
```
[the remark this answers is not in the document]
ravisharda: It's meant to be used for unit testing only. In RESTAuthHelperTest I'm registering a FakeAuthHandler so that I unit test auth functionality in RESTAuthHelper.

ravisharda:
Working on adding some more tests.

ravisharda:
It should be noted that the authorization model changes warrants some changes on how authorization rules are specified for the default PasswordAuthHandler. The following table lists some examples scenarios: 

|S.No.| User Scenario | Statement | Remarks|
|:----:|:------|:----------|:---------|
|i.| A highly privileged user | ``privilegeduser:<encrypted-password(ep)>:*,READ_UPDATE``| This user has all permissions across the entire system.|
|ii.| An admin who can create new scopes | ``scopecreator:<ep>:/,READ_UPDATE``|-|
|iii.| An admin who can not only create but also delete all scopes | ``scopemanager:<ep>:/,READ_UPDATE;/*,READ_UPDATE``| Since delete requires READ_UPDATE permission on the specific scope, the acl for resource ``/*`` is necessary.|
|iv.| A user who can list all scopes | ``scopelister:<ep>:/,READ``<br/>``;/*,READ``|-|
|iv.| A user who can list only some scopes | ``limitedscopelister:<ep>:/,READ;scope1,READ;scope5,READ``|This user will see only scope1 and scope5 when listing the scopes|
|v.| A user who can create streams within a given project's scope | ``streamsinascopecreater:<ep>:scope1,READ_UPDATE``|This user can do all stream-related (and other) operations for scope ``scope1``|

Also, The following table lists the resource strings after the authorization model changes as decided in #3277 (option 2). 

|S.No.| Action |	Resource String for Auth (in this PR)  | Required Permission | gRPC method (in ``ControllerServiceImpl``) | REST controller method (in ``StreamMetadataResourceImpl``| 
|:--:|:-----------|:---------------:|:---------------:|:---------------:|:----------------------|
|1.| List scopes |  ``/``, and <br/>``<scope-name>`` for each scope|READ |-|``listScopes``| 
|2.| Get a scope |  ``<scope-name>`` |READ |-|``getScope``| 
|3.| Create a scope | ``/`` |READ_UPDATE |``createScope``|``createScope``| 
|4.| Delete a scope | ``<scope-name>`` |READ_UPDATE |``deleteScope``|``deleteScope``| 
|5.| List streams | ``<scope-name>`` and <br/> ``<stream-name>`` for each stream|READ |``listStreamsInScope``|``listStreams``| 
|6.| Get a stream | ``<scope-name>/<stream-name>`` |READ |-|``getStream`` |
|7.| Create a stream |  ``<scope-name>`` |READ_UPDATE |``createStream``|``createStream``| 
|8.| Update a stream |  ``<scope-name>/<stream-name>`` |READ_UPDATE |``updateStream``|``updateStream``| 
|9.| Update a stream's state |  ``<scope-name>/<stream-name>`` |READ_UPDATE |-|``updateStreamState``| 
|10.| Delete a stream |  ``<scope-name>/<stream-name>`` |READ_UPDATE |``deleteStream``|``deleteStream``| 
|11.| Truncate a stream |  ``<scope-name>/<stream-name>`` |READ_UPDATE |``truncateStream``|-| 
|12.| Seal a stream |  ``<scope-name>/<stream-name>`` |READ_UPDATE |``sealStream``|-| 
|13.| List reader groups | ``<scope-name>`` |READ |-|``listReaderGroups``| 
|14.| Get a reader group |  ``<scope-name>/<reader-group-name>`` |READ |-|``getReaderGroup``| 
|15.| List scaling events of a stream |  ``<scope-name>/<stream-name>`` |READ |-|``getScalingEvents``| 
|16.| List current stream segments | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``getCurrentSegments`` | -|
|17.| List stream segments | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``getSegments`` | -|
|18. | List segments following| ``<scope-name>/<stream-name>``| READ | ``getSegmentsImmediatlyFollowing``|-|
|19.| List segments between | ``<scope-name>/<stream-name>``| READ | ``getSegmentsBetween``|-|
|20.| Scale (a stream) | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``scale``|-|
|21.| Check scale status | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``checkScale``|-|
|22.| Get URI for segment | ``<scope-name>/<stream-name>`` | READ | ``getURI``|-|
|23.| Is Segment valid | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``isSegmentValid``|-|
|24.| Is stream cut valid | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``isStreamCutValid``|-|
|25.| Create a transaction | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``createTransaction``|-|
|26.| Commit transaction | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``commitTransaction``|-|
|27.| Abort transaction | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``abortTransaction``|-|
|28.| Ping transaction | ``<scope-name>/<stream-name>`` | READ | ``pingTransaction``|-|
|29.| Check transaction state | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``checkTransactionState``|-|  
|30.| Get delegation token | ``<scope-name>/<stream-name>`` | READ_UPDATE | ``getDelegationToken``|-|

sarlaccpit:
Does the baked in password file in `config` need to be updated to reflect the changes for the admin users? 

https://github.com/pravega/pravega/blob/6f977dd75e3dc9a9158c17fbd554dc78480bc79e/config/passwd

sarlaccpit:
In the table,  #4 and #7  are overlapping.   The ability to delete a scope and create a stream in that scope should not be dictated by the same permission (READ_UPDATE on the scope).

ravisharda:
> Does the baked in password file in `config` need to be updated to reflect the changes for the admin users?
> 
> https://github.com/pravega/pravega/blob/6f977dd75e3dc9a9158c17fbd554dc78480bc79e/config/passwd

No, it doesn't, as the file specifies access control entry for a user ``admin`` using `*,READ_UPDATE`. Like earlier, this provides the user all permissions across the entire system.

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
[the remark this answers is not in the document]
tkaitchuck: If this map is only for testing, then it should be located in test code. The best way to do this is to wrap or subclass the real class in the test class and inject the modified implementation.

sarlaccpit:
I think that in general, this PR improves the situation even if it doesn't solve everything.   For example, my comment about the issue or ambiguity (depending how we look at it) with rows 4 and 7 of the table is already present today.  Not made better or worse by this PR.

What this PR adds is refactors that will help future work (intangible for the net outcome of functionality that this PR brings, but good investment nonetheless IMHO).

In terms of net changes/improvement, the 2 that jump out to me are:

- create resource is now consistent ( create scope asks for READ_UPDATE on /, create stream asks for READ_UPDATE_ on /scopename).
- list scopes is filtered by who has  read access on such and such scope. --> this allows plugin implementations to also add logic which lets "admins" list everything under / even if they don't have READ explicitly on each scope.   Other plugin implementation can decide to require READ on each scope no matter what; point is, the interface allows the flexibility for the implementation to choose.  In general, in an ideal world, the more of these flexibilities are left to the implementation of the plugin, the better.

So I feel these are plusses that we should get in, even if there is probably more work to do on the overall authz model which, I think we're finding, can only be pushed so far with the current mechanics of READ and READ_UPDATE.

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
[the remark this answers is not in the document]
ravisharda: The map `handlerMap` itself is used in the main code as well. It holds entries comprising of custom auth handlers by handler name. The method `registerHandler(..)` was the one that was added to aid unit testing, in order to make it possible to inject an auth handler entry into `PravegaAuthManager`.  
I took a cue from your suggestion to also explore if it'd be useful to create an inherited sub-class which adds this sole method `registerHandler`, but then felt it'd be overkill in this case. Besides, having this method right here makes it easier for others to spot it and use it in their own tests when necessary. A comment might be useful here - I'll add it.
Please let me know if you agree.

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaAuthManager.java
```diff
@@ -127,6 +130,19 @@ public boolean authorize(String resource, Principal principal, String credential
...
+
+    /**
+     * This method is not only visible for testing, but also intended to be used solely for testing. It allows tests
+     * to inject and register custom auth handlers. Also, this method is idempotent.
+     *
+     * @param authHandler the {@code AuthHandler} implementation to register
+     * @Throws NullPointerException {@code authHandler} is null
+     */
+    @VisibleForTesting
+    public synchronized void registerHandler(AuthHandler authHandler) {
```
fpj: If you want to avoid having this method public, you could make it protected and create an extended class for testing.

controller/src/main/java/io/pravega/controller/server/rpc/auth/RESTAuthHelper.java
```diff
@@ -0,0 +1,131 @@
...
+     * The delegate used for performing authentication and authorization.
+     */
+    private final PravegaAuthManager pravegaAuthManager;
+
+    public RESTAuthHelper(PravegaAuthManager pravegaAuthManager) {
+        this.pravegaAuthManager = pravegaAuthManager;
+    }
+
+    /**
+     * Determines whether the given {@code principal} has specified {@code permission} on the given {@code resource}.
```
fpj: To be consistent with the definition here, I want to confirm that we are using `subject` to denote the abstract entity being authorized and `principal` being the code object, is that it?

controller/src/test/java/io/pravega/controller/rest/v1/SecureStreamMetaDataTests.java
```diff
@@ -33,10 +33,17 @@ public void setup() throws Exception {
...
-            writer.write("user1:" + passwd + ":/,READ;scope1,READ_UPDATE;scope2,READ_UPDATE;\n");
+
+            // User "user1" can:
+            //    - list, create and delete scopes
+            //    - Create and delete streams within scopes "scope1" and "scope2". Also if "user1" lists scopes,
+            //      she'll see those scopes, but not "scope3".
+            writer.write("user1:" + passwd + ":/,READ_UPDATE;scope1,READ_UPDATE;scope1/*,READ_UPDATE;scope2,READ_UPDATE;scope2/*,READ_UPDATE;\n");
+
             writer.write("user2:" + passwd + ":/,READ;scope3,READ_UPDATE;\n");
-            writer.close();
```
fpj: Why are you removing this `close` call?

controller/src/test/java/io/pravega/controller/rest/v1/StreamMetaDataAuthFocusedTests.java
```diff
@@ -0,0 +1,616 @@
...
+    @SuppressWarnings("checkstyle:StaticVariableName")
+    private static Client client;
+
+    @SuppressWarnings("checkstyle:StaticVariableName")
+    private static File passwordHandlerInputFile;
+
+    // We want to ensure that the tests in this class are run one after another (in no particular sequence), as we
+    // are using a shared server (for execution efficiency). We use this in setup and teardown method initiazers
+    // for ensuring the desired behavior.
+    Lock sequential = new ReentrantLock();
```
fpj: I'm wondering if this locking is strictly necessary. Would it be sufficient that we use different resource names (scope and stream) to avoid conflicts?

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+        // Keeping the read timeout large so that there is ample time for reading the event even in
+        // case of abnormal delays in test environments.
+        String readEvent1 = reader.readNextEvent(10000).getEvent();
+        log.info("Done reading event [{}]", readEvent1);
+
+        String readEvent2 = reader.readNextEvent(10000).getEvent();
+        log.info("Done reading event [{}]", readEvent2);
+
+        assertEquals(writeEvent1, readEvent1);
```
fpj: Can we get a checkpoint event here?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -368,33 +392,39 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request, StreamO
...
-                                                    }),
-                responseObserver, requestTag);
+        authenticateExecuteAndProcessResults(
+                () -> this.authHelper.checkAuthorization(AuthResourceRepresentation.ofAScope(scope),
+                        AuthHandler.Permissions.READ),
+                delegationToken -> controllerService.listStreams(
+                        scope, request.getContinuationToken().getToken(), listStreamsInScopeLimit)
+                        .thenApply(response -> {
+                             List<StreamInfo> streams = response.getKey().stream()
+                                     .filter(m -> authHelper.isAuthorized(m, AuthHandler.Permissions.READ))
```
fpj: This is one of the semantic changes, it filters the elements of the list according to visibility, right?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -156,16 +159,18 @@ public void deleteStream(StreamInfo request, StreamObserver<DeleteStreamStatus>
 
         log.info(requestTag.getRequestId(), "deleteStream called for stream {}/{}.",
                 request.getScope(), request.getStream());
-        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(request.getScope() + "/" +
-                        request.getStream(), AuthHandler.Permissions.READ_UPDATE),
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
+                AuthResourceRepresentation.ofAStreamInScope(request.getScope(), request.getStream()),
```
fpj: This one of the semantic changes proposed here, that delete depends on the permission over scope and stream right?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -358,7 +381,8 @@ public void checkTransactionState(TxnRequest request, StreamObserver<TxnState> r
     public void createScope(ScopeInfo request, StreamObserver<CreateScopeStatus> responseObserver) {
         RequestTag requestTag = requestTracker.initializeAndTrackRequestTag(requestIdGenerator.get(), "createScope", request.getScope());
         log.info(requestTag.getRequestId(), "createScope called for scope {}.", request.getScope());
-        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(request.getScope(), AuthHandler.Permissions.READ_UPDATE),
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
+                AuthResourceRepresentation.ofScopes(), AuthHandler.Permissions.READ_UPDATE),
```
fpj: Another one of the semantic changes, the authorization to create a scope depends on the permission over `ofScopes` which is `/`, right?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: Yes, that's true.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: Yes, that's right. We now use the parent resource representation for authorization of create operations such as this one.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: Yes, the delete depends on the permission over the respective resource. That's not a semantic change though, as the behavior was the same in the pre-PR code, as well. I had modified it to use parent resource in this PR but then changed it back based on our auth model discussions in the respective issue.

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
ravisharda: There shouldn't be an occurrence of a checkpoint event here, `disableAutomaticCheckpoints()` is invoked on `readerGroupConfig` in line 115:

https://github.com/pravega/pravega/blob/f5ba3f3139563fd7da5cc58babccb2ae4f77da28/standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java#L113-L116

controller/src/test/java/io/pravega/controller/rest/v1/StreamMetaDataAuthFocusedTests.java
ravisharda: I wanted to avoid the possibility of invisible/indirect dependencies among tests and race conditions leading to flaky and hard-to-debug tests. Also, depending on due care taken in naming resources alone don't appear to be sufficient, IMHO. Moreover, the presence of this lock and its usage makes it visibly apparent that the intention is to run the tests in this class one at a time, and that it'll be true even if it were decided that the rest of the tests are run using parallel execution.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHelper.java
```diff
@@ -31,22 +31,29 @@ public static AuthHelper getDisabledAuthHelper() {
...
-            if (allowedLevel.ordinal() < expectedLevel.ordinal()) {
-                throw new RuntimeException(new AuthorizationException("Access not allowed"));
-            }
+            return (allowedLevel.ordinal() < permission.ordinal()) ? false : true;
+        } else {
+            return true;
+        }
+    }
+
+    public String checkAuthorization(String resource, AuthHandler.Permissions expectedLevel) {
```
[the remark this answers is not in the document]
ravisharda: `checkAuthorization` was an existing method used in a lot of places. `isAuthorized` is a new method I added for list operations for checking authorization for each scope/stream without having to depend on occurrence of an exception to determine failed authorization. I moved some of the logic inside `checkAuthorization` into the new `isAuthorized` method. I didn't want to modify existing code that depends on checkAuthorization, to avoid changing any semantics in the authorization behavior, and also because that change'd have been irrelevant to problem at hand.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Creates a resource representation for use in authorization of actions pertaining to the specified scope.
+     *
+     * @param scopeName the name of the scope
+     * @return a string representing the scope with the specified name
+     * @throws NullPointerException if {@code scopeName} is null
+     * @throws IllegalArgumentException if {@code scopeName} is empty
+     */
+    public static String ofAScope(String scopeName) {
```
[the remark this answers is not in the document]
ravisharda: Done. Have made the change as you've suggested. Will push the change soon.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,110 @@
...
+        Exceptions.checkNotNullOrEmpty(scopeName, "scopeName");
+        return scopeName;
+    }
+
+    /**
+     * Creates a resource representation for use in authorization of actions pertaining to the collection of streams
+     * within the specified scope.
+     *
+     * @param scopeName the name of the scope
+     * @return a string representing the collection of streams within the scope with the specified name
```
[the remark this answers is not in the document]
ravisharda: Sure, have modified it to say this instead. Please let me know if you don't like it:

`a string representing the collection of streams under the scope with the specified name`

The string itself is really just the scope name technically, but in so far as the caller is concerned, it is asking for a resource representing the collection of streams to which it wants to add a new stream (for example).

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param scopeName the name of the scope
+     * @param readerGroupName the name of the reader group
+     * @return a string representing the specified reader group
+     * @throws NullPointerException if {@code scopeName} or {@code streamName} are null
+     * @throws IllegalArgumentException if {@code scopeName} or {@code streamName} are empty
+     */
+    public static String ofAReaderGroupInScope(String scopeName, String readerGroupName) {
+        Exceptions.checkNotNullOrEmpty(readerGroupName, "readerGroupName");
+        return String.format("%s/%s", ofReaderGroupsInScope(scopeName), readerGroupName);
+    }
```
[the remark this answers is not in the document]
ravisharda: I think representing the target of authorization as strings keeps the model very simple and easy to understand. IMHO, we should enhance it if/when the need arises, in a separate PR.

controller/src/main/java/io/pravega/controller/server/AuthResourceRepresentation.java
```diff
@@ -0,0 +1,110 @@
...
+     * Creates a resource representation for use in authorization of actions pertaining to the specified stream within
+     * the specified scope.
+     *
+     * @param scopeName the name of the scope
+     * @param streamName the name of the stream
+     * @return a string representing the specified stream within the specified scope
+     * @throws NullPointerException if {@code scopeName} or {@code streamName} are null
+     * @throws IllegalArgumentException if {@code scopeName} or {@code streamName} are empty
+     */
+    public static String ofAStreamInScope(String scopeName, String streamName) {
```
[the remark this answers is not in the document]
ravisharda: My intention of using that `A` was to make it easy to immediately distinguish the two kinds of methods, such as: `ofScopes`, `ofAScope`. From your comment, I realize they are easily distinguishable anyway as the parameters are different. Have removed the `A` from all methods, accordingly.

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
```diff
@@ -256,7 +217,8 @@ public void deleteScope(final String scopeName, final SecurityContext securityCo
         long traceId = LoggerHelpers.traceEnter(log, "deleteScope");
 
         try {
-            authenticateAuthorize(scopeName, READ_UPDATE);
+            restAuthHelper.authenticateAuthorize(getAuthorizationHeader(),
+                    AuthResourceRepresentation.ofScope(scopeName), READ_UPDATE);
```
EronWright: I thought that we agreed that deleting a scope would be considered an operation on the parent (`/`).   The goal is for scopes to be provisioned by the administrator; the scope user shouldn't be able to delete it.    Therefore the solutions are:

```
restAuthHelper.authenticateAuthorize(getAuthorizationHeader(),
                    AuthResourceRepresentation.ofScopes(), READ_UPDATE);
```

or introduce a new verb:
```
restAuthHelper.authenticateAuthorize(getAuthorizationHeader(),
                    AuthResourceRepresentation.ofScope(scopeName), DELETE);
```

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
```diff
@@ -299,7 +261,10 @@ public void deleteStream(final String scopeName, final String streamName, final
         long traceId = LoggerHelpers.traceEnter(log, "deleteStream");
 
         try {
-            authenticateAuthorize(scopeName + "/" + streamName, READ_UPDATE);
+            restAuthHelper.authenticateAuthorize(
+                    getAuthorizationHeader(),
+                    AuthResourceRepresentation.ofStreamInScope(scopeName, streamName),
+                    READ_UPDATE);
```
EronWright: I thought we agreed that stream deletion was an action taken on the parent scope.

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
fpj: If we add a new `DELETE` permission, then we should at least have a `CREATE` too. I'd rather not introduce only one. If we are to change the set of permissions we can grant, I'd rather rethink the model entirely and propose a new design. For this PR, my preference is that we stick to `READ` and `READ_UPDATE`, and accept its limitations.

Please check the discussion in the issue from this comment onward:

https://github.com/pravega/pravega/issues/3277#issuecomment-466900097

PR #3387, opened 2019-02-18
Issue 3227: Fix controller recovery of Zookeeper session

### LINKED ISSUE #3227: grpc auth handler is not initialized properly upon controller restart
**Problem description**
While running longevity test in PKS  cluster with moderate IO (medium-scale) workload on `pravega:issue3131-reopen-0.5.0-2058.9dae626` branch, observed controller restart after `~72hrs` due to ZK session expiry.
```
# grep -r "Creating store client" .
./controller_2018-12-21.21.log:2018-12-21 13:43:58,018 276988618 [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Creating store client
./controller_2018-12-18.0.log:2018-12-18 08:47:30,238 838  [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Creating store client
 ```
Upon controller restart, it is expected to resume the IO/workload. But here looks like after controller restart grpc auth handler is not initialized properly due to that IO's are failing with below error
```
2018-12-21 21:10:51,009 303801609 [grpc-default-executor-360] ERROR i.p.c.s.r.g.v.ControllerServiceImpl - Controller api failed with authenticator error
2018-12-21 21:10:51,081 303801681 [grpc-default-executor-360] INFO  i.p.c.s.r.g.v.ControllerServiceImpl - getURI called for segment longevity/mediumScale/16660178147157.
2018-12-21 21:10:51,082 303801682 [grpc-default-executor-360] ERROR i.p.c.s.r.g.v.ControllerServiceImpl - Controller api failed with authenticator error
2018-12-21 21:10:59,844 303810444 [controllerpool-66] WARN  i.p.c.s.r.StreamCutBucketService - [requestId=8141109484465893907] Exception thrown while performing auto retention for stream StreamImpl(scope=longevity, streamName=mediumScale)
java.util.concurrent.CompletionException: io.pravega.common.util.RetriesExhaustedException: java.util.concurrent.CompletionException: io.pravega.controller.store.host.HostStoreException: Could not find host for container id: 2
```

### PR DESCRIPTION
**Pull request by RaulGracia**
**Change log description**  
Allow ZKClientFactory to correctly handle unexpected changes in Zookeeper connection parameters and close the ZK client when this occurs.

**Purpose of the change**  
Fixes #3227.

**What the code does**  
When the Controller detects a ZK session expiration, it attempts to re-use the existing ZK client. But before actually re-using the same ZK client, the Controller first checks that all the parameters of the former ZK client are still the same (see `ZKClientFactory:newZooKeeper`):
```
Preconditions.checkArgument(this.connectString.equals(connectString), "connectString differs");
Preconditions.checkArgument(this.sessionTimeout == sessionTimeout, "sessionTimeout differs");
Preconditions.checkArgument(this.canBeReadOnly == canBeReadOnly, "canBeReadOnly differs");
Preconditions.checkNotNull(watcher, "watcher");
```
However, there is a semantic problem with the first precondition regarding ZK `connectString`: On the one hand, when the Controller is instantiated, the Pravega operator provides it with a "hostname" as a ZK connection endpoint (e.g., `pravega-zookeeper-client:2181`). On the other hand, during the subsequent session expirations, the Curator Framework provides a series of ZK server IPs to connect with (e.g., `xxx.8:2181,xxx.11:2181,xxx.13:2181`). As can be observed, the first precondition will never be met, given the disparate format in which both the Pravega operator and Curator pass the ZK connection endpoints to the Controller. As a consequence of this mismatch, the Controller gets stuck and cannot neither re-connect to ZK nor restart, and it eventually blocks IO from a client perspective. 

To fix this issue, this PR allows `ZKClientFactory` to correctly handle `IllegalArgumentException`s related to precondition check failures and closes the ZK client, so the Controller is forced to restart.

**How to verify it**  
After applying this PR to a longevity experiment, it has been running for +4 days. We actually observed the same situation that made the Controller to get stuck before, now being handled correctly:
1) _Zookeeper connectivity problems_: First, we observed exceptions in Zookeeper logs that are responsible for connection problems in other components of the system:
```
(zookeeper-0)
2019-02-15 19:36:35,590 [myid:1] - ERROR [LearnerHandler-/xxx.11:49368:LearnerHandler@624] - Unexpected exception causing shutdown while sock still open
java.io.EOFException
	at java.io.DataInputStream.readInt(DataInputStream.java:392)
	at org.apache.jute.BinaryInputArchive.readInt(BinaryInputArchive.java:63)
	at org.apache.zookeeper.server.quorum.QuorumPacket.deserialize(QuorumPacket.java:85)
	at org.apache.jute.BinaryInputArchive.readRecord(BinaryInputArchive.java:99)
	at org.apache.zookeeper.server.quorum.LearnerHandler.run(LearnerHandler.java:534)
2019-02-15 19:36:35,592 [myid:1] - WARN  [LearnerHandler-/xxx.11:49368:LearnerHandler@639] - ******* GOODBYE /xxx.11:49368 ********
2019-02-15 19:36:35,594 [myid:1] - INFO  [NIOWorkerThread-4:ZooKeeperServer@1046] - Client attempting to renew session 0x10005eaf60b26ed at /xxx.12:47044
2019-02-15 19:36:35,595 [myid:1] - INFO  [NIOWorkerThread-6:ZooKeeperServer@1046] - Client attempting to renew session 0x10005eaf60b0016 at /xxx.6:49192
2019-02-15 19:36:35,596 [myid:1] - INFO  [NIOWorkerThread-3:ZooKeeperServer@1046] - Client attempting to renew session 0x30005ec374d211b at /xxx.7:54772
```
``` 
(zookeeper-1)
2019-02-15 19:36:21,809 [myid:2] - WARN  [QuorumPeer[myid=2](plain=/0.0.0.0:2181)(secure=disabled):Follower@96] - Exception when following the leader
java.net.SocketTimeoutException: Read timed out
	at java.net.SocketInputStream.socketRead0(Native Method)
	at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
	at java.net.SocketInputStream.read(SocketInputStream.java:171)
	at java.net.SocketInputStream.read(SocketInputStream.java:141)
	at java.io.BufferedInputStream.fill(BufferedInputStream.java:246)
	at java.io.BufferedInputStream.read(BufferedInputStream.java:265)
	at java.io.DataInputStream.readInt(DataInputStream.java:387)
	at org.apache.jute.BinaryInputArchive.readInt(BinaryInputArchive.java:63)
	at org.apache.zookeeper.server.quorum.QuorumPacket.deserialize(QuorumPacket.java:85)
	at org.apache.jute.BinaryInputArchive.readRecord(BinaryInputArchive.java:99)
	at org.apache.zookeeper.server.quorum.Learner.readPacket(Learner.java:157)
	at org.apache.zookeeper.server.quorum.Follower.followLeader(Follower.java:92)
	at org.apache.zookeeper.server.quorum.QuorumPeer.run(QuorumPeer.java:1214)
2019-02-15 19:36:21,809 [myid:2] - INFO  [QuorumPeer[myid=2](plain=/0.0.0.0:2181)(secure=disabled):MBeanRegistry@128] - Unregister MBean [org.apache.ZooKeeperService:name0=ReplicatedServer_id2,name1=replica.2,name2=Follower]
2019-02-15 19:36:21,809 [myid:2] - INFO  [QuorumPeer[myid=2](plain=/0.0.0.0:2181)(secure=disabled):Follower@201] - shutdown called
java.lang.Exception: shutdown Follower
	at org.apache.zookeeper.server.quorum.Follower.shutdown(Follower.java:201)
	at org.apache.zookeeper.server.quorum.QuorumPeer.run(QuorumPeer.java:1218)
```
 
2) _Controller experiences ZK session problems_: At this point, the ZK clients detects connectivity problems with ZK servers that translates into the state of the current session:
```
2019-02-15 19:36:21,812 109440868 [ControllerServiceMain-SendThread(nautilus-pravega-zookeeper-client:2181)] INFO  org.apache.zookeeper.ClientCnxn - Unable to read additional data from server sessionid 0x10005eaf60b001f, likely server has closed socket, closing socket connection and attempting reconnect
2019-02-15 19:36:21,914 109440970 [ControllerServiceMain-EventThread] INFO  org.apache.curator.ConnectionState - Connection string changed to: nautilus-pravega-zookeeper-client:2181
2019-02-15 19:36:21,916 109440972 [ControllerServiceMain-EventThread] INFO  o.a.c.f.s.ConnectionStateManager - State change: SUSPENDED
2019-02-15 19:36:21,917 109440973 [Curator-ConnectionStateManager-0] INFO  i.p.c.fault.SegmentMonitorLeader - Zookeeper connection state changed to: SUSPENDED
```

3) _Allow Controller to execute `ZKClientFactory.newZooKeeper`_: With this PR, the Controller does not check the ZK `connectString` and is able to continue trying to reconnect to ZK:
```
2019-02-15 19:36:31,920 109450976 [ControllerServiceMain-EventThread] INFO  i.p.c.s.client.StoreClientFactory - Connection Strings: nautilus-pravega-zookeeper-client:2181 vs xxx.8:2181,xxx.11:2181,xxx.13:2181
2019-02-15 19:36:31,920 109450976 [ControllerServiceMain-EventThread] INFO  i.p.c.s.client.StoreClientFactory - canBeReadOnly: false vs false
2019-02-15 19:36:31,920 109450976 [ControllerServiceMain-EventThread] INFO  i.p.c.s.client.StoreClientFactory - sessionTimeout: 10000 vs 10000
2019-02-15 19:36:31,920 109450976 [ControllerServiceMain-EventThread] INFO  i.p.c.s.client.StoreClientFactory - watcher is null: false
```

4) _Eventually, the Controller gracefully restarts_: After several attempts to reconnect to ZK, the Controller restarts and the new instance is able to continue serving IO to clients:
```
(old Controller)
2019-02-15 19:36:31,926 109450982 [Curator-ConnectionStateManager-0] INFO  o.a.zookeeper.ZooKeeperTestable - injectSessionExpiration() called
2019-02-15 19:36:31,926 109450982 [ControllerServiceMain-EventThread] INFO  o.a.c.f.s.ConnectionStateManager - State change: LOST
2019-02-15 19:36:31,926 109450982 [Curator-ConnectionStateManager-0] INFO  i.p.c.fault.SegmentMonitorLeader - Zookeeper connection state changed to: LOST
2019-02-15 19:36:31,926 109450982 [Curator-ConnectionStateManager-0] WARN  i.p.c.fault.SegmentContainerMonitor - Connection to zookeeper lost, attempting to interrrupt the leader thread
2019-02-15 19:36:31,926 109450982 [ControllerServiceMain-EventThread] INFO  org.apache.zookeeper.ClientCnxn - EventThread shut down for session: 0x10005eaf60b001f
2019-02-15 19:36:31,926 109450982 [controllerpool-36] ERROR i.p.c.cluster.zkImpl.ClusterZKImpl - Connection lost with Zookeeper
2019-02-15 19:36:31,926 109450982 [controllerpool-36] INFO  i.p.c.f.ControllerClusterListener - Received error event when monitoring the controller host cluster, ignoring...
2019-02-15 19:36:31,926 109450982 [Curator-LeaderSelector-0] WARN  i.p.c.fault.SegmentMonitorLeader - Leadership interrupted, releasing monitor thread
2019-02-15 19:36:31,928 109450984 [Curator-PathChildrenCache-2] WARN  i.p.c.s.b.ZooKeeperBucketManager - RetentionService: Received connectivity error
2019-02-15 19:36:31,928 109450984 [Curator-PathChildrenCache-0] ERROR i.p.c.cluster.zkImpl.ClusterZKImpl - Connection lost with Zookeeper
2019-02-15 19:36:31,928 109450984 [Curator-PathChildrenCache-0] INFO  i.p.c.fault.SegmentMonitorLeader - Received error event when monitoring the pravega host cluster, ignoring...
2019-02-15 19:36:31,928 109450984 [Curator-PathChildrenCache-1] WARN  i.p.c.server.bucket.BucketService - RetentionService: StreamNotification for connectivity error
2019-02-15 19:36:31,928 109450984 [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - ZK session expired
2019-02-15 19:36:31,929 109450985 [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Stopping ControllerServiceStarter
2019-02-15 19:36:31,931 109450987 [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Awaiting termination of ControllerServiceStarter
2019-02-15 19:36:31,932 109450988 [ControllerServiceStarter STOPPING] INFO  i.p.c.s.ControllerServiceStarter - Initiating controller service shutDown
2019-02-15 19:36:31,932 109450988 [ControllerServiceStarter STOPPING] INFO  i.p.c.s.ControllerServiceStarter - Stopping controller event processors
2019-02-15 19:36:31,933 109450989 [RESTServer STOPPING] INFO  i.p.c.server.rest.RESTServer - Stopping REST server listening on port: 10080
2019-02-15 19:36:31,934 109450990 [GRPCServer STOPPING] INFO  i.p.c.server.rpc.grpc.GRPCServer - Stopping gRPC server listening on port: 9090
2019-02-15 19:36:31,934 109450990 [ControllerServiceStarter STOPPING] INFO  i.p.c.s.ControllerServiceStarter - Stopping the segment container monitor
2019-02-15 19:36:31,935 109450991 [ControllerServiceStarter STOPPING] INFO  i.p.c.s.ControllerServiceStarter - Stopping controller cluster listener
2019-02-15 19:36:31,935 109450991 [ControllerServiceStarter STOPPING] INFO  i.p.c.s.ControllerServiceStarter - Controller cluster listener shutdown
2019-02-15 19:36:31,936 109450992 [ControllerServiceStarter STOPPING] INFO  i.p.c.s.ControllerServiceStarter - Stopping auto retention service
2019-02-15 19:36:31,936 109450992 [ControllerServiceStarter STOPPING] INFO  i.p.c.server.bucket.BucketManager - RetentionService: Stop request received for bucket manager
2019-02-15 19:36:31,936 109450992 [ControllerClusterListener STOPPING] INFO  i.p.c.f.ControllerClusterListener - Deregistering host xxx.4:9090:5c484ee8-90cb-45e0-9802-4c7c273329a3 from controller cluster
2019-02-15 19:36:31,937 109450993 [ControllerServiceStarter STOPPING] INFO  i.p.c.server.bucket.BucketService - RetentionService: Stop request received for bucket 0
...
```
```
(new Controller)
2019-02-15 19:37:33,034 756  [ControllerServiceMain] WARN  o.a.curator.CuratorZookeeperClient - session timeout [10000] is less than connection timeout [15000]
2019-02-15 19:37:33,051 773  [ControllerServiceMain] INFO  o.a.c.f.imps.CuratorFrameworkImpl - Starting
2019-02-15 19:37:33,057 779  [ControllerServiceMain] INFO  org.apache.zookeeper.ZooKeeper - Initiating client connection, connectString=nautilus-pravega-zookeeper-client:2181 sessionTimeout=10000 watcher=org.apache.curator.ConnectionState@18a40023
2019-02-15 19:37:33,067 789  [ControllerServiceMain] INFO  o.apache.zookeeper.ClientCnxnSocket - jute.maxbuffer value is 4194304 Bytes
2019-02-15 19:37:33,079 801  [ControllerServiceMain-SendThread(nautilus-pravega-zookeeper-client:2181)] INFO  org.apache.zookeeper.ClientCnxn - Opening socket connection to server nautilus-pravega-zookeeper-client/yyy.233:2181. Will not attempt to authenticate using SASL (unknown error)
2019-02-15 19:37:33,087 809  [ControllerServiceMain] INFO  o.a.c.f.imps.CuratorFrameworkImpl - Default schema
2019-02-15 19:37:33,088 810  [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Awaiting ZK client connection to ZK server
2019-02-15 19:37:33,089 811  [ControllerServiceMain-SendThread(nautilus-pravega-zookeeper-client:2181)] INFO  org.apache.zookeeper.ClientCnxn - Socket connection established, initiating session, client: /xxx.4:37994, server: nautilus-pravega-zookeeper-client/yyy.233:2181
2019-02-15 19:37:33,099 821  [ControllerServiceMain-SendThread(nautilus-pravega-zookeeper-client:2181)] INFO  org.apache.zookeeper.ClientCnxn - Session establishment complete on server nautilus-pravega-zookeeper-client/yyy.233:2181, sessionid = 0x3000c71a4a50011, negotiated timeout = 10000
2019-02-15 19:37:33,105 827  [ControllerServiceMain-EventThread] INFO  o.a.c.f.s.ConnectionStateManager - State change: CONNECTED
2019-02-15 19:37:33,106 828  [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Awaiting ZK session expiry or termination trigger for ControllerServiceMain
2019-02-15 19:37:33,107 829  [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Starting controller services
2019-02-15 19:37:33,109 831  [ControllerServiceMain] INFO  i.p.c.server.ControllerServiceMain - Awaiting controller services start
2019-02-15 19:37:33,110 832  [ControllerServiceStarter STARTING] INFO  i.p.c.s.ControllerServiceStarter - Initiating controller service startUp
2019-02-15 19:37:33,110 832  [ControllerServiceStarter STARTING] INFO  i.p.c.s.ControllerServiceStarter - Event processors enabled = true
2019-02-15 19:37:33,110 832  [ControllerServiceStarter STARTING] INFO  i.p.c.s.ControllerServiceStarter - Cluster listener enabled = true
2019-02-15 19:37:33,110 832  [ControllerServiceStarter STARTING] INFO  i.p.c.s.ControllerServiceStarter -     Host monitor enabled = true
2019-02-15 19:37:33,110 832  [ControllerServiceStarter STARTING] INFO  i.p.c.s.ControllerServiceStarter -      gRPC server enabled = true
2019-02-15 19:37:33,110 832  [ControllerServiceStarter STARTING] INFO  i.p.c.s.ControllerServiceStarter -      REST server enabled = true
2019-02-15 19:37:33,111 833  [ControllerServiceStarter STARTING] INFO  i.p.c.s.ControllerServiceStarter - Creating the stream store
...
2019-02-18 07:26:01,988 215309710 [grpc-default-executor-3165] INFO  i.p.c.s.r.g.v.ControllerServiceImpl - getURI called for segment longevity/mediumScale/19696720026463.
2019-02-18 07:26:02,220 215309942 [controllerpool-11] INFO  i.p.c.s.r.r.StreamMetadataResourceImpl - Successfully fetched required scaling events for scope: longevity, stream: mediumScale
2019-02-18 07:26:02,261 215309983 [grpc-default-executor-3165] INFO  i.p.c.s.r.g.v.ControllerServiceImpl - getURI called for segment longevity/mediumScale/19722489830248.
```

### CODE REVIEW

File: controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java [hunk 1]
```diff
@@ -115,7 +115,7 @@ public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher
                 this.client = new ZooKeeper(connectString, sessionTimeout, watcher, canBeReadOnly);
                 return this.client;
             } else {
-                Preconditions.checkArgument(this.connectString.equals(connectString), "connectString differs");
+                log.info("Former ZK connection string: {}. New ZK connection string {}.", this.connectString, connectString);
```
fpj: If the connect string differs, then I believe the underlying zk client won't have that connect string (we don't recreate the underlying client). As such, I have two questions:

- Is there any condition under which the connect string actually changes? I'm not talking about giving IP addresses rather than hostnames like described in this PR, but instead really changing the list of servers. A zk ensemble can change its composition with dynamic configuration, but I don't know whether curator will be propagating it through this API.

shiveshr: in this case, say curator propagates new zk ensemble via this api, then we can simply call `client.updateZkServers` with new connection string. 

However, from what i have experimented with, the `newZookeeper` on factory is called only during session "lost" from curator as it attempts to create a new client with a new zksession. 
Since we do not wish to create a new session, any other parameter being sent here is of no interest to us irrespective of whether it is with same or different connection string. so we could ignore the connection string altogether.

RaulGracia:
[ZK-related issue in longevity run](https://trello.com/c/OzWWJqm8/73-zk-related-issue-in-longevity-run)

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
fpj: Here is the `LOST` documentation:

https://curator.apache.org/apidocs/org/apache/curator/framework/state/ConnectionState.html#LOST

and `newZooKeeper` is expected to create a new zk client instance:

https://curator.apache.org/apidocs/org/apache/curator/utils/ZookeeperFactory.html#newZooKeeper-java.lang.String-int-org.apache.zookeeper.Watcher-boolean-

Now, if this is only invoked when the session has expired, then executing this method is pointless, and I'm concerned that we will be leaving the controller instance in a bad state as it will keep trying to get a valid zk client instance and getting a bad zk client instance instead.

 Why don't we simply abort this controller instance in the case it asks for a new zookeeper instance and let the controller instance restart?

shiveshr: Thats exactly what happens - as we are restarting controller, any attempt from curator to silently create a new zookeeper client and hence a session is thwarted by this little trick of having our own factory.

Note: as shutdown is happening, different modules continue their work until it is all cancelled. And if they get a handle to a working zookeeper client during this time, they may be able to perform store updates despite a session expiration.. hence we need to prevent it.

RaulGracia: According to the behavior observed in the longevity run and a discussion with @shiveshr:

If the old `connectString` is invalid after it has changed, then the current Curator client will eventually set the connection state to `LOST` (the ZK ensemble will be unreachable) and this will induce a graceful restart of the Controller. With the restart, a new client will be created with the new `connectString`. This is the current behavior of the Controller with this PR as observed in the longevity run, which I believe that is the expected one. Moreover, I feel this approach more natural than forcing an instance abort directly in `newZookeeper`.

fpj: From what you are saying, if I replace the implementation of that `newZooKeeper` method with:

```
public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher watcher, boolean canBeReadOnly) {
   return null
}
```

I have the precise behavior you are describing and there is no ambiguity. If that's true, then there is no reason for us to try to overengineer that method and potentially create ambiguity.

shiveshr: No this factory is called even for the first time that zookeeper client needs to be created.. so we just need to create one and always return the same.

RaulGracia: Apart from that, we also need to be sure that no exception is raised in that method; otherwise we will encounter the same issue that makes the Controller to get blocked.

shiveshr:
@adrianmo 
> Is there a reason why we attempt to reuse a ZK client object when the session expires? Maybe it would be better to just create a new client object using the original configuration.

Yes, each controller process has cluster membership which is lost whenever zk session expires. Other controller instances attempt to failover outstanding work of this controller instance upon receiving the signal. for this reason we do not want to silently change the zk client and establish a new connection.

The approach we have chosen to follow is to release all held resources and clean up all objects and let effectively let a new controller instance be spawned (as an optimization, we locally shutdown modules and restart the modules without exiting the process). 

When internal restart is attempted, as new modules are created, a new curator (thereby new zk client) is created as well which is then used for communication with zookeeper.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
fpj: We are transforming a precondition into a log message. We are also keeping the precondition checks for all other parameters. If we are to ignore changes to the input, then we need to remove all these precondition equality checks.

One alternative that looks cleaner to me is something like the following:

```
        try {
                Preconditions.checkArgument(this.connectString.equals(connectString), "connectString differs");
                Preconditions.checkArgument(this.sessionTimeout == sessionTimeout, "sessionTimeout differs");
                Preconditions.checkArgument(this.canBeReadOnly == canBeReadOnly, "canBeReadOnly differs");
                Preconditions.checkNotNull(watcher, "watcher");
                this.client.register(watcher);
         } catch (IllegalArgumentException e) {
                log.warn("Input argument for new ZooKeeper client changed with respect to existing client", e);
                this.client.close();
         } finally {
                return this.client;
         }
```

In the case something doesn't match, we make sure that the session is invalid from this point onwards and create no confusion.

RaulGracia:
Yes @fpj, I'm working on this right now.

RaulGracia:
I added 1 test (`StoreClientFactoryTest`) and modified an existing one (`ZKControllerServiceMainTest`) to validate the behavior of both the `StoreClientFactory` and `ControllerServiceMain` in the presence of exceptions due to updated Zookpeer client parameters. Please, let me know if I'm missing something @fpj @shiveshr @adrianmo.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -115,14 +121,51 @@ public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher
...
-                Preconditions.checkNotNull(watcher, "watcher");
-                this.client.register(watcher);
+                try {
+                    tryInjectFailure();
+                    Preconditions.checkArgument(this.connectString.equals(connectString), "connectString differs");
+                    Preconditions.checkArgument(this.sessionTimeout == sessionTimeout, "sessionTimeout differs");
+                    Preconditions.checkArgument(this.canBeReadOnly == canBeReadOnly, "canBeReadOnly differs");
+                    Preconditions.checkNotNull(watcher, "watcher");
+                    this.client.register(watcher);
+                } catch (IllegalArgumentException | NullPointerException e) {
```
fpj: See the comment above, we don't need to catch NPE here.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -115,14 +121,51 @@ public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher
                 this.client = new ZooKeeper(connectString, sessionTimeout, watcher, canBeReadOnly);
                 return this.client;
             } else {
-                Preconditions.checkArgument(this.connectString.equals(connectString), "connectString differs");
-                Preconditions.checkArgument(this.sessionTimeout == sessionTimeout, "sessionTimeout differs");
-                Preconditions.checkArgument(this.canBeReadOnly == canBeReadOnly, "canBeReadOnly differs");
-                Preconditions.checkNotNull(watcher, "watcher");
-                this.client.register(watcher);
+                try {
+                    tryInjectFailure();
```
fpj: One option here is to replace this `tryInjectFailure` call with:

```
Preconditions.checkArgument(this.injectParameterUpdateFailure.get() == true, "...");
```

I'm assuming this variable is an atomic boolean as suggested below, and we wouldn't need `tryInjectFailure` any longer.

I'm wondering if we can't instead mock curator so that we don't have to have any hack in the middle of the code. You are really trying to test two separate behaviors, which are of course connected, but still separate:

1- that `newZooKeeper` throws illegal argument in the case of a parameter change and closes the zk client
2- that the controller reacts correctly to `newZooKeeper` closing the client

If we separate the behavior like that, will it be simpler to test? For (1), we simply invoke `newZooKeeper` with a different parameter compared to the ones of the existing client. For (2), we close the zk client and check that the controller restarts correctly.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -115,14 +121,51 @@ public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher
...
+                log.error("Problems attempting to close ZooKeeper client.", e);
+            }
+        }
+
+        /**
+         * Simulates an error related to a change in ZooKeeper client parameters. The error is injected only once and
+         * this class automatically sets this flag to false when the exception is thrown.
+         */
+        @VisibleForTesting
+        @Synchronized
```
fpj: Also, you may want to use an atomic boolean rather than synchronize the method.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -115,14 +121,51 @@ public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher
...
-                Preconditions.checkArgument(this.sessionTimeout == sessionTimeout, "sessionTimeout differs");
-                Preconditions.checkArgument(this.canBeReadOnly == canBeReadOnly, "canBeReadOnly differs");
-                Preconditions.checkNotNull(watcher, "watcher");
-                this.client.register(watcher);
+                try {
+                    tryInjectFailure();
+                    Preconditions.checkArgument(this.connectString.equals(connectString), "connectString differs");
+                    Preconditions.checkArgument(this.sessionTimeout == sessionTimeout, "sessionTimeout differs");
+                    Preconditions.checkArgument(this.canBeReadOnly == canBeReadOnly, "canBeReadOnly differs");
+                    Preconditions.checkNotNull(watcher, "watcher");
```
fpj: This check is actually not necessary as the watcher is optional and is about the application, not the ZK configuration.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
RaulGracia: Done.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
RaulGracia: I used an `AtomicBoolean` to inject an `IllegalArgumentException` in `StoreClientFactory` as suggested. And, as you correctly described, we are already testing both behaviors separately:

1- `StoreClientFactoryTest:testZkSessionExpiryWithChangedParameters`: This test verifies that in the case of an unexpected parameter change, we handle the exception correctly and close the zk client. Moreover, we also verify that further calls to Curator client return an exception that the Controller is able to handle correctly (`SessionExpiredException`).

2-`ZKControllerServiceMainTest:testZKSessionExpiry`: This test has been modified to also test that the Controller recovers correctly from a ZK session expiration in which ZK client parameters have been updated. This also involves the fact that the ZK client is closed.

For testing both scenarios, we added the `injectParameterUpdateFailure` flag (in a similar spirit to the `injectSessionExpiration()` method in ZK client). I feel that is convenient because it emulates the exact issue we are facing. Specifically, for test case number 2, we could close the zk client from the test code itself. However, we would not be in control of exactly "when" the client is closed: we may be closing it before or after the invocation to `newZooKeeper` upon a session expiration. This is because the invocation to `newZooKeeper` is internally managed by Curator client. This may lead the test to execute a different situation that is not exactly the one that we are facing in the original issue. I thought that exposing the `injectParameterUpdateFailure` flag is an acceptable way to have control over `StoreClientFactory` to reproduce the original issue with precision.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
RaulGracia: Done.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
RaulGracia: Removed check.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -94,11 +102,14 @@ static CuratorFramework createZKClient(ZKClientConfig zkClientConfig, Supplier<B
     }
 
     @VisibleForTesting
-    static class ZKClientFactory implements ZookeeperFactory {
+    public static class ZKClientFactory implements ZookeeperFactory {
```
fpj: We should be able to test this class without making it public. The scope of package protected should be sufficient to test it. Let's rework the test cases so that we don't have to do it.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -115,14 +126,30 @@ public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher
...
                 return this.client;
             } else {
-                Preconditions.checkArgument(this.connectString.equals(connectString), "connectString differs");
-                Preconditions.checkArgument(this.sessionTimeout == sessionTimeout, "sessionTimeout differs");
-                Preconditions.checkArgument(this.canBeReadOnly == canBeReadOnly, "canBeReadOnly differs");
-                Preconditions.checkNotNull(watcher, "watcher");
-                this.client.register(watcher);
+                try {
+                    // Simulates an error related to a change in ZooKeeper client parameters.
+                    Preconditions.checkArgument(this.injectParameterUpdateFailure.get(), "Simulating a parameter update in ZooKeeper client.");
```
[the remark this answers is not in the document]
RaulGracia: You are right. In an attempt to address your comments, I actually removed the "injection" of failures in this class and created a mock class (extending this one) that can be instructed to close the client after executing the `newZooKeeper` method.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
RaulGracia: At the moment, I could restrict the visibility of this nested class to `protected`, so a mock class for test can extend its functionality and close the ZK client when desired in a test. Let's see if this solution is better than the previous one.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -102,14 +101,13 @@ public static CuratorFramework createZKClient(ZKClientConfig zkClientConfig, Sup
     }
 
     @VisibleForTesting
-    public static class ZKClientFactory implements ZookeeperFactory {
+    protected static class ZKClientFactory implements ZookeeperFactory {
```
fpj: I was suggesting that we simply do not change the modifier and leave it package protected. My suggestion is that we change the signature back to:

```
static class ZKClientFactory implements ZookeeperFactory
```

RaulGracia: The idea behind setting `ZKClientFactory` as `protected` was to be able of creating a mock object extending `StoreClientFactory` that could close the ZK client right after the invocation to `newZookeeperClient`. However, I changed that and simplified the testing:
1) `StoreClientFactoryTest`: Emulates the change of a ZK connection parameter and checks that the client is correctly closed when invoking `newZookeeperClient`, and further requests to the ZK client return a `SessionExpiredException`.
2) `ZKControllerServiceMainTest`: Now, this test also checks that the Controller can restart with the ZK client being closed, which is what happens when any of the ZK parameters change and `newZookeeperClient` is invoked (we close the ZK client after injecting a session expiration from the test itself).

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
```diff
@@ -108,19 +117,31 @@ public ZooKeeper newZooKeeper(String connectString, int sessionTimeout, Watcher
...
+            }
+            return this.client;
+        }
+
+        private void closeClient(ZooKeeper client) {
+            try {
+                client.close();
+            } catch (Exception e) {
+                // We prevent throwing uncontrolled exceptions here, which may lead Curator to retry indefinitely.
+                log.error("Problems attempting to close ZooKeeper client.", e);
```
fpj: Say instead:

"Exception while attempting to close ZooKeeper client."

controller/src/test/java/io/pravega/controller/server/ZKControllerServiceMainTest.java
```diff
@@ -91,105 +91,120 @@ protected void shutDown() throws Exception {
...
     }
 
-    @Test(timeout = 100000)
+    /**
+     * This test verifies that the Controller mechanism for handling a ZK session expiration works correctly. Moreover,
+     * this verification is done both assuming that Zookeeper client connection parameters do not change and simulating
+     * a change on them.
+     */
+    @Test(timeout = 120000)
     public void testZKSessionExpiry() throws Exception {
```
fpj: The changes to this test case are making it super long. We don't have to do it in this PR, as it would be good to unblock these changes, but I'd recommend creating an issue to improve it in the future and separate the test cases.

controller/src/main/java/io/pravega/controller/store/client/StoreClientFactory.java
RaulGracia: Done.

controller/src/test/java/io/pravega/controller/server/ZKControllerServiceMainTest.java
RaulGracia: Here is the issue: https://github.com/pravega/pravega/issues/3449

PR #3312, opened 2019-01-25
Issue 3278: List Streams in scope

shiveshr:
Signed-off-by: Shivesh Ranjan <shivesh.ranjan@gmail.com>

**Change log description**  
Add listStreamsInScope api to `StreamManager` and `Controller Grpc`

**Purpose of the change**  
Fixed #3278 

**What the code does**  
This PR adds a new api to Controller client (grpc) and StreamManager interfaces. This is paginated api on controller client.

* `StreamManager` returns an iterator _I_ to the application.
* Controller client returns an async iterator _AI_ that _I_ builds upon. 
* Grpc returns a continuation token, which _AI_ uses to fetch next batch.

On the service side, instead of fetching all streams under scope, we fetch only a limited number of streams and use the last fetched stream as the continuation token. The streams in scope are added as a separate listing stored under a hierarchical znode formed of 2 digists + 4 digits. 
A continuation token is essentially a serialized integer that is broken into three parts - 
msb 2 bytes, middle 4 bytes and lsb 4 bytes. Stream references are stored as /01/2345/stream-6789. 
So effectively all streams under the scope are ordered by the stream position. Stream position is monotonically increasing and any new stream that is added to the scope will be done at a higher position.  
So now the continuation token basically signifies the position of last stream that was returned in previous iteration. 
And continuing from the continuation token, we simply retrieve next `limit` number of streams under the scope. 
Since the streams are stored under the aforesaid hierarchy, we start with all children of the `middle` part of continuation token and only include streams whose position is greater than token.
If we are not able to get `limit` number of streams from this, we go to the next higher available `middle` znode and fetch its children. If middle is exhausted, we increment the `msb` and repeat until we have found `limit` number of streams or reached the end. 

**How to verify it**  
unit tests added

Issue #3278, opened 2019-01-14
Add listStreams() method to the gRPC client


bourgeoisor:
**Problem description**

The current implementation of the StreamManager for Java allows users to manage scopes and streams, but is missing an important method: listing streams. This is particularly useful when wanting to delete a scope that contains streams (this fails and streams need to be manually purged). There is a `.deleteStream()` method in the implementation but no methods to actually list out the current streams.

This is available to the REST API, but not to the Java gRPC-enabled client.

**Problem location**

`io.pravega.client.admin.impl.StreamManagerImpl`

**Suggestions for an improvement**

Add the following method to the implementation of the StreamManager for the Java client:

`streamManager.listStreams()`

shiveshr:
Since we have createScope and deleteScope available in grpc, so i dont see any reason why we should not have listStreamsInScope in grpc. 
I will add the grpc. @tkaitchuck @shrids IMO this api also ought to be exposed on StreamManager interface and i can make that change as well unless there is any objection to it.

shiveshr:
![](https://github.trello.services/images/mini-trello-icon.png) [List Streams in scope with pagination](https://trello.com/c/XBbsA0Hn/99-list-streams-in-scope-with-pagination)

PR #3392, opened 2019-02-20
[WIP] Issue 2895:  Enable TableStore iterator apis on SegmentHelper.

### LINKED ISSUE #2895: Client and Wire Protocol for Tables
**Problem description**
Implement Client code and supporting Wire Protocol for Tables. Contracts defined in #2876 .
- TableSegment
- ClientFactory

**Problem location**
Client and Wire Protocol.

shrids:
Note: This PR depends on PR #3312 for `io.pravega.common.util.ContinuationTokenAsyncIterator`

**Change log description**  
Enable TableStore iterator apis on the SegmentHelper.

**Purpose of the change**  
Fixes #2895 

**What the code does**  
This PR adds helper methods to enable the controller to access the table store iterator apis.

**How to verify it**  
All the newly added tests and existing tests should continue to pass.


shrids:
[Issue 2895: Enable TableStore iterator apis on SegmentHelper](https://trello.com/c/95H650Mv/127-issue-2895-enable-tablestore-iterator-apis-on-segmenthelper)

client/src/main/java/io/pravega/client/stream/impl/Controller.java
```diff
@@ -42,6 +44,14 @@
      */
     CompletableFuture<Boolean> createScope(final String scopeName);
 
+    /**
+     * Gets an async iterator on streams in scope.
+     *
+     * @param scopeName The name of the scope for which to list streams in.
+     * @return A future which when completed will contain an iterator for all Streams in the scope.
```
andreipaduroiu: This doesn't return a future

common/src/main/java/io/pravega/common/util/BlockingAsyncIterator.java
```diff
@@ -0,0 +1,68 @@
...
+ * This class is a wrapper over async iterator that implements java.util.Iterator interface.
+ * This is blocking wrapper on async iterator's getNext method which implements blocking hasNext and next methods of 
+ * iterator interface. 
+ * This class is threadsafe but that has the drawback of blocking on asyncIterator's getNext within a lock. 
+ */
+@ThreadSafe
+@VisibleForTesting
+public class BlockingAsyncIterator<T> implements Iterator<T> {
+    private final AsyncIterator<T> asyncIterator;
+    private T next;
```
andreipaduroiu: GuardedBy("$lock")

common/src/main/java/io/pravega/common/util/BlockingAsyncIterator.java
```diff
@@ -0,0 +1,68 @@
...
+    private boolean canHaveNext;
```
andreipaduroiu: GuardedBy("$lock")

common/src/main/java/io/pravega/common/util/BlockingAsyncIterator.java
```diff
@@ -0,0 +1,68 @@
...
+    public BlockingAsyncIterator(AsyncIterator<T> asyncIterator) {
+        this.canHaveNext = true;
+        this.next = null;
+        this.asyncIterator = asyncIterator;
+    }
+
+    @Synchronized
+    private void load() {
+        if (next == null && canHaveNext) {
+            next = asyncIterator.getNext().join();
```
andreipaduroiu: Waiting on a future while holding a lock is never a good idea. Here's my suggestion for avoiding this:
- make `next` a `CompletableFuture<T>` and still guard it by a lock
- In this method
    - While holding the lock, check `next` for null, and if so, then set it to `asyncInterator.getNext()` (no `join`)
    - After releasing the lock, invoke `join` on `next` and inspect its result.

You may have to do a bit more bookkeeping (in this method) to track whether you actually invoked `getNext` or not.

I think you can also get rid of `canHaveNext`. If `next` is completed (it should always be once this method exits), if its contents is `null`, then `hasNext()` should return `false` and `next()` should throw.

common/src/main/java/io/pravega/common/util/ContinuationTokenAsyncIterator.java
```diff
@@ -0,0 +1,110 @@
...
+
+@ThreadSafe
+@Slf4j
+/**
+ * This is a continuation token based async iterator implementation. This class takes a function that when completed will 
+ * have next batch of results with continuation token. 
+ * This class determines when to call the next iteration of function (if all existing results have been exhausted) and 
+ * ensures there is only one outstanding call. 
+ */
+public class ContinuationTokenAsyncIterator<T, U> implements AsyncIterator<U> {
```
andreipaduroiu: Please use more specific names for the type args. `T` and `U` are not very indicative.

Also, doesn't checkstyle complain about not having `@param` args for these type params?

common/src/main/java/io/pravega/common/util/ContinuationTokenAsyncIterator.java
```diff
@@ -0,0 +1,110 @@
...
+    private final Object lock = new Object();
+    @GuardedBy("lock")
+    @VisibleForTesting
+    @Getter(AccessLevel.PACKAGE)
```
andreipaduroiu: The Lombok getter is not thread safe. Write your own getter where you take care of this.

Also you should make a copy of it in your getter. Even if you return the reference to the queue, the contents can still be iterated and/or modified externally, without holding the lock.

common/src/main/java/io/pravega/common/util/ContinuationTokenAsyncIterator.java
```diff
@@ -0,0 +1,110 @@
...
+    private final Queue<U> queue;
+    @GuardedBy("lock")
+    @VisibleForTesting
+    @Getter(AccessLevel.PACKAGE)
+    private final AtomicReference<T> token;
```
andreipaduroiu: Why do you have an `AtomicReference` that is guarded by a synchronized lock?

And why do you expose the whole `AtomicReference` to the outside via a getter? You should have explicit methods that access only its contents.

common/src/main/java/io/pravega/common/util/ContinuationTokenAsyncIterator.java
```diff
@@ -0,0 +1,110 @@
...
+    private final Function<T, CompletableFuture<Map.Entry<T, Collection<U>>>> function;
+    @GuardedBy("lock")
+    private CompletableFuture<Void> outstanding;
+    private final AtomicBoolean canContinue;
+    private final AtomicBoolean isOutstanding;
+    
+    public ContinuationTokenAsyncIterator(Function<T, CompletableFuture<Map.Entry<T, Collection<U>>>> function, T tokenIdentity) {
```
andreipaduroiu: 1. Check for nulls in the args.
2. Javadoc for constructor

common/src/main/java/io/pravega/common/util/ContinuationTokenAsyncIterator.java
```diff
@@ -0,0 +1,110 @@
...
+        this.isOutstanding = new AtomicBoolean(false);
+    }
+
+    @Override
+    public CompletableFuture<U> getNext() {
+        final T continuationToken;
+        boolean toCall = false;
+        synchronized (lock) {
+            // if the result is available, return it without making function call
+            if (!queue.isEmpty()) {
```
andreipaduroiu: So you check this queue here and if it's not empty, you return the first element, else you invoke the function and add to the queue and reinvoke this method.

I think you can greatly simplify this class by removing the queue altogether and invoking your function (in a loop, maybe) and returning the result. You can also get rid of a few other fields as well.

common/src/test/java/io/pravega/common/util/IteratorTest.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.stream.IntStream;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.atLeast;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+public class IteratorTest {
```
andreipaduroiu: This should be named `ContinuationTokenIteratorTest`

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1035,6 +1044,199 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
+                                                                                    final ConnectionFactory clientCF,
+                                                                                    final String delegationToken,
+                                                                                    final long clientRequestId) {
+
+        final Controller.NodeUri uri = getSegmentUri(scope, stream, 0L, hostControllerStore);
+        final String qualifiedName = getScopedStreamName(scope, stream);
+        final WireCommandType type = WireCommandType.READ_TABLE_KEYS;
+        final long requestId = (clientRequestId == RequestTag.NON_EXISTENT_ID) ? idGenerator.get() : clientRequestId;
+
+        return new ContinuationTokenAsyncIterator<>(token -> {
```
andreipaduroiu: You have a lambda that contains an anonymous implementation of an interface. Let's refactor this code and make this a separate method, and the `FailingReplyProcessor` its own sub-class. That will make reading this much easier.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1035,6 +1044,199 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
+                                                                               final ConnectionFactory clientCF,
+                                                                               final String delegationToken,
+                                                                               final long clientRequestId) {
+
+        final Controller.NodeUri uri = getSegmentUri(scope, stream, 0L, hostControllerStore);
+        final String qualifiedName = getScopedStreamName(scope, stream);
+        final WireCommandType type = WireCommandType.READ_TABLE_ENTRIES;
+        final long requestId = (clientRequestId == RequestTag.NON_EXISTENT_ID) ? idGenerator.get() : clientRequestId;
+
+        return new ContinuationTokenAsyncIterator<>(token -> {
```
andreipaduroiu: Same here.

controller/src/main/java/io/pravega/controller/store/stream/Scope.java
```diff
@@ -39,6 +43,35 @@
...
     CompletableFuture<Void> deleteScope();
 
+    /**
+     * Api to add stream under a scope for listing streams. 
+     * 
+     * @param name name of stream
+     * @param creationTime  creation time of stream.
+     * @return Future, which upon completion will indicate that the stream has been added to the scope. 
+     */
+    CompletableFuture<Void> addStreamToScope(final String name, final long creationTime);
```
andreipaduroiu: How are these related to this pull request?

controller/src/main/java/io/pravega/controller/store/stream/Scope.java
```diff
@@ -52,4 +85,26 @@
      */
     void refresh();
 
+    static String encodeStreamInScope(String name, long creationTime) {
```
andreipaduroiu: These 3 methods would look much nicer if encapsulated (non-statically) into their own class:
- Two constructors: one with name and creation time and one with encoded name. Both constructors would immediately encode and store in a field.
- The class would implement comparable, which is what you have in `compareStreamInScope`.


Another question? Where is this encoding going? Is it stored anywhere? If so, have you considered versioning it?

controller/src/main/java/io/pravega/controller/store/stream/StreamMetadataStore.java
```diff
@@ -189,6 +190,19 @@
...
 
+    /**
+     * List existing streams in scopes with pagination. This api continues listing streams from the supplied continuation token
+     * and returns a count limited list of streams and a new continuation token.
+     *
+     * @param scopeName Name of the scope
+     * @param continuationToken continuation token
+     * @param limit limit on number of streams to return.
+     * @param executor 
+     * @return A map of streams in scope to their configurations
```
andreipaduroiu: This does not return a map.

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
```diff
@@ -40,13 +60,90 @@ public String getName() {
...
         return store.getChildren(scopePath);
     }
 
     @Override
-    public void refresh() {
+    public CompletableFuture<Pair<List<String>, String>> listStreamsInScope(int limit, String continuationToken, Executor executor) {
+        AtomicInteger set = new AtomicInteger(0);
+        AtomicReference<String> floor = new AtomicReference<>();
+        if (!Strings.isNullOrEmpty(continuationToken)) {
+            Pair<String, Long> decoded = Scope.decodeStreamInScope(continuationToken);
```
andreipaduroiu: If this continuation token is external, it needs to be versioned, Please see my comment in `Scope.decodeStreamInScope`

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
```diff
@@ -40,13 +60,90 @@ public String getName() {
...
+    private int findSetForStream(String stream) {
+        int hash = stream.hashCode();
+
+        return hash == Integer.MIN_VALUE ? 0 : Math.abs(hash) % SET_COUNT;
     }
 
+    private CompletableFuture<List<String>> getStreamsInSet(Integer set, int limit, String continuationToken) {
+        String setPath = getSetPath(set);
+        return store.getChildren(setPath)
+                    .exceptionally(e -> {
```
andreipaduroiu: Use `Futures.exceptionallyExpecting`

controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -30,10 +30,10 @@
 @Slf4j
 public final class Config {
     private final static com.typesafe.config.Config CONFIG = ConfigFactory.defaultApplication()
-            .withFallback(ConfigFactory.defaultOverrides().resolve(ConfigResolveOptions.defaults().setAllowUnresolved(true)))
-            .withFallback(ConfigFactory.systemEnvironment())
-            .withFallback(ConfigFactory.defaultReference())
-            .resolve(ConfigResolveOptions.defaults().setAllowUnresolved(true));
+                                                                          .withFallback(ConfigFactory.defaultOverrides().resolve(ConfigResolveOptions.defaults().setAllowUnresolved(true)))
```
andreipaduroiu: This change seems unnecessary

andreipaduroiu:
I just realized this is built on top of #3312. I added the relevant comments there too.

controller/src/main/java/io/pravega/controller/store/stream/Scope.java
shiveshr: this is coming from issue #3278

controller/src/main/java/io/pravega/controller/store/stream/Scope.java
shiveshr: @shrids please update the PR with latest from PR #3312 as these methods have been removed.

controller/src/main/java/io/pravega/controller/store/stream/StreamMetadataStore.java
shiveshr: Fixed in the branch for #3312

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
shiveshr: It has been made versioned in branch #3312

common/src/main/java/io/pravega/common/util/ContinuationTokenAsyncIterator.java
shiveshr: answered on the original branch #3312 

@shrids lets have that branch merged before we share the review here as managing comments on two PRs for same change set becomes unwieldy.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: On second thoughts, maybe we dont need to use ContinuationTokenAsync iterator here. 
Instead, you can have a simple iterator: 
```AsyncIterator<Map.Entry<IteratorState, Collection<TableSegment.IteratorItem<TableKey<byte[]>>>>>```
 that, as part of each getNext returns the pair of token and next batch of entries. 

That way you dont need to depend on other PR and the api here becomes explicit that it is giving out a token and a batch of entries rather than an "iterator".

shiveshr: the advantage is, if i want to get all keys, i can wrap it up in a continuation token Async Iterator. 
if i want to work only on a subset of keys, i can use the result as received by directly calling the api and working with a batch of results.. 
it will totally depend on the business logic that is interested in using these apis.. and there could be different ways it could be invoked in different places.

shrids: Thanks @shiveshr , will open a separate PR with the updated api.

shrids:
Closing this PR. Created a new PR https://github.com/pravega/pravega/pull/3396 which does not use ContinuationTokenAsyncIterator.

shrids:
[Issue 2895: Enable TableStore iterator apis on SegmentHelper](https://trello.com/c/95H650Mv/127-issue-2895-enable-tablestore-iterator-apis-on-segmenthelper)

PR #3396, opened 2019-02-21
Issue 2895: Enable TableStore iterator apis on SegmentHelper.

shrids:
**Change log description**  
Enable TableStore iterator apis on the SegmentHelper.

**Purpose of the change**  
Fixes#2895

**What the code does**  
- This PR adds helper methods to enable the controller to access the table store iterator apis.
- It also ensures that the usage of `io.netty.buffer.ByteBuf#array` is used only if the backing array is present. If the backing array is not present (off heap buffer) then the netty api can throw an `UnsupportedOperationException`. Also, the first byte `io.netty.buffer.ByteBuf#array` need not correspond to the first byte of the buffer since the `ByteBuf` can be a slice of other buffer or a pooled Buffer.

**How to verify it**  
All the newly added tests and existing tests should continue to pass.

### DIFF
--- a/client/src/main/java/io/pravega/client/tables/impl/IteratorState.java
+++ b/client/src/main/java/io/pravega/client/tables/impl/IteratorState.java
@@ -9,17 +9,21 @@
  */
 package io.pravega.client.tables.impl;
 
-import java.nio.ByteBuffer;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
 
 /**
  * Defines the state of a resumable iterator.
  */
-interface IteratorState {
+public interface IteratorState {
+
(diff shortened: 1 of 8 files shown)


client/src/main/java/io/pravega/client/tables/impl/IteratorState.java
```diff
@@ -9,25 +9,33 @@
...
+public interface IteratorState {
+
+    IteratorState EMPTY = new IteratorStateImpl(Unpooled.EMPTY_BUFFER);
 
     /**
      * Serializes the IteratorState instance to a compact byte array.
+     * @return byte representation..
      */
-    ByteBuffer toBytes();
+    ByteBuf toBytes();
```
shiveshr: any specific reason to choose netty byteBuf to nio.ByteBuffer?

PR #3406, opened 2019-02-26
Issue 489: Standardize Controller Config

### LINKED ISSUE #489: Configuration approach mismatch between controller and host
**Problem description**
This covers two issues relevant to config.
1. The implementation of configuration has evolved completely independently in Pravega host and controller. The host uses Java properties based approach where as the controller uses typesafe config. Because of this any config related changes need to be done twice. It will help to finalize on one way of handling config
2.  Handing config in docker containers. The overall strategy of handling config with in a docker container is through env vars. Again the approach here has evolved independently. For controller making each config available through the env needs code change. Again it will help to finalize on one automated way of handling config. 

**Problem location**
Config.java and ComponentConfig.java

andreipaduroiu:
**Change log description**  
- Changed the Controller configuration to use the same approach as in the Segment Store.
- Retired usage of `com.typesafe` along the way.

**Purpose of the change**  
Fixes #489.

**What the code does**  
- Updated `Config.java` to use the configuration API in common to load controller configuration.
- Configuration is fetched from environment variables, system properties, external file and defaulted to hardcoded values if none of that is supplied.
    - The only difference between this and the Segment Store approach is that this allows value references to other properties. This is a backward compatibility since old config properties had some sort of environment variable references attached to them.

**How to verify it**  
Tests must pass.


controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -125,14 +287,17 @@ public static GRPCServerConfig getGRPCServerConfig() {
                 .build();
     }
 
-    public static MetricsConfig getMetricsConfig() {
+    private static MetricsConfig createMetricsConfig(Properties p) {
```
shiveshr: If i remember correctly, we needed this method because we had our metrics in the config with a path prefix. We could use the same metrics name as used in segment store

andreipaduroiu:
@fpj System tests passed yesterday.

PR #3409, opened 2019-02-26
Issue #3219: Prevent stale connections for sparse writers.

### LINKED ISSUE #3219: Reduce number of connections for sparse writers
**Problem description**
If an application has a small number of RoutingKeys, there may be a segment that has no events being written to some parts of keyspace. In this case the application may be holding open more connections than necessary because the segment may have been merged. Consider a segment that is split into two, but one half receives no traffic and is later merged into an adjacent segment. In this case the writer would create a connection for the new segment, but never close it, because it never received a segment sealed after it was merged because it never tried to write any data.

**Problem location**
EventStreamWriterImpl

There are two not mutually exclusive solutions:
 * When SegmentOutputStreamImpl is idle it should periodically check if its segment is sealed. This could be implemented by using the background threadpool to get issue a request to get info on the segment. If the segment the reply indicates the segment is sealed **and** no append requests have been issued since the getInfo, then it can treat this as a segment sealed. Or we could do this server side by having the server propagate a segment sealed reply to any connections with an setup Append when the segment is sealed.
 * We could make instantiating the SegmentOutputStreamImpl for a segment lazy. So when a segment splits the new entries are added to the StreamSegments but the writer itself doesn't get created until data is actually sent to it.

tkaitchuck:
**Change log description**  
* When flush is triggered from a segment seal, send a dummy event to detect if the segment is sealed.

**Purpose of the change**  
* Fixes #3219 

**What the code does**  
Currently if a segment is sealed but the writer never writes any data to that segment (for example because their routing keys don't happen to land there) it will not notice the segment is sealed and end up keeping the connection around until the writer is closed. To some applications that sparsely write to keyspace and experience frequent scaling events this looks like a connection leak. 

The solution here is to send a zero byte append (without any header) to the segment when it is flushed. This won't actually end up writing anything to the segment, but it will force the server to respond with either an ACK or a SegmentIsSealed response, which should allow the client to move forward to newer segments.


client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -158,7 +159,16 @@ private void handleLogSealed(Segment segment) {
                               * inflight that will need to be resent to the new segment when the write lock
                               * is released. (To preserve order)
                               */
-                             flushInternal();
+                             for (SegmentOutputStream writer : selector.getWriters()) {
+                                 try {
+                                     writer.write(PendingEvent.withoutHeader(null, ByteBufferUtils.EMPTY, null));
```
shrids: Here, we are writing a zero byte append to all the segments when one of the segments receives a segment sealed response from the segment store. 
This approach implies that the EventStreamWriter would wait for the response from Segment store (a DataAppended , SegmentIsSealed) for each of the open segments. If a segment sealed is obeserved for any of the individual segment writers then EventStreamWriter would send zero byte appends to all the available segment writers again.

Also, does writing the zero byte appends to all the current segment writers before invoking a flush on all of the them help speed up things?

tkaitchuck: No, it will be slower in the case where there are no inflight events. In the case that there are or that there are some in the buffer it won't change because the 0 byte event will be lumped into the open batch (taking up no space). Hence zero extra bytes will have to go over the wire.

shrids:
https://asd-nautilus-jenkins.isus.emc.com/job/pravega-branch-builder/924/

shrids:
The system test run was successful https://asd-nautilus-jenkins.isus.emc.com/job/test-pravega-continuous/1751/ 
Note: ReadTxnWriteScaleWithFailoverTest, ReadTxnWriteAutoScaleWithFailoverTest, and MetadataScalability tests were commented out due to know issues.

shrids:
The system test run with txn tests enabled also worked fine https://asd-nautilus-jenkins.isus.emc.com/job/test-pravega-continuous-1/18/

PR #3418, opened 2019-03-04
Issue 3427: External connectivity error in PKS

### LINKED ISSUE #3427: External connectivity error for Pravega in PKS NSX-T environment.
**Problem description**
There is an error `Error: Could not find or load main class 100.64.192.23` in Segmentstore when Pravega external connectivity is enabled in PKS environment. 

The reason is that NSX-T adds a tier-1 router uplink address to load balancer ingress IP for network policy configuration. The current scripts we have don't expect to receive two IP addresses. 

**Problem location**
Problem is located in [init_kubernetes.sh](https://github.com/pravega/pravega/blob/master/docker/pravega/scripts/init_kubernetes.sh#L48)

Remove the extra address

Tristan1900:
**Change log description**  
When external connectivity in PKS environment is enabled, the value returned by this [line](https://github.com/pravega/pravega/blob/master/docker/pravega/scripts/init_kubernetes.sh#L48) will have two space-separated IPs such as `IP1 IP2`. There is an error `Error: Could not find or load main class 100.64.192.23` when specifying the space separated IPs in JAVA_OPTS.

The second IP is unique to PKS NSX-T environment. When traffic is forwarded to pods from NSX load balancer, the source IP is the tier-1 router's uplink port's IP address. This address can cause CIDR-based network policy to disallow the traffic, so it needs to be configured to be part of the allowed CIDR block. Therefore this IP is made visible to the user for configuring their network policy. 

I think the extra IP is only for the administrator to use. The change is to ignore the extra IP and pass the rest to Pravega. 

Update:
According to the [documentation](https://docs.vmware.com/en/VMware-NSX-T/2.1/com.vmware.nsxt.ncp_kubernetes.doc/GUID-FF86F9C2-E70D-418B-AFDB-B3CF8DE106C2.html), the first IP is the one that needs to be passed to Pravega.

**Purpose of the change**  
Fix #3427 

**What the code does**  
Ignore the extra IP address that PKS adds to load balancer ingress.

**How to verify it**  
Should pass all the tests

Signed-off-by: wenqi Wenqi.Mou@emc.com


Tristan1900:
@adrianmo Could you please review this? Thanks!

docker/pravega/scripts/init_kubernetes.sh
```diff
@@ -45,13 +45,13 @@ init_kubernetes() {
 
         service_type=$( k8 "${ns}" "services" "${podname}" ".spec.type" )
         if [ "${service_type}" == "LoadBalancer" ]; then
-            export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[].ip" )
+	    export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[0].ip" )
```
fpj: Remove unnecessary spaces.

Tristan1900: Got it. Thanks.

Tristan1900:
> @Tristan1900 is there any documentation that we can check to make sure that the IP at `ingress[0]` is the one that we need to use?

Thanks for the heads up @adrianmo , I am attaching the documentation link to this PR.

Tristan1900:
> We need an issue on the Pravega queue to reflect the work needed on Pravega. Issue 112 can reference such a Pravega issue once created.

Thanks for the guidance @fpj, I have opened an issue in Pravega for this error and updated the reference.

PR #3419, opened 2019-03-05
Issue 3420: Commit order guarantee for Sequential Transactions

### LINKED ISSUE #3420: Transaction commit ordering guarantee
**Problem description**
Since we consolidated multiple transactions in the same epoch at the time of commit, this had a drawback that the ordering of transactions commit did not match the order in which the commits were issued.
This is unintuitive for developers and detrimental to watermarking.
With this change, while keeping the benefits of consolidation, we have guaranteed strong ordering for "sequential" transactions committed on a per writer basis.
Sequential transactions are when a transaction writer doesnot create multiple open transactions and then commits them in arbitrary order, rather follows a scheme of create, commit, create, commit.... For such transactions, we will ensure that the transactions are committed in the order in which their commit was requested.

**Problem location**
Controller Transaction workflow

Implement strict ordering for sequential transactions.

shiveshr:
**Change log description**  

Since we consolidated multiple transactions in the same epoch at the time of commit, this had a drawback that the ordering of transactions commit did not match the order in which the commits were issued. 
This is unintuitive for developers and detrimental to watermarking. 
With this change, while keeping the benefits of consolidation, we have guaranteed strong ordering for "sequential" transactions committed on a per writer basis. 

We say that a set of transactions are sequential when no two transactions in the set are executed concurrently and each transaction begins and commits sequentially.  For such transactions, we ensure that the transactions are committed in the order in which their commit was requested. 

**Purpose of the change**  
Fixes #3420 

**What the code does**  
There are two parts to this change:
1. introduction of writerId and commitTime in transactions.
2. introduction of ordering positions for transactions. 

As transaction commits are issued on a stream, a reference to the transaction id is optimistically added to the ordered set. This ordered set is implemented as a Zookeeper based scheme (ref `ZkOrderedStore`) which uses PersistentSequential znodes to assign order to transactions as their commit is attempted. 

The transaction commit workflow (which works against a commitEvent) used to take requests per epoch and consolidate transactions marked for committing in the epoch and committed them. This could break the order. 
So instead, now we always commit transactions in smallest available epoch first before moving forward. 
So the CommitRequestHandler calls into store to startCommittingTransactions without supplying the epoch. 
The store fetches ordered positions of transactions from the commit order set. It then filters transactions in smallest epoch that are in committing state and orders them by their positions in the ordered set. 

Now by doing this we automatically guarantee order within the epoch. And since we also start with smallest epoch, this means it writer did sequential transactions, their older transactions would be on lower epochs and hence order is not broken. 
Note: for non sequential transactions, we are not guaranteeing the order. 

**How to verify it**  
unit tests added to verify transaction ordering as per expectation. 
I am also adding a new unit for ZkOrderedStore.


controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -335,13 +335,13 @@ private SegmentRange convert(final String scope,
     }
 
     public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String stream, final TxnId
-            txnId) {
+            txnId, String writerId, long time) {
```
tkaitchuck: We should be very explicit on the definition of this time parameter. I've added these same two fields to `commitTransaction` in the watermarking branch, with a different goal. Are these going to be compatible? If so we should make explicit that this is a application defined time.
Also:
Nit: Seems like writerId should go before TxnId to me

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -122,23 +123,23 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
      * an end. However, during failover, once we have created the node, we are guaranteed that it will be only that transaction that will be getting
      * committed at that time.
      */
-    private CompletableFuture<Void> tryCommitTransactions(final String scope,
+    private CompletableFuture<Integer> tryCommitTransactions(final String scope,
```
tkaitchuck: Document this result.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -278,11 +285,29 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
...
     }
 
+    private CompletableFuture<Void> transactionCommitCompletionCallback(String scope, String stream, UUID txnId,
+                                                                        Map<Long, Long> position, OperationContext context) {
+
+        return streamMetadataStore.getTransactionData(scope, stream, txnId, context, executor)
+                .thenCompose(txnRecord -> {
+                    String writerId = txnRecord.getWriterId();
+                    long commitTime = txnRecord.getCreationTime();
+                    // TODO: shivesh: watermarking: generate a writer mark. We have writer id, time and position
```
tkaitchuck: There are multiple comments related to watermarking. Should we target the watermarking branch?

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -564,12 +564,12 @@ public OperationContext createContext(String scope, String name) {
         return withCompletion(getStream(scopeName, streamName, context).checkTransactionStatus(txId), executor);
     }
 
-    @Override
+    @VisibleForTesting
```
tkaitchuck: Does this no longer override anything?

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -680,18 +709,57 @@ private void handleStreamMetadataExists(final long timestamp, CompletableFuture<
...
+                                    if (smallestEpoch.get() > epoch) {
+                                        smallestEpoch.set(epoch);
+                                    }
+                                } else {
+                                    toPurge.add(order);
+                                }
+                                break;
+                            case OPEN:  // do nothing
+                                break;
+                            default:
```
tkaitchuck: I believe we can now use exhaustive enums. This way you can enumerate all the cases here and it will be a compiler error if someone adds a new TxnStatus without updating this code.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1112,21 +1121,45 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
...
+                    time, Long.MIN_VALUE));
+        }
+        return future.thenCompose(updated -> {
+            final Data data = new Data(updated.toBytes(), version);
+            return updateActiveTx(txId, data);  
+        });
     }
 
-    @Override
+    @VisibleForTesting
```
tkaitchuck: Override?

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
```diff
@@ -408,11 +408,13 @@
      * @param txId    transaction identifier.
      * @param commit  whether to commit or abort the specified transaction.
      * @param version optional expected version of transaction data node to validate before updating it.
+     * @param writerId
```
tkaitchuck: This is empty

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
shiveshr: no.. the transactions are comitted in a batch.. so the commit transaction api on the store interface was only used in tests.. so i removed it from the interface and marked this method visible for testing

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: same as above comment about removing from the interface.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
shiveshr: I do not need these fields for transaction ordering.. 
i was updating the transaction metadata (ActiveTxRecord) so i thought i will update it once with all the fields that we know we will need for watermarking. 
So they represent the same thing as has been discussed in watermarking PDP.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
shiveshr: With this branch i have just marked place holders for watermarking. 
This branch only deals with transaction ordering.
The only thing related to watermarking is the `writerId` and `time` fields have also been included in controller metadata and grpc with their default values being sent as of now.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -87,29 +86,34 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
...
+                        if (r >= 0) {
+                            log.debug("Successfully committed transactions on epoch {} on stream {}/{}", r, scope, stream);
+                        } else {
+                            log.debug("No transactions found in committing state on stream {}/{}", r, scope, stream);
+                        }
                         if (processedEvents != null) {
                             processedEvents.offer(event);
                         }
-                        future.complete(r);
+                        future.complete(null);
```
andreipaduroiu: Just a quick note here: it may be possible that `processedEvents.offer` may throw (but double-check), in which case this future would never complete.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1686,23 +1696,41 @@ private StreamSegmentRecord newSegmentRecord(int epoch, int segmentNumber, long
...
 
-    abstract CompletableFuture<Data> getActiveTx(final int epoch, final UUID txId);
+    abstract CompletableFuture<Data> getActiveTx(final UUID txId);
 
-    abstract CompletableFuture<Version> updateActiveTx(final int epoch,
-                                                       final UUID txId,
-                                                       final Data data);
+    abstract CompletableFuture<Version> updateActiveTx(final UUID txId, final Data data);
+
+    // This method is called for adding a transcation to committing order.
```
andreipaduroiu: typo in transaction

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1686,23 +1696,41 @@ private StreamSegmentRecord newSegmentRecord(int epoch, int segmentNumber, long
...
+    abstract CompletableFuture<Version> updateActiveTx(final UUID txId, final Data data);
+
+    // This method is called for adding a transcation to committing order. 
+    // It is important to note that this method is called before marking transaction as committing in active txn record. 
+    // So retrieving a transaction from this ordered list is no guarantee that the transaction has been set to committing. 
+    // Similarly, there could be duplicate entries for the same transaction if retried. 
+    // So the ordered ist merely captures the order in which a request may have been received and is only opportunistic 
+    // and makes no strong claim of its consistency with active txn record. 
+    // 1. we could have duplicate entries for same transaction in commit order.
+    // 2. we could have transcations that are marked for aborting added to commit order
```
andreipaduroiu: and here

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
andreipaduroiu: Also why don't you use Javadoc for method descriptions?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1686,23 +1696,41 @@ private StreamSegmentRecord newSegmentRecord(int epoch, int segmentNumber, long
...
 
     abstract CompletableFuture<Void> createCompletedTxEntry(final UUID txId, byte[] data);
 
     abstract CompletableFuture<Map<String, Data>> getCurrentTxns();
 
-    abstract CompletableFuture<Map<String, Data>> getTxnInEpoch(int epoch);
+    /**
+     * This method finds transactions to commit in lowest epoch and returns a sorted list of transaction ids, 
+     * sorted by their order of commits. 
+     */
```
andreipaduroiu: I'm sure this will introduce a new Javadoc warning due to missing `@return` tag.

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
```diff
@@ -31,13 +31,15 @@
...
 import java.util.concurrent.Executor;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
  * ZK stream metadata store.
  */
 @Slf4j
 class ZKStreamMetadataStore extends AbstractStreamMetadataStore implements AutoCloseable {
+    static final Predicate<Throwable> DATA_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException;
```
andreipaduroiu: FYI, `Futures.exceptionallyExpecting` already unwraps.

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
```diff
@@ -0,0 +1,292 @@
...
+     * @return CompletableFuture which when completed will contain all positions to entities in the set. 
+     */
+    public CompletableFuture<Map<Long, String>> getEntitiesWithPosition(String scope, String stream) {
+        Map<Long, String> result = new ConcurrentHashMap<>();
+        return Futures.exceptionallyExpecting(storeHelper.getChildren(getStreamPath(scope, stream)), DATA_NOT_FOUND_PREDICATE, Collections.emptyList())
+                          .thenCompose(children -> {
+                              // start with smallest queue and collect records
+                              Iterator<Integer> iterator = children.stream().mapToInt(Integer::parseInt).iterator();
+
+                              return Futures.loop(iterator::hasNext, () -> {
```
andreipaduroiu: There is an overload to `Futures.loop` that takes in an iterator and does the legwork for you. I've seen you using this in a few other places too.

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
shiveshr: But this is a predicate that could be used elsewhere too.

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndTransactionOrderTest.java
```diff
@@ -0,0 +1,245 @@
...
+        controllerWrapper.close();
+        autoScaleMonitor.close();
+        clientFactory.close();
+        internalCF.close();
+        connectionFactory.close();
+        server.close();
+        zkTestServer.close();
+    }
+    
+    @Test(timeout = 200000)
```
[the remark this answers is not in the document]
shiveshr: the test takes about 20 seconds on my laptop.. so maybe around 30-40 seconds on travis (i dont know. sometimes travis is really slow). 
so kept the timeout at a larger value in worst case scenario. but it should not really be taking that much time hopefully. 
will reduce it to 100 seconds.

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndTransactionOrderTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private CompletableFuture<Void> waitTillCommitted(Controller controller, Stream s, UUID key, ConcurrentSet<UUID> uncommitted) {
+        AtomicBoolean committed = new AtomicBoolean(false);
+        AtomicInteger counter = new AtomicInteger(0);
+        // check 6 times with 5 second gap until transaction is committed. if it is not committed, declare it uncommitted
+        return Futures.loop(() -> !committed.get() && counter.getAndIncrement() < 5,
+                () -> Futures.delayedFuture(() -> controller.checkTransactionStatus(s, key)
+                                                            .thenAccept(status -> {
+                                                                committed.set(status.equals(Transaction.Status.COMMITTED));
+                                                            }), 5000, executor), executor)
```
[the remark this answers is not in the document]
shiveshr: the problem with that is since we are creating very large number of transcations in this test (around 1000s), we would hammer our threadpool with too many requests if we do it more aggressively.

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
```diff
@@ -0,0 +1,291 @@
...
+                                       if (position > rollOverAfter) {
+                                           // if newly created position exceeds rollover limit, we need to delete that entry 
+                                           // and roll over. 
+                                           // 1. delete newly created path
+                                           return storeHelper.deletePath(positionPath, false)
+                                                             // 2. seal latest queue
+                                                             .thenCompose(v -> storeHelper.createZNodeIfNotExist(
+                                                                     getQueueSealedPath(scope, stream, latestQueueNum)))
+                                                             // 3. call addEntity recursively
+                                                             .thenCompose(v -> addEntity(scope, stream, entity))
```
fpj: The flow implemented makes it a bit hard to do it, but one option to consider is to create the next last queue and add the entity atomically to avoid the recursive call. ZK provides such a feature:

https://zookeeper.apache.org/doc/r3.4.13/api/org/apache/zookeeper/Transaction.html

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -87,29 +86,38 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
...
+                            log.debug("Successfully committed transactions on epoch {} on stream {}/{}", r, scope, stream);
+                        } else {
+                            log.debug("No transactions found in committing state on stream {}/{}", r, scope, stream);
+                        }
                         if (processedEvents != null) {
-                            processedEvents.offer(event);
+                            try {
+                                processedEvents.offer(event);
+                            } catch (Exception ex) {
+                                // ignore, this processed events is only added for enabling unit testing this class
```
fpj: What's `processedEvents`, is that state of the object only used for testing?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1686,23 +1699,45 @@ private StreamSegmentRecord newSegmentRecord(int epoch, int segmentNumber, long
...
-                                                       final UUID txId,
-                                                       final Data data);
+    abstract CompletableFuture<Version> updateActiveTx(final UUID txId, final Data data);
+
+    /**
+     * This method is called for adding a transaction to committing order. 
+     * It is important to note that this method is called before marking transaction as committing in active txn record. 
+     * So retrieving a transaction from this ordered list is no guarantee that the transaction has been set to committing. 
+     * Similarly, there could be duplicate entries for the same transaction if retried. 
+     * So the ordered ist merely captures the order in which a request may have been received and is only opportunistic
```
fpj: Typo: `ist -> list`

fpj: I'm wondering if the following situation is possible:

1- We add txn to commit order
2- We get the smallest epoch and find that txn there
3- But, we can't commit it because it is not in `COMMITTING` state yet
4- We remove it from the commit order because we processed its epoch
5- Txn transitions to `COMMITTING`, but now we lost its order information.

How do we avoid such a scenario?

controller/src/main/java/io/pravega/controller/store/stream/VersionedTransactionData.java
```diff
@@ -21,12 +21,15 @@
 @AllArgsConstructor
 public class VersionedTransactionData {
```
fpj: Is an object of this class persisted anywhere? Do we need to worry about version compatibility?

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -433,21 +448,105 @@
...
                                       e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException, EMPTY_DATA)))
                               ).thenApply(txnMap -> txnMap.entrySet().stream().filter(x -> !x.getValue().equals(EMPTY_DATA))
                                                           .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)))
                       );
     }
 
     @Override
-    CompletableFuture<Version> createNewTransaction(final int epoch, final UUID txId, final byte[] txnRecord) {
-        final String activePath = getActiveTxPath(epoch, txId.toString());
+    public CompletableFuture<List<Map.Entry<UUID, ActiveTxnRecord>>> getOrderedCommittingTxnInLowestEpoch() {
```
fpj: This one looks like an expensive call. Is it going to be a problem?

controller/src/test/java/io/pravega/controller/store/stream/StreamMetadataStoreTest.java
```diff
@@ -914,6 +915,199 @@ public void scaleWithTxnForInconsistentScanerios() throws Exception {
...
+        streamObj.addTxnToCommitOrder(tx02).join();
+
+        Map<Long, UUID> positions = streamObj.getAllOrderedCommittingTxns().join();
+        assertEquals(4, positions.size());
+        assertEquals(positions.get(0L), tx00);
+        assertEquals(positions.get(1L), tx00);
+        assertEquals(positions.get(2L), tx01);
+        assertEquals(positions.get(3L), tx02);
+        
+        // verify that when we retrive transactions from lowest epoch we get tx00
```
fpj: Typo: `retrive -> retrieve`

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
```diff
@@ -0,0 +1,291 @@
...
+    @VisibleForTesting
+    ZkOrderedStore(String ordererName, ZKStoreHelper storeHelper, Executor executor, int rollOverAfter) {
+        this.ordererName = ordererName;
+        this.storeHelper = storeHelper;
+        this.executor = executor;
+        this.rollOverAfter = rollOverAfter;
+    }
+
+    /**
+     * Method to add new entity to the ordered set. Note: Same entity could be added to the set multiple times.
```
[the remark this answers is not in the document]
shiveshr: Ok, the `set` is a misnomer. I should just say an ordered list

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -90,25 +99,29 @@
 
     private final Cache cache;
     private final Supplier<Integer> currentBatchSupplier;
-
+    private final Executor executor;
```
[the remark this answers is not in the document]
shiveshr: We are reusing the executor that is passed to metadata store. 
This uses the executor to loop over multiple `collections` to retrieve all the entries.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
shiveshr: Yes, its only for testing. This was present in CommitEventProcessor and used in its tests. Maybe as a different issue we can clean this up and write our tests differently so as to not include test specific fields in the class.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: no because we dont do `4`. 

We only remove a transaction from committing order once its committed or aborted. 
Other special cases are if the transaction entry in orderer is found to be duplicate. 

The approach is -- 
1. fetch transactions from orderer
2. group transactions by epoch. 
3. within an epoch, fetch transaction status. 
It is important to note that for a writer using sequential txn, if there are multiple transactions in the orderer, all transactions lower than last transaction from the writer cannot be in OPEN state.

controller/src/main/java/io/pravega/controller/store/stream/VersionedTransactionData.java
shiveshr: No, this is only an in-memory object.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: As long as number of outstanding transcations to be committed for a stream are not too high, it will be ok.

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
shiveshr: Actually, we could use transactions, but the recursive call only happens in case the current "set" of almost `0.9 * Integer.Max` entries are exhausted. So the recursion will kick in few times after 1.8Billion transactions are created and committed. so i could open an issue for using zk transactions, but keep things as is for now.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
fpj: If it needs fixing or improving in a separate PR, please create an issue to record it.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
fpj: Ok, so this is O(n) where n is the number of outstanding transactions for a stream.

shiveshr: yes, however we can limit it to a fixed upper bound without breaking any guarantees. 
like say we terminate whenever we have collected 1000 transactions. So the order in average case will approximately be closer to the upper bound.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
shiveshr: https://github.com/pravega/pravega/issues/3566

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
```diff
@@ -0,0 +1,317 @@
...
+     * Returns a map of position to entity that was added to the set. 
+     * Note: Entities are ordered by position in the set but the map responded from this api is not ordered by default.
+     * Users can filter and order elements based on the position and entity id.  
+     * @param scope scope scope
+     * @param stream stream stream
+     * @return CompletableFuture which when completed will contain all positions to entities in the set. 
+     */
+    CompletableFuture<Map<Long, String>> getEntitiesWithPosition(String scope, String stream) {
+        Map<Long, String> result = new ConcurrentHashMap<>();
+        return Futures.exceptionallyExpecting(storeHelper.getChildren(getStreamPath(scope, stream)), DATA_NOT_FOUND_PREDICATE, Collections.emptyList())
```
fpj: One clarification here. We are getting the children of a given stream:

```
storeHelper.getChildren(getStreamPath(scope, stream)
```

and iterating over the children assuming they are all collections, is it really the case, isn't there any other children which are not (parents of) collections?

shiveshr: No, this is getting all children for a "stream" specific znode where all transactions that have been marked for "committing" are stored. 

The hierarchy is as follows:
```
     * This class creates a collection of collections starting with collection numbered 0.
     * A new collection is created under the znode: 
     * \<orderer-root\>/scope/stream/\<collectionNum\>/entities
     * It adds entries to lowest numbered collection using a persistent sequential znode 
     * until its capacity is exhausted. 
     * Once a collection is exhausted, a successor collection with number `current + 1` is created and current successor 
     * is marked as sealed by created following znode: 
     *     `\<orderer-root\>/scope/stream/\<collectionNum\>/sealed
```

So it gets all children of `\<orderer-root\>/scope/stream/` and recursively on 
`entities`. It uses collectionNum+entityNum to create a 64 bit "position" for each transaction.

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
```diff
@@ -0,0 +1,317 @@
...
+
+    /**
+     * Returns a map of position to entity that was added to the set. 
+     * Note: Entities are ordered by position in the set but the map responded from this api is not ordered by default.
+     * Users can filter and order elements based on the position and entity id.  
+     * @param scope scope scope
+     * @param stream stream stream
+     * @return CompletableFuture which when completed will contain all positions to entities in the set. 
+     */
+    CompletableFuture<Map<Long, String>> getEntitiesWithPosition(String scope, String stream) {
```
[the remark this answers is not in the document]
shiveshr: this is the most important method of this class, called from Stream object when it tries to collect all transactions that are in committing state for a given stream. 
Ref: ZkStream.java

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -559,12 +560,12 @@ public OperationContext createContext(String scope, String name) {
...
     public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String streamName,
```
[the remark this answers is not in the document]
shiveshr: its actually used in several tests where we want to commit the transaction directly without going through the whole workflow of commit transactions. So its used outside of the package too.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -77,6 +83,9 @@
     private final Object txnsLock = new Object();
     @GuardedBy("txnsLock")
     private final Map<UUID, VersionedMetadata<ActiveTxnRecord>> activeTxns = new HashMap<>();
+    private final AtomicLong counter = new AtomicLong();
+    @GuardedBy("txnsLock")
+    private final Map<Long, UUID> transactionCommitOrder = new HashMap<>();
```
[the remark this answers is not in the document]
shiveshr: we could do that but we access all transaction related metadata under txnsLock. And for commit order, we do a foreach where we retrieve txnRecord for all transactionIds found in the orderer which accesses ActiveTxn map. 
though we can use concurrentHashMap safely while taking the txnLock whenever retriving the activeTxnRecord.

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
shiveshr: opened new issue #3570

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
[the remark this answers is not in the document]
fpj: Ok, it was probably collapsed when I searched.

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
fpj: Oh I see, you are creating the collections directly under the `scope/stream/` znode. For organization purposes or simply to avoid having the creation of collections affecting the `cversion` of `scope/stream/`, you may want to have the collections under `scope/stream/collections`, in which case the entities path would be: `scope/stream/collections/<collectionNum>`. What do you think?

shiveshr: done

controller/src/main/java/io/pravega/controller/store/stream/ZkOrderedStore.java
TisonKun: Hi @fpj & @shiveshr , I opened new pr #4140 to this discussion.

Some questions were discovered when implementing. Thus I'd like to ask there which operations you want to do in a transaction? Note that we provide fault tolerance if concurrent seal node/entity directory creation occurs, but this is hard(even impossible now) to implement when using ZooKeeper's Transaction.

TisonKun: Besides, if our original purpose is to avoid recursive calls, I provide another possibly solution without necessary involve transaction in #3570 .

PR #3424, opened 2019-03-05
Issue 3422: BucketStore.getBucket must return a positive bucket number

### LINKED ISSUE #3422: Controller bug: Wrap java hashcode with Math.abs in BucketStore
**Problem description**
 Java hashcode can be negative. BucketStore does not wrap it under Math.abs.

**Problem location**
BucketStore

Fix

shiveshr:

**Change log description**  
 * Changes the logic in `BucketStore.getBucket` to handle negative hash codes.

**Purpose of the change**  
Fixes #3422 

**What the code does**  
Bucket store is changed to absolute value of java.hashcode() method

**How to verify it**  

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/BucketStore.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/BucketStore.java
@@ -82,7 +82,8 @@ public interface BucketStore {
 
     static int getBucket(String scope, String stream, int bucketCount) {
         String scopedStreamName = getScopedStreamName(scope, stream);
-        return Math.abs(scopedStreamName.hashCode()) % bucketCount;
+        int hash = scopedStreamName.hashCode();
+        return hash == Integer.MIN_VALUE ? 0 : Math.abs(hash) % bucketCount;
     }
 
     static String getScopedStreamName(String scope, String stream) {
--- a/controller/src/main/java/io/pravega/controller/store/stream/BucketStore.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/BucketStore.java
@@ -82,7 +82,7 @@ public interface BucketStore {
 
     static int getBucket(String scope, String stream, int bucketCount) {
         String scopedStreamName = getScopedStreamName(scope, stream);
-        return scopedStreamName.hashCode() % bucketCount;
+        return Math.abs(scopedStreamName.hashCode()) % bucketCount;
     }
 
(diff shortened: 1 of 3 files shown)


controller/src/main/java/io/pravega/controller/store/stream/BucketStore.java
```diff
@@ -82,7 +82,8 @@
 
     static int getBucket(String scope, String stream, int bucketCount) {
         String scopedStreamName = getScopedStreamName(scope, stream);
-        return scopedStreamName.hashCode() % bucketCount;
+        int hash = scopedStreamName.hashCode();
+        return hash == Integer.MIN_VALUE ? 0 : Math.abs(hash) % bucketCount;
```
fpj: What about this:

```
return (hash ^ 0x7FFFFFFF) % bucketCount;
```

andreipaduroiu: If the purpose is to yield a positive number, then what you are suggesting looks better. It's faster. But if he needs a true `abs`, then what you are proposing will not account for 2's complement (though, IMO, I think just a positive number is good enough).

shiveshr: > return (hash ^ 0x7FFFFFFF) % bucketCount;
I could use `&` instead of `^` (xor). We just need the sign bit to be flipped.

controller/src/main/java/io/pravega/controller/store/stream/BucketStore.java
```diff
@@ -82,8 +82,7 @@
 
     static int getBucket(String scope, String stream, int bucketCount) {
         String scopedStreamName = getScopedStreamName(scope, stream);
-        int hash = scopedStreamName.hashCode();
-        return hash == Integer.MIN_VALUE ? 0 : Math.abs(hash) % bucketCount;
+        return (scopedStreamName.hashCode() & 0x7fffffff) % bucketCount;
```
shrids: `java.lang.Integer.MAX_VALUE`

controller/src/main/java/io/pravega/controller/store/stream/BucketStore.java
fpj: I actually did mean to say or instead of xor, I'm glad you caught the it.

PR #3430, opened 2019-03-07
Call readerOffline before calling reader.close

### LINKED ISSUE #3429: EventProcessor call readerGroup.readerOffline before calling reader.close
**Problem description**
Currently reader.close api uses the latest position of the reader to update to the readergroup state as it goes offline. 
This has a problem that if there was a problem in processing the latest event and we call reader.close, the last event is never read back. 
1. reader.readNextEvent() --> pos1
2. reader.readNextEvent() --> pos2 // process throws exception
3. reader.close()

Upon failover recovery, event at pos2 is never processed. 

**Problem location**
Event processor cell.

Call readerOffline with checkpointed position before calling reader.close.

shiveshr:

**Change log description**  
Call readerOffline with last known position before calling reader.close() so that we continue to use our checkpointed position.

**Purpose of the change**  
Fixes #3429

**What the code does**  
EventProcessorGroup, before shutting down a reader, calls readerOffline with last reported checkpoint. 

**How to verify it**  
Integration test added


controller/src/main/java/io/pravega/controller/eventProcessor/impl/EventProcessorCell.java
```diff
@@ -285,12 +289,18 @@ final void awaitTerminated() {
 
     private EventProcessor<T> createEventProcessor(final EventProcessorConfig<T> eventProcessorConfig) {
         EventProcessor<T> eventProcessor = eventProcessorConfig.getSupplier().get();
-        eventProcessor.checkpointer = (Position position) ->
-        checkpointStore.setPosition(process, readerGroupName, readerId, position);
+        eventProcessor.checkpointer = (Position position) -> {
+            checkpoint.set(position);
```
tkaitchuck: It looks like these two lines should go in the other order. After all you don't want a race where the checkpoint is set but not set on the store.

shiveshr: Actually there is a race but not what you are thinking of. Let me explain:

Actually the store is used for failover recovery by another process that identifies the reader has failed. Store is external (zookeeper) and may not be accessible (say after a session expiration) 

The local "checkpoint" field is for reader to self declare itself offline with the latest checkpoint identifying the position till which the event has been processed before this reader self declares itself offline. 

So the local field is independent of whether we have been able to update the store or not. If the reader shutsdown gracefully, the store does not come into picture.

Now coming to the race. In case of zookeeper session expiration, the controller also loses cluster membership and is declared as failed. Other controller instances will try to recover its work by calling its reader offline at the position from the store.
At the same time this reader is declaring itself offline with its local position. 
But that is fine, because that does not break our at least once processing guarantees even if we recover from an older checkpoint (a la store). 

What you are suggesting is also fine and will guarantee that we only ever use the checkpoints that have been successfully updated in the store. but i was going for slightly more optimal position.

PR #3432, opened 2019-03-07
Issue 3100: Shutdown hook on controller process

### LINKED ISSUE #3100: Add shutdown hook to controller service
**Problem description**
Controller needs to add a new shutdown hook so that it can perform a clean shutdown if a jvm process receives a shutdown signal.
 
**Problem location**
Controller Main

Add a Runtime.shutdown hook.

shiveshr:
**Change log description**  
Added shutdown hook to Main class in Controller Service to handle clean shutdown in case the clustermanager sends kill signal. 

**Purpose of the change**  
fixes #3100 

**What the code does**  
Adds shutdown hook to Main.java which has ControllerServiceMain instance. It sends shutdown signal to ControllerServiceMain which in turn sends shutdown signal to ControllerServiceStarter. A clean shutdown without restart is performed. 

In addition, we also dump memory usage statistics and threaddumps as optional dumps if enabled via a configuration. 

**How to verify it**  
Controller should gracefully shutdown in case of a shutdown signal


controller/src/main/java/io/pravega/controller/server/Main.java
```diff
@@ -86,8 +88,28 @@ public static void main(String[] args) {
...
             controllerServiceMain.startAsync();
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
+                memoryMXBean.setVerbose(true);
+                log.error("Shutdown hook memory usage dump: Heap memory usage: {}, non heap memory usage {}", memoryMXBean.getHeapMemoryUsage(),
+                        memoryMXBean.getNonHeapMemoryUsage());
+                
+                log.info("Controller service shutting down");
+                controllerServiceMain.stopAsync();
+                controllerServiceMain.awaitTerminated();
```
andreipaduroiu: FYI if the Controller will fail (before or during shutdown) this will throw an exception, so the subsequent code won't execute. There is a helper method in `Services` but it converts everything into a Future. An alternative is to put this in a `try-finally` block and move lines 107-110 to the `finally`

controller/src/main/java/io/pravega/controller/server/Main.java
```diff
@@ -86,8 +88,30 @@ public static void main(String[] args) {
...
+            Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
+                log.error("Thread {} with stackTrace {} failed with uncaught exception", t.getName(), t.getStackTrace(), e);              
+            });
+
             ControllerServiceMain controllerServiceMain = new ControllerServiceMain(serviceConfig);
             controllerServiceMain.startAsync();
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
+                memoryMXBean.setVerbose(true);
+                log.error("Shutdown hook memory usage dump: Heap memory usage: {}, non heap memory usage {}", memoryMXBean.getHeapMemoryUsage(),
```
fpj: Why is this an error message?

controller/src/main/java/io/pravega/controller/server/Main.java
```diff
@@ -86,8 +88,30 @@ public static void main(String[] args) {
...
+                        memoryMXBean.getNonHeapMemoryUsage());
+                
+                log.info("Controller service shutting down");
+                try {
+                    controllerServiceMain.stopAsync();
+                    controllerServiceMain.awaitTerminated();
+                } finally {
+                    if (Config.DUMP_STACK_ON_SHUTDOWN) {
+                        Thread.getAllStackTraces().forEach((key, value) ->
+                                log.info("Shutdown Hook Thread dump: Thread {} stackTrace: {} ", key.getName(), value));
```
fpj: This might be a good idea, but let me try to understand the reasoning behind it. Are you assuming that the JVM is shutting down due to an error? The issue talks about a clean shutdown, so I'm trying to match what I'm seeing here to the issue description.

controller/src/main/java/io/pravega/controller/server/Main.java
```diff
@@ -86,8 +88,30 @@ public static void main(String[] args) {
                     .restServerConfig(Optional.of(restServerConfig))
                     .build();
 
+            Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
+                log.error("Thread {} with stackTrace {} failed with uncaught exception", t.getName(), t.getStackTrace(), e);
```
fpj: Is this an error or warn message? It is an error in the case the thread dying is an unrecoverable situation.

controller/src/main/java/io/pravega/controller/server/Main.java
shiveshr: Basically mesos, k8s and other package managers sometime shutdown instances if they cross resource limits. it may be good to know if when resource limits are exceeding, what was the snapshot of threads and memory footprint in general. 
but we dont always want to enable the flag.. only if it happens enough, we could ask customers to turn on the flag and have logs collected in such instances. 

By default in our system tests and longevity, having these dumped is a good idea..

controller/src/main/java/io/pravega/controller/server/Main.java
shiveshr: We are registering callback for `uncaughtException` which essentially means thread faced an erroneous case with an exception that was unhandled.. hence error level logging.

controller/src/main/java/io/pravega/controller/server/Main.java
shiveshr: this should be info. fixed it.

fpj: I don't see it fixed.

shiveshr: done

controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -148,6 +152,7 @@
     private static final Property<Integer> PROPERTY_TXN_TTL_HOURS = Property.named("transaction.ttlHours", 24);
     private static final Property<String> PROPERTY_SCALE_STREAM_NAME = Property.named("scale.streamName", "_requeststream");
     private static final Property<String> PROPERTY_SCALE_READER_GROUP = Property.named("scale.ReaderGroup", "scaleGroup");
+    private static final Property<Boolean> PROPERTY_DUMP_STACK_ON_SHUTDOWN = Property.named("dumpStackOnShutdown", false);
```
fpj: Question: is this matching the new way of setting configuration properties for the controller? Do we need to prefix the property with any name, like `service.`?

controller/src/main/java/io/pravega/controller/server/ControllerServiceMain.java
```diff
@@ -221,4 +221,14 @@ public void forceClientSessionExpiry() throws Exception {
         ((CuratorFramework) this.storeClient.getClient()).getZookeeperClient().getZooKeeper()
                                                          .getTestable().injectSessionExpiration();
     }
+
+    @Override
+    protected void shutDown() throws Exception {
```
fpj: I understand that this is invoked from `stopAsync`? Can we unit test this flow?

controller/src/main/java/io/pravega/controller/util/Config.java
shiveshr: this is same as process level configs for example:
```
    private static final Property<Boolean> PROPERTY_HOST_MONITORING_ENABLED = Property.named("hostMonitorEnabled", true);
```

controller/src/main/java/io/pravega/controller/server/ControllerServiceMain.java
shiveshr: ControllerServiceMain.stop async awaitTerminated is unit tested here:

`ControllerServiceMainTest:testControllerServiceMainStartStop`

fpj: Ok, the code coverage is complaining about the shutdown hook, but I suppose there isn't much sense in unit testing it, is there?

controller/src/main/java/io/pravega/controller/util/Config.java
fpj: For organization purposes, can we group this property with the ones of the same category (process level according to you)?

controller/src/main/java/io/pravega/controller/server/ControllerServiceMain.java
shiveshr: added a test to ControllerServiceMain to verify that onShutdown hook there is clean shutdown of controllerServiceMain.

PR #3441, opened 2019-03-08
Issue 3435: Fixed transaction-related tests to actually use transactional writers

### LINKED ISSUE #3435: Some system tests for transactions are using regular writers
**Problem description**
We realized that some of the system tests assumed to work with Transactions are actually using regular writers. We need to change these tests to use transactional writers. 

**Problem location**
`MultiReaderTxnWriterWithFailoverTest`
`ReadTxnWriteAutoScaleWithFailoverTest`
`ReadTxnWriteScaleWithFailoverTest`

Create transactional writers in these tests.

RaulGracia:
**Change log description**  
Fix transaction-related failover tests so they use transactional writers.

**Purpose of the change**  
Fixes #3435.

**What the code does**  
This PR adds some convenience methods in `AbstractReadWriteTest` to create transactional writers in the same way than regular writers. Moreover, this PR fixes the tests that should be instantiating transactional writers instead of regular ones. 

**How to verify it**  
(Waiting for a system test build to pass, DO NOT MERGE)

### DIFF
--- a/test/system/src/test/java/io/pravega/test/system/AbstractReadWriteTest.java
+++ b/test/system/src/test/java/io/pravega/test/system/AbstractReadWriteTest.java
@@ -335,21 +335,12 @@ abstract class AbstractReadWriteTest extends AbstractSystemTest {
     }
 
     void createWriters(EventStreamClientFactory clientFactory, final int writers, String scope, String stream) {
-        createWritersInternal(clientFactory, writers, scope, stream, testState.writersComplete, false);
+        createWritersInternal(clientFactory, writers, scope, stream, testState.writersComplete, testState.txnWrite);
     }
 
     void addNewWriters(EventStreamClientFactory clientFactory, final int writers, String scope, String stream) {
         Preconditions.checkNotNull(testState.writersListComplete.get(0));
-        createWritersInternal(clientFactory, writers, scope, stream, testState.newWritersComplete, false);
-    }
-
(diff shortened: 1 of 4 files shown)


RaulGracia:
[Some system tests for transactions are using regular writers](https://trello.com/c/xRghr9Sk/142-some-system-tests-for-transactions-are-using-regular-writers)

test/system/src/test/java/io/pravega/test/system/AbstractReadWriteTest.java
```diff
@@ -331,12 +335,21 @@ void createReaders(EventStreamClientFactory clientFactory, String readerGroupNam
     }
 
     void createWriters(EventStreamClientFactory clientFactory, final int writers, String scope, String stream) {
```
shrids: Can we use `TestState.txnWrite ` boolean to decide between txn v/s normal writers when `createWriters` are invoked?
All these tests do a `testState = new TestState(true);` in the @Setup method , which sets the `txnWrite` to true.

RaulGracia: Good point, @shrids. Yes, I will better use that flag which already is there to decide the type of writers to instantiate.

RaulGracia:
@fpj A system test build has been scheduled. I'm waiting for the results to see if the modified tests are passing or not. In the case they don't pass, I will evaluate if it is a problem of the changes of this PR or if we just uncovered a bug that requires a separate bugfix.

PR #3445, opened 2019-03-09
Issue 3444: Fix Abstract Processor Fairness deadlock 

### LINKED ISSUE #3444: Event Processor Fairness deadlock
**Problem description**
- commit-transactions happen on commit-event-processor
- all other request processing happen on request-event-processor

For mutual exclusion, if one processor is processing an event on a stream, it creates a fence by changing the state. Any other processor attempting to process another type of event, is fenced out and retries later. And it could potentially have continued to be fenced out upon each retry if there was a continuous large volume of incoming traffic on the other.
We saw this with large volumes of transactions stalling scale operations and hence we introduced a fairness logic that was as follows:

If a processor cannot process an event for a stream because it is fenced out, it sets a waiting processor flag with its id and postpones event processing..

Before processing a new event, each processor sees if there is a waiting processor and relinquishes processing by postponing it if there is a different processor in waiting.

now this causes deadlock in following situation:
```
1. event processing started on processor 1
2. processor 2 sets waiting processor 
3. processor 1's processing fails with retryable exception even after several retries
(or crashes mid way into processing). event is posted back (repicked) later and
attempted processing. 
4. event for processor 1 is picked again to complete the incomplete processing,
but before it starts processing, it checks the waiting processor flag and it honors
"waiting processor 2" and doesn't start execution. 
5. processor 2 cannot process any event because it is fenced out on metadata
and hence cant remove its waiting flag until it gets an opportunity to at least run once.

so `4` and `5` both back off, one for fairness, other because its fenced out. hence deadlock.
``` 

This was observed in system test run.

**Problem location**
AbstractRequestProcessor

Before processing the event first call a new method called `ignoreWaitingProcessor` which looks at the metadata fence.. and if metadata fence matches the event to be processed, it should ignore any waiting processor

shiveshr:

**Change log description**  
Fixes the deadlock where waiting processor cant make progress because it is fenced out on metadata and processor resuming a previously incomplete processing defers its processing because there is another processor in waiting. 

**Purpose of the change**  
Fixes #3444 

**What the code does**  

 Before processing the event first call a new method called `ignoreFairness` which looks at the metadata fence (state) and if the state matches the processing this event intends to perform, it ignores waiting processor. and if metadata fence matches the event to be processed, it should ignore any waiting processor

**How to verify it**  
Unit tests added.


RaulGracia:
After executing a system test build with this branch, I have found that `readtxnwritescalewithfailovertest` is now passing, but `ReadTxnWriteAutoScaleWithFailoverTest` is still failing (`Scaling did not happen within desired time`).

shiveshr:
only 20 transactions created.. 16 of them committed.. 
for last 4 transctions, after creation, the writers are not able to write to those transactions.. and since the traffic is low, the test fails eventually as it expects auto scale up to happen within some predetermined time.

tkaitchuck:
What happens if there is a third task, which is waiting for some other reason. Then ignore fairness will not apply, but it might still transitively be blocked on the first task

shiveshr:
> What happens if there is a third task, which is waiting for some other reason. Then ignore fairness will not apply, but it might still transitively be blocked on the first task

It is important to note that if processing has started, it should not wait on anything and complete always. 
So either the processing for a task started or it didnt --> if it started, it would have taken a metadata fence via "state". Else no fence is taken. 
If fence is take, ignore all fairness. 
Else it will honour fairness for whichever processor has set the waiting flag. 
It is important to note that we do not have one event processing wait on outcome of another event processing. Each processing is independent and only ever waits for fairness or mutex. 
If we change that then your comment will become relevant.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/StreamTask.java
```diff
@@ -28,4 +28,11 @@
      * @return future of processing
      */
     CompletableFuture<Void> writeBack(T event);
+
+    /**
+     * Method that indicates to the processor if it should ignore any fairness and start processing the task or not. 
+     * @param event event to process
+     * @return Completable Future which when completed will indicate if fairness should be ignore or honored. 
+     */
+    CompletableFuture<Boolean> ignoreFairness(T event);
```
fpj: All implementations of `ignoreFairness` are simply comparing state, but I'm having a hard time convincing myself that this is sufficient to determine that it is the same task as before. Why can't it be a different task that leads to the same state?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -124,10 +124,15 @@ public String getProcessorName() {
...
-                    if (waitingRequestProcessor == null || waitingRequestProcessor.equals(getProcessorName())) {
+        CompletableFuture<String> waitingProcFuture = suppressException(streamMetadataStore.getWaitingRequestProcessor(scope, stream, context, executor), null,
+                "Exception while trying to fetch waiting request. Logged and ignored.");
+        CompletableFuture<Boolean> ignoreFairnessFuture = task.ignoreFairness(event);
+        
+        CompletableFuture.allOf(waitingProcFuture, ignoreFairnessFuture)
+                .thenAccept(v -> {
+                    boolean ignoreFairness = ignoreFairnessFuture.join();
+                    String waitingRequestProcessor = waitingProcFuture.join();
+                    if (ignoreFairness || waitingRequestProcessor == null || waitingRequestProcessor.equals(getProcessorName())) {
```
fpj: `ignoreFairness` sounds bad, what we really need a predicate like `shouldProceed`, which evaluates to `true` in the case that:

- there is no other blocked processor
- there is a blocked processor, which won't be unblocked if we block the present one

and evaluates to `false` otherwise.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/StreamTask.java
shiveshr: > Why can't it be a different task that leads to the same state?

Each type of task has a specific state associated with it. The metadata fence is taken via `state`. 
Also, there is exactly one task of a particular type that can be ongoing (this is determined via task specific record). For example, ongoing scale is identified by `EpochTransitionRecord`, ongoing update is determined by `ConfigurationRecord`, ongoing truncate is identified by `TruncationRecord` etc. 

So task processing happens as follows:
1. update task specific record to ensure only one outstanding task of a particular type exists. 
2. task specific event is posted and then is picked for processing
3. task processing checks if corresponding metadata record is set and then tries to take metadata fence via `state` 

So if `state` matches the state the task expects to set as a fence, then the task processing can run.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
shiveshr: I think a better name would be - `isAlreadyStarted` as that is most explicitly stating what we are doing here.

fpj: I'm fine with the suggestion.

controller/src/test/java/io/pravega/controller/server/eventProcessor/requesthandlers/RequestProcessorTest.java
```diff
@@ -71,9 +72,11 @@ public String getKey() {
 
     public static class TestRequestProcessor1 extends AbstractRequestProcessor<TestEvent1> implements StreamTask<TestEvent1> {
         private final BlockingQueue<TestEvent1> queue;
+        private boolean toIgnoreFairness;
```
fpj: Haven't we changed from `ignoreFairness` to `hasAlreadyStarted`?

controller/src/test/java/io/pravega/controller/server/eventProcessor/requesthandlers/RequestProcessorTest.java
```diff
@@ -71,9 +72,11 @@ public String getKey() {
...
         public TestRequestProcessor1(StreamMetadataStore streamMetadataStore, ScheduledExecutorService executor, BlockingQueue<TestEvent1> queue) {
             super(streamMetadataStore, executor);
             this.queue = queue;
+            this.toIgnoreFairness = false;
```
fpj: Same comment here.

controller/src/test/java/io/pravega/controller/server/eventProcessor/requesthandlers/RequestProcessorTest.java
```diff
@@ -194,4 +208,68 @@ public void testRequestProcessor() throws InterruptedException {
         assertEquals(null, waitingProcessor);
     }
 
+    @Test(timeout = 30000)
+    public void testIgnoreFairness() throws InterruptedException {
```
fpj: Same comment here.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -124,10 +124,15 @@ public String getProcessorName() {
         CompletableFuture<Void> resultFuture = new CompletableFuture<>();
 
         OperationContext context = streamMetadataStore.createContext(scope, stream);
-        suppressException(streamMetadataStore.getWaitingRequestProcessor(scope, stream, context, executor), null,
-                "Exception while trying to fetch waiting request. Logged and ignored.")
-                .thenAccept(waitingRequestProcessor -> {
-                    if (waitingRequestProcessor == null || waitingRequestProcessor.equals(getProcessorName())) {
+        CompletableFuture<String> waitingProcFuture = suppressException(streamMetadataStore.getWaitingRequestProcessor(scope, stream, context, executor), null,
+                "Exception while trying to fetch waiting request. Logged and ignored.");
+        CompletableFuture<Boolean> ignoreFairnessFuture = task.hasTaskStarted(event);
```
fpj: maybe replace the name of this variable to reflect the new name.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -124,10 +124,15 @@ public String getProcessorName() {
...
+        
+        CompletableFuture.allOf(waitingProcFuture, ignoreFairnessFuture)
+                .thenAccept(v -> {
+                    boolean ignoreFairness = ignoreFairnessFuture.join();
+                    String waitingRequestProcessor = waitingProcFuture.join();
+                    if (ignoreFairness || waitingRequestProcessor == null || waitingRequestProcessor.equals(getProcessorName())) {
```
fpj: same comment here about the name.

controller/src/test/java/io/pravega/controller/server/eventProcessor/RequestHandlersTest.java
```diff
@@ -547,4 +552,289 @@ private void createStreamInStore(String stream) {
...
+        
+        // 6. run a new update. it should fail because of waiting processor.
+        TruncateStreamEvent event2 = new TruncateStreamEvent(fairness, fairness, 0L);
+        AssertExtensions.assertFutureThrows("", streamRequestHandler.process(event2),
+                e -> Exceptions.unwrap(e) instanceof StoreException.OperationNotAllowedException);
+        streamStore.deleteWaitingRequestConditionally(fairness, fairness, "myProcessor", null, executor).join();
+    }
+    
+    @Test
+    public void testCommitTxnIgnoreFairness() {
```
RaulGracia: If I'm not wrong, we found the problem in a test that, in addition to failovers, we interleaved scale stream and commit transaction events (`ReadTxnWriteAndAutoScaleWithFailoverTest`). Does it worth it to create a test that verifies that scale and commit events do not get deadlocked?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -306,4 +306,10 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
             return CompletableFuture.completedFuture(null);
         }
     }
+
+    @Override
+    public CompletableFuture<Boolean> hasTaskStarted(CommitEvent event) {
+        return streamMetadataStore.getState(event.getScope(), event.getStream(), true, null, executor)
+                                  .thenApply(state -> state.equals(State.COMMITTING_TXN) || state.equals(State.SEALING));
```
RaulGracia: Could be a problem the fact of having more than one task for which `hasTaskStarted` evaluate to `true` on a given moment? For instance, if the Stream is `SEALING` state, a commit task and a seal task would evaluate  `hasTaskStarted` to `true`. Could this ambiguity be harmful in some way?

controller/src/test/java/io/pravega/controller/server/eventProcessor/RequestHandlersTest.java
shiveshr: This test is doing exactly that but in slightly more generic fashion. we set waiting processor on scale and commit transcation while they have been started.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
shiveshr: This is a good observation. However, this ambiguity has no harmful effect. This ambiguity is not local to fairness but it is about mutual exclusion of tasks. 
Note 1: fairness is about ensuring each processor gets a chance to execute its workflow and doesnt wait indefinitely. Fairness does not care for mutual exclusion. So if some processor is waiting to acquire a mutex to perform its task, the fairness ensures it doesnt wait long. 

Note 2: all other tasks require mutual exclusion. However, `SealStreamTask` and `commitTxn` are a special case. They are not mutually exclusive once the state is set to sealing. 
Commits should proceed even if the state is set to SEALING. 
And seal stream task doesnt start sealing the stream until all transactions have been completed (either committed or aborted). 
So once the state is set to sealing, both tasks can run concurrently and dont require mutual exclusion with each other. 

When either task is picked for processing, and if the state is SEALING, it determines that it can be performed and does not request for a "waiting processor" for fairness or ignores one if present because the processing for this task has already been initiated and should be proceeded with until it is completed.

Issue #217, opened 2016-12-01
Connection pooling


tkaitchuck:
If a user is writing to or reading from multiple streams, there is no requirement for these to go over separate connections. As the number of streams grows this becomes more and more important.
For this reason we should implement connection pooling so that redundant connections to the same host are not needed.

shrids:
![](https://github.trello.services/images/mini-trello-icon.png) [Issue 217: Connection pooling Design document](https://trello.com/c/BtKprslb/131-issue-217-connection-pooling-design-document)

shrids:
SubTasks:
- [ ] Implement the WireCommand Changes and modify the server and client logic to handle extra fields to enable connection pooling. (In Progress)
- [ ] Implement Connection pool for Netty Channels.
- [ ] Enable connection pooling for connections between the client process and Segment store.

PR #3448, opened 2019-03-11
Issue 217: WireCommand changes to enable connection pooling.

shrids:
**Change log description**  
WireCommand changes to enable connection pooling.

**Purpose of the change**  
Issue #217 

**What the code does**  
The goal of connection pooling is to ensure a common pool of connections between the client process and the segment stores, which does not require a linear growth of the number of connections with the number of segments. 

This PR addresses only the `WireCommand` related changes that are required to enable connection pooling.

A request ID field has been added to `WireCommands` and separate fields are used to distinguish between offsets and requestIds.
The following `WireCommands` has been changed : `SegmentIsSealed, SegmentIsTruncated, NoSuchSegment, AppendBlock, AppendBlockEnd, ConditionalAppend, DataAppended, ConditionalCheckFailed, ReadSegment, SegmentRead`
The associated client logic has also been changed.

The `request id`, which is 8 bytes long, is composed of the `session id` (most significant 4 bytes) and `request sequence number` (least significant 4 bytes).  The connection pooling logic will be based on the `session id ` i.e. based on the `session id ` the data will be sent to the appropriate `ReplyProcessor`. 
Also, for a given `session id ` we can have multiple requests with different sequence numbers which is useful in the case of RawClient (this is used for conditional writes `ConditionalOutputStream` and fetching segment meta data `SegmentMetaDataClient`.)

**How to verify it**  


client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
```diff
@@ -51,6 +51,7 @@
     private final AtomicBoolean closed = new AtomicBoolean(false);
     private final Controller controller;
     private final String delegationToken;
+    private final long requestId;
```
tkaitchuck: This is a strange sort of 'requestId' it's being used more like 'writerId' on the write side in that it is not identifying the request, but rather the reader in this case.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -80,6 +81,9 @@
     private final RetryWithBackoff retrySchedule;
     private final Object writeOrderLock = new Object();
     private final String delegationToken;
+    @VisibleForTesting
+    @Getter
+    private final long requestId = RequestIdGenerator.getRequestId();
```
tkaitchuck: This does not seam useful here, as the CommandEncoder is going to transform many of these into an append block.

common/src/main/java/io/pravega/common/util/RequestIdGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.common.util;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * This is used to generated unique request ids per process.
+ */
+public class RequestIdGenerator {
```
tkaitchuck: Two different requestId generators will likely produce collisions.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -91,7 +92,7 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
                 currentBlockSize = Math.max(TYPE_PLUS_LENGTH_SIZE, blockSizeSupplier.getAppendBlockSize());
                 bytesLeftInBlock = currentBlockSize;
                 segmentBeingAppendedTo = append.segment;
-                writeMessage(new AppendBlock(session.id), out);
+                writeMessage(new AppendBlock(append.getRequestId(), session.id), out);
```
tkaitchuck: This is only the first of many IDs that may be in the block...
Perhaps we need a different abstraction.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -554,7 +568,8 @@ public static WireCommand readFrom(DataInput in, int length) throws IOException
         final ByteBuf data;
         final int numEvents;
         final long lastEventNumber;
-        final long unused; // Will be used by AppendSequence:  
+        final long unused; // Will be used by AppendSequence:
+        final long requestId;
```
tkaitchuck: I think you can take over the unused long, conditional appends went a different way.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: Yes, the Command Encoder will use the `requestorId` which is part of the `Append`  to construct the `AppendBlock`.

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
shrids: Yes, I will rename it to `requestorId`. 
This is used to identify the commands originating from a segment writer / segment reader and this is also used to ensure the responses (which also have this id)from the SSS are routed back to the correct reply processors of these segment writers/ segment readers.

common/src/main/java/io/pravega/common/util/RequestIdGenerator.java
shrids: The idea here was to have RequestIds unique per client process. Having the same request id across different client process would not cause any issues on the PRP (or client) as the reply is always sent on the connection which sent the request.
(Please let me know if I misunderstood your comment).

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
shrids: ok.

shrids:
[Connection pooling : Implement WireCommand changes and add the client and server logic to handle this.](https://trello.com/c/KUfmNgfu/145-connection-pooling-implement-wirecommand-changes-and-add-the-client-and-server-logic-to-handle-this)

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -42,6 +45,9 @@
     private final Map<Long, CompletableFuture<Reply>> requests = new HashMap<>();
     private final ResponseProcessor responseProcessor = new ResponseProcessor();
     private final AtomicBoolean closed = new AtomicBoolean(false);
+    @Getter
+    private final AtomicLong requesterId = new AtomicLong(IdGenerator.getRequesterId());
```
andreipaduroiu: Why do you have a getter on the `AtomicLong`? It is a very bad practice to expose internals to the outside. Someone can do `getRequesterId().set(<new-value-here>)` and mess up with your implementation.

Please write an explicit getter that only returns the contents. Similarly, if you need to update it, write an explicit setter for it.

andreipaduroiu: I see you use `updateAndGet` somewhere below. Then please expose the appropriate API on this class that modifies its internal state. Again, you should not expose your internal state and allow others to modify it.

Another reason this is not good is because both in this class and in your outside callers you are using `IdGenerator.getRequesterId`, therefore both of them need to know to use that. if you create an API such as `nextRequestId()` (with no args), that can encapsulate all this logic and will make your code look cleaner.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -72,14 +71,16 @@
...
     private final UUID writerId;
     private final Consumer<Segment> resendToSuccessorsCallback;
     private final State state = new State();
     private final ResponseProcessor responseProcessor = new ResponseProcessor();
     private final RetryWithBackoff retrySchedule;
     private final Object writeOrderLock = new Object();
     private final String delegationToken;
+    @VisibleForTesting
+    @Getter
+    private final long requestorId = IdGenerator.getRequesterId();
```
andreipaduroiu: Can you decide if it's going to be `requestor` or `requester`?

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
```diff
@@ -51,6 +54,9 @@
     private final AtomicBoolean closed = new AtomicBoolean(false);
     private final Controller controller;
     private final String delegationToken;
+    @VisibleForTesting
+    @Getter
+    private final long requestId = IdGenerator.getRequesterId();
```
andreipaduroiu: Is it `requestId` or `requesterId` ?

common/src/main/java/io/pravega/common/util/IdGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.common.util;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.LongUnaryOperator;
+
+/**
+ * IdGenerator is used to generated unique ids which are guaranteed to be unique per process. The id
```
andreipaduroiu: "unique ids which are guaranteed to be unique per process" -> "ids which are guaranteed to be unique per process"

common/src/test/java/io/pravega/common/util/IdGeneratorTest.java
```diff
@@ -0,0 +1,31 @@
...
+import static org.junit.Assert.assertEquals;
+
+
+public class IdGeneratorTest {
+
+    @Test
+    public void getRequesterId() {
+        long id1 = IdGenerator.getRequesterId();
+        long id2 = IdGenerator.getRequesterId();
+        assertEquals("Requester Id should be incremented", id1 >> Integer.SIZE, (id2 >> Integer.SIZE) -1);
```
andreipaduroiu: Are you sure this is a thorough enough test?

Are there other boundary cases to test? Such as numeric overflow? Negative-to-positive transitions?

common/src/main/java/io/pravega/common/util/IdGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Return a {@link LongUnaryOperator} that can be used to generate a new Request id.
+     * @return id which corresponds to the newer request id.
+     */
+    public static LongUnaryOperator getRequestId() {
+        return id -> {
+            int currentRequest = (int) id;
+            int currentRequester = (int) (id >> Integer.SIZE);
+            return ((long) currentRequester << Integer.SIZE) | ((long) (currentRequest + 1) & 0xFFFFFFFL);
```
tkaitchuck: This looks more complicated than needed. You could remove the shifts by just masking the upper bits.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
shrids: updated the api based on feedback.

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
shrids: updated the api .

common/src/test/java/io/pravega/common/util/IdGeneratorTest.java
shrids: updated the api and added tests for numeric overflows.

client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
```diff
@@ -68,7 +68,7 @@ public boolean write(ByteBuffer data, long expectedOffset) throws SegmentSealedE
                     .run(() -> {
                         if (client == null || client.isClosed()) {
                             client = new RawClient(controller, connectionFactory, segmentId);
-                            long requestId = requestIdGenerator.get();
+                            long requestId = client.getRequesterId().updateAndGet(IdGenerator.getRequestId());
```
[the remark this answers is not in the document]
shrids: updated the api.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: updated the code to use a RequestId abstraction.

client/src/main/java/io/pravega/client/RequestId.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+@ToString
+public class RequestId {
+    private static final AtomicInteger ID_GENERATOR = new AtomicInteger(0);
+    @Getter
+    private final int sessionId;
+    @GuardedBy("$LOCK")
+    private int requestSequenceNumber = 0;
+
+    public RequestId() {
```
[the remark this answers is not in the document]
shrids: done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/Append.java
```diff
@@ -22,22 +22,24 @@
     final int eventCount;
     final ByteBuf data;
     final Long expectedLength;
+    final long requestId;
```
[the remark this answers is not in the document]
shrids: updated it.

client/src/main/java/io/pravega/client/RequestId.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * The class represents a request id that is sent as part of every WireCommand. It consists of a
+ * <i>sessionId</i> and a <i>requestSequenceNumber</i>. The <i>sessionId</i> is used to represent the
+ * communication between Segment clients and the SegmentStore which uses an underlying
+ * connection pool. This <i>sessionId</i> is unique per connection pool.
+ * The <i>requestSequenceNumber</i> is used to represent the requestSequence for a given session.
+ */
+@ToString
+public class RequestId {
```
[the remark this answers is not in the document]
shrids: Updated the class and refactored the methods.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -153,6 +153,7 @@ public boolean isFailure() {
         final long requestId;
         final String segment;
         final String serverStackTrace;
+        final long offset;
```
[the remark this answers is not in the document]
shrids: done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -257,6 +263,7 @@ public boolean isFailure() {
...
```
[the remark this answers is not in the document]
shrids: this offset is used by the reader to do lookup of the read future. ref: io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl.ResponseProcessor#noSuchSegment
Updated the comment to indicate the same.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -186,6 +189,7 @@ public boolean isFailure() {
         final String segment;
         final long startOffset;
         final String serverStackTrace;
+        final long offset;
```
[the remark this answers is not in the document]
shrids: updated the comment.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -189,6 +190,7 @@ public boolean isFailure() {
         final String segment;
         final long startOffset;
         final String serverStackTrace;
+        // This represents the offset at which the segment is truncated at.
```
tkaitchuck: Comments like this, and on 268, and 156 should be made with the java doc style so they are applied to the getters.

shrids: ok, the java doc style here will be applied to the getters generated by lombok. 
Updated the comments to use java doc style.

PR #3453, opened 2019-03-11
Issue 3405: add tags into metrics with dimensions

### LINKED ISSUE #3405: Add tags to those metrics with dimensions
**Problem description**
This is the follow-up issue of issue 3319, which is to migrate Pravega metric to Micrometer based in order to support dimensional metrics.
With issue-3319 in place, we need to add tags into these metrics with dimensions, per the telegraf.conf provided by the Platform team.

**Problem location**
All the places currently adding dimension into flat metric name: nameFromContainer, nameFromHost, nameFromSegment, nameFromStream, nameFromTransaction, failMetricName etc.

Instead of flatting the metric name, start calling method (metricName, tagName, tagValue) which has been baked into the code by issue-3319 (precisely PR https://github.com/pravega/pravega/pull/3375)

kevinhan88:
**Change log description**  
Based on https://github.com/pravega/pravega/pull/3375, tags were added into those metrics with dimensions.

**Purpose of the change**  
Fixes #3405 

**What the code does**  
Tags were added into those metrics with dimensions

**How to verify it**  
Integration test such as direct InfluxDB connection shows metrics are being generated with tags.
p.s. The Platform team has taken this branch and did an end-to-end test, and said okay for the branch to be merged.


shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
...
+    public static final String TAG_SCOPE = "scope";
+    public static final String TAG_STREAM = "stream";
+    public static final String TAG_SEGMENT = "segment";
+    public static final String TAG_TRANSACTION = "transaction";
+    public static final String TAG_EPOCH = "epoch";
+
+    private static final String TRANSACTION_DELIMITER = "#transaction.";
+    private static final String EPOCH_DELIMITER = ".#epoch.";
+    /**
+     * The Transaction unique identifier is made of two parts, each having a length of 16 bytes (64 bits in Hex).
```
RaulGracia: A Transaction id is a `UUID` that `represents a 128-bit value`. Then, you can get the most/least significant bytes of the `UUID`, which are 8 bytes (64 bits) each, right?

controller/src/main/java/io/pravega/controller/metrics/HostContainerMetrics.java
```diff
@@ -66,20 +66,20 @@ public void updateHostContainerMetrics(Map<Host, Set<Integer>> oldMapping, Map<H
     }
 
     private void reportContainerCountPerHost(Host host, Set<Integer> containerIds) {
-        DYNAMIC_LOGGER.reportGaugeValue(nameFromHost(SEGMENT_STORE_HOST_CONTAINER_COUNT, host.toString()), containerIds.size());
+        DYNAMIC_LOGGER.reportGaugeValue(SEGMENT_STORE_HOST_CONTAINER_COUNT, containerIds.size(), hostTag(host.toString()));
```
RaulGracia: I see that now metrics are indexed in another way. While this looks more flexible, I have a question: are dashboards impacted by this change? (e.g., Grafana dashboards)

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    private static final int TRANSACTION_PART_LENGTH = Long.BYTES * 8 / 4;
+    /**
+     * The length of the Transaction unique identifier, in bytes (it is made of two parts).
+     */
+    private static final int TRANSACTION_ID_LENGTH = 2 * TRANSACTION_PART_LENGTH;
+
+    /**
+     * Generate string array as container tag for metric.
```
RaulGracia: The description for these methods could be a bit more descriptive, for example `Generates a tag based on the input containerId to be associated with a metric.`

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Generate string array as segment tags for metric.
+     * @param qualifiedSegmentName fully qualified segment name.
+     * @return string array as segment tag of metric.
+     */
+    public static String[] segmentTags(String qualifiedSegmentName) {
+        Preconditions.checkNotNull(qualifiedSegmentName);
+        String segmentBaseName = getSegmentBaseName(qualifiedSegmentName);
+
+        String[] tags = {TAG_SCOPE, null, TAG_STREAM, null, TAG_SEGMENT, null, TAG_EPOCH, null};
```
RaulGracia: You may want to add more inline comments of the actual inputs you expect to parse here. Also, a good testing of the different expected (and unexpected) inputs for this method is important.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return tags;
+    }
+
+    /**
+     * Get base name of segment with the potential transaction delimiter removed.
+     * @param segmentQualifiedName fully qualified segment name.
+     * @return the base name of segment.
+     */
+    private static String getSegmentBaseName(String segmentQualifiedName) {
```
RaulGracia: You may want to check out `StreamSegmentNameUtils`.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: This could be a post-migration enhancement item.

controller/src/main/java/io/pravega/controller/metrics/HostContainerMetrics.java
kevinhan88: Yes, today's hierarchical data model is not sustainable, migrating towards dimensional data model is the direction we have to go; Eron in the demo also said they are willing to accept essential breaking changes.
Having said that, effort has been made to make the migration backwards as much as possible. Metrics names, tags, prefix are kept same. Metric type suffix are removed per the demo feedback.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: Yes many test cases created for metric tags and a new MetricsTags utility class generated.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: Seems there is no visibility of StreamSegmentNameUtils from shared/metrics. The dependency is the other way around.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -247,4 +220,49 @@ public static String failMetricName(String metricName) {
...
     }
+
+    /**
+     * Convenient method to join two Strings with dot.
+     *
+     * @param element1 the first String object.
+     * @param element2 the second String object.
+     * @return String object with element1 and element2 joined with dot.
+     */
+    public static String joinWithDot(String element1, String element2) {
```
ravisharda: IMO, joinWithDot() doesn't look like representing a functional behavior of the entity `MetricNames` that this class represents. 

Is this for concatenating metric names? If so, it could perhaps be represented by the signature `concatenate(String metric1, String metric2)`. 

Or maybe, this is for concatenating two elements of different kinds and not metrics. Perhaps a better location for such a method is a utility class.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
```
ravisharda: 2019 :)

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: Fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: Will said and agree with all the principals mentioned:
1. Agree the method should go to utility class. MetricsNames.java is actually a sort of utility class, just without the "Util" on the name. It is called "MetricsNames" instead of "MetricName" to follow the Java Utility class convention, such as Strings.
2. Also struggling with the method name "join" or "concatenate" because the name doesn't call out the "dot" in-between.
3. Trying to make a perfect balance, I tried to move this method into NameUtils.java or StringUtil.java, but due to the dependency relationship, metric related codes do not have access to these utility classes.

Hope this explains.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: Fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
ravisharda: Sounds good.

kevinhan88:
Close this PR because https://github.com/pravega/pravega/pull/3504 is created to replace this PR:

Why to have PR https://github.com/pravega/pravega/pull/3504 to replace this one?

1. Somehow the clock on my VirtualBox VM was days behind the real time (from this PR you can see the commit time is about a week behind), this makes re-basing not functioning well and caused a big mess.
In addition this PR is not showing difference based on the latest master, which makes code review very difficult.
2. If we stick with this PR, the future merge into master may experience issues again due to the disorder of commits.
3. There are just a few review comments left on this PR, which have been all addressed.
4. I have to take personal off tomorrow so I'd like to have a clean PR for the team, which is https://github.com/pravega/pravega/pull/3504.

Sorry for inconvenience or confusion this may cause.

PR #3459, opened 2019-03-13
Issue 3458: Controller Metrics Config Fix

### LINKED ISSUE #3458: Controller metrics config is not picked up from config file/command line
**Problem description**
Controller metrics config is not picked up from config file/command line because it replaces the prefix `controller.metrics` with `metrics`, but when populating the `MetricsConfig.builder`, that `metrics` prefix is added again.

**Problem location**
Config.java

Do not add `metrics` as a prefix again.

andreipaduroiu:
**Change log description**  
Fixing `Config.java#createMetricsConfig` to not prefix metrics-related config values with "metrics" - that is already done by the `MetricsConfig` builder.

**Purpose of the change**  
Fixes #3458.

**What the code does**  
Instead of replacing `controllermetrics.` with `metrics.` it simply removes `controller.metrics.` from the property name. The `MetricsConfig.builder()` will take care of the prefix.

**How to verify it**  
New unit test to verify metrics config for controller.

### DIFF
--- a/controller/src/test/resources/controller.config.properties
+++ b/controller/src/test/resources/controller.config.properties
@@ -46,3 +46,4 @@ controller.transaction.maxLeaseValue=30000
 #controller.transaction.ttlHours=
 controller.scale.streamName=_requeststream
 #controller.scale.readerGroup=scaleGroup
+controller.metrics.statsDHost=no-host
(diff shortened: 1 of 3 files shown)


controller/src/test/resources/controller.config.properties
```diff
@@ -46,3 +46,4 @@ controller.transaction.maxLeaseValue=30000
 #controller.transaction.ttlHours=
 controller.scale.streamName=_requeststream
 #controller.scale.readerGroup=scaleGroup
+controller.metrics.statsDHost=no-host
```
tkaitchuck: newline

PR #3468, opened 2019-03-15
Issue 3467: Monitor Pravega pods for restart during a system test run.

### LINKED ISSUE #3467: Fail system test if Pravega container restart is observed.
**Problem description**
System tests currently deploys Pravega on a Kubernetes cluster and runs tests against it. The tests should fail if there is a restart of the Pravega pods during a specific run.

The restart of pod can happen due to muliple reasons like
- K8s sees that the pod is not responding. (liveness probe failure)
- The jvm process crashes due to OOM or any other exception.
- the Pravega pods are consuming more resources limits set.

**Problem location**
K8SequentialExecutor

Improve the framework.

shrids:
**Change log description**  
Monitor Pravega pod for restart during every system test run. A pod restart during the test run will fail the corresponding test and log the reason for the pravega pod failure.

**Purpose of the change**  
Fixes #3467 

**What the code does**  
The code now fetches the container Status of Pravega pods during every system test run. Incase the restart count of the Pravega pod is greater than what was observed during the start of the test then the  corresponding test is failed with the details.
Example output while simulating a pod restart  (used `docker kill <containerId>`  to simulate a Pravega pod failure)
```
io.pravega.test.system.PravegaTest > simpleTest FAILED
    java.lang.AssertionError: Failing test due to Pravega pod restart.
{pravega-pravega-controller-6fcf5658c8-7c6t6=class V1ContainerStatus {
        containerID: docker://3d9ad5cde103e5fbffc72a3f410aa49a4f081bf914d84b1ebef53608c9a2e3f7
        lastState: class V1ContainerState {
            running: null
            terminated: class V1ContainerStateTerminated {
                containerID: docker://2b2aed2e0ef4b5c54d9d006ae93314e92876d2b93b8ec9dc38e564d8a86a1d48
                exitCode: 137
                finishedAt: 2019-03-15T04:17:59.000-04:00
                message: null
                reason: Error
                signal: null
                startedAt: 2019-03-15T03:49:18.000-04:00
            }
   restartCount: 5       
   ....   
        state: class V1ContainerState {
            running: class V1ContainerStateRunning {
                startedAt: 2019-03-15T04:18:05.000-04:00
            }
            terminated: null
            waiting: null
        }
    }}
``` 
This also logs the container id, which can be used to fetch the logs from `fluentd` to understand more about the cause of failure.

**How to verify it**  
Verified the behavior by performing `docker kill <containerId>` , which causes K8s to restart the pod causing the test to fail with this error.

### DIFF
--- a/test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
+++ b/test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
@@ -193,8 +193,25 @@ public class K8sClient {
      * @param labelValue Value of the label.
      * @return Future representing the list of pod status.
      */
-    @SneakyThrows(ApiException.class)
     public CompletableFuture<List<V1PodStatus>> getStatusOfPodWithLabel(final String namespace, final String labelName, final String labelValue) {
+        CompletableFuture<V1PodList> future = getPodsWithLabel(namespace, labelName, labelValue);
+        return future
+                .thenApply(v1PodList -> {
+                    List<V1Pod> podList = v1PodList.getItems();
+                    log.debug("{} pod(s) found with label {}={}.", podList.size(), labelName, labelValue);
+                    return podList.stream().map(V1Pod::getStatus).collect(Collectors.toList());
+                });
+    }
+
+    /**
+     * Method to fetch all pods which match a label.
+     * @param namespace Namespace on which the pod(s) reside.
+     * @param labelName Name of the label.
+     * @param labelValue Value of the label.
+     * @return Future representing the list of pod status.
+     */
+    @SneakyThrows(ApiException.class)
+    public CompletableFuture<V1PodList> getPodsWithLabel(String namespace, String labelName, String labelValue) {
         CoreV1Api api = new CoreV1Api();
 
         // Workaround for okhttp issue, tracked by https://github.com/pravega/pravega/issues/3361
@@ -204,12 +221,23 @@ public class K8sClient {
         K8AsyncCallback<V1PodList> callback = new K8AsyncCallback<>("listPods");
         api.listNamespacedPodAsync(namespace, PRETTY_PRINT, null, null, true, labelName + "=" + labelValue, null,
                                    null, null, false, callback);
-        return callback.getFuture()
-                       .thenApply(v1PodList -> {
-                           List<V1Pod> podList = v1PodList.getItems();
-                           log.debug("{} pod(s) found with label {}={}.", podList.size(), labelName, labelValue);
-                           return podList.stream().map(V1Pod::getStatus).collect(Collectors.toList());
-                       });
+        return callback.getFuture();
+    }
+
+    /**
+     * Method to fetch all restarted pods which match a label.
+     * @param namespace Namespace on which the pod(s) reside.
+     * @param labelName Name of the label.
+     * @param labelValue Value of the label.
+     * @return Future representing the list of pod status.
+     */
+    public CompletableFuture<Map<String, V1ContainerStatus>> getRestartedPods(String namespace, String labelName, String labelValue) {
+        return getPodsWithLabel(namespace, labelName, labelValue)
+                     .thenApply(v1PodList -> v1PodList.getItems().stream()
+                                                      .filter(pod -> pod.getStatus().getContainerStatuses().size() != 0 && (pod.getStatus().getContainerStatuses().get(0).getRestartCount() != 0))
+                                                      .collect(Collectors.toMap(pod -> pod.getMetadata().getName(),
+                                                                                // currently this supports only one container per pod.
+                                                                                pod -> pod.getStatus().getContainerStatuses().get(0))));
     }
 
     /**
--- a/test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
+++ b/test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
@@ -185,7 +185,6 @@ public class K8sClient {
                        });
     }
 
-
     /**
      * Method to fetch the status of all pods which match a label.
      * @param namespace Namespace on which the pod(s) reside.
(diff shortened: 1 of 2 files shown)


test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
```diff
@@ -204,12 +219,23 @@ public V1Namespace createNamespace(final String namespace) {
...
+    }
+
+    /**
+     * Method to fetch all restarted pods which match a label.
+     * @param namespace Namespace on which the pod(s) reside.
+     * @param labelName Name of the label.
+     * @param labelValue Value of the label.
+     * @return Future representing the list of pod status.
+     */
+    public CompletableFuture<Map<String, V1ContainerStatus>> getRestartedPods(String namespace, String labelName, String labelValue) {
```
shiveshr: is it possible that this will cascade to other tests once first test fails with a restart?

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
```diff
@@ -204,12 +219,23 @@ public V1Namespace createNamespace(final String namespace) {
...
+        return getPodsWithLabel(namespace, labelName, labelValue)
+                     .thenApply(v1PodList -> v1PodList.getItems().stream()
+                                                      .filter(pod -> pod.getStatus().getContainerStatuses().size() != 0 && (pod.getStatus().getContainerStatuses().get(0).getRestartCount() != 0))
```
RaulGracia: Maybe `!pod.getStatus().getContainerStatuses().isEmpty()` could be clearer here.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
```diff
@@ -75,6 +83,23 @@
 
     }
 
+    private Map<String, V1ContainerStatus> getPravegaPodStatus(K8sClient client) {
+        return Futures.getAndHandleExceptions(
+                client.getRestartedPods("default", "app", "pravega-cluster"),
```
RaulGracia: We need to be sure of what we need to include to check for restarts. I agree that a Controller and Segment Store restart should be interpreted as a failure. This may be also extended to unexpected ZK and BK pod restarts. However, we need to make sure that we do not include any of the other pods available by default in the cluster (e.g., fluentd).

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
```diff
@@ -185,16 +185,31 @@ public V1Namespace createNamespace(final String namespace) {
...
     /**
      * Method to fetch the status of all pods which match a label.
      * @param namespace Namespace on which the pod(s) reside.
      * @param labelName Name of the label.
      * @param labelValue Value of the label.
      * @return Future representing the list of pod status.
      */
-    @SneakyThrows(ApiException.class)
     public CompletableFuture<List<V1PodStatus>> getStatusOfPodWithLabel(final String namespace, final String labelName, final String labelValue) {
+        return getPodsWithLabel(namespace, labelName, labelValue)
```
adrianmo: Even if we are using only one label at the moment, this method should support multiple labels to cover future usages.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
```diff
@@ -204,12 +219,23 @@ public V1Namespace createNamespace(final String namespace) {
...
+     * @param labelName Name of the label.
+     * @param labelValue Value of the label.
+     * @return Future representing the list of pod status.
+     */
+    public CompletableFuture<Map<String, V1ContainerStatus>> getRestartedPods(String namespace, String labelName, String labelValue) {
+        return getPodsWithLabel(namespace, labelName, labelValue)
+                     .thenApply(v1PodList -> v1PodList.getItems().stream()
+                                                      .filter(pod -> pod.getStatus().getContainerStatuses().size() != 0 && (pod.getStatus().getContainerStatuses().get(0).getRestartCount() != 0))
+                                                      .collect(Collectors.toMap(pod -> pod.getMetadata().getName(),
+                                                                                // currently this supports only one container per pod.
```
adrianmo: Include this comment to the method Javadoc to let users know about this limitation.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
shrids: No, the code only fails a test if there is a restart of pods (with the provided label) during the course of execution of the test by comparing the restart count of the pods before and after the tests.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
shrids: At present with the label `app=pravega-cluster` tracks the restart counts for bookies, Segment store and Controller.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
shrids: fixed.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
shrids: Created a new method.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
shrids: Updated the code to track zk pod status .

PR #3469, opened 2019-03-15
Issue 3462: Fix Segmentstore's Kubernetes bootstrapping

### LINKED ISSUE #3462: No SegmentStore docker init failure when deploying with failed services in K8s
**Problem description**
Deploying Pravega cluster with external access enabled and LoadBalancer as the service type,
I'm observing segment store servers to be in a running state when LoadBalancer services have not come up properly.

Checking the SegmentStore logs, I'm observing the following:
```
Running in a Kubernetes environment and managed by the Pravega Operator with external access enabled
Trying to obtain the external service endpoint...
Found external service endpoint: null:12345
Checking whether NFS mounting is required
```
This means that the `init_kubernetes.sh` script does not fail when `PUBLISHED_ADDRESS` variable comes back with a `null` value.

**Problem location**
SegmentStore docker init script 
https://github.com/pravega/pravega/blob/master/docker/pravega/scripts/init_kubernetes.sh

`init_kubernetes.sh` should fail if either `PUBLISHED_ADDRESS` or `PUBLISHED_PORT` get a `null` value.

adrianmo:
**Change log description**  
- Replace "null" return values with "" (empty string) to consider the value as invalid and keep the process in the retry loop.
- Refactor how the external service endpoint is obtained to reduce duplicated code.

**Purpose of the change**  
Fix #3462 

**What the code does**  
Updates the `init_kubernetes.sh` file.

**How to verify it**  
@jkhalack will be able to test the change in an environment where this error can be easily reproduced. 

---
Signed-off-by: Adrián Moreno <adrian@morenomartinez.com>

### DIFF
--- a/docker/pravega/scripts/init_kubernetes.sh
+++ b/docker/pravega/scripts/init_kubernetes.sh
@@ -22,6 +22,10 @@ k8() {
         retval=$( curl --cacert ${cacert} -H "Authorization: Bearer ${bearer}" https://kubernetes.default.svc/api/v1/namespaces/${namespace}/${resource_type}/${resource_name} 2> /dev/null | jq -rM "${jsonpath}" 2> /dev/null )
     fi
 
+    if [ "$retval" == "null" ]; then
+        retval=""
+    fi
+
     echo "$retval"
 }
 
--- a/docker/pravega/scripts/init_kubernetes.sh
+++ b/docker/pravega/scripts/init_kubernetes.sh
@@ -32,7 +32,6 @@ k8() {
 init_kubernetes() {
     if [ -d "/var/run/secrets/kubernetes.io" ] && [ ! -z "${K8_EXTERNAL_ACCESS}" ]; then
         echo "Running in a Kubernetes environment and managed by the Pravega Operator with external access enabled"
-        echo "Trying to obtain the external service endpoint..."
 
         if [ -z "${POD_NAMESPACE}" ]; then
           echo "POD_NAMESPACE variable not set. Exiting..."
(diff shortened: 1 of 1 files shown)


adrianmo:
@jkhalack I have built the Docker images from this branch to facilitate your testing:
```
adrianmo/pravega:0.5.0-2156.6adddf0-SNAPSHOT
adrianmo/bookkeeper:0.5.0-2156.6adddf0-SNAPSHOT
```
Thanks!

docker/pravega/scripts/init_kubernetes.sh
```diff
@@ -42,29 +45,25 @@ init_kubernetes() {
...
         if [ "${service_type}" == "LoadBalancer" ]; then
-            export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[0].ip" )
-            export PUBLISHED_PORT=$( k8 "${ns}" "services" "${podname}" ".spec.ports[].port" )
             while [ -z ${PUBLISHED_ADDRESS} ] || [ -z ${PUBLISHED_PORT} ]
             do
-                echo "LoadBalancer external address and port could not be obtained. Waiting 30 seconds and trying again..."
-                sleep 30
+                echo "Trying to obtain LoadBalancer external endpoint..."
+                sleep 10
                 export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[0].ip" )
```
Tristan1900: Might be a problem here. When `PUBLISHED_ADDRESS` is `null`, it can still exit the while loop. Maybe we can add something like `export PUBLISTSH_ADDRESS="null"` and `[ "${PUBLISHED_ADDRESS}" != "null" ]` . What do you think?

adrianmo: @Tristan1900 with the changes made to the `k8` function, `PUBLISHED_ADDRESS` and `PUBLISHED_PORT` should never be "null". Am I missing something?

Tristan1900: Right! Sorry didn't see that.

Tristan1900:
LGTM!

Tristan1900:
Update for verification. 
I ran HelloWorldReader/Writer sample against this PR and it works as expected in NightShift2 and GKE. NightShift1 for some reasons cannot provide an external address so segmentstore stuck at init stage forever. I think the problem is in NightShift1 and has nothing to do with this PR.

adrianmo:
Thanks @Tristan1900 and @jkhalack for testing the changes!

adrianmo:
@fpj this is good to go.

Issue #3344, opened 2019-02-05
Implement watermarks


tkaitchuck:
Following the design in [PDP 33](https://github.com/pravega/pravega/wiki/PDP-33:-Watermarking) implement watermarking.
This is an umbrella issue to cover all the features.
To allow for work in parallel, send pull requests for each component to the `feature-watermarking` branch. This way all the changes can be merged at once without creating a massive code review.

Todo:
- [x] Collect positions in the writer.
- [x] Add controller API to accept the streamcuts from the client.
- [x] Add support for reading from the marks segment on the reader.
- [ ] Test to validate the client end-to-end with a mock controller.
- [x] Implement single-controller-per-stream protocol.
- [x] Test to verify controller failover mechanism.
- [x] Implement aggregation logic.
- [x] Implement serialization/deserialization of aggregated marks.
- [ ] Test to validate client can decode data as written by controller.
- [x] Test to validate controller end-to-end with a mock client.
- [x] Test to validate the whole system end-to-end.
- [ ] Test to validate failover end to end.
- [ ] Add system test for Watermarks that includes failover.
- [x] Add public API
- [ ] Add documentation

tkaitchuck:
As an add on item:
- [ ] Aggregate writes from multiple requests into a single call to the controller.
- [ ] Aggregate writes from multiple writer calls to the controller into a single metadata store call.

PR #3480, opened 2019-03-19
Issue #3344: (Watermarking) Add server side support for watermarking

tkaitchuck:
**NOTE: This is merging to the Watermarking branch not master.**

**Change log description**  
* Add server side support and new wirecommands needed for watermarking
* Add plumbing on the client to get the offsets where they are needed.

**Purpose of the change**  
Sub item of #3344

**What the code does**  
Changes appends and segment merges to return the length of the segment after the data has been appended. This will be collected by the writer and passed to the controller when time is noted.

**How to verify it**  
Not yet ready for end-to-end testing as client impl is not complete.

### DIFF
--- a/segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/StreamSegmentStore.java
+++ b/segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/StreamSegmentStore.java
@@ -26,44 +26,53 @@ import java.util.concurrent.CompletableFuture;
  * * To delete an Attribute, set its value to Attributes.NULL_ATTRIBUTE_VALUE.
  */
 public interface StreamSegmentStore {
+    
     /**
-     * Appends a range of bytes at the end of a StreamSegment and atomically updates the given attributes. The byte range
-     * will be appended as a contiguous block, however there is no guarantee of ordering between different calls to this
-     * method.
+     * Appends a range of bytes at the end of a StreamSegment and atomically updates the given
+     * attributes. The byte range will be appended as a contiguous block, however there is no
+     * guarantee of ordering between different calls to this method.
      *
      * @param streamSegmentName The name of the StreamSegment to append to.
      * @param data              The data to add.
-     * @param attributeUpdates  A Collection of Attribute-Values to set or update. May be null (which indicates no updates).
-     *                          See Notes about AttributeUpdates in the interface Javadoc.
+     * @param attributeUpdates  A Collection of Attribute-Values to set or update. May be null (which
+     *                          indicates no updates). See Notes about AttributeUpdates in the interface Javadoc.
      * @param timeout           Timeout for the operation
-     * @return A CompletableFuture that, will completed normally, if the add was added. If the
-     * operation failed, the future will be failed with the causing exception.
-     * @throws NullPointerException     If any of the arguments are null, except attributeUpdates.
+     * @return A CompletableFuture that, when completed normally, will indicate the append completed
+     *         successfully and contains the new length of the segment. If the operation failed, the
+     *         future will be failed with the causing exception. (NOTE: the length is not
+     *         necessarily the same as offset immediately following the data because the append may
+     *         have been batched together with others internally.)
+     * @throws NullPointerException If any of the arguments are null, except attributeUpdates.
      * @throws IllegalArgumentException If the StreamSegment Name is invalid (NOTE: this doesn't
-     *                                  check if the StreamSegment does not exist - that exception will be set in the
-     *                                  returned CompletableFuture).
+     *                                  check if the StreamSegment does not exist - that exception 
+     *                                  will be set in the returned CompletableFuture).
      */
-    CompletableFuture<Void> append(String streamSegmentName, byte[] data, Collection<AttributeUpdate> attributeUpdates, Duration timeout);
+    CompletableFuture<Long> append(String streamSegmentName, byte[] data, Collection<AttributeUpdate> attributeUpdates, Duration timeout);
 
     /**
-     * Appends a range of bytes at the end of a StreamSegment an atomically updates the given attributes, but only if the
-     * current length of the StreamSegment equals a certain value. The byte range will be appended as a contiguous block.
-     * This method guarantees ordering (among subsequent calls).
+     * Appends a range of bytes at the end of a StreamSegment an atomically updates the given
+     * attributes, but only if the current length of the StreamSegment equals a certain value. The
+     * byte range will be appended as a contiguous block. This method guarantees ordering (among
+     * subsequent calls).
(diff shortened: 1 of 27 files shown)


segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/StreamSegmentStore.java
```diff
@@ -27,43 +27,51 @@
...
     /**
-     * Appends a range of bytes at the end of a StreamSegment and atomically updates the given attributes. The byte range
-     * will be appended as a contiguous block, however there is no guarantee of ordering between different calls to this
-     * method.
+     * Appends a range of bytes at the end of a StreamSegment and atomically updates the given
+     * attributes. The byte range will be appended as a contiguous block, however there is no
+     * guarantee of ordering between different calls to this method.
      *
      * @param streamSegmentName The name of the StreamSegment to append to.
-     * @param data              The data to add.
```
andreipaduroiu: I liked the indentation like it was before. Made for easier reading.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/StreamSegmentStore.java
```diff
@@ -27,43 +27,51 @@
...
+     *         future will be failed with the causing exception. (NOTE: the length is not
+     *         necessarily the same as offset immediately following the data because the append may
+     *         have been batched together with others internally.)
+     * @throws NullPointerException If any of the arguments are null, except attributeUpdates.
+     * @throws IllegalArgumentException If the StreamSegment Name is invalid (NOTE: this doesn't
+     *             check if the StreamSegment does not exist - that exception will be set in the
+     *             returned CompletableFuture).
      */
-    CompletableFuture<Void> append(String streamSegmentName, long offset, byte[] data, Collection<AttributeUpdate> attributeUpdates, Duration timeout);
+    CompletableFuture<Long> append(String streamSegmentName, long offset, byte[] data, Collection<AttributeUpdate> attributeUpdates, Duration timeout);
```
andreipaduroiu: Technically append-with-offset doesn't need to return this (as it should be the same as `offset`. Did you put this here just for consistency?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -349,12 +351,15 @@ public boolean isOffline() {
         return this.metadataStore.getOrAssignSegmentId(streamSegmentName, timer.getRemaining(),
                 streamSegmentId -> {
                     StreamSegmentAppendOperation operation = new StreamSegmentAppendOperation(streamSegmentId, data, attributeUpdates);
-                    return processAttributeUpdaterOperation(operation, timer);
+                    return processAttributeUpdaterOperation(operation, timer).thenApply(v -> {
+                        SegmentMetadata segmentMetadata = this.metadata.getStreamSegmentMetadata(streamSegmentId);
```
andreipaduroiu: No need for this. `StreamSegmentAppendOperation` will contain the offset where this was added. just return `operation.getStreamSegmentOffset()`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -363,7 +368,10 @@ public boolean isOffline() {
         return this.metadataStore.getOrAssignSegmentId(streamSegmentName, timer.getRemaining(),
                 streamSegmentId -> {
                     StreamSegmentAppendOperation operation = new StreamSegmentAppendOperation(streamSegmentId, offset, data, attributeUpdates);
-                    return processAttributeUpdaterOperation(operation, timer);
+                    return processAttributeUpdaterOperation(operation, timer).thenApply(v -> {
```
andreipaduroiu: Same here

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -521,7 +529,11 @@ public boolean isOffline() {
                     this.durableLog.add(new MergeSegmentOperation(targetSegmentId, sourceSegmentId), timer.getRemaining()));
         }
 
-        return result.thenApply(v -> sourceMetadata.getSnapshot());
+        return result.thenApply(v -> {
+            SegmentProperties sourceProperties = sourceMetadata.getSnapshot();
+            return new MergeStreamSegmentResultImpl(this.metadata.getStreamSegmentMetadata(targetSegmentId).getLength(),
```
andreipaduroiu: `MergeSegmentOperation` also has both a `getStreamSegmentOffset` and `getLength`, which can be used instead of querying the metadata. The only case where you really do need to get the target length is when we delete the source (line 519).

(I prefer not to query the metadata too much as it's shared by all internal components of the Container, and every access requires acquiring a lock)

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentContainerTests.java
```diff
@@ -235,7 +235,7 @@ public void testSegmentRegularOperations() throws Exception {
                 attributeUpdates.add(new AttributeUpdate(attributeReplaceIfEquals,
                         i == 0 ? AttributeUpdateType.Replace : AttributeUpdateType.ReplaceIfEquals, i + 1, i));
                 byte[] appendData = getAppendData(segmentName, i);
-                opFutures.add(context.container.append(segmentName, appendData, attributeUpdates, TIMEOUT));
+                opFutures.add(Futures.toVoid(context.container.append(segmentName, appendData, attributeUpdates, TIMEOUT)));
```
andreipaduroiu: In this test only (not the rest), you should add some code to verify that these futures actually return the offset. All of these are executed in order and you know the length (appendData), so it should be easy to store a list of expected offsets (in order) and then compare with the results from `opFutures`.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/StreamSegmentStore.java
tkaitchuck: Yes. They get handled the same in the caller, so it was easier for them to work the same way.

andreipaduroiu:
Looks like  checkstyle issue is preventing the build from completing. @tkaitchuck can you please give it a look?

fpj:
@tkaitchuck Is it the case that code coverage won't report because the build of the base branch is currently failing?

Issue #3479, opened 2019-03-19
Increase client throttling limit


vijikarthi:
**Problem description**
The client throttling limit for the append entry request is very less and it has a direct impact to the Pravega performance. The current limit that we have in [Append Processor](https://github.com/pravega/pravega/blob/master/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java#L73) is an arbitrary number and from the performance testing results we have observed that increasing the throttle limit by a factor of ~~10~~ 5 has improved the performance a lot.

**Problem location**
Segment Store [Append Processor](https://github.com/pravega/pravega/blob/master/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java#L73)

**Suggestions for an improvement**

RaulGracia:
Closed via #3482.

PR #3482, opened 2019-03-19
Issue 3479: Increase client throttling limit

vijikarthi:
Signed-off-by: Vijay Srinivasaraghavan <vijayaraghavan.srinivasaraghavan@emc.com>

**Change log description**  
  * increased the throttling limit by a factor of ~~10~~ 5 to ensure the clients are not blocked 

**Purpose of the change**  
To address the issue https://github.com/pravega/pravega/issues/3479

**What the code does**  
Increased the throttling configuration setting in `AppendProcessor`  

**How to verify it**  
test by writing more/frequent data with/without the fix and observe the write throughput difference

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -70,8 +70,8 @@ public class AppendProcessor extends DelegatingRequestProcessor {
     //region Members
 
     static final Duration TIMEOUT = Duration.ofMinutes(1);
-    private static final int HIGH_WATER_MARK = 128 * 10 * 1024;
-    private static final int LOW_WATER_MARK = 64 * 10 * 1024;
+    private static final int HIGH_WATER_MARK = 640 * 1024; // 640KB
+    private static final int LOW_WATER_MARK = 320 * 1024;  // 320KB
     private static final String EMPTY_STACK_TRACE = "";
     private final StreamSegmentStore store;
     private final ServerConnection connection;
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -70,8 +70,8 @@ public class AppendProcessor extends DelegatingRequestProcessor {
     //region Members
 
     static final Duration TIMEOUT = Duration.ofMinutes(1);
-    private static final int HIGH_WATER_MARK = 128 * 1024;
-    private static final int LOW_WATER_MARK = 64 * 1024;
+    private static final int HIGH_WATER_MARK = 128 * 10 * 1024;
+    private static final int LOW_WATER_MARK = 64 * 10 * 1024;
     private static final String EMPTY_STACK_TRACE = "";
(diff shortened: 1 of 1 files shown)


vijikarthi:
> Do you observe a performance difference with this change?

Yes.

RaulGracia:
@kmgowda @vijikarthi perhaps adding in the PR some numeric values/improvement percentages specifying the workload used may help reviewers to better understand the extent of this change.

kmgowda:
please find the detailed results of the performance difference between the master and this pull request 
note that there major improvement in the throughput for the events sizes 1K and above
However , for 100 bytes I have not any seen any difference in terms of throughput and latencies

vijikarthi:
> I'm not convinced this is a good idea for the following reasons:
> 
> * The watermarks changed here determine when to pause/resume a connection for throttling purposes. If we increase it, then we risk overwhelming the server. Increasing it by 10x possibly means that we can handle 10x fewer concurrent writers pounding the server (I'm speculating).
> * Accumulating more data might have an impact to latency. I can understand how making more data available to the server increases throughput, though.

There are pros and cons to the settings. If we keep the value to current (which looks like an arbitrary value), then we expect to run more number of writers to compensate the ingest throughput limitation. Increasing the number of writers will increase the no of TCP connections (total segments x connection) and I am not sure if that will have any impact to the system? We need to strike some balance and with the changed settings, I have seen a very decent performance improvement for the Writer.

fpj:
@vijikarthi 

> If we keep the value to current (which looks like an arbitrary value), then we expect to run more number of writers to compensate the ingest throughput limitation. 

Agreed, we have not done a sensitive analysis to determine the optimal value. I do believe we need to optimize for multiple writers, which is likely to be the common case for us.

> Increasing the number of writers will increase the no of TCP connections (total segments x connection) and I am not sure if that will have any impact to the system?

The connection pooling work that is going on should address this concern at least partially.

> We need to strike some balance and with the changed settings, I have seen a very decent performance improvement for the Writer.

Right, that's possible, but it looks like we have tried a single value, could we do a bit more exploration to see what other values give us?

vijikarthi:
I have tested 10KB event size for a total of 10GB records for varying writer(s) with and without the fix. Here is the throughput comparison.

| writer(s) | Throughput (with-fix) | Throughput (without-fix) |
| --- | --- | --- |
| 1 | 170 MB/s | 127 MB/s |
| 5 | 240 MB/s | 198 MB/s |
| 10 | 264 MB/s | 225 MB/s |
| 20 | 264 MB/s | 232 MB/s |

tkaitchuck:
@vijikarthi
The fact that this is increasing throughput at 20 writers does indicate the current value is too small. 
What happens with other values? Like say 4x increase instead of 10x?

kmgowda:
Hi Tom, 
    please see the "pravega-perf-testing" channel.. I have published the results with 1x, 2x, 4x, 6x, 8x and 10x (this PR changes).

kmgowda:
Hi Tom/vijay,
   I did the testing with 10 producers with 10 segments.
There is not much improvement in case of multi producers.
The detailed results are published in the pravega-perf-testing channel.

vijikarthi:
Tested with 10 producers and 10 segments for varying watermark/throttling limits ranging from 2x to 10x from the original values (HIGH_WATER_MARK  128 KB and LOW_WATER_MARK  64 KB) suggests that up to 5x, the performance is looking good and afterwards not much of improvement. The throttling limit hence is changed to 5x of the original value.  

image

PR #3483, opened 2019-03-19
Issue 3464: (SegmentStore) Properly calculating StorageWriter acks

### LINKED ISSUE #3464: Data corruption exception in System tests.
**Problem description**
In some system test runs such as ReadTxnWriteScaleWithFailoverTest, when bookies are repeatedly failing, a `DataCorruptionException` occurs. Example:

```
segmentstore_2019-03-14.59.log:2019-03-14 06:25:48,526 1412251 [core-18] INFO  i.p.s.s.h.ZKSegmentContainerMonitor - Container Changes: Desired = [0, 1, 2, 3], Current = [0, 1, 2, 3], PendingTasks = [], ToStart = [], ToStop = [].
2019-03-14 06:25:52,893 1416618 [core-25] WARN  i.p.s.s.c.StreamSegmentContainer - SegmentContainer[1]: Writer failed. Shutting down StreamSegmentContainer.
java.util.concurrent.CompletionException: io.pravega.segmentstore.server.DataCorruptionException: Unable to retrieve CacheContents for 'AggregatedAppend: SegmentId = 1, Offsets = [892517-2320514), Se
qNo = 18494'.
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
  Caused by: io.pravega.segmentstore.server.DataCorruptionException: Unable to retrieve CacheContents for 'AggregatedAppend: SegmentId = 1, Offsets = [892517-2320514), SeqNo = 18494'.
        at io.pravega.segmentstore.server.writer.SegmentAggregator.getFlushArgs(SegmentAggregator.java:826)
        at io.pravega.segmentstore.server.writer.SegmentAggregator.lambda$flushExcess$14(SegmentAggregator.java:696)
        ... 11 common frames omitted
        
segmentstore_2019-03-14.60.log:2019-03-14 06:25:58,526 1422251 [core-24] INFO  i.p.s.s.h.ZKSegmentContainerMonitor - Container Changes: Desired = [0, 1, 2, 3], Current = [0, 2, 3], PendingTasks = [], ToStart = [1], ToStop = [].
```

andreipaduroiu:
**Change log description**  
- Fixed a bug in `StorageWriter` where it wasn't correctly calculating the ack sequence number for Table Segments.

**Purpose of the change**  
Fixes #3464.

**What the code does**  
- For Table Segments, there are two types of processors: `SegmentAggregator` (that moves data to Tier2) and `WriterTableProcessor` (which indexes table updates)
- When determining the Sequence Number (SN) that's appropriate to truncate Tier 1 at, for each Table Segment, it needs to get the minimum of the SNs that have not yet been processed. 
- If a particular processor is fully caught up, it returns a special value `Operation.NO_SEQUENCE_NUMBER`, which also happens to be negative, hence smaller than any real SN.
- The current code was confused about this and if the Table Processor was ahead of the Segment Aggregator (and fully caught up), it would incorrectly report that both were fully caught up. As such, the Storage Writer would potentially truncate more from Tier 1 than has been moved to Tier 2, which would cause data loss (corruption) upon recovery.
- This was fixed so that the min-calculation ignores fully-caught-up processors and only returns that special value if ALL processors are fully caught up.

**How to verify it**  
New unit test added to test corner case.
System test must pass (1744 did).

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
@@ -543,7 +543,7 @@ class StorageWriter extends AbstractThreadPoolService implements Writer {
     /**
      * Wraps a collection of WriterSegmentProcessors, including the main Segment Aggregator.
      */
-    private static class ProcessorCollection implements WriterSegmentProcessor {
+    private class ProcessorCollection implements WriterSegmentProcessor {
         private final SegmentAggregator aggregator;
         private final List<WriterSegmentProcessor> processors;
 
@@ -594,11 +594,9 @@ class StorageWriter extends AbstractThreadPoolService implements Writer {
 
         @Override
         public long getLowestUncommittedSequenceNumber() {
-            // We collect the lowest value across all processors.
-            return this.processors.stream()
-                                  .mapToLong(WriterSegmentProcessor::getLowestUncommittedSequenceNumber)
-                                  .min()
-                                  .orElse(Operation.NO_SEQUENCE_NUMBER);
+            return this.processors.size() == 1
+                    ? this.processors.get(0).getLowestUncommittedSequenceNumber()
+                    : StorageWriter.this.ackCalculator.getLowestUncommittedSequenceNumber(this.processors);
         }
 
         @Override
--- a/segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/AckCalculatorTests.java
+++ b/segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/AckCalculatorTests.java
@@ -13,6 +13,7 @@ import io.pravega.common.hash.RandomFactory;
 import io.pravega.segmentstore.server.SegmentOperation;
 import io.pravega.segmentstore.server.WriterFlushResult;
 import io.pravega.segmentstore.server.WriterSegmentProcessor;
+import io.pravega.segmentstore.server.logs.operations.Operation;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Random;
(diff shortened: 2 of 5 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
```diff
@@ -594,11 +594,9 @@ public boolean isClosed() {
 
         @Override
         public long getLowestUncommittedSequenceNumber() {
-            // We collect the lowest value across all processors.
-            return this.processors.stream()
-                                  .mapToLong(WriterSegmentProcessor::getLowestUncommittedSequenceNumber)
-                                  .min()
-                                  .orElse(Operation.NO_SEQUENCE_NUMBER);
+            return this.processors.size() == 1
```
tkaitchuck: In the docs for this method it says: "As opposed from {@link #getHighestCommittedSequenceNumber}, this method should be called for {@link WriterSegmentProcessor} instances that deal with the same Segment"

But both paths here are calling `getLowestUncommittedSequenceNumber`

andreipaduroiu: The Javadoc you are referring to is on the `AckCalculator` class. This is inside the `StorageWriter` class and is invoking that method correctly.

This is inside a private nested class named `ProcessorCollection` that holds all the processors for the same Segment. 

The difference here is if we only have one processor (every Stream Segment) then don't bother calling into `AckCalculator` and just return the result from its `SegmentAggregator`. For the (small minority of) TableProcessors, we go down the second path, which invokes the `AckCalculator`.

If you look inside the `AckCalculator.getLowestUncommitedSequenceNumber()` you'll see that you'll get the same result if you only pass in 1 item in the argument.

fpj: For my own understanding, is it right that the check for single processor is only an optimization?  Would it be correct to replace:

```
this.processors.size() == 1
                    ? this.processors.get(0).getLowestUncommittedSequenceNumber()
                    : StorageWriter.this.ackCalculator.getLowestUncommittedSequenceNumber(this.processors);
```

with:

```
StorageWriter.this.ackCalculator.getLowestUncommittedSequenceNumber(this.processors);
```

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/AckCalculatorTests.java
```diff
@@ -94,6 +95,36 @@ public void testGetHighestCommittedSequenceNumber() {
...
+        for (int i = 0; i < processorCount; i++) {
+            processors.add(new TestProcessor());
+        }
+
+        // Everything up-to-date.
+        processors.forEach(p -> p.setLowestUncommittedSequenceNumber(Operation.NO_SEQUENCE_NUMBER));
+        long lusn = calc.getLowestUncommittedSequenceNumber(processors);
+        Assert.assertEquals("Unexpected result when all processors up-to-date.", Operation.NO_SEQUENCE_NUMBER, lusn);
+
+        // One processor not up-to-date.
```
fpj: Without the change of this PR, is it right that this case and the one starting in line 122 would fail?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
andreipaduroiu: yes , they are equivalent. The vast majority of segments will be going down the first path.

andreipaduroiu: yes, they are equivalent. The vast majority of segments will go down the first path.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/AckCalculatorTests.java
andreipaduroiu: No, just this one would have failed. The previous code worked correctly for checks at line 112 and 122.

fpj:
I checked the codecov report, and it is complaining about the string formatting code.

PR #3491, opened 2019-03-20
Issue 3489: FileSystemStorage should write data synchronously to avoid data loss

### LINKED ISSUE #3489: FileSystemStorage should write data synchronously to avoid data loss
**Problem description**
Currently the FileSystemStorage implementation uses FileChannel to write data. However it does not use force() to sync data to disk. 
In other words, data is not guaranteed to be written to disk immediately, because data is transferred only to the  buffer cache. This can result in data loss if the server crashes before data is written to the disk.

**Problem location**
io.pravega.storage.filesystem.FileSystemStorage

Call FileChannel.force after each write
or open the file using [StandardOpenOption.SYNC](https://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html#SYNC).

sachin-j-joshi:
Call FileChannel.force() at the end of write and concat.

**Change log description** 
Call FileChannel.force() at the end of write and concat.
This prevents data loss if the server crashes before data is written to the disk. Without this the data written is not guaranteed to be written to disk immediately, as data is transferred only to the buffer cache. 

**Purpose of the change**  
Fixes #3489 

**What the code does**  
Call FileChannel.force() at the end of write and concat.

**How to verify it**  
System tests should pass. There is should be no DataCorruptionException in the logs.

### DIFF
--- a/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
+++ b/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
@@ -319,6 +319,7 @@ public class FileSystemStorage implements SyncStorage {
                     totalBytesWritten += bytesWritten;
                     length -= bytesWritten;
                 }
+                channel.force(false);
             }
             FileSystemMetrics.WRITE_LATENCY.reportSuccessEvent(timer.getElapsed());
             FileSystemMetrics.WRITE_BYTES.add(totalBytesWritten);
@@ -377,6 +378,7 @@ public class FileSystemStorage implements SyncStorage {
                 offset += bytesTransferred;
                 length -= bytesTransferred;
             }
+            targetChannel.force(false);
             Files.delete(sourcePath);
             LoggerHelpers.traceLeave(log, "concat", traceId);
             return null;
--- a/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
+++ b/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
@@ -303,29 +303,29 @@ public class FileSystemStorage implements SyncStorage {
             throw new StreamSegmentSealedException(handle.getSegmentName());
         }
 
-        long fileSize = path.toFile().length();
-        if (fileSize < offset) {
-            throw new BadOffsetException(handle.getSegmentName(), fileSize, offset);
-        } else {
-            long totalBytesWritten = 0;
-            try (FileChannel channel = FileChannel.open(path, StandardOpenOption.WRITE)) {
-                // Wrap the input data into a ReadableByteChannel, but do not close it. Doing so will result in closing
-                // the underlying InputStream, which is not desirable if it is to be reused.
-                ReadableByteChannel sourceChannel = Channels.newChannel(data);
-                while (length != 0) {
-                    long bytesWritten = channel.transferFrom(sourceChannel, offset, length);
-                    assert bytesWritten > 0 : "Unable to make any progress transferring data.";
-                    offset += bytesWritten;
-                    totalBytesWritten += bytesWritten;
-                    length -= bytesWritten;
-                }
-                channel.force(false);
+        long totalBytesWritten = 0;
+        try (FileChannel channel = FileChannel.open(path, StandardOpenOption.WRITE)) {
+            long fileSize = channel.size();
+            if (fileSize != offset) {
+                throw new BadOffsetException(handle.getSegmentName(), fileSize, offset);
             }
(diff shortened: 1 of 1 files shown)


bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -319,6 +319,7 @@ private Void doWrite(SegmentHandle handle, long offset, InputStream data, int le
                     totalBytesWritten += bytesWritten;
                     length -= bytesWritten;
                 }
+                channel.force(true);
```
fpj: Can you justify why we also need to force a metadata sync? You need to reason about what part of the metadata you are actually updating.

sachin-j-joshi: I think you are right. We don't need to explicitly synchronize metadata to update file size on disk, it should be already done as part of data sync itself.

sachin-j-joshi: doublechecked fdatasync() documentation, filesize changes automatically trigger metadata flush . 

> It does not flush modified metadata _unless that metadata is needed in order to allow a subsequent data retrieval to be correctly handled_. On the other hand, a change to the file size, would require a metadata flush.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -319,6 +319,7 @@ private Void doWrite(SegmentHandle handle, long offset, InputStream data, int le
                     totalBytesWritten += bytesWritten;
                     length -= bytesWritten;
                 }
+                channel.force(false);
```
andreipaduroiu: Actually, since you are in here, would you mind fixing the File size check at line 307?

We can check the file size and throw Bad Offset exception after we've opened the FileChannel by checking `FileChannel.size`.

At least this should reduce the number of IO calls a bit.

sachin-j-joshi: sure.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -303,28 +303,29 @@ private Void doWrite(SegmentHandle handle, long offset, InputStream data, int le
...
-                    long bytesWritten = channel.transferFrom(sourceChannel, offset, length);
-                    assert bytesWritten > 0 : "Unable to make any progress transferring data.";
-                    offset += bytesWritten;
-                    totalBytesWritten += bytesWritten;
-                    length -= bytesWritten;
-                }
+        long totalBytesWritten = 0;
+        try (FileChannel channel = FileChannel.open(path, StandardOpenOption.WRITE)) {
+            long fileSize = channel.size();
+            if (fileSize != offset) {
```
fpj: The condition in the old code to throw bad offset is that the filesize is smaller than than the offset, but here it is changing to different (`!=`), why?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -303,28 +303,29 @@ private Void doWrite(SegmentHandle handle, long offset, InputStream data, int le
...
+            // the underlying InputStream, which is not desirable if it is to be reused.
+            ReadableByteChannel sourceChannel = Channels.newChannel(data);
+            while (length != 0) {
+                long bytesWritten = channel.transferFrom(sourceChannel, offset, length);
+                assert bytesWritten > 0 : "Unable to make any progress transferring data.";
+                offset += bytesWritten;
+                totalBytesWritten += bytesWritten;
+                length -= bytesWritten;
+            }
+            channel.force(false);
```
fpj: Ok, we are going with only forcing the metadata update in the case it is strictly necessary, which we are delegating to the underlying OS implementation, yes?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
sachin-j-joshi: To make it consistent with how other storage (HDFS and S3) are checking the same conditions. @andreipaduroiu suggested the change and I agree with the suggestion.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
sachin-j-joshi: Yes. In this case the underlying OS will most sync metadata. (Which will happen most of the time with mostly append-only writes in our case.)

PR #3492, opened 2019-03-21
Issue 3411: Improve permissions required for Controller gRPC API 

### LINKED ISSUE #3411: Inconsistencies in the required permissions of operations
**Problem description**
In the [table of operations, permissions, and rights](https://github.com/pravega/pravega/issues/3277#issuecomment-466900097) of issue #3277 by @ravisharda , there are some other inconsistencies:
1- The following operations require `READ_UPDATE`, but they are really about reading the state of the resource:

* List current stream segments (16)
* List stream segments (17)
* Check scale status (21)
* Is segment valid (23)
* Check transaction state (29)
* Get delegation token (30)

Cases 21 and 29 are arguable because they are part of operations that change the state of streams, but they are not changing the status of anything. About 30, I don't understand why it needs to be `READ_UPDATE`.

2- About case 28: Ping transaction. Pinging a transaction actually changes the state of the transaction, so I think it should be `READ_UPDATE`.

**Problem location**
Security.

Revisit the code for these operations and change required permissions.

ravisharda:
**Change log description**  

Some of the operations of the Controller gRPC interface require a higher permission level `READ_UPDATE`, even though they are read operations and do not modify the state of the resource. And then, there is one operation that does modify the state, but only requires `READ` permission. 

**Purpose of the change**  
Fixes #3411. 

**What the code does**  

The code fixes the permissions required for the operations listed below. We continue to use the same serial numbers (S.No.) as mentioned in the issue so that it is easier for readers to keep track of the changes. 

|S.No.| Action |	Resource String for Auth (in this PR)  | Required Permission - "OLD" | Required Permission - "NEW"| gRPC method (in ``ControllerServiceImpl``) | REST controller method (in ``StreamMetadataResourceImpl``| 
|:--:|:-----------|:---------------:|:---------------:|:---------------:|:---------------:|:----------------------|
|21.| Check scale status | ``<scope-name>/<stream-name>`` | READ_UPDATE | READ | ``checkScale``|-|
|22.| Get URI for segment | ``<scope-name>/<stream-name>`` | READ_UPDATE | READ | ``getURI``|-|
|23.| Is Segment valid | ``<scope-name>/<stream-name>`` | READ_UPDATE | READ | ``isSegmentValid``|-|
|24.| Is stream cut valid | ``<scope-name>/<stream-name>`` | READ_UPDATE | READ | ``isStreamCutValid``|-|
|28.| Ping transaction | ``<scope-name>/<stream-name>`` | READ | READ_UPDATE | ``pingTransaction``|-|
|29.| Check transaction state | ``<scope-name>/<stream-name>`` | READ_UPDATE | READ | ``checkTransactionState``|-|  

Also, a new test class `ControllerGrpcAuthFocusedTest` has been added for verifying authorization logic of the Controller gRPC service implementation `ControllerServiceImpl`.

**Note:**
* The permission required for `Get delegation token`/`getDelegationToken` has been left as-is (`READ_UPDATE`) as the operation is state-mutative.
* The permission required for `List current stream segments`/`getCurrentSegments()` and `List stream segments`/`getSegments()` have been left as-is (`READ_UPDATE`) as the operations use the permission to create a delegation token for downstream processing.

**How to verify it**  

All tests should pass. 

For convenience, I also create a temporary test suite comprising of a subset of test classes as shown below, for verifying code when there are any auth-related changes in the controller interfaces:

```java
import io.pravega.controller.rest.v1.SecureStreamMetaDataTests;
import io.pravega.controller.rest.v1.StreamMetaDataAuthFocusedTests;
import io.pravega.controller.rest.v1.StreamMetaDataTests;
import io.pravega.controller.rest.v1.UserSecureStreamMetaDataTests;
import io.pravega.controller.server.v1.InMemoryControllerServiceImplTest;
import io.pravega.controller.server.v1.ZKControllerServiceImplTest;

@RunWith(Suite.class)
@Suite.SuiteClasses({
        StreamMetaDataAuthFocusedTests.class,
        StreamMetaDataTests.class,
        SecureStreamMetaDataTests.class,
        UserSecureStreamMetaDataTests.class,
        InMemoryControllerServiceImplTest.class,
        ZKControllerServiceImplTest.class,
        RESTAuthHelperTest.class,
        PravegaAuthManagerTest.class,
        ControllerGrpcAuthFocusedTest.class
})

// Run this class on your IDE to execute the test classes in the suite. 
public class AuthTestSuite {
}
```

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
@@ -108,7 +108,7 @@ public class PasswordAuthHandler implements AuthHandler {
      *                     by this instance.
      */
     @VisibleForTesting
-    public void initialize(String passwordFile) {
+    void initialize(String passwordFile) {
         loadPasswordFile(passwordFile);
     }
 
--- a/controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
@@ -9,6 +9,7 @@
  */
 package io.pravega.controller.server.rpc.auth;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
@@ -97,6 +98,18 @@ public class PasswordAuthHandler implements AuthHandler {
             throw new CompletionException(new AuthenticationException(userName));
         }
         return authorizeForUser(userMap.get(userName), resource);
+    }
+
+    /**
+     * This method exists expressly for unit testing purposes. It loads the contents of the specified
+     * {@code passwordFile} into this object.
+     *
+     * @param passwordFile the file with a list of users, their encrypted passwords and their ACLs that is to be used
+     *                     by this instance.
+     */
+    @VisibleForTesting
+    public void initialize(String passwordFile) {
+        loadPasswordFile(passwordFile);
     }
 
     @Override
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -170,7 +170,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
         log.info("getCurrentSegments called for stream {}/{}.", request.getScope(), request.getStream());
         authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorizationAndCreateToken(
                 AuthResourceRepresentation.ofStreamInScope(request.getScope(), request.getStream()),
-                AuthHandler.Permissions.READ_UPDATE),
+                AuthHandler.Permissions.READ),
                 delegationToken -> controllerService.getCurrentSegments(request.getScope(), request.getStream())
                                        .thenApply(segmentRanges -> SegmentRanges.newBuilder()
                                                                                 .addAllSegmentRanges(segmentRanges)
@@ -186,7 +186,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
         authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorizationAndCreateToken(
                 AuthResourceRepresentation.ofStreamInScope(request.getStreamInfo().getScope(),
                         request.getStreamInfo().getStream()),
-                AuthHandler.Permissions.READ_UPDATE),
+                AuthHandler.Permissions.READ),
                 delegationToken -> controllerService.getSegmentsAtHead(request.getStreamInfo().getScope(),
                         request.getStreamInfo().getStream())
                                        .thenApply(segments -> {
@@ -257,7 +257,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                 request.getStreamInfo().getStream());
         authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
                 AuthResourceRepresentation.ofStreamInScope(request.getStreamInfo().getScope(), request.getStreamInfo().getStream()),
-                AuthHandler.Permissions.READ_UPDATE),
+                AuthHandler.Permissions.READ),
                 delegationToken -> controllerService.checkScale(request.getStreamInfo().getScope(), request.getStreamInfo().getStream(),
                         request.getEpoch()), responseObserver);
     }
@@ -370,7 +370,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                 request.getStreamInfo().getStream(), request.getTxnId());
         authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
                 AuthResourceRepresentation.ofStreamInScope(request.getStreamInfo().getScope(), request.getStreamInfo().getStream()),
-                AuthHandler.Permissions.READ_UPDATE),
+                AuthHandler.Permissions.READ),
                 delegationToken -> controllerService.checkTransactionStatus(request.getStreamInfo().getScope(),
                         request.getStreamInfo().getStream(),
                         request.getTxnId()),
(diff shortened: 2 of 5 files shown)


ravisharda:
@shiveshr yes, changing the permission for `getCurrentSegments()` from `READ` to `READ_UPDATE` was causing some tests in package `io.pravega.local` to block. The issue was occurring as the token being generated for this method was being passed along as delegation token for SegmentRanges. I've modified the permission back to `READ_UPDATE` and will take a relook again.

ravisharda:
In addition to existing tests that verify authorization, new unit tests should be added to test the changes more comprehensively, as suggested by @shiveshr. That task is pending.

ravisharda:
> ```
> io.pravega.local.SecurePravegaClusterTest > failingCreateTestStream PASSED
> No output has been received in the last 35m0s, this potentially indicates a stalled build or something wrong with the build itself.
> Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
> ```
> Some test seems to be stuck without failing because of these permissions change.. can you please investigate.
> 
> Also add specific unit tests if possible to see the impact of these permission changes.

@shiveshr As suggested, I have added tests (in a new test class `io.pravega.controller.server.rpc.auth.ControllerGrpcAuthFocusedTest`) to verify permission changes.

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
```diff
@@ -0,0 +1,26 @@
...
+public class AuthFileUtils {
+
+    public static String credentialsAndAclAsString(String username, String password, String acl) {
+        Preconditions.checkArgument(!Strings.isNullOrEmpty(username)
+                && !Strings.isNullOrEmpty(password)
+                && acl != null
+                && !acl.startsWith(":"));
+
+        // This will return a string that looks like this:"<username>:<pasword>:acl\n"
+        return String.format("%s:%s:%s%n", username, password, acl);
```
fpj: Is there a typo here? Should it be `%s:%s:%s\n` instead?

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
```diff
@@ -0,0 +1,478 @@
...
+    @AfterClass
+    public static void classTearDown() {
+        if (AUTH_FILE != null && AUTH_FILE.exists()) {
+            AUTH_FILE.delete();
+        }
+        ExecutorServiceHelpers.shutdown(EXECUTOR);
+    }
+
+    @Before
+    public void setup() throws IOException {
```
fpj: The tests in this class are really short (which is a plus), but this setup is really long. I'm wondering whether we can shorten the per-test setup by making some of these elements common.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -356,7 +356,7 @@ public void pingTransaction(PingTxnRequest request, StreamObserver<PingTxnStatus
                 request.getStreamInfo().getStream(), request.getTxnId());
         authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
                 AuthResourceRepresentation.ofStreamInScope(request.getStreamInfo().getScope(), request.getStreamInfo().getStream()),
-                AuthHandler.Permissions.READ),
+                AuthHandler.Permissions.READ_UPDATE),
```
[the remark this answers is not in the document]
ravisharda: This change in permission here implies that the user account will now require `READ_UPDATE` permission on the resource (`<scope>/<stream>`) as opposed to `READ`. If a client application was using a credential that only had `READ` permission on the stream, a call to this method will now be unauthorized and fail. In that sense, it is not a backward compatible change. Does that answer your question?

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
ravisharda: `%n` outputs a platform-independent newline character as per the following Oracle tutorial, which boils down to `\n` on most platforms.  It is not a typo. https://docs.oracle.com/javase/tutorial/java/data/numberformat.html

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
ravisharda: The tests were pretty fast, as we are:
* creating in-memory Pravega objects (like host store, stream store, etc.), 
* using some mock objects (such as Cluster), and 
* using a special in-process gRPC server, which is used in some of the unit test samples in gRPC repo (for example, https://github.com/grpc/grpc-java/blob/master/examples/src/test/java/io/grpc/examples/header/HeaderServerInterceptorTest.java). 

Running test initializers (@Before method) before each test makes the tests independent and parallelizable. I think the efficiency of the initializers isn't much of a concern in the tests in this class.

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
```diff
@@ -0,0 +1,26 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.controller.auth;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+
+public class AuthFileUtils {
```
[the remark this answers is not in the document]
ravisharda: This is specific to the controller tests, where I used it to create entries for the file used by the Password Auth Hander in a more convenient manner.

controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
```diff
@@ -99,6 +100,18 @@ public Permissions authorize(String resource, Principal principal) {
...
 
+    /**
+     * This method exists expressly for unit testing purposes. It loads the contents of the specified
+     * {@code passwordFile} into this object.
+     *
+     * @param passwordFile the file with a list of users, their encrypted passwords and their ACLs that is to be used
+     *                     by this instance.
+     */
+    @VisibleForTesting
+    public void initialize(String passwordFile) {
```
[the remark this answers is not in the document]
ravisharda: Yes, you are right: the class `ControllerGrpcAuthFocusedTest` that uses it is in the same package as the class `PasswordAuthHandler`. So, the visibility of this method can be reduced. 

Having said that, this method shall find use outside of the package as well - for tests in other places. Should we leave it public or make it public when such a need arises? I am leaning towards the former. Will make the change in the next commit.

PR #3496, opened 2019-03-21
Issue 3495: Split metadata scalability test into two separate tests

### LINKED ISSUE #3495: Break metadata scalability test into two separate tests
**Problem description**
Metadata scalability test is very resource hungry as we create 10k segments and then scale the stream over 1000 times in a loop. 

The intent of the test is to test two separate capabilities -
1. verify if we can create very large number of segments for a stream
2. verify that metadata queries work even after 1000+ scales (as epoch history time series and sealed segment map shards chunk after 1000 entries)

**Problem location**
MetadataScalabilityTest system test

Break it up into two separate tests.

shiveshr:
**Change log description**  
Split metadata scalability test into two separate tests so that they are individually not resource hungry and test intended scenarios. 

**Purpose of the change**  
Fixes #3495

**What the code does**  
1. Metadata operations on stream with large number of segments like scale and truncate
2. Metadata operation on stream with large number of epochs (perform 1000s of scales) and then truncate. 

One additional change of config file having incorrect default zookeeper port is fixed. 

PS: running truncation on large stream has a huge memory footprint because we loop over all segments in streamcut and that actually reads epoch record from cache and deserializes it for each segment. so that happens 10k times and puts a huge memory pressure on controller. I have fixed that issue as part of issue 3478. Will include truncation related changes for large stream in that branch.

This large epoch record

**How to verify it**  
System test should pass

### DIFF
--- a/controller/src/main/java/io/pravega/controller/util/Config.java
+++ b/controller/src/main/java/io/pravega/controller/util/Config.java
@@ -139,7 +139,7 @@ public final class Config {
     private static final Property<String> PROPERTY_TLS_TRUST_STORE = Property.named("auth.tlsTrustStore", "");
     private static final Property<String> PROPERTY_TLS_KEY_FILE = Property.named("auth.tlsKeyFile", "");
     private static final Property<String> PROPERTY_TOKEN_SIGNING_KEY = Property.named("auth.tokenSigningKey", "");
-    private static final Property<String> PROPERTY_ZK_URL = Property.named("zk.url", "localhost:2121");
+    private static final Property<String> PROPERTY_ZK_URL = Property.named("zk.url", "localhost:2181");
     private static final Property<Integer> PROPERTY_ZK_RETRY_MILLIS = Property.named("zk.retryIntervalMillis", 5000);
     private static final Property<Integer> PROPERTY_ZK_MAX_RETRY_COUNT = Property.named("maxRetries", 5);
     private static final Property<Integer> PROPERTY_ZK_SESSION_TIMEOUT_MILLIS = Property.named("sessionTimeoutMillis", 10000);
--- a/test/system/src/test/java/io/pravega/test/system/MetadataScalabilityLargeNumSegmentsTest.java
+++ b/test/system/src/test/java/io/pravega/test/system/MetadataScalabilityLargeNumSegmentsTest.java
@@ -1,0 +1,89 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.test.system;
+
+import io.pravega.client.segment.impl.Segment;
+import io.pravega.client.stream.ScalingPolicy;
+import io.pravega.client.stream.StreamConfiguration;
+import io.pravega.client.stream.impl.ControllerImpl;
+import io.pravega.shared.segment.StreamSegmentNameUtils;
+import io.pravega.test.system.framework.SystemTestRunner;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+/**
+ * This test creates a stream with 10k segments and then rapidly scales it 10 times.
+ * Then it performs truncation a random number of times.
+ */
+@Slf4j
+@RunWith(SystemTestRunner.class)
+public class MetadataScalabilityLargeNumSegmentsTest extends MetadataScalabilityTest {
+    private static final String STREAM_NAME = "metadataScalabilitySegments";
+    private static final int NUM_SEGMENTS = 10000;
+    private static final StreamConfiguration CONFIG = StreamConfiguration.builder()
+                                                                         .scalingPolicy(ScalingPolicy.fixed(NUM_SEGMENTS)).build();
+    private static final int SCALES_TO_PERFORM = 10;
+    
+    private final AtomicInteger counter = new AtomicInteger(0);
+
+    @Override
+    String getStreamName() {
+        return STREAM_NAME;
+    }
+
+    @Override
+    StreamConfiguration getStreamConfig() {
+        return CONFIG;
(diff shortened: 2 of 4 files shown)


controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -139,7 +139,7 @@
     private static final Property<String> PROPERTY_TLS_TRUST_STORE = Property.named("auth.tlsTrustStore", "");
     private static final Property<String> PROPERTY_TLS_KEY_FILE = Property.named("auth.tlsKeyFile", "");
     private static final Property<String> PROPERTY_TOKEN_SIGNING_KEY = Property.named("auth.tokenSigningKey", "");
-    private static final Property<String> PROPERTY_ZK_URL = Property.named("zk.url", "localhost:2121");
+    private static final Property<String> PROPERTY_ZK_URL = Property.named("zk.url", "localhost:2181");
```
RaulGracia: Unrelated but necessary to change, I also found this problematic in a local deployment.

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityLargeNumSegmentsTest.java
```diff
@@ -0,0 +1,89 @@
...
+                                          .map(Segment::getSegmentId).collect(Collectors.toList());
+        Map<Double, Double> newRanges = new HashMap<>();
+        double delta = 1.0 / NUM_SEGMENTS;
+        newRanges.put(delta * (i - 1), delta * i);
+
+        return new ImmutablePair<>(segmentsToSeal, newRanges);
+    }
+
+    @Test
+    public void largeSegmentsScalability() {
```
RaulGracia: Maybe the test could be renamed as `parallelSegmentsScalabilityTest` or something similar, as I suppose that the goal of the test is to have many segments in parallel, not large ones.

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityLargeNumSegmentsTest.java
```diff
@@ -0,0 +1,89 @@
...
+        testState = new TestState(false);
+
+        ControllerImpl controller = getController();
+
+        List<List<Segment>> listOfEpochs = scale(controller);
+        // TODO: uncomment truncation as part of #3478
```
RaulGracia: So, is there a cyclic dependency here? PR 3478 assumes this to be merged first, but then this depends on 3478?

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityTest.java
```diff
@@ -47,24 +46,14 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
-/**
- * This test creates a stream with 10k segments and then rapidly scales it 1010 times.
- * Then it performs truncation a random number (less than 1010) of times. 
- */
 @Slf4j
```
RaulGracia: Adding some comments to describe the general purpose of this abstract class would be great.

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityTest.java
```diff
@@ -47,24 +46,14 @@
...
-public class MetadataScalabilityTest extends AbstractScaleTests {
-    private static final String STREAM_NAME = "metadataScalability";
-    private static final int NUM_SEGMENTS = 10000;
-    private static final StreamConfiguration CONFIG = StreamConfiguration.builder()
-                                                                         .scalingPolicy(ScalingPolicy.fixed(NUM_SEGMENTS)).build();
-    private static final int SCALES_TO_PERFORM = 1010;
-
+public abstract class MetadataScalabilityTest extends AbstractScaleTests {
     @Rule
     public Timeout globalTimeout = Timeout.seconds(60 * 60);
```
RaulGracia: Are these tests going to take up to 1 hour each? We need to think to the system test build time as well.

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityTest.java
```diff
@@ -101,35 +90,37 @@ public void tearDown() {
...
+    abstract StreamConfiguration getStreamConfig();
+
+    abstract String getStreamName();
 
-        ControllerImpl controller = getController();
+    abstract int getScalesToPerform();
+
+    abstract Pair<List<Long>, Map<Double, Double>> getScaleInput(ArrayList<Segment> sortedCurrentSegments);
+
+    List<List<Segment>>  scale(ControllerImpl controller) {
```
RaulGracia: Extra space in method signature.

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityLargeNumSegmentsTest.java
shiveshr: PR 3478 is built on top of this branch. 
in PR 3478 we uncomment the truncation.

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityTest.java
shiveshr: Observation is that it takes about 25 minutes

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityLargeNumSegmentsTest.java
shiveshr: the goal is to test large number of segments per epoch. 
i could rename it to large number of segments. parallel segments seems odd.

test/system/src/test/java/io/pravega/test/system/MetadataScalabilityTest.java
```diff
@@ -138,40 +135,52 @@ public void scalability() {
...
+         // try SCALES_TO_PERFORM randomly generated stream cuts and truncate stream at those 
+         // stream cuts. 
+         List<AtomicInteger> indexes = new LinkedList<>();
+         Random rand = new Random();
+         for (int i = 0; i < numSegments; i++) {
+             indexes.add(new AtomicInteger(1));
+         }
+         Futures.loop(() -> indexes.stream().allMatch(x -> x.get() < scalesToPerform - 1), () -> {
+             // randomly generate a stream cut. 
+             // Note: From epoch 1 till epoch SCALES_TO_PERFORM each epoch is made up of numOfSegments segments
```
fpj: Let's improve this description. Here is my recommendation:

```
We randomly generate a stream cut in each iteration of this loop. A valid stream
cut in this scenario contains for each position i in [0, numSegments -1], a segment
from one of the scale epochs of the stream. For each position i, we randomly
choose an epoch and pick the segment at position i. It increments the epoch
index accordingly (indexes list) so that in the next iteration it chooses a later
epoch for the same i.

Because the segment in position i always contain the range [d * (i-1), d * i],
where d = 1 / (number of segments), the stream cut is guaranteed to cover
the entire key space. 
```

Issue #1761, opened 2017-08-18
Multiple controllers send differing segment count/splits/merges metrics on the same stream


jkhalack:
**Problem description**

On clusters with two Pravega controller instances, if two controllers are reporting `segments_count`, `segment_splits`, and `segment_merges` metrics on the same stream, the numbers being reported differ between the controllers.

One of two controllers would report stale metrics as of the moment when the stream was created (initial number of segments for `segments_count`, and zeros for `segment_splits` and `segment_merges`).
The second controller reports the actual number of active segments in `segments_count`, but much bigger numbers than the actual metrics for `segment_splits` and `segment_merges` would have to be.

###### Example:
One controller is reporting 
```
53e392ef-27ad-4b8b-be0b-198a14b7b7b1-0004.pravega_controller__b9051a32-5ff8-4083-a343-702f4ae1a34a.66cd021c-23da-46eb-88a3-d64166d5c1ab.pravega.controller.segments_count.hulk.smallScale.Gauge:1|g
```
The other controller is reporting
```
53e392ef-27ad-4b8b-be0b-198a14b7b7b1-0004.pravega_controller__9ea16bb9-05b6-424e-9066-e799a51b9134.5810a308-b5f6-4adf-ac65-b6a680cb7fbb.pravega.controller.segments_count.hulk.hulk.smallScale.Gauge:15|g
```
*(the correct number of active segments is 15, as confirmed by `scaling-events` API)*

###### Another example:
One controller is reporting 
```
c41e8d11-b058-4bf1-883a-829185d8134d-0004.pravega_controller__b5eed754-a70a-4200-8299-1c570b57902b.67f3abd2-4365-45bf-9bc9-3b9184b5330c.pravega.controller.segment_splits.hulk.smallScale.Counter:0|g
```
The other controller is reporting
```
c41e8d11-b058-4bf1-883a-829185d8134d-0004.pravega_controller__2a23b6bb-cca4-4543-a432-991d8fa62b6d.48d93a29-27f7-4257-9711-c4c5f1fa1c09.pravega.controller.segment_splits.hulk.smallScale.Counter:2453|g
``` 
*(the correct number would be below 400, as the total number of ever created segments is less than 400 for the stream)*

If only one of the two controllers is reporting the metrics for a given stream, then all the three metrics are reported correctly.

**Problem location**
Pravega controller - multiple instances

**Suggestions for an improvement**
Multiple controllers should report identical metrics;
[...]

PR #3504, opened 2019-03-22
Issue 3405:  Add tags to metrics with dimensions

### LINKED ISSUE #3405: Add tags to those metrics with dimensions

kevinhan88:
**Change log description**  
This is the PR to replace the original PR for issue 3405: https://github.com/pravega/pravega/pull/3453

Why to have this PR to replace the original one (PR 3453)?
1. Somehow the clock on my VirtualBox VM was days behind the real time (from the PR mentioned above you can see the commit time is about a week behind), this makes re-base not functioning well and caused big mess.
2. The PR 3453 is based on issue-3319 PR but was submitted before its dependent PR was merged; possibly due to this reason, it is not showing difference based on the latest master (e.g. there should be 31 files changed but PR 3453 always shows 55 files changed). This makes code review difficult.
3. If we stick with the PR 3453, its merging into master may likely encounter issues again as we experienced for issue-3319. Instead of trying 2nd time by that time, I'm doing it now.
4. There are just a few review comments left on PR 3453, which have been all addressed.
5. I have to take personal off tomorrow so I'd like to have this decent clean PR for the team.

Sorry for the inconvenience or confusion it may cause.

**Purpose of the change**  
Closes #3405 

**What the code does**  
PR for issue-3319 has been merged, which baked the API changes of Micrometer into Pravega.
This PR starts using the new API:

Whenever the code operates on metrics with dimensions (those with host, containerId, stream, segments, transaction), the old way is to call utility method such as nameFromContainer, nameFromHost, nameFromStream etc to form a flat metric name.
With the new API, application code calls hostTag, streamTag, segmentTag etc to supply tags, together with the metric name to identify/locate a unique metric object.
As metric tags (dimensions) are provided directly by Pravega, it is now possible to skip Telegraf entirely which is used to map the flat metric name back to tags. (This has been manually tested end-to-end by the Platform team)

**How to verify it**  
All the tests should pass.


jkhalack:
Could you add the `statsdHost` => `statsDHost` change to https://github.com/kevinhan88/pravega/blob/master/config/config.properties file, please?
Same for `statsDHport` and `enableStatsDReporter`. Thanks.

Also, it would be great to update the metrics names in the documentation:
https://github.com/kevinhan88/pravega/blob/master/documentation/src/docs/metrics.md

documentation/src/docs/metrics.md
```diff
@@ -7,46 +7,63 @@ You may obtain a copy of the License at
...
+[Distribution Summary](https://micrometer.io/docs/concepts#_distribution_summaries)); 
+while OpStatsLogger is a sub-metric for complex ones (Timer/Distribution Summary).
 ## 1.1. Metrics Service Provider — Interface StatsProvider
-The starting point of Pravega Metric framework is the StatsProvider interface, it provides start and stop method for Metric service. Regarding the reporters, currently we have support for CSV reporter and StatsD reporter.
+The starting point of Pravega Metric framework is the StatsProvider interface, it provides start and stop method for Metric service.
+It also provides startWithoutExporting() for unit testing purpose which only store metrics in memory without exporting them
+to outside destinations. 
+Currently we have support StatsD and InfluxDB registries.
 ```java
 public interface StatsProvider {
```
andreipaduroiu: Can you please remove this code and add a link to the `StatsProvider` interfaces in the codebase?

documentation/src/docs/metrics.md
```diff
@@ -7,46 +7,63 @@ You may obtain a copy of the License at
...
 ## 1.2. Metric Logger — interface StatsLogger
-Using this interface we can register required metrics for simple types like [Counter](http://metrics.dropwizard.io/3.1.0/manual/core/#counters) and [Gauge](http://metrics.dropwizard.io/3.1.0/manual/core/#gauges) and some complex statistics type of Metric OpStatsLogger, through which we provide [Timer](http://metrics.dropwizard.io/3.1.0/manual/core/#timers) and [Histogram](http://metrics.dropwizard.io/3.1.0/manual/core/#histograms).
+Using this interface we can register required metrics for simple types like 
+[Counter](https://micrometer.io/docs/concepts#_counters) and 
+[Gauge](https://micrometer.io/docs/concepts#_gauges) 
+and some complex statistics type of Metric OpStatsLogger, through which we provide 
+[Timer](https://micrometer.io/docs/concepts#_timers) and 
+[Distribution Summary](https://micrometer.io/docs/concepts#_distribution_summaries).
 ```java
 public interface StatsLogger {
```
andreipaduroiu: Same here

documentation/src/docs/metrics.md
```diff
@@ -7,46 +7,63 @@ You may obtain a copy of the License at
...
 - createMeter(): Create and register a Meter metric.
 - registerGauge(): Register a Gauge metric.
 - createScopeLogger(): Create the stats logger under given scope name.
 
 ### 1.3. Metric Sub Logger — OpStatsLogger
-OpStatsLogger provides complex statistics type of Metric, usually it is used in operations such as CreateSegment, ReadSegment, we could use it to record the number of operation, time/duration of each operation.
+OpStatsLogger provides complex statistics type of Metric, usually it is used in operations such as CreateSegment, 
+ReadSegment, we could use it to record the number of operation, time/duration of each operation.
 ```java
 public interface OpStatsLogger {
```
andreipaduroiu: And here

documentation/src/docs/metrics.md
```diff
@@ -116,204 +134,169 @@ public final class ServiceStarter {
 ```
 
 ## 2.1. Example for Dynamic Counter and OpStatsLogger(Timer)
-This is an example from io.pravega.segmentstore.server.host.handler.PravegaRequestProcessor. In this class, we registered two metrics: One timer (createStreamSegment), one dynamic counter (segmentReadBytes).
+This is an example from io.pravega.segmentstore.server.host.stat.SegmentStatsRecorderImpl.java. In this class, we 
+registered two metrics: One timer (createStreamSegment), one dynamic counter (dynamicLogger).
 ```java
-public class PravegaRequestProcessor extends FailingRequestProcessor implements RequestProcessor {
+public class SegmentStatsRecorderImpl implements SegmentStatsRecorder {
```
andreipaduroiu: And this.

Most of this code is irrelevant to metrics.

documentation/src/docs/metrics.md
```diff
@@ -116,204 +134,169 @@ public final class ServiceStarter {
...
 
 
-### 2.2. Example for Dynamic Gauge and OpStatsLogger(Histogram) 
-This is an example from io.pravega.controller.store.stream.AbstractStreamMetadataStore. In this class, we report a Dynamic Gauge which represents the open transactions and  one histogram (CREATE_STREAM).
+### 2.2. Example for Dynamic Gauge 
+This is an example from io.pravega.controller.metrics.StreamMetrics. In this class, we report 
+a dynamic gauge which represents the open transactions of a stream.
 ```java
-public abstract class AbstractStreamMetadataStore implements StreamMetadataStore  {
+public final class StreamMetrics extends AbstractControllerMetrics implements AutoCloseable {
```
andreipaduroiu: And here. Really, I don't understand the need for copy-pasting code into this document.

documentation/src/docs/metrics.md
```diff
@@ -116,204 +134,169 @@ public final class ServiceStarter {
...
+To Easy unit testing, Micrometer also provides SimpleMeterRegistry, which has memory only storage but no exporting; 
+call startWithoutExporting() of StatsProvider to use this feature in test codes. 
+
+Currently Pravega supports StatsD registry in Telegraf flavor; dimensional metrics data model (or metric tags) 
+is supported. Communication protocol is UDP.
+Direct InfluxDB connection is also supported.
+
 The reporter could be configured through MetricsConfig.
 ```java
 public class MetricsConfig extends ComponentConfig {
```
andreipaduroiu: Link to this class in the codebase.

documentation/src/docs/metrics.md
```diff
@@ -116,204 +134,169 @@ public final class ServiceStarter {
...
+    public static final String COMPONENT_CODE = "metrics";
     ...
 }
 ```
 
 # 4. Steps to add your own Metrics
-* Step 1. When start a segment store/controller service, start a Metrics service as a sub service. Reference above example in ServiceStarter.start()
+
 ```java
+// Step 1. When start a segment store/controller service, start a Metrics service as a sub service. Reference above example in ServiceStarter.start()
```
andreipaduroiu: When starting

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessorTest.java
```diff
@@ -239,7 +239,7 @@ public void testReadSegmentWithCancellationException() {
     @Test(timeout = 20000)
     public void testReadSegmentTruncated() {
         // Set up PravegaRequestProcessor instance to execute read segment request against
-        String streamSegmentName = "testReadSegment";
+        String streamSegmentName = "scope/stream/testReadSegment";
```
andreipaduroiu: Why did you change all of these names?

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -269,33 +233,24 @@ public static String joinWithDot(String element1, String element2) {
...
-    public static MetricKey metricKey(String metric, String typeSuffix, String... tags) {
+    public static MetricKey metricKey(String metric, String... tags) {
 
-        StringBuilder sb = new StringBuilder(metric);
-        if (tags == null || tags.length == 0) {  //no tags: append suffix to form cache & registry key
-            sb.append(typeSuffix);
-            String key = sb.toString();
-            return new MetricKey(key, key);
-        } else { //tags: append tag values and suffix to form cache key, original name is registry key
+        if (tags == null || tags.length == 0) {  //without tags: the original metric name is cache and registry key
```
andreipaduroiu: This comment seems like it's missing something

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
...
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.shared;
+
+import com.google.common.base.Preconditions;
+
+public class MetricsTags {
```
andreipaduroiu: `public final class`

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
...
+    // Metric Tag Names
+    public static final String TAG_CONTAINER = "container";
+    public static final String TAG_HOST = "host";
+    public static final String TAG_SCOPE = "scope";
+    public static final String TAG_STREAM = "stream";
+    public static final String TAG_SEGMENT = "segment";
+    public static final String TAG_TRANSACTION = "transaction";
+    public static final String TAG_EPOCH = "epoch";
+
+    private static final String TRANSACTION_DELIMITER = "#transaction.";
```
andreipaduroiu: Please take these out of this class and make use of methods in `StreamSegmentNameUtils`. All stream naming logic should reside in that class.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return tags;
+    }
+
+    /**
+     * Get base name of segment with the potential transaction delimiter removed.
+     * @param segmentQualifiedName fully qualified segment name.
+     * @return the base name of segment.
+     */
+    private static String getSegmentBaseName(String segmentQualifiedName) {
```
andreipaduroiu: This should go into `StreamSegmentNameUtils`

jkhalack:
Tested the branch by writing some data to different kinds of streams.
Looks like none of the `segment_write_bytes`, `segment_write_events`, or `segment_read_bytes` metrics are being reported (written to InfluxDB) on the streams with transactional writes (they **are** reported on the streams with non-transactional writes).

jkhalack:
Testing the on the stream with dynamic scaling, it looks like `segment_splits` and `segment_merges` metrics are not reported correctly.
These are the charts based on InfluxDB-stored metrics:
splits_merges_metrics
You can see that the number of segments fluctuates between 17 and 22, so the number of splits and merges per time interval (10 seconds here) should be in single digits. However, it is reported as 40+ splits and 20+ merges (when the scaling events do happen). Something is wrong with those counters.

To compare, these are the data for the same period of time, as reported by the controller `scaling-events` REST API:
split_merges_API
As you can see, there are not more than 5 splits or merges happening per any given minute.

fpj:
@jkhalack

> You can see that the number of segments fluctuates between 17 and 22, so the number of splits and merges per time interval (10 seconds here) should be in single digits. However, it is reported as 40+ splits and 20+ merges (when the scaling events do happen). Something is wrong with those counters.

You have reported this issue elsewhere, right? #1761 

I don't expect this PR to be solving that issue. The discussion around that issue should happen in the issue and corresponding PR, unless of course there is some breaking change that this particular PR is introducing. 

To narrow down the scope of the discussion, is this an issue created by the changes to add dimensionality?

jkhalack:
@fpj I'm frankly not sure.
I'm used to see the splits/merges count issues with multicontroller configuration, but with this branch I do see it with a single-controller setup. 
Probably more testing is needed on the master branch to see if it happens there too.

And I agree that it should be dealt with in a separate issue (either amend #1761 or create a new one for single-controller setup).

documentation/src/docs/metrics.md
kevinhan88: It seems the intention of the original author is to help readers read from beginning to end without the need to leave the doc, by adding(duplicating) some reference contents directly into the doc.
Yes this approach is opposite to java-doc style, where reference link is always enforced, and copy-pasting/duplication is prohibited.

@andreipaduroiu, are we sure we want to change the style drastically?

Meanwhile, I'm removing the less relevant lines from the doc so it only contains the "distilled" metrics operations.

documentation/src/docs/metrics.md
kevinhan88: Though I'm not the original author of the document, my understanding is that the interface is provided here so readers don't have to leave the doc (to get interface definition) when they reach the bullet points following the interface.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: Good catch! Yes this class has nothing to be extended.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: I already tried. Metric code is on the lowest level of the dependency relationship defined, it doesn't have access to those classes.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
kevinhan88: Explained above.

kevinhan88:
@jkhalack Yes it is a bug: metric name (without tags) is used to lookup cache, which caused "aggregation" effect for counters (as stream specific metrics share the common metric name). The bug is fixed.
Before this PR is merged, please note the latest change is inside branch "issue-3405-2nd-attempt" instead of branch "issue-3405".

kevinhan88:
> Could you add the `statsdHost` => `statsDHost` change to https://github.com/kevinhan88/pravega/blob/master/config/config.properties file, please?
> Same for `statsDHport` and `enableStatsDReporter`. Thanks.
> 
> Also, it would be great to update the metrics names in the documentation:
> https://github.com/kevinhan88/pravega/blob/master/documentation/src/docs/metrics.md

Fixed.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessorTest.java
kevinhan88: Segment name must be qualified containing at least stream.
The old unit test code works with unqualified name because the old code doesn't have to retrieve stream/segment information from the name.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: Fixed.

kevinhan88:
> Tested the branch by writing some data to different kinds of streams.
> Looks like none of the `segment_write_bytes`, `segment_write_events`, or `segment_read_bytes` metrics are being reported (written to InfluxDB) on the streams with transactional writes (they **are** reported on the streams with non-transactional writes).

@jkhalack Yes Julia, for those transnational segment operations, now we report them with its parent segment metrics only if the transaction is merged.

documentation/src/docs/metrics.md
kevinhan88: Done

documentation/src/docs/metrics.md
kevinhan88: Done

documentation/src/docs/metrics.md
kevinhan88: All irrelevant lines removed.

config/config.properties
```diff
@@ -201,14 +201,17 @@ metrics.enableStatistics=false
...
+#metrics.influxDBName=pravega
 
 # InfluxDB user name.
 #metrics.influxDBUserName=
 
 # InfluxDB password.
 #metrics.influxDBPassword=
 
+# InfluxDB retention policy (e.g. 2h, 52w).
+#metrics.influxDBRetention=
```
RaulGracia: Would it be interesting to set a safe default here?

documentation/src/docs/metrics.md
```diff
@@ -7,57 +7,50 @@ You may obtain a copy of the License at
...
-public interface StatsProvider {
-    void start();
-    void close();
-    StatsLogger createStatsLogger(String scope);
-    DynamicLogger createDynamicLogger();
-}
-```
-
-- start(): Initializes [MetricRegistry](http://metrics.dropwizard.io/3.1.0/manual/core/#metric-registries) and reporters for our Metrics service. 
+The starting point of Pravega Metric framework is the StatsProvider interface, it provides start and stop method for Metric service.
```
RaulGracia: `start and stop methods...`

documentation/src/docs/metrics.md
```diff
@@ -7,57 +7,50 @@ You may obtain a copy of the License at
...
+It also provides startWithoutExporting() for unit testing purpose which only store metrics in memory without exporting them
```
RaulGracia: `testing purposes, which only stores metrics...`

documentation/src/docs/metrics.md
```diff
@@ -7,57 +7,50 @@ You may obtain a copy of the License at
...
+to outside destinations.
```
RaulGracia: `Outside destination` -> External system?

documentation/src/docs/metrics.md
```diff
@@ -7,57 +7,50 @@ You may obtain a copy of the License at
...
-    StatsLogger createStatsLogger(String scope);
-    DynamicLogger createDynamicLogger();
-}
-```
-
-- start(): Initializes [MetricRegistry](http://metrics.dropwizard.io/3.1.0/manual/core/#metric-registries) and reporters for our Metrics service. 
+The starting point of Pravega Metric framework is the StatsProvider interface, it provides start and stop method for Metric service.
+It also provides startWithoutExporting() for unit testing purpose which only store metrics in memory without exporting them
+to outside destinations. 
+Currently we have support StatsD and InfluxDB registries.
```
RaulGracia: `we have support for...`

documentation/src/docs/metrics.md
```diff
@@ -7,57 +7,50 @@ You may obtain a copy of the License at
...
-public interface OpStatsLogger {
-    void reportSuccessEvent(Duration duration);
-    void reportFailEvent(Duration duration);
-    void reportSuccessValue(long value);
-    void reportFailValue(long value);
-    OpStatsData toOpStatsData();
-    void clear();
-}
-```
+OpStatsLogger provides complex statistics type of Metric, usually it is used in operations such as CreateSegment,
```
RaulGracia: By "complex statistics" you mean a "histogram", right? If so, you may want to use the concept of histogram that looks more concrete. You can say that if the user is interested on measuring the latency of operations (`CreateSegment`, `ReadSegment`,...) this is the metric type to choose.

documentation/src/docs/metrics.md
```diff
@@ -7,57 +7,50 @@ You may obtain a copy of the License at
...
-    void reportSuccessEvent(Duration duration);
-    void reportFailEvent(Duration duration);
-    void reportSuccessValue(long value);
-    void reportFailValue(long value);
-    OpStatsData toOpStatsData();
-    void clear();
-}
-```
+OpStatsLogger provides complex statistics type of Metric, usually it is used in operations such as CreateSegment, 
+ReadSegment, we could use it to record the number of operation, time/duration of each operation.
```
RaulGracia: `number of operations...`

documentation/src/docs/metrics.md
```diff
@@ -116,204 +101,158 @@ public final class ServiceStarter {
...
+    public void createStream(String scope, String streamName, int minNumSegments, Duration latency) {
+        DYNAMIC_LOGGER.reportGaugeValue(OPEN_TRANSACTIONS, 0, streamTags(scope, streamName));
+    }
     ...
  }
 ```
 
 ### 2.3 Example for Dynamic Meter
-This is an example from io.pravega.segmentstore.server.SegmentStoreMetrics. In this class, we report a Dynamic Meter which represents the segments created.
+This is an example from io.pravega.segmentstore.server.SegmentStoreMetrics. In this class, we report a Dynamic Meter
```
RaulGracia: Sometimes you refer to the classes in caps (e.h., `Dynamic Meter`) and other you don't. Please, look for a consistent naming.

documentation/src/docs/metrics.md
```diff
@@ -116,204 +101,158 @@ public final class ServiceStarter {
...
 }
 ```
  
-# 3. Metric reporter and Configurations
-Reporters are the way through which we export all the measurements being made by the metrics. We currently provide StatsD and CSV output. It is not difficult to add new output formats, such as JMX/SLF4J.
-CSV reporter will export each Metric out into one file. 
-StatsD reporter will export Metrics through UDP/TCP to a StatsD server.
+# 3. Metric Registries and Configurations
+With Micrometer, each meter registry is responsible for both storage and exporting of metrics objects. 
+In order to have a unified interface Micrometer provides CompositeMeterRegistry for application to interact with,
```
RaulGracia: Consider to use `this format` with the class names (`CompositeMeterRegistry`). We are trying to enforce a consistent style in the docs progressively.

documentation/src/docs/metrics.md
```diff
@@ -477,36 +416,36 @@ controller.stream.truncated_latency_ms
 ```
 controller.stream.created.Counter
 controller.stream.create_failed_global.Counter
```
RaulGracia: In some places the metrics end with `.Counter` and in others they don't. Please, let's keep a consistent way of referring to the metrics names.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
```diff
@@ -97,4 +103,16 @@ public DynamicLogger createDynamicLogger() {
         init();
         return new DynamicLoggerImpl(conf, metrics, new StatsLoggerImpl(getMetrics(), "pravega"));
     }
+
+    private String[] getHostTag() {
```
RaulGracia: Looks like this may be used in other places, for which perhaps can be exposed as an utility function in another place.

config/config.properties
kevinhan88: Here the retention policy name is arbitrary, similar to influxDBUserName, no default value for it.

documentation/src/docs/metrics.md
kevinhan88: fixed

documentation/src/docs/metrics.md
kevinhan88: fixed

documentation/src/docs/metrics.md
kevinhan88: fixed

documentation/src/docs/metrics.md
kevinhan88: fixed

documentation/src/docs/metrics.md
kevinhan88: fixed

documentation/src/docs/metrics.md
kevinhan88: fixed.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsProviderImpl.java
kevinhan88: fixed.

documentation/src/docs/metrics.md
Aparnarr: Accepted suggestion

documentation/src/docs/metrics.md
Aparnarr: Accepted suggestion

documentation/src/docs/metrics.md
Aparnarr: Accepted suggestion

documentation/src/docs/metrics.md
Aparnarr: Accepted Suggestion

documentation/src/docs/metrics.md
Aparnarr: Accepted suggestion

config/config.properties
jkhalack: The default in InfluxDB is no user (authentication is not enabled) and no password.

As to the retention policy, InfluxDB databases have `default` retention policies (name of the default retention policy is set per database, by InfluxDB), so if it not specified here, then the default one will be used.

documentation/src/docs/metrics.md
```diff
@@ -362,14 +301,14 @@ segmentstore.segment.write_latency_ms
 - Segment Store global and per-segment read/write metrics (counters):
 ```
 // Global counters
-segmentstore.segment.read_bytes_global.Counter
-segmentstore.segment.write_bytes_global.Counter
-segmentstore.segment.write_events_global.Counter
+segmentstore.segment.read_bytes_global
```
jkhalack: When you inspect the metrics in InfluxDB, the corresponding metric has the name `pravega_segmentstore_segmentstore_segment_read_bytes_global`, and the same for the rest of the metrics.
They all have prefixes like `pravega_segmentstore`, `pravega_controller`, or `pravega` as compared to the metrics manes reported here. 
The same for CSV metrics, with the difference that CSV might be not converting dots to underscores, as InfluxDB (or Micrometer?) does.

The prefixes are very confusing to whoever has to infer the actual metrics names from whatever is listed in this file, especially because there are at least three different prefixes.

Could you please add the explanation to how the prefixes are added (or update the metrics names to the be ones that actually reported)?

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -114,20 +111,20 @@ public Counter call() throws Exception {
...
         }
         counter.add(value);
-        countersCache.put(name, counter);
+        countersCache.put(keys.getCacheKey(), counter);
     }
 
     @Override
     public void freezeCounter(String name, String... tags) {
-        MetricsNames.MetricKey keys = metricKey(name, COUNTER_SUFFIX, tags);
+        MetricsNames.MetricKey keys = metricKey(name, tags);
```
RaulGracia: Same here and below.

documentation/src/docs/metrics.md
kevinhan88: There is a specific PR following this PR to update the document, so let's address all document issues there in order to not delay this PR further.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -114,20 +111,20 @@ public Counter call() throws Exception {
     @Override
     public void updateCounterValue(String name, long value, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
-        MetricsNames.MetricKey keys = metricKey(name, COUNTER_SUFFIX, tags);
```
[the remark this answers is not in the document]
kevinhan88: keys means the object contains both cache key and registry key.

documentation/src/docs/metrics.md
jkhalack: Sounds good.

RaulGracia:
@kevinhan88 Travis build is failing in a related test to this PR: `io.pravega.test.integration.ControllerMetricsTest > streamMetricsTest FAILED`

jkhalack:
Jenkins build still fails:
```
Execution failed for task ':segmentstore:server:test'
```

segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
```diff
@@ -148,20 +150,20 @@ private void report() {
          */
         private final OpStatsLogger processOperationsLatency;
         private final OpStatsLogger processOperationsBatchSize;
-        private final String operationLogSize;
+        private final String[] containerTag;
```
fpj: Really minor comment, this should be `containerTags` plural as it is an array of tags.

test/testcommon/src/main/java/io/pravega/test/common/TestUtils.java
```diff
@@ -41,7 +41,7 @@
      *
      * @return free port.
      */
-    public static int getAvailableListenPort() {
+    public synchronized static int getAvailableListenPort() {
```
fpj: This change seems to be gratuitous.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
kevinhan88: Yes, one tag is a "pair" of strings. Here containerTag and hostTag returns "one" pair of strings as "a" tag, while all others, such as streamTags, returns multiple pairs of strings as multiple tag"s".

So precisely the method might be called "containerTagPair", "streamTagPairs". The method name determines plural or single from "pair" perspective instead of String perspective, but I agree it's not straightforward sometimes to readers.

test/testcommon/src/main/java/io/pravega/test/common/TestUtils.java
kevinhan88: This method is to get a free port but without synchronization two threads may call at same time and get same port which eventually cause port collision. I experienced this problem once.

shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java
```diff
@@ -27,13 +27,14 @@
     public final static Property<Integer> DYNAMIC_CACHE_EVICTION_DURATION_MINUTES = Property.named("dynamicCacheEvictionDurationMinutes", 30);
     public final static Property<Integer> OUTPUT_FREQUENCY = Property.named("outputFrequencySeconds", 60);
     public final static Property<String> METRICS_PREFIX = Property.named("metricsPrefix", "pravega");
-    public final static Property<String> STATSD_HOST = Property.named("statsdHost", "localhost");
-    public final static Property<Integer> STATSD_PORT = Property.named("statsdPort", 8125);
+    public final static Property<String> STATSD_HOST = Property.named("statsDHost", "localhost");
```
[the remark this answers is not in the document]
kevinhan88: In the previous PR I introduced an change (from "statsD" to "statsd") to address one code review comments (to make all "statsD" consistent), but it turns out that change is not backwards compatible. 
This change is to revert that breaking change, hence this is not a breaking change.

kevinhan88:
@fpj to the comments over StreamSegmentNameUtils.java access - seems "shared.metrics" doesn't have access to "shared.protocol" where SSNU is located.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
```diff
@@ -197,11 +199,11 @@ public void memoryCommit(int metadataUpdateTxnCount, Duration elapsed) {
         }
 
         public void operationLogTruncate(int count) {
-            DYNAMIC_LOGGER.incCounterValue(this.operationLogSize, -count);
+            DYNAMIC_LOGGER.incCounterValue(MetricsNames.OPERATION_LOG_SIZE, -count, this.containerTag);
```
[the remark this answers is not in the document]
kevinhan88: No, we don't.

test/testcommon/src/main/java/io/pravega/test/common/TestUtils.java
kevinhan88: And removed this synchronization to a separate issue/PR, per the conversation with Flavio.

PR #3509, opened 2019-03-22
Issue 3478: Reduce controller's memory footprint

### LINKED ISSUE #3478: Contoller memory footprint improvement
**Problem description**
Controller's metadata store caches byte[] from the underlying store (zookeeper). 
Everytime the entry is read from the cache, it is deserialized again to convert it to a data object. 
For non cached reads, creation of new objects is inevitable. However, for cached reads, we could avoid duplicate deserialization and object creation by caching the deserialized object instead of byte[]. 

Secondly, cache has an expiration time of 10 minutes, which means objects will be GC'd only after 10 minutes and can continue to put memory pressure on controller process. 
So reduce cache period. 

Thirdly, use a shared cache across multiple streams. Currently each stream has its own cache. If there are large number of streams, this could lead to huge memory pressure on controller process. 

**Problem location**
StreamMetadataStore 

Cache

shiveshr:
**Change log description**  
Implement several improvements to controller's memory usage - avoid redundant objects, avoid redundant copies, avoid multiple deserializations.

**Purpose of the change**  
Fixes #3478 
Also includes changes from #3496 #3512 #3516. So they should be merged before this. 

**What the code does**  
There are several improvements done as part of this change set:

1. One of the big culprits was the ZkHostStore which actually read the serialized host-container map from store everytime we needed to find segment to host mapping and then deserialized it. 
this was highly wasteful. Changed it to use NodeCacheListener. 
PS: We are only fixing using nodeCache in this PR. The fix for the serialization (using our custom serialization) with node cache is done as part of #3511 in PR #3512 
2. Caching:
Other major improvement, esp relevant for streams with very large metadata, is to cache the java objects rather than byte arrays that we fetch from the store. 
This has the advantage that we dont have to deserialize objects everytime we read it from the cache (which can happen multiple times). 
Saves on both compute and memory (because we avoid creation of duplicate object). The downside is, cached objects can stay referenced for longer even when they are not needed. But we have realized that upside weighs more than downside and has bigger perf implication. 
In order to mitigate the cached object memory footprint, I have also reduced cache duration from 10 minutes to 2 minutes. Objects will be evicted from the cache 2 minutes after last access. 

So basically the contract that store class had with its store-specific implementation was to deliver a wrapper `Data` object that had byte[] and version in it. With this change we have removed the `Data` class. instead we have changed that contract to use `VersionedMetadata` (which has generic {Object+version}). This applies to cache as well. 
So `Data.java` is deleted and `PersistentStreamBase.java` and its derived classes (`ZkStream.java` and `InMemoryStream.java` now have contract using typed `VersionedMetadata<MetadataType>` params. 

3. `SegmentHelper`:
`SegmentHelper` used to create a new connection with a new replyProcessor every time it contacted segmentStore. It had to do this because our library ties in connection with a reply processor. 
This means we created a lot of connection objects unnecessarily. With this change we have introduced a new Connection manager class. This is not a connection pooling (basically there is no multiplexing on same connection). Its reuse connections. 

This class simply tries to reuse available connections once previous processing on it is complete. For this is creates a new passthru "reusable reply processor" and `SegmentHelper` returns connections back to it after each processing. 
`ConnectionManager` class: This class opens new connections and reclaims connections when users return the connection back to it after use. The connection is not destroyed immediately, but kept around for reuse.  When a subsequent request for a new connection comes, an existing connection (if available) a new replyProcessor is associated with the connection. This connection manager does not do Connection Pooling. It simply gives out connections, and if there are no available connections, it creates a new one. It also imposes a bound on number of concurrent connections, so as to ensure that we do not overwhelm the system on resource front. 

`ConnectionManager` keeps a cache of connectionpools to different segment stores. 
If a segment store has not been contacted with in a while, this entry is evicted from the cache. 
So there could possibly be multiple connectionpools for the same segmentstore node present because of cache eviction. But the evicted connectionpools is marked for shutdown and as it reclaims connections it will close them. Also, only transiently, a caller may have access to an evicting connectionManager and may request for new connections. This is harmless as the object will return a valid connection to the caller. But soon after eviction, segmentHelper would have no reference to this object to be able to request newer connections. And since its marked for shutdown, it will eventually shutdown. 

3.1 segmenthelper also has additional buffer fixes - byteBuf used in table apis needs to be released as it is reference counted buffer and causes leaks otherwise. 
3.2 replyProcessor.processingFailure can also receive ConnectionFailed rather than explicit ConnectionDropped. ConnectionFailed needs to be retryable. So wrapping all processing failures under WireCommandFailureException. 

4. StreamSegmentRecord object was being transformed into Segment object and vice versa which was unnecessary and we have now exposed `StreamSegmentRecord` to store apis instead of `Segment` object. This saves on unnecessary creation of duplicate objects. 

5. `copyCollections` flag on Metadata objects with collections. All objects created via deserialization will no longer make another "immutable" copy of the collection. For very large objects, we were creating a lot of copies of collection objects. This results in a lot of `Node` objects being created. 
Refer to class in package `io.pravega.controller.store.stream.records`.

6. Few minor improvements: 
PersistentStreamBase.isStreamCutValid method is optimized to not make too many getEpochRecord calls to cache. 
AbstractStreamMetadataStore does not compute splits and merges after every scale anymore as that is a very costly computation. 

Analysis of improvements:

Through profiling on controller process, i found that the most substantial allocations were for char[] and byte[] (30-40% each). These were followed by HashMap.Node, Long, List.Node and CompletableFuture objects. These independently contributed to footprint of 5-10% which would shrink after each GC cycle. 

Improvements to reduce byte[] allocation: Bulk of byte[] allocation is originating from serialization and deserialization of our metadata objects and sending these over the wire. Since we reduced the unnecessary serialization and deserialization, bulk of byte[] allocations have also come down. But beyond a point, it is not possible to shrink this down without looking to reduce amount of information in our objects. 
Most popularly serialized and deserialized object is `StreamSegmentRecord`. This is a denormalized inner record present in multiple record types like epochRecord, truncation record etc. 
With caching of objects we have reduced the footprint of byte[] reasonably. For same workload, from an earlier 150-250 mb range, the footprint now ranges between 40-100mbs.

The char[] allocations mostly come from logger (~75-80%) and micrometer (5-10%) and CommandDecoder. 
Most of it is generated in log.error messages. Most of the error messages come in because we hammer segment store with 10k create segment requests and several connections fail and are retried. Similarly we issue that many concurrent truncate and delete requests. 
With segmentStore connection pooling changes we have ensured that we dont face a lot of connection failures resulting in such large stacks being printed and taking up char arrays. 
This has given marked improvement to overall memory footprint. char[] allocations, esp when we hammer segment store for very large streams.  

There are `TaskStartException` that can be logged quite aggressively too. This is also an unnecessary and avoidable exception. This is partially fixed in KVS branch. I will take that up as a separate issue and clean up the possibility of throwing start exceptions. 

Future improvements:
We can look to reduce number of allocations for `StreamSegmentRecord` as this object for a segment is duplicated in each epoch record and a clone object gets created every time a new epoch record (or some of the other records) get created. But this will not be very easy to do. I will take a stab at it in a different PR. 

**How to verify it**  
All tests should continue to pass. 
New test added for connection management to segment store. 

Uncommented truncate operation in LargeSegmentsSystem test as that will also be optimal now.


fpj:
@shivesh you have some interesting analysis of memory allocation in the description. do you have a sense of how much you have been able to improve with these changes? what have you actually used to profile the controller?

shiveshr:
> do you have a sense of how much you have been able to improve with these changes? what have you actually used to profile the controller?
I used JProfiler. Ran segmentStore (with inmemory store), controller processes and zookeeper on my VM.


> I'm concerned about the length of the PR given that making these changes can easily break Pravega if we don't pay close attention to them. I'm wondering if you'd be willing to break this PR into multiple changes.
I agree this is not ideal and i can break it into several PRs.
In fact one of the changes in this PR is already taken out into a separate PR (#3512). [its present in this PR too but once that branch is merged, this PR will shrink by 5-10 files). 

Similarly, I can easily take out segment connection manager changes from this as it is pretty contained in a single class with a single test class.

shiveshr:
breaking it into different PRs

PR #3512, opened 2019-03-22
Issue 3511: Use versioned serialization for cluster container monitor

### LINKED ISSUE #3511: Host to container mapping uses java native serialization
**Problem description**
Host to container mapping is stored in zookeeper by serializing `Map<Host, Set<Integer>>`. 
All of this uses java's native serialization which is not versioned. 

**Problem location**
Host.java, ZkHostStore.java, ZKSegmentContainerMonitor.java

Use pravega's versioned serialization instead of java's native serialization.

shiveshr:

**Change log description**  
Host to container mapping is stored in zookeeper by serializing Map<Host, Set<Integer>>.
All of this uses java's native serialization which is not versioned.

Also, in ZkHostStore use NodeCache instead of querying zookeeper each time for resolving container to host mapping. 

**Purpose of the change**  
Fixes #3511 

**What the code does**  
Creates a new Class called HostToContainerMap which uses our versioned serialization. 
Also changes Host.java to follow versioned serialization scheme.

**How to verify it**  
SerializationTest added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/host/ZKHostStore.java
+++ b/controller/src/main/java/io/pravega/controller/store/host/ZKHostStore.java
@@ -11,19 +11,19 @@ package io.pravega.controller.store.host;
 
 import com.google.common.base.Preconditions;
 import io.pravega.common.cluster.Host;
+import io.pravega.common.cluster.HostContainerMap;
 import io.pravega.controller.util.ZKUtils;
 import io.pravega.shared.segment.SegmentToContainerMapper;
-import java.io.Serializable;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
 import io.pravega.shared.segment.StreamSegmentNameUtils;
+import lombok.SneakyThrows;
 import lombok.Synchronized;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.SerializationUtils;
 import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.recipes.cache.NodeCache;
 import org.apache.curator.utils.ZKPaths;
 
 /**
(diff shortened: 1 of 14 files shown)


shared/cluster/src/main/java/io/pravega/common/cluster/ContainerSet.java
```diff
@@ -0,0 +1,63 @@
...
+        protected void declareVersions() {
+            version(0).revision(0, this::write00, this::read00);
+        }
+
+        private void read00(RevisionDataInput revisionDataInput, ContainerSetBuilder ContainerSetBuilder) throws IOException {
+            ContainerSetBuilder
+                    .containerSet(revisionDataInput.readCollection(DataInput::readInt, HashSet::new));
+        }
+
+        private void write00(ContainerSet ContainerSet, RevisionDataOutput revisionDataOutput) throws IOException {
```
andreipaduroiu: please use proper Java casing for variables: `ContainerSet` -> `containerSet`

controller/src/main/java/io/pravega/controller/store/host/ZKHostStore.java
```diff
@@ -53,15 +55,19 @@
...
         zkClient = client;
         zkPath = ZKPaths.makePath("cluster", "segmentContainerHostMapping");
+        hostContainerMapNode = new NodeCache(zkClient, zkPath);
+
         segmentMapper = new SegmentToContainerMapper(containerCount);
     }
 
     //Ensure required zk node is present in zookeeper.
     @Synchronized
+    @SneakyThrows(Exception.class)
```
andreipaduroiu: Can we narrow this down to `IOException.class`? I can't tell from this view where the exception might be coming from.

shiveshr: Curator has signature for NodeCache.start() as throws Exception

PR #3516, opened 2019-03-24
Issue 3515: ControllerFailover test case is incorrect and stale 

### LINKED ISSUE #3515: ControllerFailover Systemtest is incorrect and stale
**Problem description**
Controller failover test is stale. It was written at a time when scale waited for transactions on old epochs which was no longer the case with rolling transactions. 
The test creates a transaction and then initiates a scale and expects the scale to stall. 
it checks once if the scale is complete or not. And then goes on to invoke a controller failover. 

So there are two problems here -
1. assumption that the scale would wait on transaction. 
This had been passing merely by luck as when we check `isScaleDone` in the test, the scale is not complete. But there is nothing preventing the scale to complete.
2. the test starts with 2 controller instances. There is nothing that tells this test which controller is running the scale. So reducing number of controller instances has no guarantee that the scale processing for the said stream actually failsover. 

**Problem location**
ControllerFailover system test

Remove this test.

shiveshr:

**Change log description**  
* Refactors the test case to validate that scaling happens after stopping and restarting the controller.

**Purpose of the change**  
Fixes #3515 

**What the code does**  
 Controller failover test is stale. It was written at a time when scale waited for transactions on old epochs which was no longer the case with rolling transactions.
The test creates a transaction and then initiates a scale and expects the scale to stall.
it checks once if the scale is complete or not. And then goes on to invoke a controller failover.

So there are two problems here -
1.     assumption that the scale would wait on transaction.
    This had been passing merely by luck as when we check isScaleDone in the test, the scale is not complete. But there is nothing preventing the scale to complete.
2.    the test starts with 2 controller instances. There is nothing that tells this test which controller is running the scale. So reducing number of controller instances has no guarantee that the scale processing for the said stream actually fails over.

We have removed the incorrect scale check. To make the test meaningful (but not deterministic, we stop a controller and restart it. We verify that the scale operation is able to complete despite the restart.

**How to verify it**  
System test should consistently pass

### DIFF
--- a/test/system/src/test/java/io/pravega/test/system/ControllerFailoverTest.java
+++ b/test/system/src/test/java/io/pravega/test/system/ControllerFailoverTest.java
@@ -12,14 +12,13 @@ package io.pravega.test.system;
 import io.pravega.client.ClientConfig;
 import io.pravega.client.stream.ScalingPolicy;
 import io.pravega.client.stream.StreamConfiguration;
-import io.pravega.client.stream.Transaction;
 import io.pravega.client.stream.impl.Controller;
 import io.pravega.client.stream.impl.ControllerImpl;
 import io.pravega.client.stream.impl.ControllerImplConfig;
 import io.pravega.client.stream.impl.StreamImpl;
 import io.pravega.client.stream.impl.StreamSegments;
-import io.pravega.client.stream.impl.TxnSegments;
 import io.pravega.common.concurrent.Futures;
+import io.pravega.shared.segment.StreamSegmentNameUtils;
 import io.pravega.test.system.framework.Environment;
 import io.pravega.test.system.framework.SystemTestRunner;
 import io.pravega.test.system.framework.Utils;
@@ -32,7 +31,6 @@ import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import lombok.Cleanup;
 import lombok.extern.slf4j.Slf4j;
@@ -63,7 +61,7 @@ public class ControllerFailoverTest extends AbstractSystemTest {
     public static void initialize() throws MarathonException, ExecutionException {
         URI zkUri = startZookeeperInstance();
         startBookkeeperInstances(zkUri);
-        URI controllerUri = startPravegaControllerInstances(zkUri, 2);
+        URI controllerUri = startPravegaControllerInstances(zkUri, 1);
         ensureSegmentStoreRunning(zkUri, controllerUri);
     }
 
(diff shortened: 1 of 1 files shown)


shiveshr:
the problem is once scale starts on a controller, there is no guarantee that by the time we bring it down the scale would not have completed. so i dont know what or how we can make this test more deterministic..

fpj:
The name controller failover is a bit misleading as it is attempting to validate a very specific scenario. Questions:

- Are we testing controller failure scenarios in other tests?
- Would it make sense to change this test case to test failover more generally?
- If there is nothing we can do to make this test meaningful, then should we remove it?

shiveshr:
I believe this test was created before event processors.. Where scale was done on task frameworks. So you could send request to a specific controller and it would have the scale processing started. And it would stall because of presence of transaction.. 
So it would test the failover specifically by killing controller1 and seeing that controller 2 picks and completes the work after transaction based latch was removed (via txn abort). 

We cant do this kind of thing deterministically anymore as we dont know from the test which controller has a specific work. If we had a single controller then the determinism is there.. But we dont know if the work is incomplete when we shut it down and is completed only after failover completes.. thats because workflows dont depend on each other and any workflow started can be finished before we are able to shutdown. 

What we can test instead is following:
1. run workflow against 1st controller. 
2. fail controller.. bring up new controller.. 
3. run a new workflow against 2nd controller.. 
This thing is fairly regularly tested in all our other failover related tests where we bring down and bring up instances.

fpj:
> This thing is fairly regularly tested in all our other failover related tests where we bring down and bring up instances.

I think you are suggesting to remove the test case, and it might be fine. can you point me to the other test cases that are already exercising the functionality you are referring to?

shiveshr:
All tests with "failover" in their name typically start with 3 controller instances and scale it down to 1 and repeats this several times. Since kubernetes can take down any instance when scaling down a service instance, the basic failover of event processor happens with a high degree of probability. 

As i said earlier, we could change this test to have 1 controller instance.. perform some operation there.. then stop and start controller and have some operation performed via second controller instance.. 
it will not exactly test a failover of a partially completed workflow but it will deterministically test that segment ownership for internal streams moves from one controller to the next.

fpj:
> we could change this test to have 1 controller instance.. perform some operation there.. then stop and start controller and have some operation performed via second controller instance..
it will not exactly test a failover of a partially completed workflow but it will deterministically test that segment ownership for internal streams moves from one controller to the next.

ok, let's go with that.

test/system/src/test/java/io/pravega/test/system/ControllerFailoverTest.java
```diff
@@ -159,12 +155,12 @@ public void failoverTest() throws InterruptedException, ExecutionException {
...
         }
 
-        // Scale operation should now complete on the second controller instance.
         // Note: if scale does not complete within desired time, test will timeout. 
-        while (!scaleStatus) {
-            scaleStatus = controller2.checkScaleStatus(stream1, 0).join();
-            Thread.sleep(30000);
-        }
+        AtomicBoolean scaleDone = new AtomicBoolean(false);
+        Futures.loop(() -> !scaleDone.get(),
```
andreipaduroiu: This is no different than a simple `while` busy-wait-loop. You can just call `join` on `controller2.checkScaleStatus` and decide from there. You'll be blocking this thread regardless, but the code will look a lot simpler.

shiveshr: I know. But we have had comments asking not to use Thread.sleep in the past. So have been using this workaround in tests where we want to check something in a loop with sleeps.

PR #3518, opened 2019-03-24
Issue 3513: Regression: AbstractRequestProcessor should complete future in withCompletion on all paths

### LINKED ISSUE #3513: Regression: StreamsAndScopesManagementTest timedout
**Problem description**
StreamsAndScopesManagementTest is timing out while trying to perform any operation on a recreated stream. This is happening because of a regression caused by a recent change #3444 .
The reason is, we have added a `hasStarted` method in the withCompletion method of AbstractProcessor. If hasStarted throws any exception, the future is not completed, which causes the processing to be stuck. 

**Problem location**
AbstractRequestProcessor.java

Fix the issue.

shiveshr:
Signed-off-by: Shivesh Ranjan shivesh.ranjan@gmail.com

**Change log description**
Fixes regression caused by a recent change #3444.
This regression causes system test failure for ScopesAndStreamsManagementSystemTest

**Purpose of the change**
Fixes #3513

**What the code does**
We added a `hasStarted` call in `withCompletion` method of `AbstractProcessor`. If `hasStarted` fails exceptionally, the result future is never completed, which causes the processing for a stream to be stuck.

Also, fixed a cascading theoretical issue where if any synchronous termination occurs rather than future with asynchronous failure then serializedRequestProcessor will fall into same issue as above.

**How to verify it**
Unit tests added for both processors

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
+++ b/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
@@ -156,6 +156,9 @@ public abstract class AbstractRequestProcessor<T extends ControllerEvent> extend
                                 StoreException.create(StoreException.Type.OPERATION_NOT_ALLOWED, "Postponed "
                                         + event + " so that waiting processor" + waitingRequestProcessor + " can work. "));
                     }
+                }).exceptionally(e -> {
+                    resultFuture.completeExceptionally(e);
+                    return null;
                 });
 
         return resultFuture;
--- a/controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
+++ b/controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
@@ -276,8 +276,8 @@ public class SerializedRequestHandlerTest extends ThreadPooledTestSuite {
         // signal 2nd processing to complete
         signal2.complete(null);
         
-        // we should be able to process addiional events. 
-        AssertExtensions.assertFutureThrows("Second processing should should have failed in procesEvent method with synchronous exception",
+        // we should be able to process additional events. 
+        AssertExtensions.assertFutureThrows("Second processing should have failed in procesEvent method with synchronous exception",
                 future2, e -> Exceptions.unwrap(e) instanceof RuntimeException && Exceptions.unwrap(e).getMessage().equals("2"));
         
         signal3.complete(null);
--- a/controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
+++ b/controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
@@ -31,7 +31,11 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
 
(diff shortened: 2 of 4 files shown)


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -156,6 +156,9 @@ public String getProcessorName() {
                                 StoreException.create(StoreException.Type.OPERATION_NOT_ALLOWED, "Postponed "
                                         + event + " so that waiting processor" + waitingRequestProcessor + " can work. "));
                     }
+                }).exceptionally(e -> {
+                    resultFuture.completeExceptionally(e);
```
fpj: Isn't it fine to return the future as we have completed exceptionally?

controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
```diff
@@ -226,6 +229,62 @@ public boolean toPostpone(TestEvent event, long pickupTime, Throwable exception)
...
+        queue = throwingRequestHandler.getEventQueueForKey(event.getKey());
+        assertFalse(queue.isEmpty());
+        assertEquals(queue.size(), 1);
+        assertEquals(queue.get(0).getKey().eventId, "3");
+
+        assertFalse(wait3.isDone());
+        // signal 2nd processing to complete
+        signal2.complete(null);
+        
+        // we should be able to process addiional events.
```
fpj: Typo in `additional`.

controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
```diff
@@ -226,6 +229,62 @@ public boolean toPostpone(TestEvent event, long pickupTime, Throwable exception)
...
+        assertFalse(queue.isEmpty());
+        assertEquals(queue.size(), 1);
+        assertEquals(queue.get(0).getKey().eventId, "3");
+
+        assertFalse(wait3.isDone());
+        // signal 2nd processing to complete
+        signal2.complete(null);
+        
+        // we should be able to process addiional events. 
+        AssertExtensions.assertFutureThrows("Second processing should should have failed in procesEvent method with synchronous exception",
```
fpj: Duplicate `should`.

controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
```diff
@@ -226,6 +229,62 @@ public boolean toPostpone(TestEvent event, long pickupTime, Throwable exception)
         stop.set(true);
     }
 
+    @Test(timeout = 10000)
+    public void testThrowsSynchronousException() throws InterruptedException, ExecutionException {
```
fpj: I don't understand the synchronous exception issue that we are supposedly fixing in this PR and testing here. Could you please elaborate?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
shiveshr: the future called `resultFuture` is returned to the caller already. we are just completing it here.

controller/src/test/java/io/pravega/controller/eventProcessor/impl/SerializedRequestHandlerTest.java
shiveshr: Lets say i call a method that returns a future.. and that future could pass or fail. 
But if the method throws an exception rather than return a future that fails. For example, a precondition check in the method fails before it creates the exception. 

Example:
```
CompletableFuture<Void> decrementByOne(int input) {
if (input <= 0) throw new RuntimeException("value should be greater than 1");
return new CompletableFuture.completedFuture(input - 1);
}

// following will get a future back
val x = incrementByOne(100);

// following will fail synchronously and 
val x = incrementByOne(-1);
```

PR #3525, opened 2019-03-26
Issue 3478: Memory footprint improvement: Segment helper connection management

shiveshr:

**Change log description**  
Manages connections to segment store from controller. 
Also, since connectionFactory was getting refactored to take in as constructor param, also refactored host store to take in as constructor param. 

**Purpose of the change**  
Fixes one part of 3478

**What the code does**  
Contains perf improvements to segmenthelper class. This has two improvements - 
1. Connection Management
2. segment helper table api fixes: ByteBuf release + table URI
3. Refactor segmentHelper to take hoststore and connectionfactory as constructor params. 
4. So HostStore and ConnectionFactory dont need to be passed to streamMetadataTask and streamTransactionMetadataTask
5. handle connection failed received in processing failure to fail the connection and send wireCommandFailed response.

Details: Focus on ResourcePool.javam, SegmentStoreConnectionManager.java and SegmentHelper.java. Everything else in this change set is param refactoring. 

ConnectionManagement:
SegmentHelper used to create a new connection with a new replyProcessor every time it contacted segmentStore. It had to do this because our library ties in connection with a reply processor.
    This means we created a lot of connection objects unnecessarily. With this change we have introduced a new Connection manager class. This is not a connection pooling (basically there is no multiplexing on same connection). Its reuse connections.

This class simply tries to reuse available connections once previous processing on it is complete. For this is creates a new passthru "reusable reply processor" and SegmentHelper returns connections back to it after each processing.
ConnectionManager class: It is backed by new ResourcePool class. This class opens new connections and reclaims connections when users return the connection back to it after use. The connection is not destroyed immediately, but kept around for reuse. When a subsequent request for a new connection comes, an existing connection (if available) a new replyProcessor is associated with the connection. This connection manager does not do Connection Pooling. It simply gives out connections, and if there are no available connections, it creates a new one. It also imposes a bound on number of concurrent connections, so as to ensure that we do not overwhelm the system on resource front.

ConnectionManager keeps a cache of connectionpools to different segment stores.
If a segment store has not been contacted with in a while, this entry is evicted from the cache.
So there could possibly be multiple connectionpools for the same segmentstore node present because of cache eviction. But the evicted connectionpools is marked for shutdown and as it reclaims connections it will close them. Also, only transiently, a caller may have access to an evicting connectionpool and may request for new connections. This is harmless as the object will return a valid connection to the caller. But soon after eviction, connection manager would have no reference to this object to be able to request newer connections. And since its marked for shutdown, it will eventually shutdown.
Resource pool is a new utility class that is introduced to maintain a pool of generic type of reusable resources. It returns a resource wrapped under a ClosableResource object. Once close on such a resource is called, the resource is returned to the resource pool. 
The pool maintains a max concurrent resource creation and max idle resource set and tries to fulfil new requests from the idle set or returned resources before attempting to create a new resource. 

3.1 segmenthelper also has additional buffer fixes - byteBuf used in table apis needs to be released as it is reference counted buffer and causes leaks otherwise.
3.2 replyProcessor. In case connection is closed after manager gives it out, the reply processor will receive processingFailure - ConnectionFailedException. ConnectionFailed needs to be retryable. So wrapping all processing failures under WireCommandFailureException.
And before returning a connection to the pool, the caller lets it know that the connection has failed explicitly so as to avoid giving out failed connections that will be unnecessarily retried. 
3.3 removing "segment number `0`" from uri queries for table segments. 
The request uses table segment name while uri lookup includes segment number 0 which is incorrect. 

**How to verify it**  
Unit tests added/updated.


controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1236,28 +1225,43 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
             }
         });
         resultFuture.whenComplete((result, e) -> {
-            connectionFuture.thenAccept(ClientConnection::close);
+            // when processing completes, return the connection back to connection manager asynchronously.
+            // Note: If result future is complete, connectionFuture is definitely complete. 
+            if (resultFuture.isCompletedExceptionally()) {
+                resultFuture.exceptionally(ex -> {
+                    Throwable unwrap = Exceptions.unwrap(ex);
+                    if (unwrap instanceof WireCommandFailedException &&
```
tkaitchuck: This strikes me a somewhat fragile. Could we make this a whitelist instead of a blacklist?

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,653 @@
...
+     * waiting requests, this class tries to maintain an available connection pool of predetermined size. If more number of 
+     * connections than max available size is returned to it, the classes closes those connections to free up resources.  
+     *
+     * It is important to note that this is not a connection pooling class. 
+     * It simply reuses already created connections to send additional commands over it. It doesnt multiplex commands on the 
+     * same connection concurrently.  
+     * As users finish their processing, they should return the connection back to this class. 
+     *
+     * The connectionManager can be shutdown as well. However, the shutdown trigger does not prevent from callers to attempt 
+     * to create new connections and new connections will be served. Shutdown ensures that
```
tkaitchuck: This doc ends mid sentence

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,653 @@
...
+     * The connectionManager can be shutdown as well. However, the shutdown trigger does not prevent from callers to attempt 
+     * to create new connections and new connections will be served. Shutdown ensures that 
+     */
+    static class SegmentStoreConnectionPool {
+        private static final int MAX_CONCURRENT_CONNECTIONS = 500;
+        private static final int MAX_AVAILABLE_CONNECTIONS = 100;
+        private final PravegaNodeUri uri;
+        private final Object lock = new Object();
+        @GuardedBy("lock")
+        private final ArrayDeque<ConnectionObject> availableConnections;
```
tkaitchuck: This class seems very elaborate given that it's not doing true pooling which is going to be done by another component underneath it. Could we simplify things by instead of having a pool having a 'reusableConnection' which we just return to the cache directly when done?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: i dont quite follow the comment. ?.
The check here is:
if future completed exceptionally with `connectionFailed` or `connectionDropped` then we want to inform the pool about the failure before returning the connection back to it.
else we return the connection as is because the connection is still valid.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
shiveshr: yes, this class is slightly more complicated. 
It ensures 2 things -
1. limit the max concurrent connections created to a given segment store instance
2. once connections are returned to the pool, keep a ceiling on number of available connections. 

but i had a discussion with @shrids and with the multiplexing of requests over single connection, i dont need to maintain any connection logic. 
the connection factory should automatically make use of the same connection to send multiple requests. 
So once #3531 is merged, it will simplify things quite a bit and this class can be discarded. 
So let me wait on that change and once its merged, i will update this PR.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,653 @@
...
+                                log.debug("Returned connection object is discarded as available list is full");
+                                connectionCount--;
+                                toClose = true;
+                            }
+                        }
+                    }
+                }
+
+                if (toClose) {
+                    if (connectionListener != null) {
```
andreipaduroiu: Why would `connectionListener` be null? Don't you pass it in through the constructor?

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,653 @@
...
+
+            if (waiting != null) {
+                log.debug("Creating new connection for {}", uri);
+                ReusableReplyProcessor rp = new ReusableReplyProcessor();
+                rp.initialize(waiting.getReplyProcessor());
+                Futures.completeAfter(
+                        () -> clientCF.establishConnection(uri, rp)
+                                      .thenApply(connection -> new ConnectionObject(connection, rp))
+                                      .whenComplete((r, e) -> {
+                                          if (connectionListener != null) {
```
andreipaduroiu: What if an error occurred? Do you still want to notify of a new connection?

Also, `thenApply` and `whenComplete` can be merged together.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,653 @@
...
+        }
+
+        // unset reply processor
+        void uninitialize() {
+            replyProcessor.set(null);
+        }
+
+        @Override
+        public void hello(WireCommands.Hello hello) {
+            ReplyProcessor rp = replyProcessor.get();
```
andreipaduroiu: Can you just write a method that does this pattern instead of rewriting the same code over and over again?

```
    private <T> void execute(BiConsumer<ReplyProcessor,T> toInvoke, T arg){
        ReplyProcessor rp = replyProcessor.get();
        if (rp != null) {
            toInvoke.accept(rp, arg);
        }
    }
```

then you call it like this:
```
        execute(ReplyProcessor::hello, hello);
```

controller/src/test/java/io/pravega/controller/timeout/TimeoutServiceTest.java
```diff
@@ -107,18 +107,24 @@ public TimeoutServiceTest() throws Exception {
         TaskMetadataStore taskMetadataStore = TaskStoreFactory.createStore(storeClient, executor);
 
         ConnectionFactoryImpl connectionFactory = new ConnectionFactoryImpl(ClientConfig.builder().build());
-        streamMetadataTasks = new StreamMetadataTasks(streamStore, StreamStoreFactory.createInMemoryBucketStore(), hostStore, taskMetadataStore,
-                new SegmentHelper(), executor, hostId, connectionFactory, AuthHelper.getDisabledAuthHelper(), requestTracker);
+        streamMetadataTasks = new StreamMetadataTasks(streamStore, StreamStoreFactory.createInMemoryBucketStore(), taskMetadataStore,
+                new SegmentHelper(connectionFactory, hostStore), executor, hostId, AuthHelper.getDisabledAuthHelper(), requestTracker);
         streamTransactionMetadataTasks = new StreamTransactionMetadataTasks(streamStore, hostStore,
                 SegmentHelperMock.getSegmentHelperMock(), executor, hostId, TimeoutServiceConfig.defaultConfig(),
+<<<<<<< HEAD
```
andreipaduroiu: Why is this showing up here? Did you forget to merge conflicts?

Below too...

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,653 @@
...
+ * Connection Manager class that maintains a cache of connection pools to connection to different segment stores. 
+ * Users of connection manager class can request for a connection pool and then request connections on the pool and return 
+ * the connections when done. 
+ * This maintains a cache of connection pools to segment stores. Any segmentstore for which no new connection request
+ * comes for a while, its pool is evicted from the cache which triggers a shutdown on the pool. 
+ * If newer requests are received for the said segmentstore, a new pool gets created. However, any callers using the existing 
+ * pool can continue to do so. The pool in shutdown mode simply drains all available connections. And when it has no references left
+ * it can be garbage collected. 
+ */
+class SegmentStoreConnectionManager {
```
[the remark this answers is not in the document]
shiveshr: Cache eviction leads to shutdown being invoked on the pool object. 
cache cleanup will also result in all entries being evicted and shutdown invoked. 
shutdown cleans up all outstanding connections. So resources are all freed up. 

Autoclosable's close is invoked only if we have try(object){} kind of block. 
We can make this closeable and in turn make segment helper closable which invokes its close method.
And in the close method we can call cache.invalidateAll and cache.cleanup. 
I will make that change.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,653 @@
...
+         *
+         * @param processor reply processor to use
+         * @return A completableFuture which when completed will have the connection object that the caller can use to
+         * communicate with segment store and receive response on the supplied reply processor.
+         */
+        CompletableFuture<ConnectionObject> getConnection(ReplyProcessor processor) {
+            CompletableFuture<ConnectionObject> connectionFuture;
+            boolean tryCreateNewConnection = false;
+            synchronized (lock) {
+                ConnectionObject obj = availableConnections.poll();
```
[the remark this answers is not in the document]
shiveshr: I dont want to block on the queue. that would be detrimental and can block a lot of threads.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
shiveshr: its passed through a constructor thats only created for testing. 
production code does pass null

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
shiveshr: since we assigned this new connection creation to a request, that request will get the failure and will retry after returning the failed connection to this class.

shiveshr:
@tkaitchuck @andreipaduroiu thanks for the review comments. I am addressing them. However, I will close this PR for now.. Will reopen it after @shrids #3531 is complete. 
That will simplify the whole connection logic as we can then multiplex over a single connection and connectionFactory will internally maintain the state of connection.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -845,16 +826,21 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
             return new AbstractMap.SimpleImmutableEntry<>(key, value);
         }).collect(Collectors.toList());
 
         WireCommands.UpdateTableEntries request = new WireCommands.UpdateTableEntries(requestId, qualifiedName, delegationToken,
                                                                                       new WireCommands.TableEntries(wireCommandEntries));
-        sendRequestAsync(request, replyProcessor, result, clientCF, ModelHelper.encode(uri));
-        return result;
+        sendRequestAsync(request, replyProcessor, result, ModelHelper.encode(uri));
+        return result
+                .whenComplete((r, e) -> buffersToRelease.forEach(ReferenceCounted::release));
```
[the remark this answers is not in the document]
shiveshr: i have added try catch block inside sendRequestAsync to catch all synchronous failures so that result future is always completed.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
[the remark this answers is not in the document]
shiveshr: I have abstracted out ResourcePool in common.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1236,28 +1225,43 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
-                    }
-                });                
+                connection.sendAsync(request, resultFuture);
             }
         });
         resultFuture.whenComplete((result, e) -> {
-            connectionFuture.thenAccept(ClientConnection::close);
+            // when processing completes, return the connection back to connection manager asynchronously.
+            // Note: If result future is complete, connectionFuture is definitely complete. 
+            if (resultFuture.isCompletedExceptionally()) {
```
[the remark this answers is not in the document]
shiveshr: Fixed it.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
shiveshr: On @andreipaduroiu 's suggestion i have moved the resource pooling logic to a new ResourcePool class.

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
shiveshr: Fixed

andreipaduroiu:
@shiveshr You closed it and then reopened it, and you said it should be closed. What is the final status? Shall we keep looking at this?

shiveshr:
@andreipaduroiu please review. i had closed it because i thought i would take a dependence on #3531 but that is still WIP so lets get this in and i will make incremental changes whenever #3531 gets in.

common/src/main/java/io/pravega/common/util/ResourcePool.java
```diff
@@ -0,0 +1,286 @@
...
+    /**
+     * Method to get a resource initialized with supplied arg.
+     * This method attempts to find an existing available resource.
+     * If not found, it submits a new waiting request for whenever a resource becomes available. A resource could become available
+     * because such a resource was returned to the pool or a new resource was created.
+     * It also opportunistically submits a request to create a new resource if required.
+     *
+     * @return A completableFuture which when completed will have the resource object that the caller requested.
+     */
+    public CompletableFuture<T> getResource() {
```
andreipaduroiu: This looks good. However I have a suggestion that may improve the usability of this. 

Currently we need to call `getResource()` and then `returnResource` when we're done with a resource. This will force the caller to hold a reference to this pool, which, while doable, may not produce the nicest code.

One thing I've done in the past is to define a nested class `AutoReturnResource<T> implements AutoCloseable`. This can only be created in the `ResourcePool` class and will be returned by `getResource`. This new class will have a getter providing access to the resource, but upon calling `close` it will internally invoke the pool's `returnResource` (which will now be made private).

Both implementations are equivalent, but IMO the `AutoReturnResource` looks a bit more elegant and makes it obvious to the user that it's a finite resource and needs to be cleaned up because it implements `AutoCloseable`.

common/src/main/java/io/pravega/common/util/ResourcePool.java
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        if (waiting != null) {
+            try {
+                tSupplier.get().whenComplete((t, e) -> {
+                    if (e != null) {
+                        waiting.future.completeExceptionally(e);
+                    } else {
+                        if (listener != null) {
+                            listener.notify(Event.Created);
```
andreipaduroiu: Is there a chance that `listener.notify` will throw? If so, `waiting.future` may never complete and be left hanging. Consider adding a try-catch in here to cover this case.

andreipaduroiu: `Futures.completeAfter` does something similar, though I think it won't help you much here.

common/src/main/java/io/pravega/common/util/ResourcePool.java
```diff
@@ -0,0 +1,286 @@
...
+                        }
+                        waiting.future.complete(t);
+                    }
+                });
+            } catch (Exception e) {
```
andreipaduroiu: An assertion error may also be thrown, in which case your future will be left hanging. I suggest catching `Throwable` here. Check out `Futures.completeAfter`

common/src/main/java/io/pravega/common/util/ResourcePool.java
shiveshr: i have made the listener an inner class here from an interface.. so it simply writes an event into a queue. its not supposed to throw. 
and its only used for signaling to the unit test.

common/src/main/java/io/pravega/common/util/ResourcePool.java
```diff
@@ -281,6 +279,48 @@ public void shutdown() {
...
-        private final CompletableFuture<T> future;
+        private final CompletableFuture<ClosableResource<T>> future;
+    }
+
+    /**
+     * A closeable resource wrapper class which returns the resource back to the pool automatically once it is closed.
+     * Its close method is idempotent and can be invoked multiple times without returning the resource more than once. 
+     * @param <T> Type of underlying resource
+     */
+    public static class ClosableResource<T> implements AutoCloseable {
```
andreipaduroiu: `Closable` -> `Closeable`. Even though my browser's spellcheck complains about this, I prefer to be consistent with `AutoCloseable`.

andreipaduroiu: plus you used `closeable` in line 286 :)

common/src/main/java/io/pravega/common/util/ResourcePool.java
```diff
@@ -281,6 +279,48 @@ public void shutdown() {
...
+        }
+        
+        @Override
+        @Synchronized
+        public void close() {
+            // Close is idempotent. 
+            // If close had already been invoked on this resource wrapper, then we do not return the resource to the pool.  
+            if (!isClosed) {
+                if (invalid.get()) {
+                    resourcePool.returnResource(resource, false);
```
andreipaduroiu: 1. `resourcePool.returnResource(resource, !invalid.get())`. That way you can also eliminate `returnResource()` (with no args).

2. Be very careful with this `synchronized` method. You're acquiring a lock here and acquiring a different lock in `resourcePool.returnResource`. If not used carefully, this could lead to a deadlock. You are completing a future in `returnResource` while holding this sub-class' lock, so that is a a potential cause for concern. Would we be able to solve the same problem by changing `isClosed` to an `AtomicBoolean` and using `.compareAndSet(false, true)` to ensure `close` is called only once?

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -117,21 +118,66 @@ public void close() throws IOException {
...
+
+        @VisibleForTesting
+        ReplyProcessor getReplyProcessor() {
+            return resource.getResource().reusableReplyProcessor.replyProcessor.get();
+        }
+        // endregion
+
+        @Override
+        @Synchronized
+        public void close() {
```
andreipaduroiu: You're also holding a lock here while closing your resource, which in turn may complete a future while holding it.

common/src/main/java/io/pravega/common/util/ResourcePool.java
shiveshr: done

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
shiveshr: fixed

controller/src/main/java/io/pravega/controller/server/SegmentStoreConnectionManager.java
```diff
@@ -0,0 +1,421 @@
...
+        @VisibleForTesting
+        ReplyProcessor getReplyProcessor() {
+            return resource.getResource().reusableReplyProcessor.replyProcessor.get();
+        }
+        // endregion
+
+        @Override
+        public void close() {
+            boolean toClose = false;
+            synchronized (lock) {
```
shrids: could we use the following pattern?
```
AtomicBoolean closed = new AtomicBoolean(false);
  if(!closed.getAndSet(true)) {
                 ConnectionObject connectionObject = resource.getResource();
                connectionObject.reusableReplyProcessor.uninitialize();
                if (!connectionObject.state.get().equals(ConnectionObject.ConnectionState.CONNECTED)) {
                    resource.invalidate();
                }
    }
```

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
tkaitchuck: Yes, I was suggesting we generalize this to include other types of exceptions especially unknown exceptions. If there are exceptions for which we want to consider the connection still valid, then we should have a conditional on those to return the connection.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1220,29 +1224,88 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
+     * @param e Exception, if any, thrown from the request processing. 
+     */
+    private void requestCompleteCallback(CompletableFuture<ConnectionWrapper> connectionFuture, 
+                                         Throwable e) {
+        // when processing completes, return the connection back to connection manager asynchronously.
+        // Note: If result future is complete, connectionFuture is definitely complete. if connectionFuture had failed,
+        // we would not have received a connection object anyway. 
+        if (e != null) {
+            Throwable unwrap = Exceptions.unwrap(e);
+            if (hasConnectionFailed(unwrap)) {
```
tkaitchuck: See my earlier comment. I think any unknown exception should be treated as a failure so that the next caller will receive a clean connection.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: opened issue #3577

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: opened seperate issue https://github.com/pravega/pravega/issues/3577 where we will tackle it.

PR #3528, opened 2019-03-26
Issue 3527: Ensure K8sClient.waitUntilPodCompletes method completes when all retries are exhausted.

### LINKED ISSUE #3527: K8sClient should not retry on  UnknownHostException
**Problem description**
Currently `K8sClient` retries if an IOException is observed while downloading logs and tracking the status of a pod. The K8sClient should not retry if it faces `UnknownHostException` which also of type `IOException`. 

**Problem location**
K8sClient.

On observing an UnknownHostException the test framework should fail the test.

shrids:
**Change log description**  
Fix issue with `K8sClient.waitUntilPodCompletes` method which causes the future returned by it to never complete.

**Purpose of the change**  
Fixes #3527 

**What the code does**  
Currently K8sClient retries incase of an `IOException` while tracking the status of the test pod for completion and download individual test logs. Now after all the retries have been exhausted the method `K8sClient.waitUntilPodCompletes` did not complete the future returned by it exceptionally causing the framework to stall.

**How to verify it**  
All the tests should continue to pass. Verified the error scenarios by simulating the `UnknownHostException`

### DIFF
--- a/test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
+++ b/test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
@@ -47,7 +47,6 @@ import io.pravega.common.util.Retry;
 import io.pravega.test.system.framework.TestFrameworkException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.UnknownHostException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.time.Duration;
@@ -55,7 +54,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
```diff
@@ -614,7 +617,13 @@ public void deletePVC(String namespace, String name) {
                             String logFile = toFile + "-" + retryCount.incrementAndGet() + ".log";
                             Files.copy(logStream, Paths.get(logFile));
                             log.debug("Logs downloaded from pod {} to {}", podName, logFile);
+                            return null;
                         } catch (ApiException | IOException e) {
+                            if (e instanceof UnknownHostException) {
+                                // do not retry if an UnknownHostException is observed.
```
fpj: Should we retry a bounded number of times? That would accommodate some transient failures, but not block the job indefinitely.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
```diff
@@ -498,11 +500,12 @@ public void deletePVC(String namespace, String name) {
         Retry.RetryAndThrowConditionally retryConfig = retryWithBackoff
                 .retryWhen(t -> {
                     Throwable ex = Exceptions.unwrap(t);
-                    if (ex.getCause() instanceof IOException) {
+                    if (ex.getCause() instanceof IOException && !(ex.getCause() instanceof UnknownHostException)) {
```
fpj: It is odd that we are not handling these IO exceptions in any way. What reason do we have to believe that they are benign?

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
shrids: The retry is  done a bounded number of times. I have updated the code and description of the PR.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
shrids: In case of IO exception retry is performed a bounded number of times.

PR #3530, opened 2019-03-27
Issue 3104: Clean up gradle properties

### LINKED ISSUE #3104: Clean up gradle.properties
**Problem description**
There are some variables with default values that we should revisit and clean up in `gradle.properties`:

```
#Default values for cluster name, master IP and number of slaves in the cluster.
#To override this use -PCLUSTER_NAME=<> -PMASTER=<> -PNUM_SLAVES=<> while running gradle task
CLUSTER_NAME=ant-man
MASTER=127.0.0.1
NUM_SLAVES=3

#Default host,user path values for pravega.io server
#To override this use -Phostname=<> -PMASTER=<> -Pusername=<> while running gradle task
hostname=0.0.0.0
username=root

#Default config values for system test execution on AWS
#To override this use -Paws_access_key=<> and so on... while running gradle task `startSystemTestsWithDocker`
aws_access_key=null
aws_secret_key=null
aws_region=us-east-2
aws_key_name=aws-key-pair
cred_path=/home/ubuntu
config_path=/home/ubuntu
pravega_org=pravega
pravega_branch=master
travis_commit=0
```

**Problem location**
Gradle properties

Clean up variables and values.

fpj:
Signed-off-by: Flavio Junqueira (fpj) <flavio.junqueira@emc.com>

**Change log description**  
* Updates gradle.properties to remove properties related tests that are needed only when running system tests.

**Purpose of the change**  
Fixes #3104 

**What the code does**  
There is no code change. This pull request is removing a bunch of gradle properties from the `gradle.properties` file, which are specific to running tests and can be passed via command line.

Although I have removed those properties altogether (and posted them to the wiki), it would also be fine to leave them without any parameters. I removed because I felt that they are cluttering the properties file unnecessarily.

**How to verify it**  
There is no functional change, but it should build regularly.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -829,7 +829,7 @@ project('test:system') {
         if(ext.getProperty('awsExecution').toString().equals("true")) {
             commandLine './aws/preTestScript.sh'
             args "$aws_access_key", "$aws_secret_key", "$aws_region", "$aws_key_name", "$cred_path", "$config_path", "$pravega_org", "$pravega_branch"
-        } else {
+        } else if (project.hasProperty('CLUSTER_NAME')) {
             commandLine "./preTestScript.sh"
             args "$CLUSTER_NAME", "$MASTER", "$NUM_SLAVES"
         }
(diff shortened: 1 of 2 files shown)


fpj:
Ok, it sounds like I can't simply remove them as gradle will complain. I'll fix it.

PR #3541, opened 2019-03-29
Issue 3478: Memory footprint improvement: Stream store improvements

shiveshr:
**Change log description**  
Implement several improvements to controller's memory usage in stream metadata store - avoid redundant objects, avoid redundant copies, avoid multiple deserializations.

**Purpose of the change**  
Fixes parts of #3478 for metadata store related improvements

**What the code does**  
1. Caching:
Cache the java objects rather than byte arrays that we fetch from the store. 
This has the advantage that we dont have to deserialize objects everytime we read it from the cache (which can happen multiple times). 
Saves on both compute and memory (because we avoid creation of duplicate object). The downside is, cached objects can stay referenced for longer even when they are not needed. But we have realized that upside weighs more than downside and has bigger perf implication. 
In order to mitigate the cached object memory footprint, I have also reduced cache duration from 10 minutes to 2 minutes. Objects will be evicted from the cache 2 minutes after last access. 

So basically the contract that store class had with its store-specific implementation was to deliver a wrapper `Data` object that had byte[] and version in it. With this change we have removed the `Data` class. instead we have changed that contract to use `VersionedMetadata` (which has generic {Object+version}). This applies to cache as well. 
So `Data.java` is deleted and `PersistentStreamBase.java` and its derived classes (`ZkStream.java` and `InMemoryStream.java` now have contract using typed `VersionedMetadata<MetadataType>` params. 

2. StreamSegmentRecord object was being transformed into Segment object and vice versa which was unnecessary and we have now exposed `StreamSegmentRecord` to store apis instead of `Segment` object. This saves on unnecessary creation of duplicate objects. 

3. `copyCollections` flag on Metadata objects with collections. All objects created via deserialization will no longer make another "immutable" copy of the collection. For very large objects, we were creating a lot of copies of collection objects. This results in a lot of `Node` objects being created. 
Refer to class in package `io.pravega.controller.store.stream.records`.
So what we have done is, for collections created internally, we dont make a copy. 
for externally supplied collections we make a copy.

getters always wrap collections under unmodifiable collections. so our data objects continue to be immutable while we save on copying the collection unnecessarily. 

4. transaction UUID to String to UUID conversion is avoided. PersistentStreamBase makes use of UUID as transaction id as do all derived classes. Only before writing to zk path and reading from zk path is a UUID converted to and from a string. 

5. Few minor improvements: 
PersistentStreamBase.isStreamCutValid method is optimized to not make too many getEpochRecord calls to cache. 
AbstractStreamMetadataStore does not compute splits and merges after every scale anymore as that is a very costly computation. 
Removed Open transactions computation. that is a very costly operation. 

**How to verify it**  
All tests should pass.


controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -434,12 +434,10 @@ public OperationContext createContext(String scope, String name) {
                                                        final Executor executor) {
         CompletableFuture<Void> future = withCompletion(getStream(scope, name, context).scaleOldSegmentsSealed(sealedSegmentSizes, record), executor);
 
-        future.thenCompose(result -> CompletableFuture.allOf(
+        future.thenCompose(result -> 
                 getActiveSegments(scope, name, null, executor).thenAccept(list ->
-                        StreamMetrics.reportActiveSegments(scope, name, list.size())),
-                findNumSplitsMerges(scope, name, executor).thenAccept(simpleEntry ->
```
fpj: This is removing the reporting of splits and merges, but it is not listed in the PR description. Is there any real impact to the reporting of splits and merges the fact that we aren't doing on every scale call. In particular, I'm wondering about this issue #1761 .

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -760,20 +758,7 @@ protected Scope getScope(final String scopeName) {
 
         return result;
     }
-
-    private CompletableFuture<SimpleEntry<Long, Long>> findNumSplitsMerges(String scopeName, String streamName, Executor executor) {
```
fpj: Where is this moving to? A browser search is not giving me anything...

controller/src/main/java/io/pravega/controller/store/stream/Cache.java
```diff
@@ -15,46 +15,50 @@
...
-                    public CompletableFuture<Data> load(final String key) {
-                        CompletableFuture<Data> result = loader.get(key);
-                        result.exceptionally(ex -> {
-                            invalidateCache(key);
-                            return null;
-                        });
-                        return result;
-                    }
-                });
+                            .maximumSize(MAX_CACHE_SIZE)
```
fpj: This max cache size is in number of objects, right? What about the size of objects? It is not the same caching 10k objects of 1 byte vs. 10k objects of 1MB each.

controller/src/main/java/io/pravega/controller/store/stream/Cache.java
```diff
@@ -15,46 +15,50 @@
...
+                            .expireAfterAccess(2, TimeUnit.MINUTES)
```
fpj: This is looking like a magic number, we don't know whether 2 minutes or 10 minutes or anything is a good value, is it right?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -70,7 +71,7 @@
     private final String name;
     private final AtomicInteger historyChunkSize;
     private final AtomicInteger shardSize;
-
+
```
fpj: Unnecessary spaces/tab.

controller/src/main/java/io/pravega/controller/store/stream/ZKGarbageCollector.java
```diff
@@ -159,7 +159,7 @@ int getLatestBatch() {
 
     @VisibleForTesting
     CompletableFuture<Void> fetchVersion() {
-        return zkStoreHelper.getData(guardPath)
+        return zkStoreHelper.getData(guardPath, x -> x)
```
fpj: Out of curiosity, wouldn't we be able to reduce this lambda to a single call to `getData` by replacing the `x -> x` argument with the function that deserializes it in line 163?

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
```diff
@@ -50,14 +50,19 @@
     private final Map<Long, StreamSegmentRecord> segmentMap = new HashMap<>();
     
     @Builder
-    EpochRecord(int epoch, int referenceEpoch, List<StreamSegmentRecord> segments, long creationTime) {
+    private EpochRecord(int epoch, int referenceEpoch, List<StreamSegmentRecord> segments, long creationTime, boolean copyCollections) {
         this.epoch = epoch;
         this.referenceEpoch = referenceEpoch;
-        this.segments = ImmutableList.copyOf(segments);
+        this.segments = copyCollections ? ImmutableList.copyOf(segments) : segments;
```
fpj: This looks like bug prone. `EpochRecord` is a data class, right? As such, don't we expect the state of its objects to be immutable?

controller/src/main/java/io/pravega/controller/store/stream/records/EpochTransitionRecord.java
```diff
@@ -55,16 +54,23 @@
      * Key ranges for new segments to be created.
      */
     final Map<Long, Map.Entry<Double, Double>> newSegmentsWithRange;
-
+
```
fpj: Unnecessary spaces/tab.

controller/src/main/java/io/pravega/controller/store/stream/records/EpochTransitionRecord.java
```diff
@@ -55,16 +54,23 @@
...
 
-    public EpochTransitionRecord(int activeEpoch, long time, Set<Long> segmentsToSeal, Map<Long, Map.Entry<Double, Double>> newSegmentsWithRange) {
+    @Builder
+    private EpochTransitionRecord(int activeEpoch, long time, Set<Long> segmentsToSeal, Map<Long, Map.Entry<Double, Double>> newSegmentsWithRange, 
+                                 boolean copyCollection) {
         this.activeEpoch = activeEpoch;
         this.time = time;
-        this.segmentsToSeal = ImmutableSet.copyOf(segmentsToSeal);
-        this.newSegmentsWithRange = ImmutableMap.copyOf(newSegmentsWithRange);
+        this.segmentsToSeal = copyCollection ? ImmutableSet.copyOf(segmentsToSeal) : segmentsToSeal;
```
fpj: We seem to be repeating this pattern, I'm not sure this is a good idea.

controller/src/main/java/io/pravega/controller/store/stream/records/HistoryTimeSeriesRecord.java
```diff
@@ -61,11 +61,18 @@
         }
         this.epoch = epoch;
         this.referenceEpoch = referenceEpoch;
-        this.segmentsSealed = segmentsSealed == null ? null : ImmutableList.copyOf(segmentsSealed);
-        this.segmentsCreated = segmentsCreated == null ? null : ImmutableList.copyOf(segmentsCreated);
+        List<StreamSegmentRecord> segmentsSealedList = copyCollections ? ImmutableList.copyOf(segmentsSealed) : segmentsSealed;
```
fpj: Same here.

controller/src/main/java/io/pravega/controller/store/stream/records/StreamCutRecord.java
```diff
@@ -48,10 +47,16 @@
      */
     final Map<Long, Long> streamCut;
 
-    public StreamCutRecord(long recordingTime, long recordingSize, Map<Long, Long> streamCut) {
+    @Builder
+    private StreamCutRecord(long recordingTime, long recordingSize, Map<Long, Long> streamCut, boolean copyCollections) {
         this.recordingTime = recordingTime;
         this.recordingSize = recordingSize;
-        this.streamCut = ImmutableMap.copyOf(streamCut);
+        this.streamCut = copyCollections ? ImmutableMap.copyOf(streamCut) : streamCut;
```
fpj: Same here.

controller/src/main/java/io/pravega/controller/store/stream/records/StreamTruncationRecord.java
```diff
@@ -72,7 +72,7 @@
      * This is non empty while truncation is ongoing.
      * This is reset to empty once truncation completes by calling mergeDeleted method.
      */
-    private final ImmutableSet<Long> toDelete;
+    private final Set<Long> toDelete;
```
fpj: Same here about changing the immutability of the data structure.

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
```diff
@@ -580,6 +581,57 @@ public void testGetActiveTxn() throws Exception {
...
+        final StreamConfiguration configuration1 = StreamConfiguration.builder().scalingPolicy(policy1).build();
+        stream.create(configuration1, System.currentTimeMillis(), startingSegmentNumber).join();
+        stream.createStreamPositionNodeIfAbsent(0).join();
+        stream.updateState(State.ACTIVE).join();
+        Long creationTime = stream.getCreationTime().join();
+        Integer position = stream.getStreamPosition().join();
+        assertEquals(0, position.intValue());
+        ZKStoreHelper.ZkCacheKey<Integer> key = new ZKStoreHelper.ZkCacheKey<>(stream.getCreationPath(), position.toString(), x -> BitConverter.readInt(x, 0));
+        VersionedMetadata<?> cachedCreationTime = storeHelper.getCache().getCachedData(key).join();
+        // verify that both times are same
```
fpj: `verify that both timestamps are the same`

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
```diff
@@ -580,6 +581,57 @@ public void testGetActiveTxn() throws Exception {
...
+        assertEquals(creationTime, cachedCreationTime.getObject());
+        // delete stream.
+        stream.updateState(State.SEALING).join();
+        stream.updateState(State.SEALED).join();
+        stream.deleteStream().join();
+
+        // refresh the stream object to indicate new request context
+        stream.refresh();
+        // verify that metadata doesnt exist in the store.
```
fpj: `doesn't`

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
```diff
@@ -580,6 +581,57 @@ public void testGetActiveTxn() throws Exception {
...
+        stream.createStreamPositionNodeIfAbsent(1).join();
+        stream.updateState(State.ACTIVE).join();
+
+        Long creationTimeNew = stream.getCreationTime().join();
+        Integer positionNew = stream.getStreamPosition().join();
+        assertEquals(1, positionNew.intValue());
+
+        ZKStoreHelper.ZkCacheKey<Integer> keyNew = new ZKStoreHelper.ZkCacheKey<>(stream.getCreationPath(), positionNew.toString(), x -> BitConverter.readInt(x, 0));
+        VersionedMetadata<?> cachedCreationTimeNew = storeHelper.getCache().getCachedData(keyNew).join();
+        // verify that both times are different
```
fpj: This is a good test case. What about the case in which the stream is deleted and the cached entry still exists? Will the cached entry for the deleted stream be visible to a request?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -626,9 +627,22 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
 
     @Override
     public CompletableFuture<Boolean> isStreamCutValid(Map<Long, Long> streamCut) {
-        return Futures.allOfWithResults(streamCut.keySet().stream().map(x -> getSegment(x).thenApply(segment ->
-                new SimpleEntry<>(segment.getKeyStart(), segment.getKeyEnd())))
-                                                 .collect(Collectors.toList()))
+        Map<Integer, List<Long>> groupByEpoch = streamCut.keySet().stream().collect(groupingBy(StreamSegmentNameUtils::getEpoch));
```
fpj: According to the PR description, this change is avoiding many `getEpochRecord` calls to the cache. Could you elaborate? I don't see it.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -64,32 +76,29 @@
...
-    private final String retentionSetPath;
-    private final String retentionStreamCutRecordPathFormat;
-    private final String currentEpochRecordPath;
-    private final String epochRecordPathFormat;
-    private final String historyTimeSeriesChunkPathFormat;
-    private final String segmentSealedEpochPathFormat;
-    private final String segmentsSealedSizeMapShardPathFormat;
-
-    private final Cache cache;
+    private final AtomicReference<String> creationPath = new AtomicReference<>();
```
fpj: This is changing the immutability of the state of objects of this class. Could we optimize without breaking immutability?

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -504,137 +699,177 @@
...
+        store.invalidateCache(getStatePath(), id);
+        store.invalidateCache(getConfigurationPath(), id);
+        store.invalidateCache(getTruncationPath(), id);
+        store.invalidateCache(getEpochTransitionPath(), id);
+        store.invalidateCache(getCommittingTxnsPath(), id);
+        store.invalidateCache(getCurrentEpochRecordPath(), id);
+    }
+
+    /**
+     * Method to retrieve unique Id for the stream. We use streamPosition as the unique id for the stream.
```
fpj: I don't understand this observation, how can we use `streamPosition` as an id? Specifically, I'm wondering about why it is correct to make an id out of a position.

fpj: `... value, else`

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
shiveshr: There are two very bad metrics computation --
1. splits and merges
2. count open transactions
both are costly and scan metadata. 

Splits and merges can be stored in the metadata and updated with every scale to make it efficient. 

Count open transaction: since transactions are getting created and committed/aborted continuously, the open transaction count is very costly as it reads metadata for all transactions and then filters those that are open..
since create transaction request can be sent to any controller instance, so there can be no coordinated count. 
keeping a common counter in metadata will also be detrimental as it will need to be updated on each of the above mentioned operation (create/commit/abort).

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
shiveshr: I have removed this.

controller/src/main/java/io/pravega/controller/store/stream/Cache.java
shiveshr: In guava cache we cannot have a size based cache. its only number of elements.

controller/src/main/java/io/pravega/controller/store/stream/Cache.java
shiveshr: Yeah its a magic number.. Typically if value is regularly accessed, it should be in the cache, or LRU will be removed if the cache is full.

controller/src/main/java/io/pravega/controller/store/stream/ZKGarbageCollector.java
shiveshr: we dont want any deserialization in this case hence whatever byte array we get, we do a no op. hence the lambda x->x

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
shiveshr: Yes, its a data class. But someone constructing it gives it a collection. So we used to make an immutable copy of the collection by default.
When deserializing, the serializer gives us a collection that we know for sure will not be mutated because the deserializer is owned within this data class. so we avoid a duplicate copy of collection here by using this copy collections flag.

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
shiveshr: when a stream is deleted, the controller instance where the delete took place cleans up the cache.. but the entry may have been cached on other controller instances that are unaware of the delete.. 
the stream has a unique 'id' upon each new creation.. we use this `id` while caching.. so any cache for an old stream is not visible to anyone requesting data for a new stream with a different id

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: we dont store segmentRecords at the top level but within an epochRecord. 
EpochRecords are immutable and cached. 
Here, the original implementation called getEpochRecord for all segments. So if there were multiple segments from the same epoch, the call will be made multiple times. 

by first grouping segments by their epochs, and then making a single call per epoch, we improve the efficiency there.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: Actually all these strings are created per Stream object irrespective of whether we want to access those paths or not. This was wasteful. 
Hence i loaded them in an atomic reference on the need basis. so now its loaded once and used. 
since this is an internal private field which is not exposed to anyone outside this class, so breaking the immutability for this field for the gains we get IMO is ok.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: position is unique for each stream within a scope..
 making position as id makes it safe against recreated streams and cached values (as explained in the comment above)

controller/src/main/java/io/pravega/controller/eventProcessor/impl/ConcurrentEventProcessor.java
```diff
@@ -46,7 +46,7 @@
 @Slf4j
 public class ConcurrentEventProcessor<R extends ControllerEvent, H extends RequestHandler<R>>
         extends EventProcessor<R> {
-    private static final int MAX_CONCURRENT = 10000;
+    private static final int MAX_CONCURRENT = 1000;
```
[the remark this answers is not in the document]
shiveshr: by reducing concurrent processing of events by 10x, in peak load conditions, the memory footprint also goes down by 10x. 
10000 is a lot anyway. If each processing has a footprint of only 100kb then with 10k events being picked and pocessed, the load on memory will be 1Gb.
the actual footprint will be higher because events for which processing has completed will have their footprint shaved off only after a GC cycle. 

It will impact our throughput of events processed. but these are background asynchronous tasks anyway and we are processing 1000 events concurrently as opposed to 10k. 
We can have reasonably large number of segments on our internal streams and then the load gets be distributed across multiple controller instances to achieve high throughput. 
but we dont want to overwhelm a single controller instance by picking more work than its memory can reasonably sustain. hence 1000 seemed like a reasonable number even for an average 1mb memory footprint for each event's processing.

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
fpj: The confusing part is what is the one on what is happening to the metrics. You need to clarify in the description whether the optimization has any impact to the consumer of metrics.

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
fpj: I was not actually asking you to explain the case, but about the presence of a test case for it, but since you got into it, let's talk about it:

> any cache for an old stream is not visible to anyone requesting data for a new stream with a different id

I'm not sure I understand. A client gets the id from the controller, and given that we can have stale cache entries in other controller instances, the client could get a stale id for the same stream name, no?

shiveshr: Clients deal with stream names only. 
The id is something controllers use internally for caching only for different generations of streams with same name. 
The metadata is stored at a path that only depends on the stream name and nothing changes there.

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
shiveshr: Yes, i have removed the metrics update altogether. we dont want to continue computing these metrics whose computation is typically more expensive than a workflow execution. 

For scale we have an explicit api which the UI makes use of. 
number of open transactions is something we can add a new api or do something else. But in its current form it is not sustainable, esp with our kvs implementation where this costly computation hammers segmentstore and has detrimental effect on segment store's performance.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
fpj: You are right about them being private, but why do you want to lazily load these strings rather than loading them in the constructor like you are doing at the moment?

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
fpj: Remind me the definition of stream position so that I have more context for this, please. I still don't understand how it can be unique, but it is probably because I don't have enough context.

shiveshr: Stream position is the unique number given to a stream at the time of its creation under the "StreamsInScope" tree. Everytime a stream gets created it gets a new higher number.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: its just a minor optimization to lazily load the evaluated strings on the need basis rather than load them at the time of construction of the object. most of the time we will only need to answer only a small subset of queries and would not need to have all the strings computed but only those that are needed by those queries.

tkaitchuck: I don't understand the 'most of the time' part. Why would such an object be created over and over? Can we at least factor out this 'get__Path' logic into a helper or something that can be long lived. It's weird to intermix that with a data object.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -111,35 +120,205 @@
...
+            } else {
+                return x;
+            }
+        });
+    }
+
     // region overrides
 
     @Override
     public CompletableFuture<Integer> getNumberOfOngoingTransactions() {
```
tkaitchuck: The logic below this point seems to have little relationship to the previous 200 lines. Can we separate these?

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
```diff
@@ -46,21 +45,37 @@
...
+     * All getters of this class that return a collection always wrap them under Collections.unmodifiableCollection so that
+     * no one can change the data object from outside.  
+     */
+    private EpochRecord(int epoch, int referenceEpoch, List<StreamSegmentRecord> segments, long creationTime, boolean copyCollections) {
         this.epoch = epoch;
         this.referenceEpoch = referenceEpoch;
-        this.segments = ImmutableList.copyOf(segments);
+        this.segments = copyCollections ? ImmutableList.copyOf(segments) : segments;
         this.creationTime = creationTime;
         this.segmentMap.putAll(segments.stream().collect(Collectors.toMap(StreamSegmentRecord::segmentId, x -> x)));
```
tkaitchuck: Why is this doing a collect into a map, only to make a copy into another map ?

controller/src/main/java/io/pravega/controller/store/stream/records/HistoryTimeSeries.java
```diff
@@ -37,8 +38,26 @@
...
+     * This is a private constructor that is only directly used by the builder during the deserialization. 
+     * The deserialization passes @param copyCollections as false so that we do not make an immutable copy of the collection
+     * for the collection passed to the constructor via deserialization. 
+     *
+     * The all other constructors, the value of copyCollections flag is true and we make an immutable collection copy of 
+     * the supplied collection. 
+     * All getters of this class that return a collection always wrap them under Collections.unmodifiableCollection so that
+     * no one can change the data object from outside.  
+     */
+    private HistoryTimeSeries(List<HistoryTimeSeriesRecord> historyRecords, boolean copyCollection) {
```
tkaitchuck: This pattern is not a good one. Just choose one way or the other.

controller/src/main/java/io/pravega/controller/store/stream/records/StreamTruncationRecord.java
```diff
@@ -81,17 +82,44 @@
...
+     * for the collection passed to the constructor via deserialization. 
+     *
+     * The all other constructors, the value of copyCollections flag is true and we make an immutable collection copy of 
+     * the supplied collection. 
+     * All getters of this class that return a collection always wrap them under Collections.unmodifiableCollection so that
+     * no one can change the data object from outside.  
+     */
+    private StreamTruncationRecord(Map<Long, Long> streamCut, Map<StreamSegmentRecord, Integer> span,
+                                  Set<Long> deletedSegments, Set<Long> toDelete, long sizeTill, boolean updating, boolean copyCollections) {
+        this.streamCut = copyCollections ? ImmutableMap.copyOf(streamCut) : streamCut;
```
tkaitchuck: All of this just makes the code more complicated. The constructor should have a contract and the callers should abide by it, making it conditional is just weird. If you really want to have it two ways you can overload the constructor, but honestly it seems like in such a case it's probably better to just make the copy.

controller/src/main/java/io/pravega/controller/store/stream/records/StreamTruncationRecord.java
```diff
@@ -81,17 +82,44 @@
...
         this.spanEpochHigh = span.values().stream().max(Comparator.naturalOrder()).orElse(Integer.MIN_VALUE);
     }
+    
+    public StreamTruncationRecord(Map<Long, Long> streamCut, Map<StreamSegmentRecord, Integer> span,
+                                  Set<Long> deletedSegments, Set<Long> toDelete, long sizeTill, boolean updating) {
+        this(streamCut, span, deletedSegments, toDelete, sizeTill, updating, true);
+    }
+
+    public Map<Long, Long> getStreamCut() {
+        return Collections.unmodifiableMap(streamCut);
```
tkaitchuck: If the value passed into the constructor were always copied to an immutable map this would not be needed.

controller/src/main/java/io/pravega/controller/store/stream/records/StreamTruncationRecord.java
shiveshr: the constructor contract on public constructors is a simple one. 
This is a private constructor where we are doing the optimization. 

This whole thing is being done for reducing creation of avoidable collection.Node objects.

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
shiveshr: This becomes more relevant for streams with very large number of concurrent segments. 
For a stream with 10k segments (like we create in metadata scalability system tests), this avoids creation of 10k duplication collection node for each such object. 

A rough calculation where 32 bytes per object (16 for the object and 16 for the reference fields) will translate to ~320kb of raw collection node footprint. The java object with its paddings and alignments takes more. Maybe around 500kb or 0.5Mb. This is just arising from creating a copy of the collection which can be avoided altogether. 

For each such object that we create, we could have large number of such collection nodes that will have a memory footprint until garbage collected.

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
shiveshr: we compute and keep this map from a list so that this data object can efficiently respond to queries for `getSegmentForId`

controller/src/main/java/io/pravega/controller/store/stream/records/StreamTruncationRecord.java
shiveshr: yes, but thats exactly what we are trying to avoid. creation of duplicate collection nodes. please refer to the comment above where i have explained the rationale and approach.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: the logic below is encapsulated under "region overrides". Is that sufficient or are you suggesting something else?

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: let me explain what we are doing here:
a stream object has multiple types of metadata records. each stored at a different path. 
most queries only deal with specific metadata records. 
For example writers or readers asking for successors. 
We were originally computing all metadata paths for a stream in the constructor of this Stream object. Which meant we were computing paths that may not even be required by queries coming in. 
It is important to note that controller doesnt keep these stream objects forever in its memory. 
They are created on demand, and kept in a cache for a brief period until they are regularly accessed. 
The controller will instantiate the object, and this object will be shared across multiple queries. Once no activity happens on the object for a while, it will be GC'd. 
So `most of the time` is about the kind of queries coming in during the lifespan of the object. 
If the queries coming in are typically `gets` that only require access to a small set of paths, then evaluating all the paths before hand may be wasteful. 

So for a query like getSuccessor we only need a handful of these paths to be evaluated. So evaluating and keeping all the strings in the object is wasteful. Each string being evaluated is about 50-100 chars or having a footprint of 100-200 bytes. The atomic reference object has a smaller footprint in comparison. So with 20 odd string paths, we maybe saving on maybe around 1000-2000 bytes per stream object. I dont know if the gains are significant enough to do this. 
but that is the reason for this choice. 

> Can we at least factor out this 'get__Path' logic into a helper 
So we can create a helper for computing the paths. but paths are specific to the stream in question.. so the helper cannot keep all the paths for all the streams.. and we dont want to compute the path every time before accessing the path. that would be more wasteful. 
so we want to keep the computed path in the stream object for reuse.

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
fpj: We had to resolve this issue offline, it is not trivial to understand why this is correct.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: On second thoughts, the gains here are not much and the complexity increase is more.. i will revert these changes.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: reverted the changes above

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
shiveshr: employed the approach to use immutable collections in our serializers.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -548,7 +554,7 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
                             });
                         }
                     });
-                }).thenApply(x -> segments);
+                }).thenApply(x -> ImmutableSet.copyOf(segments));
```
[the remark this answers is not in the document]
shiveshr: yes, the segments list coming in is from outside the store. so we create a copy here.

controller/src/main/java/io/pravega/controller/store/stream/records/CommittingTransactionsRecord.java
```diff
@@ -46,47 +44,33 @@
...
     }
 
-    public CommittingTransactionsRecord(int epoch, List<UUID> transactionsToCommit, int activeEpoch) {
-        this(epoch, transactionsToCommit, Optional.of(activeEpoch), true);
+    public CommittingTransactionsRecord(int epoch, @NonNull ImmutableList<UUID> transactionsToCommit, int activeEpoch) {
+        this(epoch, transactionsToCommit, Optional.of(activeEpoch));
     }
     
     @Builder
-    /**
```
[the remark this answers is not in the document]
shiveshr: yes, the comment block was no longer relevant.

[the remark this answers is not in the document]
shiveshr: this comment is no longer valid as we have made all params immutable and our serialization also deserializes into an immutable collection. so all the `copyCollection` business is no longer applicable

common/src/main/java/io/pravega/common/io/serialization/RevisionDataInput.java
```diff
@@ -159,6 +178,23 @@
...
+     *
+     * @param keyDeserializer   A Function that will decode a single Key of the Map from the given RevisionDataInput.
+     * @param valueDeserializer A Function that will decode a single Value of the Map from the given RevisionDataInput.
+     * @param newMapBuilder     An ImmutableMapBuilder that will create a new instance of the ImmutableMap type desired.
+     * @param <K>               Type of the Keys in the Map.
+     * @param <V>               Type of the Values in the Map.
+     * @return A new Map. If the original Map passed to RevisionDataOutput.writeMap() was null, this will return an empty map.
+     * @throws IOException If an IOException occurred.
+     */
+    <K, V> ImmutableMap<K, V> readMap(ElementDeserializer<K> keyDeserializer, ElementDeserializer<V> valueDeserializer,
```
andreipaduroiu: Why didn't you use the same pattern as for `readCollection` above? (`C extends ...`). The `readCollection` was consistent with the other `readCollection` and `readMap`, but this `readMap` stands out as being different. Since this is a core, generic class, we should make it as flexible as possible.

common/src/main/java/io/pravega/common/io/serialization/RevisionDataInputStream.java
```diff
@@ -143,6 +146,14 @@ public UUID readUUID() throws IOException {
         }
         return result;
     }
+    
+    @Override
+    public <T, C extends ImmutableCollection<T>> void readCollection(ElementDeserializer<T> elementDeserializer, C.Builder<T> newCollection) throws IOException {
```
andreipaduroiu: `newCollection` -> `builder`

common/src/main/java/io/pravega/common/io/serialization/RevisionDataInputStream.java
```diff
@@ -178,5 +189,15 @@ public UUID readUUID() throws IOException {
         return result;
     }
 
+    @Override
+    public <K, V> ImmutableMap<K, V> readMap(ElementDeserializer<K> keyDeserializer, ElementDeserializer<V> valueDeserializer, ImmutableMap.Builder<K, V> newMap) throws IOException {
```
andreipaduroiu: `newMap` -> `builder`

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1461,9 +1467,13 @@ private int getShardNumber(long segmentId) {
         return StreamSegmentNameUtils.getEpoch(segmentId) / shardSize.get();
     }
 
-    private Map<StreamSegmentRecord, Integer> convertToSpan(EpochRecord epochRecord) {
-        return epochRecord.getSegments().stream()
-                          .collect(Collectors.toMap(x -> x, x -> epochRecord.getEpoch()));
+    private ImmutableMap<StreamSegmentRecord, Integer> convertToSpan(EpochRecord epochRecord) {
+        ImmutableMap.Builder<StreamSegmentRecord, Integer> builder = ImmutableMap.builder();
+        epochRecord.getSegments().stream()
+                          .forEach(x -> {
```
andreipaduroiu: no need for braces (`{}`)

common/src/main/java/io/pravega/common/io/serialization/RevisionDataInput.java
shiveshr: ImmutableCollection is base class for ImmutableSets, ImmutableLists, etc. 
i was trying to use the same pattern as immutableMap but couldnt find a way. 
The problem is, the builder that we give out is still typed as ImmutableCollection.Builder. 
Doing a build on it we get ImmutableCollection and not the derived type. I dont want to do an unchecked typecast just because i know the builder comes from derived type. 

I tried patterns like
```    public <T, C extends ImmutableCollection<T>> C readCollection(ElementDeserializer<T> elementDeserializer, C.Builder<T> builder) 
```

```
    public <T, B extends ImmutableCollection.Builder<T>> ImmutableCollection<T> readCollection(ElementDeserializer<T> elementDeserializer, B builder) 
```

So i decided to let the builder be populated.. the caller created the builder with correct type and when it calls build on it, the derived class' build method is invoked and we get correctly typed collection.

shiveshr: Actually readMap is consistent with others in that it returns the map. 
the readCollection doesnt return a collection so that deviates from other methods..

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -569,12 +564,7 @@ public OperationContext createContext(String scope, String name) {
                                                           final UUID txId, final OperationContext context,
                                                           final Executor executor) {
         Stream stream = getStream(scope, streamName, context);
-        return withCompletion(stream.commitTransaction(txId), executor)
-                .thenApply(result -> {
-                    stream.getNumberOfOngoingTransactions().thenAccept(count ->
-                            TransactionMetrics.reportOpenTransactions(scope, streamName, count));
-                    return result;
-                });
+        return withCompletion(stream.commitTransaction(txId), executor);
```
RaulGracia: You may want to check the impact of removing this with the platform team. Just wondering if they use this metric extensively, removing it without an alternative may might be an issue for them.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -42,47 +48,47 @@
...
+    private VersionedMetadata<EpochRecord> currentEpochRecord;
     @GuardedBy("lock")
-    private Map<Integer, Data> epochRecords = new HashMap<>();
+    private Map<Integer, VersionedMetadata<EpochRecord>> epochRecords = new HashMap<>();
     @GuardedBy("lock")
-    private Map<Integer, Data> historyTimeSeries = new HashMap<>();
+    private Map<Integer, VersionedMetadata<HistoryTimeSeries>> historyTimeSeries = new HashMap<>();
     @GuardedBy("lock")
-    private Data retentionSet;;
+    private VersionedMetadata<RetentionSet> retentionSet;;
```
RaulGracia: There are two `;` here.

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
```diff
@@ -130,12 +145,28 @@ public ZKStoreHelper(final CuratorFramework cf, Executor executor) {
         return result;
     }
 
-    public CompletableFuture<Data> getData(final String path) {
-        final CompletableFuture<Data> result = new CompletableFuture<>();
+    /**
+     * Method to retrieve entity from zookeeper and then deserialize it using the supplied `fromBytes` function.
```
RaulGracia: `to retrieve an entity...`

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
```diff
@@ -341,4 +372,54 @@ private BackgroundCallback callback(Consumer<CuratorEvent> result, Consumer<Thro
...
+        // Since cache is untyped and holds all types of deserialized objects, we typecast it to the requested object type
+        // based on the type in caller's supplied Deserialization function. 
+        return new VersionedMetadata<>((T) v.getObject(), v.getVersion());
+    }
+
+    void invalidateCache(String path, String id) {
+        cache.invalidateCache(new ZkCacheKey<>(path, id, x -> null));
+    }
+
+    @Data
```
RaulGracia: I think that normally annotations go immediately over the class declaration, right?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1202,33 +1221,22 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
-    @Override
-    public CompletableFuture<Map<UUID, ActiveTxnRecord>> getActiveTxns() {
-        return getCurrentTxns()
-                .thenApply(x -> x.entrySet()
-                                 .stream()
-                                 .collect(toMap(k -> UUID.fromString(k.getKey()),
-                                         v -> ActiveTxnRecord.fromBytes(v.getValue().getData()))));
-    }
-
+
```
RaulGracia: Actually, I though on using this as a part of the information printed to describe a Stream from the CLI...

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -605,14 +644,41 @@
...
+     * The id of a stream is fixed for lifecycle of a stream and only changes when the stream is deleted and recreated. 
+     * The id is used for caching entities and safeguarding against stream recreation. 
+     * @return CompletableFuture which when completed contains stream's position as the id. 
+     */
+    private CompletableFuture<String> getId() {
+        String id = this.idRef.get();
+        if (!Strings.isNullOrEmpty(id)) {
+            return CompletableFuture.completedFuture(id);
+        } else {
+            // We will return empty string as id if the position does not exist.
```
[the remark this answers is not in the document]
shiveshr: its just going to result in value being loaded into cache against an empty id + path. which is harmless. when id becomes available, it will get cached against key with id and previous cache key will eventually be evicted from the cache.

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
shiveshr: I have moved it to a different point where it is computed. 
After completion of a commit workflow. 
Look at persistentStreamBase

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: The method is not removed. Its implementation is pushed down to the derived classes. 

This method was calling getCurrentTxns and converting String to UUID. 
getCurrentTxns used to be an abstract method in PersistentStreamBase and had store specific implementation.
I removed getCurrentTxns and made the implementation of getActiveTxns in ZkStream and InMemoryStream which now directly deals with UUID rather than returning String which again gets converted to UUID in this class.
Done the same for all txn related apis where we dont do one more scan of the collection to convert string to UUID

PR #3542, opened 2019-03-29
Issue 3533: Fix StreamTransactionMetadataTasks bugs

### LINKED ISSUE #3533: Illegal State Exception can be thrown from StreamTransactionMetadataTasks
**Problem description**
StreamTransactionMetadataTasks get event writers initialized after internal streams are created. 
This has an effect that if any request comes to it before the initialization, it throws an IllegalStateException.
It has this check for requests like createTxn and pingTxn where we dont rely on writers. 
And even in requests like commit and abort where we do, we can avoid throwing IllegalStateException

Another issue in stream transaction metadata tasks is inconsistent usage of routing key. 

**Problem location**
StreamTransactionMetadataTasks

Fix the issue.

shiveshr:
**Change log description**  
Fixes couple of issues in stream transaction metadata tasks -
1. Removes unnecessary checkready that results in IllegalState Exception. Replaces it with writer future. 
2. Fixes issue in inconsistent usage of routing key for writing events into commit and abort streams. 

**Purpose of the change**  
Fixes #3533 

**What the code does**  
Earlier streamTransaction metadata tasks class would wait until writers were initialized before doing any operation. The writers are created externally and passed to this class (We will fix that part of it later in a technical debt issue). For now, it doesnt have to do that as `createTxn` and `pingTxn` do not depend on internal stream writer's initialization. They can proceed before initialization happens. 
And commit and abort writers initialization can be wrapped under a promise as a CompletableFuture and the requests can be chained to it. 
So now, even if we receive any requests before writer initialization, we dont throw illegal state exception anymore. We simply return a future which will be completed when the writers are initialized. 

Another issue we fix is where in one of writeCommitEvent methods, the routing key used was incorrect. So to avoid such mistakes, all writeEvents of a type end up calling the same method which does the actual write into the stream using the correct routing key. 

**How to verify it**  
Unit test added.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
@@ -597,7 +597,7 @@ public class StreamTransactionMetadataTasks implements AutoCloseable {
         return commitWriterFuture
                 .thenCompose(commitWriter -> commitWriter.writeEvent(event.getKey(), event));
     }
-    
+
     CompletableFuture<TxnStatus> writeCommitEvent(String scope, String stream, int epoch, UUID txnId, TxnStatus status) {
         CommitEvent event = new CommitEvent(scope, stream, epoch);
         return TaskStepsRetryHelper.withRetries(() -> writeCommitEvent(event)
@@ -628,7 +628,6 @@ public class StreamTransactionMetadataTasks implements AutoCloseable {
     }
 
     private CompletableFuture<Void> notifyTxnCreation(final String scope, final String stream,
-                                 
                                                       final List<Segment> segments, final UUID txnId) {
         return Futures.allOf(segments.stream()
                 .parallel()
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
@@ -77,10 +77,6 @@ public class StreamTransactionMetadataTasks implements AutoCloseable {
      */
     private static final int MAX_EXECUTION_TIME_MULTIPLIER = 1000;
 
-    protected EventStreamWriter<CommitEvent> commitEventEventStreamWriter;
-    protected EventStreamWriter<AbortEvent> abortEventEventStreamWriter;
-    protected String commitStreamName;
-    protected String abortStreamName;
     protected final String hostId;
     protected final ScheduledExecutorService executor;
 
@@ -95,6 +91,8 @@ public class StreamTransactionMetadataTasks implements AutoCloseable {
 
     private volatile boolean ready;
     private final CountDownLatch readyLatch;
+    private final CompletableFuture<EventStreamWriter<CommitEvent>> commitWriterFuture;
+    private final CompletableFuture<EventStreamWriter<AbortEvent>> abortWriterFuture;
 
     @VisibleForTesting
     public StreamTransactionMetadataTasks(final StreamMetadataStore streamMetadataStore,
@@ -115,6 +113,8 @@ public class StreamTransactionMetadataTasks implements AutoCloseable {
         this.authHelper = authHelper;
         this.timeoutService = new TimerWheelTimeoutService(this, timeoutServiceConfig, taskCompletionQueue);
         readyLatch = new CountDownLatch(1);
+        this.commitWriterFuture = new CompletableFuture<>();
+        this.abortWriterFuture = new CompletableFuture<>();
     }
 
     public StreamTransactionMetadataTasks(final StreamMetadataStore streamMetadataStore,
@@ -124,15 +124,8 @@ public class StreamTransactionMetadataTasks implements AutoCloseable {
                                           final String hostId,
                                           final TimeoutServiceConfig timeoutServiceConfig,
                                           final ConnectionFactory connectionFactory, AuthHelper authHelper) {
-        this.hostId = hostId;
-        this.executor = executor;
-        this.streamMetadataStore = streamMetadataStore;
-        this.hostControllerStore = hostControllerStore;
-        this.segmentHelper = segmentHelper;
-        this.connectionFactory = connectionFactory;
-        this.timeoutService = new TimerWheelTimeoutService(this, timeoutServiceConfig);
-        this.authHelper = authHelper;
-        readyLatch = new CountDownLatch(1);
+        this(streamMetadataStore, hostControllerStore, segmentHelper, executor, hostId, timeoutServiceConfig, null, 
+                connectionFactory, authHelper);
     }
 
     public StreamTransactionMetadataTasks(final StreamMetadataStore streamMetadataStore,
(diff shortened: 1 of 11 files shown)


controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -171,31 +165,35 @@ public void awaitInitialization() throws InterruptedException {
...
 
     @VisibleForTesting
+    @Synchronized
     public Void initializeStreamWriters(final String commitStreamName, final EventStreamWriter<CommitEvent> commitWriter,
                                         final String abortStreamName, final EventStreamWriter<AbortEvent> abortWriter) {
-        this.commitStreamName = commitStreamName;
-        this.commitEventEventStreamWriter = commitWriter;
-        this.abortStreamName = abortStreamName;
-        this.abortEventEventStreamWriter = abortWriter;
+        if (!commitWriterFuture.isDone()) {
```
andreipaduroiu: I believe that you don't need to check if `isDone()` returns true. If you call `complete` it will ignore your request if the future is already completed (normally or exceptionally).

andreipaduroiu: Same above.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -171,31 +165,35 @@ public void awaitInitialization() throws InterruptedException {
...
+        }
         this.setReady();
         return null;
     }
 
     @VisibleForTesting
+    @Synchronized
     public Void initializeStreamWriters(final String commitStreamName, final EventStreamWriter<CommitEvent> commitWriter,
                                         final String abortStreamName, final EventStreamWriter<AbortEvent> abortWriter) {
-        this.commitStreamName = commitStreamName;
```
andreipaduroiu: It looks like a number of these arguments are no longer used. Can you trim them off the method signature?

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -613,40 +603,37 @@ private PingTxnStatus createStatus(Status status) {
...
+                .thenCompose(commitWriter -> commitWriter.writeEvent(event.getKey(), event));
     }
 
     CompletableFuture<TxnStatus> writeCommitEvent(String scope, String stream, int epoch, UUID txnId, TxnStatus status) {
-        String key = scope + stream;
         CommitEvent event = new CommitEvent(scope, stream, epoch);
-        return TaskStepsRetryHelper.withRetries(() -> writeEvent(commitEventEventStreamWriter, commitStreamName,
-                key, event, txnId, status), executor);
+        return TaskStepsRetryHelper.withRetries(() -> writeCommitEvent(event)
+                .thenAccept(v -> {
```
andreipaduroiu: In cases where `log.isDebugEnabled()` is false (log level is set to INFO - most likely production code), you will unnecessarily execute this callback for no reason, which will eat away CPU cycles.

I suggest either adding it only if `log.isDebugEnabled() == true` or merging this and `exceptionally` below into a `whenComplete` or `handle` so you have one callback.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -613,40 +603,37 @@ private PingTxnStatus createStatus(Status status) {
...
-        log.debug("Txn={}, state={}, sending request to {}", txnId, txnStatus, streamName);
-        return streamWriter.writeEvent(key, event)
-                .thenApplyAsync(v -> {
-                    log.debug("Transaction {}, sent request to {}", txnId, streamName);
-                    return txnStatus;
-                }, executor)
+    CompletableFuture<TxnStatus> writeAbortEvent(String scope, String stream, int epoch, UUID txnId, TxnStatus status) {
+        AbortEvent event = new AbortEvent(scope, stream, epoch, txnId);
+        return TaskStepsRetryHelper.withRetries(() -> writeAbortEvent(event)
+                .thenAccept(v -> {
```
andreipaduroiu: same here

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -686,14 +665,24 @@ public String retrieveDelegationToken() {
...
-            commitEventEventStreamWriter.close();
+        CompletableFuture<Void> commitCloseFuture = commitWriterFuture.thenAccept(EventStreamWriter::close);
+        CompletableFuture<Void> abortCloseFuture = abortWriterFuture.thenAccept(EventStreamWriter::close);
+
+        // since we do the checks under synchronized block, we know the promise for writers cannot be completed 
+        // by anyone if we have the lock and we can simply cancel the futures.  
+        if (commitWriterFuture.isDone()) {
+            commitCloseFuture.join();
+        } else {
+            commitWriterFuture.cancel(true);
```
andreipaduroiu: This won't cancel the underlying task (`EventStreamWriter::close`) so it won't have any effect.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -686,14 +665,24 @@ public String retrieveDelegationToken() {
...
         timeoutService.stopAsync();
         timeoutService.awaitTerminated();
-        if (commitEventEventStreamWriter != null) {
-            commitEventEventStreamWriter.close();
+        CompletableFuture<Void> commitCloseFuture = commitWriterFuture.thenAccept(EventStreamWriter::close);
+        CompletableFuture<Void> abortCloseFuture = abortWriterFuture.thenAccept(EventStreamWriter::close);
+
+        // since we do the checks under synchronized block, we know the promise for writers cannot be completed 
+        // by anyone if we have the lock and we can simply cancel the futures.  
+        if (commitWriterFuture.isDone()) {
```
andreipaduroiu: Wouldn't it be better if you wait a bit for these guys to finish? perhaps have a shutdown timeout of a few seconds, then invoke this with `get(Timeout)` so you don't have to do this complicated thing here.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
shiveshr: the reason is, we create the new writer unnecessarily if it is already done. 
isDone check avoids that.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
shiveshr: Close is synchronized. So for the production code path, either initialized or close is called. 
if close is called before initialized, then writers are never created. 
if initialize is called before close, writers are initialized and futures are complete. 
So we dont have to wait anyway.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
shiveshr: yes. refer to previous comment. we cancel the future to complete any calls like commitTxn that is waiting on the future.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -95,6 +92,8 @@
 
     private volatile boolean ready;
     private final CountDownLatch readyLatch;
```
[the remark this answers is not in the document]
shiveshr: This is still used by the sweeper to wait on readyLatch. I choose not to change that as part of this change set. Sweeper can wait for that initialization before beginning the sweep

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -115,6 +114,8 @@ public StreamTransactionMetadataTasks(final StreamMetadataStore streamMetadataSt
         this.authHelper = authHelper;
         this.timeoutService = new TimerWheelTimeoutService(this, timeoutServiceConfig, taskCompletionQueue);
         readyLatch = new CountDownLatch(1);
```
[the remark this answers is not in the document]
shiveshr: same as above. 
TxnSweeper waits on StreamTransactionMetadataTasks to become ready by calling awaitinitializatin method that waits on readyLatch.

PR #3544, opened 2019-03-29
Issue 3532: User called workflow completion checks should be tighter and include state as well

### LINKED ISSUE #3532: CheckScale for manual scale should compare epoch against EpochTransitionRecord too
**Problem description**
When manual scale finishes it performs finishing updates in following order:
1. update current epoch record to new epoch
2. reset epoch transition record
3. reset state

In checkScale method we only check against active epoch with the scale epoch. 
So if epoch transition record is not reset, the scale is not fully complete and someone attempting to submit a new scale will get ConflictException. 

We should check for epoch transition too apart from active epoch record. 

This was observed in SystemTest for scalability where we do back to back scale operations. 
If previous scale has not fully completed and next scale is attempted we get ScaleConflict. 

**Problem location**
StreamMetadataTasks

Fix the issue

shiveshr:

**Change log description**
Fixes the cause for sporadic System test failure for ScopesAndStreamManagementTest where we send update followed by seal and if update is complete but state is not updated, then seal fails and test fails. 
So we need to have stricter checks and only declare workflow complete once the state has been reset. 

**Purpose of the change**
Fixes #3532

**What the code does**

For each of the externally submitted workflows, it now says workflow is complete only after the state is also reset. For check scale, `isUpdated` and `isTruncated` until the final step of workflow (state is reset) is performed we will continue to return in-progress. 

CheckScaleDone:
Original condition: 

```
if (activeEpoch.referenceEpoch == checkScaleEpoch) return `inProgress`.
```

now new condition is:

```
if (activeEpoch.referenceEpoch == checkScaleEpoch ||
(checkScaleEpoch == activeEpoch.referenceEpoch + 1 // indicates metadata updated && 
state =SCALING // indicates exact match for our requested check scale)) return `inprogress`
```

Note: if we want even tighter check we can include ETR as well. Otherwise, when next scale starts, etr is updated and state is set to SCALING and until active epoch is updated, we will continue to get Inprogress for previous scale. I have included the ETR check as well. where if 
epoch + 1 == activeEpoch && state == scaling then we check the epoch transition record and if it is either empty or matches our previous scale then return inprogress. 

`IsTruncated`:
old condition: 

```
truncationProp.streamCut == truncationCut && !truncationProp.isUpdating ==>inProgress
```

new condition: 

```
1. truncationProp.isUpdating 
    !truncationProp.streamcut.equals(ourStreamCut) ==> done 
     else ==> InProgress
2. !truncationProp.isUpdating 
    truncationProp.streamCut == ourStreamCut && state == TRUNCATING // if truncation record is updated but state is not reset ==> InProgress
    else ==> done
```

`IsUpdated`:
```
old condition: ConfigProp.config == newConfig && !configProp.isUpdating ==> inProgress
```

new condition:
```
1. configProp.isUpdating 
    !configProp.config.equals(ourConfig) ==> done 
     else ==> InProgress
2. !configProp.isUpdating 
    configProp.config == ourConfig && state == UPDATING // if config record is updated but state is not reset ==> InProgress
    else ==> done
```

**How to verify it**
Unit test added.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
@@ -40,6 +40,7 @@ import io.pravega.controller.store.stream.StoreException;
 import io.pravega.controller.store.stream.StreamMetadataStore;
 import io.pravega.controller.store.stream.VersionedMetadata;
 import io.pravega.controller.store.stream.records.EpochRecord;
+import io.pravega.controller.store.stream.records.EpochTransitionRecord;
 import io.pravega.controller.store.stream.records.RetentionSet;
 import io.pravega.controller.store.stream.records.StreamConfigurationRecord;
 import io.pravega.controller.store.stream.records.StreamCutRecord;
@@ -571,6 +572,8 @@ public class StreamMetadataTasks extends TaskBase {
                 streamMetadataStore.getActiveEpoch(scope, stream, context, true, executor);
         CompletableFuture<State> stateFuture =
                 streamMetadataStore.getState(scope, stream, true, context, executor);
+        CompletableFuture<EpochTransitionRecord> etrFuture =
+                streamMetadataStore.getEpochTransition(scope, stream, context, executor).thenApply(VersionedMetadata::getObject);
         return CompletableFuture.allOf(stateFuture, activeEpochFuture)
                         .handle((r, ex) -> {
                             ScaleStatusResponse.Builder response = ScaleStatusResponse.newBuilder();
@@ -585,6 +588,7 @@ public class StreamMetadataTasks extends TaskBase {
                             } else {
                                 EpochRecord activeEpoch = activeEpochFuture.join();
                                 State state = stateFuture.join();
+                                EpochTransitionRecord etr = etrFuture.join();
                                 if (epoch > activeEpoch.getEpoch()) {
                                     response.setStatus(ScaleStatusResponse.ScaleStatus.INVALID_INPUT);
                                 } else if (activeEpoch.getEpoch() == epoch || activeEpoch.getReferenceEpoch() == epoch) {
@@ -592,7 +596,8 @@ public class StreamMetadataTasks extends TaskBase {
                                 } else {
                                     // active epoch == scale epoch + 1 but the state is scaling, the previous workflow 
                                     // has not completed.
-                                    if (epoch + 1 == activeEpoch.getReferenceEpoch() && state.equals(State.SCALING)) {
+                                    if (epoch + 1 == activeEpoch.getReferenceEpoch() && state.equals(State.SCALING) &&
+                                            (etr.equals(EpochTransitionRecord.EMPTY) || etr.getNewEpoch() == activeEpoch.getEpoch())) {
                                         response.setStatus(ScaleStatusResponse.ScaleStatus.IN_PROGRESS);
                                     } else {
                                         response.setStatus(ScaleStatusResponse.ScaleStatus.SUCCESS);
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
@@ -216,7 +216,7 @@ public class StreamMetadataTasks extends TaskBase {
                                     State state = stateFuture.join();
                                     StreamConfigurationRecord configProperty = configPropertyFuture.join();
 
-                                    // if property is updating and doesnt match our request, its a subsequent update
+                                    // if property is updating and doesn't match our request, it's a subsequent update
                                     if (configProperty.isUpdating()) {
                                         return !configProperty.getStreamConfiguration().equals(newConfig);
                                     } else {
@@ -412,7 +412,7 @@ public class StreamMetadataTasks extends TaskBase {
                                     State state = stateFuture.join();
                                     StreamTruncationRecord truncationRecord = configPropertyFuture.join();
 
-                                    // if property is updating and doesnt match our request, its a subsequent update
+                                    // if property is updating and doesn't match our request, it's a subsequent update
                                     if (truncationRecord.isUpdating()) {
                                         return !truncationRecord.getStreamCut().equals(streamCut);
                                     } else {
@@ -513,7 +513,7 @@ public class StreamMetadataTasks extends TaskBase {
     /**
      * Helper method to perform scale operation against an scale request.
      * This method posts a request in the request stream and then starts the scale operation while
-     * tracking its progress. Eventually, after scale completion, it sends a response to the caller.
+     * tracking it's progress. Eventually, after scale completion, it sends a response to the caller.
      *
      * @param scope          scope.
      * @param stream         stream name.
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
@@ -38,6 +38,9 @@ import io.pravega.controller.store.stream.Segment;
 import io.pravega.controller.store.stream.State;
 import io.pravega.controller.store.stream.StoreException;
 import io.pravega.controller.store.stream.StreamMetadataStore;
+import io.pravega.controller.store.stream.VersionedMetadata;
+import io.pravega.controller.store.stream.records.EpochRecord;
+import io.pravega.controller.store.stream.records.EpochTransitionRecord;
 import io.pravega.controller.store.stream.records.RetentionSet;
 import io.pravega.controller.store.stream.records.StreamCutRecord;
 import io.pravega.controller.store.stream.records.StreamCutReferenceRecord;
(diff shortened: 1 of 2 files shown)


controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -381,9 +402,25 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                                    State state = stateFuture.join();
+                                    StreamTruncationRecord truncationRecord = configPropertyFuture.join();
+
+                                    // if property is updating and doesnt match our request, its a subsequent update
+                                    if (truncationRecord.isUpdating()) {
+                                        return !truncationRecord.getStreamCut().equals(streamCut);
+                                    } else {
+                                        // if truncate-barrier is not updating, then truncate is complete if property matches our expectation 
+                                        // and state is not updating 
+                                        return !(truncationRecord.getStreamCut().equals(streamCut) && state.equals(State.TRUNCATING));
```
fpj: Something is not write here. The PR description says the following for `isTruncated`:

```
truncationProp.streamCut == truncationCut && !truncationProp.isUpdating
 && state!= UPDATING
```

while the predicate here is:

```
!(truncationRecord.getStreamCut().equals(streamCut) && state.equals(State.TRUNCATING))
```

In the case the stream cuts match and the state is `TRUNCATING`, then this returns false (not truncated), which seems right. But, if either the stream cuts do not match or the state isn't `TRUNCATING`, then it returns true (is truncated), which I'm not sure is correct. Is it correct to evaluate to true in the case the stream cuts do not match?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -201,9 +206,25 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+    CompletableFuture<Boolean> isUpdated(String scope, String stream, StreamConfiguration newConfig, OperationContext context) {
+        CompletableFuture<State> stateFuture = streamMetadataStore.getState(scope, stream, true, context, executor);
+        CompletableFuture<StreamConfigurationRecord> configPropertyFuture
+                = streamMetadataStore.getConfigurationRecord(scope, stream, context, executor).thenApply(VersionedMetadata::getObject);
+        return CompletableFuture.allOf(stateFuture, configPropertyFuture)
+                                .thenApply(v -> {
+                                    State state = stateFuture.join();
+                                    StreamConfigurationRecord configProperty = configPropertyFuture.join();
+
+                                    // if property is updating and doesnt match our request, its a subsequent update
```
fpj: `doesnt` -> `doesn't`, `its` -> `it's`

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -201,9 +206,25 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                                    if (configProperty.isUpdating()) {
+                                        return !configProperty.getStreamConfiguration().equals(newConfig);
+                                    } else {
+                                        // if update-barrier is not updating, then update is complete if property matches our expectation 
+                                        // and state is not updating 
+                                        return !(configProperty.getStreamConfiguration().equals(newConfig) && state.equals(State.UPDATING));
```
fpj: The new condition in the description:

```
ConfigProp.config == newConfig && !configProp.isUpdating ==> inProgress
&& state != TRUNCATING // if truncation record is updated but scale is not reset ==> updating
```

does not seem to match the one here. In particular, if `ConfigProp.config != newConfig` (configs differ), then we end up evaluating to true, which I don't understand whether it is correct. The condition in the code is equivalent to:

```
ConfigProp.config != newConfig || !configProp.isUpdating || state != UPDATING ==> updated
```

or 

```
ConfigProp.config == newConfig && configProp.isUpdating && state == UPDATING ==> is updating
```

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -543,19 +586,26 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                                EpochTransitionRecord etr = etrFuture.join();
                                 if (epoch > activeEpoch.getEpoch()) {
                                     response.setStatus(ScaleStatusResponse.ScaleStatus.INVALID_INPUT);
                                 } else if (activeEpoch.getEpoch() == epoch || activeEpoch.getReferenceEpoch() == epoch) {
                                     response.setStatus(ScaleStatusResponse.ScaleStatus.IN_PROGRESS);
                                 } else {
-                                    response.setStatus(ScaleStatusResponse.ScaleStatus.SUCCESS);
+                                    // active epoch == scale epoch + 1 but the state is scaling, the previous workflow 
+                                    // has not completed.
+                                    if (epoch + 1 == activeEpoch.getReferenceEpoch() && state.equals(State.SCALING) &&
```
fpj: This seems ok, although we could have done a better with the description to match to what is in here.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: When we submit a new truncation, we set the truncationRecord to updating with new streamcut in it. 

So first condition to check is `truncationProp.isUpdating` 

case 1 : truncationProp.isUpdating = true 
that means there is an ongoing truncation request. 
If ongoing request doesnt have the same streamCut as that we requested, then our request has definitely completed else our request is definitely incomplete. 

case 2: truncationProp.isUpdating = false
Otherwise ongoing truncation is complete. Stream State may or may not have been reset.
So, if `truncationRecord.streamCut` matches our request and state is truncating, then we completed the truncation but are yet to reset the state. we will return InProgress. 



If there is no ongoing truncationRequest

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: truncation and update have symmetrical checks. the comment above applies here too.

When we submit a new update, we set the ConfigurationRecord to updating with new configuration in it.

So first condition to check is `configurationRecord.isUpdating`

case 1 : configurationRecord.isUpdating = true
that means there is an ongoing update request.
If ongoing request doesnt have the same configuration as that we requested, then our request has definitely completed else our request is definitely incomplete.

case 2: configurationRecord.isUpdating = false
Otherwise ongoing update work is complete. Stream State may or may not have been reset.
So, if configurationRecord.config matches our request and state is UPDATING, then we completed the update but are yet to reset the state. we will return InProgress.

shiveshr: > ConfigProp.config == newConfig && configProp.isUpdating && state == UPDATING ==> is updating
Note: this is an incomplete condition. Update to `configProp` and `state` are not atomic. 
So `configProp.isUpdating` maybe false but state may be updating. 
We want to return InProgress in such cases. That is what this PR is doing, earlier we would return success by only checking `configProp` and not include `state`.

shiveshr: > In particular, if ConfigProp.config != newConfig (configs differ), then we end up evaluating to true, which I don't understand whether it is correct. The condition in the code is equivalent to:

this is because there can only be one outstanding update request. 
So if the outstanding update request does not match the one we submitted then our requested udpate is definitely done

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: The PR description was over simplification and incorrect. Corrected the PR description with exact checks

fpj: I have pretty much the same problem with `IsTruncated`. Focusing on step 2, we have the following:

```
 truncationProp.streamCut == ourStreamCut && state!= TRUNCATING // if truncation record is updated but state is not reset ==> InProgress
    else ==> done
```

while in the code I see:

```
!(truncationRecord.getStreamCut().equals(streamCut) && state.equals(State.TRUNCATING))
```

In more abstract terms, I see in the description:

```
boolean isUpdated = A && !B
```

while the code says:

```
boolean isUpdated = !(A && B)
```

where:
`A`: `truncationRecord.getStreamCut().equals(streamCut)`
`B`: `state.equals(State.TRUNCATING)`

shiveshr: let me first simplify and state in words when a work is done.. then we can tally it with code. 
Done conditions:
1. property.record != ourRecord   // this can only happen if our property update completed and at least one more subsequent update was performed after ours which may or may not have completed yet. 
So our update attempt is done. 
2. if property.record == ourRecord // this means no subsequent update was even attempted. now our update may or may not have completed. 
2.1 If property.isUpdating == true then our update of metadata is not complete. 
2.2 if state == truncating then our metadata update is complete but our state is not reset yet. 

Now lets tally it with the code for `update` and see if above is satisfied or not:
```
if (configProperty.isUpdating()) {
// this tests condition 2.1
    return !configProperty.getStreamConfiguration().equals(newConfig);
} else {
// this checks condition 2.2 and 1
    return !(configProperty.getStreamConfiguration().equals(newConfig) && state.equals(State.UPDATING)); 
```

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: oops sorry again.. 
``` configProp.config == ourConfig && state!= UPDATING // if config record is updated but state is not reset   
 ==> InProgress
 else ==> done
```
there is problem in the description.. the first condition has `state == UPDATING`.. i wrote it correctly in words.. but not in the pseudo code..

PR #3549, opened 2019-03-31
Issue 3547 : Ensure the key version is set to NOT_EXISTS if the key is not present.

### LINKED ISSUE #3547: Table Segment Wire Protocol Bugs
**Problem description**
1. `PravegaRequestProcessor.getTableEntriesCommand`
    - If result is null then the version to send back should be `TableKey.NOT_EXISTS`. It currently is `TableKey.NO_VERSION`.
2. If a Table Key does not exist, the `PravegaRequestProcessor` encodes a zero-length buffer over the wire. The underlying `TableStore` will return `null` in this case. Since a zero-length value is a valid entry, there is no way for a client to differentiate between a key not existing and a key with a zero-length.

**Problem location**
PravegaRequestProcessor and SegmentHelper.

Fix the aforementioned bugs.

shrids:
**Change log description**  
* If the key does not exist during the readTable Operation then the resulting TableEntry should have a version corresponding to `io.pravega.segmentstore.contracts.tables.TableKey#NOT_EXISTS`

**Purpose of the change**  
Fixes #3547 

**What the code does**  
- When a `readTable` operation is performed on the TableStore and if the key does not exist then the resulting TableEntry will have a zero length value and the key version should be TableKey#NOT_EXISTS.
-  Since`TableStore`  does not throw `KeyNotExistsException` for `TableStore#get` , the `SegmentHelper#readTableupdated ` has been updated.

 **How to verify it**  

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
+++ b/controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
@@ -863,7 +863,8 @@ public class SegmentHelper {
      * @param scope               Stream scope.
      * @param stream              Stream name.
      * @param keys                List of {@link TableKey}s to be removed. Only if all the elements in the list has version as
-     *                            {@link KeyVersion#NOT_EXISTS} then an unconditional update/removal is performed. Else an atomic conditional
+     *                            {@link KeyVersion#NO_VERSION} then an unconditional update/removal is performed. Else an atomic
+     *                            conditional
      *                            update (removal) is performed.
      * @param hostControllerStore Host controller store.
      * @param clientCF            Client connection factory.
@@ -961,6 +962,7 @@ public class SegmentHelper {
      * a value corresponding to the latest version. If the operation failed, the future will be failed with the
      * causing exception. If the exception can be retried then the future will be failed with
      * {@link WireCommandFailedException}.
+     * Note: TableKeyDoesNotExist is not thrown by the readTable Command.
      */
     public CompletableFuture<List<TableEntry<byte[], byte[]>>> readTable(final String scope,
                                                                          final String stream,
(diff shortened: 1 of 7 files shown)


controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1000,22 +1000,11 @@ public void noSuchSegment(WireCommands.NoSuchSegment noSuchSegment) {
...
-                                                                                     new TableKeyImpl<>(getArray(k.getData()),
-                                                                                                        new KeyVersionImpl(k.getKeyVersion()));
-                                                                             return new TableEntryImpl<>(tableKey, getArray(e.getValue().getData()));
-                                                                         }).collect(Collectors.toList());
+                                                                         .map(e -> new TableEntryImpl<>(convertFromWireCommand(e.getKey()), getArray(e.getValue().getData())))
+                                                                         .collect(Collectors.toList());
                 result.complete(tableEntries);
             }
 
-            @Override
```
shiveshr: why are you removing this? table key does not exist is an expected response that should be handled. right?

shrids: `io.pravega.shared.protocol.netty.FailingReplyProcessor#tableKeyDoesNotExist`  call back is invoked when `WireCommands.TableKeyDoesNotExist` is received by the client. This Wirecommand is sent by the PRP(PravegaRequestProcessor) incase `io.pravega.segmentstore.contracts.tables.KeyNotExistsException` is thrown by the TableStore apis. 
Now, `io.pravega.segmentstore.contracts.tables.TableStore#get` does not throw `KeyNotExistsException` incase the key does not exist, it returns a null. The PRP now returns an Entry where the KeyVersion corresponds to `TableKey.NOT_EXISTS` if the key does not exist.

shiveshr: it is still a good idea to keep that response available, in case something changes on segment store in future. because that can be a legitimate reponse to read queries.. right?

Also, can we have the same issue with readTableEntries?

shrids: The default implementation for `ReplyProcessor#tableKeyDoesNotExist` is to throw an `IllegalStateException`.
Updated the code to remove redundant implementation of `tableKeyDoesNotExist` in `readTableKeys(..)` and `readTableEntries(..)`.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -863,7 +863,8 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
      * @param scope               Stream scope.
      * @param stream              Stream name.
      * @param keys                List of {@link TableKey}s to be removed. Only if all the elements in the list has version as
-     *                            {@link KeyVersion#NOT_EXISTS} then an unconditional update/removal is performed. Else an atomic conditional
+     *                            {@link KeyVersion#NO_VERSION} then an unconditional update/removal is performed. Else an atomic
+     *                            conditional
```
andreipaduroiu: Please fix the Javadoc indentation on these two lines.

shrids: done.

PR #3553, opened 2019-04-01
Issue 3550: Correct maximum event size

### LINKED ISSUE #3550: Writing event of exactly 1 MB
**Problem description**
After updating to v0.4.0 it is impossible to write event size of exactly 1048576 bytes (1MB), only 1048568 bytes (1MB - 8bytes). Exception: "Write size too large"
It is caused by adding header to every event, which takes 8 bytes.

**Problem location**
PendingEvent.java

Change condition check into "data.readableBytes() <= MAX_WRITE_SIZE + 8"  OR change "MAX_WRITE_SIZE" into "1024*1024+8"

akurilov:
**Change log description**  
Allow max event size payload = 1MB

**Purpose of the change**  
Fixes #3550
Pravega v0.4.0 brakes the compatibility. It was possible to write events with size up to 1MB (including this value). Since v0.4.0 the upper limit is 8 bytes less due to the additional headers introduced to distinguish byte streams and event streams. 

**What the code does**  
Change the max event size to 1MB + 8 bytes

**How to verify it**  
Try to write an event with payload size = 1MB

### DIFF
--- a/test/system/src/test/java/io/pravega/test/system/MetadataScalabilityTest.java
+++ b/test/system/src/test/java/io/pravega/test/system/MetadataScalabilityTest.java
@@ -46,7 +46,6 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 @Slf4j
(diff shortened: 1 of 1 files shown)


fpj:
hi @akurilov, thanks for your contribution. All checks are currently failing. You need to make sure to:

- Sign all commits, otherwise the DCO bot is going to complain
- Check the travis build to see why the build is failing

client/src/main/java/io/pravega/client/stream/impl/PendingEvent.java
```diff
@@ -24,7 +24,11 @@
  */
 @Data
 public class PendingEvent {
-    public static final int MAX_WRITE_SIZE = Serializer.MAX_EVENT_SIZE;
+
+    /**
+     * The wire protocol adds extra 8 bytes to the event payload
```
tkaitchuck: Checkstyle is failing because this sentence must end in a "."

PR #3555, opened 2019-04-01
Issue 3554: Add Zookeeper environment variable for Controller service definition in Docker Compose Configuration file

### LINKED ISSUE #3554: Controller fails to connect to Zookeeper in Docker Compose based deployment
**Problem description**
`ZK_URL` environment variable is missing from the `controller` service definition in `docker-compose.yml` file. As a result, the controller fails to bootstrap properly, and the deployment becomes unusable. 

To reproduce the issue: 
1. Clone Pravega into a VM that has Docker and Docker Compose installed: `git clone https://github.com/pravega/pravega.git`
2. Execute command to navigate to directory containing the `docker-compose.yml file`: `cd ./pravega/docker/compose`
3. Execute this command to set an environment variable:  `export HOST_IP=<host-ip>` (where `<host-ip>` is the IP address of the host in which you are deploying Pravega.)
4. Execute this command to start the containers: `docker-compose up -d`
5. Verify that the containers are up (check the `state` column): `docker-compose ps`
6. Run a client sample application against Controller URI: `tcp://<host-ip>:9090/`. Observe the failures. 
7. Now, run `docker-compose logs`. Observe an error like the following. Notice that the controller is trying to locate Zookeeper at `127.0.0.1:2181`, which is incorrect (should have been the `<host-ip>:2181`. 
```
controller_1    | 2019-04-01 11:14:25,150 31781 [ControllerServiceMain-SendThread(localhost:2181)] INFO  org.apache.zookeeper.ClientCnxn - Opening socket connection to server localhost/127.0.0.1:2181. Will not attempt to authenticate using SASL (unknown error)
controller_1    | 2019-04-01 11:14:25,152 31783 [ControllerServiceMain-SendThread(localhost:2181)] WARN  org.apache.zookeeper.ClientCnxn - Session 0x0 for server null, unexpected error, closing socket connection and attempting reconnect
controller_1    | java.net.ConnectException: Connection refused
controller_1    |       at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
controller_1    |       at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:717)
controller_1    |       at org.apache.zookeeper.ClientCnxnSocketNIO.doTransport(ClientCnxnSocketNIO.java:357)
controller_1    |       at org.apache.zookeeper.ClientCnxn$SendThread.run(ClientCnxn.java:1214)
```

**Problem location**
`/path/to/pravega/docker/compose/docker-compose.yml` file. 

Add an environment variable `ZK_URL: zookeeper:2181` in the controller service definition section in the `docker-compose.yml` file.

ravisharda:
Signed-off-by: Ravi Sharda <ravi.sharda@emc.com>

**Change log description**  
In the Docker Compose configuration file, add ZK_URL as env. variable so that the Controller is linked to the appropriate Zookeeper container. 

**Purpose of the change**  
Fixes #3554. 

**What the code does**  
Adds `ZK_URL` environment variable to the Controller service definition in the Docker Compose application configuration file `docker-compose.yml`.

**How to verify it**  
Here are the steps to verify the fix:

1. Clone Pravega into a VM that has Docker and Docker Compose installed: `git clone https://github.com/pravega/pravega.git`
2. Execute command to navigate to the directory containing the docker-compose.yml file: `cd ./pravega/docker/compose``
3. Execute this command to set an environment variable:  `export HOST_IP=<host-ip>` (where `<host-ip>` is the IP address of the host in which you are deploying Pravega).
4. Execute this command to start the containers: `docker-compose up -d`
5. Verify that the containers are up (check the `state` column): `docker-compose ps`
6. Run a reader/writer client application against Controller URI: `tcp://<host-ip>:9090/`. The client should work correctly with the specified Controller URI. 
7. To be doubly sure, check out the container logs: `docker-compose logs`

### DIFF
--- a/docker/compose/docker-compose.yml
+++ b/docker/compose/docker-compose.yml
@@ -70,7 +70,6 @@ services:
       JAVA_OPTS: |
         -Dcontroller.service.port=9090
         -Dcontroller.service.restPort=10080
-        -Dcontroller.zk.url=zookeeper:2181
         -Dconfig.controller.metricenableCSVReporter=false
         -Xmx512m
         -XX:OnError="kill -9 p%"
--- a/docker/compose/docker-compose.yml
+++ b/docker/compose/docker-compose.yml
@@ -66,6 +66,7 @@ services:
     command: controller
     environment:
       WAIT_FOR: zookeeper:2181
+      ZK_URL: zookeeper:2181
       JAVA_OPTS: |
         -Dcontroller.service.port=9090
         -Dcontroller.service.restPort=10080
--- a/docker/compose/docker-compose-extendeds3.yml
+++ b/docker/compose/docker-compose-extendeds3.yml
@@ -52,10 +52,10 @@ services:
     command: controller
     environment:
       WAIT_FOR: zookeeper:2181
+      ZK_URL: zookeeper:2181
       JAVA_OPTS: |
         -Dcontroller.service.port=9090
         -Dcontroller.service.restPort=10080
-        -Dcontroller.zk.url=zookeeper:2181
         -Dconfig.controller.metricenableCSVReporter=false
         -Xmx512m
         -XX:OnError="kill -9 p%"
--- a/docker/compose/docker-compose-nfs-mount.yml
+++ b/docker/compose/docker-compose-nfs-mount.yml
@@ -52,10 +52,10 @@ services:
     command: controller
     environment:
       WAIT_FOR: zookeeper:2181
+      ZK_URL: zookeeper:2181
       JAVA_OPTS: |
         -Dcontroller.service.port=9090
         -Dcontroller.service.restPort=10080
-        -Dcontroller.zk.url=zookeeper:2181
         -Dconfig.controller.metricenableCSVReporter=false
         -Xmx512m
         -XX:OnError="kill -9 p%"
--- a/docker/compose/docker-compose-nfs.yml
+++ b/docker/compose/docker-compose-nfs.yml
@@ -59,10 +59,10 @@ services:
     command: controller
     environment:
       WAIT_FOR: zookeeper:2181
+      ZK_URL: zookeeper:2181
       JAVA_OPTS: |
         -Dcontroller.service.port=9090
         -Dcontroller.service.restPort=10080
-        -Dcontroller.zk.url=zookeeper:2181
         -Dconfig.controller.metricenableCSVReporter=false
         -Xmx512m
         -XX:OnError="kill -9 p%"


docker/compose/docker-compose.yml
```diff
@@ -66,6 +66,7 @@ services:
     command: controller
     environment:
       WAIT_FOR: zookeeper:2181
+      ZK_URL: zookeeper:2181
```
fpj: This looks ok, but I'm wondering about the `zookeeper` host name. Is it correct?

docker/compose/docker-compose.yml
```diff
@@ -66,6 +66,7 @@ services:
...
       JAVA_OPTS: |
         -Dcontroller.service.port=9090
         -Dcontroller.service.restPort=10080
```
adrianmo: Since you are setting the ZK endpoint on `ZK_URL`, you can remove the `-Dcontroller.zk.url=zookeeper:2181` parameter sent on the `JAVA_OPTS`.

docker/compose/docker-compose.yml
adrianmo: Yes, @fpj, the `zookeeper` host name is configured a few lines below in the `links` block, making the `zookeeper` host name available to the Controller.

docker/compose/docker-compose.yml
ravisharda: Sure, thanks. I've pushed a commit with the suggested change.

docker/compose/docker-compose.yml
ravisharda: To add a bit more color, `zookeeper` is the name of the service defined later in the docker-compose.yml file. That service is linked to the Controller service via the `controller` service's `links` option. Since containers for linked services are reachable at a hostname identical to the alias or the service name, the hostname for zookeeper is indeed `zookeeper`.

ravisharda:
> @ravisharda the change looks good.
> 
> I've added a comment below. Also, there are other Compose files in the same directory that have the same issue. Can you apply the fix to them as well? Thanks.

Thanks, @adrianmo. I've applied the fix to the rest of the Compose files as well, as you suggested.

docker/compose/docker-compose.yml
ravisharda: To verify the discussion about the hostname, I ran a few commands to check the `hostname` of the Zookeeper container on a deployment. (The name of the container was `compose_zookeeper_1`)

```
 (Run the hostname command inside the Zookeeper container)
..:~/..#  docker exec -it compose_zookeeper_1 hostname
99665322b64f
```
The above shows that the hostname visible inside the container is `99665322b64f`. 

Then, I ran the following commands: 
```
(Start a shell inside another container, so that we are in a container which shares the same  
network as the Zookeeper container.)
# docker exec -it compose_controller_1 sh

(Now, ping `zookeeper`. If the host exists, we should see a typical ping response. ) 
# ping zookeeper
PING zookeeper (172.20.0.2): 56 data bytes
64 bytes from 172.20.0.2: seq=0 ttl=64 time=0.067 ms
64 bytes from 172.20.0.2: seq=1 ttl=64 time=0.204 ms
```
So, our assertions are correct.

PR #3591, opened 2019-04-09
Issue 3590: Improve the latency of small size events

### LINKED ISSUE #3590: Improve the latency for small message sizes
**Problem description**
Currently for 100 bytes event size, the average latency is 8 ms observed in white cluster (lag unitas).
event after this PR : https://github.com/pravega/pravega/pull/3482 
the performance improvement is observed only in 1K and above size messages.

**Problem location**
The problem location is pravega client. the pravega client batch size needs to be tuned so that
for small size message there will be improvement in the latencies.

Actually, the batch size can be set as  producer property; so that user application can configure it according to the event size requirements.

kmgowda:
**Change log description**  
The MAX_BATCH_SIZE is set to 8K at the pravega client to reduce the latency of small size events.

**Purpose of the change**  
fixes #3590 

**What the code does**  
The pravega client accumulates upto 8K (earlier it was 32K) and sends the messages/events to pravega cluster. with 32K , the median latency was 8ms ; by setting it to 8K , the median latency is 6ms.

**How to verify it**  
The Benchmark tests are executed to confirm the latency improvement by this change.

Signed-off-by: Keshava Munegowda <keshava.munegowda@dell.com>

### DIFF
--- a/client/src/main/java/io/pravega/client/netty/impl/AppendBatchSizeTrackerImpl.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/AppendBatchSizeTrackerImpl.java
@@ -30,7 +30,7 @@ import java.util.function.Supplier;
  */
 class AppendBatchSizeTrackerImpl implements AppendBatchSizeTracker {
     private static final int MAX_BATCH_TIME_MILLIS = 100;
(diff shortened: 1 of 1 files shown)


kmgowda:
Ok, I will close this PR. but, issue 3590 remains as it is.

PR #3594, opened 2019-04-09
Issue 3593: Removed the duplication from metric names.

### LINKED ISSUE #3593: Get rid of duplication in metric names
**Problem description**
These are quite come metrics containing duplicated segments, such as:
pravega_controller_controller_stream_created_latency_ms
pravega_controller_controller_transactions_committed_latency_ms
pravega_segmentstore_segmentstore_container_operation_commit_latency_ms
pravega_segmentstore_segmentstore_segment_create_latency_ms
pravega_segmentstore_segmentstore_tablesegment_get_latency_ms
...
These metrics names occupy more disk space, increase network traffic, and it's hard on the back-end monitoring system to display the complete name.

**Problem location**
MetricsNames.java

Remove the duplicated segment from those metrics.

kevinhan88:

**Change log description**  
Removed the duplicated part from histogram metric names: 
e.g. segmentstore.segmentstore.segment.write_latency_ms is changed to segmentstore.segment.write_latency_ms.

**Purpose of the change**  
Fixes #3593 

**What the code does**  
Changed histogram metric names inside MetricsNames.java by removing "segmentstore" or "controller" if the metric's context already contains "segmentstore" or "controller".

**How to verify it**  
Metrics back-end should stop seeing metrics name contains "segmentstore.segmentstore" or "controller.controller".


shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -95,14 +95,14 @@
     public static final String CACHE_GET_LATENCY = "segmentstore.cache.get_latency_ms";         // Histogram
```
RaulGracia: I think that this should be fixed as well, eight? Please, double check all the fields with a `Histogram` comment.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -95,14 +95,14 @@
...
     public static final String CACHE_READ_BYTES = "segmentstore.cache.read_bytes";              // Counter
     public static final String CACHE_TOTAL_SIZE_BYTES = "segmentstore.cache.size_bytes";        // Gauge
-    public static final String CACHE_GENERATION_SPREAD = "segmentstore.cache.gen";              // Histogram
+    public static final String CACHE_GENERATION_SPREAD = "cache.gen";              // Histogram
 
     // DurableDataLog (Tier1) stats
-    public static final String BK_TOTAL_WRITE_LATENCY = "segmentstore.bookkeeper.total_write_latency_ms";   // Including Queue. Per-container Histogram.
-    public static final String BK_WRITE_LATENCY = "segmentstore.bookkeeper.write_latency_ms";               // Exclusively the write to BK. Per-container Histogram.
+    public static final String BK_TOTAL_WRITE_LATENCY = "total_write_latency_ms";   // Including Queue. Per-container Histogram.
```
RaulGracia: Here (and below) you are removing `bookkeeper` too. Is the tag `bookkeeper` added by the histogram? Otherwise, we may confuse this with the write latency of the segment store.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -119,52 +119,52 @@
...
-    public static final String THREAD_POOL_QUEUE_SIZE = "segmentstore.thread_pool.queue_size";          // Histogram
-    public static final String THREAD_POOL_ACTIVE_THREADS = "segmentstore.thread_pool.active_threads";  // Histogram
+    public static final String THREAD_POOL_QUEUE_SIZE = "thread_pool.queue_size";          // Histogram
+    public static final String THREAD_POOL_ACTIVE_THREADS = "thread_pool.active_threads";  // Histogram
 
     // Metrics in Controller
     // Stream request counts
     public static final String CREATE_STREAM = "controller.stream.created";                         // Counter
-    public static final String CREATE_STREAM_LATENCY = "controller.stream.created_latency_ms";      // Histogram
+    public static final String CREATE_STREAM_LATENCY = "stream.created_latency_ms";      // Histogram
```
RaulGracia: Minor: indentation of comments.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -119,52 +119,52 @@
...
-    public static final String COMMIT_TRANSACTION_LATENCY = "controller.transactions.committed_latency_ms"; // Histogram
+    public static final String COMMIT_TRANSACTION_LATENCY = "transactions.committed_latency_ms"; // Histogram
     public static final String COMMIT_TRANSACTION_FAILED = "controller.transactions.commit_failed";         // Counter, Per-stream Counter, Per-transaction Counter
     public static final String ABORT_TRANSACTION = "controller.transactions.aborted";                       // Counter and Per-stream Counter
-    public static final String ABORT_TRANSACTION_LATENCY = "controller.transactions.aborted_latency_ms";    // Histogram
+    public static final String ABORT_TRANSACTION_LATENCY = "transactions.aborted_latency_ms";    // Histogram
     public static final String ABORT_TRANSACTION_FAILED = "controller.transactions.abort_failed";           // Counter, Per-stream Counter, Per-transaction Counter
     public static final String OPEN_TRANSACTIONS = "controller.transactions.opened";                        // Per-stream Gauge
     public static final String TIMEDOUT_TRANSACTIONS = "controller.transactions.timedout";                  // Per-stream Counter
```
shiveshr: same comment as @RaulGracia about removing `controller.` prefix and segmentstore prefix from all names.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: This belongs to RocksDB which reads as "rocksdb.segmentstore.cache.get_latency_ms", there is no duplication.

kevinhan88:
@shiveshr @RaulGracia Yes this is a request from the Platform team complaining the metrics name with duplication are too long, cannot easily fit in screen. Because they are making changes to adopt the micrometer migration, so they want to have this name change to be done together.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: This is the ultimate change with the name, please let me know if there's still concern:
before: bookkeeper.segmentstore.bookkeeper.total_write_latency_ms
after: bookkeeper.write_latency_ms.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: @shiveshr only those metrics belong to a StatsLogger may have duplication in names, because the loggerName and metric name may both contain things such as "controller" or "segmentstore".
The dynamic metrics don't have statsLogger prefix, so it is essential to have "controller"/"segmentstore" inside names to indicate the origin of the metrics.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: fixed.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -95,14 +95,14 @@
...
+    public static final String BK_TOTAL_WRITE_LATENCY = "total_write_latency_ms";                           // Including Queue. Per-container Histogram.
+    public static final String BK_WRITE_LATENCY = "write_latency_ms";                                       // Exclusively the write to BK. Per-container Histogram.
     public static final String BK_WRITE_BYTES = "segmentstore.bookkeeper.write_bytes";                      // Counter
-    public static final String BK_WRITE_QUEUE_SIZE = "segmentstore.bookkeeper.write_queue_size";            // Per-container Histogram
-    public static final String BK_WRITE_QUEUE_FILL_RATE = "segmentstore.bookkeeper.write_queue_fill";       // Per-container Histogram
+    public static final String BK_WRITE_QUEUE_SIZE = "write_queue_size";                                    // Per-container Histogram
+    public static final String BK_WRITE_QUEUE_FILL_RATE = "write_queue_fill";                               // Per-container Histogram
     public static final String BK_LEDGER_COUNT = "segmentstore.bookkeeper.bookkeeper_ledger_count";         // Per-container Gauge
```
jkhalack: This seems to be an overkill too (it works out to`pravega_segmentstore_bookkeeper_bookkeeper_ledger_count` in InfluxDB). 
Could you remove one of `bookkeeper` strings, please?

kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -95,14 +95,14 @@
     public static final String CACHE_GET_LATENCY = "segmentstore.cache.get_latency_ms";         // Histogram
     public static final String CACHE_READ_BYTES = "segmentstore.cache.read_bytes";              // Counter
     public static final String CACHE_TOTAL_SIZE_BYTES = "segmentstore.cache.size_bytes";        // Gauge
-    public static final String CACHE_GENERATION_SPREAD = "segmentstore.cache.gen";              // Histogram
+    public static final String CACHE_GENERATION_SPREAD = "cache.gen";                           // Histogram
 
     // DurableDataLog (Tier1) stats
-    public static final String BK_TOTAL_WRITE_LATENCY = "segmentstore.bookkeeper.total_write_latency_ms";   // Including Queue. Per-container Histogram.
```
jkhalack: This metric is currently reported to InfluxDB as `pravega_bookkeeper_segmentstore_bookkeeper_write_latency_ms`,
and it is bookkeeper client metric, if I am not mistaken.

Are you sure that removing `segmentstore` from here will not lead to confusion with Bookkeeper server metrics if they were to be saved to the same database?

kevinhan88: Agree, only the repeating bookkeeper is removed.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
EronWright: Is the goal to reduce "duplication" or to improve "consistency".    This PR certainly doesn't look very consistent.

EronWright:
The change seems to be adhoc as though there is no governing principle behind it.     I fear that this PR is making things worse.

kevinhan88:
@EronWright histogram metrics names are in the format of: systemPrefix(such as "pravega") + statsLoggerName(e.g. "controller", "segmentstore") + metricName.
Inside MetricsNames.java, where neither system prefix nor logger prefix appear, some metric names may indeed contain repeating element, without thinking how the ultimate metric name looks. 
Here we only remove the obvious repeating element from the ultimate name, to minimum the impact. As you mentioned the metric names are actually the contract of the Pravega metrics so this PR is per the request from your team trying to help the metric migration effort - thoughts?

jkhalack:
@kevinhan88  It might make sense to group the metrics by the prefix they will be getting, something like: 
- Segmentstore metrics without statsLogger
- Segmentstore metrics with `segmentstore` statsLogger
- Segmentstore metrics with `bookkeeper` statsLogger
- Segmentstore metrics with `filesystem` statsLogger
- Segmentstore metrics with `hdfs` statsLogger
...
- Controller metrics without statsLogger
- Controller metrics with `controller` statsLogger

(or whatever suits Pravega logic better, just make it understood where the prefixes come from).

kevinhan88:
I added the logger name into the comments of those histogram metrics.
Also added the naming convention. The description of this PR is also updated with the naming convention to make it easier to review.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -83,27 +90,27 @@
...
-    public static final String BK_WRITE_LATENCY = "segmentstore.bookkeeper.write_latency_ms";               // Exclusively the write to BK. Per-container Histogram.
-    public static final String BK_WRITE_BYTES = "segmentstore.bookkeeper.write_bytes";                      // Counter
-    public static final String BK_WRITE_QUEUE_SIZE = "segmentstore.bookkeeper.write_queue_size";            // Per-container Histogram
-    public static final String BK_WRITE_QUEUE_FILL_RATE = "segmentstore.bookkeeper.write_queue_fill";       // Per-container Histogram
-    public static final String BK_LEDGER_COUNT = "segmentstore.bookkeeper.bookkeeper_ledger_count";         // Per-container Gauge
+    public static final String BK_TOTAL_WRITE_LATENCY = "segmentstore.bookkeeper.total_write_latency_ms";              // Including Queue. Per-container Histogram. Logger: "bookkeeper"
```
jkhalack: So this will keep the double `bookkeeper` string?
As in  `bookkeeper_segmentstore_bookkeeper`?

jkhalack: I would understand `bookkeeper_segmentstore` or `segmentstore_bookkeeper`, but why `bookkeeper_segmentstore_bookkeeper`?

RaulGracia:
As I see this PR, it looks that we are trying to adapt metric names to an underlying inconsistency in the way of treating prefixes. In line with @EronWright comments, we need a consistent way of defining the name and prefixes for metrics. Otherwise, this will be error prone and will be make it worse as the more metrics are added to Pravega.

It would be great to agree on a solution at least on the following topics:
- How to add _component prefixes_ to metrics?
- How to _keep order_ in the metric name structure?
- How to add the _base prefix_ to all metrics?

1) How to add _component prefixes_ to metrics:

Inspecting the code, it seems that the root cause for the underlying issue is the following. When we create a logger, we have the following options:
```
@Override
public DynamicLogger createDynamicLogger() {
     init();
     return new DynamicLoggerImpl(conf, metrics, new StatsLoggerImpl(getMetrics(), "pravega"));
}
@Override
public StatsLogger createStatsLogger(String name) {
     init();
     return new StatsLoggerImpl(getMetrics(), "pravega." + name);
}
```
On the one hand, when we create a `DynamicLogger` instance, we do not add any "prefix" to it other than the base one ("pravega"). This requires us to add the desired "component prefix" in the metric name itself. An example of this is, for instance, `ACTIVE_SEGMENT_COUNT` that directly interacts with a `DynamicLogger` object and the name of the metric should contain the component prefix (i.e., "segmentstore").

On the other hand, we have those classes that instantiate a `StatsLogger`, whose constructor takes as input a String that is actually used as prefix (namely `basename` in `StatsLoggerImpl`). Internally, `StatsLoggerImpl` is using this prefix along with the metric name (like for Histograms). This seems to be the root cause for the naming inconsistencies we are witnessing in `MetricsNames.java`.

In my humble opinion, we need to keep a consistent prefix behavior across all the `XXXLogger` classes. That is, either we change `DynamicLogger` to add a component prefix to metrics or we remove the use of prefixes in `StatsLogger`. I don't have a strong preference here, so I leave to the platform team and the author of the PR the solution that is more convenient to people dealing with metrics names. By doing this, there will be a single way of defining metrics names (with or without prefix), irrespective of the metric type.

2) How to keep order in the metric name structure?

I see some metric in which the metric naming structure is changed. For instance:
i) Metric `bookkeeper.total_write_latency_ms` is a Segment Store metric, but it is not incorporating the `segmentstore` prefix with the current change.
ii) Metric `rocksdb.segmentstore.cache.get_latency_ms` is placing first the "sub-component" (RocksDB) to the component it belongs to ("Segment Store").

We need to define a consistent way of structuring metric names (I think that has been already defined) and follow it. In my humble opinion, as the previous examples are Segment Store metrics, both should have  `segmentstore` as the component name first, followed by `bookkeeper` and `rocksdb`, respectively. 

3) How to add the _base prefix_ to all metrics?

In the code snippet above, you can observe that the base prefix for metrics (i.e., "pravega") is hard-coded. I think that is not good, specially when dealing with multiple Pravega deployments that are publishing metrics to the same metrics back-end. For consistency, I would suggest to replace the "pravega" String hardcoded in the `createDynamicLogger()` and `createStatsLogger()` methods by the `pravegaservice.clusterName` property available in the Pravega configuration.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -12,11 +12,16 @@
...
- * Class that defines all names used for metrics that Pravega exports. All metrics have a prefix that indicates
- * the component that is exporting it. We use a second level to indicate the sub-component (e.g., storage) or
- * abstraction (e.g., transactions) the metric is related to.
+ * Class that defines all names used for metrics that Pravega exports.
+ * All metric names are in the format of
+ * "Metrics Prefix + Component Origin + Sub-Component (or Abstraction) + Metric Base Name":
  *
- * The current prefix being used are:
+ *     Metric Prefix: "pravega" by default, configurable.
+ *     Component Origin: indicate which component generates the metric, such as "segmentstore", "controller".
```
RaulGracia: `indicate` -> `indicates`.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -30,22 +35,12 @@
  * - controller.hosts: metrics related to Pravega servers in the cluster (e.g., number of servers, failures)
  * - controller.container: metrics related to container lifecycle (e.g., failovers)
  *
- * We have two types of metrics:
```
RaulGracia: Why has this been deleted?

kevinhan88: It was mentioned in the past because metrics were named according to the different type. Metric names are now consistent regardless of the type.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: fixed.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
RaulGracia: We are still using `_global` metrics, right? In fact, this is written in the paragraph below: `global version of it by adding "_global" suffix`. These two paragraphs are just elaborating on why we have metrics with an without `_global` suffix. Still, I don't see the need for removing this explanation, as it is linked with the paragraph below.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
kevinhan88: Yes, added back.

kevinhan88:
There are two questions remaining: 
1. It's reporeted that the txn segment not being reflected to "parent" segment. Local test showed its working as expected, screenshots have been updated in the slack channel. Will work with the Platform team to figure out the root cause on their end. 
2. Table segments metrics are showing differently than other "regular" segments. This is because the name structure of table segment is different. PR https://github.com/pravega/pravega/pull/3629 is pending to add the missing scope. Once PR 3629 is merged, the problem should be fixed automatically.

RaulGracia:
Thanks @kevinhan88. As I understand, this PR depends on #3629 to define the new naming structure for internal Table Segments. Meanwhile, before approving this PR I would like to see some test coverage to verify that the `segmentTags` method is dealing well with the naming structure proposed in #3629. This can be done in parallel to speed up things.

shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -36,6 +36,12 @@
      */
     private static final int TRANSACTION_ID_LENGTH = 2 * TRANSACTION_PART_LENGTH;
 
+    // For table segment name parsing, will depend on StreamSegmentNameUtils.
```
RaulGracia: I don't know if this is the place to set these constants. Please, keep an eye to PR #3629 and consider using (or even extending) the logic defined at `StreamSegmentNameUtils.java`. We need to avoid duplicating constants and naming patterns that already defined in one place.

kevinhan88:
It's been raised a couple of times that Metrics should depends on StreamSegmentNameUtil.java to avoid internal stream/segment definition exposed. Please note metrics does not have access to the class with the current dependency relationship.
An issue is created to enhance the dependency relationship: https://github.com/pravega/pravega/issues/3631

kevinhan88:
Branch build test passed: https://asd-nautilus-jenkins.isus.emc.com/view/Pravega/job/pravega-branch-builder/1312/

RaulGracia:
Thanks @kevinhan88, at this point I will approve the PR as it is. I miss an integration test to validate the changes from PR #3629, but this can be done as a separate issue to speed up things. I created an issue to keep track of this: #3634.

PR #3603, opened 2019-04-10
Issue 3597: Change bookies health check command

### LINKED ISSUE #3597: Docker swarm tests are failing to start bookkeeper services
**Problem description**
bookies are getting stopped after a few seconds with error 
```
"Err": "task: non-zero exit (137): dockerexec: unhealthy container",
```
**Problem location**
Docker swarm system test framework

Command used to check bookie health check has to be changed. Also, need to increase max attempts of bookie restart.

anishakj:
Signed-off-by: anishakj <anisha.kj@dell.com>

**Change log description**  
1. changed docker client version to 8.15.1
2. Increased segment store memory to 2000m
3. in docker client update service(), service spec is updated to pass networks.
4. updated DOCKER_API_MIN_VERSION as 1.37
5. Changed the default health check command of bookie
6. Increased the max  restart attempts to 1 and changed the restart policy 

**Purpose of the change**  
Fixes #3597

**What the code does**  
we were facing "no route to host" exceptions with docker swarm tests. This was because task termination order was not deterministic in the docker versions we were using. upgraded the docker versions.
Framework with the updated docker client jar is not able to scale the services. Updated the service spec with networks() to solve.it
Increased segment store memory provisioning to eliminate out of memory errors.
Deterministic task termination order is supported from Docker API version 1.37
with the bookkeeper image change, ```netstat  ``` command is not available in BK container, changed the command to use `ss` tool. Also there was a race when bookies are tries to start simultaneously, increased the max attempts to restart as `1`.

**How to verify it**  
Verified by running system tests.


test/system/src/main/java/io/pravega/test/system/framework/services/docker/BookkeeperDockerService.java
```diff
@@ -81,7 +82,7 @@ private ServiceSpec setServiceSpec() {
         stringList.add(env4);
 
         final TaskSpec taskSpec = TaskSpec
-                .builder().restartPolicy(RestartPolicy.builder().maxAttempts(0).condition("none").build())
+                .builder().restartPolicy(RestartPolicy.builder().maxAttempts(1).condition(RESTART_POLICY_ANY).build())
```
shrids: I have a couple of clarification questions.
- Could provide more information about the "any" restart policy v/s "on_failure"?
- Does setting maxAttempts imply it is restarted only on the first failure?
- Currently in the K8s based test framework a bookie / or any other pravega container restarts would lead to a test failure. Would the behavior be the same here ?

anishakj: "on-failure" will restart the container only if it exited with error."any" will restart the container , even if the container exited with no errors. In our bookie tests, one of the bookie was got into running state and others were shutting down without any errors.

Max attempts will restart the container only on the first failure.

I am not sure whether the restart of container cause test failure.

RaulGracia: In a conversation with @adrianmo, it looks like this change is necessary due to the following reason:
1) With the new version of Bookkeeper, if all the bookies are started at once they will try to acquire the Zookeeper lock to create the necessary metadata.
2) Only one of the Bookies will succeed on doing so, and the rest of Bookies just terminate instead of waiting until the metadata is initialized. 
3) The expectation for the rest of Bookies is that once they are restarted, the metadata will be initialized by the Bookie that acquired the lock in the first place.
Is this correct @anishakj? Does this clarifies the motivation for this change @shrids?

shrids: yes it does. Thanks @RaulGracia @anishakj

PR #3615, opened 2019-04-11
Issue 3537:  Fix backward compatibility issue with WireCommands.ConditionalAppend

### LINKED ISSUE #3537: Ensure backward Compatability of WireCommands.ConditionalAppend.
**Problem description**
`WireCommands.ConditionalAppend` should be backward compatible with older versions of the client.
Pravega clients before commit id `8e96dc1ca5364e5164846736b69006a1312291d0` will not be able to decode the latest version of `WireCommands.ConditionalAppend`

**Problem location**
WireCommands.java

Fix the check @ https://github.com/pravega/pravega/blob/master/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java#L639

shrids:
**Change log description**  
Fix backward compatibility issue with `WireCommands.ConditionalAppend`

**Purpose of the change**  
Fixes #3537 

**What the code does**  
Ensure Pravega is backward compatible with 0.4.x version. This fixes the issue which prevented users from using 0.4.x version of Pravega client with the 0.5.x version.

*Details:*
In the version 8 of the `WireCommands` a new field of type `Long` has been added to `ConditionalAppend`.

The fields of ConditionalAppend (version 8) are 
```
final UUID writerId;
final long eventNumber;
final long expectedOffset;
final Event event;
final long requestId; // added in version 8.
```

Now when a `ConditionalAppend` corresponding to version 7 or lower is sent over wire to Pravega( with version 8) then  `ConditionalAppend.readFrom(Stream in, length)` method is invoked.
This  which internally invokes
```Event.readFrom(in, length - 5 * Long.Bytes). ```

Now, `Event.readFrom()` method has a check if the `Event.length` is exactly equal to `length` passed (current value is length - 5 * Long.Bytes).
This check will fail if a `ConditionalAppend` with version 7 or lower is sent because length of an Event matches `length - 4 * Long.Bytes.`

To solve this the PR has changed the check in Event.readFrom() to verify if the encoded Event length is `less than or equal` to the length of the byte stream, thereby ensuring `ConditionalAppend` is backward compatible.

**How to verify it**  
The existing and newly added tests should continue to pass.

### CODE REVIEW

File: shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java [hunk 1]
```diff
@@ -486,7 +486,7 @@ public static Event readFrom(DataInput in, int length) throws IOException {
                 throw new InvalidMessageException("Was expecting EVENT but found: " + typeCode);
             }
             int eventLength = in.readInt();
-            if (eventLength != length - TYPE_PLUS_LENGTH_SIZE) {
+            if (eventLength > length - TYPE_PLUS_LENGTH_SIZE) {
```
fpj: This is making the assumption that if the event length is greater than the expected length, then it is valid. That's not correct. Instead, the logic should be such that:

```
if (version == x) and (event length == expected length of version x)
then ok

if (version == y) and (event length == expected length of version y)
then ok

else throw
```

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -636,7 +636,7 @@ public static WireCommand readFrom(ByteBufInputStream in, int length) throws IOE
             UUID writerId = new UUID(in.readLong(), in.readLong());
             long eventNumber = in.readLong();
             long expectedOffset = in.readLong();
-            Event event = Event.readFrom(in, length - 8 * 5);
+            Event event = Event.readFrom(in, length - Long.BYTES * 4);
```
fpj: I´m not sure I follow why we reducing the multiplier from 5 to 4.

andreipaduroiu: There are 4 Long fields in the header and 1 long field in the footer. After you read 32 bytes from the header you need to read the event, without the 8 byte footer. The following line reads the request id, but only if there's anything left to read.

I agree with you this is odd. Why did this work originally with a multiplier of 5. If I read the code right, the request Id would never be read and would always be -1. This should have been covered in a unit test.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
shrids: The check here is to validate if the length of the `Event` read from the byte stream is less than `length` of the byte stream passed.  If length of `Event` is greater than the length of the byte stream then a `InvalidMessageException` will be thrown.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
shrids: If a Pravega Client with the previous version  (<=7) sends a `ConditionalAppend` to Pravega running version (8) then the length passed to read the `Event` would be `length = Long.Bytes * 5` ). 
This would cause the `Event.readFrom(..)` to throw an `InvalidMessageException` . 
Hence the count is reduced to `Long.Bytes * 4`

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
fpj: This PR is about backward compatibility, so I'm assuming you need to change the verification here because the length changes with the version. The trouble is that if you make the verification less precise, you are opening up for other problematic scenarios, which is what I'm trying to avoid.

Starting from the basics, explain to me please why you need to change from `!=` to `>` in this validation? I checked both PR description and issue, and none of those explains why we need to do it.

fpj: For example, if `eventLength` is zero, then we won't throw which is clearly incorrect.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
fpj: I think what you are saying is making the length the minimum between the two versions, 7 and 8. Instead, as I have pointed above, the right way is to pass the correct length per version and validate according to the version.

shrids: Thanks for the review comment Flavio. As I understand the version is never passed as part of the WireCommand. (do let know if I misunderstood your comment)

fpj: What about this:

https://github.com/pravega/pravega/blob/master/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java#L54

https://github.com/pravega/pravega/blob/master/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java#L103

shrids: Yes, this version is passed only as part of the `io.pravega.shared.protocol.netty.WireCommands.Hello` command. This version is not sent for a ConditionalAppend and the version that was sent during `WireCommands.Hello` is not accessible to `ConditionalAppend`

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
shrids: Yes, if the `eventLength` is read as zero due to data corruption then InvalidMessageException is not thrown (will check if there is a better solution to this issue).

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -636,11 +622,29 @@ public static WireCommand readFrom(ByteBufInputStream in, int length) throws IOE
             UUID writerId = new UUID(in.readLong(), in.readLong());
             long eventNumber = in.readLong();
             long expectedOffset = in.readLong();
-            Event event = Event.readFrom(in, length - 8 * 5);
+            Event event = readEvent(in, in.available());
             long requestId = in.available() >= Long.BYTES ? in.readLong() : -1L;
+            // All the available bytes should be read.
+            if (in.available() > 0) {
+                throw new InvalidMessageException("Was expecting zero length but found: " + in.available());
```
tkaitchuck: This should not throw. It prevents us from adding new fields to this message in the future.

fpj: At the same time, if we remove, then we could be missing an ill-formed message. I think we should be checking precisely the length according to version. If we want to relax the verification to be forward compatible, then we are opening ourselves to other problems. I'm in favor of supporting only backward compatibility on the server side, which means that we only accept messages with the same version of earlier, and we validate precisely the length of messages per version.

shrids: @tkaitchuck @fpj I have improved the check to validate length. Does it make it any better now ?

tkaitchuck: Our current policy is to allow additional data for forward compatibility: https://github.com/pravega/pravega/wiki/PDP-21:-Protocol-revisioning
We should stick with that for this PR.

shrids: Yes, with this check the older server can throw an error if there is a newer version of `Conditional Append`. Removed this check.

fpj:
The code coverage report is indicating that we are not testing the throw lines. Can you write a couple of unit tests that uses assert throws?

PR #3620, opened 2019-04-12
Issue 3608: Sporadic build failures in PravegaTablesScaleRequestHandlerTest

### LINKED ISSUE #3608: Sporadic build failure: PravegaTablesScaleRequestHandlerTest
**Problem description**
PravegaTablesScaleRequestHandlerTest fails sporadically in different tests, typically with state being found as SCALING even at the end of a successful processing of scale request. 

**Problem location**
PravegaTablesScaleRequestHandlerTest.java

Find and fix the issue.

shiveshr:

**Change log description**  
This change set fixes the root cause that was plaguing PravegaTablesScaleRequestHandlerTest. 
The problem is that loading of cache, if triggered before invalidate is called, but completes after invalidate, may end up loading a stale value. 

**Purpose of the change**  
Fixes #3608 

**What the code does**  
We use guava's LoadingCache to cache both mutable and immutable metadata records. We invalidate mutable cached values upon either update or start of a new "operation" so as to load their latest values in cache whenever requested next. A caller may explicitly request to ignore cached values too by using "ignoreCached" flag on apis with mutable records. 

So if a caller requested a get with ignore cached flag, we used to first invalidate the cache, followed by calling getCachedData. The idea was that since we have invalidated the cache, getCachedData will load it from the underlying store. 
This is problematic. Explanation below:

The cache invalidation is insufficient because, if thread 1 calls invalidate followed by attempt to load value into cache, any subsequent invalidate calls between start of load and end of load are meaningless because cache has empty slot for the said key with a background loading being performed. 
Note: if there is an outstanding load request, guava cache does not schedule a subsequent load following an invalidate unless previous load has already completed and value is loaded against the key. 

This becomes interesting in following situation: 
1. The start of load could have been triggered before an update operation. So it has a chance that it will load a value from before the update. 
2. Any attempt from update to call invalidate on cache is rendered meaningless because cache hasnt loaded the value and thinks the slot is empty. 
3. when load completes and value is loaded into the cache, since the call to store was made before update operation, it may get the value from before the update. 

Why was this manifesting during scale related unit tests -- 
Scale workflow, after having completed metadata updates, but before resetting state, starts a background query to update scale related metrics. These metrics were started under a new operation context instead of reusing the scale's operation context. This invalidated the cached "state" followed by attempt to load the "state" when it was "SCALING". 
In the meantime, scale workflow also updated the state to ACTIVE and invalidated the cache. 
Then our test requests for the "state" with a new operation context. This also calls invalidate. 
However, now these invalidate were racing with the "load" from metrics thread. 
If invalidates were called before metrics thread could complete load, the invalidates were useless.  
And metrics thread, which had already begun the load, had read the value "SCALING" from before the state update but had not completed the load into the cache.
 
This meant that our test actually found the state to be SCALING whenever such a race occurred. 
This has a much higher probability of happening with our pravegaTables unit testing set up, where the future returned (and loaded) into the cache contained the value even before the future was loaded into the cache. With zookeeper, we made a remote call and loaded the future into cache corresponding to that remote call. So the value was retrieved from the store with a delay and hence we did not encounter it there. 

Solution:
Code fix:
- If caller explicitly asks the store to ignore cache, we will continue to invalidate the cache, but instead of reading the value from the cache, we read it explicitly from the store. 
This is done for both ZkStream and PravegaTableStream for all mutable records. 

Minor side optimization:
-  For reporting scale metrics, we do not need to create a new context. We should reuse the context from scale operation. So i have also made that change. 

Test fix:
- This change set also fixes the test code which was requesting the "state" from the cache with "ignoreCached" flag set to false. 

**How to verify it**  
No more sporadic failures.


controller/src/test/java/io/pravega/controller/server/eventProcessor/ScaleRequestHandlerTest.java
```diff
@@ -293,7 +293,7 @@ public void testScaleWithTransactionRequest() throws InterruptedException {
                 Lists.newArrayList(new AbstractMap.SimpleEntry<>(0.0, 1.0)), false, System.currentTimeMillis(), System.currentTimeMillis())).join();
 
         // 3. verify that scale is complete
-        State state = streamStore.getState(scope, stream, false, null, executor).join();
+        State state = streamStore.getState(scope, stream, true, null, executor).join();
```
fpj: I don't have a good sense of how often we end up ignoring the cache in the code. I wonder if we do it often enough that it doesn't make sense to have a cache at all.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -309,12 +310,16 @@
     @Override
     CompletableFuture<VersionedMetadata<EpochRecord>> getCurrentEpochRecordData(boolean ignoreCached) {
         return getId().thenCompose(id -> {
+
+            CompletableFuture<VersionedMetadata<Integer>> future;
             if (ignoreCached) {
                 store.invalidateCache(currentEpochRecordPath, id);
+                future = store.getData(currentEpochRecordPath, x -> BitConverter.readInt(x, 0));
```
fpj: Getting the data directly from the store when `ignoreCached` is true makes sense, but the semantics of `ignoreCached` is odd. `ignoreCached` is trying to make invalidate + repopulate atomic, but these two operations are really independent. My take is that `ignoreCached` should really be just that, *bypass the cache and get it from the store directly*. Invalidation should be performed separately when a new update happens.

shiveshr: Yes, i was thinking of doing that. Ignore cached need not invalidate. I decided to keep the changes to a minimum, hence let it be there. I will remove the invalidate from these getters for ignoreCached part.

controller/src/test/java/io/pravega/controller/server/eventProcessor/ScaleRequestHandlerTest.java
shiveshr: We use the cache for all immutable records and mutable records for getters. 
For example if multiple clients request for head or tail of the stream, they are all served from the cache. 
So truncationRecord for head and currentEpochRecord for tail are cached and these getters use ignoreCached = false. 

For all workflows, we always retrieve the value from the store and do not rely on cached value.

PR #3623, opened 2019-04-12
Issue-3621: use RawClient to simplify SegmentHelper implementation

### LINKED ISSUE #3621: Use RawClient to simplify SegmentHelper implementation
**Problem description**
"SegmentHelper was written before the RawClient existed. So it's doing it's job in a lot more complicated way than it needs to. The goal for the task should be to: Modify SegmentHelper to use the RawClient to remove a lot of it's boiler plate code."

**Problem location**
SegmentHelper
RawClient
and related testcases

huide9:
Signed-off-by: huide9 <huide9@gmail.com>

**Change log description**  
-use PraveagNodeUri to replace controller in RawClient constructor. 
-modify the related callers and the test cases to adapt the new impl
- ( this is first part - to modify the RawClient and functions which return Boolean)

**Purpose of the change**  
Fixes #3621 (part I)

**What the code does**  
change RawClient(controler, connectionFactory, segmentId) to:
             RawClient(PraveagaNodeUri, connectionFactory, segmentId)
which is the real needed by RawClient, so the followed SegmentHelper changes can easily utilize RawClient to simplify its impl

**How to verify it**  
RawClientTest.java


client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
```diff
@@ -66,7 +66,7 @@ public boolean write(ByteBuffer data, long expectedOffset) throws SegmentSealedE
                     .throwingOn(SegmentSealedException.class)
                     .run(() -> {
                         if (client == null || client.isClosed()) {
-                            client = new RawClient(controller, connectionFactory, segmentId);
+                            client = new RawClient(controller.getEndpointForSegment(segmentId.getScopedName()).get(), connectionFactory, segmentId);
```
shiveshr: controller.getEndpointForSegment is a remote call and you are blocking on it here before passing it as constructor param. You can make it asynchronous by doing 
controller.getEndpointForSegment().thenApply(uri -> new RawClient(uri, connectionFactory, segmentId))

client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
```diff
@@ -88,7 +91,11 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
     RawClient getConnection() {
         synchronized (lock) {
             if (client == null || client.isClosed()) {
-                client = new RawClient(controller, connectionFactory, segmentId);
+                try {
```
shiveshr: same as above

huide9:
working on the test case failure now

client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
huide9: thanks Shivesh,for the comments do you mean to replace with the following statement:
          controller.getEndpointForSegment(**segmentId.getScopedName()**).thenApply(uri -> new RawClient(uri, connectionFactory, segmentId));

client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
huide9: save as above - leave it untouched.

client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
```diff
@@ -88,7 +88,7 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
     RawClient getConnection() {
         synchronized (lock) {
             if (client == null || client.isClosed()) {
-                client = new RawClient(controller, connectionFactory, segmentId);
```
tkaitchuck: Don't need to change this

huide9: fixed

huide9:
> This part look good, but I don't see SegmentHelper being updated here. Perhaps there is a file missing?

it'll be in another PR, the previous consideration is to commit the less modified RawClient and then other parts.

huide9:
> Please see if you can create raw client object asynchronously in the callback for controller.getEndpointForSegment

this concern has been addressed. the modification has been removed. instead a new constructor added, so it doesn't has to modify these code.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -70,9 +79,15 @@
 
     private final SegmentStoreConnectionManager connectionManager;
     private final HostControllerStore hostStore;
-    
+    private final ConnectionFactory connectionFactory;
+
+    private final Object lock = new Object();
+    @GuardedBy("lock")
+    private RawClient client = null;
```
tkaitchuck: This does not need to be a member variable. All the callers of getConnection use the returned value, so there is no need to store it here.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -56,12 +56,24 @@ public void process(Reply reply) {
...
+                log.info("TableSegmentNotEmpty");
+                closeConnection(new ConnectionFailedException(reply.toString()));
             } else {
                 log.debug("Received reply {}", reply);
                 reply(reply);
             }
         }
 
+        @Override
+        public void tableSegmentNotEmpty(WireCommands.TableSegmentNotEmpty tableSegmentNotEmpty) {
```
tkaitchuck: This method is never going to be invoked. `process` above is handling the reply and it does not call this method.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -80,6 +92,12 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
         }
     }
 
+    public RawClient(PravegaNodeUri uri, ConnectionFactory connectionFactory) {
+        this.segmentId = null;
```
tkaitchuck: If segmentId is going to be null sometimes, you might as well just remove the field. It's only used in the log message when closing the connection anyway.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -52,10 +47,24 @@
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+
+import io.pravega.shared.protocol.netty.WireCommands;
```
tkaitchuck: Our style guide dictates that imports are to be in alphabetical order. Followed by static imports in alphabetical order. You can configure your IDE to do this for you. Please do that and have it re-order these.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +103,77 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
+    public CompletableFuture<Boolean> createSegment(final String scope,
+                                                    final String stream,
+                                                    final long segmentId,
+                                                    final ScalingPolicy policy,
+                                                    String controllerToken,
+                                                    final long clientRequestId) {
+        final String qualifiedStreamSegmentName = getQualifiedStreamSegmentName(scope, stream, segmentId);
+        final Controller.NodeUri uri = getSegmentUri(scope, stream, segmentId);
+        //final long requestId = (clientRequestId == RequestTag.NON_EXISTENT_ID) ? idGenerator.get() : clientRequestId;
+        final long requestId = (clientRequestId == RequestTag.NON_EXISTENT_ID) ? idGenerator.get() : idGenerator.get() + clientRequestId;
```
shrids: The usage of idGenerator is incorrect as there can be an id collision, the id is used by the connection pooling logic to ensure reposes to every `Flow` are correctly handled. 

The id (type `long`) contains two parts [Flow Id `int` + Sequence Number `int` ] .  
The flow id is used by the connection pooling logic to ensure correct `ReplyProcessor` is invoked. (In this case the ReplyProcessor of the RawClient should be invoked for the given given `Flow Id`.)

The pattern used inside `SegmentMetadataClientImpl`, which is very similar to the pattern this PR implements with the SegmentHelper, can be reused here. This will ensure all the operation performed in the SegmentHelper can be handled by reusing the underlying RawClient(s).

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +103,77 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
-        final FailingReplyProcessor replyProcessor = new FailingReplyProcessor() {
-            @Override
-            public void connectionDropped() {
-                log.warn(requestId, "CreateSegment {} Connection dropped", qualifiedStreamSegmentName);
-                result.completeExceptionally(
-                        new WireCommandFailedException(type, WireCommandFailedException.Reason.ConnectionDropped));
+    RawClient getConnection(PravegaNodeUri nodeUri) {
+        synchronized (lock) {
+            if (client == null || client.isClosed()) {
+                client = new RawClient(nodeUri, connectionFactory);
```
shiveshr: do we multiplex all requests on the same client?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +103,77 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
         Pair<Byte, Integer> extracted = extractFromPolicy(policy);
 
-        WireCommands.CreateSegment request = new WireCommands.CreateSegment(requestId, qualifiedStreamSegmentName,
-                extracted.getLeft(), extracted.getRight(), controllerToken);
-        sendRequestAsync(request, replyProcessor, result, ModelHelper.encode(uri));
-        return result;
+        Class[] expectedArray = {WireCommands.SegmentCreated.class,
+                WireCommands.SegmentAlreadyExists.class};
+
+        return connection.sendRequest(requestId, new WireCommands.CreateSegment(requestId, qualifiedStreamSegmentName,
```
shiveshr: we are losing all the logging specific to the response received. 
We should find a way to add appropriate logs.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: this has been fixed in recent commit

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: as per my understanding, that's one of the benefits to use RawClient; in recent commit, the 'requestId' has been changed to use `connection.getFlow().getNextSequenceNumber`.  @shrids could you please give more explanation for the question?

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -56,12 +56,24 @@ public void process(Reply reply) {
                 if (hello.getLowVersion() > WireCommands.WIRE_VERSION || hello.getHighVersion() < WireCommands.OLDEST_COMPATIBLE_VERSION) {
                     closeConnection(new IllegalStateException("Incompatible wire protocol versions " + hello));
                 }
+            } else if (reply instanceof WireCommands.WrongHost) {
+                log.info("WrongHost");
+                closeConnection(new ConnectionFailedException(reply.toString()));
+            } else if (reply instanceof WireCommands.TableSegmentNotEmpty) {
+                log.info("TableSegmentNotEmpty");
```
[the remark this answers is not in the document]
huide9: before the modification, this only appeared in `SegmentHelper.deleteTableSegment` the original code copied below:
```
 @Override
            public void tableSegmentNotEmpty(WireCommands.TableSegmentNotEmpty tableSegmentNotEmpty) {
                log.warn(requestId, "deleteTableSegment {} TableSegmentNotEmpty.", tableName);
                result.completeExceptionally(new WireCommandFailedException(type, WireCommandFailedException.Reason.TableSegmentNotEmpty));
            }
```

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shrids: @huide9 the question more related to the performance aspect. 
```
Before this PR the SegmentHelper used  `SegmentStoreConnectionManager` to manage 
connections with SegmentStore. This maintains a cache of connection pools to every 
SegmentStore where a pool could contain 
`SegmentStoreConnectionManager#MAX_CONCURRENT_CONNECTIONS` (500) network connections
per Segment store.
```

The current code is replacing this with a single RawClient per segment store. The question is do we need to create a pool of RawClient per segment store. 
As per my current understanding this is not required since having a pool of RawClients per Segment store would also translate into using same underlying network connection.

cc: / @shiveshr @tkaitchuck

shiveshr: The question is more about -- a client is specific to segment store instance.. right?
shouldnt we maintain a map of nodeUri to client. and return an existing client for that node uri. 
here it appears we could get back a client which could be for a different pravegaNodeUri.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -151,4 +172,9 @@ public void close() {
         closeConnection(new ConnectionClosedException());
     }
 
+    @VisibleForTesting
+    public Long retreiveRequestId() {
```
tkaitchuck: This shouldn't need to be here. The connection is being mocked, the mock can retrieve the IDs. This will avoid modifying the code.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: shall we put this RawClient modification in a separate PR, so the current one (SegmentHelper related) can go ahead - I can go ahead file an issue to track this, please let me know. thanks

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
huide9: removed

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
huide9: how about use a separate PR to remove segmentId from RawClient and all the callers and their test cases, so we can go ahead merge this PR first. thank you.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: removed

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -70,9 +74,12 @@
 
     private final SegmentStoreConnectionManager connectionManager;
     private final HostControllerStore hostStore;
-    
+    private final ConnectionFactory connectionFactory;
+    private final Object lock = new Object();
```
tkaitchuck: This lock doesn't appear to be guarding anything any more.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +95,67 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
         return Controller.NodeUri.newBuilder().setEndpoint(host.getIpAddr()).setPort(host.getPort()).build();
     }
 
+    private void closeConnection(Reply reply, RawClient client) {
+        log.info("Closing connection as a result of receiving: {}", reply);
+        RawClient c;
+        synchronized (lock) {
+            c = client;
+            client = null;
```
tkaitchuck: This line has no effect.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +95,67 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
+        }
+        if (c != null) {
```
tkaitchuck: The only way c could be null here is if the client passed into this method was null in the first place.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +95,67 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
+        for (Class<?> c : classArray) {
+            if (c.isAssignableFrom(reply.getClass())) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @SneakyThrows(ConnectionFailedException.class)
+    private Boolean transformReplyToBoolean(Reply reply, Class[] classArray, RawClient client) {
```
tkaitchuck: This needs a java doc explaining the return value.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +95,67 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
+        if (isReplyWanted(reply, classArray)) {
+            return true;
+        }
+
+        closeConnection(reply, client);
```
tkaitchuck: It's not obvious from the method name that it will close the connection. It's even less obvious that it does so only if the reply is not wanted.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +95,67 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
+        if (reply instanceof WireCommands.NoSuchSegment) {
+            throw new NoSuchSegmentException(reply.toString());
+        } else if (reply instanceof WireCommands.WrongHost) {
+            throw new ConnectionFailedException(reply.toString());
+        } else {
+            throw new ConnectionFailedException("Unexpected reply of " + reply + " when expecting a "
```
tkaitchuck: The semantics of this function are strange. It either returns `true` of throws. If it never returns false, why have a return value at all?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +95,67 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
 
+        RawClient connection = new RawClient(ModelHelper.encode(uri), connectionFactory);
+        final long requestId = connection.getFlow().getNextSequenceNumber();
         Pair<Byte, Integer> extracted = extractFromPolicy(policy);
 
-        WireCommands.CreateSegment request = new WireCommands.CreateSegment(requestId, qualifiedStreamSegmentName,
-                extracted.getLeft(), extracted.getRight(), controllerToken);
-        sendRequestAsync(request, replyProcessor, result, ModelHelper.encode(uri));
-        return result;
+        Class[] expectedArray = {WireCommands.SegmentCreated.class,
```
tkaitchuck: I think this should be called 'expectedReplies'. It's not helpful to put the type into the name, but it's actually confusing because it implies that you are expecting all of them.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -154,113 +164,29 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
-        };
-
-        WireCommands.DeleteSegment request = new WireCommands.DeleteSegment(requestId, qualifiedName, delegationToken);
-        sendRequestAsync(request, replyProcessor, result, ModelHelper.encode(uri));
-        return result;
+        RawClient connection = new RawClient(ModelHelper.encode(uri), connectionFactory);
+        final long requestId = connection.getFlow().getNextSequenceNumber();
+        Class[] expectedReply = {WireCommands.SegmentDeleted.class};
+        return connection.sendRequest(requestId, new WireCommands.DeleteSegment(requestId, qualifiedName, delegationToken))
+                .thenApply(r -> transformReplyToBoolean(r, expectedReply, connection));
```
tkaitchuck: I think `WireCommands.NoSuchSegment` is also an acceptable reply. After all if it is not there it's deleted right?

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -80,6 +82,12 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
         }
     }
 
+    public RawClient(PravegaNodeUri uri, ConnectionFactory connectionFactory) {
```
shrids: Before this change the RawClient was always tied to specific segment. With this change we are making a RawClient specific to a `PravegaNodeUri`. This implies a given RawClient can be used to perform operations on all the segments owned by a specific Segment Store.  (Hence we could have a pool of RawClient instances where there is atleast one instance per Segment Store. )

Also, `connectionFactory.establishConnection(uri, responseProcessor);` in line 87 below internally establishes a new connection , `establishConnection()` is invoked by this method and it disables Flow/ConnectionPooling.

Hence we should create a connection using `connectionFactory.establishConnection(flow, uri, responseProcessor)` similar to the segment specific RawClient constructor.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: Do we want to have a map of expected response and expected completion result.. also possible logging. 
that will keep parity with existing implementation. 

in some specific cases, we may want to fail request will a specific error. 
this implementation will limit that possibility. but having a map will keep those possibilities open.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: i'm taking the following as the reference: `ConditionalOutputStreamImpl` and `SegmentMetadataClientImple`;

consider the original `SegmentHelper` implementation: if the reply is not wanted: the default way in replyProcessor is throw exception.

my think is: return false or throw exception - throw exception for getting more detailed logs. 

maybe change `transfromReplyToBoolean` => `transformReply` to avoid misleading caused by function name?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +94,73 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
+                client.close();
+            } catch (Exception e) {
+                log.warn("Exception tearing down connection: ", e);
+            }
+        }
+    }
+
+    private boolean isReplyWanted(Reply reply, Class[] expectedReplies) {
+        for (Class<?> c : expectedReplies) {
+            if (c.isAssignableFrom(reply.getClass())) {
```
shiveshr: instead of isAssignableFrom, we can do an exact match by sending a Set of expectedReplies and checking if set.contains(reply.class)

That will be more efficient and not require a loop.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -88,64 +94,73 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
+            return true;
+        } else {
+            throw handleUnexpectedReply(reply, expectedReplies, client);
+        }
+    }
+
+    @SneakyThrows(ConnectionFailedException.class)
+    private RuntimeException handleUnexpectedReply(Reply reply, Class[] expectedReplies, RawClient client) {
+        closeConnection(reply, client);
+        if (reply instanceof WireCommands.NoSuchSegment) {
```
shiveshr: do we want to close the connection for NoSuchSegment response? (ref: line 135)

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -154,113 +125,39 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
-                result.complete(true);
-            }
-
-            @Override
-            public void processingFailure(Exception error) {
-                log.error(requestId, "deleteSegment {} failed", qualifiedName, error);
-                handleError(error, result, type);
-            }
+        RawClient connection = new RawClient(ModelHelper.encode(uri), connectionFactory);
+        final long requestId = connection.getFlow().getNextSequenceNumber();
```
shrids: We can also use a requestId with value `connection.getFlow().asLong()` as we are creating a new RawClient instance. A similar comment can applies to all other methods of SegmentHelper.

shrids: We can also use a requestId with value `connection.getFlow().asLong()` as we are creating a new RawClient instance. A similar comment can applies to all other methods of SegmentHelper.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -280,60 +177,24 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
+
+        Map<Class, SegmentHelperLog> expectedReplies = new HashMap<>();
+        expectedReplies.put(WireCommands.SegmentSealed.class,
+                () -> log.info(requestId, "sealSegment {} segmentSealed", qualifiedName));
+        expectedReplies.put(WireCommands.SegmentIsSealed.class,
+                () -> log.info(requestId, "sealSegment {} SegmentIsSealed", qualifiedName));
+
+        return connection.sendRequest(requestId, new WireCommands.SealSegment(requestId, qualifiedName, delegationToken))
+                .thenApply(r -> transformReply(r, expectedReplies, connection));
```
shrids: `io.pravega.client.netty.impl.RawClient#close` should be invoked after all the operations are complete.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -632,54 +493,19 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
     public CompletableFuture<Boolean> createTableSegment(final String tableName,
                                                          String delegationToken,
                                                          final long clientRequestId) {
-        final CompletableFuture<Boolean> result = new CompletableFuture<>();
```
shrids: Observing `io.pravega.controller.server.SegmentHelper#createTransaction` still uses `sendRequestAsync` to send the request instead of RawClient. Any specific reason why we are not using the RawClient here?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: you mean it's an option or you suggest to use getFlow().asLong() instead getFlow().getNextSequenceNumber()?

shrids: yes, we can use `getFlow().asLong() ` since we are using the a new RawClient for every request.
`getFlow().getNextSequenceNumber()` is not required since we are not reusing the `RawClient`.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: this changes are only for functions that return CompletableFuture<Boolean>; the reason to narrow down is to review less code to speed up merge (thank you all); after this part get merged, all the other function will be modified together in another PR.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: Would `close` close the connection too?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1303,6 +1086,56 @@ private boolean hasConnectionFailed(Throwable unwrap) {
         return new ImmutablePair<>(rateType, desiredRate);
     }
 
+    private interface SegmentHelperLog {
```
shiveshr: Nit: we dont need an additional interface here and can use `Consumer` interface for these purposes.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -280,60 +177,24 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
+        expectedReplies.put(WireCommands.SegmentSealed.class,
+                () -> log.info(requestId, "sealSegment {} segmentSealed", qualifiedName));
+        expectedReplies.put(WireCommands.SegmentIsSealed.class,
+                () -> log.info(requestId, "sealSegment {} SegmentIsSealed", qualifiedName));
+
+        return connection.sendRequest(requestId, new WireCommands.SealSegment(requestId, qualifiedName, delegationToken))
+                .thenApply(r -> transformReply(r, expectedReplies, connection));
     }
 
     public CompletableFuture<UUID> createTransaction(final String scope,
```
shiveshr: Any specific reason for excluding `transaction` related methods, `updatePolicy` from following the same pattern as rest of the methods?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: thank you Shivesh, the goal is to have ()->void  instead Consumer's (T)->void - as we don't have input argument to be handled here.

tkaitchuck: I think this is equivalent to the builtin Runnable. You can just change the type in the map and everything should work as-is without this being declared.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: this modification is for the functions that return CompleatbleFuture<Boolean>, the left part will be commit after this get merged.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1303,6 +1086,52 @@ private boolean hasConnectionFailed(Throwable unwrap) {
...
+                                                   Map<? extends Class, Runnable> expectedReplies,
+                                                   RawClient client) {
+        closeConnection(reply, client);
+        if (reply instanceof WireCommands.NoSuchSegment) {
+            throw new NoSuchSegmentException(reply.toString());
+        } else if (reply instanceof WireCommands.WrongHost) {
+            throw new ConnectionFailedException(reply.toString());
+        } else {
+            throw new ConnectionFailedException("Unexpected reply of " + reply + " when expecting one of "
+                    + expectedReplies.keySet().toString());
```
andreipaduroiu: Are you sure this will have the output you expect? I suggest using `String.join` to compose a proper String.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -94,58 +101,22 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
 
+        RawClient connection = new RawClient(ModelHelper.encode(uri), connectionFactory);
+        final long requestId = connection.getFlow().asLong();
         Pair<Byte, Integer> extracted = extractFromPolicy(policy);
 
-        WireCommands.CreateSegment request = new WireCommands.CreateSegment(requestId, qualifiedStreamSegmentName,
-                extracted.getLeft(), extracted.getRight(), controllerToken);
-        sendRequestAsync(request, replyProcessor, result, ModelHelper.encode(uri));
-        return result;
+        Map<Class, Runnable> expectedReplies = new HashMap<>();
```
andreipaduroiu: You are repeating the same pattern over and over again and creating a new Map just to put two elements in each. This makes the code look bad and is inefficient.

I recommend you modify your `transformReply` method to extract the necessary info (`Class.getSimpleName()` will get you the name of the class, and you can pass in a String with the operation name) and do a more generic error handler.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
huide9: tested, it works, an extra test case also added in SegmentHelperTest.java

```at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:604)
	at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:577)
	at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
	at io.pravega.client.netty.impl.RawClient.reply(RawClient.java:104)
	...
	at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:46)
	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:55)
Caused by: io.pravega.shared.protocol.netty.ConnectionFailedException: Unexpected reply of WireCommands.SegmentDeleted(type=SEGMENT_DELETED, requestId=17179869184, segment=/0.#epoch.0) when expecting one of [class io.pravega.shared.protocol.netty.WireCommands$SegmentAlreadyExists, class io.pravega.shared.protocol.netty.WireCommands$SegmentCreated]```

File: controller/src/main/java/io/pravega/controller/server/SegmentHelper.java (hunk 30 above)
Comment (huide9) [reply controller/src/main/java/io/pravega/controller/server/SegmentHelper.java]: Thanks for the comments Andrei. this has been fixed

File: controller/src/main/java/io/pravega/controller/server/SegmentHelper.java [hunk 31]
```diff
@@ -108,7 +109,7 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
         final long requestId = connection.getFlow().asLong();
         Pair<Byte, Integer> extracted = extractFromPolicy(policy);
 
-        Map<Class, Runnable> expectedReplies = new HashMap<>();
+        expectedReplies.clear();
```
andreipaduroiu: This is not what I asked.

I asked you to create a static, final, read-only HashMap which you create once (statically or in a constructor) and where you pre-populate all these expected responses.

The code you just wrote is not thread safe and it doesn't improve the performance over the previous iteration.

huide9: thanks Andrei,
• each request expects different different reply i.e., `createSegment` expects `SegmentAlreadyExistes` or `SegmentCreated` but not `SegmentSealed`, if put all these expected responses in a constructor it will be a problem when `SegmentSealed` returns to `createSegment` - as a special case.
• the RawClient instance will be create in each function call, I think a static final HashMap can not take the one time creation benefit.
please correct me.

huide9: a member Map to hold all the expected replies has been added for avoiding repeat adding expected replies into `expectedReplies` map at every function call, and the related handling also updated. thanks for the suggestion.

huide9:
> All looks good. I am waiting for confirmation that System Tests have passed successfully.
System test passed:
https://asd-nautilus-jenkins.isus.emc.com/view/Pravega/job/pravega-branch-builder/1538/

PR #3629, opened 2019-04-13
Issue 3628 : Change Table Segment Name schema to include system scope

### LINKED ISSUE #3628: Change table segment name schema on controller
**Problem description**
Currently the name scheme for table segments is `scope/table-<streamName>-tableName`.
The metrics generated on segment store for per segment generates metrics for these segments which can be visible to users as segments created under their scopes.

**Problem location**
Controller Pravega Table metadata store

Change name schema to use System scope.

shiveshr:
**Change log description**  
 Currently the name scheme for table segments is `<scope>/table-<streamName>-<tableName>`.
The metrics generated on segment store for per segment generates metrics for these segments which can be visible to users as segments created under their scopes.

Instead, we have moved table segment name to `_system/_tables/<scope>{optional}/<stream>{optional}/<table-name>`

**Purpose of the change**  
Fixes #3628 

**What the code does**  
The table name schema is changed such that all internal tables are created under system scope. 
StreamSegmentNameUtils has new utility functions to create fully qualified names for table segments which takes a scope and arbitrary number of tokens. Using these tokens, we compose the name such that its under `scope/_tables/token1/token2...`. 
These functions are `createQualifiedTableName(String scope, String... tokens), List<String> extractTokensFromQualifiedTableName(String name), boolean isTableSegment(String name).

Controller always creates all tables under _system scope. 

PravegaTableStoreHelper and SegmentHelper api signatures for table apis is changed to take fully qualified table names rather than scope+table name as earlier. 
HostStore api for getHostForTableSegment also takes qualified segment name as opposed to scope and table name earlier. 

The table names are computed in PravegaTablesStreamMetadataStore for all top level tables. 
These tables are _system/_tables/scopes, _system/_tables/deletedStreams, _system/_tables/completedTxnBatches and _system/_tables/completedTxnBatch-`number`

PravegaTablesScope has _system/_tables/`scope`/streamsInScope-`id`

PravegaTablesStream has _system/_tables/`scope`/`stream`/metadataTable-`id`, 
_system/_tables/`scope`/`stream`/epochsWithTxn-`id`, 
PravegaTablesStream has _system/_tables/`scope`/`stream`/txnsInEpoch-`id`

**How to verify it**  
All tests should pass. 
Unit tests added for StreamSegmentNameUtils and HostControllerStore.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
@@ -39,7 +39,7 @@ import static io.pravega.controller.store.stream.PravegaTablesStreamMetadataStor
  * Each such scope table is protected against recreation of scope by attaching a unique id to the scope when it is created. 
  */
 public class PravegaTableScope implements Scope {
-    private static final String STREAMS_IN_SCOPE_TABLE_FORMAT = "Table" + SEPARATOR + "streamsInScope" + SEPARATOR + "%s";
+    private static final String STREAMS_IN_SCOPE_TABLE_FORMAT = "Table" + SEPARATOR + "%s" + SEPARATOR + "streamsInScope" + SEPARATOR + "%s";
     private final String scopeName;
     private final PravegaTablesStoreHelper storeHelper;
     private final AtomicReference<UUID> idRef;
@@ -71,7 +71,7 @@ public class PravegaTableScope implements Scope {
                                      return storeHelper.addNewEntryIfAbsent(NameUtils.INTERNAL_SCOPE_NAME, SCOPES_TABLE, scopeName, newId());
                                  }))
                       .thenCompose(v -> getStreamsInScopeTableName())
-                      .thenCompose(tableName -> storeHelper.createTable(scopeName, tableName)
+                      .thenCompose(tableName -> storeHelper.createTable(NameUtils.INTERNAL_SCOPE_NAME, tableName)
                                                            .thenAccept(v -> log.debug("table created {}/{}", scopeName, tableName)));
     }
 
@@ -83,7 +83,7 @@ public class PravegaTableScope implements Scope {
 
     CompletableFuture<String> getStreamsInScopeTableName() {
         return getId().thenApply(id -> {
-            return String.format(STREAMS_IN_SCOPE_TABLE_FORMAT, id.toString());
+            return String.format(STREAMS_IN_SCOPE_TABLE_FORMAT, scopeName, id.toString());
         });
     }
 
@@ -104,7 +104,7 @@ public class PravegaTableScope implements Scope {
     @Override
     public CompletableFuture<Void> deleteScope() {
         return getStreamsInScopeTableName()
-                .thenCompose(tableName -> storeHelper.deleteTable(scopeName, tableName, true)
+                .thenCompose(tableName -> storeHelper.deleteTable(NameUtils.INTERNAL_SCOPE_NAME, tableName, true)
                                                      .thenAccept(v -> log.debug("table deleted {}/{}", scopeName, tableName)))
                 .thenCompose(deleted -> storeHelper.removeEntry(NameUtils.INTERNAL_SCOPE_NAME, SCOPES_TABLE, scopeName));
     }
@@ -115,7 +115,7 @@ public class PravegaTableScope implements Scope {
         AtomicReference<String> token = new AtomicReference<>(continuationToken);
         AtomicBoolean canContinue = new AtomicBoolean(true);
         return getStreamsInScopeTableName()
-                .thenCompose(entry -> storeHelper.getKeysPaginated(scopeName, entry,
+                .thenCompose(entry -> storeHelper.getKeysPaginated(NameUtils.INTERNAL_SCOPE_NAME, entry,
                         Unpooled.wrappedBuffer(Base64.getDecoder().decode(token.get())), limit)
                                                  .thenApply(result -> {
                                                      if (result.getValue().isEmpty()) {
@@ -132,7 +132,7 @@ public class PravegaTableScope implements Scope {
     public CompletableFuture<List<String>> listStreamsInScope() {
         List<String> result = new ArrayList<>();
         return getStreamsInScopeTableName()
-                .thenCompose(tableName -> storeHelper.getAllKeys(scopeName, tableName).collectRemaining(result::add)
+                .thenCompose(tableName -> storeHelper.getAllKeys(NameUtils.INTERNAL_SCOPE_NAME, tableName).collectRemaining(result::add)
                                                      .thenApply(v -> result));
     }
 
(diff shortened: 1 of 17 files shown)


controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
```diff
@@ -39,7 +39,7 @@
  * Each such scope table is protected against recreation of scope by attaching a unique id to the scope when it is created. 
  */
 public class PravegaTableScope implements Scope {
-    private static final String STREAMS_IN_SCOPE_TABLE_FORMAT = "Table" + SEPARATOR + "streamsInScope" + SEPARATOR + "%s";
+    private static final String STREAMS_IN_SCOPE_TABLE_FORMAT = "Table" + SEPARATOR + "%s" + SEPARATOR + "streamsInScope" + SEPARATOR + "%s";
```
fpj: One clarification. I can see that you are adding another `%s` to the string format, which makes sense, we need an additional one for the scope according to this PR. But, the description says that the table name should be:

```
_system/Table-<scope>-<stream>-<table-name>
```

which has three input strings rather than two, what am I missing?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -66,7 +66,7 @@
  * This class is coded for transaction ids that follow the scheme that msb 32 bits represent epoch
  */
 class PravegaTablesStream extends PersistentStreamBase {
-    private static final String STREAM_TABLE_PREFIX = "Table" + SEPARATOR + "%s" + SEPARATOR; // stream name
+    private static final String STREAM_TABLE_PREFIX = "Table" + SEPARATOR + "%s" + SEPARATOR + "%s" + SEPARATOR; // stream name
```
fpj: Another clarification question. I have the same question about this format containing two `%s` and not three. Can you clarify what the format is supposed to be for these constants?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
RaulGracia: I understand that the motivation for changing the naming structure of Table Segments used for metadata is that, currently, the metrics for these segments are not correctly reporting the `scope` tag, right? If this is the case, I have two questions:
1) Have you tested that the method `segmentTags()` in `MetricsTags.java` is correctly reporting the scope as a tag for metadata Table Segments? Maybe we need a test for that, which is related to this comment: https://github.com/pravega/pravega/pull/3594#pullrequestreview-226346168
2) As I understand the underlying problem, all the Table Segments created by users that do not follow the same structure as the one used here, will also have the problem of not correctly reporting the `scope` tag?

fpj: @RaulGracia I'll let @shiveshr answer you questions, but I want to address this specific point:

> I understand that the motivation for changing the naming structure of Table Segments used for metadata is that, currently, the metrics for these segments are not correctly reporting the scope tag

For me, the main motivation is that the tables of streams in scopes are in the user scopes themselves, and I think they shouldn't because that's really system metadata and should not live in user scopes. Moving the tables to the system scope also affects the metrics names as you are well pointing out.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
```diff
@@ -39,7 +39,7 @@
  * Each such scope table is protected against recreation of scope by attaching a unique id to the scope when it is created. 
  */
 public class PravegaTableScope implements Scope {
```
fpj: Can you opportunistically fix the name here to `PravegaTablesScope` to be consistent with the naming of other classes about pravda tables?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
fpj: Ok, I think I get this one now. This class is about scopes, there is no stream. The format for the name of the scope table is:

```
_system/Table-<scope>-<table-name>
```

is this right?

shiveshr: Yes

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
shiveshr: It’s the prefix that’s used in next three table name formats

controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
shiveshr: These are internal tables created by controller where it needs a predictable functional computation for the table name from scope and stream.. 
the user created table segments will have scope/table-name.


I am not sure about tags though. I have not tested tags behaviour explicitly.  I will check with you on it and update here

fpj: My sense is that tags is out of scope of this PR, although clearly it is something we need to look into.

fpj:
@andreipaduroiu 

> I'm concerned that we have too flat of a structure in these internal names. We'll be having a lot of files under _system, which could pose problems for some Tier 2 implementations which don't do flat hierarchies well.
> 
> Could we come up with a different scheme? Perhaps: _system/streams/<scope>/....?

I'm not sure I follow. Are you suggesting that we use tables that point to tables, like in virtual memory pagination? And if we do it, is the layout of the files/objects in Tier 2 such that they will be organized hierarchically?

I agree with the observation about a large number of say files under the same directory, good point, but it is unclear to me how we can leverage any kind of hierarchy given that we are simply naming tables. Could you clarify, please?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
shiveshr: I agree that it may be out of scope wrt this change set. But I want to understand how the tags are extracted and used so that we can evaluate whether we are breaking some other expectation vis a vis tags. 

So @RaulGracia could you please summarize how do metrics infrastructure apply scope tags? And what scope tag do we want for internal table segments. 

If the scope tag is extracted from the Fully Qualified segment name, then the scope tag earlier would have been user scope and now would be system scope. 

What should be the expectation there? If its system scope, then we do not need to do anything additional.

shiveshr:
@fpj I think what andrei is referring to is that if our segment naming scheme follows a directory structure, then a tier-2 implementation would make use of it to store tier-2 files/objects to follow that directory structure. 
Which means, by directly placing all metadata tables under _system, we can have very large number of files in that directory in tier-2. 

So maybe a naming scheme that looks as follows could be better:
`/_system/metadata/<scope>/<stream>/<tableName>`

@andreipaduroiu is that correct?

RaulGracia:
@shiveshr I think that the proposed naming structure for Controller Table Segments would improve the balance of load across directories in Tier 2. I would suggest to move the naming structure and constants (e.g., `Table` prefix) you plan to use to `shared` package, as it will be easier to i) parse the Table Segment format in `segmentTags` method for PR #3594, and ii) allow other components to follow the same or similar structure.

fpj:
I'm finding the discussion here a bit speculative. I'm not aware of us leveraging any such structure in Tier 2, but frankly, I'd need to refresh my cache about the bindings to remind myself about the data layout.

Keep in mind that introducing this structure in the name is not the only way, though. We could hash names to directories and create the same effect without having to introduce any structure to the name.

Can we close on the changes of the PR rather than speculating about future changes?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTableScope.java
shiveshr: done

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
```diff
@@ -303,4 +308,59 @@ private static StringBuffer getScopedStreamNameInternal(String scope, String str
...
+    }
+
+    /**
+     * Method to check if given segment name is a table name generated using getQualifiedTableName.
+     * @param qualifiedName qualified table name
+     * @return true if the name is generated using getQualifiedTableName. False otherwise
+     */
+    public static boolean isTableSegment(String qualifiedName) {
+        Preconditions.checkNotNull(qualifiedName);
+        return qualifiedName.contains(TABLES);
```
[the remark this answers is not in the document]
shiveshr: made it stronger.

PR #3639, opened 2019-04-16
Issue #3632: Add session to connection before it is established

### LINKED ISSUE #3632: EndToEndChannelLeakTest is fails due to race in connection pool
**Problem description**
EndToEndChannelLeakTest sometimes fails because there are two few connections.
This is due to a race in the connectionpool where if two connections are requested to the same host in rapid succession, the connection can be established (and the connection made available to the pool) before the session is added. This results in the pool falsely treating the newly created connection as an idle one, and assigning another session to be added to it.

**Problem location**
ConnectionPoolImpl

Add the session to the connection before the connection is added to the list in the pool.

tkaitchuck:

**Change log description**  
* Add first session to connection before the connection is established so the pool will know it is used.

**Purpose of the change**  
Fixes #3632 

**What the code does**  
Previously the connection pool was adding the session to the connection once the connection was established. This meant that a concurrent call to open a connection could see the connection as unused if the session had not been added yet. This caused two sessions to be added to the same newly created connection, which made ConnectionLeakTest flaky because the same number of connections were not always used. This change splits out the connection future from the session handler on the connection object. This way the session can be added to the session handler before the connection is established. 

Additionally changes are made to the connection leak test to make it deterministic.

**How to verify it**  
Connection leak test should no longer be flaky.

### DIFF
--- a/client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
@@ -72,8 +72,8 @@ public class ConnectionPoolImpl implements ConnectionPool {
     private static final Comparator<Connection> COMPARATOR = new Comparator<Connection>() {
         @Override
         public int compare(Connection c1, Connection c2) {
-            int v1 = c1.getSessionCount().orElse(Integer.MAX_VALUE);
-            int v2 = c2.getSessionCount().orElse(Integer.MAX_VALUE);
+            int v1 = Futures.isSuccessful(c1.getConnected()) ? c1.getSessionCount() : Integer.MAX_VALUE;
+            int v2 = Futures.isSuccessful(c2.getConnected()) ? c2.getSessionCount() : Integer.MAX_VALUE;
             return Integer.compare(v1, v2);
         }
     }; 
@@ -105,8 +105,7 @@ public class ConnectionPoolImpl implements ConnectionPool {
         // remove connections for which the underlying network connection is disconnected.
         List<Connection> prunedConnectionList = connectionList.stream().filter(connection -> {
             // Filter out Connection objects which have been completed exceptionally or have been disconnected.
-            CompletableFuture<SessionHandler> r = connection.getSessionHandler();
-            return !r.isDone() || (Futures.isSuccessful(r) && r.join().isConnectionEstablished());
+            return !connection.getConnected().isDone() || (Futures.isSuccessful(connection.getConnected()) && connection.getSessionHandler().isConnectionEstablished());
         }).collect(Collectors.toList());
         log.debug("List of connections to {} that can be used: {}", location, prunedConnectionList);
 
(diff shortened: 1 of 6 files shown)


client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
```diff
@@ -115,19 +114,20 @@ public ConnectionPoolImpl(ClientConfig clientConfig) {
...
+            final FlowHandler handler = new FlowHandler(location.getEndpoint(), batchSizeTracker);
+            CompletableFuture<Void> establishedFuture = establishConnection(location, handler);
+            connection = new Connection(location, handler, establishedFuture);
+            prunedConnectionList.add(connection);
         }
-        prunedConnectionList.add(connection);
+        ClientConnection result = connection.getFlowHandler().createFlow(flow, rp);
         connectionMap.put(location, prunedConnectionList);
-        return connection.getFlowHandler().thenApply(flowHandler -> flowHandler.createFlow(flow, rp));
+        return connection.getConnected().thenApply(v -> result);
```
shrids: We should create a `Flow` only once the connection is established.
i.e. 
```
 connection.getConnected().thenApply(v -> connection.getFlowHandler().createFlow(flow, rp));
```
Is there any reason it is being done before?

client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
```diff
@@ -137,25 +137,29 @@ public ConnectionPoolImpl(ClientConfig clientConfig) {
...
         // create a new connection.
-        CompletableFuture<FlowHandler> flowHandlerFuture = establishConnection(location);
-        Connection connection = new Connection(location, flowHandlerFuture);
-        return connection.getFlowHandler().thenApply(flowHandler -> flowHandler.createConnectionWithFlowDisabled(rp));
+        final AppendBatchSizeTracker batchSizeTracker = new AppendBatchSizeTrackerImpl();
+        final FlowHandler handler = new FlowHandler(location.getEndpoint(), batchSizeTracker);
+        CompletableFuture<Void> connectedFuture = establishConnection(location, handler);
+        Connection connection = new Connection(location, handler, connectedFuture);
+        ClientConnection result = connection.getFlowHandler().createConnectionWithFlowDisabled(rp);
+        return connectedFuture.thenApply(v -> result);
```
shrids: Similar question here.

client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
tkaitchuck: That is the race condition causing the tests to fail that this PR is attempting to fix. See the attached issue.

shrids:
https://asd-nautilus-jenkins.isus.emc.com/job/test-pravega-continuous-1/115/ one of the tests failed with these changes.

shrids:
the next run passed https://asd-nautilus-jenkins.isus.emc.com/jenkins/job/test-pravega-continuous/1897/

PR #3648, opened 2019-04-18
Issue 3596:  Improved Exception handling for ReplyProcessor.

### LINKED ISSUE #3596: Ensure `io.pravega.shared.protocol.netty.ReplyProcessor#processingFailure` takes a Throwable as a parameter.
**Problem description**
We can update the `io.pravega.shared.protocol.netty.ReplyProcessor#processingFailure` method to take a `Throwable` as a parameter.

**Problem location**
All ReplyProcessor implementations.

Refactor the api signature to remove redundant wrapping of exceptions.

shrids:
**Change log description**  
Defensive exception handling for netty callbacks and `io.pravega.shared.protocol.netty.ReplyProcessor` invocations.

**Purpose of the change**  
Fixes #3596

**What the code does**  
This code does not modify/change any functionality, it enables defensive exception handling for netty related operations.

This enhancement request originally was to ensure `Throwable` is  caught for netty related operations. But on closer inspection of the code observed that catching `Throwable` is undesirable and catching `java.lang.Exception` is infact enough.
- `channel.writeAndFlush(..)` : This is now wrapped within  `try {} catch {}` block to handle any Exceptions.
- Reply processor invocation is now wrapped under a `try {} catch {}` block to ensure any exceptions (if thrown) do not cause other invocations to be skipped.
It should be noted that no exceptions were observed for any the cases handled in the pull request. This change enables defensive exception handling to ensure there are no code breakages with any future versions of the ReplyProcessor. 

**How to verify it**  

### DIFF
--- a/client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
@@ -175,7 +175,7 @@ public class FlowHandler extends ChannelInboundHandlerAdapter implements AutoClo
                 rp.connectionDropped();
             } catch (Exception e) {
                 // Suppressing exception which prevents all ReplyProcessor.connectionDropped from being invoked.
-                log.warn("Encountered exception invoking ReplyProcessor for flow id " + flowId, e);
+                log.warn("Encountered exception invoking ReplyProcessor for flow id {}", flowId, e);
             }
         });
         super.channelUnregistered(ctx);
@@ -192,7 +192,7 @@ public class FlowHandler extends ChannelInboundHandlerAdapter implements AutoClo
                     rp.hello((WireCommands.Hello) cmd);
                 } catch (Exception e) {
                     // Suppressing exception which prevents all ReplyProcessor.hello from being invoked.
-                    log.warn("Encountered exception invoking ReplyProcessor.hello for flow id " + flowId, e);
+                    log.warn("Encountered exception invoking ReplyProcessor.hello for flow id {}", flowId, e);
                 }
             });
             return;
@@ -206,7 +206,7 @@ public class FlowHandler extends ChannelInboundHandlerAdapter implements AutoClo
             try {
                 processor.process(cmd);
             } catch (Exception e) {
-                log.error("ReplyProcessor.process failed for reply {} due to {}", cmd, e.getMessage());
+                log.warn("ReplyProcessor.process failed for reply {} due to {}", cmd, e.getMessage());
                 processor.processingFailure(e);
             }
         });
@@ -220,7 +220,7 @@ public class FlowHandler extends ChannelInboundHandlerAdapter implements AutoClo
                 rp.processingFailure(new ConnectionFailedException(cause));
             } catch (Exception e) {
                 // Suppressing exception which prevents all ReplyProcessor.processingFailure from being invoked.
-                log.warn("Encountered exception invoking ReplyProcessor.processingFailure for flow id " + flowId, e);
+                log.warn("Encountered exception invoking ReplyProcessor.processingFailure for flow id {}", flowId, e);
             }
         });
     }
--- a/client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
@@ -206,7 +206,7 @@ public class FlowHandler extends ChannelInboundHandlerAdapter implements AutoClo
             try {
                 processor.process(cmd);
             } catch (Exception e) {
-                log.error("Encountered exception invoking ReplyProcessor.process for reply " + cmd, e);
+                log.error("ReplyProcessor.process failed for reply {} due to {}", cmd, e.getMessage());
                 processor.processingFailure(e);
             }
         });
--- a/client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
@@ -170,8 +170,13 @@ public class FlowHandler extends ChannelInboundHandlerAdapter implements AutoClo
         }
         channel.set(null);
         flowIdReplyProcessorMap.forEach((flowId, rp) -> {
-            rp.connectionDropped();
-            log.debug("Connection dropped for flow id {}", flowId);
+            try {
+                log.debug("Connection dropped for flow id {}", flowId);
+                rp.connectionDropped();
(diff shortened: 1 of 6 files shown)


client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -202,8 +215,13 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         flowIdReplyProcessorMap.forEach((flowId, rp) -> {
-            rp.processingFailure(new ConnectionFailedException(cause));
-            log.debug("Exception observed for flow id {}", flowId);
+            try {
+                log.debug("Exception observed for flow id {}", flowId, cause);
```
andreipaduroiu: Do you want to log the entire stack trace or just the exception message?

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -194,6 +206,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
             try {
                 processor.process(cmd);
             } catch (Exception e) {
+                log.error("Encountered exception invoking ReplyProcessor.process for reply " + cmd, e);
```
tkaitchuck: Not sure this is needed. The processor should be logging this with more/better context. We'll just end up with duplicated stack traces.

shrids: updated it.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -170,8 +170,13 @@ public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
...
         channel.set(null);
         flowIdReplyProcessorMap.forEach((flowId, rp) -> {
-            rp.connectionDropped();
-            log.debug("Connection dropped for flow id {}", flowId);
+            try {
+                log.debug("Connection dropped for flow id {}", flowId);
+                rp.connectionDropped();
+            } catch (Exception e) {
+                // Suppressing exception which prevents all ReplyProcessor.connectionDropped from being invoked.
+                log.warn("Encountered exception invoking ReplyProcessor for flow id " + flowId, e);
```
RaulGracia: Is any exception caught here equally important? Just thinking in the case that there could be expected and harmless exceptions being caught here, so we could change the log level and do no print the whole stack trace for them (for the rest, this way of logging may be ok).

RaulGracia: Please, use `{}` for logging arguments in log lines (e.g., `flowId`).

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -182,7 +187,14 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
         log.debug(connectionName + " processing reply {} with flow {}", cmd, Flow.from(cmd.getRequestId()));
 
         if (cmd instanceof WireCommands.Hello) {
-            flowIdReplyProcessorMap.forEach((flowId, rp) -> rp.hello((WireCommands.Hello) cmd));
+            flowIdReplyProcessorMap.forEach((flowId, rp) -> {
+                try {
+                    rp.hello((WireCommands.Hello) cmd);
+                } catch (Exception e) {
+                    // Suppressing exception which prevents all ReplyProcessor.hello from being invoked.
+                    log.warn("Encountered exception invoking ReplyProcessor.hello for flow id " + flowId, e);
```
RaulGracia: Same comment here as before.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -194,6 +206,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
             try {
                 processor.process(cmd);
             } catch (Exception e) {
+                log.error("ReplyProcessor.process failed for reply {} due to {}", cmd, e.getMessage());
```
RaulGracia: Same consideration here: for some expected exceptions logging the "message" may be enough (in this case, maybe `error` would be too severe), but for other unexpected exceptions it would be also useful to have the stack trace.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -202,8 +215,13 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         flowIdReplyProcessorMap.forEach((flowId, rp) -> {
-            rp.processingFailure(new ConnectionFailedException(cause));
-            log.debug("Exception observed for flow id {}", flowId);
+            try {
+                log.debug("Exception observed for flow id {} due to {}", flowId, cause.getMessage());
+                rp.processingFailure(new ConnectionFailedException(cause));
+            } catch (Exception e) {
+                // Suppressing exception which prevents all ReplyProcessor.processingFailure from being invoked.
+                log.warn("Encountered exception invoking ReplyProcessor.processingFailure for flow id " + flowId, e);
```
RaulGracia: Please, use `{}` for the argument instead of String concatenation.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
shrids: Ideally exceptions should not be thrown by the ReplyProcessor implementations (the current ReplyProcessors do not throw it too). If thrown then it should be logged as it indicates that the newer ReplyProcessor s are incorrect and we prevent breakages with the future version of ReplyProcessors. With this in context feel that WARN with the stack trace is helpful.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
shrids: Will change the log to warn level and since `processor.processingFailure(e)` is invoked next the stack trace will be logged there for unexpected exception.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
shrids: done.

Issue #3638, opened 2019-04-16
System test failures since the readers were not able to read all events.


andreipaduroiu:
**Problem description**

Observing that a we can recieve empty `SEGMENT_READ`. (my understanding was that SSS responds only when there is data)
```2019-04-11 02:47:31,330 1349655 [epollEventLoopGroup-4-2] TRACE i.p.s.protocol.netty.CommandDecoder - Decode a message on connection: 10.200.90.219/10.200.90.219:12345. Message was WireCommands.SegmentRead(type=SEGMENT_READ, segment=testReadWriteAndScaleScope1144952233/testReadWriteAndScaleStream/2.#epoch.1, offset=223630, atTail=true, endOfSegment=false, data=java.nio.HeapByteBuffer[pos=0 lim=0 cap=0], requestId=304942678016)```

**Problem location**
Client / Test file.

**Suggestions for an improvement**
Needs to be investigated.

shrids:
Closing this issue since this has been fixed as part of PR: https://github.com/pravega/pravega/pull/3639

shrids:
reopening the issue as it was observed again.

shrids:
The system test failure was root caused to repeated connection failures. 
Details:
- The reader were not able to read all the events since the connection to Segmentstore is failing
```2019-04-25 13:35:26,569 797688 [clientInternal-1] ERROR i.p.c.s.i.AsyncSegmentInputStreamImpl - Error while sending request WireCommands.ReadSegment(type=READ_SEGMENT, segment=testReadWriteAndScaleScope1839109582/_RGtestReadWriteAndScaleReaderGroup1345371180/0.#epoch.0, offset=2115, suggestedLength=1048576, delegationToken=, requestId=111669149696)
io.pravega.shared.protocol.netty.ConnectionFailedException: Connection to 10.200.90.138 is not established.
        at io.pravega.client.netty.impl.FlowHandler.getChannel(FlowHandler.java:122)
        at io.pravega.client.netty.impl.ClientConnectionImpl.sendAsync(ClientConnectionImpl.java:69)
        at io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl.sendRequestOverConnection(AsyncSegmentInputStreamImpl.java:202)
        at io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl.lambda$null$2(AsyncSegmentInputStreamImpl.java:188)```

Detailed logs: 
[...]

PR #3654, opened 2019-04-19
Issue 3638: Ensure the AsyncSegmentInputStreamImpl handles ConnectionFailedException during sendAsync.

shrids:
**Change log description**  
- Ensure AsyncSegmentInputStreamImpl closes the connection incase a `ConnectionFailedException` is observed during `io.pravega.client.netty.impl.ClientConnection#sendAsync()`

**Purpose of the change**  
Related to issue #3638

**What the code does**  
AsycnSegmentInputStreamImpl invokes `ClientConnection.sendAsync(...)` to send commands like `WireCommand.ReadSegment`. This method can throw a `ConnectionFailedException` incase the underlying Netty Channel was set to null due to `channelUnregistered()` callback by netty. This can cause the AsyncSegmentInputStream to retry until the replyProcessor.connectionDropped() is invoked by `FlowHandler`.

**How to verify it**  

### DIFF
--- a/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
@@ -168,7 +168,6 @@ class AsyncSegmentInputStreamImpl extends AsyncSegmentInputStream {
         Exceptions.checkNotClosed(closed.get(), this);
         WireCommands.ReadSegment request = new WireCommands.ReadSegment(segmentId.getScopedName(), offset, length,
                                                                         this.delegationToken, requestId);
-
         return backoffSchedule.retryWhen(t -> {
             Throwable ex = Exceptions.unwrap(t);
             if (closed.get()) {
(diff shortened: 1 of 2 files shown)


client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
```diff
@@ -181,11 +180,17 @@ public boolean isClosed() {
...
                         if (ex != null) {
-                            log.warn("Exception while establishing connection with Pravega " +
-                                    "node", ex);
+                            log.warn("Exception while establishing connection with Pravega node", ex);
                             closeConnection(new ConnectionFailedException(ex));
                         }
-                    }).thenCompose(c -> sendRequestOverConnection(request, c));
+                    }).thenCompose(c -> sendRequestOverConnection(request, c)
+                            .whenComplete((reply, ex) -> {
+                                if (ex instanceof ConnectionFailedException) {
```
andreipaduroiu: Shouldn't you unwrap this exception?

andreipaduroiu: Consider using `Futures.exceptionallyExpecting` which will handle all of this and make your code more compact.

shrids:
Closing this PR, since we are not observing with the latest client code.

shrids:
Reopening this as this error was observed again.

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
shrids: The future returned  `sendRequestOverConnecton` is used by the overall retry logic. Hence 
```
Futures.exceptionallyExpecting(future, t -> t instanceof ConnectionFailedException, value)
``` 
cannot be used. Do let me know if there is a better way to compact the code.
(the only other option is to invoke `closeConnection(...)` as part of the predicate)

andreipaduroiu: Ok. I missed the fact that this is not really returning a value but executing something.

Issue #3653, opened 2019-04-19
Exceptions raised by controller gRPC API are overly abstract


ravisharda:
**Problem description**
The Controller gRPC API throws the rather general [StatusRuntimeException](https://grpc.io/grpc-java/javadoc/io/grpc/StatusRuntimeException.html), which is problematic in these respects:
* *Leaky abstraction*. One has to inspect the inner exceptions and/or the message itself to understand the exception condition, which is perhaps OK for human users. But, the client app shouldn't need to depend on exception messages or inner exceptions, as those things can/should be allowed to change without any compatibility concerns.

* *Lack of expressiveness*. The exception name `StatusRuntimeException` is not expressive, even though it is perfectly OK as a top-level exception (as explained below). For instance, receiving the generic `StatusRuntimeException` when auth has failed, doesn't tell you what really went wrong. One can look at the message, but its really the exception name itself which should convey the meaning of the condition leading to the exception. Besides, not to mention, for exception handling on the client app, it really can only handle exceptions by catching the exception itself. 

* *Difficulty in error handling*. `StatusRuntimeException` homogenizes several classes of exception conditions, which is useful for callers if they don't care about what has failed. This is much like Java's `RuntimeException`,  `SQLException` or `IOException`. 

   It's not useful to throw such top-level exceptions directly though: you'd rather throw subtypes like `IllegalArgumentException`, `SQLIntegrityConstraintViolationException` or ` FileNotFoundException`. Similarly, it'd be much more useful if subtypes of `StatusRuntimeException` (and not `StatusRuntimeException` itself) were thrown, each representing a precise exception condition. 

  A negative side-effect of this sub-problem can be seen in the client-side proxy `ControllerImpl` in the lines shown below. It retries calls to the server upon encountering `StatusRuntimeException`, irrespective of the underlying exception condition and whether it is recoverable or not. It has to do that because the exception encapsulates both recoverable and irrecoverable conditions. However, there is no point in retrying if authentication or authorization has failed. It might make sense to do so if the condition was a transient connectivity issue. A side effect of automatic retries on irrecoverable errors is that it takes much longer than expected for a client application to get the feedback later rather than sooner. The root of this problem ultimately lies in the absence of precise exceptions being thrown by the server.
https://github.com/pravega/pravega/blob/cb253c8c73b076adbe76cd87da8418c09a8989b3/client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java#L156-L159

**Problem location**
[...]

PR #3656, opened 2019-04-19
Issue 3643: Fix reliability issues of standalone mode tests

### LINKED ISSUE #3643: Auth enabled Standalone tests are not checking for right exceptions
**Problem description**
1. When i run the `AuthEnabledInProcPravegaClusterTest.testCreateStreamFailsWhenCredentialsAreInvalid` tests with auth enabled, i see errors of following kind show up in the log:
```Apr 17, 2019 1:28:28 AM io.grpc.internal.SerializingExecutor run
SEVERE: Exception while executing runnable io.grpc.internal.ServerImpl$ServerTransportListenerImpl$1StreamCreated@46f751b9
java.lang.NullPointerException: startCall() returned a null listener for method io.pravega.controller.stream.api.grpc.v1.ControllerService/createScope
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl.startWrappedCall(ServerImpl.java:565)
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl.startCall(ServerImpl.java:544)
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl.access$1800(ServerImpl.java:364)
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl$1StreamCreated.runInContext(ServerImpl.java:480)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
```

The test simply verifies if the exception thrown is of the kind RuntimeException and declares success. 
I suspect that the exception that has been logged is not what the test expects and this points to a problem. The exception logged is NPE which doesnt seem right. 

2. SecurePravegaClusterTest.failingCreateTestStream fails in createScope with following error: 
```
java.util.concurrent.CompletionException: io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
io.pravega.common.util.RetriesExhaustedException: java.util.concurrent.CompletionException: io.grpc.StatusRuntimeException: UNAVAILABLE: io exception

....
Caused by: javax.net.ssl.SSLHandshakeException: General OpenSslEngine problem
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext$AbstractCertificateVerifier.verify(ReferenceCountedOpenSslContext.java:629)
....
Caused by: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
...
Caused by: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

And the test again simply verifies if a RuntimeException is thrown and claims success. 
The error seems to be expected, but IMO the test should do a stricter verification to avoid over generalization and miss out if a regression is introduced in this path. 

3. Please add timeouts to these tests.

**Problem location**
AuthEnabledInProcPravegaClusterTest

ravisharda:
**Change log description**  

Improve the standalone mode/in-proc cluster tests so that:

* They are more likely to catch regression issues, and 
* They do not block builds for too long.

**Purpose of the change**  
Fixes #3643.

**What the code does**  

1. Refactors the test code in order to make it easier to do the rest of the items below well and to reduce duplication of code. 
    
   Also, it moves the primary happy path test to the base test class, so that the same scenario is tested for all 4 cases: a) TLS and auth disabled, b) both enabled, c) TLS enabled, but Auth disabled, and d) Auth disabled, but TLS disabled. In effect, this adds more tests to the suite. 

2. Adds timeouts to tests that verify exceptional scenarios, so that they don't block builds. 
   
   Note that the timeouts used in the tests are kept large. This is because the retry mechanism on the client side kicks in even when the exception condition is irrecoverable and therefore it can take some time before the client application encounters the exception. This issue has been reported along with other issues in #3653.

3. Checks for specific exception conditions when verifying exceptional scenarios. Earlier those tests were expecting the RuntimeException, which can prevent the tests from catching regression issues, as described in issue #3643.

4. Removes a test from SecurePravegaClusterTest and adds one in TlsEnabledInProcPravegaClusterTest instead, to test a more specific scenario and to make the intent of the test explicit. 

**How to verify it**  
All standalone mode tests must pass (`./gradlew :standalone:test`).


standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
```diff
@@ -52,105 +38,61 @@ public void setUp() throws Exception {
...
+            streamManager.createScope(scopeName());
+        } catch (RetriesExhaustedException e) {
+            if (!hasAuthExceptionAsRootCause(e)) {
+                throw e;
+            }
+        }
+    }
+
+    private boolean hasAuthExceptionAsRootCause(Throwable e) {
+        Throwable innermostException = ExceptionUtils.getRootCause(e);
```
shiveshr: Please use our utility class called Exceptions.unwrap

ravisharda: The `Exceptions.unwrap` won't work here, as it only unwraps `CompletionException`  or `ExecutionException`, but here we have a `RetriesExhaustedException`.

andreipaduroiu: I suppose we could add `RetriesExhaustedException` to `Exceptions.unwrap`. I don't see an issue with that.

@shiveshr and @tkaitchuck: do you anticipate any hidden bugs if we do this? There is no value to a `RetriesExhaustedException` **as long as** it contains a meaningful inner exception (if it doesn't, then we shouldn't unwrap)

ravisharda: I tried adding `RetriesExhaustedException` to the whitelist used in Exceptions.unwrap(), but there was an an unintended semantical side-effect: the `StreamManager.createScope(...)` method then returns `StatusRuntimeException` as opposed to `RetriesExhaustedException` as it does today. There might be other side-effects in other paths.

All I need here is to fetch the innermost/root exception and check if it is related to auth. The `org.apache.commons.lang3.exception.ExceptionUtils.getRootCause(e)` used here does just that. That being said, if we do need to abstract out the unwrapping of the root exception through our own `Exceptions` utility, one way to do that is to write a new method in the `Exceptions` class that delegates the call to `ExceptionUtils.getRootCause()`; I don't see much benefit of that though unless there is a logic that we needed to add on top of that.  

What do you guys say?

shiveshr: My suggestion is we separate the two in different issues -
for the sake of this test, we use Exceptions.unwrap + RetriesExhausted 

And have a separate issue dealing with RetriesExhausted as part of Exceptions.unwrap where we can evaluate if it causes any regressions or is safe to do so.

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
```diff
@@ -52,105 +38,61 @@ public void setUp() throws Exception {
...
                 .credentials(new DefaultCredentials("", ""))
                 .controllerURI(URI.create(localPravega.getInProcPravegaCluster().getControllerURI()))
                 .build();
 
         @Cleanup
         StreamManager streamManager = StreamManager.create(clientConfig);
 
-        AssertExtensions.assertThrows(RuntimeException.class,
-                () -> streamManager.createScope(scope));
+        try {
```
shiveshr: Also, please use assertExtensions.assertThrows with the predicate check for thrown exception

ravisharda: Thanks for catching the bug in the test. Have fixed it now in the latest commit. The assertion code is also looking simpler with that change.

standalone/src/test/java/io/pravega/local/AuthEnabledInProcPravegaClusterTest.java
ravisharda: Please take a look at the code in the latest commit. I still use the ExceptionUtils for reasons mentioned in my last comment in the thread. This should cover the intent of the first paragraph of @shiveshr's comments. Please unresolve this conversation if you are not satisfied.

ravisharda:
@andreipaduroiu I've responded to your last comment inline with the same comment in the thread. Please review the same.

PR #3662, opened 2019-04-22
Issue 3575: Enable running of system tests in a Distributed Pravega Cluster with basic authentication enabled.

### LINKED ISSUE #3575: Pravega Security: System tests to verify authentication on a distributed cluster, between client and controller
**Problem description**
Add tests to the System Test Suite to verify authentication of client with Controller and Segment Store, for a distributed cluster deployment.

**Problem location**
See parent issue: 
https://github.com/pravega/pravega/issues/1669

This is only a basic auth test for a distributed cluster. Need to add tests for each supported auth mechanism and deployment scenarios in Pravega.

pbelgundi:

**Change log description**  
Allow users to run all system tests on a Pravega Distributed Cluster, with basic authentication enabled . 

**Purpose of the change**  
Fixes #3575  
This change enables running of system tests with authentication enabled in Controller and Segment Store. The basic 'admin' username & password provided via default password file in the controller image (/opt/pravega/conf/passwd) are used for authentication.

**What the code does**  
Moved the system properties being set inside AbstractService.java into a properties file in the resources folder. These are set to their default values when security is disabled, and include the appropriate additional properties when security is enabled.
The current gradle task 'startK8SystemTests' can now be invoked with "-DsecurityEnabled=true" flag to run all system tests with security (auth) enabled. When not set (default), the value of this flag would be 'false' and tests would run with security 'disabled'. Jenkins jobs would also have the option of running system tests with security enabled or disabled.

**How to verify it**  

After this test run, verified options set in pravega cluster using :

'kubectl get pravegacluster pravega -o yaml'
And the following security properties were present in the pravega spec under 'options':
"controller.auth.enabled" "true"
"controller.auth.userPasswordFile" "/opt/pravega/conf/passwd"
"controller.auth.tokenSigningKey" "secret"
"autoScale.authEnabled" "true"
"autoScale.tokenSigningKey" "secret"

Also checked the Controller and segment store logs for these system properties to be set at startup.

Executed all system tests successfully with security disabled.
Test Results: https://asd-nautilus-jenkins.isus.emc.com/job/test-pravega-continuous/1960/System_20Tests/
StreamCutsTest fails with security enabled and disabled.
But failure is not due to authentication in both cases.

On executing tests with security(auth) enabled,  the following tests failed with security enabled:
AutoScaleTest - Authentication between Controller and Segment Store fails.
ControllerRestAPITest - fails even when default basic auth credentials are added to jersey client and tls is enabled.
Readtxnwriteautoscalewithfailovertest - Authentication between Controller and Segment Store fails.

These require fixes on Controller and Segment Store and would be addressed as part of issue: https://github.com/pravega/pravega/issues/3777 and https://github.com/pravega/pravega/issues/3779

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -818,7 +818,9 @@ project('test:system') {
         if (System.getProperty("imageVersion") == null) {
             systemProperty "imageVersion", "INVALID_IMAGE_VERSION"
         }
-
+        
+        systemProperty "securityEnabled", System.getProperty("securityEnabled", "false")
+       
         maxParallelForks = 1
     }
 
--- a/test/system/src/main/java/io/pravega/test/system/framework/Utils.java
+++ b/test/system/src/main/java/io/pravega/test/system/framework/Utils.java
@@ -9,6 +9,9 @@
  */
 package io.pravega.test.system.framework;
 
+import com.google.common.collect.ImmutableMap;
+import io.pravega.client.ClientConfig;
+import io.pravega.client.stream.impl.DefaultCredentials;
 import io.pravega.test.system.framework.services.Service;
 import io.pravega.test.system.framework.services.docker.BookkeeperDockerService;
 import io.pravega.test.system.framework.services.docker.HDFSDockerService;
@@ -24,6 +27,8 @@ import io.pravega.test.system.framework.services.marathon.PravegaControllerServi
 import io.pravega.test.system.framework.services.marathon.PravegaSegmentStoreService;
 import io.pravega.test.system.framework.services.marathon.ZookeeperService;
 
+import java.io.IOException;
+import java.util.Properties;
 import lombok.extern.slf4j.Slf4j;
 
 import java.net.URI;
@@ -42,7 +47,10 @@ public class Utils {
     public static final int ALTERNATIVE_CONTROLLER_PORT = 9093;
     public static final int ALTERNATIVE_REST_PORT = 9094;
     public static final TestExecutorFactory.TestExecutorType EXECUTOR_TYPE = TestExecutorFactory.getTestExecutionType();
-
+    public static final boolean AUTH_ENABLED = isAuthEnabled();
+    public static final String PROPERTIES_FILE = "pravega.properties";
+    public static final String PROPERTIES_FILE_WITH_AUTH = "pravega_withAuth.properties";
+    public static final ImmutableMap<String, String> PRAVEGA_PROPERTIES = readPravegaProperties();
 
     /**
      * Get Configuration from environment or system property.
@@ -77,7 +85,7 @@ public class Utils {
                 return new BookkeeperDockerService(serviceId, zkUri);
             case KUBERNETES:
             default:
-                return new BookkeeperK8sService(serviceId, zkUri);
+                return new BookkeeperK8sService(serviceId, zkUri, getPravegaProperties());
         }
     }
 
@@ -89,7 +97,7 @@ public class Utils {
                 return new PravegaControllerDockerService(serviceName, zkUri);
             case KUBERNETES:
             default:
-                return new PravegaControllerK8sService(serviceName, zkUri);
+                return new PravegaControllerK8sService(serviceName, zkUri, getPravegaProperties());
         }
 
     }
@@ -116,8 +124,43 @@ public class Utils {
                 return  new PravegaSegmentStoreDockerService(serviceId, zkUri, hdfsUri, contUri);
             case KUBERNETES:
             default:
-                return new PravegaSegmentStoreK8sService(serviceId, zkUri);
+                return new PravegaSegmentStoreK8sService(serviceId, zkUri, getPravegaProperties());
         }
+    }
+
+    private static ImmutableMap<String, String> getPravegaProperties() {
+        return PRAVEGA_PROPERTIES;
+    }
+
+    private static ImmutableMap<String, String> readPravegaProperties() {
+        String resourceName = PROPERTIES_FILE;
+        if (AUTH_ENABLED) {
+            resourceName = PROPERTIES_FILE_WITH_AUTH;
+        }
+        Properties props = new Properties();
+        try {
+            props.load(Utils.class.getClassLoader().getResourceAsStream(resourceName));
+        } catch (IOException e) {
+            log.error("Error reading properties file.", e);
+        }
+        ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+        props.forEach((key, value) -> builder.put(key.toString(), value.toString()));
+        return builder.build();
+    }
+
+    public static ClientConfig buildClientConfig(URI controllerUri) {
+        if (!AUTH_ENABLED) {
+            log.debug("Generating config with auth disabled.");
+            return ClientConfig.builder().controllerURI(controllerUri).build();
+        } else {
+            log.debug("Generating config with auth enabled.");
+            return ClientConfig.builder()
+                               // auth
+                               .credentials(new DefaultCredentials("1111_aaaa", "admin"))
+                               .controllerURI(controllerUri)
+                               .build();
+        }
+
     }
 
     /**
@@ -130,6 +173,7 @@ public class Utils {
      * Default value is false
      * @return true if skipServiceInstallation is set, false otherwise.
      */
+
     public static boolean isSkipServiceInstallationEnabled() {
         String config = getConfig("skipServiceInstallation", "false");
         return config.trim().equalsIgnoreCase("true") ? true : false;
@@ -144,6 +188,13 @@ public class Utils {
     public static boolean isAwsExecution() {
         String dockerConfig = getConfig("awsExec", "false");
         return dockerConfig.trim().equalsIgnoreCase("true") ?  true : false;
-
     }
+
+    private static boolean isAuthEnabled() {
+        String securityEnabled = Utils.getConfig("securityEnabled", "false");
+        return Boolean.valueOf(securityEnabled);
+    }
+
+
+
 }
--- a/test/system/src/test/java/io/pravega/test/system/ControllerRestApiTest.java
+++ b/test/system/src/test/java/io/pravega/test/system/ControllerRestApiTest.java
@@ -41,6 +41,7 @@ import io.pravega.test.system.framework.Environment;
 import io.pravega.test.system.framework.SystemTestRunner;
 import io.pravega.test.system.framework.Utils;
 import io.pravega.test.system.framework.services.Service;
+
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
@@ -51,6 +52,8 @@ import javax.ws.rs.client.Invocation;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import org.glassfish.jersey.client.authentication.HttpAuthenticationFeature;
+
 import lombok.Cleanup;
 import lombok.extern.slf4j.Slf4j;
 import mesosphere.marathon.client.MarathonException;
@@ -68,6 +71,7 @@ import static javax.ws.rs.core.Response.Status.OK;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+
 @Slf4j
 @RunWith(SystemTestRunner.class)
 public class ControllerRestApiTest extends AbstractSystemTest {
@@ -81,9 +85,16 @@ public class ControllerRestApiTest extends AbstractSystemTest {
     private String resourceURl;
 
     public ControllerRestApiTest() {
+
         org.glassfish.jersey.client.ClientConfig clientConfig = new org.glassfish.jersey.client.ClientConfig();
         clientConfig.register(JacksonJsonProvider.class);
         clientConfig.property("sun.net.http.allowRestrictedHeaders", "true");
+
+        if (Utils.AUTH_ENABLED) {
+            HttpAuthenticationFeature feature = HttpAuthenticationFeature.basic("admin", "1111_aaaa");
+            clientConfig.register(feature);
+        }
+
         client = ClientBuilder.newClient(clientConfig);
     }
 
@@ -128,6 +139,7 @@ public class ControllerRestApiTest extends AbstractSystemTest {
         // TEST CreateScope POST http://controllerURI:Port/v1/scopes
         resourceURl = new StringBuilder(restServerURI).append("/v1/scopes").toString();
         webTarget = client.target(resourceURl);
+
         final CreateScopeRequest createScopeRequest = new CreateScopeRequest();
         createScopeRequest.setScopeName(scope1);
         builder = webTarget.request(MediaType.APPLICATION_JSON_TYPE);
@@ -248,7 +260,10 @@ public class ControllerRestApiTest extends AbstractSystemTest {
         final String testStream1 = RandomStringUtils.randomAlphanumeric(10);
         final String testStream2 = RandomStringUtils.randomAlphanumeric(10);
         URI controllerUri = ctlURIs.get(0);
-        try (StreamManager streamManager = new StreamManagerImpl(ClientConfig.builder().controllerURI(controllerUri).build())) {
+
+        final ClientConfig clientConfig = Utils.buildClientConfig(controllerUri);
+
+        try (StreamManager streamManager = new StreamManagerImpl(clientConfig)) {
             log.info("Creating scope: {}", testScope);
             streamManager.createScope(testScope);
 
@@ -265,14 +280,14 @@ public class ControllerRestApiTest extends AbstractSystemTest {
         final String readerGroupName2 = RandomStringUtils.randomAlphanumeric(10);
         final String reader1 = RandomStringUtils.randomAlphanumeric(10);
         final String reader2 = RandomStringUtils.randomAlphanumeric(10);
+        
         @Cleanup("shutdown")
         InlineExecutor executor = new InlineExecutor();
         Controller controller = new ControllerImpl(ControllerImplConfig.builder()
-                                     .clientConfig(ClientConfig.builder().controllerURI(controllerUri).build())
+                                     .clientConfig(clientConfig)
                                      .build(), executor);
         try (ClientFactoryImpl clientFactory = new ClientFactoryImpl(testScope, controller);
-             ReaderGroupManager readerGroupManager = ReaderGroupManager.withScope(testScope,
-                     ClientConfig.builder().controllerURI(controllerUri).build())) {
+             ReaderGroupManager readerGroupManager = ReaderGroupManager.withScope(testScope, clientConfig)) {
             final ReaderGroupConfig config = ReaderGroupConfig.builder()
                                                        .stream(Stream.of(testScope, testStream1))
                                                        .stream(Stream.of(testScope, testStream2))
--- a/test/system/src/test/java/io/pravega/test/system/MultiReaderTxnWriterWithFailoverTest.java
+++ b/test/system/src/test/java/io/pravega/test/system/MultiReaderTxnWriterWithFailoverTest.java
@@ -1,11 +1,10 @@
 /**
  * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
- *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  */
 package io.pravega.test.system;
 
@@ -13,6 +12,7 @@ import io.pravega.client.ClientConfig;
 import io.pravega.client.admin.ReaderGroupManager;
 import io.pravega.client.admin.StreamManager;
 import io.pravega.client.admin.impl.StreamManagerImpl;
+import io.pravega.client.netty.impl.ConnectionFactoryImpl;
 import io.pravega.client.stream.ScalingPolicy;
 import io.pravega.client.stream.StreamConfiguration;
 import io.pravega.client.stream.impl.ClientFactoryImpl;
@@ -95,23 +95,20 @@ public class MultiReaderTxnWriterWithFailoverTest extends AbstractFailoverTests 
         assertTrue(segmentStoreInstance.isRunning());
         log.info("Pravega Segmentstore service instance details: {}", segmentStoreInstance.getServiceDetails());
         //executor service
-        executorService = ExecutorServiceHelpers.newScheduledThreadPool(NUM_READERS + NUM_WRITERS + 2,
-                                                                        "MultiReaderTxnWriterWithFailoverTest-main");
-        controllerExecutorService = ExecutorServiceHelpers.newScheduledThreadPool(2,
-                                                                                  "MultiReaderTxnWriterWithFailoverTest-controller");
+        executorService = ExecutorServiceHelpers.newScheduledThreadPool(NUM_READERS + NUM_WRITERS + 2, "MultiReaderTxnWriterWithFailoverTest-main");
+        controllerExecutorService = ExecutorServiceHelpers.newScheduledThreadPool(2, "MultiReaderTxnWriterWithFailoverTest-controller");
+        final ClientConfig clientConfig = Utils.buildClientConfig(controllerURIDirect);
         //get Controller Uri
-        controller = new ControllerImpl(
-                                        ControllerImplConfig.builder()
-                                                            .clientConfig(ClientConfig.builder().controllerURI(controllerURIDirect).build())
-                                                            .maxBackoffMillis(5000).build(),
-                                        controllerExecutorService);
+        controller = new ControllerImpl(ControllerImplConfig.builder()
+                                                            .clientConfig(clientConfig)
+                                                            .maxBackoffMillis(5000).build(), controllerExecutorService);
         testState = new TestState(true);
         //read and write count variables
-        streamManager = new StreamManagerImpl( ClientConfig.builder().controllerURI(controllerURIDirect).build());
+        streamManager = new StreamManagerImpl(clientConfig);
         createScopeAndStream(scope, STREAM_NAME, config, streamManager);
         log.info("Scope passed to client factory {}", scope);
-        clientFactory = new ClientFactoryImpl(scope, controller);
-        readerGroupManager = ReaderGroupManager.withScope(scope,  ClientConfig.builder().controllerURI(controllerURIDirect).build());
+        clientFactory = new ClientFactoryImpl(scope, controller, new ConnectionFactoryImpl(clientConfig));
+        readerGroupManager = ReaderGroupManager.withScope(scope, clientConfig);
     }
 
     @After
@@ -128,7 +125,7 @@ public class MultiReaderTxnWriterWithFailoverTest extends AbstractFailoverTests 
         //scale the controller and segmentStore back to 1 instance.
         Futures.getAndHandleExceptions(controllerInstance.scaleService(1), ExecutionException::new);
         Futures.getAndHandleExceptions(segmentStoreInstance.scaleService(1), ExecutionException::new);
-}
+    }
 
     @Test
     public void multiReaderTxnWriterWithFailOverTest() throws Exception {
--- a/test/system/src/test/java/io/pravega/test/system/PravegaTest.java
+++ b/test/system/src/test/java/io/pravega/test/system/PravegaTest.java
@@ -9,7 +9,6 @@
  */
 package io.pravega.test.system;
 
-import io.pravega.client.ClientConfig;
 import io.pravega.client.EventStreamClientFactory;
 import io.pravega.client.admin.ReaderGroupManager;
 import io.pravega.client.netty.impl.ConnectionFactory;
@@ -36,8 +35,6 @@ import java.util.List;
 import java.util.UUID;
 import lombok.Cleanup;
 import lombok.extern.slf4j.Slf4j;
-import org.junit.Before;
-import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.Timeout;
@@ -75,34 +72,6 @@ public class PravegaTest extends AbstractReadWriteTest {
         ensureSegmentStoreRunning(zkUri, controllerUri);
     }
 
-    @BeforeClass
-    public static void beforeClass() {
-        // This is the placeholder to perform any operation on the services before executing the system tests
-    }
-
-    /**
-     * Invoke the createStream method, ensure we are able to create stream.
-     */
-    @Before
-    public void createStream() {
-
-        Service conService = Utils.createPravegaControllerService(null);
-
-        List<URI> ctlURIs = conService.getServiceDetails();
-        URI controllerUri = ctlURIs.get(0);
-
-        log.info("Invoking create stream with Controller URI: {}", controllerUri);
-        @Cleanup
-        ConnectionFactory connectionFactory = new ConnectionFactoryImpl(ClientConfig.builder().build());
-        @Cleanup
-        ControllerImpl controller = new ControllerImpl(ControllerImplConfig.builder()
-                                    .clientConfig(ClientConfig.builder().controllerURI(controllerUri).build())
-                                    .build(), connectionFactory.getInternalExecutor());
-
-        assertTrue(controller.createScope(STREAM_SCOPE).join());
-        assertTrue(controller.createStream(STREAM_SCOPE, STREAM_NAME, config).join());
-    }
-
     /**
      * Invoke the simpleTest, ensure we are able to produce  events.
      * The test fails incase of exceptions while writing to the stream.
@@ -110,14 +79,26 @@ public class PravegaTest extends AbstractReadWriteTest {
      */
     @Test
     public void simpleTest() {
-
         Service conService = Utils.createPravegaControllerService(null);
         List<URI> ctlURIs = conService.getServiceDetails();
         URI controllerUri = ctlURIs.get(0);
 
+        log.info("Invoking create stream with Controller URI: {}", controllerUri);
+
         @Cleanup
-        EventStreamClientFactory clientFactory = EventStreamClientFactory.withScope(STREAM_SCOPE, ClientConfig.builder().controllerURI(controllerUri).build());
+        ConnectionFactory connectionFactory = new ConnectionFactoryImpl(Utils.buildClientConfig(controllerUri));
+        @Cleanup
+        ControllerImpl controller = new ControllerImpl(ControllerImplConfig.builder()
+                                                                           .clientConfig(Utils.buildClientConfig(controllerUri))
+                                                                           .build(), connectionFactory.getInternalExecutor());
+
+        assertTrue(controller.createScope(STREAM_SCOPE).join());
+        assertTrue(controller.createStream(STREAM_SCOPE, STREAM_NAME, config).join());
+
+        @Cleanup
+        EventStreamClientFactory clientFactory = EventStreamClientFactory.withScope(STREAM_SCOPE, Utils.buildClientConfig(controllerUri));
         log.info("Invoking Writer test with Controller URI: {}", controllerUri);
+
         @Cleanup
         EventStreamWriter<Serializable> writer = clientFactory.createEventWriter(STREAM_NAME,
                                                                                  new JavaSerializer<>(),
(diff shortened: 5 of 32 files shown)


pbelgundi:
Failing tests:
ControllerRestAPITest
"2019-05-02 07:59:59,631 2414 [main] ERROR i.p.t.s.f.TestFrameworkException - TestFramework Exception. Type: InternalError, Details: Exception while running test method: restApiTests
java.lang.AssertionError: Create scope status expected:<201> but was:<401>
        at io.pravega.test.system.ControllerRestApiTest.restApiTests(ControllerRestApiTest.java:136)
"

AutoScaleTest
"2019-05-02 11:34:06,033 291143 [AutoScaleTest-main-3] DEBUG i.p.t.system.AbstractReadWriteTest - Writing event efe89f8b-3684-40dd-aecc-597c7713c0a3:287
2019-05-02 11:34:06,033 291143 [AutoScaleTest-main-3] DEBUG i.p.t.system.AbstractReadWriteTest - Event write count 9121
2019-05-02 11:34:06,033 291143 [grpc-default-executor-1] ERROR i.p.test.system.AbstractScaleTests - test ScaleUp failed with exception {}
io.pravega.common.util.RetriesExhaustedException: java.util.concurrent.CompletionException: io.pravega.test.system.AbstractScaleTests$ScaleOperationNotDoneException
        at io.pravega.common.util.Retry$RetryAndThrowBase.lambda$null$6(Retry.java:245)
        at java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:870)
        at java.util.concurrent.CompletableFuture$UniExceptionally.tryFire(CompletableFuture.java:852)
        at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
        at java.util.concurrent.CompletableFuture.uniAccept(CompletableFuture.java:656)
        at java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:632)
        at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
"
StreamCutsTest
"
2019-05-02 10:26:48,174 95692 [epollEventLoopGroup-4-2] DEBUG i.p.client.netty.impl.FlowHandler - Connection dropped for flow id 104
2019-05-02 10:26:48,175 95693 [main] ERROR i.p.t.s.f.TestFrameworkException - TestFramework Exception. Type: InternalError, Details: Exception while running test method: streamCutsTest
java.lang.AssertionError: Expected events read:  expected:<2000> but was:<1535>
        at io.pravega.test.system.StreamCutsTest.combineSlicesAndVerify(StreamCutsTest.java:264)
        at io.pravega.test.system.StreamCutsTest.writeEventsAndCheckSlices(StreamCutsTest.java:206)
        at io.pravega.test.system.StreamCutsTest.streamCutsTest(StreamCutsTest.java:155)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
"

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -42,7 +45,7 @@
     public static final int ALTERNATIVE_CONTROLLER_PORT = 9093;
     public static final int ALTERNATIVE_REST_PORT = 9094;
     public static final TestExecutorFactory.TestExecutorType EXECUTOR_TYPE = TestExecutorFactory.getTestExecutionType();
-
+    public static final boolean SECURITY_ENABLED = isSecurityEnabled();
```
ravisharda: `Security enabled` is an abstract term and today it means TLS and Auth are enabled, but tomorrow it might encompass other things. Please make it more precise here by breaking it down into two separate flags - one for auth and another for TLS. Since the system test framework currently only supports `auth`, you might use just the former.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -116,10 +119,36 @@ public static Service createPravegaSegmentStoreService(final URI zkUri, final UR
...
+
+    public static PravegaProperties getPravegaProperties() {
+        if (!SECURITY_ENABLED) {
+            return PravegaProperties.builder().build();
+        } else {
+            return PravegaProperties.builder().buildWithSecurity();
         }
     }
 
+    public static ClientConfig getClientConfig(URI controllerUri) {
```
ravisharda: `get...()` seems to give an impression that the returned object is pre-created/cached and that it is inexpensive to call it again and again. In this case, the code is preparing/building an object upon call, so it might make more sense to call it `build..()/prepare...()`.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -116,10 +119,36 @@ public static Service createPravegaSegmentStoreService(final URI zkUri, final UR
...
+        if (!SECURITY_ENABLED) {
+            log.info("Generating config with auth disabled.");
+            return ClientConfig.builder().controllerURI(controllerUri).build();
+        } else {
+            log.info("Generating config with auth enabled.");
+            return ClientConfig.builder()
+                               // auth
+                               .credentials(new DefaultCredentials("1111_aaaa", "admin"))
```
ravisharda: Can the credential data be overridden through environment variables and/ JVM/system properties? There should be a way to do that.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -116,10 +119,36 @@ public static Service createPravegaSegmentStoreService(final URI zkUri, final UR
...
+                               // tls
```
ravisharda: Please remove commented out code. It adds additional cognitive burden on the reader.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -146,4 +176,9 @@ public static boolean isAwsExecution() {
         return dockerConfig.trim().equalsIgnoreCase("true") ?  true : false;
 
     }
+
+    public static boolean isSecurityEnabled() {
```
ravisharda: Make the method more specific like `isAuthEnabled()`, etc.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
```
ravisharda: Since it is a new class, use `2019` here.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Map;
+import lombok.Builder;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * This class encapsulates all system properties to be passed to the operator for a Pravega deployment.
+ * */
+@Slf4j
+@Builder
+public class PravegaProperties {
```
ravisharda: Please make the class name more specific and easier to understand. `PravegaProperties` is both vague and ambiguous.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * This class overrides the lombok builder.
+     * Add all methods for building PravegaProperties (with/without Auth/TLS) to this class
+     */
+    public static final class PravegaPropertiesBuilder {
+        private static final PravegaProperties DEFAULT_INSTANCE = new PravegaProperties();
+
+        public static PravegaProperties buildWithSecurity() {
+            log.debug("Security Builder invoked.");
```
ravisharda: The utility of this log statement is suspect, assuming you have log statements down the call hierarchy. If you do need it, reduce the level to trace.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
...
+            PravegaProperties ctx = new PravegaProperties();
```
ravisharda: The variable name `ctx` looks inappropriate here.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
...
+            addControllerProperties(ctx);
+            addSegmentStoreProperties(ctx);
+            return ctx;
+        }
+
+        public static PravegaProperties build() {
+            log.debug("Default Builder invoked.");
```
ravisharda: `Trace` or remove.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        private static void addControllerProperties(PravegaProperties ctx) {
+            //auth
+            ctx.addProperty("controller.auth.enabled", "true");
+            ctx.addProperty("controller.auth.userPasswordFile", "/opt/pravega/conf/passwd");
+            ctx.addProperty("controller.auth.tokenSigningKey", "secret");
+
+            //tls
+            /*
```
ravisharda: Remove commented out code.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+
+        private static void addSegmentStoreProperties(PravegaProperties ctx) {
+            // auth
+            ctx.addProperty("autoScale.authEnabled", "true");
+            ctx.addProperty("autoScale.tokenSigningKey", "secret");
+
+            // tls
+            /*
```
ravisharda: Remove commented out code.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        public static PravegaProperties build() {
+            log.debug("Default Builder invoked.");
+            return DEFAULT_INSTANCE;
+        }
+
+        private static void addControllerProperties(PravegaProperties ctx) {
+            //auth
+            ctx.addProperty("controller.auth.enabled", "true");
```
ravisharda: Should we move the names of the properties to a constants file? That might make it easier to inspect them, as well as modify them as these things change overtime in Pravega.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: Taking credentials through environment variables is already supported by ClientConfigBuilder. Could use that builder when required.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: Intentionally left this in, as it needs to be added in near future.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: Same method would be used once we have Auth+TLS environment for security system tests.
Does not make sense to have different methods for AuthEnabled and TLSEnabled unless we plan to have different environment setups with only auth enabled or only tls enabled.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: ok will change.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: ok

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: Don't expect names to change too often, as we always need to ensure backward compatibility when changing names. Could be done in the next set of PRs when we have more properties to deal with.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: As per our earlier discussions and documentation, we plan to have only 1 environment for system tests with "security" enabled (for now this will be only auth enabled). Once I have tested with Auth +TLS enabled and it starts working fine, it will be auth + tls enabled pravega environment, hence the name isSecurityEnabled.
This variable controls running of all system tests with/without security.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: see my earlier comment.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: Ok. will change.

test/system/src/test/java/io/pravega/test/system/PravegaTest.java
```diff
@@ -72,55 +69,42 @@
      */
     @Environment
     public static void initialize() {
+        log.info("Environment: Invoking tests with security enabled: {}", Utils.SECURITY_ENABLED);
```
ravisharda: Did you mean to say something like "Setting up environment with auth enabled"? The existing log message can be misleading if the environment creation fails (since it says "Invoking tests..."). Also, why are you adding the "Environment" prefix to the log statement? I think it is unnecessary as method names can be automatically logged via an appropriate log format configuration via appender encoder pattern in the logging configuration file.

test/system/src/test/java/io/pravega/test/system/PravegaTest.java
```diff
@@ -72,55 +69,42 @@
...
-     * Invoke the createStream method, ensure we are able to create stream.
+     * Invoke the simpleTest, ensure we are able to produce  events.
+     * The test fails incase of exceptions while writing to the stream.
+     *
      */
-    @Before
-    public void createStream() {
+    @Test
+    public void simpleTest() {
+        log.info("simpleTest:: with security enabled: {}", Utils.SECURITY_ENABLED);
```
ravisharda: Please avoid adding method names in log statements. Also, should this really be an info level log? Debug or trace level looks more appropriate to me here. 

Can we use a better name for the test - one that reveals the intent of the test?

test/system/src/test/java/io/pravega/test/system/ReadTxnWriteScaleWithFailoverTest.java
```diff
@@ -106,18 +106,18 @@ public void setup() {
                 "ReadTxnWriteScaleWithFailoverTest-main");
         controllerExecutorService = ExecutorServiceHelpers.newScheduledThreadPool(2,
                 "ReadTxnWriteScaleWithFailoverTest-controller");
+        ClientConfig clientConfig = Utils.getClientConfig(controllerURIDirect);
         //get Controller Uri
         controller = new ControllerImpl(ControllerImplConfig.builder()
-                                    .clientConfig( ClientConfig.builder().controllerURI(controllerURIDirect).build())
+                                    .clientConfig(clientConfig)
                                     .maxBackoffMillis(5000).build(),
```
ravisharda: This value repeats in several places. Can we move it as a constant to some file?

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
ravisharda: In this PR the support is for `auth` only, so as of now using a "security enabled" flag is misleading. Sure, I understand that the plan is to add support for TLS as well, but the code needs to represent what's in there today - not what it is supposed to to be tomorrow. A good way to tackle this might be to use separate flags but bring together using a method like isSecurityEnabled() for convenience.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
ravisharda: Sounds good.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
ravisharda: I'd rather that we remove unused code. You might want to store it in your local repository in some way.

test/system/src/test/java/io/pravega/test/system/ReadTxnWriteScaleWithFailoverTest.java
pbelgundi: Didn't get this..??

test/system/src/test/java/io/pravega/test/system/PravegaTest.java
pbelgundi: Will remove method name from log and change to debug.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
ravisharda: To make this intent clear, it might be a good idea to create more specific methods like `isAuthEnabled` and then use a wrapper method like `isSecurityEnabled()` that brings them together for convenience and making the intention clear. That being said, I'll leave it up to you.

test/system/src/test/java/io/pravega/test/system/PravegaTest.java
pbelgundi: Added this log to make sure system property (securityEnabled) set by gradle when starting the tests, is received correctly inside initialize().
Nevertheless, the test method does not automatically receive this value unless it is set on the test pod at the time of creation... see changes to class K8SequentialExecutor

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
ravisharda: Sounds good.

test/system/src/test/java/io/pravega/test/system/PravegaTest.java
ravisharda: Sure. Having a log statement here makes sense. The issue I have is the message in the log entry, which is misleading (At this point, it is not that the tests are being invoked.)

test/system/src/test/java/io/pravega/test/system/ReadTxnWriteScaleWithFailoverTest.java
ravisharda: I meant the number `5000`. It repeats everywhere, so it might make sense to use a descriptive constant to represent it, in case we need to change this later.

build.gradle
```diff
@@ -818,7 +819,10 @@ project('test:system') {
         if (System.getProperty("imageVersion") == null) {
             systemProperty "imageVersion", "INVALID_IMAGE_VERSION"
         }
-
+        
+        systemProperty "securityEnabled", System.getProperty("securityEnabled", "false")
+        println "Is security enabled for tests:" + System.getProperty("securityEnabled")
```
shrids: The print line will printed for all for all the gradle tasks, this is because this line will be executed during configuration phase of gradle's build lifecycle.
We should remove it or associate it with a gradle action...

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -116,8 +119,31 @@ public static Service createPravegaSegmentStoreService(final URI zkUri, final UR
                 return  new PravegaSegmentStoreDockerService(serviceId, zkUri, hdfsUri, contUri);
             case KUBERNETES:
             default:
-                return new PravegaSegmentStoreK8sService(serviceId, zkUri);
+                return new PravegaSegmentStoreK8sService(serviceId, zkUri, getPravegaProperties());
+        }
+    }
+
+    public static PravegaProperties getPravegaProperties() {
```
shrids: we can change the scope to private.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
shrids: The credentials will be read from the env/ system properties only if the `credentials`  is not explicitly set.
ref: `io.pravega.client.ClientConfig.ClientConfigBuilder#extractCredentials` 
We could always read the credentials via the environment, right? This would avoid the hardcoding of credentials in the framework.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -146,4 +173,10 @@ public static boolean isAwsExecution() {
         return dockerConfig.trim().equalsIgnoreCase("true") ?  true : false;
 
     }
+
```
shrids: we can change the access level to private.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,100 @@
...
+            ctx.addProperty("controller.auth.userPasswordFile", "/opt/pravega/conf/passwd");
+            ctx.addProperty("controller.auth.tokenSigningKey", "secret");
+        }
+
+
+        private static void addSegmentStoreProperties(PravegaProperties ctx) {
+            // auth
+            ctx.addProperty("autoScale.authEnabled", "true");
+            ctx.addProperty("autoScale.tokenSigningKey", "secret");
```
shrids: should we make the `tokenSigningKey` configurable?

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: Why? That will break calls to this method from other classes.
This method was not added in this PR.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: Making configurable would help if we're going to execute system tests with different security setups.
Right now that is ruled out as all tests would just execute on 2 environments - 1 with all security enabled and 2 with security disabled.
Could make it configurable at a later point if we have the need to have multiple security environments.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: Taking credentials through environment variables would require passing these variables ( both username and password) in clear text via gradle at run time. This is not very different from directly hardcoding it in the code.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: done

build.gradle
pbelgundi: Sure will remove this.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: ok. done

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: To make it truly configurable should be read from a properties file supplied at runtime. This can be done if needed at a later stage when we have the need to support mulitple security environments.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
shrids: The comment was for the method `isSecurityEnabled()`, we can change the access level to private

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
shrids: ok.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
shrids: ok.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,100 @@
...
+            strBuilder.append(e);
+            strBuilder.append(":");
+            strBuilder.append(this.getProperties().get(e));
+            strBuilder.append("|");
+        }
+        return strBuilder.toString();
+    }
+
+    /**
+     * This class overrides the lombok builder.
```
shiveshr: why do we need to annotate with builder on PravegaProperties class if we want to override it ?

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,100 @@
...
+            props.addProperty("controller.auth.enabled", "true");
+            props.addProperty("controller.auth.userPasswordFile", "/opt/pravega/conf/passwd");
+            props.addProperty("controller.auth.tokenSigningKey", "secret");
+        }
+
+
+        private static void addSegmentStoreProperties(PravegaProperties props) {
+            // auth
+            props.addProperty("autoScale.authEnabled", "true");
+            props.addProperty("autoScale.tokenSigningKey", "secret");
```
shiveshr: dont we need to pass password file or some other configuration that enable client to talk to controller?

test/system/src/test/java/io/pravega/test/system/ControllerRestApiTest.java
```diff
@@ -304,12 +319,15 @@ public void restApiTests() {
         assertEquals("Get readergroup properties status", OK.getStatusCode(), response.getStatus());
         ReaderGroupProperty readerGroupProperty = response.readEntity(ReaderGroupProperty.class);
         log.debug("ReaderGroup properties {}", readerGroupProperty);
+
```
shiveshr: please revert unchanged lines

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: Sure :)

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: For basic(default) auth, file is already present in the controller docker image at the said location.
For system tests, more complex type of security setups won't be supported. 
See: https://github.com/pbelgundi/pravega-resources/blob/master/Executing_Security_System_Tests.md

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: "override" can work only if it is annotated. 
Without the annotation I cannot call: PravegaProperties.builder().build();
Though I could still call PravegaProperties.PravegaPropertiesBuilder.build(), I thought former looks more elegant.

build.gradle
```diff
@@ -788,6 +788,7 @@ project('test:system') {
 
     task startK8SystemTests(type: Test) {
         tasks.withType(Exec) { environment "tier2Type", System.getProperty("tier2Type", "nfs") }
+
```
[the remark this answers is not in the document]
pbelgundi: ok

test/system/src/test/java/io/pravega/test/system/ReadTxnWriteScaleWithFailoverTest.java
pbelgundi: There are a lot of such constants all over system tests and that need refactor. Can be done as a separate refactoring exercise rather than include in this PR which is not related.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: I think the name represents exactly what it is.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: Done.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: Removed the logs.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: done

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: done

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: removed logs

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: done

test/system/src/test/java/io/pravega/test/system/PravegaTest.java
pbelgundi: removed log

ravisharda:
@pbelgundi I don't see any new commits in this PR for the changes you might have made based on the review comments: this PR only has a single commit. Could you please push the changes?

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: done

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: done

test/system/src/test/java/io/pravega/test/system/MultiReaderTxnWriterWithFailoverTest.java
```diff
@@ -99,19 +100,20 @@ public void setup() {
                                                                         "MultiReaderTxnWriterWithFailoverTest-main");
         controllerExecutorService = ExecutorServiceHelpers.newScheduledThreadPool(2,
                                                                                   "MultiReaderTxnWriterWithFailoverTest-controller");
+        final ClientConfig clientConfig = Utils.getClientConfig(controllerURIDirect);
         //get Controller Uri
         controller = new ControllerImpl(
```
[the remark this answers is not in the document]
pbelgundi: done.

build.gradle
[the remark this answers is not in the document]
pbelgundi: done

test/system/src/test/java/io/pravega/test/system/ControllerRestApiTest.java
pbelgundi: ok. done.

pbelgundi:
> @pbelgundi I don't see any new commits in this PR for the changes you might have made based on the review comments: this PR only has a single commit. Could you please push the changes?

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
ravisharda: Ok, if you are convinced.

ravisharda:
LGTM.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -130,6 +156,7 @@ public static Service createPravegaSegmentStoreService(final URI zkUri, final UR
      * Default value is false
      * @return true if skipServiceInstallation is set, false otherwise.
      */
+
```
fpj: Please remove this new line.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,98 @@
...
+            return DEFAULT_INSTANCE;
+        }
+
+        private static void addControllerProperties(PravegaProperties props) {
+            //auth
+            props.addProperty("controller.auth.enabled", "true");
+            props.addProperty("controller.auth.userPasswordFile", "/opt/pravega/conf/passwd");
+            props.addProperty("controller.auth.tokenSigningKey", "secret");
+        }
+
```
fpj: Remove new line.

test/system/src/test/java/io/pravega/test/system/MultiReaderTxnWriterWithFailoverTest.java
```diff
@@ -1,18 +1,19 @@
 /**
  * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
- *
+ * <p>
```
fpj: Please undo these license header changes.

test/system/src/test/java/io/pravega/test/system/MultiReaderTxnWriterWithFailoverTest.java
```diff
@@ -95,23 +96,21 @@ public void setup() {
...
+        final ClientConfig clientConfig = Utils.buildClientConfig(controllerURIDirect);
         //get Controller Uri
-        controller = new ControllerImpl(
-                                        ControllerImplConfig.builder()
-                                                            .clientConfig(ClientConfig.builder().controllerURI(controllerURIDirect).build())
+        controller = new ControllerImpl(ControllerImplConfig.builder()
+                                                            .clientConfig(clientConfig)
                                                             .maxBackoffMillis(5000).build(),
-                                        controllerExecutorService);
+                controllerExecutorService);
```
fpj: Please undo this reformatting change.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        return strBuilder.toString();
+    }
+
+    /**
+     * This class overrides the lombok builder.
+     * Add all methods for building PravegaProperties (with/without Auth/TLS) to this class
+     */
+    public static final class PravegaPropertiesBuilder {
+        private static final PravegaProperties DEFAULT_INSTANCE = new PravegaProperties();
```
fpj: I'm a bit confused about this declaration, it feels like `builder().build()` will always return the same instance, while `builder().buildWithSecurity()` returns a new instance each time. Can you elaborate on the behavior you are after with this `DEFAULT_INSTANCE`?

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
pbelgundi: ok

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: DEFAULT_INSTANCE has properties that will result in a pravega cluster being created without security (security disabled).
buildWithSecurity() returns an instance with all properties including those required for security enablement.
Both are methods return immutable instance of PravegaProperties.  
buildWithSecurity() could be optimized to return a static instance instead of creating a new instance every time.

Still better would be to read all properties from a properties file in the resources folder. There could be multiple resource files, one for each deployment flavour (no security, with auth, with tls+auth) and each builder would read the appropriate resource file.

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: ok

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
fpj: > DEFAULT_INSTANCE has properties that will result in a pravega cluster being created without security (security disabled).

perhaps for the needs of our current tests, this is sufficient. it doesn't seem to be flexible, though. the `PravegaProperties` instance is static and final, which does not enable any customization. I agree that changing properties via code is not ideal, and config files might work, but I wonder if we won't have the need of configuring Pravega separately for different test cases.

to sum up, it seems to work fine for the current state of affairs, but it might be limiting for our future use, and I'd like to know your view since you have been working on this code and thinking about it. what should we do?

test/system/src/test/java/io/pravega/test/system/MultiReaderTxnWriterWithFailoverTest.java
pbelgundi: ok

pbelgundi: ok

test/system/src/test/java/io/pravega/test/system/MultiReaderTxnWriterWithFailoverTest.java
pbelgundi: ok

test/system/src/main/java/io/pravega/test/system/framework/services/PravegaProperties.java
pbelgundi: I have removed the 'PravegaProperties' class and added a method to Utils class to read properties from the appropriate properties file(s) under resources folder.

PR #3693, opened 2019-04-25
Issue 3669: Update default TLS material

### LINKED ISSUE #3669: Improve the default TLS/PKI material
**Problem description**
The default TLS/PKI material included in the installation (standalone mode as well as distributed mode) works, but suffers from the following problems:

1. The default server keystore jks (`standalone.keystore.jks`) file should have had two entries: the first one representing the CA signed certificate of the server, and the second one representing the CA. It only contains one today: the latter. Also, the entry type of the certificate is incorrect. 

    ```
    $ keytool -list -v -keystore standalone.keystore.jks -storepass 1111_aaaa
    ...
    Alias name: caroot
    Entry type: PrivateKeyEntry
    ```
2. The CA certificate and key files are not packaged in the installation, while the CA is referenced in the material. Also, if you want to generate and sign new certificates, you don't have a ready CA to work with. 

3. The subject name of the server certificate is `localhost`, which means you can't use a remote client unless you disable hostname verification. A more general hostname like `server.pravega.io` would allow you to define a mapping of this hostname to IP addresses on a deployment environment through the OS hosts file and use it for remote clients.

4. The names of the files are confusing and hard to decipher. The names of the files should be self-explanatory. List the contents of the `pravega/config` directory and see if you can make sense of the PKI material out there.

**Problem location**
`/path/to/pravega/config`

ravisharda:
**Change log description**  
Update the default TLS material in these respects: 
1. Fix the server keystore to have a CA-signed server certificate as well as the CA's certificate. Fix the entry types too. (See issue #3669 for more details).

2. Include a CA (in the form of a self-signed certificate and a private key) in the config directory. 
   
    The CA cert is present in the client trust store used as truststore by server components when playing the role of a client while connecting to other server components. It is also used as the truststore in tests, to use `chain of trust`. Earlier, the server certificate was being used as truststure for establishing direct trust.
3. Add multiple subject names - `localhost` and a generic `server.pravegastack.io` - through the SAN extension field in the certificate. The latter will allow clients to connect to a remote server over TLS with hostname verification enabled, using a hostname/IP address mapping entry in the OS hosts file. 
   
   It should be noted that this is useful in the case of Docker Compose based deployments, but not standalone mode. Standalone mode Controller gRPC server doesn't listen on a non-loopback address (unlike the REST server, which does), so a remote client cannot connect to it. 

4. Make the names of the files more specific and easy to understand. Also, use correct file extensions for certificates and keys. The .crt and .key files are standard extensions for certificates and keys for both PEM and DER encoded files. We use PEM encoded files. (Read more about it [here](https://support.ssl.com/Knowledgebase/Article/View/19/0/der-vs-crt-vs-cer-vs-pem-certificates-and-how-to-convert-them)). 

**Purpose of the change**  
Resolves #3669.

**What the code does**  
* Fixes the TLS material as explained above.
* Moves the default security config strings to a class with constants. Use those constants in the tests. This was necessary, as the strings were spread out throughout the code base, and it was hard to ensure any changes to names/paths were applies globally.
* Updates tests to make use of the new material.

**How to verify it**  
All tests should pass. 

Also, inspect the TLS material using these commands (using the default password on password prompts)

| Filename | Purpose | Inspecting |
|:----------:|:----------|:------------|
| `ca-cert.crt` | PEM-encoded X.509 certificate of the CA | `openssl x509 -in ca-cert.crt -text -noout` |
| `ca-key.key` | PEM-encoded file containing the CA's encrypted private key | `openssl pkcs8 -inform PEM -in ca-key.key -topk8` |
| `client.truststore.jks` | A password-protected truststore file containing the CA's certificate | `keytool -list -v -keystore client.truststore.jks` |
| `server.keystore.jks` | The server keystore file in .jks format | `keytool -list -v -keystore server.keystore.jks` |
| `server-cert.crt` | PEM-encoded CA-signed server certificate file | `openssl x509 -in server-cert.crt -text -noout` |
| `server-key.key` | PEM-encoded file containing the server's encrypted private key | `openssl pkcs8 -inform PEM -in server-key.key -topk8` |


ravisharda:
This PR is not ready for review yet. I'm working on fixing the build issue.

ravisharda:
All the tests are now passing in the CI build. This PR is ready for review now. 

The code coverage error showing up is a non-issue in this case, as the code it is pointing to is a class with constants, and IMO it doesn't make sense to add unit tests for them.

EronWright:
Are the Java keystore files in JKS or in PKCS12 format?   Note that JKS is deprecated as of JDK 8.

ravisharda:
> Are the Java keystore files in JKS or in PKCS12 format? Note that JKS is deprecated as of JDK 8.

@EronWright thanks for pointing that out. The keystores and truststores are still in JKS (`.jks`) format. We should certainly move to PKCS12 (`.p12` or `.pfx`) format sooner or later, as you have suggested. 

My reading of http://openjdk.java.net/jeps/229 (which changes the default keystore type from `JKS` to `PKCS12`, in Java 9) indicates that PKCS12 files should work as drop-in replacements for JKS files. But, to be sure, an extensive verification exercise will need to be done, including for TLS-enabled interfaces in Zookeeper, Bookkeeper, etc. Since, JKS format still works as of Java 11, I believe we have some leeway on when we perform that exercise. I'll create a separate enhancement issue to support PKCS12 files for that exercise.

common/src/main/java/io/pravega/common/SecurityConfigDefaults.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static final String TLS_CA_PRIVATE_KEY_PATH = "../config/ca-key.key";
+
+    public static final String AUTH_HANDLER_INPUT_PATH = "../config/passwd";
+
+    public static final String AUTH_ADMIN_USERNAME = "admin";
+
+    public static final String AUTH_ADMIN_PASSWORD = "1111_aaaa";
+
+
```
fpj: Remove unnecessary new line.

common/src/main/java/io/pravega/common/SecurityConfigDefaults.java
```diff
@@ -0,0 +1,44 @@
+/**
+ * Copyright (c) 2019 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.common;
```
fpj: `common` does not seem to be the right place for these definitions. `common` is more about classes that can be used across components of Pravega and even outside the project. these definitions here are fairly specific to Pravega, so I'd say it should be in `shared` instead.

common/src/main/java/io/pravega/common/SecurityConfigDefaults.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Holds default security configuration values.
+ */
+public class SecurityConfigDefaults {
```
fpj: I see these definitions being used in tests and standalone. are those defaults for distributed Pravega too?

common/src/main/java/io/pravega/common/SecurityConfigDefaults.java
ravisharda: Fixed.

common/src/main/java/io/pravega/common/SecurityConfigDefaults.java
ravisharda: They are set as the defaults in standalone mode config file. They can also be used in distributed mode, but not configured as defaults in the configuration files used in distributed mode.

common/src/main/java/io/pravega/common/SecurityConfigDefaults.java
ravisharda: Sure, thanks for pointing that out. I read the readme file there and noticed that idea as well, but got misled by a couple of other examples. 

I tried moving it to the `shared` > `cluster` subproject, but moving there would cause a series of build changes. After reflecting a bit and looking around for other test utilities/helpers, I felt that the best home for it is in `pravega` > `test` > `testcommon` Gradle subproject in the package `io.pravega.test.common`. I've made that change in the latest commit. Hope that makes sense.

fpj: What about `io.pravega.shared.test`? We don't have a `test` package under `shared`, but we could create one in this PR and it would make more sense to me because of the following:

- `shared` is supposed to contain classes that are pravda-specific and shared across components.
- `test.common` is borderline ok for this. Even though we are using standard security primitives, I don't see these constants being useful to anyone outside pravega. To determine whether goes into `common` or `test.common`, I ask myself: could anyone be interested in using this outside the context of pravega?

ravisharda: Given the purpose of the `pravega/shared` Gradle subproject, it makes sense to have a `pravega/shared/test` sub-project with the classes under `io.pravega.shared.test` package/namespace. 

That being said, all the classes that are used across tests in other Gradle sub-projects lie inside of the `pravega/test/testcommon` subproject today inside the namespace `io.pravega.test.common`. Witness the `io.pravega.test.common.AssertExtensions` class that lies in the same package as `io.pravega.test.common.SecurityConfigDefaults` class that I have introduced. This scheme also makes sense to me, as all test related stuff is cohesively located in one place. Arguably, this scheme is even better as it favors functional cohesion over logical cohesion. 

I think we move `SecurityConfigDefaults` elsewhere, it'd make sense to move other existing classes in testcommon, as well (with consequential build script changes as well). Doing so will take the scope of this PR way well beyond the original intention. 

Therefore, I'd like to propose we keep it as is right now. What do you say?

fpj: `io.pravega.test.common.AssertExtensions` is something that I see others outside Pravega benefitting from. `test/common` is the right place for it.

I can't recall any class for tests like `io.pravega.test.common.SecurityConfigDefaults`, so I'm tempted to say that it is the first time we are encountering this situation.

But, if you tell me that we need deeper changes than just moving the file to have those definitions in `shared/test`, then I can be convinced to postpone the reorganization. Otherwise, I'd rather avoid the tech debt.

ravisharda: A lot of tests across several Gradle subprojects use the `SecurityConfigDefaults`. There will be deeper changes required to move it (like Gradle build changes, etc.), and therefore I'd request that we postpone this reorganization.

PR #3701, opened 2019-04-25
Issue 3637: (SegmentStore) Enable compaction on all Controller Table Segments

### LINKED ISSUE #3637: Enable Table Segment Compaction
**Problem description**
Enable Table Segment Compaction:
- Default MinUtilization=75%;
- Set a sensible Tier 2 Rolling Segment Size.

Run a few system tests to verify this doesn't cause adverse effects

**Problem location**
Tables.

None.

andreipaduroiu:
**Change log description**  
- Enabled compaction on all Table Segments. Set MIN_UTILIZATION to 75% and ROLLOVER_SIZE to about 40MB (40 * max entry size).

**Purpose of the change**  
Fixes #3637.

**What the code does**  
- Set Segment Attributes to non-default values that enable compaction:
    - `MIN_UTILIZATION` set to 75 which means that once the utilization metric falls below 75%, a compaction is triggered.
    - `ROLLOVER_SIZE` set to around 40MB so that data can be truncated out of Tier 2. A smaller value would not have been wise since it would have caused frequent rollovers in Tier 2 for these segments.

**How to verify it**  
All unit tests must pass.
A few system tests runs must pass (that verify failover and good functioning after recovery).
Longevity tests should also work well (since they will end up creating enough data to trigger this frequently).


andreipaduroiu:
System test run 196 passed

PR #3703, opened 2019-04-26
Issue 3702: Prevent suppression of stack traces in tests

### LINKED ISSUE #3702: Prevent suppression of stack traces in tests
**Problem description**
Sometimes tests have the following pattern:
```
try {
 // some operation
} catch (SomeException e) {
    Assert.fail("should not get an exception");
}
```
This is unnecessarily verbose, but also prevents the stack trace or message of the exception from being logged. If the exception is unexpected this is critical debugging information that we could miss. 

**Problem location**
Numerous files

Remove the try catch and just throw the exception.

tkaitchuck:
**Change log description**  
* Prevent tests from suppressing exception stack traces 

**Purpose of the change**  
Fixes #3702

**What the code does**  
Removes unneeded try catch code from tests. No functional changes.

### DIFF
--- a/controller/src/test/java/io/pravega/controller/server/ControllerServiceStarterTest.java
+++ b/controller/src/test/java/io/pravega/controller/server/ControllerServiceStarterTest.java
@@ -53,34 +53,21 @@ public abstract class ControllerServiceStarterTest {
     }
 
     @Before
-    public abstract void setup();
+    public abstract void setup() throws Exception;
 
     @After
-    public abstract void tearDown();
+    public abstract void tearDown() throws Exception;
 
     @Test
-    public void testStartStop() {
+    public void testStartStop() throws URISyntaxException {
         Assert.assertNotNull(storeClient);
         ControllerServiceStarter starter = new ControllerServiceStarter(createControllerServiceConfig(), storeClient, 
                 SegmentHelperMock.getSegmentHelperMockForTables(executor));
         starter.startAsync();
-
-        try {
-            starter.awaitRunning();
-        } catch (IllegalStateException e) {
-            log.error("Error awaiting starter to get ready");
-            Assert.fail("Error awaiting starter to get ready");
-        }
+        starter.awaitRunning();
 
         // Now, that starter has started, perform some rpc operations.
-        URI uri;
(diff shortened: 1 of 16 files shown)


fpj:
I have posted a comment in the issue, to see if we can agree on the guidelines for handling exceptions in test code.

controller/src/test/java/io/pravega/controller/server/PravegaTablesControllerServiceMainTest.java
```diff
@@ -104,46 +92,22 @@ public void testZKSessionExpiry() throws Exception {
...
-            Assert.fail("Failed waiting for controllerServiceStarter to get ready");
-            return;
-        }
+        controllerServiceStarter.awaitRunning();
 
         assertEquals(1, clientQueue.size());
         CuratorFramework curatorClient = (CuratorFramework) clientQueue.poll().getClient();
         // Simulate ZK session timeout
         // we will submit zk session expiration and 
-        CompletableFuture.runAsync(() -> {
```
shiveshr: this was originally injecting the session expiration and making the zk client call asynchronously. 

although looking at the code, it doesnt make difference.

controller/src/test/java/io/pravega/controller/server/ControllerServiceStarterTest.java
```diff
@@ -53,34 +53,21 @@
...
         // Now, that starter has started, perform some rpc operations.
-        URI uri;
-        try {
-            uri = new URI( (enableAuth ? "tls" : "tcp") + "://localhost:" + grpcPort);
-        } catch (URISyntaxException e) {
-            log.error("Error creating controller URI", e);
-            Assert.fail("Error creating controller URI");
-            return;
-        }
+        URI uri = new URI( (enableAuth ? "tls" : "tcp") + "://localhost:" + grpcPort);
```
RaulGracia: Extra space between parentheses.

fpj: This is not adding the space, it is from the previous code, no reformatting is required here.

PR #3711, opened 2019-04-26
Issue 3710: (SegmentStore) Delete Segment if Empty Bug Fix

### LINKED ISSUE #3710: TableSegment.delete(mustBeEmpty==true) fails to detect empty tables in all situations.
**Problem description**
Consider this scenario, executed in short sequence (so all operations go into the cache; no indexing involved):
1. Update Key1=Value1
2. Delete Key1
3. Update Key2=Value2
4. Delete(mustBeEmpty=true) succeeds.

The table segment is deleted incorrectly in this scenario as it is not empty (`Key2` is still present).

**Problem location**
ContainerKeyIndex/ContainerKeyCache/SegmentKeyCache.

Count correctly.

andreipaduroiu:
**Change log description**  
Fixed a bug in `ContainerKeyIndex` where it would incorrectly determine that a Table Segment is empty when it actually is not.

**Purpose of the change**  
Fixes #3710.

**What the code does**  
Cause:
- The cause for this was an incorrect way of calculating the number of Table Entries in a Segment, given both indexed and non-indexed entries.
- The previous calculation methodology was incorrectly assuming that all Table Key removals recorded in the cache are for keys that actually exist. 
    - While this may be true for conditional removals, unconditional removals perform no precondition checks and simply record the action (unconditionally removing an inexistent key is a valid operation)
- All removals were recorded in the `SegmentKeyCache` and when asked to calculate a delta of bucket updates (how many buckets added minus how many removed) this calculation would be thrown off by this.

This has been changed in the following way:
- The `ContainerKeyIndex` fetches the list of all bucket updates from the tail. 
- It filters out all Buckets that are marked for removal and queries the Index to see if they previously existed. Only in that case can they be discounted from the total number of buckets.

**How to verify it**  
Added a new unit test to cover this scenario.

### DIFF
--- a/segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImplTests.java
+++ b/segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImplTests.java
@@ -14,6 +14,7 @@ import io.pravega.common.Exceptions;
 import io.pravega.common.concurrent.Futures;
 import io.pravega.common.util.ArrayView;
 import io.pravega.common.util.AsyncIterator;
+import io.pravega.common.util.ByteArraySegment;
 import io.pravega.common.util.HashedArray;
 import io.pravega.segmentstore.contracts.AttributeUpdate;
 import io.pravega.segmentstore.contracts.ReadResult;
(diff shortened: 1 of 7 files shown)


segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImplTests.java
```diff
@@ -122,6 +123,51 @@ public void testCreateDelete() {
                 ex -> ex instanceof StreamSegmentNotExistsException);
     }
 
+    /**
+     * Tests the ability to delete a TableSegment, but only if it is empty.
+     */
+    @Test
+    public void testDeleteIfEmpty() {
```
fpj: The new logic to count buckets in `ContainerKeyIndex` involves tail and indexed buckets. It is not clear to me that this test case is making sure that buckets are indexed, and consequently, that all cases in `ContainerKeyIndex.isTableSegmentEmpty` are being exercised. Could you clarify, please?

andreipaduroiu: The point is not to count buckets, but to figure out if a table segment is empty or not. The old technique was trying to count buckets and see if the result was greater than zero, which is harder to do correctly in all cases. But since all I need is to figure out if a table is empty, the problem is simpler and can be reduced to a number of cases:
- If the tail index (cache) has at least one update, then regardless of whether we also have deletions or other data in the segment, it is not empty.
- Otherwise, if the tail index has no deletions, then we need to see how many entries are in the segment. If more than zero, then it is not empty, otherwise it's empty.
- If we have at least one deletion, then it must figure out if those deletions are for existing keys or not. It needs to discount those deletions for inexistent keys and subtract the remaining count from the number of entries in the table.

DeleteIfEmpty is already tested throughout the rest of this test suite. Every test deletes the table segment at the end and makes sure it's empty. This particular test case covers the remaining of the scenarios. If you run this in the IDE with test coverage you'll see all the lines in that method are covered (codecov.io disagrees but the IDE is more accurate (I can hit breakpoints in there too)).

fpj: The code coverage report actually does show all cases being covered, and thanks for confirming via IDE as well.

PR #3727, opened 2019-05-01
Issue 3726: (SegmentStore) StorageWriter can't handle updating attributes after a segment merge

### LINKED ISSUE #3726: StorageWriter can't handle Updating Attributes after Merging a Segment
**Problem description**
If a `MergeStreamSegmentOperation` is followed by a `UpdateAttributesOperation`, the `StorageWriter` will enter an error state from which it cannot recover until a container failover. It manifests itself using this exception:
```
java.lang.IllegalArgumentException: startOffset must refer to an offset beyond the Segment's StorageLength offset.
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:141)
	at io.pravega.segmentstore.server.reading.StreamSegmentReadIndex.readDirect(StreamSegmentReadIndex.java:626)
	at io.pravega.segmentstore.server.reading.ContainerReadIndex.readDirect(ContainerReadIndex.java:184)
	at io.pravega.segmentstore.server.writer.StorageWriterFactory$StorageWriterDataSource.getAppendData(StorageWriterFactory.java:146)
	at io.pravega.segmentstore.server.writer.SegmentAggregator.getFlushArgs(SegmentAggregator.java:820)
	at io.pravega.segmentstore.server.writer.SegmentAggregator.lambda$flushFully$11(SegmentAggregator.java:665)
	at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:712)
	at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:681)
	at io.pravega.common.concurrent.Futures.runOrFail(Futures.java:572)
	at io.pravega.common.concurrent.Futures$Loop.run(Futures.java:725)
```

**Problem location**
`SegmentAggregator.java`

This is caused by the `addUpdateAttributesOperation` in `SegmentAggregator` which reuses or creates an `AggregatedAppend` using the current length of the segment in Storage. It reuses if the last operation is already an `AggregatedAppend`, otherwise it creates a new one. This works well when there is already an `AggregatedAppend` created by a `StreamSegmentAppendOperation` (which sets the offset properly, and it is reused), but if there is no such operation in the current queue, it creates a new one. The value it sets as offset may not be accurate if there are other non-append/attribute operations in the queue, such as `MergeStreamSegmentOperation`.

This should be fixed so that it picks up the correct offset.

andreipaduroiu:
**Change log description**  
Fixed a bug in SegmentAggregator where a MergeTransactionOperation followed by an UpdateAttributesOperation would cause the SegmentAggregator to request invalid data from the ReadIndex and prevent progress from being made.

**Purpose of the change**  
Fixes #3726.

**What the code does**  
- The SegmentAggregator aggregates multiple operations of similar types together. This includes `UpdateAttributesOperation` and `SegmentAppendOperation`, but it does not include `MergeSegmentOperation`s.
- The first operation that is included will determine the starting offset of the resulting `AggregatedOperation`. For `SegmentAppendOperations` this works well, since `SegmentAppendOperation`s have well defined offsets. However, `UpdateAttributeOperations` do not, hence the code was assigning the current length of the Segment in Tier 2.
- This would be a problem in the following case:
    - A `MergeSegmentOperation` is queued up, then an `UpdateAttributesOperation`, then a `SegmentAppendOperation`
    - In the beginning the StorageLength of the segment is L1. 
    - As such, the `MergeSegmentOperation` also has an offset of L1 (correct), however, when queuing up the `UpdateAttributesOperation`, its corresponding offset would also be set to L1 (the merge hasn't been applied yet). 
    - The subsequent `SegmentAppendOperation` will be aggregated into this one as well, and its own offset would be disregarded.
    - When the `MergeSegmentOperation` is applied to Tier 2, the storage length of the segment would be incremented by M1 to become L1+M1.
    - When needing to flush the `AggregatedAppendOperation` made up of `UpdateAttributesOperation` and the `SegmentAppendOperation`, since that operation has the offset set to L1, it would attempt to read data from the ReadIndex beginning at L1, and it is correctly rejected due to L1 being less than L1+M1. If this had been allowed to proceed, we'd end up writing corrupted data in Tier 2.
- The solution is to use the last processed offset in the StorageWriter as the `UpdateAttributesOperation` offset when needing to begin an `AggregatedOperation` with it.

**How to verify it**  
New unit test added to verify this scenario.

### DIFF
--- a/segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/HDFSIntegrationTest.java
+++ b/segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/HDFSIntegrationTest.java
@@ -63,6 +63,11 @@ public class HDFSIntegrationTest extends BookKeeperIntegrationTestBase {
     //region StreamSegmentStoreTestBase Implementation
 
     @Override
+    protected boolean appendAfterMerging() {
+        return false; // HDFS is slow enough as it is; adding this would cause the test to take even longer.
+    }
+
+    @Override
     protected ServiceBuilder createBuilder(ServiceBuilderConfig.Builder configBuilder, int instanceId) {
         ServiceBuilderConfig builderConfig = getBuilderConfig(configBuilder, instanceId);
         return ServiceBuilder
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
@@ -117,7 +117,7 @@ class SegmentAggregator implements WriterSegmentProcessor, AutoCloseable {
         this.timer = Preconditions.checkNotNull(timer, "timer");
         this.executor = Preconditions.checkNotNull(executor, "executor");
         this.lastFlush = new AtomicReference<>(timer.getElapsed());
-        this.lastAddedOffset = new AtomicLong(-1); // Will be set properly in initialize().
+        this.lastAddedOffset = new AtomicLong(-1); // Will be set properly after we process a StorageOperation.
         this.mergeTransactionCount = new AtomicInteger();
         this.truncateCount = new AtomicInteger();
         this.hasSealPending = new AtomicBoolean();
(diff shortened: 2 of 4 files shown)


andreipaduroiu:
Yes that is correct. `MergeSegmentOperation` and `SegmentAppendOperation` are `StorageOperation`s; upon adding an operation of either type to a `SegmentAggregator` the `lastAddedOffset` will be updated to the last byte that they encompass. If no such operation was encountered prior to an `UpdateAttributesOperation`, then Tier 2 is fully caught up (to that operation) so it will be set (in a different part of the code) to `metadata.getStorageLength()` (which would be the correct value in that case).

PR #3732, opened 2019-05-02
Issue 3715: replaced the non thread safe map with SimpleMeterRegistry

### LINKED ISSUE #3715: Sporadic failure in EndToEndStatsTest.testStatsCount
**Problem description**
Observed Travis failure in `EndToEndStatsTest` integration test:
```
io.pravega.test.integration.endtoendtest.EndToEndStatsTest > testStatsCount FAILED
        at io.pravega.test.integration.endtoendtest.EndToEndStatsTest.testStatsCount(EndToEndStatsTest.java:117)
```

**Problem location**
`EndToEndStatsTest`

Fix the test.

kevinhan88:
**Change log description**  
Replaced the non-thread safe map inside the TestStatsRecorder with solid SimpleMeterRegistry to fix the sporadic multi-threading failure.

**Purpose of the change**  
Fixes #3715 

**What the code does**  
1. Inside `EndToEndStatsTest.TestStatsRecorder`, replaces the non thread safe map with Micrometer `SimpleMeterRegistry`.
2. Use Micrometer API to retrieve counters, instead of the unsafe map get operation.
3. Added `SEGMENT_WRITE_BYTES` tests, in addition to the existing `SEGMENT_WRITE_EVENTS` tests.
4. Added a set to hold strong references to metric objects, to further consolidate the code for the most intensive runtime environment.

**How to verify it**  
Integration test should pass, no matter how many times it repeats.
(The old implementation usually fails after a few hundreds repeats)

### DIFF
--- a/test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndStatsTest.java
+++ b/test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndStatsTest.java
@@ -39,8 +39,9 @@ import io.pravega.test.common.TestUtils;
 import io.pravega.test.common.TestingServerStarter;
 import io.pravega.test.integration.demo.ControllerWrapper;
 import java.time.Duration;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 import lombok.Cleanup;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
@@ -141,9 +142,9 @@ public class EndToEndStatsTest {
         @Getter
         SimpleMeterRegistry registry = new SimpleMeterRegistry();
 
-        // A set to keep strong references to metric objects, as caching is skipped in the test.
+        // A placeholder to keep strong references to metric objects, as caching is skipped in the test.
         // Note Micrometer registry holds weak references only, so there is chance metric objects without strong references might be garbage collected.
-        private Set<Meter> metrics = new HashSet<>();
+        private List<Meter> references = Collections.synchronizedList(new LinkedList<Meter>());
 
         @Override
         public void createSegment(String streamSegmentName, byte type, int targetRate, Duration elapsed) {
@@ -170,8 +171,8 @@ public class EndToEndStatsTest {
             if (!StreamSegmentNameUtils.isTransactionSegment(streamSegmentName)) {
                 Counter eventCounter = registry.counter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
                 Counter byteCounter = registry.counter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
-                metrics.add(eventCounter);
-                metrics.add(byteCounter);
+                references.add(eventCounter);
+                references.add(byteCounter);
                 eventCounter.increment(numOfEvents);
                 byteCounter.increment(dataLength);
             }
@@ -181,8 +182,8 @@ public class EndToEndStatsTest {
         public void merge(String streamSegmentName, long dataLength, int numOfEvents, long txnCreationTime) {
             Counter eventCounter = registry.counter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
             Counter byteCounter = registry.counter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
-            metrics.add(eventCounter);
-            metrics.add(byteCounter);
+            references.add(eventCounter);
+            references.add(byteCounter);
             eventCounter.increment(numOfEvents);
             byteCounter.increment(dataLength);
         }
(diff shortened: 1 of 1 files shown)


test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndStatsTest.java
```diff
@@ -109,32 +115,39 @@ public void testStatsCount() throws Exception {
         EventStreamWriter<String> test = clientFactory.createEventWriter("test", new JavaSerializer<>(),
                 EventWriterConfig.builder().transactionTimeoutTime(10000).build());
 
+        String[] tags = segmentTags(StreamSegmentNameUtils.getQualifiedStreamSegmentName("test", "test", 0L));
+
         for (int i = 0; i < 10; i++) {
             test.writeEvent("test").get();
         }
-        assertEventuallyEquals(10, () -> statsRecorder.getSegments().get(StreamSegmentNameUtils.getQualifiedStreamSegmentName("test", "test", 0L)).get(), 2000);
+        assertEventuallyEquals(10, () -> (int) (statsRecorder.getRegistry().counter(SEGMENT_WRITE_EVENTS, tags).count()), 2000);
```
RaulGracia: Why the timeouts (last argument) in all the invocations to `assertEventuallyEquals` are different? A 1 or 2 seconds timeout seems reasonable for all the cases, right?

kevinhan88: @RaulGracia I repeated the test 517 times and finally triggered the problem. The bottom line is that HashMap is not thread safe, it shouldn't be used in a multiple thread env - and indeed the problem happened.

Yes 2 second for regular write assertion, but 10 seconds for transaction write to play it safe (I didn't set these 2 values but I'm ok with the values set).
For the additional Write Byte assertion I added, however, I give it 100ms max waiting instead of 2 second, because this assertion is after Write Event assertion, which should have already waited sufficient time.

RaulGracia: `HashMap` is not thread-safe, you are right, in the same way that the recently added `HashSet` is neither thread-safe (and you are not using synchronization), right? In the changes of this PR, I see that now there is no "removal" of elements from the `metrics` data structure, which may alleviate the problem. But note that if you had left the "remove" operation (or other update concurrent operations on that data structure), we may probably observe problems too.

kevinhan88: The strong reference place holder (HashSet) has no consumer hence no concurrent concern of it, however it doesn't hurt to synchronize it; also I changed the HashSet to LinkedList to improve the performance and reduce the memory footprint a bit.
I also changed the name "metrics" to "references" to emphasis it is a place holder, instead of a metrics registry, so there is no need to manage its life-cycle. (The test case will create about 10+ metrics objects, and all the objects will be released when tear down the test case)

RaulGracia:
@kevinhan88 please, let's not change more the code and try to execute Travis and Jenkins build multiple times to verify that the original issue is not reproduced (the version before the last commit has been executed 5 times without errors).

kevinhan88:
@RaulGracia I don't think running multiple Travis/Jenkins is sufficient to trigger this multi-threading issues. I set it up in IDE and the problem was only triggered after hundreds times of test case execution.

kevinhan88:
I'd like to point out one more thing: in addition to fixing the multi-threading issue, I'm also trying to make this particular integration test to be as close to production as possible.
The original code uses a raw HashMap to store counters of metrics, and use map api (instead of metrics api) to access counters, which lost the point of testing, because it could only prove map api works, not metrics api.

RaulGracia:
@kevinhan88 maybe I did not get my point across properly: the original issue (test failure) has been manifesting every few Travis/Jenkins builds (it does not need hundreds of executions). My point is: let's execute multiple times the Travis builds and the Jenkins builds to verify that the original issue is not reproduced again with the changes of this PR. Does it sound reasonable?

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndStatsTest.java
```diff
@@ -109,32 +116,39 @@ public void testStatsCount() throws Exception {
...
 
     private static class TestStatsRecorder implements SegmentStatsRecorder {
+
         @Getter
-        HashMap<String, AtomicInteger> segments = new HashMap<>();
+        SimpleMeterRegistry registry = new SimpleMeterRegistry();
+
+        // A placeholder to keep strong references to metric objects, as caching is skipped in the test.
+        // Note Micrometer registry holds weak references only, so there is chance metric objects without strong references might be garbage collected.
+        private List<Meter> references = Collections.synchronizedList(new LinkedList<Meter>());
```
tkaitchuck: Please use ArrayList not linked list.

kevinhan88: fixed

RaulGracia:
@tkaitchuck do you have more comments for this PR?

PR #3740, opened 2019-05-03
Issue 3735: Update StreamCut Validation inside a ReaderGroup to handle completed Segments.

### LINKED ISSUE #3735: System test failure StreamCut Test
**Problem description**
`io.pravega.test.system.StreamCutsTest` test failed for branch ` issue-3719-cherry-pick-for-r0.5-0.5.0-2236.137147`  with the following error.

```
2019-04-30 22:16:06,178 74382 [main] ERROR i.p.t.s.f.TestFrameworkException - TestFramework Exception. Type: InternalError, Details: Exception while running test method: streamCutsTest
java.lang.IllegalArgumentException: Segment offset in startStreamCut should be <= segment offset in endStreamCut.
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:141)
	at io.pravega.client.stream.ReaderGroupConfig$ReaderGroupConfigBuilder.lambda$verifyStartAndEndStreamCuts$2(ReaderGroupConfig.java:223)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
```

This issue is observed even on the master build.

**Problem location**
ReaderGroup initialization with StreamCuts.

With PR https://github.com/pravega/pravega/pull/3364 StreamCut can have segments(completed) with an offset as `-1L`. This causes

* ReaderGroupConfig's streamCut validation to fail in the case of startStreamCut and endStreamCut both consisting of the same `Segment`. If the endStreamCut points to the end of the `Segment` with offset `-1L` then the `io.pravega.client.stream.ReaderGroupConfig.ReaderGroupConfigBuilder#verifyStartAndEndStreamCuts` will fail.

* Having -1L in StreamCut can cause issues while serializing it  as `io.pravega.client.stream.impl.StreamCutImpl.StreamCutSerializer` serializes the offset as `out.writeCompactLong()`. (-1L cannot be represented with compact long.

* The bounded Processing logic reader group needs to be tweaked to handle this scenario where segments with -1L are passed.

shrids:
**Change log description**  
- Ensure bounded stream processing validation handles StreamCuts with completed Segments.
-  StreamCutTest now uses `io.pravega.client.stream.ReaderGroup#generateStreamCuts` to generate StreamCuts.

**Purpose of the change**  
Fixes #3735 

**What the code does**  
With PR #3364 StreamCut can have completed Segments i.e. it can contain segments with an offset as -1L. 
This causes
*  ReaderGroupConfig's `StreamCut` validation to fail in the case of startStreamCut and endStreamCut both having of the same Segment, where the `endStreamCut` points to the end of the Segment with offset -1L. 

*    Having -1L in a `StreamCut` can cause issues while serializing it as `io.pravega.client.stream.impl.StreamCutImpl.StreamCutSerializer` serializes the offset as `out.writeCompactLong()` , -1L cannot be represented with compact long.

* The bounded stream processing logic needs to be tweaked to handle this scenario where segments with -1L are passed as the upper bound of the stream.

**How to verify it**  
Note: System test run is in progress.


client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -308,7 +308,8 @@ private boolean shouldAcquireSegment() throws ReaderNotInReaderGroupException {
             if (!state.isReaderOnline(readerId)) {
                 throw new ReaderNotInReaderGroupException(readerId);
             }
-            if (acquireTimer.hasRemaining()) {
+            // Acquire segments if no segments assigned to a reader and there are unassigned segments that can be acquired.
+            if ((!(state.getSegments(readerId).isEmpty() && state.getNumberOfUnassignedSegments() != 0)) && acquireTimer.hasRemaining()) {
```
tkaitchuck: We already cover the case of unassigned segments == 0 below. This can be simplified to
`!state.getSegments(readerId).isEmpty() && acquireTimer.hasRemaining()`
But I'm not sure this is what we actually want to do. For example notice that acquireTimer gets updated with `UPDATE_WINDOW` to perform the update below. This could result in multiple threads from the same reader all trying to acquire segments at the same time. 

What is the problem that prompted this change?

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
```diff
@@ -172,15 +172,15 @@ private void read00(RevisionDataInput revisionDataInput, StreamCutBuilder builde
             builder.stream(stream);
             Map<Segment, Long> map = revisionDataInput.readMap(in -> new Segment(stream.getScope(),
                                                                                  stream.getStreamName(), in.readCompactLong()),
-                                                               in -> in.readCompactLong());
+                                                               in -> in.readLong());
```
tkaitchuck: Is this a compatible change?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
shrids: Re-verified the need for this change using an integration test, it is not required.
(The current `UPDATE_WINDOW` has a duration of 30 seconds)

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -220,8 +220,15 @@ private void verifyStartAndEndStreamCuts(final StreamCut startStreamCut, final S
            final Map<Segment, Long> endPositions = endStreamCut.asImpl().getPositions();
            //check offsets for overlapping segments.
            startPositions.keySet().stream().filter(endPositions::containsKey)
-                        .forEach(s -> Preconditions.checkArgument(startPositions.get(s) <= endPositions.get(s),
-                                "Segment offset in startStreamCut should be <= segment offset in endStreamCut."));
+                        .forEach(s -> {
+                            if (startPositions.get(s) == -1) {
```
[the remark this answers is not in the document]
shrids: Thanks, I was able to simulate the issue you described. 
I have updated the code to improve segment completed handling. Now the client can differentiate between a seal and reaching upper bound.

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
shrids: Enabled serialization compatibility.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -220,8 +220,15 @@ private void verifyStartAndEndStreamCuts(final StreamCut startStreamCut, final S
...
+                                Preconditions.checkArgument(endPositions.get(s) == -1,
```
fpj: We use `-1` to say that we have reached the end of a segment, and we release completed segments at checkpoint. This case seems to be saying that it is forbidden to use two stream cuts for begin and end such that no checkpoint happened between the two. Is that the semantics we are trying to enforce, and if so, isn't that unnecessarily limiting?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -220,7 +220,9 @@ public void resetReaderGroup(ReaderGroupConfig config) {
 
         return listOfMaps.stream()
                          .flatMap(map -> map.entrySet().stream())
-                         .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+                         .collect(Collectors.toMap(Entry::getKey,
+                                                   // A value of -1L implies read until the end of the segment.
+                                                   entry -> (entry.getValue() == -1L) ? Long.MAX_VALUE : entry.getValue()));
```
fpj: I don't understand why we need this conversion from `-1` to `MAX_VALUE`.

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
```diff
@@ -184,6 +186,37 @@ private void write00(StreamCutInternal cut, RevisionDataOutput revisionDataOutpu
...
+            return 1;
+        }
+
+        @Override
+        protected void declareVersions() {
+            super.declareVersions();
+            version(1).revision(0, this::write10, this::read10);
+        }
+
+        private void read10(RevisionDataInput revisionDataInput, StreamCutBuilder builder) throws IOException {
```
fpj: We are adding a new version, which enables us to deserialize a data structure with an older format, but we are not doing anything with it. In particular, if we get a stream cut with `00` format, then we should be executing the previous logic or perhaps implementing a reasonable workaround.

Also, it might be a more elegant solution to extend `RevisionDataOutput` to include something like `writeCompactSignedLong` and `writeCompactSignedLong` that uses a bit for the signal rather than use stock `java.io`.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
shrids: `getEndSegmentsForStreams(ReaderGroupConfig config)` returns the list of EndSegments and its corresponding end offsets upto which the segment should read. This `endOffset` is passed as parameter for `SegmentInputStreamImpl` and `Long.MAX_VALUE` corresponds to `SegmentInputStreamImpl#UNBOUNDED_END_OFFSET`

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
```diff
@@ -184,6 +186,37 @@ private void write00(StreamCutInternal cut, RevisionDataOutput revisionDataOutpu
         }
     }
 
+    // StreamCut serializer for version 1.
+    public static class StreamCutSerializer01 extends StreamCutSerializer {
```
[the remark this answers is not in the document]
shrids: This ensures that the correct `Serializer` implementation is used based on the version i.e. if we have a `StreamCut` serialized with `version 0` then invoking the newer serializer `StreamCutSerializer01().deserialize(..)` would ensure that the serializer corresponding to `version 0` is invoked by `io.pravega.common.io.serialization.VersionedSerializer.SingleType#deserializeContents`

The junit `.StreamCutTest#testStreamCutSerializationCompatabilityV0` verifies this behavior.

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
shrids: Yes, the older logic is invoked based on the version number. the Junit `io.pravega.client.stream.StreamCutTest#testStreamCutSerializationCompatabilityV0` validates this behaviour.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
shrids: This check is verified only on the segments that are present in both start  and end `StreamCut`. 
If both start and end `StreamCut`s have the same segment then the following invariants are valid
- if segment's position in start `StreamCut`  is `-1L`  then segment's position in the end `StreamCut` should also be `-1L`
- segment's Positions in start `StreamCut` ( where position is not `-1L`) should be less or equal to the segment's Position in end `StreamCut`, provided the segment position in end `StreamCut` is not `-1L`

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
```diff
@@ -184,6 +186,37 @@ private void write00(StreamCutInternal cut, RevisionDataOutput revisionDataOutpu
         }
     }
 
+    // StreamCut serializer for version 1.
```
tkaitchuck: Unrelated to the discussion below, by convention this should be called StreamCutSerializer10 not StreamCutSerializer01

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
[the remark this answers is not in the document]
tkaitchuck: Check with @andreipaduroiu, but I don't think that is needed. It should be possible to do with a single class and just creating new methods.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: @fpj Perhaps an easier way to reason about why the above is correct is: "If you are starting at the end of a segment, and the end cut has the same segment, it cannot possibly be any offset other than the end of the segment"

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
shrids: fixed it.

RaulGracia:
One question @shrids: is this change going to impact the Flink/Hadoop connectors that already make use of StreamCuts? Do we need any kind of additional validation in this regard? cc/ @vijikarthi.

shrids:
@RaulGracia This change should not impact the Flink/Hadoop connectors w.r.t to usage of StreamCuts. Also, no new validation is required from the connector end in this regard. 
The only point to note is that StreamCut serialized in 0.5 version of Pravega is not compatible with 0.4 version of Pravega.

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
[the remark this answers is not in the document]
andreipaduroiu: You can do them in one class. The `getVersion` returns the **serialization version used for writing data** and not for reading data. You can declare multiple versions (and version-revisions) in a single serialization class.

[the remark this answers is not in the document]
andreipaduroiu: Please refer to the [Versioned Serializer Javadoc](https://github.com/pravega/pravega/blob/master/common/src/main/java/io/pravega/common/io/serialization/VersionedSerializer.java#L410) which has examples of how this should be done.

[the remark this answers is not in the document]
shrids: Thanks @andreipaduroiu @tkaitchuck @fpj . Refactored the code to use a single serialization class.

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
shrids: Updated the code to use the same VersionedSerializer, `io.pravega.client.stream.impl.StreamCutImpl.StreamCutSerializer` for both the versions.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
fpj: The overloading of `-1` is confusing me. Could we use a different value to represent `UNBOUNDED`? Could we use `Long.MAX_VALUE` directly rather than convert some other value like `-1` to `Long.MAX_VALUE`?

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
fpj: @shrids you have not responded to this:

> Also, it might be a more elegant solution to extend RevisionDataOutput to include something like writeCompactSignedLong and writeCompactSignedLong that uses a bit for the signal rather than use stock java.io.

In the case it is not clear what I'm suggesting, I'd rather see something like this in `write10`:

```
revisionDataOutput.writeMap(map, (out, s) -> out.writeCompactLong(s.getSegmentId()),
                                 (out, offset) -> out.writeSignedCompactLong(offset));
```

shrids: Pull request https://github.com/pravega/pravega/pull/3782/files is addressing this. Will update the PR once the changes are in.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
shrids: @fpj @tkaitchuck my reasons for not changing this is listed below. 
Please review it and let me know your opinion.

With PR https://github.com/pravega/pravega/pull/3364, `-1L` offset now represents a completely read segments and this can be part of a `StreamCut`. Now for bounded stream processing this StreamCut can used to represent the lower bound or the upper bound.

 In case of lower bound `-1L` implies that the we should start reading from the successors of the segments (with offset -1L) and in the case of upper bound the semantic changes to read until the end of the a segment.
Hence, changing -1L to Long.MAX_VALUE will transfer the confusion for lower bound.

Also, using `Long.MAX_VALUE` implies we will not be able to use `RevisionDataOutput#writeCompactLong` since it has a limit of `2^62-1`

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
shrids: fixed it.

Issue #3745, opened 2019-05-03
Move security constants to :shared:test


fpj:
**Problem description**
According to the conversation in this thread of #3693:

https://github.com/pravega/pravega/pull/3693#discussion_r279546589

we want to move security constants in: 

test/testcommon/src/main/java/io/pravega/test/common/SecurityConfigDefaults.java

to :shared:test, which is a package that doesn't exist currently. These defaults are specific to Pravega and as such shouldn't be in `:test:common`. `:test:common` is for utility classes that can be used outside the scope of Pravega, while `:shared` is where we have shared classes that are useful only in the scope of Pravega.

**Problem location**
Security defaults for testing purposes.

**Suggestions for an improvement**
Make the necessary gradle changes to accomodate `:shared:test` and move `SecurityConfigDefaults` there.

Issue #3296, opened 2019-01-18
Authentication failure messages in logs with auth-enabled Standalone mode


ravisharda:
**Problem description**
1. Enable ``auth`` in ``standalone-config.properties``:
```
singlenode.enableAuth=true
singlenode.userName=admin
singlenode.passwd=1111_aaaa
singlenode.passwdFile=../config/passwd
```
2. Start Pravega in standalone mode: ``./gradlew startStandalone``

3. Observe stack traces like the following showing up again and again in the logs: 
```
[INFO ] 2019-01-18 00:00:09.964 [core-17] i.p.s.s.h.ZKSegmentContainerMonitor - Container Changes: Desired = [0, 1, 2, 3], Current = [0, 1, 2, 3], PendingTasks = [], ToStart = [], ToStop = [].
[INFO ] 2019-01-18 00:00:10.274 [grpc-default-executor-12] i.p.c.s.r.g.v.ControllerServiceImpl - getCurrentSegments called for stream _system/_requeststream.
[ERROR] 2019-01-18 00:00:10.275 [grpc-default-executor-12] i.p.c.s.r.g.v.ControllerServiceImpl - Controller api failed with authenticator error
[WARN ] 2019-01-18 00:00:10.278 [grpc-default-executor-13] i.p.c.s.i.ControllerImpl - gRPC call for getCurrentSegments with trace id 0 failed with server error.
io.grpc.StatusRuntimeException: UNAUTHENTICATED: Authentication failed
	at io.grpc.Status.asRuntimeException(Status.java:526)
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:418)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
	at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:663)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
	at io.grpc.internal.CensusTracingModule$TracingClientInterceptor$1$1.onClose(CensusTracingModule.java:392)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:443)
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:63)
[...]

PR #3756, opened 2019-05-07
Issue 3728: Fix for batch client auth issues

### LINKED ISSUE #3728: Batch client auth issues
**Problem description**
Batch client API calls encounter failure when Auth (authentication and authorization) is enabled. 

**Problem location**
See comments by @vijikarthi in https://github.com/pravega/pravega/issues/3296#issuecomment-475811508 for clues.

ravisharda:
**Change log description**  
Fixes the issue that was causing auth failures when using batch clients with auth-enabled Pravega cluster. 

**Purpose of the change**  
Fixes #3728. 

**What the code does**  
Here are the major changes:
* Main code: 
   - Fixes the issue by setting delegation token in the `ControllerServiceImpl.getSegmentsBetween()` method after authorization. 
   - Improves error logging (albeit slightly) in one place. The existing logged message can be misleading at times.
* Test code:
   - Rearranges the content of `io.pravega.test.integration.BatchClientTest` to allow for different security configurations to be set up by subclasses. 
   - Introduces a new subclass `io.pravega.test.integration.BatchClientWithAuthTest` of the `BatchClientTest`. It configures an auth-enabled environment and reuses the existing tests present in `BatchClientTest` to verify that batch clients work properly with auth-enabled. 
   - Introduces a new test helper class `PasswordAuthHandlerInput` to make it easier to generate and inject input file for the default AuthHandler - PasswordAuthHandler. 

**How to verify it**  
* All unit tests should pass.
* These integration tests in the following classes should pass: 
   - `io.pravega.test.integration.BatchClientWithAuthTest`
   - `io.pravega.test.integration.BatchClientTest`

**Additional Details:**

The root cause turned out to be missing delegation token, as shown below. 

```
2019-05-05 20:04:01,923 92884 [epollEventLoopGroup-6-7] TRACE i.p.s.protocol.netty.CommandDecoder - Decode a message on connection: /127.0.0.1:49114. Message was WireCommands.GetStreamSegmentInfo(type=GET_STREAM_SEGMENT_INFO, requestId=274877906945, segmentName=basicTestScope/basicTestStream/0.#epoch.0, delegationToken=)
2019-05-05 20:04:01,923 92884 [epollEventLoopGroup-6-7] DEBUG i.p.s.s.h.h.ServerConnectionInboundHandler - Processing request: WireCommands.GetStreamSegmentInfo(type=GET_STREAM_SEGMENT_INFO, requestId=274877906945, segmentName=basicTestScope/basicTestStream/0.#epoch.0, delegationToken=)
2019-05-05 20:04:01,923 92884 [epollEventLoopGroup-6-7] ERROR i.p.s.s.h.h.ServerConnectionInboundHandler - Caught exception on connection: 
java.lang.IllegalArgumentException: JWT String argument cannot be null or empty.
	at io.jsonwebtoken.lang.Assert.hasText(Assert.java:135)
	at io.jsonwebtoken.impl.DefaultJwtParser.parse(DefaultJwtParser.java:479)
	at io.jsonwebtoken.impl.DefaultJwtParser.parseClaimsJws(DefaultJwtParser.java:541)
	at io.pravega.segmentstore.server.host.delegationtoken.TokenVerifierImpl.verifyToken(TokenVerifierImpl.java:36)
	at io.pravega.segmentstore.server.host.handler.PravegaRequestProcessor.verifyToken(PravegaRequestProcessor.java:207)
	at io.pravega.segmentstore.server.host.handler.PravegaRequestProcessor.getStreamSegmentInfo(PravegaRequestProcessor.java:400)
	at io.pravega.shared.protocol.netty.DelegatingRequestProcessor.getStreamSegmentInfo(DelegatingRequestProcessor.java:60)
	at io.pravega.shared.protocol.netty.WireCommands$GetStreamSegmentInfo.process(WireCommands.java:966)
at 
```

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -13,6 +13,7 @@ import com.google.common.base.Throwables;
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import io.pravega.auth.AuthHandler;
+import io.pravega.auth.AuthorizationException;
 import io.pravega.client.stream.impl.ModelHelper;
 import io.pravega.common.Exceptions;
 import io.pravega.common.hash.RandomFactory;
@@ -223,7 +224,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
         log.info("getSegmentsBetweenStreamCuts called for stream {} for cuts from {} to {}", request.getStreamInfo(), request.getFromMap(), request.getToMap());
         String scope = request.getStreamInfo().getScope();
         String stream = request.getStreamInfo().getStream();
-        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorizationAndCreateToken(
                 AuthResourceRepresentation.ofStreamInScope(scope, stream), AuthHandler.Permissions.READ),
                 delegationToken -> controllerService.getSegmentsBetweenStreamCuts(request)
                         .thenApply(segments -> ModelHelper.createStreamCutRangeResponse(scope, stream,
@@ -456,12 +457,20 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                             streamObserver.onCompleted();
                         }
                     });
-        } catch (Exception e) {
-            log.error("Controller api failed with authenticator error");
+        } catch (RuntimeException e) {
+            log.warn(e.getMessage(), e);
             logAndUntrackRequestTag(requestTag);
-            streamObserver.onError(Status.UNAUTHENTICATED
-                    .withDescription("Authentication failed")
-                    .asRuntimeException());
+
+            Throwable cause = e.getCause();
+            if (cause instanceof AuthorizationException) {
+                streamObserver.onError(Status.UNAUTHENTICATED
+                        .withDescription("Authentication failed")
+                        .asRuntimeException());
+            } else {
+                streamObserver.onError(Status.INTERNAL
+                        .withDescription(cause.getMessage())
+                        .asRuntimeException());
+            }
         }
     }
 
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -204,7 +204,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
     }
 
     @Override
-    public void getSegmentsImmediatlyFollowing(SegmentId segmentId, StreamObserver<SuccessorResponse> responseObserver) {
+    public void getSegmentsImmediatelyFollowing(SegmentId segmentId, StreamObserver<SuccessorResponse> responseObserver) {
         log.info("getSegmentsImmediatelyFollowing called for segment {} ", segmentId);
         authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
                 AuthResourceRepresentation.ofStreamInScope(segmentId.getStreamInfo().getScope(),
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -13,7 +13,6 @@ import com.google.common.base.Throwables;
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import io.pravega.auth.AuthHandler;
-import io.pravega.auth.AuthorizationException;
 import io.pravega.client.stream.impl.ModelHelper;
 import io.pravega.common.Exceptions;
 import io.pravega.common.hash.RandomFactory;
@@ -446,7 +445,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                         logAndUntrackRequestTag(requestTag);
                         if (ex != null) {
                             Throwable cause = Exceptions.unwrap(ex);
-                            log.error("Controller api failed with error: ", ex);
+                            log.error("Controller api failed with error. " + ex.getMessage(), ex);
                             String errorDescription = replyWithStackTraceOnError ? "controllerStackTrace=" + Throwables.getStackTraceAsString(ex) : cause.getMessage();
                             streamObserver.onError(Status.INTERNAL
                                     .withCause(cause)
@@ -457,20 +456,12 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                             streamObserver.onCompleted();
                         }
                     });
-        } catch (RuntimeException e) {
-            log.warn(e.getMessage(), e);
+        } catch (Exception e) {
+            log.error(e.getMessage(), e);
             logAndUntrackRequestTag(requestTag);
-
-            Throwable cause = e.getCause();
-            if (cause instanceof AuthorizationException) {
-                streamObserver.onError(Status.UNAUTHENTICATED
-                        .withDescription("Authentication failed")
-                        .asRuntimeException());
-            } else {
-                streamObserver.onError(Status.INTERNAL
-                        .withDescription(cause.getMessage())
-                        .asRuntimeException());
-            }
+            streamObserver.onError(Status.UNAUTHENTICATED
+                    .withDescription("Authentication failed")
+                    .asRuntimeException());
         }
     }
 
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -445,7 +445,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                         logAndUntrackRequestTag(requestTag);
                         if (ex != null) {
                             Throwable cause = Exceptions.unwrap(ex);
-                            log.error("Controller api failed with error. " + ex.getMessage(), ex);
+                            log.error("Controller api failed with error: {}", ex.getMessage(), ex);
                             String errorDescription = replyWithStackTraceOnError ? "controllerStackTrace=" + Throwables.getStackTraceAsString(ex) : cause.getMessage();
                             streamObserver.onError(Status.INTERNAL
                                     .withCause(cause)
--- a/test/integration/src/main/java/io/pravega/test/integration/utils/PasswordAuthHandlerInput.java
+++ b/test/integration/src/main/java/io/pravega/test/integration/utils/PasswordAuthHandlerInput.java
@@ -7,7 +7,7 @@
  *
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
-package io.pravega.test.common;
+package io.pravega.test.integration.utils;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
--- a/test/integration/src/main/java/io/pravega/test/integration/utils/PasswordAuthHandlerInput.java
+++ b/test/integration/src/main/java/io/pravega/test/integration/utils/PasswordAuthHandlerInput.java
@@ -11,6 +11,7 @@ package io.pravega.test.integration.utils;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import io.pravega.common.Exceptions;
 import lombok.Data;
 import lombok.Getter;
 
@@ -34,6 +35,8 @@ public class PasswordAuthHandlerInput {
     }
 
     public PasswordAuthHandlerInput(String fileName, String extension) {
+        Exceptions.checkNotNullOrEmpty(fileName, "fileName");
+        Exceptions.checkNotNullOrEmpty(extension, "extension");
         try {
             file = File.createTempFile(fileName, extension);
         } catch (IOException e) {
@@ -42,10 +45,12 @@ public class PasswordAuthHandlerInput {
     }
 
     public void postEntry(Entry entry) {
+        Preconditions.checkNotNull(entry, "Specified entry is null.");
         postEntries(Arrays.asList(entry));
     }
 
     public void postEntries(List<Entry> entries) {
+        Exceptions.checkNotNullOrEmpty(entries, "entries");
         try (FileWriter writer = new FileWriter(file.getAbsolutePath())) {
             entries.forEach(e -> {
                 try {
--- a/test/integration/src/test/java/io/pravega/test/integration/BatchClientTest.java
+++ b/test/integration/src/test/java/io/pravega/test/integration/BatchClientTest.java
@@ -80,9 +80,10 @@ public class BatchClientTest {
     protected final int servicePort = TestUtils.getAvailableListenPort();
     protected final int containerCount = 4;
 
+    protected TestingServer zkTestServer;
+
     private final Random random = RandomFactory.create();
 
-    protected TestingServer zkTestServer;
     private PravegaConnectionListener server;
     private ControllerWrapper controllerWrapper;
     private ServiceBuilder serviceBuilder;
--- a/test/integration/src/test/java/io/pravega/test/integration/BatchClientTest.java
+++ b/test/integration/src/test/java/io/pravega/test/integration/BatchClientTest.java
@@ -18,6 +18,7 @@ import io.pravega.client.admin.StreamInfo;
 import io.pravega.client.admin.StreamManager;
 import io.pravega.client.batch.SegmentIterator;
 import io.pravega.client.batch.SegmentRange;
+import io.pravega.client.batch.impl.BatchClientFactoryImpl;
 import io.pravega.client.batch.impl.SegmentRangeImpl;
 import io.pravega.client.segment.impl.Segment;
 import io.pravega.client.stream.EventStreamWriter;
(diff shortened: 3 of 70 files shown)


test/testcommon/src/main/java/io/pravega/test/common/PasswordAuthHandlerInput.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * This is a helper class for tests and may be used for generating the input file for the PasswordAuthHandler - the
+ * default AuthHandler implementation.
+ */
+public class PasswordAuthHandlerInput {
```
fpj: This class is currently used only in integration tests, and as such, there is no need to share. It also has the issue we discussed in the other PR that `test:common` is not the place for Pravega-specific classes.

I'd rather have it in `test:integration` for the time being.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -223,7 +223,7 @@ public void getSegmentsBetween(Controller.StreamCutRange request, StreamObserver
         log.info("getSegmentsBetweenStreamCuts called for stream {} for cuts from {} to {}", request.getStreamInfo(), request.getFromMap(), request.getToMap());
         String scope = request.getStreamInfo().getScope();
         String stream = request.getStreamInfo().getStream();
-        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorizationAndCreateToken(
```
fpj: Some methods in this very same class invoke `checkAuthorization` (e.g., the method above with the typo in the name) while others invoke `check AuthorizationAndCreateToken`. I'm not sure what the criteria are for deciding what to call.

test/integration/src/test/java/io/pravega/test/integration/BatchClientWithAuthTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    protected ControllerWrapper createControllerWrapper() {
+        return new ControllerWrapper(zkTestServer.getConnectString(),
+                false, true,
+                controllerPort, serviceHost, servicePort, containerCount, -1,
+                true, PASSWORD_AUTHHANDLER_INPUT.getPath(), "secret");
+    }
+
+    @Test
+    @Override
```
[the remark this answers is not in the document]
ravisharda: Yes, creating these methods isn't necessary for automated runs. I added them here to have the flexibility of running them individually while debugging.

test/testcommon/src/main/java/io/pravega/test/common/PasswordAuthHandlerInput.java
ravisharda: I kept it here as I am aware of existing test classes other than integration tests that can benefit from it (`ControllerGrpcAuthFocusedTest` is a case in point). I have not updated those existing test classes right now, though. I was thinking we'll move it to the shared project later when we issue #3745. 
What do you say?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: The former checks for authorization only, while the latter sets a delegation token apart from checking for authorization. The delegation token is used subsequently by the client when talking to the segment store.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -445,7 +445,7 @@ public void getDelegationToken(StreamInfo request, StreamObserver<DelegationToke
                         logAndUntrackRequestTag(requestTag);
                         if (ex != null) {
                             Throwable cause = Exceptions.unwrap(ex);
-                            log.error("Controller api failed with error: ", ex);
+                            log.error("Controller api failed with error. " + ex.getMessage(), ex);
```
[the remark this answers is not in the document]
ravisharda: I've fixed it in the latest commit.

test/integration/src/test/java/io/pravega/test/integration/BatchClientWithAuthTest.java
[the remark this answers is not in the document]
fpj: Are you planning on keeping or removing them? I'd say we remove them.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
fpj: I get that basic behavior @ravisharda , that's not the point. Why do we need to create a token for some calls and others we don't ? What's different in the call path that makes one require the creation of a token while the other didn't?

Clearly this distinction is what caused the bug because whoever wrote this code did not make the right assessment on whether a token needed to be created or not. I'm asking for some explicit criteria so that we can perhaps review the calls in this very same class and others and validate that they are doing the right thing.

test/testcommon/src/main/java/io/pravega/test/common/PasswordAuthHandlerInput.java
fpj: Let's not abuse the exception. There is no need for this class to be here, and if there is a need to share it, which is not the case in this PR, let's do the work of enabling `shared:test`.

test/integration/src/test/java/io/pravega/test/integration/BatchClientWithAuthTest.java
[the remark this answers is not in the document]
ravisharda: I've removed the overriding test methods in the latest commit.

test/testcommon/src/main/java/io/pravega/test/common/PasswordAuthHandlerInput.java
ravisharda: :) Sure. As suggested, I've moved this class to the project `test:integration` under the package `io.pravega.test.integration.utils`. Hope that works.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: I completely agree with you. 

As of now, I don't have a complete rule-set that governs whether to set the delegation token or not. I'd need to discover those rules through experimentation and document them. I'd propose that we write up a new Wiki document that details out the auth design, beyond the very high-level design information covered under [this](https://github.com/pravega/pravega/wiki/PDP-23:-Pravega-security----encryption-and-Role-Based-Access-Control) PDP. In that new document, we can cover both the high-level design as well as important rules like these. 

We can cover this separately as part of a different issue if you agree.

vijikarthi: I am not sure if the fix provided here addresses all the API calls of batch client in its entirety (for e.g., the deprecated APIs of `BatchClientFactoryImpl` class)?

test/integration/src/main/java/io/pravega/test/integration/utils/PasswordAuthHandlerInput.java
```diff
@@ -0,0 +1,77 @@
...
+    public PasswordAuthHandlerInput(String fileName, String extension) {
+        try {
+            file = File.createTempFile(fileName, extension);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void postEntry(Entry entry) {
+        postEntries(Arrays.asList(entry));
```
vijikarthi: It is good to have a precondition to validate the `entry` for not being null

test/integration/src/test/java/io/pravega/test/integration/BatchClientTest.java
```diff
@@ -119,13 +121,42 @@ public void tearDown() throws Exception {
...
 
+    //region Factory methods that may be overridden by subclasses.
+
+    protected ServiceBuilder createServiceBuilder() {
+        return ServiceBuilder.newInMemoryBuilder(ServiceBuilderConfig.getDefaultConfig());
+    }
+
+    protected ControllerWrapper createControllerWrapper() {
+        return new ControllerWrapper(zkTestServer.getConnectString(),
+                false, true,
```
vijikarthi: A variable name explaining what those boolean values represents can be useful instead of passing `true` or `false`

test/integration/src/test/java/io/pravega/test/integration/BatchClientWithAuthTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * The tests in this class are intended to verify whether Batch Client works with a Pravega cluster
+ * that has "Auth" (short for authentication and authorization) enabled.
+ *
+ * This class inherits the tests of the parent class. Some of the test methods of the parent are reproduced here as
+ * handles, to enable running an individual test interactively (for debugging purposes).
+ */
+public class BatchClientWithAuthTest extends BatchClientTest {
+
```
vijikarthi: Do we also have a test that includes both `Auth` and `TLS`?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: @vijikarthi, there seems to be no deprecated API method in `BatchFactoryImpl` interface, There is indeed a deprecated API method in its implementation class `BatchClientFactoryImpl`. I'll verify it to check if works with auth enabled. Could you also specify if there are other batch client API methods that you suspect won't work with auth?

test/integration/src/test/java/io/pravega/test/integration/BatchClientWithAuthTest.java
ravisharda: We don't have a batch client test class that includes both `auth` and `TLS` or with only `TLS`. I just added one for `auth` in this PR. I didn't add test classes for TLS, as that change, although desirable, wouldn't be relevant for the current issue and PR. Or maybe it is: how do you know this will work in tandem with `TLS` otherwise. Just thinking out loud. 

What do you say?

test/integration/src/test/java/io/pravega/test/integration/BatchClientTest.java
ravisharda: Since it is a simple argument (and not a complex condition/expression), IDE parameter name hints should be enough, assuming the parameter names are aptly-named - which, luckily is the case here as shown in the screenshot below.  

image

test/integration/src/main/java/io/pravega/test/integration/utils/PasswordAuthHandlerInput.java
ravisharda: Sure. I've pushed a new commit that checks arguments for this and other public methods of the class.

ravisharda:
W.r.t. the comment below, please note that this PR is intended to fix only the auth issues specific to batch client API and not the condition that results in the exception below, which in turn shows up whenever auth is enabled, whether you use batch/regular client or even if you don't use any client. 

This exception showed up in my verification of this fix as well but didn't cause the batch client auth tests in class ` `io.pravega.test.integration.BatchClientWithAuthTest` to fail. Also, this exception shows up when running the standalone mode (without the batch client), as you pointed out: the regular client API continue to function properly despite this error. Not to say that the exception is not a problem (it certainly is), but rather that it is orthogonal to this issue and should be addressed separately. Issue #3296, which you have reopened, needs to be addressed for that issue. 

As for the Hadoop connector tests failing with the code in this PR, there might be a different reason for it. I'll need to investigate further.

> I took this branch and tested with Hadoop Connector to see if it resolves the issues mentioned in [#3296 (comment)](https://github.com/pravega/pravega/issues/3296#issuecomment-475811508) but I was still not able to run those secure test case. After that, I tried to run the standalone from this branch (`./gradlew startStandalone`) and I am seeing lot of these warnings.
> 
> ```
> 2019-05-07 16:31:39,766 11294 [grpc-default-executor-1] WARN  i.p.c.stream.impl.ControllerImpl - gRPC call for getCurrentSegments with trace id 0 failed with server error.
> io.grpc.StatusRuntimeException: UNAUTHENTICATED: Authentication failed
>         at io.grpc.Status.asRuntimeException(Status.java:530)
>         at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:434)
>         at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
>         at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
>         at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
>         at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
>         at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
>         at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
>         at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:694)
>         at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
>         at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
>         at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
>         at io.grpc.internal.CensusTracingModule$TracingClientInterceptor$1$1.onClose(CensusTracingModule.java:397)
>         at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:459)
>         at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:63)
>         at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:546)
>         at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.access$600(ClientCallImpl.java:467)
>         at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:584)
>         at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
>         at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
>         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
>         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
>         at java.lang.Thread.run(Thread.java:748)
> ```

ravisharda:
@vijikarthi After digging into the issue, I've found that the root cause for auth not working with Hadoop Connectors against a server built using the code from this PR is really a bug in Hadoop Connectors class `io.pravega.connectors.hadoop.SecurityHelper.prepareClientConfig()` method. 

Here's what I did to encounter the issue:
1. Configure Hadoop Connector to use auth, by setting the corresponding flag in `SetupUtils` class. 
2. Configure Pravega standalone mode to use auth. Start the server.
3. Run `io.pravega.connectors.hadoop.PravegaConnectorLocalJobITCase.testPravegaConnectorInput()` test. Target the test to run against the standalone mode server.
4. Observer the `auth` errors, especially the ones pertaining to scope/stream that the test creates. 

In the about scenario, the problem appears to be that when `PravegaInputFormat.getSplits()` invokes `io.pravega.connectors.hadoop.prepareClientConfig()`, the called method does not set the credentials properly and as a result, the Batch client API calls fail. If I modify the code inside the method to set the credentials properly, the test passes consistently. 

I believe we need to modify the respective code in Hadoop Connectors project to fix the auth issue encountered there, and that issue is orthogonal to the issue fixed by this PR. In this PR we are only addressing the auth issues specific to batch client API. 

Please take a look and let me know what you think.

ravisharda:
The reason for Hadoop Connectors tests failing when auth (short for authentication and authorization) was enabled, turned out to be an invalid specification for auth method. The test succeeded against a server built using this PR branch, after the system property `pravega.client.auth.method` was changed from to `Basic` (from `Default`). 

I just pushed a new integration test that demonstrates how to configure client credentials via system properties (as opposed to supplying a DefaultCredentials Object to ClientConfig directly).  

https://github.com/pravega/pravega/blob/17e5dee8733b2e342be0cb8a15f2cf98d14a8ebe/test/integration/src/test/java/io/pravega/test/integration/BatchClientAuthTest.java#L82

vijikarthi:
@ravisharda I took your branch and tested with Hadoop Connector by passing the security configuration as system properties. The authentication part worked but it failed on authorization. If I bypass the authorization verification it works properly with your branch. While debugging, I have noticed that the `BatchClientAuthTest` is not using the `SetupUtils` to load the in-memory Pravega cluster instead it is directly making use of `ControllerWrapper` class and it uses the password auth file which was created by the `BatchClientAuthTest::createAuthFile() which basically puts the user credential along with the ACL`. Is there any reason why we can't use `SetupUtils`?

ravisharda:
> Just had one question specific to the timeouts, otherwise the changes look good and I have tested it with Hadoop connector integration test.

Since the client retries for ten times by default regardless of the error, it can take some time before the test encounters an error. [Issue 3653](https://github.com/pravega/pravega/issues/3653) covers this among other issues.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -457,7 +457,7 @@ public void getDelegationToken(StreamInfo request, StreamObserver<DelegationToke
                         }
                     });
         } catch (Exception e) {
-            log.error("Controller api failed with authenticator error", e);
+            log.error(e.getMessage(), e);
```
shiveshr: we dont need to print exception.message and full error stack both.

PR #3761, opened 2019-05-08
Issue 3670: Service side Grpc implementation for watermarking related apis

### LINKED ISSUE #3670: Watermark: Add Controller gRPC apis for writers to report time
**Problem description**
For watermarks, writers need to report time to controller via `mark` and `commitTxn` apis. 
For this we need to add a new mark api and add writer id and time to commitTxn api. 

The apis implementation should record time for writerId in the metadata. 
Also, committransaction workflow should make use of commit time to record time for writerId in metadata. 

**Problem location**
Controller grpc

Add new apis

shiveshr:
**Change log description**  
Controller grpc api implementation and metadata support for watermarking. 

**Purpose of the change**  
Fixes  #3670 

**What the code does** 
This PR adds an implementation for grpc apis `noteTimeForWriter` and `removeWriter`.  Also, it updates `commitTxn` api to take writerId, time and position. 

We have added new metadata for the writer mark. For PravegaTables, the writer mark for a stream go into a new table called `_system/_tables/stream/WritersMarkTable.#<id>`.

The key is `writerId` and the value is the mark for the writer with `writerId`. The writer mark is a new metadata record type that is introduced in this change set. It has the timestamp and the position of the writer. 

The writer mark can be updated via explicit request from the writer or during a transaction commit. 
For transaction commits, whenever we merge a transaction segment into its parent segment, we record the segment offset after the merge completes. The segment offset is stored in the metadata record of the active transaction. 

When all transactions in a batch are merged and before marking transactions as committed, we report marks for each transaction where its record contains commit offsets, then it invokes `store.noteTimeForWriter(writerId, time, commitOffsets)`. 

It is important to note that it is possible that a writer calls `removeWriter`, and that there are attempts to `noteTimeForWriter` after the writer is removed. This situation can arise in at least two ways:

1-  Writer calls `noteTimeForWriter` explicitly, in which case we will create a new record.
2- Transaction commits after remove-writer call completes.  

We do not handle this in any special way and let the marks be reported for writers. 

The responsibility for removing stale writers from known writers set will lie with watermarking periodic workflow implementation and will be tackled in a subsequent PR. 

**How to verify it**  
New unit tests added to test new metadata records.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/ControllerService.java
+++ b/controller/src/main/java/io/pravega/controller/server/ControllerService.java
@@ -510,8 +510,8 @@ public class ControllerService {
         }
     }
 
-    public CompletableFuture<Controller.TimestampResponse> noteTimestampFromWriter(String scope, String stream, String writer, long timestamp, Map<Long, Long> streamCut) {
-        return streamStore.noteWriterMark(scope, stream, writer, timestamp, streamCut, null, executor)
+    public CompletableFuture<Controller.TimestampResponse> noteTimestampFromWriter(String scope, String stream, String writerId, long timestamp, Map<Long, Long> streamCut) {
+        return streamStore.noteWriterMark(scope, stream, writerId, timestamp, streamCut, null, executor)
                 .handle((r, e) -> {
                     Controller.TimestampResponse.Builder response = Controller.TimestampResponse.newBuilder();
                     if (e != null) {
--- a/controller/src/main/java/io/pravega/controller/server/ControllerService.java
+++ b/controller/src/main/java/io/pravega/controller/server/ControllerService.java
@@ -21,6 +21,7 @@ import io.pravega.controller.metrics.StreamMetrics;
 import io.pravega.controller.metrics.TransactionMetrics;
 import io.pravega.controller.store.stream.OperationContext;
 import io.pravega.controller.store.stream.ScaleMetadata;
+import io.pravega.controller.store.stream.StoreException;
 import io.pravega.controller.store.stream.StreamMetadataStore;
 import io.pravega.controller.store.stream.VersionedTransactionData;
 import io.pravega.controller.store.stream.records.StreamSegmentRecord;
@@ -331,14 +332,14 @@ public class ControllerService {
         return listOfSegment;
     }
 
-    public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String stream, final TxnId
-            txnId) {
+    public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String stream, final TxnId txnId, 
+                                                          final String writerId, final long timestamp) {
         Exceptions.checkNotNullOrEmpty(scope, "scope");
         Exceptions.checkNotNullOrEmpty(stream, "stream");
         Preconditions.checkNotNull(txnId, "txnId");
         Timer timer = new Timer();
         UUID txId = ModelHelper.encode(txnId);
-        return streamTransactionMetadataTasks.commitTxn(scope, stream, txId, null)
+        return streamTransactionMetadataTasks.commitTxn(scope, stream, txId, writerId, timestamp, null)
                 .handle((ok, ex) -> {
                     if (ex != null) {
                         log.warn("Transaction commit failed", ex);
@@ -507,6 +508,49 @@ public class ControllerService {
         } else if (status.equals(DeleteStreamStatus.Status.FAILURE)) {
             streamMetrics.deleteStreamFailed(scope, streamName);
         }
+    }
+
+    public CompletableFuture<Controller.TimestampResponse> noteTimestampFromWriter(String scope, String stream, String writer, long timestamp, Map<Long, Long> streamCut) {
+        return streamStore.noteWriterMark(scope, stream, writer, timestamp, streamCut, null, executor)
+                .handle((r, e) -> {
+                    Controller.TimestampResponse.Builder response = Controller.TimestampResponse.newBuilder();
+                    if (e != null) {
+                        response.setResult(Controller.TimestampResponse.Status.INTERNAL_ERROR);
+                    } else {
+                        switch (r) {
+                            case SUCCESS:
+                                response.setResult(Controller.TimestampResponse.Status.SUCCESS);
+                                break;
+                            case INVALID_TIME:
+                                response.setResult(Controller.TimestampResponse.Status.INVALID_TIME);
+                                break;
+                            case INVALID_POSITION:
+                                response.setResult(Controller.TimestampResponse.Status.INVALID_POSITION);
+                                break;
+                            default:
+                                response.setResult(Controller.TimestampResponse.Status.INTERNAL_ERROR);
+                                break;
+                        }
+                    }
+                    return response.build();
+                });
+    }
+ 
+    public CompletableFuture<Controller.WriterShutdownResponse> shutdownWriter(String scope, String stream, String writer) {
+        return streamStore.removeWriter(scope, stream, writer, null, executor)
+                .handle((r, e) -> {
+                    Controller.WriterShutdownResponse.Builder response = Controller.WriterShutdownResponse.newBuilder();
+                    if (e != null) {
+                        if (Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException) {
+                            response.setResult(Controller.WriterShutdownResponse.Status.STREAM_DOES_NOT_EXIST);
+                        } else {
+                            response.setResult(Controller.WriterShutdownResponse.Status.INTERNAL_ERROR);
+                        }
+                    } else {
+                        response.setResult(Controller.WriterShutdownResponse.Status.SUCCESS);
+                    }
+                    return response.build();
+                });
     }
 
     // End metrics reporting region
--- a/controller/src/main/java/io/pravega/controller/server/ControllerService.java
+++ b/controller/src/main/java/io/pravega/controller/server/ControllerService.java
@@ -537,12 +537,12 @@ public class ControllerService {
     }
  
     public CompletableFuture<Controller.WriterShutdownResponse> shutdownWriter(String scope, String stream, String writer) {
-        return streamStore.removeWriter(scope, stream, writer, null, executor)
+        return streamStore.shutdownWriter(scope, stream, writer, null, executor)
                 .handle((r, e) -> {
                     Controller.WriterShutdownResponse.Builder response = Controller.WriterShutdownResponse.newBuilder();
                     if (e != null) {
                         if (Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException) {
-                            response.setResult(Controller.WriterShutdownResponse.Status.STREAM_DOES_NOT_EXIST);
+                            response.setResult(Controller.WriterShutdownResponse.Status.UNKNOWN_WRITER);
                         } else {
                             response.setResult(Controller.WriterShutdownResponse.Status.INTERNAL_ERROR);
                         }
--- a/controller/src/main/java/io/pravega/controller/server/ControllerService.java
+++ b/controller/src/main/java/io/pravega/controller/server/ControllerService.java
@@ -536,18 +536,18 @@ public class ControllerService {
                 });
     }
  
-    public CompletableFuture<Controller.WriterShutdownResponse> shutdownWriter(String scope, String stream, String writer) {
+    public CompletableFuture<Controller.RemoveWriterResponse> removeWriter(String scope, String stream, String writer) {
         return streamStore.shutdownWriter(scope, stream, writer, null, executor)
                 .handle((r, e) -> {
-                    Controller.WriterShutdownResponse.Builder response = Controller.WriterShutdownResponse.newBuilder();
+                    Controller.RemoveWriterResponse.Builder response = Controller.RemoveWriterResponse.newBuilder();
                     if (e != null) {
                         if (Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException) {
-                            response.setResult(Controller.WriterShutdownResponse.Status.UNKNOWN_WRITER);
+                            response.setResult(Controller.RemoveWriterResponse.Status.UNKNOWN_WRITER);
                         } else {
-                            response.setResult(Controller.WriterShutdownResponse.Status.INTERNAL_ERROR);
+                            response.setResult(Controller.RemoveWriterResponse.Status.INTERNAL_ERROR);
                         }
                     } else {
-                        response.setResult(Controller.WriterShutdownResponse.Status.SUCCESS);
+                        response.setResult(Controller.RemoveWriterResponse.Status.SUCCESS);
                     }
                     return response.build();
                 });
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -448,14 +448,14 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
     }
 
     @Override
-    public void writerShutdown(Controller.WriterShutdownRequest request, StreamObserver<Controller.WriterShutdownResponse> responseObserver) {
+    public void removeWriter(Controller.RemoveWriterRequest request, StreamObserver<Controller.RemoveWriterResponse> responseObserver) {
         StreamInfo streamInfo = request.getStream();
         log.info("writerShutdown called for stream {}/{}, writer={}", streamInfo.getScope(),
                 streamInfo.getStream(), request.getWriter());
         authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
                 AuthResourceRepresentation.ofStreamInScope(streamInfo.getScope(), streamInfo.getStream()),
                 AuthHandler.Permissions.READ_UPDATE),
-                delegationToken  -> controllerService.shutdownWriter(streamInfo.getScope(),
+                delegationToken  -> controllerService.removeWriter(streamInfo.getScope(),
                         streamInfo.getStream(), request.getWriter()),
                 responseObserver);
     }
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -333,7 +333,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                 AuthHandler.Permissions.READ_UPDATE),
                 delegationToken -> controllerService.commitTransaction(request.getStreamInfo().getScope(),
                         request.getStreamInfo().getStream(),
-                        request.getTxnId()),
+                        request.getTxnId(), request.getWriterId(), request.getTimestamp()),
                 responseObserver);
     }
 
@@ -432,6 +432,37 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                 responseObserver);
     }
 
+    // region watermarking apis
+
+    @Override
+    public void writerShutdown(Controller.WriterShutdownRequest request, StreamObserver<Controller.WriterShutdownResponse> responseObserver) {
+        StreamInfo streamInfo = request.getStream();
+        log.info("writerShutdown called for stream {}/{}, writer={}", streamInfo.getScope(),
+                streamInfo.getStream(), request.getWriter());
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
+                AuthResourceRepresentation.ofStreamInScope(streamInfo.getScope(), streamInfo.getStream()),
+                AuthHandler.Permissions.READ_UPDATE),
+                delegationToken  -> controllerService.shutdownWriter(streamInfo.getScope(),
+                        streamInfo.getStream(), request.getWriter()),
+                responseObserver);
+    }
+
+    @Override
+    public void noteTimestampFromWriter(Controller.TimestampFromWriter request, StreamObserver<Controller.TimestampResponse> responseObserver) {
+        StreamInfo streamInfo = request.getPosition().getStreamInfo();
+        log.info("noteWriterMark called for stream {}/{}, writer={} time={}", streamInfo.getScope(),
+                streamInfo.getStream(), request.getWriter(), request.getTimestamp());
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
+                AuthResourceRepresentation.ofStreamInScope(streamInfo.getScope(), streamInfo.getStream()),
+                AuthHandler.Permissions.READ_UPDATE),
+                delegationToken  -> controllerService.noteTimestampFromWriter(streamInfo.getScope(),
+                        streamInfo.getStream(), request.getWriter(), request.getTimestamp(), request.getPosition().getCutMap()),
+                responseObserver);
+    }
+
+
+    // endregion
+    
     // Convert responses from CompletableFuture to gRPC's Observer pattern.
     private <T> void authenticateExecuteAndProcessResults(Supplier<String> authenticator, Function<String, CompletableFuture<T>> call,
                                                           final StreamObserver<T> streamObserver, RequestTag requestTag) {
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -433,20 +433,7 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
     }
 
     // region watermarking apis
-
-    @Override
-    public void writerShutdown(Controller.WriterShutdownRequest request, StreamObserver<Controller.WriterShutdownResponse> responseObserver) {
-        StreamInfo streamInfo = request.getStream();
-        log.info("writerShutdown called for stream {}/{}, writer={}", streamInfo.getScope(),
-                streamInfo.getStream(), request.getWriter());
-        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
-                AuthResourceRepresentation.ofStreamInScope(streamInfo.getScope(), streamInfo.getStream()),
-                AuthHandler.Permissions.READ_UPDATE),
-                delegationToken  -> controllerService.shutdownWriter(streamInfo.getScope(),
-                        streamInfo.getStream(), request.getWriter()),
-                responseObserver);
-    }
-
+    
     @Override
     public void noteTimestampFromWriter(Controller.TimestampFromWriter request, StreamObserver<Controller.TimestampResponse> responseObserver) {
         StreamInfo streamInfo = request.getPosition().getStreamInfo();
@@ -460,7 +447,18 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                 responseObserver);
     }
 
-
+    @Override
+    public void writerShutdown(Controller.WriterShutdownRequest request, StreamObserver<Controller.WriterShutdownResponse> responseObserver) {
+        StreamInfo streamInfo = request.getStream();
+        log.info("writerShutdown called for stream {}/{}, writer={}", streamInfo.getScope(),
+                streamInfo.getStream(), request.getWriter());
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
+                AuthResourceRepresentation.ofStreamInScope(streamInfo.getScope(), streamInfo.getStream()),
+                AuthHandler.Permissions.READ_UPDATE),
+                delegationToken  -> controllerService.shutdownWriter(streamInfo.getScope(),
+                        streamInfo.getStream(), request.getWriter()),
+                responseObserver);
+    }
     // endregion
     
     // Convert responses from CompletableFuture to gRPC's Observer pattern.
--- a/controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
@@ -981,7 +981,7 @@ public class InMemoryStream extends PersistentStreamBase {
             }
         }
         
-        return CompletableFuture.completedFuture(null);
+        return result;
     }
 
     @Override
@@ -990,6 +990,7 @@ public class InMemoryStream extends PersistentStreamBase {
         
         synchronized (writersLock) {
             writerMarks.remove(writer);
+            result.complete(null);
         }
         
         return result;
@@ -1031,7 +1032,7 @@ public class InMemoryStream extends PersistentStreamBase {
             } else if (!Objects.equals(existing.getVersion(), version)) {
                 result.completeExceptionally(StoreException.create(StoreException.Type.WRITE_CONFLICT, "writer mark version mismatch"));
             } else {
-                this.writerMarks.put(writer, existing);
+                this.writerMarks.put(writer, updatedCopy);
                 result.complete(null);
             }
         }
--- a/controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
@@ -13,6 +13,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.ImmutableMap;
 import io.pravega.client.stream.StreamConfiguration;
 import io.pravega.common.concurrent.Futures;
 import io.pravega.controller.store.stream.records.ActiveTxnRecord;
@@ -28,6 +29,7 @@ import io.pravega.controller.store.stream.records.StateRecord;
 import io.pravega.controller.store.stream.records.StreamConfigurationRecord;
 import io.pravega.controller.store.stream.records.StreamCutRecord;
 import io.pravega.controller.store.stream.records.StreamTruncationRecord;
+import io.pravega.controller.store.stream.records.WriterMark;
 import io.pravega.controller.util.Config;
 
 import javax.annotation.concurrent.GuardedBy;
@@ -98,6 +100,11 @@ public class InMemoryStream extends PersistentStreamBase {
      */
     @GuardedBy("txnsLock")
     private final Map<Integer, Set<UUID>> epochTxnMap = new HashMap<>();
+
+    private final Object writersLock = new Object();
+
+    @GuardedBy("writersLock")
+    private final Map<String, VersionedMetadata<WriterMark>> writerMarks = new HashMap<>();
 
     InMemoryStream(String scope, String name) {
         this(scope, name, Duration.ofHours(Config.COMPLETED_TRANSACTION_TTL_IN_HOURS).toMillis());
@@ -710,7 +717,7 @@ public class InMemoryStream extends PersistentStreamBase {
                     activeTxns.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, x -> x.getValue().getObject()))));
         }
     }
-
+    
     @Override
     CompletableFuture<List<Map.Entry<UUID, ActiveTxnRecord>>> getOrderedCommittingTxnInLowestEpoch() {
         List<Long> toPurge = new ArrayList<>();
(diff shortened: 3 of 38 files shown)


controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -959,6 +966,79 @@ private void handleStreamMetadataExists(final long timestamp, CompletableFuture<
...
+                result.complete(null);
+            }
+        }
+        
+        return result;
+    }
+
+    @Override
+    public CompletableFuture<Void> removeWriter(String writer) {
+        CompletableFuture<Void> result = new CompletableFuture<>();
```
tkaitchuck: Here and in the other methods in this file, you can use `CompletableFuture.completedFuture(null)`

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1173,6 +1180,65 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
         });
     }
 
+    @Override
+    public CompletableFuture<Void> recordCommitOffsets(final UUID txnId, final Map<Long, Long> commitOffsets) {
+        // The transaction may already have been committed and its record removed and this could be an idempotent run.
```
tkaitchuck: How does this handle the case were the transaction was committed and its record was removed but this was not successfully run?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1173,6 +1180,65 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
...
+     * transaction record for which a writer with time and position information is available. 
+     */
+    CompletableFuture<Void> generateMarksForTransactions(CommittingTransactionsRecord committingTransactionsRecord) {
+        return Futures.allOf(committingTransactionsRecord.getTransactionsToCommit().stream().map(txId -> {
+            int epoch = RecordHelper.getTransactionEpoch(txId);
+            // Ignore data not found exceptions. DataNotFound Exceptions can be thrown because transaction record no longer 
+            // exists and this is an idempotent case. DataNotFound can also be thrown because writer's mark was deleted 
+            // as we attempted to update an existing record. Note: Delete can be triggered by writer explicitly calling
+            // shutdownWriter api. 
+            return Futures.exceptionallyExpecting(getActiveTx(epoch, txId)
```
tkaitchuck: A local variable would help with the indentation here.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1486,89 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
+                });
+    }
+
+
+    /**
+     * Compares two given positions and returns true if position 2 is ahead of position 1. 
+     * 
+     * Note: This is not absolutely correct implementation because that would be a very costly computation. 
+     * This method simply checks if position 2 has a higher max segment number than position 1. 
+     * If those are equal, then it simply checks the offsets in each segment in position 2 to be ahead of same segment in
```
tkaitchuck: The implementation of the method actually is stronger than this comment. It compares all segment that overlap have increased.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1486,89 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
+     * @return return true if position 2 is ahead of/or equal to position 1. false otherwise. 
+     */
+    @VisibleForTesting
+    boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> position2) {
+        ArrayList<Map.Entry<Long, Long>> sortedPosition1 = new ArrayList<>(position1.entrySet());
+        sortedPosition1.sort(Collections.reverseOrder(Comparator.comparingLong(Map.Entry::getKey)));
+        ArrayList<Map.Entry<Long, Long>> sortedPosition2 = new ArrayList<>(position2.entrySet());
+        sortedPosition2.sort(Collections.reverseOrder(Comparator.comparingLong(Map.Entry::getKey)));
+
+        long maxInPos2Only = position2.keySet().stream().filter(position1::containsKey).max(Long::compare).orElse(Long.MIN_VALUE);
```
tkaitchuck: nit: "only" isn't really needed in the name.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -666,6 +672,49 @@
...
+        String writerPath = getWriterPath(writer);
+        return store.getData(writerPath, WriterMark::fromBytes);
+    }
+
+    @Override
+    CompletableFuture<Void> updateWriterMarkRecord(String writer, long timestamp, ImmutableMap<Long, Long> position, Version version) {
+        String writerPath = getWriterPath(writer);
+        WriterMark mark = new WriterMark(timestamp, position);
+
+        return Futures.toVoid(store.setData(writerPath, mark.toBytes(), version));
```
tkaitchuck: It seems like this would expensive. This is going to get called a LOT.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -509,5 +510,48 @@ private void reportDeleteStreamMetrics(String scope, String streamName, DeleteSt
...
+                    return response.build();
+                });
+    }
+ 
+    public CompletableFuture<Controller.WriterShutdownResponse> shutdownWriter(String scope, String stream, String writer) {
+        return streamStore.removeWriter(scope, stream, writer, null, executor)
+                .handle((r, e) -> {
+                    Controller.WriterShutdownResponse.Builder response = Controller.WriterShutdownResponse.newBuilder();
+                    if (e != null) {
+                        if (Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException) {
```
tkaitchuck: This needs test coverage.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -959,6 +966,79 @@ private void handleStreamMetadataExists(final long timestamp, CompletableFuture<
...
 
+    @Override
+    CompletableFuture<Void> createWriterMarkRecord(String writer, long timestamp, ImmutableMap<Long, Long> position) {
+        CompletableFuture<Void> result = new CompletableFuture<>();
+        WriterMark mark = new WriterMark(timestamp, position);
+
+        synchronized (writersLock) {
+            VersionedMetadata<WriterMark> existing = writerMarks.get(writer);
+            if (existing != null) {
+                result.completeExceptionally(StoreException.create(StoreException.Type.DATA_EXISTS, "writer mark exists"));
```
andreipaduroiu: You can just return `new CompletableFuture(...)` or `Futures.failedFuture(...)` instead of creating a `result` variable. That will force whomever will modify this in the future to always return something (there is a possibility that they may just return an incomplete Future).

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
andreipaduroiu: Both this and the next lines are guaranteed to be O(n). You can make this at most O(n) by looping through the already sorted positions (descending) which will prevent you from doing unnecessary comparisons.

andreipaduroiu: Actually why do you sort them anyway? You never seem to use `sortedPosition1` or ``sortedPosition2` anywhere.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1486,89 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
+        long maxInPos1Only = position1.keySet().stream().filter(position2::containsKey).max(Long::compare).orElse(Long.MIN_VALUE);
+        
+        boolean compareMaxes = maxInPos2Only >= maxInPos1Only;
+        
+        boolean compareOverlaps = position2.entrySet().stream().filter(x -> position1.containsKey(x.getKey()))
```
andreipaduroiu: Do this computation lazily, only if `compareMaxes` is false.

andreipaduroiu: Also you are double-getting from `position1`. Once in `filter` and once in `allMatch`.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -509,5 +510,48 @@ private void reportDeleteStreamMetrics(String scope, String streamName, DeleteSt
         }
     }
 
+    public CompletableFuture<Controller.TimestampResponse> noteTimestampFromWriter(String scope, String stream, String writer, long timestamp, Map<Long, Long> streamCut) {
+        return streamStore.noteWriterMark(scope, stream, writer, timestamp, streamCut, null, executor)
```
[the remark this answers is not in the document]
shiveshr: The problem in doing it at controller is, these request could land up different controller instances. 
A writer is best suited to handle it IMO

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: Txn record removal is the last step. Record commit offsets is an intermediate step during commit workflow. So if this was not run successfully then txn removal would not have happened. 

Here is the flow:
we first merge transaction segments. Then we fetch the size. Then update the commitOffsets in transaction's metadata. 
Once we have completed above operations for a batch, for each transaction in the committing set, we generate their marks and then move them to committed state (remove activeTxnRecord and create a new CompletedTxnRecord).

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: The comment mentions that
>       * If those are equal, then it simply checks the offsets in each segment in position 2 to be ahead of same segment in

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: Actually you are right.. I was doing more checks with sorting.. but i realized those checks are inconclusive. I intended to not have the sorted as it is useless.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: Yes, but its inevitable. We need to store marks as these requests could be received on different controller instances. Besides we dont hold all of this in memory anyway.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: Actually, this was a miss on my part. I had a simple check in the first commit. Then i tried improving the comparison check for two positions and realized that without retrieving metadata information about ranges corresponding to segments, it would not be possible to draw any stronger conclusions so i reverted my changes but did not clean it properly. I dont need sorted position sets. I just need to find max and compare maxes.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
shiveshr: Checkstyle sometimes complains about number of `returns` from a method.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
tkaitchuck: Right, it doesn't just do that check "if they are equal" it does it all the time.

shiveshr: Fixed the impl to do the latter iff they are equal

tkaitchuck: That's the opposite of what I was suggesting. The impl as it was was fine. I just wanted the doc to reflect it.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
tkaitchuck: There should be a limit, but that limit is not 1. So having one for each branch of the if is ok.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1173,6 +1180,64 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
...
+     * transaction record for which a writer with time and position information is available. 
+     */
+    CompletableFuture<Void> generateMarksForTransactions(CommittingTransactionsRecord committingTransactionsRecord) {
+        return Futures.allOf(committingTransactionsRecord.getTransactionsToCommit().stream().map(txId -> {
+            int epoch = RecordHelper.getTransactionEpoch(txId);
+            // Ignore data not found exceptions. DataNotFound Exceptions can be thrown because transaction record no longer 
+            // exists and this is an idempotent case. DataNotFound can also be thrown because writer's mark was deleted 
+            // as we attempted to update an existing record. Note: Delete can be triggered by writer explicitly calling
+            // shutdownWriter api. 
+            return Futures.exceptionallyExpecting(getActiveTx(epoch, txId).thenCompose(txnRecord -> {
```
tkaitchuck: I think it would help clarity to have a local variable and do the return Futures.exceptionallyExpecting on the next line.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1485,87 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
+                    if (record == null) {
+                        // Attempt to create a new record. It is possible that while we are attempting to create the record, 
+                        // a concurrent request could have created it. 
+                        // For example: a writer sends a request to one of controller instances to note writer mark and 
+                        // the connection is dropped. It could then send the request to noteTime to a different controller 
+                        // instance. So two controller instances could attempt concurrent creation of mark for the said writer.
+                        // In this case, we may have found mark to be inexistent when we did a get but it may have been created 
+                        // when we attempt to create it. So immediately after attempting a create, if we get DataExists, 
+                        // we will call noteWriterMark recursively. 
+                        return Futures.exceptionallyComposeExpecting(
```
tkaitchuck: same here

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1485,87 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
+     * @return return true if position 2 is ahead of/or equal to position 1. false otherwise. 
+     */
+    @VisibleForTesting
+    boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> position2) {
+        long maxInPos2 = position2.keySet().stream().filter(position1::containsKey).max(Long::compare).orElse(Long.MIN_VALUE);
+        long maxInPos1 = position1.keySet().stream().filter(position2::containsKey).max(Long::compare).orElse(Long.MIN_VALUE);
+        
+        boolean compareMaxes = maxInPos2 >= maxInPos1;
+        
+        if (compareMaxes) {
```
tkaitchuck: The previous implementation was better. See my above comment.

controller/src/main/java/io/pravega/controller/store/stream/records/WriterMark.java
```diff
@@ -0,0 +1,84 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.SneakyThrows;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+
+@Data
+public class WriterMark {
```
RaulGracia: Missing javadoc.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: In previous implementation, i was trying some additional checks but realized they could not be done without making this computation more costly ( as it would require fetching segment ranges from store). So i had reverted them. But there were some partial changes that had lingered on in the first iteration of the PR (like sorting the segments in the position object) which was not being used. 

Do you have suggestions for additional checks apart from that which is present now that we can include? Note: we do not want to make too many calls into store for each mark that is reported by a writer.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -331,14 +332,14 @@ private SegmentRange convert(final String scope,
         return listOfSegment;
     }
 
-    public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String stream, final TxnId
-            txnId) {
+    public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String stream, final TxnId txnId, 
+                                                          final String writerId, final long timestamp) {
         Exceptions.checkNotNullOrEmpty(scope, "scope");
```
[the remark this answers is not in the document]
shiveshr: WriterId could be null, or timestamp negative if the request comes from an older client which does not have watermarking supported.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -331,14 +332,14 @@ private SegmentRange convert(final String scope,
         return listOfSegment;
     }
 
-    public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String stream, final TxnId
-            txnId) {
+    public CompletableFuture<TxnStatus> commitTransaction(final String scope, final String stream, final TxnId txnId, 
+                                                          final String writerId, final long timestamp) {
```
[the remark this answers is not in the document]
shiveshr: This is not on grpc. This is an internal method. 
Grpc contract is both forward and backward compatible because we add new fields to the rpc api contract.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1485,87 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
                 });
     }
 
+    @Override
+    public CompletableFuture<WriterTimestampResponse> noteWriterMark(String writer, long timestamp, Map<Long, Long> position) {
```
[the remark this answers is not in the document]
shiveshr: > Could it be the case, with an unfortunate succession of re-connections, that a writer executes two close-by noteWriterMark calls to two different Controller instances 

yes

> and the watermarks are stored in the metadata in a different order compared to how the writer sent them

a writer has only its latest mark reported. And updates to store use optimistic concurrency. 
So its not about order. Its about one of the marks being stored first.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1485,87 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
+                                .thenApply(v -> WriterTimestampResponse.SUCCESS);
+                    }
+                });
+    }
+
+
+    /**
+     * Compares two given positions and returns true if position 2 is ahead of position 1. 
+     * 
+     * Note: This is not absolutely correct implementation because that would be a very costly computation.
```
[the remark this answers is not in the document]
shiveshr: Typically a writer reporting a mark will always advance time as well with each new mark. 

And if this method is called from commit-transaction, then the commit happens against a specific epoch. So the highest segment number check suffices there. 

And as long as writer is advancing time in its transaction commits as well, the check to compare writer marks is more of a sanity check to prevent rouge writers or tackle some bug.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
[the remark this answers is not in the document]
shiveshr: a writer is aggregating writes only for itself. 
controller recieves writes from multiple writers and stores them.
A background job periodically scans the reported marks from across writers and consolidates them and emits a watermark.

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
```diff
@@ -462,14 +462,14 @@ public void testTransaction() throws Exception {
                 context, executor).get();
```
fpj: I don't see unit tests for the new methods introduced in `ZkStream`.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -919,6 +919,11 @@ private SegmentRange convert(String scope, String stream, Map.Entry<Long, Map.En
                 this.retrieveDelegationToken()), executor);
     }
 
+    public CompletableFuture<Map<Long, Long>> getCurrentSegmentsSize(String scope, String stream, List<Long> segments) {
```
fpj: `getCurrentSegmentSizes`?

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
```diff
@@ -960,7 +960,8 @@ public void sealStreamWithTxnTest() throws Exception {
                 .get().getKey();
```
fpj: I'm not seeing `getCurrentSegmentSizes` declared in this PR being unit tested.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
shiveshr: No, because writers call can reach any arbitrary controller. So either we build some mechanism with which all writers are redirected to the same controller instance to report marks for a given stream. 
Or we durably store the writer marks and then perform some computation on them asynchronously with some periodicity.
We decided to do the latter in the PDP. 
The disadvantage of doing former is, we need to make explicit grpc call into a specific controller instance or redirect the request from one controller to another. The controller will need to maintain in memory state for all writers, which in turn will put memory pressure on it. And in case of failure, we lose marks reported by all writers and that can lead to missing writer positions in the watermark computation if the writer doesnt report another mark.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: I replied to the other comment, the implementation was doing unnecessary sorting and not making use of it. I changed that. And as for two conditions that are being checked, if first condition fails, we no longer check second condition. the java doc reflects that.

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
shiveshr: Its merely calling getSegmentOffset which is tested.

controller/src/test/java/io/pravega/controller/store/stream/ZkStreamTest.java
shiveshr: The tests are introduced at base Stream test level (StreamTestBase) and it gets tested for all three store types.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -432,6 +432,35 @@ public void getDelegationToken(StreamInfo request, StreamObserver<DelegationToke
...
     }
 
+    // region watermarking apis
+    
+    @Override
+    public void noteTimestampFromWriter(Controller.TimestampFromWriter request, StreamObserver<Controller.TimestampResponse> responseObserver) {
+        StreamInfo streamInfo = request.getPosition().getStreamInfo();
+        log.info("noteWriterMark called for stream {}/{}, writer={} time={}", streamInfo.getScope(),
+                streamInfo.getStream(), request.getWriter(), request.getTimestamp());
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
```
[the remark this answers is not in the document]
shiveshr: CheckAuthorizationAndCreateToken is used when we want to send token back to client to enable them to connect to segment store nodes. 
So it is relevant for `getCurrentSegments`, `getSuccessor` like apis. For apis where client simply asks controller to perform some action on its behalf, it could be a get or an update request, there is no need to generate token.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -432,6 +432,35 @@ public void getDelegationToken(StreamInfo request, StreamObserver<DelegationToke
...
+                        streamInfo.getStream(), request.getWriter(), request.getTimestamp(), request.getPosition().getCutMap()),
+                responseObserver);
+    }
+
+    @Override
+    public void writerShutdown(Controller.WriterShutdownRequest request, StreamObserver<Controller.WriterShutdownResponse> responseObserver) {
+        StreamInfo streamInfo = request.getStream();
+        log.info("writerShutdown called for stream {}/{}, writer={}", streamInfo.getScope(),
+                streamInfo.getStream(), request.getWriter());
+        authenticateExecuteAndProcessResults(() -> this.authHelper.checkAuthorization(
```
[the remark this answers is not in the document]
shiveshr: replied above

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -509,5 +510,48 @@ private void reportDeleteStreamMetrics(String scope, String streamName, DeleteSt
...
+                                response.setResult(Controller.TimestampResponse.Status.INTERNAL_ERROR);
+                                break;
+                        }
+                    }
+                    return response.build();
+                });
+    }
+ 
+    public CompletableFuture<Controller.WriterShutdownResponse> shutdownWriter(String scope, String stream, String writer) {
+        return streamStore.removeWriter(scope, stream, writer, null, executor)
```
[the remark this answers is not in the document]
shiveshr: Shutdown writer is the grpc call. Its connotation is to tell controller to stop performing any action on behalf of the said writer wrt watermarking. 
The way controller makes use of writers is to store their marks durably in the store. And then a periodic workflow scans the store to retrieve latest marks stored against all known writers and then compute and emit a watermark (if any). 
So to shutdown computation for a writer really is about removing the writer from the underlying store so that we do not consider it for computation. Hence remove writer.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -509,5 +510,48 @@ private void reportDeleteStreamMetrics(String scope, String streamName, DeleteSt
...
+                            default:
+                                response.setResult(Controller.TimestampResponse.Status.INTERNAL_ERROR);
+                                break;
+                        }
+                    }
+                    return response.build();
+                });
+    }
+ 
+    public CompletableFuture<Controller.WriterShutdownResponse> shutdownWriter(String scope, String stream, String writer) {
```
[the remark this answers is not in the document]
shiveshr: replied here https://github.com/pravega/pravega/pull/3761#discussion_r287914194

controller/src/main/java/io/pravega/controller/server/ControllerService.java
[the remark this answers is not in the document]
shiveshr: In fact, i just realized, we should have both shutdown and remove semantics on the store. 
This is because of the following:
if a writer reports a writer mark, its expectation is that this mark will be included in the watermark computation.
But watermark computation happens asynchronously. 
So if the writer shutdown is called before the asynchronous computation is performed, and we remove the writer mark, then we do not include the writer's mark in the computation. 
We should mark the writer for shutdown and it should be removed from computation if it becomes inactive and not before that.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -432,6 +432,35 @@ public void getDelegationToken(StreamInfo request, StreamObserver<DelegationToke
...
+                AuthResourceRepresentation.ofStreamInScope(streamInfo.getScope(), streamInfo.getStream()),
+                AuthHandler.Permissions.READ_UPDATE),
+                delegationToken  -> controllerService.noteTimestampFromWriter(streamInfo.getScope(),
+                        streamInfo.getStream(), request.getWriter(), request.getTimestamp(), request.getPosition().getCutMap()),
+                responseObserver);
+    }
+
+    @Override
+    public void writerShutdown(Controller.WriterShutdownRequest request, StreamObserver<Controller.WriterShutdownResponse> responseObserver) {
```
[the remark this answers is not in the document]
shiveshr: replied above. I am adding a shutdown semantics on top of remove writer .

Shutdown is called from writerShutdown which will only mark writer to be shutdown. 
A shutdown writer is removed eventually once its marks are no longer relevant.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1173,6 +1180,67 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
...
+                        return Futures.toVoid(updateActiveTx(epoch, txnId, new VersionedMetadata<>(updated, txnRecord.getVersion())));
+                    } else {
+                        return CompletableFuture.completedFuture(null);
+                    }
+                }), DATA_NOT_FOUND_PREDICATE, null);
+    }
+
+    /**
+     * This method takes the list of transactions in the committing transactions record, and tries to report marks
+     * for writers for these transactions, if they exist.
```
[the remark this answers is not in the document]
shiveshr: Basically older writers may not report their commit time and writer id. In which case we will not attempt to report their time.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -99,6 +101,11 @@
     @GuardedBy("txnsLock")
     private final Map<Integer, Set<UUID>> epochTxnMap = new HashMap<>();
 
+    private final Object writersLock = new Object();
+
+    @GuardedBy("writersLock")
```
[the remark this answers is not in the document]
shiveshr: I have followed that same pattern across the class. Basically in case of failure scenarios like Data version mismatch etc, handling it inside computeIfAbsent/Present makes the code more complex. 
This is more readable.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -959,6 +966,73 @@ private void handleStreamMetadataExists(final long timestamp, CompletableFuture<
         return CompletableFuture.completedFuture(null);
     }
 
+    @Override
+    CompletableFuture<Void> createWriterMarkRecord(String writer, long timestamp, ImmutableMap<Long, Long> position) {
+        WriterMark mark = new WriterMark(timestamp, position);
+
+        synchronized (writersLock) {
```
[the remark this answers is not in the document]
shiveshr: Replied above

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1173,6 +1180,67 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
...
+                }), DATA_NOT_FOUND_PREDICATE, null);
+    }
+
+    /**
+     * This method takes the list of transactions in the committing transactions record, and tries to report marks
+     * for writers for these transactions, if they exist. 
+     * WriterId, commit time and commit offsets are recorded in ActiveTxnRecord for each transaction. 
+     * For transactions where above fields are present, a mark is recorded for them. 
+     * This method will ignore any INVALID_TIME or INVALID_POSITION related failures in noting marks for writers.
+     * This is because those will typically arise from idempotent commit case where this and a transaction with higher
```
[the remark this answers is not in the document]
shiveshr: Say we have 3 txns from writer1 in following order txn1 (time = t1), txn2 (time = t2), txn3 (time = t3). 
We batch them together and commit in order :
1. commit txn1 at pos1
2. note marker for pos1, 
3. commit txn2 at pos2
4. note marker for pos2

Say processing failed here and the whole commit workflow is retried. 

So writer's latest noted position is pos2 and time t2. 

now if commit workflow is tried again, txn1 commit will have pos1 which is before pos2 and noteWriterTime will fail with INVALID_TIME/INVALID_POSITION.
This will be ignored and we continue with txn2. This will also fail with INVALID_TIME. 

Then we proceed to txn3, we commit it at pos3. The writermark is updated with pos3/t3.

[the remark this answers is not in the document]
shiveshr: Maybe you can suggest a better phrasing for the above.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1420,6 +1488,89 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
+                        // instance. So two controller instances could attempt concurrent creation of mark for the said writer.
+                        // In this case, we may have found mark to be inexistent when we did a get but it may have been created 
+                        // when we attempt to create it. So immediately after attempting a create, if we get DataExists, 
+                        // we will call noteWriterMark recursively. 
+                        CompletableFuture<WriterTimestampResponse> future = createWriterMarkRecord(writer, timestamp, newPosition)
+                                .thenApply(v -> WriterTimestampResponse.SUCCESS);
+                        
+                        return Futures.exceptionallyComposeExpecting(future,
+                                e -> Exceptions.unwrap(e) instanceof StoreException.DataExistsException, 
+                                () -> noteWriterMark(writer, timestamp, position));
```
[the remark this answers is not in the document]
shiveshr: The method does following:
get --> if exists --> `update` --> else `create`

So we recurse only if create gets DataExists. Basically we could have called get and update but the method already does that so we call recursively. 

Even if two controllers try simultaneously, once they find that record exists, they will either be able to successfully update or fail with BadVersion. So they will terminate.

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
```diff
@@ -555,6 +570,36 @@
...
+     * @param writer writer id
+     * @return A completableFuture, which when completed, will contain writer's mark.  
+     */
+    CompletableFuture<WriterMark> getWriterMark(String writer);
+
+    /**
+     * Method to retrieve latest recorded mark for all known writers.  
+     * @return A completableFuture, which when completed, will contain map of writer to respective marks.  
+     */
+    CompletableFuture<Map<String, WriterMark>> getAllWritersMarks();
```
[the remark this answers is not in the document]
shiveshr: `WriterMarks` may imply there could be multiple marks for the same writer. 
`Writers``Marks` says we get marks for all writers.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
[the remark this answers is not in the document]
fpj: Can you update the comment, please?

controller/src/main/java/io/pravega/controller/server/ControllerService.java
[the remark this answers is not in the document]
fpj: There are two parts to this discussion:

*Method name*
`remove` is a much more natural term to refer to what we are doing here. We have a group of writers for which we are collecting watermarks. One element of the group is being decommissioned and needs to be removed. Whether that element is going to be shutdown or not is being the scope of this call. I think we should not use `shutdown` here.

*Semantics of removal*
It is correct that the asynchronous nature of the computation is such that operations can happen out of order. But, instead of creating another call to fix the problem with semantics, we should instead serialize the operations in the case the property is really needed. I'm not in favor of a second call, it seems unnecessary and confusing.

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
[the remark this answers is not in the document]
fpj: Compound nouns in English are not supposed to plural the qualifier, which is `writer` in this case. There is no call to get all marks of a single writer, so there is no ambiguity. In the case we do want to have a call for all marks of writers and one for all marks of a single writer, we can use:

- `getWriterMarks`: all marks of a single writer
- `getAllWriterMarks`: all marks of all writers

The javadocs can help with clarifying the semantics.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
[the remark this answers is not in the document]
shiveshr: > Semantics of removal
> It is correct that the asynchronous nature of the computation is such that operations can happen out of order. But, instead of creating another call to fix the problem with semantics, we should instead serialize the operations in the case the property is really needed. I'm not in favor of a second call, it seems unnecessary and confusing.

You misunderstood perhaps. The writers report their marks. Once they have reported it, their expectation would be that it should be included in watermarking computation. 
They could then call shutdown/remove. If we remove the writer mark before its included in the watermark computation then we break the expectation. 
Since watermark computation is asynchronous and periodic, the way i have implemented the logic is as follows (this is in the followup PR and you can refer to the draft PR open here: #3839 ). 
When a writer requests for a shutdown, Controller simply updates a flag inside the writer's mark that it has been shutdown. 
During watermarking computation, after having consumed the writer's mark, if the writer has been marked for shutdown, the workflow also removes the writer. 

So shutdown and remove have separate semantics from metadata store's perspective. 
From application's perspective, they have the notion of `shutdown` of a writer. We could rename it to some other verb like remove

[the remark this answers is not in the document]
fpj: I believe I understood the problem right, but perhaps did not express my thought correctly. I understood that there can be a race in the case a writer notes a mark followed by removing itself. Due to asynchronous and unordered nature of the computation, it is possible that the removal is processed before the mark is noted. What I am saying is that the way to solve it is to make sure the operations are linearized. Even if they are executed asynchronously, they can still following submission order.

The part I have probably misunderstood is your proposal. Reading again, you say:

> i just realized, we should have both shutdown and remove semantics on the store.

which I initially interpreted as a separate API call altogether, put perhaps you meant to say that this is a way to avoid the race.

> From application's perspective, they have the notion of shutdown of a writer. We could rename it to some other verb like remove

The application might be closing a client, and the client needs to remove itself from this note group. Any synonym of "remove" will do for me. :-)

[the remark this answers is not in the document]
shiveshr: So Remove Writer and Watermark computation are linearized in my proposal. 

Call from caller merely marks the writer for shutdown. 
The removal is performed by next watermark computation after having consumed the writer's last reported mark. 

So for name change -- I could rename the grpc api from `ShutdownWriter` to `CloseWriter`.

RaulGracia:
@shiveshr any idea of why codecov check is not there?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1875,5 +2026,13 @@ private StreamSegmentRecord newSegmentRecord(int epoch, int segmentNumber, long
 
     abstract CompletableFuture<Void> deleteWaitingRequestNode();
     // endregion
+
+    // region watermarking
+    abstract CompletableFuture<Void> createWriterMarkRecord(String writer, long timestamp, ImmutableMap<Long, Long> position);
```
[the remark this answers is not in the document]
RaulGracia: Good point. Moreover, I think that we could elaborate a little bit more on the description of these methods (semantics, guarantees) and move the comment on the `DataExistsException` to the description of the method as well (if there was a checked exception, then you could add it in the `@thows` block, but without it maybe that explanation fits better in the description of the method).

PR #3763, opened 2019-05-08
Issue 3762: Reduced frequency of logging in ZKSegmentContainerMonitor 

### LINKED ISSUE #3762: ZKSegmentContainerMonitor log output is too verbose
**Problem description**
`ZKSegmentContainerMonitor.checkAssignment` outputs every 10 seconds, even if there is nothing to do.

The longevity runs logs are filled with these log lines (if set to INFO) level and it's hard to find anything else.

**Problem location**
`ZKSegmentContainerMonitor`.

1. Only output if there is something to do.
2. Only output every 10 minutes if there is nothing to do.

andreipaduroiu:
**Change log description**  
-  Changed `ZKSegmentContainerMonitor.checkAssignment` to only log if there is a pending task in progress or a Segment Container change desired. If none, then log every 10 minutes.

**Purpose of the change**  
Fixes #3762.

**What the code does**  
This reduces the litter in the logs. It previously logged an entry every 10 seconds and filled up entire logs with exactly the same message.

**How to verify it**  
Build and tests must pass.

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java
@@ -9,6 +9,8 @@
  */
 package io.pravega.segmentstore.server.host;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import io.pravega.common.Exceptions;
 import io.pravega.common.LoggerHelpers;
 import io.pravega.common.cluster.Host;
@@ -17,8 +19,6 @@ import io.pravega.common.concurrent.Futures;
 import io.pravega.common.util.CollectionHelpers;
 import io.pravega.segmentstore.server.ContainerHandle;
 import io.pravega.segmentstore.server.SegmentContainerRegistry;
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java
```diff
@@ -175,9 +180,14 @@ private void checkAssignment() {
                 Collection<Integer> containersToBeStopped = CollectionHelpers.filterOut(runningContainers, desiredList);
                 containersToBeStopped = CollectionHelpers.filterOut(containersToBeStopped, containersPendingTasks);
 
-                log.info("Container Changes: Desired = {}, Current = {}, PendingTasks = {}, ToStart = {}, ToStop = {}.",
-                        desiredList, runningContainers, containersPendingTasks, containersToBeStarted,
-                        containersToBeStopped);
+                // Only report if we have any changes or if sufficient time has elapsed since the last report.
+                boolean logReport = !(containersPendingTasks.isEmpty() || containersToBeStarted.isEmpty() || containersToBeStopped.isEmpty());
```
RaulGracia: As far as I understand, we want to log if there is "any" change in the container assignment or state, right? In this case, we don't want to log if "all" of the collections containing the changes in the containers are empty. Thus, shouldn't the condition be `!(containersPendingTasks.isEmpty() && containersToBeStarted.isEmpty() && containersToBeStopped.isEmpty())`? If not, as long as there is a single empty collection, the log wouldn't be reported (despite there could be changes stored in the rest of collections).

andreipaduroiu: Yes, you are right. I need some coffee...

PR #3767, opened 2019-05-09
Issue 3463: Ensure completed transactions are not pinged by the client.

### LINKED ISSUE #3463: Pravega client continues to ping the same transaction forever
**Problem description**

The Pravega client repeatedly shows messages such as the one below. The message is repeated several times a minute over at least an 8 hour period. All transactions that have ever been opened in the client process are included. This suggests that the transaction pinger is never removing transactions from the list of transactions to ping, even after they have been committed.

```
WARN  [2019-03-14 09:26:08.088] [grpc-default-executor-11] i.p.c.stream.impl.ControllerImpl: gRPC call for pingTransaction with trace id 0 failed with server error.
io.grpc.StatusRuntimeException: INTERNAL: /transactions/activeTx/examples20/STREAMPROCESSING2/0/00000000-0000-0000-0000-000000000001
        at io.grpc.Status.asRuntimeException(Status.java:526)
        at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:418)
        at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
        at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
        at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:663)
        at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
        at io.grpc.internal.CensusTracingModule$TracingClientInterceptor$1$1.onClose(CensusTracingModule.java:392)
        at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:443)
        at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:63)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:525)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.access$600(ClientCallImpl.java:446)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:557)
        at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
        at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
```

The following code reproduces this error:
https://github.com/claudiofahey/pravega-samples/blob/issue-191-streamprocessing/pravega-client-examples/src/main/java/io/pravega/example/streamprocessing/ExactlyOnceMultithreadedProcessor.java#L420

Of particular note, the client that begins the transaction does not commit or abort it. It relies on another process to get the transaction by transaction ID and commit it. For this reason, the client never calls `Pinger.stopPing`, causing transactions to never be removed from the Pinger list.

This occurs with Pravega 0.4.0. Other versions have not been tested. Pravega was started using `docker-compose up`.

**Problem location**

Possibly `io.pravega.client.stream.impl.Pinger.pingTransactions`.

`io.pravega.client.stream.impl.Pinger.pingTransactions` should remove non-open transactions from `txnList`. It should not rely on the client calling `abort` or `commit`.
Alternatively, make `Transaction` closeable and add `stopPing` to the `close`. This allows the pinger to stop pinging the transaction even when we don't want that client to abort or commit.

shrids:
**Change log description**  
- Ensure Pravega client does not ping transactions that have been completed.

**Purpose of the change**  
Fixes #3463

**What the code does**  
`io.pravega.client.stream.impl.Pinger` is used to ping transactions by the client. With this change set the `Pinger` stops pinging transactions that are in 
`Transaction.Status#COMMITTING`, `Transaction.Status#COMMITTED`, `Transaction.Status#ABORTING` or `Transaction.Status#ABORTED` state.

**How to verify it**  


client/src/main/java/io/pravega/client/stream/impl/Pinger.java
```diff
@@ -89,14 +98,21 @@ private void startPeriodicPingTxn() {
      */
     private void pingTransactions() {
         log.info("Start sending transaction pings.");
+
+        List<UUID> stopPingTxns = new ArrayList<>();
+        completedTxns.drainTo(stopPingTxns);
+        txnList.removeAll(stopPingTxns); // remove all completed transacations from the txn list which need to be pinged.
```
RaulGracia: Typo `transacations` -> `transactions`

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
```diff
@@ -66,11 +75,11 @@ void stopPing(UUID txnID) {
 
     private long getPingInterval(long txnLeaseMillis) {
         double pingInterval = txnLeaseMillis * PING_INTERVAL_FACTOR;
-        if (pingInterval < TimeUnit.SECONDS.toMillis(5)) {
+        if (pingInterval < TimeUnit.SECONDS.toMillis(10)) {
```
RaulGracia: I have seen this number (`10`) used in several places. Please, consider using a constant to refer to it.

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
shrids: fixed it.

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
shrids: done.

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
```diff
@@ -89,14 +99,21 @@ private void startPeriodicPingTxn() {
      */
     private void pingTransactions() {
         log.info("Start sending transaction pings.");
-        txnList.stream().forEach(uuid -> {
+
+        List<UUID> stopPingTxns = new ArrayList<>();
```
andreipaduroiu: 1. This is using `ArrayList` and `LinkedBlockingQueue`, both of which have O(n) runtime for all operations except `add`. I don't see you using any feature that requires this to be a List or a Queue (no ordering anywhere, no getFirst/Last), therefore I suggest you make both of these `HashSet`s and use set-specific methods on that class to do your computation.
2. I also suggest you use a common lock object (could be one of the sets) to synchronize access to these sets. This method executes at a fixed rate (not interval) which means it is possible that two or more executions overlap thus corrupting your state. 
3. Consider using `SequentialProcessor` from `common.concurrent` to avoid overlapping execution of requests if what I said above (at 2.) can cause issues.

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
```diff
@@ -40,6 +47,9 @@
     private ScheduledExecutorService executor = ExecutorServiceHelpers.newScheduledThreadPool(1,
             "pingTxnThread");
     private final List<UUID> txnList = Collections.synchronizedList(new ArrayList<>());
+    @Getter(value = AccessLevel.PACKAGE)
```
andreipaduroiu: This is about the line above (47):
1. Why do you need a dedicated thread pool? Each execution of this class' methods is pretty quick so we are wasting a thread for nothing.
2. You have a constructor that takes in an external executor, yet in `close` you shut it down. You shouldn't do that as someone else may be using it.

If all you need an Executor for is to be able to stop the pinger, then `scheduleAtFixedRate` returns a `ScheduledFuture` which can be cancelled when you close this class.

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
shrids: Done, changed to use `HashSet`. 
Also, `java.util.concurrent.ScheduledExecutorService#scheduleAtFixedRate` ensure that there are no concurrent executions of the command, pingTransactions(), hence will not use `SequentialProcessor`.

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
shrids: Yes, there is no need of an extra thread pool. 
Refactored the implementation to reuse the internal executor provided by `io.pravega.client.netty.impl.ConnectionFactory#getInternalExecutor`

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
```diff
@@ -87,8 +88,8 @@ private long getPingInterval(long txnLeaseMillis) {
     private void startPeriodicPingTxn() {
         if (!isStarted.get()) {
             log.info("Starting Pinger at an interval of {}ms ", this.pingIntervalMillis);
-            executor.scheduleAtFixedRate(this::pingTransactions, 10, this.pingIntervalMillis,
-                    TimeUnit.MILLISECONDS);
+            // scheduleAtFixedRate ensure that there are no concurrent executions of the command, pingTransactions()
+            scheduledFuture = executor.scheduleAtFixedRate(this::pingTransactions, 10, this.pingIntervalMillis, TimeUnit.MILLISECONDS);
```
andreipaduroiu: `scheduledFuture` is not guarded by anything so it's not thread safe. 

If you make this `AtomicReference<ScheduledFuture>` then you will have solved that and eliminate the need for `isStarted` (you can just check if this object has anything in it).

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
```diff
@@ -99,32 +100,34 @@ private void startPeriodicPingTxn() {
...
+            });
+        }
         log.trace("Completed sending transaction pings.");
     }
 
     @Override
     public void close() {
         log.info("Closing Pinger periodic task");
-        ExecutorServiceHelpers.shutdown(executor);
+        if (scheduledFuture != null) {
```
andreipaduroiu: This is not thread safe. In between line 129 and 130, someone could have executed some code to set this to null.

If you change this to `AtomicReference`, then you can do:
```
val toCancel = scheduledFuture.getAndSet(null);
if(toCancel != null){
     toCancel.cancel(true);
}
```

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
```diff
@@ -306,7 +306,7 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
     }
 
     private void sendAndVerifyEvent(UUID cid, ClientConnection connection, SegmentOutputStreamImpl output,
-            ByteBuffer data, int num) throws SegmentSealedException, ConnectionFailedException {
+            ByteBuffer data, int    num) throws SegmentSealedException, ConnectionFailedException {
```
andreipaduroiu: is there an extra space here?

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
shrids: Used `AtomicReference` for the `ScheduledFuture` and used `AtomicBoolean isStarted` to remove @Synchronized.

client/src/main/java/io/pravega/client/stream/impl/Pinger.java
shrids: done, used `AtomicReference`.

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
shrids: removed this change.

Issue #3725, opened 2019-04-30
Sporadic test failure MetricsTest.metricsTimeBasedCacheEvictionTest


RaulGracia:
**Problem description**
We are observing `MetricsTest.metricsTimeBasedCacheEvictionTest` test failing sporadically both on Travis:
```
io.pravega.test.integration.MetricsTest > metricsTimeBasedCacheEvictionTest FAILED
    java.lang.NullPointerException
        at io.pravega.test.integration.MetricsTest.lambda$metricsTimeBasedCacheEvictionTest$1(MetricsTest.java:247)
        at io.pravega.test.common.AssertExtensions.assertEventuallyEquals(AssertExtensions.java:48)
        at io.pravega.test.integration.MetricsTest.metricsTimeBasedCacheEvictionTest(MetricsTest.java:246)
```
and Jenkins:
```
java.lang.NullPointerException
	at io.pravega.test.integration.MetricsTest.lambda$metricsTimeBasedCacheEvictionTest$0(MetricsTest.java:203)
	at io.pravega.test.common.AssertExtensions.assertEventuallyEquals(AssertExtensions.java:48)
	at io.pravega.test.integration.MetricsTest.metricsTimeBasedCacheEvictionTest(MetricsTest.java:202)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```
See the [complete log for Jenkins failure](https://github.com/pravega/pravega/files/3133206/failure_log.txt).

**Problem location**
`MetricsTest.metricsTimeBasedCacheEvictionTest`

**Suggestions for an improvement**
Fix the test.

kevinhan88:
@RaulGracia could you please add at least the commit id of the Travis/Jenkins job for the job failure?

RaulGracia:
@kevinhan88 I attached the full log for the test. I got that log from build number 4 in the `pravega-0.5` Jenkins job. What else do you need?

kevinhan88:
Thanks @RaulGracia

kevinhan88:
The context of the issue:
Changes
Issue 3659: Set hdfs.replaceDataNodesOnFailure = false in K8s system (commit: c0bee07) (detail / githubweb)
[...]

PR #3769, opened 2019-05-09
Issue 3768: Increase timeout for pod/task-pv-pod creation

### LINKED ISSUE #3768: Timeout waiting for task-pv-pod to become ready
**Problem description**

When running the system tests in Kubernetes, the script that prepares the environment to run the system tests fails because the wait condition times out after 30 seconds. This failure is intermittent and depends on how fast Kubernetes is able to create the pod. Typically, it takes a bit less than 30 seconds, but occasionally it takes a bit more at the build fails.

```
$ kubectl wait --help
...
--timeout=30s: The length of time to wait before giving up.  Zero means check once and don't wait, negative means wait for a week.
...
```

**Problem location**

https://github.com/pravega/pravega/blob/a81147232c1e427b418e8131c73955d3e786295d/test/system/kubernetes/setupTestPod.sh#L95

Increase the timeout from the default 30 seconds to 1 minute to have a safety margin when the pod creation takes longer than normal.

adrianmo:
**Change log description**  
- Increase timeout for test pod (`pod/task-pv-pod`) creation from 30 seconds (default value) to 1 minute

**Purpose of the change**  
Fix #3768 

**What the code does**  
Applies the changes to the `test/system/kubernetes/setupTestPod.sh` script.

**How to verify it**  
System test should pass consistently. 

---
Signed-off-by: Adrián Moreno <adrian@morenomartinez.com>

### DIFF
--- a/test/system/kubernetes/setupTestPod.sh
+++ b/test/system/kubernetes/setupTestPod.sh
@@ -92,7 +92,7 @@ spec:
         - mountPath: "/data"
           name: task-pv-storage
 EOF
-kubectl wait --for=condition=Ready pod/task-pv-pod
+kubectl wait --timeout=1m --for=condition=Ready pod/task-pv-pod
 
 #Step 7: Compute the checksum of the local test artifact and the artifact on the persistent volume. Copy test artifact only if required.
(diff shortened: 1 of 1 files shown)


RaulGracia:
You are right @shrids, there is an issue for this #3725 cc/ @kevinhan88.

PR #3773, opened 2019-05-10
Issue 3772: (SegmentStore) Fixed DurableLogTests.testTruncateWithoutRecovery

andreipaduroiu:
**Change log description**  
- Fixed a unit test that was assuming that acks coming from the DurableLog were issued only after all internal memory structures are updated. 

**Purpose of the change**  
Fixes #3771.

**What the code does**  
- This unit test was assuming that acks from `DurableLog.add` were issued only after all internal memory structures were updated (the in-memory log, for example).
- However the DurableLog/OperationProcessor acks as soon as the operation is committed to Tier 1, and before it updates internal structures. As such, the in-memory log could be updated after a caller takes action on that ack.
- In this unit test, the action after the ack was to immediately truncate, which would have eliminated the operation from the log. However the operation was not there yet, so it wasn't removed. Instead, in some cases it would end up in the log just after the truncation executed, hence it would only have been removed with the next truncation.
- This is not a bug since the Tier 1 truncation is correct, and having an extra operation or two in memory for a bit is not a problem. They would eventually be cleaned up with a recovery or whenever the next truncation is issued.

**How to verify it**  
Unit test must pass.

### DIFF
--- a/segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
+++ b/segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
@@ -1077,10 +1077,7 @@ public class DurableLogTests extends OperationLogTestBase {
 
             List<OperationWithCompletion> completionFutures = processOperations(queuedOperations, durableLog);
             OperationWithCompletion.allOf(completionFutures).join();
-            TestUtils.await(() -> {
-                val allOps = readUpToSequenceNumber(durableLog, metadata.getOperationSequenceNumber());
-                return allOps.size() > 0 && allOps.get(allOps.size() - 1) == lastOp;
-            }, 10, TIMEOUT.toMillis());
+            awaitLastOperationAdded(durableLog, metadata);
 
             // Get a list of all the operations, before truncation.
             List<Operation> originalOperations = readUpToSequenceNumber(durableLog, metadata.getOperationSequenceNumber());
@@ -1100,6 +1097,7 @@ public class DurableLogTests extends OperationLogTestBase {
 
                     // Perform the truncation.
                     durableLog.truncate(currentOperation.getSequenceNumber(), TIMEOUT).join();
+                    awaitLastOperationAdded(durableLog, metadata);
                     if (!isTruncationPointFirstOperation) {
                         Assert.assertTrue("No truncation occurred even though a valid Truncation Point was passed: " + currentOperation.getSequenceNumber(), truncationOccurred.get());
                     }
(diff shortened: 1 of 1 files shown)


segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
```diff
@@ -1133,10 +1137,12 @@ public void testTruncateWithoutRecovery() {
                 // We were not able to do a full truncation before. Do one now, since we are guaranteed to have a new DataFrame available.
                 MetadataCheckpointOperation lastCheckpoint = new MetadataCheckpointOperation();
                 durableLog.add(lastCheckpoint, TIMEOUT).join();
+                awaitLastOperationAdded(durableLog, metadata);
```
fpj: The PR description says:

> it was added just after the truncation executed

but here we are waiting before the truncation. What am I missing?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
```diff
@@ -1423,6 +1429,26 @@ private void assertRecoveredOperationsMatch(List<Operation> expected, List<Opera
...
+     * {@link UpdateableContainerMetadata#getOperationSequenceNumber()} is present in the given {@link OperationLog}'s
+     * operations.
+     *
+     * This is helpful if we want to make sure that all the background async operations have completed before moving on
+     * to a next step. The {@link OperationProcessor} (inside {@link DurableLog}) acknowledges operations before adding
+     * them to the internal memory structures, so it is possible that we act on an ack before an operation that we need
+     * is present in the {@link DurableLog}.
+     */
+    @SneakyThrows(TimeoutException.class)
+    private void awaitLastOperationAdded(OperationLog durableLog, UpdateableContainerMetadata metadata) {
```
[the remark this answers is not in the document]
andreipaduroiu: Because the metadata always has the last added Operation Sequence Number. All this does is wait for that operation to appear in the log.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
andreipaduroiu: The pipeline works in these steps:
1. The "operation" is queued up.
2. It is processed by the OP.
3. The OP acks it (after persisting into Tier 1).
4. The OP adds it to in-memory structures.

We proceed with the test after step 3, and step 4 executes concurrently in the background. It's possible that we initiate and complete the truncation after 3 and before 4 is completed, thus leading to this problem.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
```diff
@@ -1071,10 +1072,12 @@ public void testTruncateWithoutRecovery() {
             // recovery, it wipes away all existing metadata).
             Set<Long> streamSegmentIds = createStreamSegmentsWithOperations(streamSegmentCount, durableLog);
             List<Operation> queuedOperations = generateOperations(streamSegmentIds, new HashMap<>(), appendsPerStreamSegment, METADATA_CHECKPOINT_EVERY, false, false);
-            queuedOperations.add(new MetadataCheckpointOperation()); // Add one of these at the end to ensure we can truncate everything.
+            val lastOp = new MetadataCheckpointOperation();
+            queuedOperations.add(lastOp); // Add one of these at the end to ensure we can truncate everything.
 
             List<OperationWithCompletion> completionFutures = processOperations(queuedOperations, durableLog);
             OperationWithCompletion.allOf(completionFutures).join();
+            awaitLastOperationAdded(durableLog, metadata);
```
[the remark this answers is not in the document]
andreipaduroiu: First one: we need to read all operations in the log, and we want to make sure all are accounted for.
Second one: we await the `StorageMetadataCheckpointOperation` that is added during a truncation because we then need to read it from the log.
Third one: we add a `MetadataCheckpointOperation` and then need to truncate on it. As such, we need that to be present in the log before truncating.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
fpj: That's not what I'm asking, let me add more detail to see if it clarifies. The PR description says:

> it was added just after the truncation executed

which is precisely what I see in lines 1099 and 1100. I see a different order in lines 1140 and 1141, and I'm wondering why the truncate in line 1141 isn't before the await in line 1140.

andreipaduroiu: Sorry, I had a confusing description. I reworded that sentence to make it clear that the operation would asynchronously end up in the log after we issue a truncation. We add it to the log **before**, then it is ack-ed, then we do a truncation, and due to some thread scheduling, the initial operation would only be added to the internal log after the truncation.

PR #3782, opened 2019-05-14
Issue 3781: VersionedSerializer support for Compacted Signed Longs

### LINKED ISSUE #3781: VersionedSerializer: Support for serializing compact signed Longs
**Problem description**
Need to add the following to `RevisionDataOutputStream`
- `writeCompactSignedLong`

Need to add the following to `RevisionDataInputStream`
- `readCompactSignedLong`

**Problem location**
VersionedSerializer.java

None.

andreipaduroiu:
**Change log description**  
- Added `RevisionDataOutput.writeCompactSignedLong()` and `RevisionDataInput.readCompactSignedLong()`.
- Improved Javadoc in `RevisionDataOutput` and `RevisionDataInput` by adding Javadoc links and fixing various references.

**Purpose of the change**  
Fixes #3781.

**What the code does**  
- `writeCompactLong`/`readCompactLong` only supported non-negative numbers up to 2^62-1. This change adds support for negative numbers as well.
- Negative numbers require an extra bit to store the sign, as such only numbers within the range [2^61, 2^61) can be encoded (3 bits are reserved for sign and length).

**How to verify it**  
New unit tests added.

### DIFF
--- a/common/src/main/java/io/pravega/common/io/serialization/RevisionDataOutput.java
+++ b/common/src/main/java/io/pravega/common/io/serialization/RevisionDataOutput.java
@@ -10,6 +10,7 @@
 package io.pravega.common.io.serialization;
 
 import io.pravega.common.util.ByteArraySegment;
+import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.io.OutputStream;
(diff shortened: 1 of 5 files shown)


common/src/main/java/io/pravega/common/io/serialization/RevisionDataOutput.java
```diff
@@ -23,26 +24,36 @@
...
 
     /**
-     * Minimum value that can be encoded using writeCompactLong().
+     * Minimum value that can be encoded using {@link #writeCompactLong}.
      */
     long COMPACT_LONG_MIN = 0L;
 
     /**
-     * Maximum value that can be encoded using writeCompactInt().
+     * Maximum value that can be encoded using {@link #writeCompactSignedLong}.
```
fpj: If the acceptable range is  `[-2^61, 2^61)`,  then shouldn't the maximum value that can be encoded be `0x1FFF_FFFF_FFFF_FFFFL - 1`?

It is just a matter of convention, but my natural reaction is to think that the maximum is inclusive as opposed to the maximum exclusive of the range. The options I suggest are:

1- Make the maximum `0x1FFF_FFFF_FFFF_FFFFL - 1`
2- Clarify in the comment that this is the top of the range, not inclusive (the range is open-ended).

andreipaduroiu: Good points. I adjusted the values (subtracted 1) and fixed the code to handle it.

PR #3804, opened 2019-05-20
Issue 3724: (SegmentStore) BugFix - Creating an already existing Segment in Tier 2

### LINKED ISSUE #3724: SegmentStore : Several NullPointerException in SegmentAggregator while running against ECS
**Problem description**
When running system tests against ECS, during the StorageWriter iteration several NullPointerException are observed (more than 8K)

Below is one stack trace
> 2019-04-29 14:58:02,446 51039 [Thread-2] ERROR i.p.s.server.writer.StorageWriter - StorageWriter[3]: Iteration[79].Error.
> java.lang.NullPointerException: null
> 	at io.pravega.segmentstore.storage.AsyncStorageWrapper.write(AsyncStorageWrapper.java:93)
> 	at io.pravega.segmentstore.server.writer.SegmentAggregator.lambda$flushPendingAppends$17(SegmentAggregator.java:775)
> 	at io.pravega.segmentstore.server.writer.SegmentAggregator.lambda$createSegmentIfNecessary$40(SegmentAggregator.java:1138)
> 	at java.util.concurrent.CompletableFuture.uniCompose(CompletableFuture.java:952)
> 	at java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:926)
> 	at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:442)
> 	at java.lang.Thread.run(Thread.java:748)

**Problem location**
The problem seems to be createSegmentIfNecessary in SegmentAggregator.java
The code is expecting the segment to be not present, but finds the segment present .
This results in null being passed to next lamda

https://github.com/pravega/pravega/blob/40f6fe09474b5b4f4a5e0bbeff54d798f7042516/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java#L1125

> return Futures
>                     .exceptionallyExpecting(
>                             this.storage.create(this.metadata.getName(), rollingPolicy, timeout),
>                             ex -> ex instanceof StreamSegmentExistsException,
>                             **null)** //====> passing null
>                     .thenComposeAsync(handle -> {
>                         **_this.handle.set(handle);_**   //===> The handle is set to null here. which ultimately throws
>                         return toRun.get();
>                     });
> 

- Handle null correctly.
- This seems to be symptom of a deeper problem of mismatch between what segment store thinks the state of the segment is vs what is actually in storage.

andreipaduroiu:
**Change log description**  
- Fixed a bug in `SegmentAggregator` where it was not able to properly recover from a Segment being created behind the scenes.
- Fixes a minor nuisance in `SegmentAttributeBTreeIndex` where `toString()` was not implemented, hence making log parsing more difficult.

**Purpose of the change**  
Fixes #3724.

**What the code does**  
- Fixes the following bug: https://github.com/pravega/pravega/issues/3724#issuecomment-494099244
- If the `SegmentAggregator` gets a `StreamSegmentExistsException` while trying to create a new segment in Tier 2 Storage, it calls `openWrite` to get a handle to it. This will allow subsequent operations to continue executing.
    - Note 1: If there is a fencing event, `openWrite` will throw the appropriate exception and the upstream code will handle it as it does if it arose from other places.
    - Note 2: If there is a mismatch between the data in Tier 2 and what the metadata indicates, the `SegmentAggregator` will detect that upon executing the next operation and enter its usual `ReconciliationMode`.

**How to verify it**  
New unit test.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
@@ -1177,7 +1177,9 @@ class SegmentAggregator implements WriterSegmentProcessor, AutoCloseable {
                                 // of them managed to create the Segment (and write something to it), but the other still assumed
                                 // the Segment did not exist - so we end up in here. We need to get a handle of the segment
                                 // and continue with whatever we were doing. If there is a mismatch (length, sealed, etc.),
-                                // then the normal reconciliation algorithm will kick in once it is discovered.
+                                // then the normal reconciliation algorithm will kick in once it is discovered and if the
+                                // segment has already been fenced out, openWrite() will throw the appropriate exception
+                                // which will be handled upstream.
                                 log.info("{}: Segment did not exist in Storage when initialized() was called, but does now.", this.traceObjectId);
                                 return this.storage.openWrite(this.metadata.getName());
                             })
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
@@ -1167,27 +1167,24 @@ class SegmentAggregator implements WriterSegmentProcessor, AutoCloseable {
             assert this.metadata.getStorageLength() == 0 : "no handle yet but metadata indicates Storage Segment not empty";
             long rolloverSize = this.metadata.getAttributes().getOrDefault(Attributes.ROLLOVER_SIZE, SegmentRollingPolicy.NO_ROLLING.getMaxLength());
             return Futures
-                    .exceptionallyExpecting(
+                    .exceptionallyComposeExpecting(
                             this.storage.create(this.metadata.getName(), new SegmentRollingPolicy(rolloverSize), timeout),
                             ex -> ex instanceof StreamSegmentExistsException,
-                            null)
+                            () -> {
+                                // This happens if we have more than one concurrent instances of the owning SegmentContainer
+                                // running at the same time. Both SegmentAggregator instances were initialized when the Segment
+                                // did not exist, and both knew about an append that would eventually make it to Storage. One
+                                // of them managed to create the Segment (and write something to it), but the other still assumed
+                                // the Segment did not exist - so we end up in here. We need to get a handle of the segment
+                                // and continue with whatever we were doing. If there is a mismatch (length, sealed, etc.),
+                                // then the normal reconciliation algorithm will kick in once it is discovered.
+                                log.info("{}: Segment did not exist in Storage when initialized() was called, but does now.", this.traceObjectId);
+                                return this.storage.openWrite(this.metadata.getName());
+                            })
                     .thenComposeAsync(handle -> {
-                        if (handle == null) {
-                            // This happens if we have more than one concurrent instances of the owning SegmentContainer
-                            // running at the same time. Both SegmentAggregator instances were initialized when the Segment
-                            // did not exist, and both knew about an append that would eventually make it to Storage. One
-                            // of them managed to create the Segment (and write something to it), but the other still assumed
-                            // the Segment did not exist - so we end up in here. We need to get a handle of the segment
-                            // and continue with whatever we were doing. If there is a mismatch (length, sealed, etc.),
-                            // then the normal reconciliation algorithm will kick in once it is discovered.
-                            log.info("{}: Segment did not exist in Storage when initialized() was called, but does now.", this.traceObjectId);
-                            return this.storage.openWrite(this.metadata.getName()).thenAccept(this.handle::set);
-                        } else {
-                            this.handle.set(handle);
-                            return CompletableFuture.completedFuture(null);
-                        }
-                    }, this.executor)
-                    .thenComposeAsync(v -> toRun.get(), this.executor);
+                        this.handle.set(handle);
+                        return toRun.get();
+                    }, this.executor);
 
         } else {
             // Segment already exists. Execute what we were supposed to.
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
@@ -1165,17 +1165,30 @@ class SegmentAggregator implements WriterSegmentProcessor, AutoCloseable {
         if (this.handle.get() == null) {
             // No handle so, the segment must not exist yet. Attempt to create it, then run what we wanted to.
             assert this.metadata.getStorageLength() == 0 : "no handle yet but metadata indicates Storage Segment not empty";
-            long rolloverSize = this.metadata.getAttributes().getOrDefault(Attributes.ROLLOVER_SIZE, -1L);
-            SegmentRollingPolicy rollingPolicy = rolloverSize < 0 ? SegmentRollingPolicy.NO_ROLLING : new SegmentRollingPolicy(rolloverSize);
+            long rolloverSize = this.metadata.getAttributes().getOrDefault(Attributes.ROLLOVER_SIZE, SegmentRollingPolicy.NO_ROLLING.getMaxLength());
             return Futures
                     .exceptionallyExpecting(
-                            this.storage.create(this.metadata.getName(), rollingPolicy, timeout),
+                            this.storage.create(this.metadata.getName(), new SegmentRollingPolicy(rolloverSize), timeout),
                             ex -> ex instanceof StreamSegmentExistsException,
                             null)
                     .thenComposeAsync(handle -> {
-                        this.handle.set(handle);
-                        return toRun.get();
-                    });
+                        if (handle == null) {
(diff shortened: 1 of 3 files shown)


segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
```diff
@@ -1633,6 +1633,51 @@ public void testSegmentMissingData() throws Exception {
 
     //region Unknown outcome operation reconciliation
 
+    /**
+     * Tests the ability of the SegmentAggregator to recover from situations when a Segment did not exist in Storage
+     * when {@link SegmentAggregator#initialize} was invoked, but exists when the first byte needs to be appended.
+     * This can happen when there are concurrent instances of the same Segment Container running at the same time and
+     * one of the managed to create the Segment in Storage after the other one was initialized; when the second one tries
```
fpj: Typo: `... one of them managed...`.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
```diff
@@ -1633,6 +1633,51 @@ public void testSegmentMissingData() throws Exception {
...
+     * to do the same, it must gracefully recover from that situation.
+     */
+    @Test
+    public void testReconcileCreateIfEmpty() throws Exception {
+        final WriterConfig config = DEFAULT_CONFIG;
+        final int appendCount = 1;
```
fpj: Where is this variable used?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
```diff
@@ -1165,17 +1165,27 @@ private void mergeCompleted(SegmentProperties segmentProperties, UpdateableSegme
...
+                            () -> {
+                                // This happens if we have more than one concurrent instances of the owning SegmentContainer
+                                // running at the same time. Both SegmentAggregator instances were initialized when the Segment
+                                // did not exist, and both knew about an append that would eventually make it to Storage. One
+                                // of them managed to create the Segment (and write something to it), but the other still assumed
+                                // the Segment did not exist - so we end up in here. We need to get a handle of the segment
+                                // and continue with whatever we were doing. If there is a mismatch (length, sealed, etc.),
+                                // then the normal reconciliation algorithm will kick in once it is discovered.
+                                log.info("{}: Segment did not exist in Storage when initialized() was called, but does now.", this.traceObjectId);
+                                return this.storage.openWrite(this.metadata.getName());
```
fpj: We are relying on the contract of `openWrite` to avoid having concurrent writes to the segment in tier 2.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
```diff
@@ -1633,6 +1633,51 @@ public void testSegmentMissingData() throws Exception {
...
+        TestContext context = new TestContext(config);
+
+        // Initialize the Segment Aggregator, but do not yet create the segment.
+        context.segmentAggregator.initialize(TIMEOUT).join();
+
+        // Write one operation.
+        @Cleanup
+        ByteArrayOutputStream writtenData = new ByteArrayOutputStream();
+        StorageOperation appendOp = generateAppendAndUpdateMetadata(0, SEGMENT_ID, context);
+        context.segmentAggregator.add(appendOp);
```
fpj: Is this going to create the segment in storage? otherwise, where do we do it? we need it so that we can validate the behavior from line 1666 onwards.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
andreipaduroiu: Fixed

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
andreipaduroiu: nowhere ... :)

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
andreipaduroiu: Yes. I updated the comment to reflect that.
All `Storage` operations that modify something can throw that, not just `openWrite`.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
andreipaduroiu: a few lines below, when we invoke `flush`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
```diff
@@ -1165,17 +1165,29 @@ private void mergeCompleted(SegmentProperties segmentProperties, UpdateableSegme
...
+            long rolloverSize = this.metadata.getAttributes().getOrDefault(Attributes.ROLLOVER_SIZE, SegmentRollingPolicy.NO_ROLLING.getMaxLength());
             return Futures
-                    .exceptionallyExpecting(
-                            this.storage.create(this.metadata.getName(), rollingPolicy, timeout),
+                    .exceptionallyComposeExpecting(
+                            this.storage.create(this.metadata.getName(), new SegmentRollingPolicy(rolloverSize), timeout),
                             ex -> ex instanceof StreamSegmentExistsException,
-                            null)
+                            () -> {
+                                // This happens if we have more than one concurrent instances of the owning SegmentContainer
```
sachin-j-joshi: This looks better.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
sachin-j-joshi: FileSystemStorage.openWrite only checks for existence. 
See. https://github.com/pravega/pravega/blob/master/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java#L214

fpj: What are you trying to imply @sachin-j-joshi ?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
fpj: I'm confused. From the comment leading this test declaration:

```
Tests the ability of the SegmentAggregator to recover from situations when a Segment did not exist in Storage
when {@link SegmentAggregator#initialize} was invoked, but exists when the first byte needs to be appended.
```

I actually see that we create the segment in storage in line 1661, so that when we invoke `flush` in line 1667, the segment already exists in storage. it is not created during the `flush` as you say, right?

andreipaduroiu: It is created during `flush`. Your comment is on `add`, which is when the operation is added to the SegmentAggregator. `flush` is invoked later.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
sachin-j-joshi: The FileSystemStorage does not implement fencing as of today. It can not detect not being a primary storage. Therefore we can not rely on this being detected in SyncStorage.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
fpj: I think I understand what the test case does, but the way you are explaining it is confusing me. Here is another attempt to explain how I'm reading it:

1- We initialize the segment aggregator without the segment existing in storage (line 1651) 
2- We create an append operation (lines 1655-1657)
3- We create the storage segment before flushing (line 1661)
4- We flush the append data to storage with the segment already created in step 3 (lines 1665-1668)

This sequence of steps is following the flow of comments, but it interprets that the segment is created in storage in line 1661, and not during the flush (line 1667) like you say above. If the segment is being created during the flush, then I don't think the test case is correct because the goal is to validate that it is able to overcome the situation in which the segment didn't exist in storage and now it does when append data.

Please ack or correct accordingly.

andreipaduroiu: No, the segment is created in 1661, when we invoke `context.storage.create`. It is also created during `flush`, in which case it will fail because it exists already and it needs to handle that gracefully. Had there be no code changes to `SegmentAggregator.createIfNecessary`, this test case would fail, hence it does exercise this scenario.

PR #3815, opened 2019-05-21
Issue 3798: Lower severity of fail to acquire lock exception

### LINKED ISSUE #3798: Soft Handle Exception: Observing Controller api failed error with "ERROR i.p.c.s.r.g.v.ControllerServiceImpl" and exception "concurrent.CompletionException: io.pravega.controller.store.task.LockFailedException" while creating IO stream
**Observing Controller api failed error with "ERROR i.p.c.s.r.g.v.ControllerServiceImpl" and exception "concurrent.CompletionException: io.pravega.controller.store.task.LockFailedException" while creating IO stream using Longevity**

Environment details: PKS / K8 with medium cluster:
```
3 master: xlarge: 4 CPU, 16 GB Ram, 32 GB Disk
5 worker: 2xlarge: 8 CPU, 32 GB Ram, 64 GB Disk
Tier-1 storage is from VSAN datastore
Tier-2 storage curved on NFS Client provisioner using Isilon as backend
```
```
Pravega version: 0.5.0-2251.e0e61de
Zookeeper Operator : pravega/zookeeper-operator:0.2.1
Pravega Operator: pravega/pravega-operator:0.3.2
```
**Controller log error:** 
[pravega-k8s-logging-2019-05-20-05-42-02.zip](https://github.com/pravega/pravega/files/3197343/pravega-k8s-logging-2019-05-20-05-42-02.zip)

```
2019-05-20 09:00:42,231 15123884 [grpc-default-executor-3] INFO  i.p.c.s.r.g.v.ControllerServiceImpl - [requestId=5869384100952760020] createStream called for stream longevity/midScale.
2019-05-20 09:00:42,304 15123957 [controllerpool-15] ERROR i.p.c.s.r.g.v.ControllerServiceImpl - Controller api failed with error:
java.util.concurrent.CompletionException: io.pravega.controller.store.task.LockFailedException: Failed locking resource longevity/midScale.
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
        at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
Caused by: io.pravega.controller.store.task.LockFailedException: Failed locking resource longevity/midScale.
        at io.pravega.controller.store.task.ZKTaskMetadataStore.acquireLock(ZKTaskMetadataStore.java:56)
        at io.pravega.controller.store.task.AbstractTaskMetadataStore.lambda$lock$0(AbstractTaskMetadataStore.java:56)
        at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1590)
        ... 7 common frames omitted
Caused by: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /pravega/pravega/taskIndex/longevity_%_midScale
        at org.apache.zookeeper.KeeperException.create(KeeperException.java:123)
        at org.apache.zookeeper.KeeperException.create(KeeperException.java:51)
        at org.apache.zookeeper.ZooKeeper.create(ZooKeeper.java:1525)
        at org.apache.curator.framework.imps.CreateBuilderImpl$17.call(CreateBuilderImpl.java:1181)
        at org.apache.curator.framework.imps.CreateBuilderImpl$17.call(CreateBuilderImpl.java:1158)
        at org.apache.curator.connection.StandardConnectionHandlingPolicy.callWithRetry(StandardConnectionHandlingPolicy.java:64)
        at org.apache.curator.RetryLoop.callWithRetry(RetryLoop.java:100)
        at org.apache.curator.framework.imps.CreateBuilderImpl.pathInForeground(CreateBuilderImpl.java:1155)
        at org.apache.curator.framework.imps.CreateBuilderImpl.protectedPathInForeground(CreateBuilderImpl.java:605)
        at org.apache.curator.framework.imps.CreateBuilderImpl.forPath(CreateBuilderImpl.java:595)
        at org.apache.curator.framework.imps.CreateBuilderImpl.forPath(CreateBuilderImpl.java:49)
        at io.pravega.controller.store.task.ZKTaskMetadataStore.acquireLock(ZKTaskMetadataStore.java:53)
        ... 9 common frames omitted
```

fpj:

**Change log description**  
* Lowers the severity of log messages when failing to acquire lock in the controller metadata store.
* Prints only the exception message, but not the stack trace to avoid polluting the log.

**Purpose of the change**  
Fixes #3798 

**What the code does**  
The exception indicating a failure to acquire a lock in the controller metadata store does not necessarily indicate a problem with the controller state, it might correspond to the situation in which multiple calls have been made, e.g., to create a stream. As such, the message does not indicate an unrecoverable error and should not be printed as such. We also remove the stack trace from the output.

The main code change is in the `ControllerServiceImpl`, and it consists of creating a private method to handle log messages based on the exception. Currently, we only print differently in the case that the exception is `LockFailedException`. Otherwise, it prints the same information as before, at the same severity level, and including the stack trace.

To test that `createStream` throws `LockFailedException` as expected, it adds some auxiliary code for tests, including three new classes:

- `TaskStoreFactoryForTests`: A factory for the following two other classes.
- `ZKTaskMetadataStoreForTests`: extends `ZKTaskMetadataStore` and implements `TaskMetadataStoreExtensions`.
- `InMemoryTaskMetadataStoreForTests`: extends `InMemoryTaskMetadataStore` and implements `TaskMetadataStoreExtensions`.
  
`TaskMetadataStoreExtensions` extends the functionality of `TaskMetadataStore` in the main code for testing. Currently, it only exposes the ability of blocking a critical to be able to test concurrent scenarios.

The additional code for testing may seem overkill for the change, but there was no good way I found to validate that the code was actually throwing what it is supposed to be throwing. Assuming it has merit, it validates the behavior of `createStream` in the presence of concurrent access. It could also be done for other operations, although out of scope for this PR.

Finally, it adds a couple of cases to `isSegmentValidTest` for test coverage.

**How to verify it**  
Build must succeed.


controller/src/test/java/io/pravega/controller/server/v1/ZKControllerServiceImplTest.java
```diff
@@ -76,7 +78,7 @@
     @Override
     public void setup() throws Exception {
         final HostControllerStore hostStore;
-        final TaskMetadataStore taskMetadataStore;
+        //final TaskMetadataStore taskMetadataStore;
```
shiveshr: nit: remove this

controller/src/test/java/io/pravega/controller/store/task/TaskStoreFactoryForTests.java
```diff
@@ -0,0 +1,141 @@
...
+        @Override
+        public CompletableFuture<Void> lock(final Resource resource,
+                                            final TaskData taskData,
+                                            final String owner,
+                                            final String tag,
+                                            final String oldOwner,
+                                            final String oldTag) {
+            CompletableFuture<Void> future = super.lock(resource, taskData, owner, tag, oldOwner, oldTag);
+            try {
+                this.latch.get().get();
```
shiveshr: you can do a `join` on the future so that you dont have to surround it with a try catch block

controller/src/test/java/io/pravega/controller/store/task/TaskStoreFactoryForTests.java
```diff
@@ -0,0 +1,141 @@
...
```
shiveshr: same here.  we could use future.join

controller/src/test/java/io/pravega/controller/store/task/TaskStoreFactoryForTests.java
```diff
@@ -0,0 +1,141 @@
...
+         */
+        void unblockCriticalSection();
+    }
+
+    /**
+     * Extends {@link ZKTaskMetadataStore} with the calls in {@link TaskMetadataStoreExtensions}.
+     */
+    public static class ZKTaskMetadataStoreForTests extends ZKTaskMetadataStore implements TaskMetadataStoreExtensions {
+
+        private final AtomicReference<CompletableFuture<Void>> latch = new AtomicReference<>(new CompletableFuture<Void>());
```
shiveshr: nit: the referenced future is completed in the constructor and not serving any purpose. We can simply do a null future here.

controller/src/test/java/io/pravega/controller/server/v1/ZKControllerServiceImplTest.java
fpj: fixed

controller/src/test/java/io/pravega/controller/store/task/TaskStoreFactoryForTests.java
fpj: fixed

controller/src/test/java/io/pravega/controller/store/task/TaskStoreFactoryForTests.java
fpj: I did it this way so that we can have a complete future that does not hold the processing of the lock in the case the test is not blocking it, which is the case for most tests as of now. If I set it to `null` during initialization, then I'd need to verify whether the value is `null` every time. This latter approach works too, but I felt it performs slightly worse.

fpj:
This test case is failing:

```
createTransactionFailureTest
```

and might be related, investigating.

PR #3818, opened 2019-05-22
Issue 3816: Sporadic failure of BookieFailoverTest in Docker Swarm system tests

### LINKED ISSUE #3816: Sporadic failure of BookieFailoverTest in Docker Swarm system tests
**Problem description**
We are observing sporadic failures of `BookieFailoverTest` in Docker Swarm system tests. The problem appears sometimes when the test scales down the number of Bookies from 3 to 2, as the test is not able to get back the number of Bookies to 3. This makes almost all subsequent system tests to fail as well.

**Problem location**
`BookieFailoverTest` (Docker Swarm).

Fix the test.

RaulGracia:
**Change log description**  
Fixes a sporadic problem in BookieFailoverTest (Docker Swarm system tests) that prevented the Bookkeeper service to scale up from 2 to 3 replicas.

**Purpose of the change**  
Fixes #3816.

**What the code does**  
This PR contains 2 changes that improve the situation of the Bookkeeper service in Docker Swarm:
- _Increase retries in the Bookkeeper restart policy_: Before, the `RestartPolicy` of Bookkeeper tasks was set to `maxAttempts=1`. However, in an unfortunate case, a Bookie may fail more than once to get started. We increased restart retries to `3`, which seems to be enough in most cases.
- _Remove unused volume for Bookies_: `BookkeeperDockerService` was mounting a volume referring to a non-existing location in the Bookkeeper image (`/opt/dl_all/distributedlog-service/logs/`). We just removed that useless volume.

Apart from that, we also upgraded the Zookeeper image used in Docker Swarm system tests (`jplock/zookeeper:3.5.1-alpha`) to the same one used in Kubernetes tests (`zookeeper:3.5.4-beta`). 

**How to verify it**  
We have seen 4 successful `BookieFailoverTest` executions in a row with these changes (including a green build).

### DIFF
--- a/test/system/src/main/java/io/pravega/test/system/framework/services/docker/ZookeeperDockerService.java
+++ b/test/system/src/main/java/io/pravega/test/system/framework/services/docker/ZookeeperDockerService.java
@@ -28,7 +28,7 @@ import static io.pravega.test.system.framework.Utils.DOCKER_NETWORK;
 @Slf4j
 public class ZookeeperDockerService extends DockerBasedService {
 
-    private static final String ZK_IMAGE = "jplock/zookeeper:3.5.1-alpha";
+    private static final String ZK_IMAGE = "zookeeper:3.5.4-beta";
     private final long instances = 1;
     private final double cpu = 1.0 * Math.pow(10.0, 9.0);
     private final long mem = 1024 * 1024 * 1024L;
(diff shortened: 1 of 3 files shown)


test/system/src/main/java/io/pravega/test/system/framework/services/docker/ZookeeperDockerService.java
```diff
@@ -28,7 +28,7 @@
 @Slf4j
 public class ZookeeperDockerService extends DockerBasedService {
 
-    private static final String ZK_IMAGE = "jplock/zookeeper:3.5.1-alpha";
+    private static final String ZK_IMAGE = "zookeeper:3.5.4-beta";
```
shrids: Now that we are using the official zookeeper docker image we can also configure the zk logs to stored similar to other services using `ZOO_LOG4J_PROP="INFO,ROLLINGFILE" ` and mounting /logs folder.
ref: https://hub.docker.com/_/zookeeper

PR #3819, opened 2019-05-22
Issue 3786: improved the gauge handling to prevent gauge object loss

### LINKED ISSUE #3786: Sporadic failure in MetricsProviderTest.testGauge
**Problem description**
Found a sporadic assertion error in `MetricsProviderTest.testGauge`:
```
Error Message
java.lang.AssertionError: expected:<2> but was:<0>
Stacktrace
java.lang.AssertionError: expected:<2> but was:<0>
	at org.junit.Assert.assertEquals(Assert.java:631)
	at io.pravega.shared.metrics.MetricsProviderTest.testGauge(MetricsProviderTest.java:193)
```

**Problem location**
`MetricsProviderTest.testGauge`

Fix the issue.

kevinhan88:
**Change log description**  
Gauge handling is different than other metrics type because it passes a function into metrics registry. Because the function is immutable, in order to update gauge, the legacy code just delete existing gauge and insert a new one.
Hence a race condition is introduced - if the gauge value is retrieved after the old object is deleted but before the new object is registered, the gauge will report 0 or NaN.
This change is to add an AtomicReference into Gauge which points to a function. To update Gauge value, simply replace the function inside the reference.

**Purpose of the change**  
Fixes #3786

**What the code does**  
Added an AtomicReference into Gauge object to hold a reference to a Supplier<Number>.
Changed the Gauge interface to expose the reference.
Changes test codes to adopt the new interface.
When to update gauge, instead of removing old object, now just replace the function(Supplier<Number>).
Since Gauge object can now be "reused", put Gauge into cache.

**How to verify it**  
All the tests should pass.

### DIFF
--- a/shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
+++ b/shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
@@ -136,18 +136,17 @@ public class DynamicLoggerImpl implements DynamicLogger {
     public <T extends Number> void reportGaugeValue(String name, T value, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
         Preconditions.checkNotNull(value);
-        MetricsNames.MetricKey keys = metricKey(name, tags);
-
-        Gauge gauge = gaugesCache.getIfPresent(keys.getCacheKey());
-        if (gauge == null) {
-            Gauge newGauge = underlying.registerGauge(keys.getRegistryKey(), value::doubleValue, tags);
-            if (newGauge == null) {
-                log.error("Unsupported Number type: {}.", value.getClass().getName());
-            } else {
-                gaugesCache.put(keys.getCacheKey(), newGauge);
-            }
-        } else {
-            gauge.supplierReference().set(value::doubleValue);
+        try {
+            MetricsNames.MetricKey keys = metricKey(name, tags);
+            Gauge gauge = gaugesCache.get(keys.getCacheKey(), new Callable<Gauge>() {
+                @Override
+                public Gauge call() throws Exception {
+                    return underlying.registerGauge(keys.getRegistryKey(), value::doubleValue, tags);
+                }
+            });
+            gauge.setSupplier(value::doubleValue);
+        } catch (ExecutionException e) {
+            log.error("Error accessing gauge through gaugesCache", e);
         }
     }
 
--- a/shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
+++ b/shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
@@ -136,27 +136,18 @@ public class DynamicLoggerImpl implements DynamicLogger {
     public <T extends Number> void reportGaugeValue(String name, T value, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
         Preconditions.checkNotNull(value);
-        Gauge newGauge = null;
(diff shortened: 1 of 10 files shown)


shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -136,27 +136,18 @@ public void freezeCounter(String name, String... tags) {
...
-            newGauge = underlying.registerGauge(keys.getRegistryKey(), value::shortValue, tags);
-        } else if (value instanceof Integer) {
-            newGauge = underlying.registerGauge(keys.getRegistryKey(), value::intValue, tags);
-        } else if (value instanceof Long) {
-            newGauge = underlying.registerGauge(keys.getRegistryKey(), value::longValue);
-        }
-
-        if (null == newGauge) {
-            log.error("Unsupported Number type: {}.", value.getClass().getName());
+        Gauge gauge = gaugesCache.getIfPresent(keys.getCacheKey());
```
andreipaduroiu: This code is not thread safe. You could be instantiating multiple gauges with the same key, but only keep track of one of them.

shared/metrics/src/main/java/io/pravega/shared/metrics/Gauge.java
```diff
@@ -9,8 +9,17 @@
...
 /**
  * Defines a Gauge, which will wrap a gauge instance and its name.
  */
 public interface Gauge extends Metric {
+
+    /**
+     * Return an AtomicReference pointing to the supplier of gauge value.
+     * @return an AtomicReference pointing to the Supplier of gauge value.
+     */
+    AtomicReference<Supplier<Number>> supplierReference();
```
andreipaduroiu: This breaks encapsulation. You allow external actors to modify this object's internals. Consider making two separate methods (`getValueSupplier` and `setValueSupplier`).

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: Fixed by supplying loader with get to make it atomic.

shared/metrics/src/main/java/io/pravega/shared/metrics/Gauge.java
kevinhan88: Good point, fixed.

shared/metrics/src/main/java/io/pravega/shared/metrics/StatsLoggerImpl.java
```diff
@@ -54,9 +55,9 @@ public Counter createCounter(String statName, String... tags) {
     }
 
     @Override
-    public <T extends Number> Gauge registerGauge(final String statName, Supplier<T> valueSupplier, String... tags) {
+    public Gauge registerGauge(final String statName, final Supplier<Number> valueSupplier, String... tags) {
         try {
-            return new GaugeImpl<>(statName, valueSupplier, tags);
+            return new GaugeImpl<>(statName, Preconditions.checkNotNull(valueSupplier), tags);
```
andreipaduroiu: `Preconditions.checkNotNull`'s exception will be swallowed up and you'll always return NULLGAUGE. Is this intentional? I think it would make more sense to throw the NullPtrEx to the caller instead of this.

kevinhan88: Yes it's by design(to have NULLMETRICS for Noop. 
We may have a separate issue for the improvement though.

Issue #3148, opened 2018-11-29
Segment Store: Lazy Segment Create and Delete in Tier2


andreipaduroiu:
**Problem description**
See #3075 for context. This is Part 1.

**Problem location**
StreamSegmentContainer.

**Suggestions for an improvement**
We need to:
- Define `MetadataStore` with these APIs
    - createSegment: creates a new Segment. This will not be directly created (as before), rather it will be an entry in the Metadata Store. Additional components may need to be updated to support this.
    - deleteSegment: deletes an existing Segment.
    - getSegmentInfo: gets Segment information for an existing segment.
    - getOrAssignId: gets the Segment Id or assigns a new one for an existing segment.
    - updateSegmentInfo: updates Segment information for an existing segment.
- Refactor `StreamSegmentMapper` and `SegmentStateStore` into `StorageMetadataStore` (name TBD)
- Update `ReadIndex` to not issue Tier 2 reads for empty segments (Metadata.getStorageLength() == 0). This may already be implemented, but we need to check.
- Update `StorageWriter` to create the segment upon the first write if not already created. Update all other operations to handle such situations, as well as reconciliation.
- Add a `DeleteSegmentOperation` that, when processed by the Writer, will delete the segment. This will also update the Metadata as well.

PR #3834, opened 2019-05-24
Issue 3832: (SegmentStore) Lazily creating the Attribute Index only on the first write.

### LINKED ISSUE #3832: Segment Attribute Index should lazily create Tier 2 Segment
**Problem description**
`ContainerAttributeIndex.forSegment` calls `SegmentAttributeBTreeIndex.initialize()` which automatically creates the Tier 2 segment file upon the first request. This includes read-only access, which can create random Tier 2 access patterns and slow down incoming requests.

**Problem location**
`SegmentAttributeBTreeIndex.java`.

Lazily create the attribute segment upon the first write.

andreipaduroiu:
**Change log description**  
- The Segment Attribute Index is now created only when writing to it for the first time.

**Purpose of the change**  
Fixes #3832.

**What the code does**  
- #3148 ensured that all segments are created in Tier 2 only when we transfer the first byte to them, however that did not address Attribute Segments (which reside only in Tier 2 and cannot be accessed from the outside). Furthermore, these were created even if we did not intend to write anything to them (when we want to get the value of an attribute, for example), which significantly increased the latency of certain external operations that relied on attributes (file creates (some cases) and setting up appends for new segments).
- The `SegmentAttributeBTreeIndex` has been modified in a similar fashion to `SegmentAggregator` to only create the sub-segment in Tier 2 when we execute the first write. Any reads or getInfo's will treat inexistent segments as a zero-length segment, and sealing an inexistent attribute segment will have no effect (it will not create it just for the sake of sealing an empty file).

**How to verify it**  
New unit tests added, existing unit test(s) updated.
All unit tests and system tests must pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
@@ -26,6 +26,7 @@ import io.pravega.common.util.btree.PageEntry;
 import io.pravega.segmentstore.contracts.Attributes;
 import io.pravega.segmentstore.contracts.BadOffsetException;
 import io.pravega.segmentstore.contracts.SegmentProperties;
+import io.pravega.segmentstore.contracts.StreamSegmentExistsException;
 import io.pravega.segmentstore.contracts.StreamSegmentNotExistsException;
 import io.pravega.segmentstore.contracts.StreamSegmentSealedException;
 import io.pravega.segmentstore.contracts.StreamSegmentTruncatedException;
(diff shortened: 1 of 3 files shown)


andreipaduroiu:
System test 188 passed.

fpj:
@andreipaduroiu could you check the code coverage report, I think it is complaining mostly because of some corner or exceptional cases not being handled:

https://codecov.io/gh/pravega/pravega/compare/8583d93dc047952a91fbaa0938bc6b110b0fe004...0c445f76a8b9b70d901318700eaa70228261f138/diff

andreipaduroiu:
@fpj the build passed. The test that used to fail had nothing to do with this change.

fpj:
What about the code coverage, do you want to do anything about it?

andreipaduroiu:
Very few non-test, non-comment LOCs were modified in this change. All the complaints from the code coverage tool are for fail-safe code that is is never expected to be executed (so I can't simulate that scenario) but is there simply for completeness.

PR #3839, opened 2019-05-27
Issue 3671: Watermarking workflow

### LINKED ISSUE #3671: Watermark: Controller: Implement background periodic watermark workflow
**Problem description**
Implement Watermark workflow on controller using the bucket infrastructure for periodic watermark computation and emitting. 
This should be implemented as a follow up to #3670  and make use of metadata recorded for each writer for the stream and compute and emit watermarks in a dedicated revisioned segment. 

**Problem location**
Controller Bucket Service

Implement watermark periodic workflow.

shiveshr:
**Change log description**  
This change set implements watermarking workflow on controller. 
Watermarking computation is modeled as a periodic background task. 
Writers can post their marks asynchronously to any controller instance. These marks are stored in metadata store. 
A periodic workflow for the stream is executed which collects all the marks reported since last watermark was emitted and computes a new watermark. 

**Purpose of the change**  
Fixes #3671 

**What the code does**  
A new class called `PeriodicWatermarking` is added which implements the periodic watermarking workflow. 
The periodic background workflow infrastructure that was built for retention workflows is reused for Watermarking service. Watermarking workflow will run on its dedicated executor thread pool whereby not interfere with other threads used by controller service. The watermarking frequency is configurable. 
The streams are divided into buckets and these buckets are distributed across different controller instances to balance the load. 
Note: There is one limitation of periodic workflow infrastructure that we will improve separately - the infrastructure does not support `rebalance` of buckets in case new controller instance is added to the cluster. 
Each controller instance runs periodic workflows for all streams belonging to buckets owned by that controller. Streams, for which writers have reported marks, are added to the watermarking bucket store. 
The periodic workflow infrastructure runs periodic jobs for these streams belonging to buckets that it owns and calls `PeriodicWatermarking.execute` method for each of these streams at the configured frequency.   

Watermarking makes uses of revisionedStreamClient to emit watermarks into watermark segment. 
It maintains a window of previously emitted watermarks by leveraging the "mark" primitive on RevisionedStreamClient. In each iteration it reads all previously emitted marks in this window. 
The workflow has a concept of `active` and `participating` writers for the watermark computation. 
The controller waits for all active writers to report their marks before performing watermarking computation. Participating writers are those that have reported their marks since the last watermark. 
So controller waits until all active writers either either participate or are deemed inactive. 

For maintaining the active writer, Controller maintains the aforesaid window. It compares writer marks with the `window.floor` watermark. Writers that have emitted marks higher than window.floor are considered as active. In each iteration, the watermarking workflow either emits a new watermark or shrinks the "window" by moving the floor to the next watermark. 

When controller has all active writers as participating writers, it performs the watermarking computation workflow where it computes the ` GLB (timestamp of all participating writer marks + last watermark)` and computes a well formed stream cut using `logical upper bound on positions of writer marks + last watermark`. 
The details of this algorithm are described in watermarking PDP. 

This PR also defines the function to compute the name of watermark segment where watermarks are emitted. 
It also defines a serializable Watermark object and a wrapper serializer that implements the client.Serializer interface. 

**How to verify it**  
Unit tests and integration test added.


shared/protocol/src/main/java/io/pravega/shared/watermarks/Watermark.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.DataOutput;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+@Data
+public class Watermark {
+    public static final WatermarkSerializer SERIALIZER = new WatermarkSerializer();
+    public static final Watermark EMPTY = new Watermark(Long.MIN_VALUE, ImmutableMap.of());
+    private final long timestamp;
+    private final ImmutableMap<Long, Long> streamCut;
```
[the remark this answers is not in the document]
shiveshr: Done

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
```diff
@@ -71,6 +74,12 @@
      */
     private static final String TABLES = "_tables";
 
+    /**
+     * Prefix for identifying system created mark segments for storing watermarks. 
+     */
+    @Getter(AccessLevel.PACKAGE)
+    private static final String MARK_PREFIX = "_MARK";
```
tkaitchuck: This looks like in belongs in `NameUtils`. Should we merge these two utility classes?

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
```diff
@@ -366,4 +375,12 @@ public static boolean isTableSegment(String qualifiedName) {
         return tokens[1].equals(TABLES);
     }
     // endregion
+    
+    // region watermark
+    public static String getMarkSegmentForStream(String stream) {
```
tkaitchuck: We should probably call this `getMarkStreamForStream` because there still will be a segment number associated with the actual segment.

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
```diff
@@ -366,4 +375,12 @@ public static boolean isTableSegment(String qualifiedName) {
...
+        StringBuffer sb = new StringBuffer();
+        sb.append(MARK_PREFIX);
+        sb.append(stream);
```
tkaitchuck: We should have a precondition that `stream` does not start with the mark_prefix already as this makes little sense. (It probably also doesn't make sense for reader groups)

shiveshr: That check should be in stream manager interface perhaps. Because controller is used to create both internal and external streams and it wouldnt know where the request is originating from.

PR #3844, opened 2019-05-28
Issue 3785: Fixes bug inside AssertExtensions

### LINKED ISSUE #3785: Sporadic failure in MetricsTest.metricsTimeBasedCacheEvictionTest
**Problem description**
Another problem has manifested on this test after solving a `NullPointerException` in #3725. In this case, the problem is a non-satisfied assertion:
```
Error Message
java.lang.AssertionError: expected:<130> but was:<-1>
Stacktrace
java.lang.AssertionError: expected:<130> but was:<-1>
	at io.pravega.test.common.AssertExtensions.assertEventuallyEquals(AssertExtensions.java:53)
	at io.pravega.test.integration.MetricsTest.metricsTimeBasedCacheEvictionTest(MetricsTest.java:253)
```
[Full log of this failure.](https://github.com/pravega/pravega/files/3188553/log.txt)

**Problem location**
MetricsTest.metricsTimeBasedCacheEvictionTest

Fix the test.

kevinhan88:
**Change log description**  
There is a bug inside AssertExtension::assertEventuallyEquals() - when compare two incoming objects, it uses == (address comparison) which fails in most cases unless the two objects are the same. Then it keeps trying until timed out and eventually hands it over to org.junit.Assert.assertEquals which correctly uses objA.equals(objB). 
The effect of this bug is that it will hold things up to timeout and then call assertEquals.

**Purpose of the change**  
Fixes #3785 

**What the code does**  
This PR fixes a bug in `AssertExtensions`, by testing null equals and then call equals of object, instead of using address comparison.

With the fix above, the test inside `MetricsTest` still failed because there's no more holding so the cache eviction didn't get chance to run; Thread.sleep is added for cache eviction to happen.

**How to verify it**  
All tests should pass.

### DIFF
--- a/test/integration/src/test/java/io/pravega/test/integration/MetricsTest.java
+++ b/test/integration/src/test/java/io/pravega/test/integration/MetricsTest.java
@@ -41,7 +41,6 @@ import io.pravega.shared.metrics.MetricRegistryUtils;
 import io.pravega.shared.metrics.MetricsConfig;
 import io.pravega.shared.metrics.MetricsProvider;
 import io.pravega.shared.metrics.StatsProvider;
-import io.pravega.test.common.AssertExtensions;
 import io.pravega.test.common.TestUtils;
 import io.pravega.test.common.TestingServerStarter;
 import io.pravega.test.common.ThreadPooledTestSuite;
@@ -62,6 +61,7 @@ import org.junit.Test;
 import static io.pravega.shared.MetricsNames.SEGMENT_READ_BYTES;
 import static io.pravega.shared.MetricsTags.segmentTags;
 import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 @Slf4j
@@ -100,7 +100,7 @@ public class MetricsTest extends ThreadPooledTestSuite {
         MetricsConfig metricsConfig = MetricsConfig.builder()
                 .with(MetricsConfig.ENABLE_STATSD_REPORTER, false)
                 .build();
-        metricsConfig.setDynamicCacheEvictionDuration(Duration.ofSeconds(3));
+        metricsConfig.setDynamicCacheEvictionDuration(Duration.ofSeconds(2));
 
         MetricsProvider.initialize(metricsConfig);
         statsProvider = MetricsProvider.getMetricsProvider();
@@ -203,13 +203,10 @@ public class MetricsTest extends ThreadPooledTestSuite {
             readAllEvents(reader1);
 
             final String[] streamTags = segmentTags(scope + "/" + STREAM_NAME + "/0.#epoch.0");
-            AssertExtensions.assertEventuallyEquals(bytesWritten, () -> {
-                return MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags) == null
-                        ? -1 : (long) MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags).count();
-            }, 2000);
+            assertEquals(bytesWritten, (long) MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags).count());
 
             //Wait for cache eviction to happen
-            Thread.sleep(5000);
+            Thread.sleep(4000);
 
             String readerGroupName2 = readerGroupName + "2";
             log.info("Creating Reader group : {}", readerGroupName2);
@@ -229,10 +226,7 @@ public class MetricsTest extends ThreadPooledTestSuite {
 
             //Metric is evicted from Cache, after cache eviction duration
             //Count starts from 0, rather than adding up to previously ready bytes, as cache is evicted.
-            AssertExtensions.assertEventuallyEquals(bytesWritten, () -> {
-                return MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags) == null
-                        ? -1 : (long) MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags).count();
-            }, 1000);
+            assertEquals(bytesWritten, (long) MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags).count());
 
             Map<Double, Double> map = new HashMap<>();
             map.put(0.0, 1.0);
@@ -253,10 +247,7 @@ public class MetricsTest extends ThreadPooledTestSuite {
             readAllEvents(reader1);
 
             final String[] streamTags2nd = segmentTags(scope + "/" + STREAM_NAME + "/1.#epoch.1");
-            AssertExtensions.assertEventuallyEquals(bytesWritten, () -> {
-                return MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags2nd) == null
-                        ? -1 : (long) MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags2nd).count();
-            }, 1000);
+            assertEquals(bytesWritten, (long) MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags2nd).count());
 
             readerGroupManager.deleteReaderGroup(readerGroupName1);
             readerGroupManager.deleteReaderGroup(readerGroupName2);
--- a/test/integration/src/test/java/io/pravega/test/integration/MetricsTest.java
+++ b/test/integration/src/test/java/io/pravega/test/integration/MetricsTest.java
@@ -9,7 +9,6 @@
  */
 package io.pravega.test.integration;
 
-import io.micrometer.core.instrument.Metrics;
 import io.pravega.client.ClientConfig;
 import io.pravega.client.admin.ReaderGroupManager;
 import io.pravega.client.admin.StreamManager;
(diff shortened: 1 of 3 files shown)


test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
```diff
@@ -45,8 +45,14 @@
     public static <T> void assertEventuallyEquals(T expected, Callable<T> eval, long timeout) throws Exception {
         long endTime = System.currentTimeMillis() + timeout;
         while (endTime > System.currentTimeMillis()) {
-            if (expected == eval.call()) {
-                return;
+            if (expected == null) {
+                if (eval.call() == null) {
+                    return;
+                }
+            } else {
```
andreipaduroiu: you can merge this into an `else if` block

test/integration/src/test/java/io/pravega/test/integration/MetricsTest.java
```diff
@@ -206,7 +206,10 @@ public void metricsTimeBasedCacheEvictionTest() throws Exception {
             AssertExtensions.assertEventuallyEquals(bytesWritten, () -> {
                 return MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags) == null
                         ? -1 : (long) MetricRegistryUtils.getCounter(SEGMENT_READ_BYTES, streamTags).count();
-            }, 10000);
+            }, 2000);
+
+            //Wait for cache eviction to happen
+            Thread.sleep(5000);
```
andreipaduroiu: Is there really no other way of simulating the passage of time? Can we not have an internal method that forces a cache eviction?

kevinhan88: No easy way at least; and this is integration test so let's do it "naturally".

test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
fpj: This comment hasn't been addressed, I have unresolved this conversation.

test/integration/src/test/java/io/pravega/test/integration/MetricsTest.java
fpj: Relying upon sleeps in tests is bad practice for at least two reasons:

- It increases the total length of the build often unnecessarily
- Due to the non-determinism of process and thread scheduling, it leads to false positives. 

Please find another way to determine that the element has been evicted from the cache.

test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
kevinhan88: it's fixed in commit 053715b493b

test/integration/src/test/java/io/pravega/test/integration/MetricsTest.java
kevinhan88: Reduced the eviction time to 2 seconds, and reduced the sleep time to 4 seconds. The total test execution time is dropped from 54 sec to 24 sec.
Note this is a "time based" integration test trying to ensure metric cache indeed evicts as expected from an end-to-end integration perspective. The unit tests of Cache evictions are already covered by SegmentStatsRecorderTest.java, which uses a TestContext as the run time, and supplies a mocked cache.
If we manipulate the cache too much in this integration test, then it's overlapping the unit test SegmentStatsRecorderTest; more importantly, it's losing the point of integration test.

kevinhan88: Just to add one more comment - I spent quite some time trying to find a way to manipulate the cache eviction so we can have more precise control of the test execution. Firstly it is near impossible to manipulate internal components of Pravega in an integration test environment; secondly when I tried it further, this test became more and more like SegmentStatsRecorderTest, a pure unit test, and starts to lose the point of integration test.

test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
```diff
@@ -45,7 +45,11 @@
     public static <T> void assertEventuallyEquals(T expected, Callable<T> eval, long timeout) throws Exception {
         long endTime = System.currentTimeMillis() + timeout;
         while (endTime > System.currentTimeMillis()) {
-            if (expected == eval.call()) {
+            if (expected == null) {
+                if (eval.call() == null) {
+                    return;
+                }
+            } else if (expected.equals(eval.call())) {
```
fpj: Is it ok if `eval.call()` is `null` here?

test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
```diff
@@ -45,7 +45,11 @@
     public static <T> void assertEventuallyEquals(T expected, Callable<T> eval, long timeout) throws Exception {
         long endTime = System.currentTimeMillis() + timeout;
         while (endTime > System.currentTimeMillis()) {
-            if (expected == eval.call()) {
+            if (expected == null) {
```
fpj: I think the logic you are trying to express here is more clearly stated as:

```
if ((expected == null && eval.call() == null) || (expected != null && expected.equals(eval.call())) {
    return;
}
```

andreipaduroiu: After looking at this method a few more times, I feel like we could move `testcommon.TestUtils.await(...)` in here and replace this method altogether. `await` is much more flexible and gives the caller the ability to implement their own condition check and fine-tune the check frequency and delay.

https://github.com/pravega/pravega/blob/master/test/testcommon/src/main/java/io/pravega/test/common/TestUtils.java#L68

test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
kevinhan88: Yes, eval.call() could return null.

test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
kevinhan88: Yes, fixed with the compact condition in the AssertExtensions.java.

kevinhan88: I added awaitFirst method into TestUtils.java and started using it inside MetricsTest.
The reason is that await() test the condition and then await, however this is cache eviction test, if the object is accessed, it changes the expiration time.
The new awaitFirst turns the things around by await first and then check the condition.

fpj: That's a good observation, @andreipaduroiu, but let me clarify what the proposal would be as I see a few options:

Option 1: we get rid of `assertEventuallyEquals` and only use `await`. That would require changing all occurrences of `assertEventuallyEquals`, which would be out of scope of this PR.
Option 2: we keep both `assertEventuallyEquals` and `await`.
   - Option 2a: we only replace `assertEventuallyEquals` with `await` in the test case we are fixing in this PR.
   - Option 2b: we replace more broadly the relevant occurrences of `assertEventuallyEquals`, but that again feels like out of scope of this PR.

Option 2a is acceptable, but then my next question is what happens to `assertEventuallyEquals`, do we fix in the way that is being proposed or leave it untouched? If the point that @kevinhan88 is raising is a concern, then we should be fixing it here in the case we keep `assertEventuallyEquals`.

What do you think?

andreipaduroiu: The scope of this PR is "Fix bug inside AssertExtensions". That is a bit broad, so I think option 1 can fall into scope. I don't want to have two different methods that are 90% similar and then we'll end up with people using both in the future. This has happened before and it was hard to fix afterwards.

There are only 17 callsites of this method. I think they can be resolved.

@kevinhan88 I don't see how using `await` will touch the cache in any other way compared to what you are doing now. You will simply pass your condition check as a `Supplier` to `await` which will have the same effect.

kevinhan88: @andreipaduroiu are these 2 methods for different purpose and usages? await() may return or throw timeout exception, while assertEventuallyEquals may return or throw AssertException. I don't see how one can replace another.

andreipaduroiu: - Both will fail the test.
- Both are designed to check a condition that is not necessarily true at the moment but may become true at one point in the future.
- `assertEventuallyEquals` can only check for equality. `await` is more flexible and allows you to verify other conditions (ex: not equals, greater, smaller, null, not null, etc.) 

- `await` is more efficient in that it doesn't call `System.currentTimeMillis` every time. 
- You can build `assertEventuallyEquals` using `await`, which will pass in the appropriate callback.

fpj: I have to take responsibility for changing the PR title, I felt that the important part was fixing the bug in the method. As is, the title of the PR is broad so even option 1 falls into the scope of the PR title. The fix of the test case is a side-effect of the change. I'm ok with going with option 1 as long as @kevinhan88 is comfortable doing it in this PR.

kevinhan88:
@andreipaduroiu I believe I got all your points and fixed whatever necessary in the a6a60f57e526891ebf42db51fb827d716a410b96.
1. assertExceptions are special exceptions as code analysis tool treat them differently, so does test report.
2. assertTrue and assertEquals are also different for code analysis and test report. assertEquals tells the value of both sides while assertTrue doesn't.
3. assertTrue doesn't return true, so we cannot pass assertTrue function into await.
For the above reasons I kept asssertEventuallyEquals, but added the intervalMillis into it, following the await style.

andreipaduroiu:
@kevinhan88 I'm ok with these changes as they are. Please file an issue so that we may unify `TestUtils.await` and `AssertExtensions.assertEventuallyEquals` since they are very similar.

kevinhan88:
https://github.com/pravega/pravega/issues/3865 created per the comment.

PR #3849, opened 2019-05-29
Issue 3847: Fix issues causing error in extracting client credentials 

### LINKED ISSUE #3847: Client credentials fail to get populated when specified through environment variables
**Problem description**
If you specify client credentials via environment variables, the following exception occurs, which in turn prevents client credentials from getting populated in the ClientConfig object: 

```
	at io.pravega.client.ClientConfig$ClientConfigBuilder.extractCredentialsFromProperties(ClientConfig.java:136)
	at io.pravega.client.ClientConfig$ClientConfigBuilder.extractCredentials(ClientConfig.java:128)
```

**Problem location**
`extractCredentials(Properties properties, Map<String, String> env)` method of `ClientConfig class`

ravisharda:
**Change log description**  
Adds null checks to prevent errors in extracting client credentials from environment variables. 

**Purpose of the change**  
Fixes #3847. 

**What the code does**  
* Adds null checks in ClientConfig `extractCredentials(...)` to prevent errors in extracting credentials from environment variables. It also adds some log statements for easier debugging. 
* Adds some tests and refactors existing one in `CredentialsExtractorTest` class for verifying client credentials extraction logic more comprehensively. 

**How to verify it**  
All tests must pass.

### DIFF
--- a/client/src/main/java/io/pravega/client/ClientConfig.java
+++ b/client/src/main/java/io/pravega/client/ClientConfig.java
@@ -122,20 +122,22 @@ public class ClientConfig implements Serializable {
         @VisibleForTesting
         ClientConfigBuilder extractCredentials(Properties properties, Map<String, String> env) {
             if (credentials != null) {
-                log.debug("Client credentials were specified by the caller.");
+                log.info("Client credentials were extracted using the explicitly supplied credentials object.");
                 return this;
             }
             if (properties != null) {
                 credentials = extractCredentialsFromProperties(properties);
                 if (credentials != null) {
-                    log.debug("Client credentials were extracted from system properties.");
+                    log.info("Client credentials were extracted from system properties. {}",
+                            "They weren't explicitly supplied as a Credentials object.");
                     return this;
                 }
             }
             if (env != null) {
                 credentials = extractCredentialsFromEnv(env);
                 if (credentials != null) {
-                    log.debug("Client credentials were extracted from environment variables");
+                    log.info("Client credentials were extracted from environment variables. {}",
+                            "They weren't explicitly supplied as a Credentials object or system properties.");
                     return this;
                 }
             }
--- a/client/src/main/java/io/pravega/client/ClientConfig.java
+++ b/client/src/main/java/io/pravega/client/ClientConfig.java
@@ -122,12 +122,22 @@ public class ClientConfig implements Serializable {
         @VisibleForTesting
         ClientConfigBuilder extractCredentials(Properties properties, Map<String, String> env) {
             if (credentials != null) {
+                log.debug("Client credentials were specified by the caller.");
                 return this;
             }
-
-            credentials = extractCredentialsFromProperties(properties);
-            if (credentials == null) {
+            if (properties != null) {
+                credentials = extractCredentialsFromProperties(properties);
+                if (credentials != null) {
+                    log.debug("Client credentials were extracted from system properties.");
+                    return this;
+                }
+            }
+            if (env != null) {
                 credentials = extractCredentialsFromEnv(env);
+                if (credentials != null) {
+                    log.debug("Client credentials were extracted from environment variables");
+                    return this;
+                }
             }
             return this;
         }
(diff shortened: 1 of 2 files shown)


client/src/main/java/io/pravega/client/ClientConfig.java
```diff
@@ -122,12 +122,22 @@ private void extractCredentials() {
...
+                credentials = extractCredentialsFromProperties(properties);
+                if (credentials != null) {
+                    log.debug("Client credentials were extracted from system properties.");
+                    return this;
+                }
+            }
+            if (env != null) {
                 credentials = extractCredentialsFromEnv(env);
+                if (credentials != null) {
+                    log.debug("Client credentials were extracted from environment variables");
```
fpj: Since this is for debugging purposes, do you want to list as part of the log message the variables that have been extracted? Same comment for the other log messages in this file.

client/src/test/java/io/pravega/client/CredentialsExtractorTest.java
```diff
@@ -10,83 +10,202 @@
...
 public class CredentialsExtractorTest {
-    @Test
-    public void testextractCredentials() {
-        //No creds defined
-        ClientConfig config = ClientConfig.builder().build();
-        assertEquals("Empty list should return null", config.getCredentials(), null);
 
-        //Test custom creds
+    @Test
+    public void extractsCredentialsFromProperties() {
```
fpj: For uniformity, can we start the test method name with `test`?

client/src/test/java/io/pravega/client/CredentialsExtractorTest.java
```diff
@@ -10,83 +10,202 @@
...
+                    }
+
+                    @Override
+                    public String getAuthenticationToken() {
+                        return "tokenSpecifiedViaExplicitObject";
+                    }
+                }).extractCredentials(properties, authEnvVariables)
+                .build();
+
+        assertEquals("Explicitly set credentials should not be overridden", "typeSpecifiedViaExplicitObject",
```
fpj: Based on this test case, I´m wondering if we should warn the user that the configuration via system properties and env variables is being ignored. I understand that we follow a given order in the presence of multiple configuration types, but perhaps we want to let the user know that she is not necessarily doing the right thing.

Note that I´m not suggesting a change to the test case, I´m suggesting that we print a warn log message in the case a user specifies multiple types and we by consequence ignore one.

client/src/main/java/io/pravega/client/ClientConfig.java
ravisharda: Since credentials are considered sensitive, they are usually not logged. It might be helpful to know whether they are present, and therefore I added the current log statements, which I think should establish conclusively whether the credentials were present. 

What do you say?

client/src/test/java/io/pravega/client/CredentialsExtractorTest.java
ravisharda: I've added the `test` prefix to all the test methods now.

client/src/test/java/io/pravega/client/CredentialsExtractorTest.java
ravisharda: Warning the user that the configuration via system properties and environment variables are being ignored would be helpful from the usability perspective. There are a couple of issues with that though:

1. The environment variable or system properties might have been specified by someone else at deployment time while the explicit credentials are specified at development time. The user might be seeing a log message as a consequence of someone else's action. 
2. There is a bit of additional processing that is required to assess the condition, in order to find whether to log the message.

I think we should leave it as is for now, especially because there are adequate log statements that make the condition clear: https://github.com/pravega/pravega/blob/7a9c3595e49cbf192ba713d6863df0a3df44775c/client/src/main/java/io/pravega/client/ClientConfig.java#L123

Would you agree?

client/src/main/java/io/pravega/client/ClientConfig.java
fpj: Ok.

client/src/test/java/io/pravega/client/CredentialsExtractorTest.java
fpj: Let's have info level messages to state what method is being used to load the credentials. That way we avoid any confusion about how we are loading it at run time.

ravisharda: Sure, done.

PR #3852, opened 2019-05-29
Issue 3851: (SegmentStore) Bugfix for Segment Id assignment never completing

### LINKED ISSUE #3851: MetadataStore.getOrAssignSegmentId may return a Future that never completes
**Problem description**
For cases when a Segment is not already mapped in the in-memory metadata, the `MetadataStore.getOrAssignSegmentId` creates a Future that is registered internally and is completed (normally or exceptionally) when the underlying assignment completes. This is the Future that is returned to the caller which is then used to do additional processing with the segment.

However, if the assignment throws a synchronous exception, this future will never be completed.

**Problem location**
`MetadataStore`

1. Fail the assignment if it fails synchronously.
2. Have `MetadataStore` implement `AutoCloseable` so it cancels pending requests when shutting down.

andreipaduroiu:
**Change log description**  
1. Fixed a bug in `MetadataStore` where a pending assignment would never complete a returned future if a sync exception is thrown.
2. Always closing the `MetadataStore` when shutting down the owning Segment Container.

**Purpose of the change**  
Fixes #3851.

**What the code does**  
- Context: 
    - Everything inside the Segment Container uses Segment Ids (not names) to operate.
    - A Segment Id is assigned the first time a segment is used, and is stored either in the in-memory metadata or in the Container Metadata Table.
    - If not in memory, the Segment Container delegates to the `MetadatStore` to get-or-assign its id, which is an async operation.
    - Multiple requests for the same segment may arrive at or around the same time (while the first one is executing), in which case subsequent ones are going to be piggybacked on the initial one.
- There was a bug in `MetadataStore.getOrAssignSegmentId` where it would never complete a returned future if the underlying calls to assign the id threw a synchronous exception (async exceptions were handled properly). In this case, whomever was waiting for this would never get any sort of completion.
    - This was triggered (in the system test that failed due to this) by the Segment Container shutting down immediately after an assignment was initiated. This was past the initial "close-check" but before downstream components performed theirs, hence a synchronous `ObjectClosedException` was thrown (and swallowed for eternity).
    - This was fixed by adding a `try-catch` block around the area of code that handled this.
- In addition, we are now closing the`MetadataStore` component when shutting down the Segment Container. This ensures that even if we have any lingering requests, those will be completed (exceptionally) instead of hanging around.

**How to verify it**  
New unit tests added to verify desired behavior.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
@@ -13,6 +13,7 @@ import com.google.common.annotations.VisibleForTesting;
 import io.pravega.common.Exceptions;
 import io.pravega.common.LoggerHelpers;
 import io.pravega.common.ObjectBuilder;
+import io.pravega.common.ObjectClosedException;
 import io.pravega.common.TimeoutTimer;
 import io.pravega.common.concurrent.Futures;
 import io.pravega.common.io.SerializationException;
@@ -61,7 +62,7 @@ import lombok.val;
  */
 @Slf4j
 @ThreadSafe
-public abstract class MetadataStore {
+public abstract class MetadataStore implements AutoCloseable {
     //region Members
 
     protected final String traceObjectId;
@@ -86,6 +87,20 @@ public abstract class MetadataStore {
         this.connector = connector;
         this.executor = executor;
         this.pendingRequests = new HashMap<>();
+    }
+
+    @Override
+    public void close() {
+        // Even though all pending requests should be cancelled from any components that they may be waiting upon, cancel
+        // them anyway to ensure the caller will get notified.
+        ArrayList<PendingRequest> toCancel;
+        synchronized (this.pendingRequests) {
+            toCancel = new ArrayList<>(this.pendingRequests.values());
+            this.pendingRequests.clear();
+        }
+
+        val ex = new ObjectClosedException(this);
+        toCancel.forEach(r -> r.completeExceptionally(ex));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


andreipaduroiu:
System test run 2008 passed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
```diff
@@ -88,6 +89,20 @@
         this.pendingRequests = new HashMap<>();
     }
 
+    @Override
+    public void close() {
+        // Even though all pending requests should be cancelled from any components that they may be waiting upon, cancel
+        // them anyway to ensure the caller will get notified.
+        ArrayList<PendingRequest> toCancel;
+        synchronized (this.pendingRequests) {
+            toCancel = new ArrayList<>(this.pendingRequests.values());
```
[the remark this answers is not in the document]
andreipaduroiu: Because you should never complete or fail a future while holding a lock. Bad things can happen.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
```diff
@@ -373,17 +388,23 @@
...
      * @param segmentName The name of the Segment to assign id for.
      * @param timeout     Timeout for the operation.
      */
     private void assignSegmentId(String segmentName, Duration timeout) {
-        TimeoutTimer timer = new TimeoutTimer(timeout);
-        Futures.exceptionListener(
-                getSegmentInfoInternal(segmentName, timer.getRemaining())
-                        .thenComposeAsync(si -> submitAssignmentWithRetry(SegmentInfo.deserialize(si), timer.getRemaining()), this.executor),
-                ex -> failAssignment(segmentName, ex));
+        try {
```
tkaitchuck: It seems like there needs to be a check if the Store is closed here.

andreipaduroiu: Either `getSegmentInfoInternal` or `submitAssignment` will throw that. The former will resort to the `Storage` adapter which will be closed (if the container is closed) and the latter will run into the `OperationProcessor`, which is the first thing that gets shut down.

The latter is exactly what triggered this problem in the first place, and is tested via the new unit test added.

PR #3861, opened 2019-05-31
Issue 3774:  make AppendBatchSizeTracker per writer/reader  and disable the flowids for segment append paths

### LINKED ISSUE #3774: Performance drop in Pravega master  and 0.5 branch
**Problem description**
The Pravega performance is droped in the master of Pravega.

**Problem location**
Following commit id is causing lower performance.

02f1c0036da29e2db34cc319714d99fd08da941a is the first bad commit
commit 02f1c0036da29e2db34cc319714d99fd08da941a
Author: Sandeep <sandeep.shridhar@emc.com>
Date:   Thu Apr 11 02:25:11 2019 +0530

   Issue 217: Enable connection pooling with SegmentStore. (#3562)

   * Implements connection pooling infrastructure and enables
   connection pooling for RawClient and Pravega Readers and
   Pravega Writers.

This commit id : 02f1c0036da29e2db34cc319714d99fd08da941a

kmgowda:
**Change log description**  
Currently there performance drop performance drop due the connection pooling implementation.
https://github.com/pravega/pravega/commit/02f1c0036da29e2db34cc319714d99fd08da941a. 
It is also identified that AppendBatchSizeTracker was used by multiple connections and thus last append number and ack numbers were incorrect. This PR fixes this AppendBatchSizeTracker  bug .
Another issue identified that  in the command encoder , whenever the segment and writer id changes, the appends are broken and thus causing performance drop.  So, This PR improves the performance of writer by disabling flowid for append path.

**Purpose of the change**  
fixes #3774 

**What the code does**  
1. The AppendBatchSizeTracker  is maintained as per writer id (for append paths) and tracker ids as for segment read paths in the netty handler.
2. The flowids are disabled for append only paths and thus there will be dedicated netty channel for segment writes and avoids the append breaks. this improves the latency and throughput.

**How to verify it**  
1. Pravega is tested with bare-metal deployment with 1 controller and 3 segment nodes on Lgaunitas (white cluster).
2. Benchmark mark tool : https://github.com/pravega/pravega-benchmark is used for performance bench-marking and ensured that the write performance is re gained and improved.

Signed-off-by: Keshava Munegowda <keshava.munegowda@dell.com>

### DIFF
--- a/client/src/main/java/io/pravega/client/netty/impl/ClientConnectionImpl.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/ClientConnectionImpl.java
@@ -34,14 +34,12 @@ public class ClientConnectionImpl implements ClientConnection {
     @VisibleForTesting
     @Getter
     private final FlowHandler nettyHandler;
-    private final AppendBatchSizeTracker batchSizeTracker;
+
     private final AtomicBoolean closed = new AtomicBoolean(false);
 
-    public ClientConnectionImpl(String connectionName, int flowId, AppendBatchSizeTracker batchSizeTracker,
-                                FlowHandler nettyHandler) {
+    public ClientConnectionImpl(String connectionName, int flowId, FlowHandler nettyHandler) {
         this.connectionName = connectionName;
         this.flowId = flowId;
-        this.batchSizeTracker = batchSizeTracker;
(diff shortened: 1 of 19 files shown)


kmgowda:
i will raise new PR with code cleanup

PR #3862, opened 2019-05-31
Issue 3853: (SegmentStore) Bugfix for ConcurrentModificationException in SegmentAttributeIterator

### LINKED ISSUE #3853: ConcurrentModificationException in SegmentAttributeIterator
**Problem description**
```
2019-05-29 21:43:29,206 9602982 [core-13] ERROR i.p.s.s.h.h.PravegaRequestProcessor - [requestId=140255] Error (Segment = '_system/_tables/testReadTxnWriteAutoScaleScope1622743218/testReadTxnWriteAutoScaleStream/transactionsInEpoch-4.#.81939009-6a2e-482b-81fb-857a5e7a3936', Operation = 'readTableKeys')
java.util.ConcurrentModificationException: null
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1442)
	at java.util.HashMap$EntryIterator.next(HashMap.java:1476)
	at java.util.HashMap$EntryIterator.next(HashMap.java:1474)
	at io.pravega.common.util.CollectionHelpers$ConvertedIterator.next(CollectionHelpers.java:271)
	at com.google.common.collect.Iterators$ConcatenatedIterator.next(Iterators.java:1363)
	at com.google.common.collect.Iterators$1.next(Iterators.java:141)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at io.pravega.segmentstore.server.containers.SegmentAttributeIterator.<init>(SegmentAttributeIterator.java:72)
	at io.pravega.segmentstore.server.containers.StreamSegmentContainer.lambda$attributeIterator$38(StreamSegmentContainer.java:749)
```

**Problem location**
SegmentAttributeIterator.

Fix.

andreipaduroiu:
**Change log description**  
- Fixed a bug in `SegmentAttributeIterator` where the `StreamSegmentMetadata` attribute iteration was not thread safe and was prone to `ConcurrentModificationException`s.

**Purpose of the change**  
Fixes #3853.

**What the code does**  
- Created `StreamSegmentMetadata.getAttributes(Filter)` which iterates and filters while holding the same lock used for modifying its contents.
- Updated `SegmentAttributeIterator` to use this method for retrieving desired attributes.

**How to verify it**  
- Concurrency cannot be easily tested in unit tests.
- Updated a `StreamSegmentMetadata` unit test to verify new method.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.BiPredicate;
 import java.util.stream.Collectors;
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
@@ -185,6 +186,13 @@ public class StreamSegmentMetadata implements UpdateableSegmentMetadata {
     }
 
     @Override
+    public synchronized Map<UUID, Long> getAttributes(BiPredicate<UUID, Long> filter) {
+        return getAttributes().entrySet().stream()
+                              .filter(e -> filter.test(e.getKey(), e.getValue()))
+                              .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    @Override
     public String toString() {
         return String.format(
                 "Id = %d, Start = %d, Length = %d, StorageLength = %d, Sealed(M/S) = %s/%s, Deleted = %s, Name = %s",
@@ -275,7 +283,7 @@ public class StreamSegmentMetadata implements UpdateableSegmentMetadata {
     @Override
     public synchronized void setLastModified(ImmutableDate date) {
         this.lastModified = date;
-        log.trace("{}: LastModified = {}.", this.lastModified);
+        log.trace("{}: LastModified = {}.", this.traceObjectId, this.lastModified);
     }
 
     @Override
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentMetadata.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentMetadata.java
@@ -10,6 +10,9 @@
 package io.pravega.segmentstore.server;
 
 import io.pravega.segmentstore.contracts.SegmentProperties;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.BiPredicate;
 
 /**
  * Defines an immutable StreamSegment Metadata.
@@ -89,4 +92,12 @@ public interface SegmentMetadata extends SegmentProperties {
      * @return True if pinned, false otherwise.
      */
     boolean isPinned();
+
+    /**
+     * Gets new Map containing all the Attributes for this Segment that match the given filter.
+     *
+     * @param filter The Key-Value filter that will be used.
+     * @return A new Map containing the result.
+     */
+    Map<UUID, Long> getAttributes(BiPredicate<UUID, Long> filter);
 }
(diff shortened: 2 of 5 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
```diff
@@ -184,6 +185,13 @@ public synchronized long getLength() {
         return new AttributesView();
     }
 
+    @Override
+    public synchronized Map<UUID, Long> getAttributes(BiPredicate<UUID, Long> filter) {
+        return getAttributes().entrySet().stream()
```
tkaitchuck: This doesn't actually remove the problematic getAttributes() method. It's right above, public, and unsafe.

andreipaduroiu: It's not removing it. The method above can be used for retrieving values. It needs to be there. It's just not good for iteration.

andreipaduroiu:
System test run 200 completed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
tkaitchuck: Then perhaps the best way to do that would be to wrap it an make it throw if you attempt to iterate over the values.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
```diff
@@ -184,6 +185,13 @@ public synchronized long getLength() {
         return new AttributesView();
     }
 
+    @Override
+    public synchronized Map<UUID, Long> getAttributes(BiPredicate<UUID, Long> filter) {
```
fpj: Is it necessary to make it a precondition that the filter argument is not null?

andreipaduroiu: Sure. I can do that. But whether I throw NullPtrEx now or whether it gets thrown on the next line really makes no difference. We'd just be adding more code to provide no extra value.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
tkaitchuck: The `getAttributes` method comes from the `SegmentProperties` interface which just says in it's javadoc "Gets a read-only Map of AttributeId-Values for this Segment.". Nothing would lead you to believe a lock was required. Looking at the code, I can see 10 places where that method is used to return a map that is iterated over, and none of them hold a lock that I can see.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
fpj: Ok.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadata.java
andreipaduroiu: The callsites where you see that being iterated over are all covered by a lock. They are only invoked by the OperationProcessor while holding its own lock (same lock used to update them). 

There's a difference between `SegmentProperties` and `SegmentMetadata`. Whomever uses the latter only does so internally, usually inside the pipeline. `SegmentProperties` is used externally and every place that is returned, we create a new object via `SegmentMetadata.getSnapshot` which is thread safe.

andreipaduroiu: @tkaitchuck After our private conversation, I added the `getAttributes` in `SegmentMetadata` and explained its behavior in its JavaDoc.

PR #3863, opened 2019-06-03
Issue 3909: AppendBatchSizeTracker bug fix 

### LINKED ISSUE #3774: Performance drop in Pravega master  and 0.5 branch

### LINKED ISSUE #3909: Bug:  Append batchsize Tracker event numbers mismatch between recordAppend and  recordAck
**Problem description**
Its observed that when multiple writer/readers (or multiple flowids)  uses the same connection , the appendbatch size tracker's recordAppend  and recordAck methods are not in sync with event numbers.

**Problem location**
CommandEncoder and Netty Flow Handlers of Pravega Client

Create AppendBatchSizeTracker per flowids. so, that event numbers are matched based on the flow ids.

kmgowda:
**Change log description**  
The AppendBatchSizeTracker is created per flowids.

**Purpose of the change**  
Fixes #3909 

**What the code does**  
1. The AppendBatchSizeTracker is maintained as per flow id id the netty handler.

**This PR does changes to following classes/Interfaces:**
*1. FlowHandler*
*2. CommandEncoder*

----
*1. FlowHandler*
following new methods are introduced.

1). **createAppendBatchSizeTrackerIfneeded**

creates the AppendBatchSize Traker per input flowid. if the appendbatchSizeTracker already exists for flowid and no need to create a new one.

Note that, the flowHandler object maintains the map of flowid-> appendBatchSizeTracker.

2). **getAppendBatchSizeTracker**

return the appendBatchSizeTracker for given flowid

*2. CommandEncoder*
This class is used to encode the message for netty channel. 
Since this class object can used among several writers and readers,
the class uses the **getAppendBatchSizeTracker** method supplied by FlowHandler to identify the appendBatchSizeTracker for append requestid(flowid).

----

**How to verify it**  
1. Pravega is tested with bare-metal deployment with 1 controller and 3 segment nodes on Lagaunitas (white cluster).
2. Benchmark mark tool : https://github.com/pravega/pravega-benchmark is used for performance bench-marking.

Signed-off-by: Keshava Munegowda keshava.munegowda@dell.com


client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -96,6 +97,33 @@ public void closeFlow(ClientConnection clientConnection) {
...
+    public void createAppendBatchSizeTracker(final UUID id) {
+        Exceptions.checkNotClosed(closed.get(), this);
+        if (idBatchSizeTrackerMap.containsKey(id)) {
+            log.info("Reusing Batch size tracker for id {}.", id);
+        } else {
+            log.info("Creating Batch size tracker for id {}.", id);
+            idBatchSizeTrackerMap.put(id, new AppendBatchSizeTrackerImpl());
+        }
+    }
+
```
andreipaduroiu: Too many empty lines

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -96,6 +97,33 @@ public void closeFlow(ClientConnection clientConnection) {
...
+
+    /**
+     * get a Batch size tracker on existing connection.
```
andreipaduroiu: "Get"

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -96,6 +97,33 @@ public void closeFlow(ClientConnection clientConnection) {
...
+     *
+     * @param id identifier.
+     * @return Batch size Tracker object.
```
andreipaduroiu: What if the id isn't registered?

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -237,6 +269,8 @@ public void close() {
                 if (openFlowCount != 0) {
                     log.warn("{} flows are not closed", openFlowCount);
                 }
+                log.debug("clearing Append Batch size Trackers {}", idBatchSizeTrackerMap);
```
andreipaduroiu: "Clearing".

Also have you checked what the output of this log instruction is? You are calling `toString` on a `Map`. That will likely just tell you how many elements are in the map.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -96,6 +97,33 @@ public void closeFlow(ClientConnection clientConnection) {
...
+    }
+
+
+    /**
+     * get a Batch size tracker on existing connection.
+     *
+     * @param id identifier.
+     * @return Batch size Tracker object.
+     */
+    public AppendBatchSizeTracker getAppendBatchSizeTracker(final UUID id) {
```
tkaitchuck: I don't think we need to tie this to the writerId. It can be based on the flowId. This will make initializing this map, and removing entries when they are no longer needed much simpler.
It will also have the added benefit that the non-append code paths aren't affected.

client/src/main/java/io/pravega/client/netty/impl/ConnectionFactory.java
```diff
@@ -24,21 +25,23 @@
...
      * This method is used to establish a client connection using a {@link Flow} on the underlying Connection
      * pool.
      * @param flow  Flow to be used to create a client connection.
+     * @param id     Identifier
      * @param endpoint The Pravega Node URI.
      * @param rp Reply Processor instance.
      * @return An instance of client connection.
      */
-    CompletableFuture<ClientConnection> establishConnection(Flow flow, PravegaNodeUri endpoint, ReplyProcessor rp);
+    CompletableFuture<ClientConnection> establishConnection(Flow flow, UUID id, PravegaNodeUri endpoint, ReplyProcessor rp);
```
tkaitchuck: I don't think we need to add UUID. See my comment below.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: Hi @tkaitchuck @shrids 
    The channelRead method of netty handler , calls the recordAck of append batch size tracker upon receiving the WireCommands.DataAppended . 
The DataAppended has the writerid;  so, if the appendbatchsizetracker per flow id; then we need to maintain the map between   "writer id -> flowid -> appendbatchsizetracker".

to do this,  writer id is required while creating the connection and flow id.
so, its better have the direct mapping of "writer id ->appendbatchsizetracker"

for append path, SegmentOutputStream   has to behave differently for state sycnhronizer and byte streamer.

with these changes,  out next changes will as as follows.

1. For state sycnhronizer - we can have SegmentOutputStream    with flowid and connection pooling.
2. for byte stream  - we can have SegmentOutputStream   with flowid disabled.


so,  i feel its better to keep the appendbatchsizetracker independent of flowid.

tkaitchuck: To send the DataAooended command to the correct SegmentOutputStream.ReplyProcessor the FlowHandler is doing this:
```
        int flowId = disableFlow.get() ? FLOW_DISABLED : Flow.from(cmd.getRequestId()).getFlowId();
        final ReplyProcessor processor = flowIdReplyProcessorMap.get(flowId);
```
I'm not sure why we can't do the same thing to maintain a map to the appendBatchSizeTracker.

kmgowda: @tkaitchuck ,
   yes , we can have this flowid -> appendsizeTracker ;
but the problem is this.
                     public void channelRead(ChannelHandlerContext ctx, Object msg) { 
                                  ....
                                  ....
                                    if (cmd instanceof WireCommands.DataAppended) {
                                                      final WireCommands.DataAppended dataAppended = 
                                                         (WireCommands.DataAppended) cmd;
                                   final AppendBatchSizeTracker batchSizeTracker = 
                                             getAppendBatchSizeTracker(dataAppended.getWriterId());
                                    if (batchSizeTracker != null) {
                                                       batchSizeTracker.recordAck(dataAppended.getEventNumber());
                                    }
        }  
                     }    


The  WireCommands.DataAppended   has the writerid , so we need mapping from writerid to appendbatchsizetracker.
so, we will end up mapping something like this
      writer id  -> flow id
and then
      flowid  -> appendbatchsize tracker.

however,  we can not avoid  writer id here.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -96,6 +97,33 @@ public void closeFlow(ClientConnection clientConnection) {
         flowIdReplyProcessorMap.remove(flow);
     }
 
+    /**
+     * Create a Batch size tracker on existing connection.
+     *
+     * @param id identifier.
+     */
+    public void createAppendBatchSizeTracker(final UUID id) {
```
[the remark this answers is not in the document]
kmgowda: yes, I will change it accordingly

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
shrids: @kmgowda 
Every Segment writer has a unique request id which is created by using 
```
private final long requestId = Flow.create().asLong();
```
The same requestId is used when the SegmentStore replies to Append requests.
Hence we can use the same method used to fetch the correct `ReplyProcessor` to fetch the correct `AppendBatchTracker`

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -43,16 +44,16 @@
     private final AtomicBoolean recentMessage = new AtomicBoolean(false);
     private final AtomicBoolean closed = new AtomicBoolean(false);
     @Getter
-    private final AppendBatchSizeTracker batchSizeTracker;
     private final ReusableFutureLatch<Void> registeredFutureLatch = new ReusableFutureLatch<>();
     @VisibleForTesting
     @Getter(AccessLevel.PACKAGE)
     private final ConcurrentHashMap<Integer, ReplyProcessor> flowIdReplyProcessorMap = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<UUID, AppendBatchSizeTracker> idBatchSizeTrackerMap = new ConcurrentHashMap<>();
```
shrids: We can use `ConcurrentHashMap<Integer, AppendBatchSizeTracker> idBatchSizeTrackerMap` to maitain a map between the segment writer and its corresponding AppendBatchSizeTracker.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -96,6 +97,33 @@ public void closeFlow(ClientConnection clientConnection) {
         flowIdReplyProcessorMap.remove(flow);
```
shrids: When the client connection is closed , we close the Flow. In such a scenario we need to remove the corresponding writer from `idBatchSizeTrackerMap ` .

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: @shrids 
But, requestID increments/changes in case of /pravega/client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java [ConditionalOutputStreamImpl class]

<pre>
ConditionalOutputStreamImpl class
public boolean write(ByteBuffer data, long expectedOffset) throws SegmentSealedException {
     ....
     ....
 <b style="color:red" >  long requestId = client.getFlow().getNextSequenceNumber();   </b>
                            log.debug("Setting up append on segment: {}", segmentId);
                            SetupAppend setup = new SetupAppend(requestId, writerId,
                                                                segmentId.getScopedName(),
                                                                delegationToken);
                            val reply = client.sendRequest(requestId, setup);  
}
</pre>

so,  by calling **getNextSequenceNumber** and using it as **requestid** will create a  collision in the map of flowid to appendbatchsizeTracker.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
tkaitchuck: @shrids Can we make that integer into an object? Possibly by defining a equals and hashcode on flow or having a flowid? I'd rather avoid a primitive as it is non-obvious what it means/is.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
tkaitchuck: If you look at `Flow` I think that long comes from two ints, only one of which changes.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
shrids: yes, we can make it into an object which represents the FlowId here.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: Thanks Tom; yes, I will have map for flowid -> appendbatchsizeTracker..
changes are in progress.. I will update the PR ASAP.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -36,23 +36,23 @@
 @Slf4j
 public class FlowHandler extends ChannelInboundHandlerAdapter implements AutoCloseable {
 
-    private static final int FLOW_DISABLED = -1;
+    private static final int FLOW_DISABLED = 0;
```
tkaitchuck: What was wrong with -1?

kmgowda:
The latest commit id: https://github.com/pravega/pravega/pull/3863/commits/1a179097677cd104ba41a2f5685d84cae4c29cb5

retains the interfaces of ConnectionPool and ConnectionFactory without UUID; thus reduces the code changes.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: removed

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: done

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: now .. i changed the prototype of this method. now , flow id needs to passed.
if its not existing then it returns null.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: this change is removed now.. along with flow , the append size tracker map entries are removed.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: done

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
kmgowda: if the flow id is disabled ; the request id should start from the value :  (-1<<32) (-1, 0). This could be confusing.. so, if the flow id is disabled , the user of flow handler need not be aware that initial value should be (-1<<32) rather it can be just 0.
it was no wrong with having -1 , 0 is better  to understand avoids bugs in case flow id is disabled.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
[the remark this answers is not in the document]
kmgowda: yes, its done now

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -96,6 +97,33 @@ public void closeFlow(ClientConnection clientConnection) {
...
 
+    /**
+     * Create a Batch size tracker on existing connection.
+     *
+     * @param id identifier.
+     */
+    public void createAppendBatchSizeTracker(final UUID id) {
+        Exceptions.checkNotClosed(closed.get(), this);
+        if (idBatchSizeTrackerMap.containsKey(id)) {
+            log.info("Reusing Batch size tracker for id {}.", id);
```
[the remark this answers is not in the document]
kmgowda: done

client/src/main/java/io/pravega/client/netty/impl/ConnectionFactory.java
kmgowda: yes its done.. file changes are reduced from 22 to 12 now.

client/src/test/java/io/pravega/client/segment/impl/SegmentMetadataClientTest.java
```diff
@@ -282,13 +282,13 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
...
-        InOrder order = Mockito.inOrder(connection1, connection2, cf);
         long length = client.fetchCurrentSegmentLength();
-        order.verify(cf, Mockito.times(2)).establishConnection(Mockito.any(Flow.class), Mockito.eq(endpoint), Mockito.any());
+        InOrder order = Mockito.inOrder(connection1, connection2, cf);
+        order.verify(cf, Mockito.times(2)).establishConnection(Mockito.any(Flow.class),  Mockito.eq(endpoint), Mockito.any());
         order.verify(connection1).sendAsync(Mockito.eq(new WireCommands.GetStreamSegmentInfo(requestIds.get(0), segment.getScopedName(), "")),
                                             Mockito.any(ClientConnection.CompletedCallback.class));
         order.verify(connection1).close();
-        order.verify(cf).establishConnection(Mockito.any(Flow.class), Mockito.eq(endpoint), Mockito.any());
+        order.verify(cf).establishConnection(Mockito.any(Flow.class),  Mockito.eq(endpoint), Mockito.any());
```
shrids: nit: remove extra space.

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndChannelLeakTest.java
```diff
@@ -170,15 +170,16 @@ public void testDetectChannelLeakSegmentSealedPooled() throws Exception {
         event = reader1.readNextEvent(10000);
         assertNotNull(event.getEvent());
 
-        assertChannelCount(5, connectionPool);
+        //the append paths increases the connections
+        assertChannelCount(12, connectionPool);
```
shrids: We are breaching the max channel count configured
`ClientConfig.builder().maxConnectionsPerSegmentStore(5)` which is supposed to be 5.

kmgowda: yes.. this is because append paths will have dedicated connection.
I will raise another PR in which except the EventWriter rest other append paths will use connection pool.

shrids: thanks for the clarification.

shrids: We could update the javadoc of `maxConnectionsPerSegmentStore` to indicate that Connection pooling is disabled for writers.

kmgowda: I will create another PR for this.

client/src/test/java/io/pravega/client/segment/impl/SegmentMetadataClientTest.java
kmgowda: done

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/AppendEncodeDecodeTest.java
```diff
@@ -66,6 +68,7 @@
     private Level origionalLogLevel;
 
     private EventExecutor executor =  new EventExecutor() {
+        private final ScheduledExecutorService schedular  = Executors.newSingleThreadScheduledExecutor();
```
fpj: There is a typo in the variable name.

client/src/main/java/io/pravega/client/netty/impl/ClientConnectionImpl.java
```diff
@@ -52,9 +51,7 @@ public void send(WireCommand cmd) throws ConnectionFailedException {
     @Override
     public void send(Append append) throws ConnectionFailedException {
         checkClientConnectionClosed();
-        final AppendBatchSizeTracker batchSizeTracker = nettyHandler.getAppendBatchSizeTracker(append.getFlowId());
```
fpj: This seems to be removing the `AppendBatchSizeTracker` altogether, is my observation and if so why?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,66 +86,57 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
         log.trace("Encoding message to send over the wire {}", msg);
         if (msg instanceof Append) {
             Append append = (Append) msg;
+            if (!append.segment.equals(segmentBeingAppendedTo) || !append.getWriterId().equals(writerIdPerformingAppends)) {
```
fpj: The description of this PR says:

```
... the class uses the getAppendBatchSizeTracker method supplied by FlowHandler to identify the appendBatchSizeTracker for append requestid(flowid).
```

but I don't see it being used here. There is a discrepancy between what this is doing and what is described in the PR.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -169,20 +158,23 @@ private void validateAppend(Append append, Session session) {
                 "Bug in CommandEncoder.encode, block is too small.");
     }
 
-    private void breakFromAppend(ByteBuf out) {
+    private void breakFromAppend(ByteBuf data, ByteBuf pendingData, ByteBuf out) {
         if (bytesLeftInBlock != 0) {
-            writeMessage(new Padding(bytesLeftInBlock - TYPE_PLUS_LENGTH_SIZE), out);
+            if (data != null) {
```
fpj: I don't understand this change. This is replacing a single line:

```
writeMessage(new Padding(bytesLeftInBlock - TYPE_PLUS_LENGTH_SIZE), out);
```

with:

```
            if (data != null) {
                 writeMessage(new PartialEvent(data), out);
            } else {
                writeMessage(new Padding(bytesLeftInBlock - TYPE_PLUS_LENGTH_SIZE), out);
            }
```

This seems to be unrelated to the tracking that is being added, please clarify.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -64,16 +65,13 @@
 @Slf4j
 public class CommandEncoder extends MessageToByteEncoder<Object> {
     private static final byte[] LENGTH_PLACEHOLDER = new byte[4];
-    private static final long RESET_TOKEN = 0;
-
     private final Function<Long, AppendBatchSizeTracker> appendTracker;
     private final Map<Map.Entry<String, UUID>, Session> setupSegments = new HashMap<>();
-    private long currentToken = RESET_TOKEN;
+    private AtomicLong tokenCounter = new AtomicLong(0);
```
fpj: I don't understand why we need this token counter. To start with, why are we introducing this concept of token here?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -64,16 +65,13 @@
 @Slf4j
 public class CommandEncoder extends MessageToByteEncoder<Object> {
```
fpj: My understanding is that the changes to this encoding are only about tracking, and I'd like a confirmation that this is not breaking compatibility with previous commits. If it does break, then we need to label the PR accordingly.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/AppendEncodeDecodeTest.java
```diff
@@ -399,6 +410,28 @@ private void testFlush(int size) throws Exception {
         assertEquals(keepAlive, two);
     }
 
+
+    @Test
+    public void testAppendWithoutBatchTracker() throws Exception {
```
fpj: I'm not sure what it means "without batch tracker" in the context of this test case.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/AppendEncodeDecodeTest.java
```diff
@@ -407,20 +440,24 @@ public void testblockTimeout() throws Exception {
         Arrays.fill(content, (byte) 1);
         Event event = new Event(Unpooled.wrappedBuffer(content));
         idBatchSizeTrackerMap.remove(1L);
-        idBatchSizeTrackerMap.put(1L, new FixedBatchSizeTracker(3));
+        idBatchSizeTrackerMap.put(1L, new FixedBatchSizeTracker(appendBlockSize));
```
fpj: where is `appendBlockSize` defined?

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/AppendEncodeDecodeTest.java
kmgowda: yes.. its changed now.. i will post the update

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: The encode method of the CommandEncoder does the following:
           if (appendTracker != null) {
                blockSizeSupplier = appendTracker.apply(append.getFlowId());
                blockSizeSupplier.recordAppend(append.getEventNumber(), msgSize);
            }
getAppendBatchSizeTracker  is supplied and appendTracker.apply inovkes the method getAppendBatchSizeTracker  to retrieve the append batch size tracker for the given flowid.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: The appendbreak happens in the two conditions
1.  event size more than the append block size (max is 32K) ; then we write the PartialEvent  (if case) 
2.   current segment id or writer id changed on the netty channel, then we pad the remaining block size. (else case)

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: This is because of connection pooling;  
whenever we start append.. we generate the token give to timer which is set to execute after append batch time out. when time out happens then we check are we at the same counter; if not then no need to write any data because its already written as part of append breaks.
if the same token number exists , then issue the blocktimeout msg to enoder of the CommandEncoder.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes.. these changes are only fix the append batch size tracker and reduces the block timeout events too. This PR does not change anything with encoding or data formatting.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/AppendEncodeDecodeTest.java
kmgowda: if there is no append batch size tracker in the CommandEncoder then every append is followed by the appendblockend. No, blockTimeout  timer configuration.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/AppendEncodeDecodeTest.java
kmgowda: its defined in the beginning of the 
public class AppendEncodeDecodeTest {

    private final int appendBlockSize = 1024;  
...
...
}

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: earlier.. (before connection pooling) .. there was dedicated channel segments writers , block time out used to check with current block size only (even that had issues) .. its all fixed now.

client/src/main/java/io/pravega/client/netty/impl/ClientConnectionImpl.java
kmgowda: no.. its moved to CommandEncoder

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -89,66 +86,57 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
         if (msg instanceof Append) {
             Append append = (Append) msg;
+            if (!append.segment.equals(segmentBeingAppendedTo) || !append.getWriterId().equals(writerIdPerformingAppends)) {
+                breakFromAppend(null, null, out);
+            }
+            ByteBuf data = append.getData().slice();
+            int msgSize = data.readableBytes();
+            AppendBatchSizeTracker blockSizeSupplier = null;
             Session session = setupSegments.get(new SimpleImmutableEntry<>(append.segment, append.getWriterId()));
             validateAppend(append, session);
```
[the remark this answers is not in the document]
kmgowda: its done now

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,56 +89,54 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
             if (!append.segment.equals(segmentBeingAppendedTo) || !append.getWriterId().equals(writerIdPerformingAppends)) {
-                breakFromAppend(out);
+                breakFromAppend(null, null, out);
+            }
+            ByteBuf data = append.getData().slice();
```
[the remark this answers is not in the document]
kmgowda: But, this will affect the user supplied buffer. 
I saw this one ByteArraySerializer .. Byte Array is used as it is for ByteBuf.wrap which does not create any copy. so, its quite dangerous use supplied buffer as it is.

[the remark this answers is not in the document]
kmgowda: however, if we take care the ByteArraySerializer  to take the byte copy , then we can remove the .slice() from the CommandEncoder. 
But, I will do it as separate PR ;  changing this requires changes in the unit tests also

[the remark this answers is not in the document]
shrids: `ByteBuf.slice()` creates a derived buffer which shares the memory region of the parent buffer.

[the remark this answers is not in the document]
kmgowda: but ,its copy.  if we dont use the .slice() then whatever changes to data will reflect the append.Data.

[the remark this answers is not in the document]
kmgowda: ok .. got it : https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#slice--

no duplicate buffer ; but indexes are separate; then we can keep the slice in CommandEncoder its good.. no additional buffer only indices differs

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -64,15 +65,15 @@
 @Slf4j
 public class CommandEncoder extends MessageToByteEncoder<Object> {
     private static final byte[] LENGTH_PLACEHOLDER = new byte[4];
-
-    private final AppendBatchSizeTracker blockSizeSupplier;
+    private final Function<Long, AppendBatchSizeTracker> appendTracker;
     private final Map<Map.Entry<String, UUID>, Session> setupSegments = new HashMap<>();
+    private AtomicLong tokenCounter = new AtomicLong(0);
```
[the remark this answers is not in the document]
kmgowda: done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,56 +89,54 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+            if (appendTracker != null) {
+                blockSizeSupplier = appendTracker.apply(append.getFlowId());
+                blockSizeSupplier.recordAppend(append.getEventNumber(), msgSize);
             }
             if (bytesLeftInBlock == 0) {
-                currentBlockSize = Math.max(TYPE_PLUS_LENGTH_SIZE, blockSizeSupplier.getAppendBlockSize());
+                if (blockSizeSupplier != null) {
+                    currentBlockSize = Math.max(TYPE_PLUS_LENGTH_SIZE, blockSizeSupplier.getAppendBlockSize());
+                } else {
+                    currentBlockSize = msgSize + TYPE_PLUS_LENGTH_SIZE;
```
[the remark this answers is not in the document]
kmgowda: Now, CommandEncoder does not assumes the batchsize tracker as mandatory ; if batchsize tracker not available then append end is followed by append block

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,56 +89,54 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+                }
                 bytesLeftInBlock = currentBlockSize;
                 segmentBeingAppendedTo = append.segment;
                 writerIdPerformingAppends = append.writerId;
                 writeMessage(new AppendBlock(session.id), out);
-                if (ctx != null) {
-                    ctx.executor().schedule(new BlockTimeouter(ctx.channel(), currentBlockSize),
+                if (ctx != null && blockSizeSupplier != null && currentBlockSize > (msgSize + TYPE_PLUS_LENGTH_SIZE)) {
```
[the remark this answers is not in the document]
kmgowda: yes done.. blockSizeSupplier checking is removed

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,56 +89,54 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+            AppendBatchSizeTracker blockSizeSupplier = null;
+            session.lastEventNumber = append.getEventNumber();
+            session.eventCount++;
+            if (appendTracker != null) {
+                blockSizeSupplier = appendTracker.apply(append.getFlowId());
+                blockSizeSupplier.recordAppend(append.getEventNumber(), msgSize);
             }
             if (bytesLeftInBlock == 0) {
-                currentBlockSize = Math.max(TYPE_PLUS_LENGTH_SIZE, blockSizeSupplier.getAppendBlockSize());
+                if (blockSizeSupplier != null) {
```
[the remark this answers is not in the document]
kmgowda: i dint create the new method ; but i have improved the verbosity.. please check

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,56 +89,54 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+                breakFromAppend(null, null, out);
+            }
+            ByteBuf data = append.getData().slice();
+            int msgSize = data.readableBytes();
+            AppendBatchSizeTracker blockSizeSupplier = null;
+            session.lastEventNumber = append.getEventNumber();
+            session.eventCount++;
+            if (appendTracker != null) {
+                blockSizeSupplier = appendTracker.apply(append.getFlowId());
+                blockSizeSupplier.recordAppend(append.getEventNumber(), msgSize);
```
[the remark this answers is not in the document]
kmgowda: earlier it was before.. so i maintained the same.. how ever i have checked putting after getting batch size ; but i dint see any improvement in the performance.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -237,6 +269,10 @@ public void close() {
                 if (openFlowCount != 0) {
                     log.warn("{} flows are not closed", openFlowCount);
                 }
+                final int appendTrackerCount = flowIDBatchSizeTrackerMap.size();
+                if (appendTrackerCount != 0) {
+                    log.warn("{} AppendBatchSizeTrackers are not closed", appendTrackerCount);
```
andreipaduroiu: So why are you not closing them?

kmgowda: ok.. i will check this. I guess., connection is getting closed only when flow id count reaches 0. however i check , if required i post another JIIRA

PR #3866, opened 2019-06-04
Issue 3344: (watermarking) Add keyspace to Position

### LINKED ISSUE #3870: Add keyspace to position
**Problem description**
For watermarking (issue #3344 ), In order to have a reader read from a mark segment and compare its position against a stream cut so as to be able to locate the correct time, it must be able to compare a stream cut to a position object. 

**Problem location**
Watermarking (#3344)

This change adds the keyspace information to the position object. This will allow it to identify the segments from the entry in a mark segment that overlap to determine which goes first.

tkaitchuck:
**Change log description**  
Adds keyspace information to the position object.

**Purpose of the change**  
Fixes #3870. This is a prerequisite to reader side watermarking support.

**What the code does**  
In order to have a reader read from a mark segment and compare it's position against a stream cut so as to be able to locate the correct time, it must be able to compare a stream cut to a position object. This change adds the keyspace information to the position object. This will allow it to identify the segments from the entry in a mark segment that overlap to determine which goes first.

**How to verify it**  
This does not yet add an implementation to read from the `mark` segment and so has no immediate impact.


client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -425,7 +424,7 @@ public void noteTime(long timestamp) {
                                              .stream()
                                              .collect(Collectors.toMap(e -> e.getKey(),
                                                                        e -> e.getValue().getLastObservedWriteOffset()));
-        Position position = new PositionImpl(offsets);
+        WriterPosition position = new WriterPosition(offsets);
```
shiveshr: What is the difference between Position and WriterPosition. 
I dont see range included as part of writerPosition.

shiveshr: Ok, you have included Range information in PositionImpl. 
But that is a super set of WriterPosition. So what is the need for writer's position object? We could have range information available with the writers too via PositionImpl. 
Do we see it diverging from Position object in future. 
Yes, its true that a writer may include multiple segments that overlap with each other. And in such cases it will only have its own assignment of subset of ranges to segments. So should we include that as part of writer position?

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -33,15 +35,31 @@
...
+            SegmentWithRange s = entry.getKey();
+            this.ownedSegments.put(s.getSegment(), entry.getValue());
+            this.segmentRanges.put(s.getSegment(), s.getRange());
+        }
+    }
+    
     @Builder(builderClassName = "PositionBuilder")
-    public PositionImpl(Map<Segment, Long> ownedSegments) {
-        this.ownedSegments = new HashMap<>(ownedSegments);
+    private PositionImpl(Map<Segment, Long> ownedSegments, Map<Segment, Range> segmentRanges) {
```
shiveshr: Should we continue to keep this constructor? 
The position object created from different constructors will either have the range information or not. Which can make it ambiguous for the user of this object. 

Why not keep the range information by default and only have that constructor?

tkaitchuck:
> the changes are looking good in general, but I am unsure of why we need two separate classes for segment without Range and with Range. Since range is an integral part of a segment, we should consolidate the two objects and only have segment class that also has range information.

Looking into that, I found it would require making changes to:
* The client <-> controller wire protocol
* The client <-> SegmentStore wire protocol
* The serialized format of checkpoints in the reader group.
* The serialized format of EventPointers that are given to clients to store. 

And of course changes on the server side to store the range information, which it is currently unaware of.

The problem with this is that either we make these changes and embed the range everywhere the segment goes (The size of eventpointers is already an issue), or we have two classes, or we use a string in the serialized from and immediately invoke an API on the controller to get the range information. 

I'm not anxious to add a call to the controller into the data path, especially if it is not needed. 

Ultimately it just doesn't make sense to store it everywhere on the client, because to the client, range just isn't that important. Until now it was only ever used in a single location (on the append path to select which segment should an event should go to).

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
tkaitchuck: We could include this information. As you note the information is not maintained because of the overlap possibility. So we would have to create a separate map/list on the writer to track the original ranges for segments, then we could take the segmentId and assemble it to a segment object from this mapping it stored from the controller, to create a position object which only lives for a single stack frame as it is immediately passed to the controller API which doesn't want Range information. So we'd then have to write another layer of code to remove it. (Or change the wire protocol to send it anyway).

So that would introduce a fair bit of code and runtime overhead that doesn't really serve any purpose.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
tkaitchuck: There already is only one _public_ constructor which requires range information. This call is private. It is needed for the codegened builder which is in-turn used in the PositionSerializer below (Which is also private). It has to be a builder due to the interface defined by the serialization framework.

shiveshr:
I was Not suggesting that we change the wire protocols. I was merely suggesting that the client's inMemory state - the Segment object (which is not serializable or sent over the wire) to be augmented with the range.
And client really SegmentInfo from controller in three paths - 
1. getCurrentSegments
2. getSegmentsImmediatelyFollowing
3. getSegmentsAtTime (this is for getting initial set of segments) followed by 

`1` and `2` return segmentRange. We can easily update `3` to also return segment range. 

And then client will have segment Range information anytime it wants to instantiate a segment object. 
So client's in memory state can always retain range information. This will only have some memory footprint implication on the client, but that is not much O(num-current-segments) * 128 bytes. 
And client anyway keeps the range information in other data structures. So its not really a memory penalty if they are all consolidated into keeping the range only in segment object.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -33,15 +35,31 @@
...
+        this.ownedSegments = ownedSegments;
+        if (segmentRanges == null) {
+            this.segmentRanges = Collections.emptyMap();
+        } else {
+            this.segmentRanges = segmentRanges;
+        }
     }
 
     static PositionImpl createEmptyPosition() {
```
shrids: this method is not invoked any where and can be removed.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -112,6 +131,35 @@ private void write00(PositionImpl position, RevisionDataOutput revisionDataOutpu
             revisionDataOutput.writeMap(map, (out, s) -> out.writeUTF(s.getScopedName()),
                                         (out, offset) -> out.writeCompactLong(offset));
         }
+        
+        private void read01(RevisionDataInput revisionDataInput, PositionBuilder builder) throws IOException {
+            Map<Segment, Range> map = revisionDataInput.readMap(in -> Segment.fromScopedName(in.readUTF()), PositionSerializer::readRange);
+            builder.segmentRanges(map);
```
shrids: this does not update `ownedSegments ` whereas `public PositionImpl(Map<SegmentWithRange, Long> segments) ` does update `ownedSegments`

tkaitchuck: This is a deserialization method. As it is a revision and not a version the deserialization process is to call read00() and then read01(). ownedSegments is updated in read00()

PR #3881, opened 2019-06-05
Issue 1478: Make getRemainingBytes asynchronous

### LINKED ISSUE #1478: Controller.getRemainingBytes should be asynchronous
**Problem description**
`getRemainingBytes` introduce in https://github.com/pravega/pravega/pull/1473 should run asynchronously in the background so as to not tie up the calling thread.

**Problem location**
ControllerImpl

A thread from the connection thread pool could be used and where possible the Async versions of the calls to the controller could be used instead of immediately calling .get() on all of the futures.

Tristan1900:
**Change log description**  
* Makes getRemainingBytes asynchronous

**Purpose of the change**  
Fixes #1478 

**What the code does**  
`getRemainingBytes` will return a `CompletableFuture<Long>` instead of `long`. 
`getUnreadBytes` will collect and compute results when all the `CompletableFuture<Long>` are finished.  

**How to verify it**  
Test should pass

Signed-off-by: wenqimou <wenqimou@gmail.com>

### DIFF
--- a/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
+++ b/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
@@ -279,7 +279,7 @@ public class ReaderGroupImpl implements ReaderGroup, ReaderGroupMetrics {
             unread = controller.getSegments(fromStreamCut, toStreamCut);
             endPositions = toStreamCut.asImpl().getPositions();
         }
-        return unread.thenApplyAsync(unreadVal -> {
+        return unread.thenApply(unreadVal -> {
             long totalLength = 0;
             for (Segment s : unreadVal.getSegments()) {
                 if (endPositions.containsKey(s)) {
--- a/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
+++ b/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
@@ -246,15 +246,19 @@ public class ReaderGroupImpl implements ReaderGroup, ReaderGroupMetrics {
         }
     }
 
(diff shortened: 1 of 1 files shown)


tkaitchuck:
@andreipaduroiu 
Your comment above is not quite right. The original code was from a (synchronous) public API making N rpc calls to the controller in the calling thread, blocking on each one, before sending the next. Now it's still using the calling thread but it's not waiting for the responses, so there may be N requests outstanding against the controller in parallel and it is then waiting on all of the results. So this should defiantly be an improvement in terms of latency. 

We could go further in two ways:
1. Using a thread pool to actually issue the queries in parallel. (It is not clear how much difference this will make)
2. Make the API call async. (This may be a good idea, but as it is a public API, it would be a breaking change.)

Tristan1900:
@tkaitchuck Thanks for the clarification and @andreipaduroiu thanks for the input. I guess the thread has to block because as Tom said the public API is synchronous. The change does not eliminate the blocking time but to reduce it by waiting for `getRemainingBytes` results all together instead of one by one. 

For the thread pool, each call to `getRemainingBytes` should return quickly with a `CompletableFuture`, so it's probably not necessary to use a thread pool.

andreipaduroiu:
Ok, now that I look closer, I agree with @tkaitchuck. I had to follow through to see how this works.

I have two follow-up questions.
1. So we are essentially using the `ConnectionFactory.getInternalExecutor` to execute multiple concurrent Controller requests (if you follow the `ReaderGroupImpl` constructor call from `ReaderGroupManagerImpl`). All of those calls are likely blocking so each will eat up a thread. Do we have enough threads in that pool to make sure we aren't going to starve something else from making progress if we get too many of these calls at once?
2. What would it take to add an async version of this method like @tkaitchuck indicates? We can modify the internal code to be all async and keep the current sync version that just invokes `get` or `join` on the async version.

tkaitchuck:
> All of those calls are likely blocking so each will eat up a thread.

Those are non-blocking calls.

Issue #15, opened 2016-07-29
Performance should be comparable to Kafka when tuned for equivalent reliability/scale


bayarsaikhan:
To be competitive with existing messaging infrastructures today (like Kafka), we should be able to provide write and read latency (tail and catch up reads) which are comparable or preferably better. 

TBD: Exact numbers, end-to-end versus broker latency...etc

pmullaney:
Was this closed because the comparison was done? Link?

shafqatevo:
Was any benchmark comparisons done between Pravega and Kafka for the same messaging use case?

PR #3887, opened 2019-06-07
Issue 2330: Delegation token expiry

### LINKED ISSUE #2330: Pravega Security: The `delegationToken` should have a reasonable timeout associated with it
**Problem description**
Pravega controller creates a delegationToken with is a JWT token. This token is then passed on to segmentstore. For revocation of privileges to work properly, this token should have a timeout associated with it.
This can be implemented using `exp` claim associated with the JWT token format. The value should be set judiciously so that there are not a lot of expirations of the token as it will slow down the data path. On the other hand, it should be granular enough so that any access revocation is verified as quickly as possible.

**Problem location**
`AuthHelper`, `PravegaInterceptor.java` and `TokenVerifierImpl.java`.

Use a reasonable default value for `exp` claim.

More information about the `exp` claim can be found here https://tools.ietf.org/html/rfc7519#section-4.1.4

ravisharda:
**Change log description**  

Access tokens, which are returned to clients by Controllers and used as delegate-level/delegation tokens by the Segment Store service, do not have expiry assigned to them. This PR introduces a time-to-live value, which is used to determine and assign expiration time to the JWT token used as the delegation token.

**Purpose of the change**  
Resolves #2330. 

**What the code does**  

Here's what the code in this PR does: 

1. It introduces delegation token expiry. 
   * Delegation tokens created at the behest of `ControllerServiceImpl` methods now have an expiration time associated with them. 
   * The delegation token's expiry time is verified in Segment Store request processors `AppendProcessor` and `PravegaRequestprocessor`. If the token is already expired, the request is rejected and an exception is propagated all the way up to the client application. The exception also informs about the specific cause (such as `TokenExpiredException`) through existing Wire command channels. 
   * New exception classes were added to represent token verification related exception conditions. 

2. It adds a new Controller configuration property representing token time-to-live (TTL), which determines what time in the future does the token expires. 
   * If the TTL is 10, the delegation token will have an expiration time 10 seconds from now. 
   * The default TTL is 600. 
   * TTL may be configured as 0, in which case the token will immediately expire. This useful mainly for testing token expiration.
   * TTL may be set as -1 (unlimited), in which case tokens will not have an expiration time. 

3. It refactors the `DelegationTokenVerifier` interface and the `TokenVerifierImpl` implementation to have two ways of verifying tokens: 
    * One that throws a specific exception that tells what went wrong, and 
    * Another one that simply says if the token is valid for the access requested. 

4. It centralizes the creation and parsing of JSON Web Tokens (JWT), in the new `JsonWebToken` class. This change ensures that the code is not repeated. Also, it makes it easier to inspect and test the logic. The PR also adds new unit tests for verifying JWT creation, parsing, and verification. 

5. It also adds a new Integration test class that verifies client and server behavior upon token expiration. 
    In order to be able to use different configurations of the servers in the same integration test class, a new ClusterWrapper class was added in the `test/integration` sub-project. 

6. Previously, the custom claims for resource permissions like `*, READ_UPDATE` overrode the rest of the JWT claims representing the subject, the audience, etc. This PR fixes that issue too as without that fix the logic would have been incorrect. 

7. While trying to verify delegation token expiry using the new integration tests, it was found that in the case of token verification failure, the exception wasn't propagated properly to the client library as the connection was being closed prematurely. Also, a bug in the client library code prevented the exception from being propagated to the client application. These issues were fixed too, in order to ensure proper exception handling. 

8. A new Gradle sub-project `:shared:security` was added to hold classes used across different modules. This was needed as a new class introduced in this PR for JWT creation and parsing is used across Controller and Segment Store code. 

**Note:**
1. Item 3 above is a breaking change for products that use Pravega with a custom token verifier. 
2. The client doesn't automatically reauthenticate and obtain a new delegation token after its expiry. Currently, it only throws an exception back to the application, which allows the application to retry if needed. I believe it is important to support the automatic refresh of tokens upon expiry for easier usability but will leave that up for another day/PR. 

**How to verify it**  
All tests should pass.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+            Set<Map.Entry<String, Object>> claims =
+                    JsonWebToken.fetchClaims(token, tokenSigningKey);
+
+            Optional<Map.Entry<String, Object>> matchingClaim = claims.stream()
+                    .filter(entry -> entryMatchesResource(entry, resource) &&
+                                     expectedLevel.compareTo(AuthHandler.Permissions.valueOf(entry.getValue().toString())) <= 0)
+                    .findFirst();
+
+            if (!matchingClaim.isPresent()) {
+                log.trace(String.format("No matching claim found for resource [%s] and permission [%s] in token [%s].",
```
fpj: `LoggerHelper.trace...`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+
+               /*
+                * Examples of matches:
+                *   - Entry = "abc/, ...", resource = "abc/xyx"
+                */
+               || entry.getKey().endsWith("/") && resource.startsWith(entry.getKey())
+
+               /*
+                * Examples of matches:
+                *    1) Entry = "_system/_requeststream, READ_UPDATE", resource = "_system/_requeststream/0.#epoch.0"
```
fpj: Nit: there is only one element in the list, so it does not make sense to number with `1)`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -346,10 +352,9 @@ private void handleException(UUID writerId, long requestId, String segment, long
             log.warn("Bad attribute update by {} on segment {}.", writerId, segment, u);
             connection.send(new InvalidEventNumber(writerId, requestId, clientReplyStackTrace));
             connection.close();
-        } else if (u instanceof AuthenticationException) {
+        } else if (u instanceof TokenException) {
```
fpj: Here we are replacing `AuthenticationException` with `TokenException`. `TokenException` extends `AuthException`, but I don't the relationship with `AuthenticationException`, could you clarify, please?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -346,10 +352,9 @@ private void handleException(UUID writerId, long requestId, String segment, long
...
             log.warn("Token check failed while being written by {} on segment {}.", writerId, segment, u);
             connection.send(new WireCommands.AuthTokenCheckFailed(requestId, clientReplyStackTrace));
-            connection.close();
```
fpj: Why are we removing the call to close the connection?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -937,10 +941,9 @@ private Void handleException(long requestId, String segment, long offset, String
             log.info(requestId, "Closing connection {} while performing {} due to {}.",
                      connection, operation, u.getMessage());
             connection.close();
-        } else if (u instanceof AuthenticationException) {
-            log.warn(requestId, "Authentication error during '{}'.", operation);
+        } else if (u instanceof TokenException) {
```
fpj: Same question here about the exception type change.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -937,10 +941,9 @@ private Void handleException(long requestId, String segment, long offset, String
...
+            log.warn(requestId, "Token verification failed during '{}'.", operation);
             invokeSafely(connection::send, new AuthTokenCheckFailed(requestId, clientReplyStackTrace), failureHandler);
-            connection.close();
```
fpj: ... and about removing the call to close the connection.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
```diff
@@ -0,0 +1,141 @@
+/**
+ * Copyright (c) 2019 Dell Inc., or its subsidiaries. All Rights Reserved.
```
fpj: Remove copyright year, pls.

shared/security/src/test/java/io/pravega/shared/security/token/JsonWebTokenTest.java
```diff
@@ -0,0 +1,82 @@
+/**
+ * Copyright (c) 2019 Dell Inc., or its subsidiaries. All Rights Reserved.
```
fpj: Copyright year.

test/integration/src/main/java/io/pravega/test/integration/demo/ClusterWrapper.java
```diff
@@ -0,0 +1,198 @@
+/**
+ * Copyright (c) 2019 Dell Inc., or its subsidiaries. All Rights Reserved.
```
fpj: Remove copyright year, pls.

test/integration/src/test/java/io/pravega/test/integration/DelegationTokenTest.java
```diff
@@ -0,0 +1,112 @@
+/**
+ * Copyright (c) 2019 Dell Inc., or its subsidiaries. All Rights Reserved.
```
fpj: Copyright year.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
```diff
@@ -0,0 +1,141 @@
...
+import lombok.NonNull;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Creates Json Web Tokens (JWT). Also, parses a JWT.
+ */
+@ToString
+@Slf4j
+public class JsonWebToken {
```
fpj: I'm not sure why you had to create this class rather than keep using `io.jsonwebtoken.Jwts`. Is my understanding right that we were using it before and that's the reason for removing all `Jwts` builder references?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+import java.util.Set;
+
+import io.pravega.shared.security.token.JsonWebToken;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class TokenVerifierImpl implements DelegationTokenVerifier {
-    private final AutoScalerConfig config;
+
+    private boolean isAuthEnabled;
```
andreipaduroiu: Both of these should be `final`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
-                log.warn("Claim verification failed for resource {} because {}", resource, e);
-                return false;
-            }
-        } else {
+    public boolean isTokenValid(String resource, String token, AuthHandler.Permissions expectedLevel) {
+        try {
+            verifyToken(resource, token, expectedLevel);
             return true;
+        } catch (TokenException e) {
+            log.warn(e.getMessage(), e);
```
andreipaduroiu: You are double-logging the error

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
-                || entry.getKey().equals("*");
+    @Override
+    public void verifyToken(String resource, String token, AuthHandler.Permissions expectedLevel)
+            throws TokenExpiredException, InvalidTokenException, InvalidClaimException, TokenException {
+        if (isAuthEnabled) {
+
+            // All key value pairs inside the payload are returned, including standard fields such as sub (for subject),
+            // aud (for audience), iat, exp, as well as custom fields of the form "<resource> -> <permission>" set by
+            // Pravega.
+            Set<Map.Entry<String, Object>> claims =
```
andreipaduroiu: You really can fit this on one line

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+                    JsonWebToken.fetchClaims(token, tokenSigningKey);
+
+            Optional<Map.Entry<String, Object>> matchingClaim = claims.stream()
+                    .filter(entry -> entryMatchesResource(entry, resource) &&
+                                     expectedLevel.compareTo(AuthHandler.Permissions.valueOf(entry.getValue().toString())) <= 0)
```
andreipaduroiu: Are you doing an ENUM value comparison? This enum does not have statically assigned values. If someone decides to reorder them in the code, your comparison will break. I suggest using a stronger check that would handle such a situation.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+                    .findFirst();
+
+            if (!matchingClaim.isPresent()) {
+                log.trace(String.format("No matching claim found for resource [%s] and permission [%s] in token [%s].",
+                        resource, expectedLevel.toString(), token));
```
andreipaduroiu: No need to call `toString` here.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+
+                throw new InvalidClaimException(String.format(
+                        "No matching claim found for resource: [%s] and permission: [%s] in the delegation token.",
+                        resource, expectedLevel.toString()));
```
andreipaduroiu: or here

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+
+            }
+        }
+    }
+
+    private boolean entryMatchesResource(Map.Entry<String, Object> entry, String resource) {
+        return resource.equals(entry.getKey()) // exact match
```
andreipaduroiu: I find this method too hard to read. Would you mind coalescing all comments into a single block above the statement? You can leave each condition on its own line, but this is a bit too spread out.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,95 @@
...
+
+                throw new InvalidClaimException(String.format(
+                        "No matching claim found for resource: [%s] and permission: [%s] in the delegation token.",
+                        resource, expectedLevel.toString()));
+
+            }
+        }
+    }
+
+    private boolean entryMatchesResource(Map.Entry<String, Object> entry, String resource) {
```
andreipaduroiu: You don't need `Map<String, Object>` here. You can just pass the key as a String.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -152,12 +152,18 @@ public void setupAppend(SetupAppend setupAppend) {
...
-                    "Update Segment Attribute", new AuthenticationException("Token verification failed"));
-            return;
+
+        if (this.tokenVerifier != null) {
+            try {
+                tokenVerifier.verifyToken(newSegment,
+                                          setupAppend.getDelegationToken(),
+                                          AuthHandler.Permissions.READ_UPDATE);
+            } catch (TokenException e) {
+                log.warn(e.getMessage(), e);
```
andreipaduroiu: double log

andreipaduroiu: Actually triple-log. `handleException` will also log this.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -204,12 +204,16 @@ public void readSegment(ReadSegment readSegment) {
...
-        if (!tokenVerifier.verifyToken(segment, delegationToken, READ)) {
-            log.warn(requestId, "Delegation token verification failed.");
-            handleException(requestId, segment, operation, new AuthenticationException("Token verification failed"));
-            return false;
+        boolean isTokenValid = false;
+        try {
+            tokenVerifier.verifyToken(segment, delegationToken, READ);
+            isTokenValid = true;
+        } catch (TokenException e) {
+            log.warn(e.getMessage(), e);
```
andreipaduroiu: same comment here

shared/authplugin/src/main/java/io/pravega/auth/InvalidClaimException.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.auth;
+
+public class InvalidClaimException extends TokenException {
```
andreipaduroiu: Javadoc

shared/authplugin/src/main/java/io/pravega/auth/InvalidTokenException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.auth;
+
+public class InvalidTokenException extends TokenException {
```
andreipaduroiu: Javadoc

shared/authplugin/src/main/java/io/pravega/auth/TokenException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.auth;
+
+public class TokenException extends AuthException {
```
andreipaduroiu: Javadoc

shared/authplugin/src/main/java/io/pravega/auth/TokenExpiredException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.auth;
+
+public class TokenExpiredException extends TokenException {
```
andreipaduroiu: Javadoc

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
andreipaduroiu: This class is not thread safe. Either make all fields final or use some synchronization method to resolve this.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private Map<String, Object> permissionsByResource;
+    private SignatureAlgorithm signatureAlgorithm = SignatureAlgorithm.HS512;
+
+    public JsonWebToken(String subject, String audience, byte[] signingKey) {
+        this(subject, audience, signingKey, Optional.empty(), null);
+    }
+
+    public JsonWebToken(@NonNull String subject, @NonNull String audience, @NonNull byte[] signingKey,
+                        @NonNull Optional<Integer> timeToLiveInSeconds,
```
andreipaduroiu: Why would an `Optional` ever be null? Please try not to use `Optional` and use the wrapped type (`Integer`). It will make your code less prone to bugs.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
```diff
@@ -0,0 +1,141 @@
...
+            // timetoLiveInSeconds = -1 implies that the token never expires.
+            // timeToLiveInSeconds = 0 implies token immediately expires.
+            Preconditions.checkArgument(timeToLiveInSeconds.get() >= -1);
+        }
+
+        this.subject = subject;
+        this.audience = audience;
+        this.signingKey = signingKey.clone();
+
+        this.currentInstant = Instant.now();
```
andreipaduroiu: Would it be better to use `System.currentTimeMillis`?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/GRPCServerConfig.java
```diff
@@ -87,6 +87,14 @@
      */
     String getTokenSigningKey();
 
+    /**
+     * Returns the delegation token time-to-live (TTL) value. The TTL in turn determines the time at which the
+     * delegation token expires.
+     *
+     * @return time-to-live value of the access token in seconds.
+     */
+    Optional<Integer> getAccessTokenTtlInSeconds();
```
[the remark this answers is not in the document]
ravisharda: Sure, have now changed it to `TTL` (capital letters) as suggested.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/DelegationTokenVerifier.java
```diff
@@ -10,17 +10,42 @@
...
+     * @param expectedLevel           maximum expected access to the given {@code resource}.
+     * @throws TokenExpiredException  if the {@code token} has expired
+     * @throws InvalidTokenException  if the {@code token} is invalid
+     * @throws InvalidClaimException  if the {@code token} does not contain the claim representing
+     *                                {@code expectedLevel} of access
+     * @throws TokenException         if any other failure condition is encountered
      */
-    boolean verifyToken(String resource, String token, AuthHandler.Permissions expectedLevel);
-}
+    void verifyToken(String resource, String token, AuthHandler.Permissions expectedLevel)
```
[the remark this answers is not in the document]
ravisharda: On second thought, we don't really need the `isTokenValid` method in the interface as it is not being used anywhere currently. So, I've removed it.

shared/authplugin/src/main/java/io/pravega/auth/TokenException.java
```diff
@@ -0,0 +1,23 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    public TokenException(String message) {
+        super(message, 401);
```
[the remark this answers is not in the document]
ravisharda: `TokenException` extends from an existing exception `AuthException`, which encapsulates HTTP response status code apart from other relevant information about the occurrence. The constructors of `AuthException` require supplying the HTTP response status code, and therefore we use one in the TokenException. 

Among the response codes, the appropriate choices were `401` and `403`. `401` seems to be the most appropriate one to me.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: :) I agree. I have modified this part to collate all the comments now.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: The newly introduced TokenException can occur either while generating the token or when verifying it. In the case of Segment Store where this code lies, this exception can occur under conditions such as: 

* The token has expired (represented by `TokenExpiredException`, a subclass of `TokenException`)
* The token is invalid (represented by `InvalidTokenException`, a subclass of `TokenException`)
* The claim in the token doesn't match the authorization claim (represented by `InvalidClaimException`, a subclass of TokenException). 

The conditions that may lead to a `TokenException` include both authentication and authorization failures (apart from failure in token validation). `AuthenticationException` did not represent all those conditions, so the new `TokenException` and its subclasses were introduced rooted in `AuthException`. Technically, I could have used `AuthException` as well, but a `TokenException` is more representative of token-related failures. 

The segment store deals with delegation token verification, and therefore its `AppendProcessor` and `PravegaResourceProcessor` now throws TokenException back to the clients. 

Hope that helps.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: Earlier, the connection was closed here, due to which the control was never reaching the `io.pravega.client.segment.impl.ResponseProcessor.authTokenCheckFailed(...)` method on the client side upon token-related failures on the segment store, which was a bug. The control was reaching the `ResponseProcessor.connectionDropped(...)` instead. As a result, upon failures in token verification, the client would hang due to subsequent retries and the corresponding exception would not be propagated to the client. 

I encountered this problem when running a new test `io.pravega.test.integration.DelegationTokenTest.testWriteFailsWhenTokenExpires(...)`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: Sure, done.

shared/authplugin/src/main/java/io/pravega/auth/TokenExpiredException.java
ravisharda: I've added Java doc comments, as suggested.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
ravisharda: I've made all the fields final. Thanks.

shared/authplugin/src/main/java/io/pravega/auth/TokenException.java
ravisharda: Done.

test/integration/src/main/java/io/pravega/test/integration/demo/ClusterWrapper.java
ravisharda: Done.

shared/security/src/test/java/io/pravega/shared/security/token/JsonWebTokenTest.java
ravisharda: Done.

test/integration/src/test/java/io/pravega/test/integration/DelegationTokenTest.java
ravisharda: Done.

build.gradle
```diff
@@ -114,6 +114,7 @@ project('common') {
         testCompile project(':test:testcommon')
         compile group: 'commons-io', name: 'commons-io', version: commonsioVersion
         compile group: 'com.google.guava', name: 'guava', version: guavaVersion
+        compile group: 'io.jsonwebtoken', name: 'jjwt', version: jjwtVersion
```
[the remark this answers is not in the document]
ravisharda: It was an oversight. We actually don't need to add jjwt dependency to `common` now. While I refactored the JJWT related code into `:shared:security`, I had missed removing this dependency. I've removed this dependency now.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHelper.java
```diff
@@ -67,17 +74,17 @@ public String checkAuthorizationAndCreateToken(String resource, AuthHandler.Perm
...
-
             claims.put(resource, String.valueOf(expectedLevel));
 
-            return Jwts.builder()
-                       .setSubject("segmentstoreresource")
-                       .setAudience("segmentstore")
-                       .setClaims(claims)
-                       .signWith(SignatureAlgorithm.HS512, tokenSigningKey.getBytes())
-                       .compact();
+            return new JsonWebToken("segmentstoreresource",
```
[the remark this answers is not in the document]
ravisharda: A majority of the parameters of `JsonWebToken` are mandatory, so using a builder here would still require the calling code to specify most of them upfront when invoking the builder (as shown below. Also, there is no use case for a step-by-step construction as of now. So, I preferred the (arguably) simpler telescopic constructor pattern instead of the builder, here. 

The following two examples show how the calling code in current use'd look like if we were to use a builder instead:

```java
JsonWebToken.builder("subject", "audience", signingKey, ttl)
     .resourcePermissionClaims(claims)
     .build() 
```
or 
```java
JsonWebToken.builder("subject", "audience", signingKey).build() 
```
What do you say?

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
ravisharda: Done.

controller/src/main/java/io/pravega/controller/server/rpc/auth/PravegaInterceptor.java
```diff
@@ -88,15 +89,10 @@
...
+        Map<String, Object> customClaims = new HashMap<>();
+        customClaims.put("*", String.valueOf(READ_UPDATE));
 
-        return Jwts.builder()
-                   .setSubject("segmentstoreresource")
-                   .setAudience("segmentstore")
-                   .setClaims(claims)
-                   .signWith(SignatureAlgorithm.HS512, tokenSigningKey.getBytes())
-                   .compact();
+        return new JsonWebToken("segmentstoreresource", "segmentstore", tokenSigningKey.getBytes(),
```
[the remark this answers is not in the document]
ravisharda: Pasting the link of a reply for a similar comment earlier: 
https://github.com/pravega/pravega/pull/3887#discussion_r294138979

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -48,6 +50,11 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
         if (publishedRPCPort != null) {
             Preconditions.checkArgument(publishedRPCPort > 0, "publishedRPCPort should be a positive integer");
         }
+        if (accessTokenTtlInSeconds != null) {
```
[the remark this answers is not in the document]
ravisharda: This is preexisting code: I didn't modify it for this PR. 

I guess the reason for this check might have been that publishedRPCPort is allowed to be null, and therefore the precondition check is only done if it is not null.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -83,6 +91,8 @@ public String toString() {
                         Strings.isNullOrEmpty(userPasswordFile) ? "unspecified" : "specified"))
                 .append(String.format("tokenSigningKey is %s, ",
                         Strings.isNullOrEmpty(tokenSigningKey) ? "unspecified" : "specified"))
+                .append(accessTokenTtlInSeconds.isPresent() ?
```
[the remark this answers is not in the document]
ravisharda: Yes, that is right. 

At a system level, if no configuration has been specified, the default configuration value of 600 seconds is applicable: when the control comes here, the object's property is already present and its value is 600. 

(https://github.com/pravega/pravega/blob/3189baaf26529b6930be9f1e66b67953d1f9c081/controller/src/main/java/io/pravega/controller/util/Config.java#L152)

On the other hand, an object of this class may also be created directly (for testing purposes), in which case this property may not have been set to a non-null value.

gradle.properties
```diff
@@ -51,6 +51,7 @@ swaggerJersey2JaxrsVersion=1.5.16
 slf4jApiVersion=1.7.25
 gradleGitPluginVersion=2.2.0
 k8ClientVersion=3.0.0
+jjwtVersion=0.9.0
```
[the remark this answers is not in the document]
ravisharda: Yes, that is right. We use it a reference in the definition for `shared:security` Gradle sub-project. I had to add this here, as it wasn't already added. The existing definitions in `build.gradle` specify the version directly in the `build.gradle` file (`compile 'io.jsonwebtoken:jjwt:0.9.0'`)

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
ravisharda: `io.pravega.shared.security.token.JsonWebToken` is a new class created in this PR for creating and parsing a JSON Web Token (JWT). This class internally uses `io.jsonwebtoken.Jwts` and encapsulates Pravega-specific logic for processing tokens. 

To explain the rationale for making this change a bit more, previously, the logic was spread and even duplicated in multiple places. It was hard to get a full picture of how Pravega creates and parses token and was more difficult than necessary to make any changes. Now, all JWT token-related domain logic is in one class. The change also made unit testing easier, and a bug was found in existing code and fixed. 

All other code in Pravega now uses this class for processing tokens, and therefore does not need any references to `io.jsonwebtoken.Jwts`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: Fixed.

ravisharda: Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: Thanks for pointing that out. When I relooked at it, it did seem unnecessarily complex. I have made the change you have suggested (among other things).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: I've made the change as suggested. It looks much better now.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHelper.java
```diff
@@ -10,25 +10,32 @@
...
 
 /**
  * Helper class containing APIs related to delegation token and authorization/authentication.
  */
 @AllArgsConstructor
 public class AuthHelper {
+
     private final boolean isAuthEnabled;
     private final String tokenSigningKey;
+    private final Optional<Integer> accessTokenTtlInSeconds;
```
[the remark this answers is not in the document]
ravisharda: I actually started out with using the `Integer` object here, as well as other places where the TTL is being defined/used, but realized that code wasn't revealing its intent well. Also, the null checks were confusing. 

I then modified the code to use Optional to clearly signal that a missing value is permitted and not depending on a value being null. It also allows for disambiguating between these scenarios (as shown in `JsonWebToken`): 
* An invalid argument null for the `Optional<Integer>` TTL parameter
* A valid input containing empty instance for the same (say, `Optional.empty()`)
* A valid input containing a non-empty instance for the same (say, `Optional.of(200)`)

The extra constructor was a remnant from my intermediate work using `Integer`, and was merely to avoid having to modify some of the tests when I went from using `Integer` to `Optional<Integer>`; I've fixed the tests now instead and removed the redundant constructor.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
ravisharda: Yes, if someone passes a null argument (and not Optional.empty() or similar ways of specifying empty Optional instance). I've explained my rationale for using Optional in another comment. Also, I have modified the code a bit. Please take a relook and share your thoughts.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
ravisharda: Using the `Instant` object reveals the intent better, both here and later in this class where we transform it. Using `System.currentTimeMillis` requires at least a bit of mental mapping.  

Therefore, I preferred the Instant here.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -283,11 +284,13 @@ private Throwable getException() {
     private final class ResponseProcessor extends FailingReplyProcessor {
         @Override
         public void connectionDropped() {
+            log.trace("connectionDropped");
```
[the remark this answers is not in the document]
ravisharda: On second thought, this log message should have been at info level like is the case some of the other existing ResponseProcessor methods. The latest commit makes that change.

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -988,7 +988,7 @@ public void close() {
         return result.thenApply( token -> token.getDelegationToken())
         .whenComplete((x, e) -> {
             if (e != null) {
-                log.warn("getCurrentSegments failed: ", e);
+                log.warn("getCurrentSegments failed: ", e.getMessage(), e);
```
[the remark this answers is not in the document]
ravisharda: Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/DelegationTokenVerifier.java
```diff
@@ -10,17 +10,42 @@
...
+    /**
+     * Determines whether the given {@code token} is valid for the specified {@code expectedLevel} of access on the
+     * given {@code resource}.
+     *
+     * @param resource       the resource for which access is desired.
+     * @param token          the access/delegation token.
+     * @param expectedLevel  maximum expected access to the given {@code resource}.
+     * @return               {@code true} if token is valid for accessing the resource, otherwise {@code false}.
+     */
+    boolean isTokenValid(String resource, String token, AuthHandler.Permissions expectedLevel);
```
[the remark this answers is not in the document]
ravisharda: I'd think we'd leave the interface cleaner if we don't keep the previous method. There doesn't seem to be any practical compatibility issues as of now.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHelper.java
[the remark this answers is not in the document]
andreipaduroiu: `Optional<>` as a method argument is an anti-pattern.

Please refer to this doc as to what the purpose of `Optional<>` is: https://community.oracle.com/docs/DOC-991686

A simple google search for `best practices with java optional` will show a good number of other documents and blog posts explaining that `Optional<>` as a method arg is a bad idea.

Consider the very good example listed as #15. here: https://www.baeldung.com/java-optional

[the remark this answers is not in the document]
ravisharda: Thanks for pointing that out. I didn't realize that there are strong opinions in the community against using Optional as method arguments. There are a number of arguments laid out against using it as method arguments in [this](https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments) discussion. There are also arguments made for using it by some ([example](https://dzone.com/articles/optional-method-parameters)). As for this case, I don't have a strong opinion for or against it - it was merely about a slightly better-looking code. I'll use the `Integer` instead.

controller/src/test/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImplTest.java
```diff
@@ -33,7 +33,16 @@ public void toStringReturnsSuccessfullyWithAllConfigSpecified() {
...
+                "", " ", "secret", 300, "/cert.pem",
+                false, true);
+        assertNotNull(config.toString());
+    }
+
+    @Test
+    public void toStringReturnsSuccessfullyWithAccessTokenTtlNull() {
+        GRPCServerConfigImpl config = new GRPCServerConfigImpl(9090, null,
+                9090, true, null, true,
+                "", " ", "secret", null, "/cert.pem",
```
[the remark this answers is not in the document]
ravisharda: Fixed it by using the builder instead.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: I have changed it to debug level as that seems to be more appropriate here. Hope that is OK.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHelper.java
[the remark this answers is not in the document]
ravisharda: I've pushed a commit with the change, so resolving the conversation now.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: The enum, as well as the comparison made here, is pre-existing code. I am thinking it is prudent to do this change separately in a different PR for the following reasons:

1. There might be unexpected repercussions. We also support AuthHandler plugins that use the Enum.
2. Plus there are other pending changes that were discussed in other issues/PRs that require changing the Permissions enum. 

What do you say?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
ravisharda: Fixed.

shared/authplugin/src/main/java/io/pravega/auth/InvalidTokenException.java
ravisharda: Done.

shared/authplugin/src/main/java/io/pravega/auth/InvalidClaimException.java
ravisharda: Done.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -31,14 +31,16 @@
     private final String tlsCertFile;
     private final String tlsKeyFile;
     private final String tokenSigningKey;
+    private final Optional<Integer> accessTokenTtlInSeconds;
```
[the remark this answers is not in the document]
ravisharda: Done.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
ravisharda: Please refer to the response for a similar question earlier: https://github.com/pravega/pravega/pull/3887#discussion_r294134259

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
ravisharda: Please refer to the following response to a similar question: 
https://github.com/pravega/pravega/pull/3887#discussion_r294135258

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHelper.java
```diff
@@ -10,25 +10,27 @@
...
 public class AuthHelper {
+
     private final boolean isAuthEnabled;
     private final String tokenSigningKey;
+    private final Integer accessTokenTTLInSeconds;
 
     @VisibleForTesting
     public static AuthHelper getDisabledAuthHelper() {
-        return new AuthHelper(false, "");
+        return new AuthHelper(false, "", -1);
```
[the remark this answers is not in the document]
ravisharda: The token TTL (`accessTokenTTLInSeconds` instance variable) that this value is used to set, uses -1 as an indicator that the token never expires. It disallows any value lower than -1. Just for completeness, it also supports 0 as TTL value (apart from a positive integer), which specifies that the token immediately expires. Both -1 and 0 are useful only for testing purposes.  In practice, at a system level, if the value is not specified in the Controller config, the default TTL value (of 600) will apply.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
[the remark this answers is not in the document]
fpj: This is not info level, this is at best debug level. Trace seems appropriate as we are tracing the execution of a specific operation.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -408,13 +412,14 @@ private void checkAckLevels(long ackLevel, long previousAckLevel) {
...
         public void processingFailure(Exception error) {
+            log.info("Received processingFailure with {}", error);
             failConnection(error);
         }
 
         @Override
         public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenCheckFailed) {
-            log.warn("Auth failed {}", authTokenCheckFailed);
-            failConnection(new AuthenticationException(authTokenCheckFailed.toString()));
+            log.info("Received authTokenCheckFailed with {}", authTokenCheckFailed);
```
fpj: Why is this info and not warn? If it is an error, but it is a recoverable error, then it should be warn level.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -283,11 +284,13 @@ private Throwable getException() {
     private final class ResponseProcessor extends FailingReplyProcessor {
         @Override
         public void connectionDropped() {
+            log.info("Received connectionDropped for writer {}", writerId);
```
fpj: Is this info or warn? In general, errors should be either `warn` or `error` level, depending on whether it is recoverable or not. In this case, it is not crystal clear because the error is dealt with automatically. I'd still prefer to make it warn, though. Same for the severity level for the other log messages changing in this file.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -48,6 +50,11 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
         if (publishedRPCPort != null) {
             Preconditions.checkArgument(publishedRPCPort > 0, "publishedRPCPort should be a positive integer");
         }
+        if (accessTokenTTLInSeconds != null) {
```
shrids: Did not understand when this can be null , given that the default value is 600?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -9,55 +9,91 @@
...
+    private boolean resourceMatchesClaimKey(String claimKey, String resource) {
+        /*
+         * Examples of the conditions when the claimKey (key of the key-value pair claim) matches the resource are:
+         *      1) claimKey = "myscope", resource = "myscope"
+         *      2) claimKey = "abc/", resource = "abc/xyx"
+         *      3) claimKey = "_system/_requeststream", resource = "_system/_requeststream/0.#epoch.0"
+         *      4) claimKey = "*" (the wildcard character)
+         */
+        return resource.equals(claimKey) // example 1
+               || claimKey.endsWith("/") && resource.startsWith(claimKey) // example 2
```
shrids: This implies claimKey "abc" and "abc/" are essentially the same. If so we can simplify the logic here.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -205,12 +205,15 @@ public void readSegment(ReadSegment readSegment) {
     }
 
     private boolean verifyToken(String segment, long requestId, String delegationToken, String operation) {
-        if (!tokenVerifier.verifyToken(segment, delegationToken, READ)) {
-            log.warn(requestId, "Delegation token verification failed.");
-            handleException(requestId, segment, operation, new AuthenticationException("Token verification failed"));
-            return false;
+        boolean isTokenValid = false;
```
shrids: nit: given that the isTokenValid is set to false only incase of `TokenException` we could as well make the default value of isTokenValid to true

ravisharda: The idea here is to be restrictive by default, and be permissive only if the conditions for token validity are positively met. So, in this case, the token will be deemed invalid even if `tokenVerifier.verifyToken(...)` throws any other exception other than the one that is being caught, or `handleException(...)` itself throws an exception.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
ravisharda: At a system-level, there is a default value in Controller configuration: when this class is instantiated the caller would set it to a non-null value. 

But, when you look at this class independently, there is no default value set for the variable, so it can indeed be null. If the value is non-null, we check for valid values, which in this case is any case is greater than or equals to -1. An example of when this value can be null is when this class is invoked directly for unit or integration tests.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/GRPCServerConfig.java
```diff
@@ -87,6 +87,14 @@
      */
     String getTokenSigningKey();
 
+    /**
+     * Returns the delegation token time-to-live (TTL) value. The TTL in turn determines the time at which the
+     * delegation token expires.
+     *
+     * @return time-to-live value of the access token in seconds.
+     */
+    Integer getAccessTokenTTLInSeconds();
```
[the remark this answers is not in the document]
ravisharda: The reason I used the `Integer` wrapper instead of the primitive `int` to allow for nulls to indicate missing configuration.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: I completely agree. This logic needs a revamp. Since it is pre-existing logic and the change needed is not directly relevant for the issue that this PR fixes, I've created a new issue for the same: https://github.com/pravega/pravega/issues/3927.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
ravisharda: When looking to make the suggested change, I realized that the `failConnection()` method was already logging the exception. So, I removed the log entries from the methods that invoke `failConnection` in order to avoid redundant logging. The `failConnection()` was logging at an info level, which I have now modified to `warn`. Hope that helps.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
ravisharda: Sure, this has changed now, as described in this comment: https://github.com/pravega/pravega/pull/3887#discussion_r295140458.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
ravisharda: There was an unnecessary statement to set the flag to false in the subsequent catch block, which I've now removed.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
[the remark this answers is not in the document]
ravisharda: This log statement was removed in the latest push as described in this comment: https://github.com/pravega/pravega/pull/3887#discussion_r295140458. This change is therefore not needed anymore.

Issue #3893, opened 2019-06-10
Have logging on thread death


tkaitchuck:
**Problem description**
In the event that a thread dies java will by default log the exception to standard error. Unfortunately this makes it hard to locate as it will not be included with the rest of the logs.

**Problem location**
Common

**Suggestions for an improvement**
Add an uncaught exception handler.

shrids:
This has been closed as part of PR #3894

Issue #3892, opened 2019-06-10
Do not suppress exceptions on a failed system test.


tkaitchuck:
**Problem description**
Currently when a system test throws an error, on several tests it goes into a finally block which makes additional assertions. If any of these fire the original cause of the failure is lost and the test fails with the assertion from the finally block.

**Problem location**
Several system tests.

**Suggestions for an improvement**
Do not make assertions in a finally block.

shrids:
Closed as part of PR #3894

Issue #1050, opened 2017-04-14
Improve logging for integration tests


tkaitchuck:
**Problem description**
Looking at RemoteSequential when a job fails, it does not include the exception that caused it. This is the most important information for debugging the problem.

**Problem location**
RemoteSequential.java:59

**Suggestions for an improvement**
 As exceptions are serializable this could be easily attached to the History object or elsewhere in the Job object. Then the failure could be attached to the exception thrown as a cause.
Additionally we could add an environmental variable for "MasterIP" so that the log line could be used to locate more detailed logs without having to look that information up.

PR #3894, opened 2019-06-10
Issue 3893: System test logging

tkaitchuck:
**Change log description**  
* Make sure tests are not suppressing exceptions
* Improve logging when threads die or get stuck

**Purpose of the change**  
Fixes #3894, #3893, #3892, and #1050

**What the code does**  
No functional changes to code outside of tests.


common/src/main/java/io/pravega/common/concurrent/ExecutorServiceHelpers.java
```diff
@@ -56,6 +72,7 @@ public static ThreadFactory getThreadFactory(String groupName) {
             @Override
             public Thread newThread(Runnable r) {
                 Thread thread = new Thread(r, groupName + "-" + threadCount.incrementAndGet());
+                thread.setUncaughtExceptionHandler(new LogUncaughtExceptions());
```
shrids: We could override `java.util.concurrent.ThreadPoolExecutor#afterExecute` of the `ScheduledExecutorService ` to log the exceptions. This is a guaranteed way of logging exceptions when using  `ThreadPoolExecutor`s.

test/system/src/main/java/io/pravega/test/system/framework/RemoteSequential.java
```diff
@@ -60,6 +60,8 @@
             if (response.status() != CREATED.code()) {
                 throw new TestFrameworkException(TestFrameworkException.Type.ConnectionFailed, "Error while starting " +
                         "test " + testMethod);
+            } else {
```
shrids: this code is executed only for tests running on a mesos setup.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceHelpers.java
tkaitchuck: We could. I'm not sure it's better because it would be in the loop on every execution (and require subclassing) as opposed to just when an exception is thrown that is not being delivered to a future and would otherwise kill the thread.

test/system/src/main/java/io/pravega/test/system/framework/RemoteSequential.java
tkaitchuck: Ack

test/system/src/main/java/io/pravega/test/system/SingleJUnitTestRunner.java
```diff
@@ -43,7 +43,19 @@ public void runMethod() {
         Method m = null;
         try {
             m = this.testClass.getDeclaredMethod(this.methodName);
-            Statement statement = methodBlock(new FrameworkMethod(m));
+            Statement statement = methodBlock(new FrameworkMethod(m) {
+                @Override
+                public Object invokeExplosively(final Object target, final Object... params) throws Throwable {
```
shrids: +1

PR #3901, opened 2019-06-12
Issue 3425: Ensure the Sealed segment writer is removed after all inflight events are resent to the newer segment writers.

### LINKED ISSUE #3425: EventStreamWriter#flush does not block in case an inflight event is being written to a SealedSegment
**Problem description**
This issue occurs when we have a `flush` and `seal` happening at the same time. The `flush` invokes `selector.getSegments()` which is synchronized and the scaling retransmit worker thread calls `selector.updateSegmentsUponSealed()` which is also synchronized. However scaling retransmit worker thread releases the lock upon return of the method before the events are added to the inflight on the new segment.

**Problem location**
EventStreamWriter.

Fix the issue.

shrids:
**Change log description**  
Ensure `SegmentOutputStream#flush` of a sealed segment throws a `SegmentSealedException` in the case of a zero inflight events.

**Purpose of the change**  
Fixes #3425

**What the code does**  
`EventStreamWriter.flush()` invokes a `flush()` on all the Segment writers. Now if any of the segment writers threw a `SegmentSealedException` during a `SegmentOutputStream.flush()` the event stream writer will fetch the latest segment writers again and perform a flush over each of them, thereby ensuring all the inflight events are written/durably persisted.

For this logic to work we need to remove the older writer only after we have resent the inflight events to the successor segments. This was not the case earlier and could cause a flush to prematurely stop blocking before the events were durably persisted by the newer segment writers.

**How to verify it**  

### DIFF
--- a/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
@@ -509,13 +509,13 @@ class SegmentOutputStreamImpl implements SegmentOutputStream {
             }
             state.waitForInflight();
             Exceptions.checkNotClosed(state.isClosed(), this);
-        }
-        /* SegmentSealedException is thrown if either of the below conditions are true
+            /* SegmentSealedException is thrown if either of the below conditions are true
                  - resendToSuccessorsCallback has been invoked.
                  - the segment corresponds to an aborted Transaction.
-         */
-        if (state.needSuccessors.get() || (StreamSegmentNameUtils.isTransactionSegment(segmentName) && state.isAlreadySealed())) {
-            throw new SegmentSealedException(segmentName + " sealed for writer " + writerId);
+             */
+            if (state.needSuccessors.get() || (StreamSegmentNameUtils.isTransactionSegment(segmentName) && state.isAlreadySealed())) {
+                throw new SegmentSealedException(segmentName + " sealed for writer " + writerId);
+            }
         }
     }
     
--- a/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
@@ -499,7 +499,7 @@ class SegmentOutputStreamImpl implements SegmentOutputStream {
             } catch (SegmentSealedException | NoSuchSegmentException e) {
                 if (StreamSegmentNameUtils.isTransactionSegment(segmentName)) {
                     log.warn("Exception observed during a flush on a transaction segment, this indicates that the transaction is " +
-                                     "commited/aborted. Details: {}", e.getMessage());
+                                     "committed/aborted. Details: {}", e.getMessage());
                     failConnection(e);
                 } else {
                     log.info("Exception observed while obtaining connection during flush. Details: {} ", e.getMessage());
@@ -509,13 +509,13 @@ class SegmentOutputStreamImpl implements SegmentOutputStream {
             }
             state.waitForInflight();
             Exceptions.checkNotClosed(state.isClosed(), this);
-            /* SegmentSealedException is thrown if either of the below conditions are true
+        }
+        /* SegmentSealedException is thrown if either of the below conditions are true
                  - resendToSuccessorsCallback has been invoked.
                  - the segment corresponds to an aborted Transaction.
-             */
-            if (state.needSuccessors.get() || (StreamSegmentNameUtils.isTransactionSegment(segmentName) && state.isAlreadySealed())) {
-                throw new SegmentSealedException(segmentName + " sealed for writer " + writerId);
-            }
+         */
+        if (state.needSuccessors.get() || (StreamSegmentNameUtils.isTransactionSegment(segmentName) && state.isAlreadySealed())) {
+            throw new SegmentSealedException(segmentName + " sealed for writer " + writerId);
         }
     }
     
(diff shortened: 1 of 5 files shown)


client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
```diff
@@ -173,16 +173,15 @@ public void testConnectAndFailedSetupAppendDueToTruncation() throws Exception {
         MockController controller = new MockController(uri.getEndpoint(), uri.getPort(), cf, true);
         ClientConnection connection = mock(ClientConnection.class);
         cf.provideConnection(uri, connection);
-        @Cleanup
```
andreipaduroiu: Why this?

This object will never be cleaned up after the test.

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
```diff
@@ -631,6 +630,25 @@ public void testSealedBeforeFlush() throws Exception {
...
+    public void testFlushForEmptyInflightSealedSegment() throws Exception {
+        UUID cid = UUID.randomUUID();
+        PravegaNodeUri uri = new PravegaNodeUri("endpoint", SERVICE_PORT);
+        MockConnectionFactoryImpl cf = new MockConnectionFactoryImpl();
+        cf.setExecutor(executorService());
+        MockController controller = new MockController(uri.getEndpoint(), uri.getPort(), cf, true);
+        ClientConnection connection = mock(ClientConnection.class);
+        cf.provideConnection(uri, connection);
+        InOrder order = Mockito.inOrder(connection);
+        SegmentOutputStreamImpl output = new SegmentOutputStreamImpl(SEGMENT, controller, cf, cid, segmentSealedCallback, RETRY_SCHEDULE, "");
```
andreipaduroiu: Here too.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -509,13 +509,13 @@ public void flush() throws SegmentSealedException {
             }
             state.waitForInflight();
             Exceptions.checkNotClosed(state.isClosed(), this);
-            /* SegmentSealedException is thrown if either of the below conditions are true
+        }
```
tkaitchuck: I'm not sure I understand the need for this change. Is there any case this prevents a problem?

client/src/main/java/io/pravega/client/stream/impl/SegmentSelector.java
```diff
@@ -142,7 +151,7 @@ public Segment getSegmentForEvent(String routingKey) {
         currentSegments = currentSegments.withReplacementRange(sealedSegment, successors);
         createMissingWriters(segmentSealedCallback, currentSegments.getDelegationToken());
         log.debug("Fetch unacked events for segment: {}, and adding new segments {}", sealedSegment, currentSegments);
-        return writers.remove(sealedSegment).getUnackedEventsOnSeal();
+        return writers.get(sealedSegment).getUnackedEventsOnSeal();
```
tkaitchuck: Please add a javadoc to this method explaining the impl and that remove needs to be called later.

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
shrids: removed this change.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: Yes this change is not required since we remove the segment writers only after the inflight events are resent to the new segments.

client/src/main/java/io/pravega/client/stream/impl/SegmentSelector.java
shrids: done.

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
shrids: removed this change.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: Yes, it is not required.

client/src/test/java/io/pravega/client/stream/impl/SegmentSelectorTest.java
```diff
@@ -217,4 +222,66 @@ public void testControllerNotReachable() {
...
+        StreamSegmentsWithPredecessors segmentsWithPredecessors = new StreamSegmentsWithPredecessors(newRange, "");
+
+        // Setup Mock.
+        SegmentOutputStream s0Writer = Mockito.mock(SegmentOutputStream.class);
+        SegmentOutputStream s1Writer = Mockito.mock(SegmentOutputStream.class);
+        SegmentOutputStream s2Writer = Mockito.mock(SegmentOutputStream.class);
+        SegmentOutputStreamFactory factory = Mockito.mock(SegmentOutputStreamFactory.class);
+        Controller controller = Mockito.mock(Controller.class);
+
+        when(s0Writer.getUnackedEventsOnSeal()).thenReturn(ImmutableList.of(pendingEvent));
```
fpj: Based on the observations of #3885 , should we be using `doAnswer` instead?

shrids: The reason for the issue in #3885 is because are trying to stub operations on a `org.mockito.Mockito#spy(T)` of a real object. Hence doing a `when(spy.methodA())).thenReturn("foo");` would imply invoking methodA() while setting up the stubbed return value. 
In this case `s0Writer` is a mock object and hence we can use `when(...).thenReturn(...)` pattern.

PR #3911, opened 2019-06-13
Issue 3904: Fix PravegaTableStreamMetadataStore's garbage collection for old batch gc

### LINKED ISSUE #3904: Controller's PravegaTableStreamMetadataStore implementation completed txn garbage collection does not retain latest two batches
**Problem description**
PravegaTableStreamMetadataStore's garbage collection logic has a bug where instead of retaining latest two, it retains two latest in the key set order as returned by pravega tables.getAllKeys. 
This can prematurely purge transaction's completed record whereby applications trying to retrieve its status may get Unknown transaction. 

Note: ZkStreamMetadataStore implementation sorted the list of batches.

**Problem location**
PravegaTableStreamMetadataStore.java

Retain the latest two batches.

shiveshr:

**Change log description**  
PravegaTableStreamMetadataStore's garbage collection had a bug where batches were not sorted by batch number and this resulted in any 2 batches being retained while everything else gc'd.

**Purpose of the change**  
Fixes #3904 

**What the code does**  
Sorts the list of batches as received from the store and then retains the latest two while purging all older batches. 

1. Fixed the issue in PravegaTableStreamMetadataStore by sorting the list of batches as received from getAllKeys on Batches table. 
2. Added a `VisibleForTesting` setter in PravegaTableStreamMetadataStore to set garbageCollector to mock new batches at the time of transaction commits. 
3. Added a new `VisibleForTesting` package level method in PravegaTableStream to get batch table name. 

**How to verify it**  
Unit test added.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
@@ -29,7 +29,6 @@ import org.apache.curator.framework.CuratorFramework;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
@@ -87,19 +86,7 @@ public class PravegaTablesStreamMetadataStore extends AbstractStreamMetadataStor
         List<String> batches = new ArrayList<>();
         return withCompletion(storeHelper.expectingDataNotFound(storeHelper.getAllKeys(COMPLETED_TRANSACTIONS_BATCHES_TABLE)
                                                          .collectRemaining(batches::add)
-                                                         .thenApply(v -> {
-                                                                     // retain latest two and delete remainder.
-                                                                     if (batches.size() > 2) {
-                                                                         List<String> list = batches
-                                                                                 .stream().sorted(Comparator.comparingLong(Long::parseLong))
-                                                                                 .collect(Collectors.toList());
-
-                                                                         return list.subList(0, batches.size() - 2);
-                                                                     } else {
-                                                                         return new ArrayList<String>();
-                                                                     }
-                                                                 }
-                                                         )
+                                                         .thenApply(v -> findStaleBatches(batches))
                                                          .thenCompose(toDeleteList -> {
                                                              log.debug("deleting batches {} on new scheme", toDeleteList);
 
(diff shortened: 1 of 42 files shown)


controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
```diff
@@ -73,19 +76,25 @@
...
                                                          .collectRemaining(batches::add)
                                                          .thenApply(v -> {
                                                                      // retain latest two and delete remainder.
                                                                      if (batches.size() > 2) {
-                                                                         return batches.subList(0, batches.size() - 2);
+                                                                         List<String> list = batches
+                                                                                 .stream().sorted(Comparator.comparingLong(Long::parseLong))
+                                                                                 .collect(Collectors.toList());
+
+                                                                         return list.subList(0, batches.size() - 2);
```
RaulGracia: Why we only keep the 2 last batches? Is this an optimal value or it is just a value that sounds reasonable? In the latter case, we may want to replace that by a constant, so it can be easily modified if necessary.

shiveshr: We keep the last two values because the latest batch is what is still being written to. 
And during the time when we transition from previous batch to new batch, callers may still be interested in transactions written to previous batch as the transactions written into it should be made available for at least the GC period. 
Everything before the latest two are guaranteed to be before the GC period so we can purge them.

fpj: What's the expected behavior in the case that `batches.size() < 2`? This is going to throw `IndexOutOfBounds`, no?

shiveshr: This is the code path if batch.size > 2

controller/src/test/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStoreTest.java
```diff
@@ -196,6 +209,176 @@ public void testSplitsMerges() throws Exception {
...
+    }
+
+    private Set<Integer> getAllBatches(PravegaTablesStreamMetadataStore testStore) {
+        Set<Integer> batches = new ConcurrentSkipListSet<>();
+        testStore.getStoreHelper().getAllKeys(COMPLETED_TRANSACTIONS_BATCHES_TABLE)
+                 .collectRemaining(x -> {
+                     batches.add(Integer.parseInt(x));
+                     return true;
+                 }).join();
+        System.err.println("shivesh:: batches = " + batches);
```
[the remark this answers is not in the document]
shiveshr: oops

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
```diff
@@ -73,19 +76,25 @@
...
+    CompletableFuture<Void> gcCompletedTxn() {
         List<String> batches = new ArrayList<>();
         return withCompletion(storeHelper.expectingDataNotFound(storeHelper.getAllKeys(COMPLETED_TRANSACTIONS_BATCHES_TABLE)
                                                          .collectRemaining(batches::add)
                                                          .thenApply(v -> {
                                                                      // retain latest two and delete remainder.
                                                                      if (batches.size() > 2) {
-                                                                         return batches.subList(0, batches.size() - 2);
+                                                                         List<String> list = batches
+                                                                                 .stream().sorted(Comparator.comparingLong(Long::parseLong))
```
andreipaduroiu: This is a very inefficient sort. On every comparison you invoke `Long.parseInt`. Please convert the list to Long and then sort it.

andreipaduroiu: And since you only need the top 2, there may be more efficient methods that can do this in one pass, without sorting.

shiveshr: Actually the list was going to be very small - typically 3 elements and this code was exercised once every day. so it would not have mattered so much. 
But i have changed it to use a one pass algorithm to exclude two largest numbers in the list.

PR #3922, opened 2019-06-18
Issue 3854 : K8 system tests fail with operator 0.4.0

### LINKED ISSUE #3854: K8 system tests fail with operator 0.4.0
**Problem description**

After the [0.4.0 release](https://github.com/pravega/pravega-operator/releases/tag/0.4.0) of the operator, Kubernetes-based system tests have started to fail (see: https://github.com/pravega/pravega-operator/issues/199). The reason for these failures is that the operator is unable to create the new admission webhook component due to lack of permissions. 

**Problem location**

https://github.com/pravega/pravega/blob/10a410a190039098536d0b873c08d181b9b8f60b/test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java#L104

The testing framework needs to create the ClusterRole and ClusterRoleBinding as defined in the below files.

https://github.com/pravega/pravega-operator/blob/master/deploy/role.yaml#L48

https://github.com/pravega/pravega-operator/blob/master/deploy/role_binding.yaml#L15

SrishT:
**Change log description**  
After the 0.4.0 release of the operator, Kubernetes-based system tests started to fail. The reason for these failures was that the operator was unable to create the new admission webhook component due to lack of permissions. This PR solves that issue by assigning the requisite permissions.

**Purpose of the change**  
Resolves #3854

**What the code does**  
Here's what the code in this PR does :
1. It adds the `ClusterRole` and `ClusterRoleBinding` into the test framework.
2. It changes the custom resource definition provided in the test file, incorporating the changes introduced in the `0.4.0` release of the operator.

**How to verify it**  
System Tests should pass


test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -133,23 +142,28 @@ public String getID() {
...
+                                                                                    .build();
+
         return ImmutableMap.<String, Object>builder()
                 .put("apiVersion", "pravega.pravega.io/v1alpha1")
                 .put("kind", CUSTOM_RESOURCE_KIND_PRAVEGA)
                 .put("metadata", ImmutableMap.of("name", PRAVEGA_ID, "namespace", NAMESPACE))
                 .put("spec", ImmutableMap.builder().put("zookeeperUri", zkLocation)
                                          .put("bookkeeper", bookkeeperSpec)
                                          .put("pravega", pravegaSpec)
+                                         .put("version", VERSION)
```
shrids: This implies the Pravega version and bookkeeper version is always the same. 
Also, `pravegaBookkeeperVersion` should be removed from build.gradle.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -133,23 +142,28 @@ public String getID() {
...
                                          .build())
                 .build();
     }
```
shrids: the method `getImageSpec(...)` should be removed if it is not used.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -77,8 +81,7 @@
     static final String ZOOKEEPER_OPERATOR_IMAGE = System.getProperty("zookeeperOperatorImage", "pravega/zookeeper-operator:latest");
     static final String IMAGE_PULL_POLICY = System.getProperty("imagePullPolicy", "Always");
     private static final String DOCKER_REGISTRY =  System.getProperty("dockerRegistryUrl", "");
-    private static final String PRAVEGA_VERSION = System.getProperty("imageVersion", "latest");
-    private static final String PRAVEGA_BOOKKEEPER_VERSION = System.getProperty("pravegaBookkeeperVersion", PRAVEGA_VERSION);
+    private static final String VERSION = System.getProperty("imageVersion", "latest");
     private static final String PRAVEGA_OPERATOR_IMAGE = System.getProperty("pravegaOperatorImage", "pravega/pravega-operator:latest");
```
shrids: This change implies the framework will not support Pravega operator versions < 4.0.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -133,23 +142,28 @@ public String getID() {
...
-                                                                                      .put("tier2", tier2Spec())
-                                                                                      .build();
+                                                                                    .put("segmentStoreReplicas", segmentStoreCount)
+                                                                                    .put("debugLogging", true)
+                                                                                    .put("cacheVolumeClaimTemplate", pravegaPersistentVolumeSpec)
+                                                                                    .put("controllerResources", getResources("2000m", "3Gi", "1000m", "1Gi"))
+                                                                                    .put("segmentStoreResources", getResources("2000m", "5Gi", "1000m", "3Gi"))
+                                                                                    //.put("options", props.getProperties())
```
shrids: the commented line can be removed...

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -195,6 +209,19 @@ public String getID() {
                 .build();
     }
 
+    private Map<String, Object> getResources(String limits_cpu, String limits_mem, String requests_cpu, String requests_mem) {
+        return ImmutableMap.<String, Object>builder()
+                .put("limits",ImmutableMap.builder()
```
RaulGracia: Minor: lacking space after comma.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -195,6 +209,19 @@ public String getID() {
...
+                                    .put("cpu", limits_cpu)
+                                    .put("memory", limits_mem)
+                                    .build())
+                .put("requests",ImmutableMap.builder()
```
RaulGracia: Same here.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -77,8 +81,7 @@
     static final String ZOOKEEPER_OPERATOR_IMAGE = System.getProperty("zookeeperOperatorImage", "pravega/zookeeper-operator:latest");
     static final String IMAGE_PULL_POLICY = System.getProperty("imagePullPolicy", "Always");
     private static final String DOCKER_REGISTRY =  System.getProperty("dockerRegistryUrl", "");
-    private static final String PRAVEGA_VERSION = System.getProperty("imageVersion", "latest");
-    private static final String PRAVEGA_BOOKKEEPER_VERSION = System.getProperty("pravegaBookkeeperVersion", PRAVEGA_VERSION);
```
RaulGracia: Similar to what @shrids has commented, why do we need to change these versions to enable Pravega operator `0.4.0` to work?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -77,8 +81,7 @@
     static final String ZOOKEEPER_OPERATOR_IMAGE = System.getProperty("zookeeperOperatorImage", "pravega/zookeeper-operator:latest");
     static final String IMAGE_PULL_POLICY = System.getProperty("imagePullPolicy", "Always");
     private static final String DOCKER_REGISTRY =  System.getProperty("dockerRegistryUrl", "");
-    private static final String PRAVEGA_VERSION = System.getProperty("imageVersion", "latest");
```
RaulGracia: Is the Pravega Operator going to work with the tags used in our snapshots? That is, using a tag like `0.6.0-2291.dd02097` is valid with the current upgrade logic?

SrishT: Yes @RaulGracia  those tags will work with the updated logic

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: Because these changes have already been incorporated in the operator 0.4.0 code and they need to be reflected in the CRD being created from the System Tests framework in order to work.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: Yes, this change does bring that restriction but needs to be incorporated since these changes are already a part of 0.4.0. If we need any changes in this logic, maybe we can create a separate issue to handle that change.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
pbelgundi: This removes the ability to specify a specific bookkeeper version and locks us into a certain version, which is not correct. Have you tested if your changes enable running of system tests with other operator versions as well?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,10 +125,14 @@ public String getID() {
         // generate BookkeeperSpec.
         final Map<String, Object> bkPersistentVolumeSpec = getPersistentVolumeClaimSpec("10Gi", "standard");
         // use the latest version of bookkeeper.
-        final Map<String, Object> bookkeeperSpec = ImmutableMap.<String, Object>builder().put("image",
-                                                                                             getImageSpec(DOCKER_REGISTRY + PREFIX + "/" + BOOKKEEPER_IMAGE_NAME, PRAVEGA_BOOKKEEPER_VERSION))
+
+        String bookkeeperImg = DOCKER_REGISTRY + PREFIX + "/" + BOOKKEEPER_IMAGE_NAME;
```
pbelgundi: Why is this change needed for getting operator 4.0 to work with system tests?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -133,23 +142,28 @@ public String getID() {
...
-                                                                                      .put("image",
-                                                                                       getImageSpec(DOCKER_REGISTRY + PREFIX + "/" + PRAVEGA_IMAGE_NAME, PRAVEGA_VERSION))
-                                                                                      .put("tier2", tier2Spec())
-                                                                                      .build();
+                                                                                    .put("segmentStoreReplicas", segmentStoreCount)
+                                                                                    .put("debugLogging", true)
+                                                                                    .put("cacheVolumeClaimTemplate", pravegaPersistentVolumeSpec)
+                                                                                    .put("controllerResources", getResources("2000m", "3Gi", "1000m", "1Gi"))
+                                                                                    .put("segmentStoreResources", getResources("2000m", "5Gi", "1000m", "3Gi"))
```
pbelgundi: Why is this necessary for using operator 4.0?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -133,23 +142,28 @@ public String getID() {
...
+                                                                                    //.put("options", props.getProperties())
+                                                                                    .put("options", props)
+                                                                                    .put("image", ImmutableMap.of("repository", pravegaImg))
```
pbelgundi: Again, this removes the ability to run system tests with a specific version of pravega, which is not acceptable.  I don't see the need for this change?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
pbelgundi: Why is this version field needed?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
RaulGracia: This change is also needed for the previous version of the Operator, but it hasn't been introduced yet. In fact, we have observed system tests that induce high load in the system to fail due to lack of resources, so we need to start specifying appropriate resources for pods to avoid these failures. While this is not necessary for the Operator to work, it is necessary for the correct execution of our tests, so I'm in favor to get these changes in as soon as possible.

SrishT:
Closing this PR as the changes in this branch cannot be merged

PR #3924, opened 2019-06-18
Issue 3692: Do not remove counter of parent segment when delete or seal txn segment

### LINKED ISSUE #3692: Segment write/read metrics not reported on committed transactions
**Problem description**
When data is written to a stream (`<transaction-stream>`) in transactional writes, none of the following metrics is reported with the `stream` tag value  `<transaction-stream>`:
```
segmentstore.segment.write_bytes
segmentstore.segment.read_bytes
segmentstore.segment.write_events
```

I am sure that the data has been actually written and read from the stream, because:
- the `transactions_committed` metrics report non-zero values for this stream;
-  the client reports data being written and read from the stream
- the `segmentstore.segment.write_bytes_global`, `segmentstore.segment.read_bytes_global`,  and `segmentstore.segment.write_events_global` metrics report excess counts as compared to the totals of all scope/stream/segment read/write counts aggregated over pravega cluster.

The issue appeared when Micrometer metrics have been implemented.

**Problem location**
Segmentstore metrics

Segmentstores should report segment read/write metrics on the parent segments after transactional writes have been committed to the stream.

kevinhan88:
**Change log description**  
- Stopped deleting counter of parent segment when delete or seal transaction segment.
- Added metrics registry level activity logs.

**Purpose of the change**  
Fixes #3692 

**What the code does**  
Inside SegmentStatsRecorderImpl, when delete or seal segment, first check if it is transaction segment.
If it is txn segment, then do not close the counter, because there is no associated counter for txn segment.

**How to verify it**  
Unit tests should pass.
In particular, longevity small-tx and other txn related tests should now see the metrics of committed txn.

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
@@ -149,7 +149,7 @@ class SegmentStatsRecorderImpl implements SegmentStatsRecorder {
 
     @Override
     public void deleteSegment(String streamSegmentName) {
-        // Do not close the counter of parent segment when delete transaction segment.
+        // Do not close the counter of parent segment when deleting transaction segment.
         if (!StreamSegmentNameUtils.isTransactionSegment(streamSegmentName)) {
             getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
             getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
@@ -159,7 +159,7 @@ class SegmentStatsRecorderImpl implements SegmentStatsRecorder {
 
     @Override
     public void sealSegment(String streamSegmentName) {
-        // Do not close the counter of parent segment when seal transaction segment.
+        // Do not close the counter of parent segment when sealing transaction segment.
         if (!StreamSegmentNameUtils.isTransactionSegment(streamSegmentName)) {
             getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
             getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
@@ -148,16 +148,22 @@ class SegmentStatsRecorderImpl implements SegmentStatsRecorder {
     }
 
     @Override
-    public void deleteSegment(String segmentName) {
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(segmentName));
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(segmentName));
-        getDynamicLogger().freezeCounter(SEGMENT_READ_BYTES, segmentTags(segmentName));
+    public void deleteSegment(String streamSegmentName) {
+        // Do not close the counter of parent segment when delete transaction segment.
+        if (!StreamSegmentNameUtils.isTransactionSegment(streamSegmentName)) {
+            getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
+            getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
+            getDynamicLogger().freezeCounter(SEGMENT_READ_BYTES, segmentTags(streamSegmentName));
+        }
     }
 
     @Override
     public void sealSegment(String streamSegmentName) {
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
+        // Do not close the counter of parent segment when seal transaction segment.
+        if (!StreamSegmentNameUtils.isTransactionSegment(streamSegmentName)) {
+            getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
+            getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
+        }
         if (getSegmentAggregate(streamSegmentName) != null) {
             cache.invalidate(streamSegmentName);
             reporter.notifySealed(streamSegmentName);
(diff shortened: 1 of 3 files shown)


shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -103,6 +103,7 @@ public Counter call() throws Exception {
                 }
             });
             counter.add(delta);
+            log.debug("Metrics: increased counter {} by {} ", keys.getCacheKey(), delta);
```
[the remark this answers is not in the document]
kevinhan88: Changed logging level to trace using log.trace() directly as lit's just logging for one line (e.g. counter.add(delta)).

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderTest.java
```diff
@@ -120,6 +120,16 @@ public void testMetrics() {
...
 
+        // Delete the 1st txn segment, this shouldn't affect the parent segment
+        context.statsRecorder.deleteSegment(txnName);
+
+        // Append the 2nd metrics txn.
+        val txnName2 = StreamSegmentNameUtils.getTransactionNameFromId(STREAM_SEGMENT_NAME, UUID.randomUUID());
+        context.statsRecorder.recordAppend(txnName2, 321L, 5, elapsed);
+
+        // Seal the 2nd txn segment, this shouldn't affect the parent segment
+        context.statsRecorder.sealSegment(txnName2);
```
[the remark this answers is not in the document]
kevinhan88: Without the fix in this PR, if the txn segment is sealed, or deleted, the counter of parent segment will be closed; then the following assertion of (parent) segment will fail.
With this fix, even we delete/seal txn segment in the middle, it shouldn't affect the following assertion.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -103,6 +103,7 @@ public Counter call() throws Exception {
                 }
             });
             counter.add(delta);
+            log.trace("Metrics: increased counter {} by {} ", keys.getCacheKey(), delta);
```
fpj: Thanks for changing to trace level, but that's not exactly what I had in mind. To quote myself:

> If you change to trace, then consider using `LoggerHelpers.traceEnter` and `LoggerHelper.traceLeave`.


Main reference:
https://github.com/pravega/pravega/blob/master/common/src/main/java/io/pravega/common/LoggerHelpers.java#L19

and examples:
https://github.com/pravega/pravega/search?q=LoggerHelpers&unscoped_q=LoggerHelpers

kevinhan88: Fixed

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
```diff
@@ -148,16 +148,22 @@ public void createSegment(String streamSegmentName, byte type, int targetRate, D
     }
 
     @Override
-    public void deleteSegment(String segmentName) {
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(segmentName));
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(segmentName));
-        getDynamicLogger().freezeCounter(SEGMENT_READ_BYTES, segmentTags(segmentName));
+    public void deleteSegment(String streamSegmentName) {
+        // Do not close the counter of parent segment when delete transaction segment.
```
andreipaduroiu: ... "deleting the transaction" ...

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
```diff
@@ -148,16 +148,22 @@ public void createSegment(String streamSegmentName, byte type, int targetRate, D
...
+            getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
+            getDynamicLogger().freezeCounter(SEGMENT_READ_BYTES, segmentTags(streamSegmentName));
+        }
     }
 
     @Override
     public void sealSegment(String streamSegmentName) {
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_BYTES, segmentTags(streamSegmentName));
-        getDynamicLogger().freezeCounter(SEGMENT_WRITE_EVENTS, segmentTags(streamSegmentName));
+        // Do not close the counter of parent segment when seal transaction segment.
```
andreipaduroiu: "sealing the"

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -94,15 +95,18 @@ public void onRemoval(RemovalNotification<String, Meter> removal) {
     public void incCounterValue(String name, long delta, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
         Preconditions.checkNotNull(delta);
+        long traceId = LoggerHelpers.traceEnter(log, "Metrics: incCounterValue", name, delta, tags);
         try {
             MetricsNames.MetricKey keys = metricKey(name, tags);
             Counter counter = countersCache.get(keys.getCacheKey(), new Callable<Counter>() {
                 @Override
                 public Counter call() throws Exception {
+                    LoggerHelpers.traceLeave(log, "Metrics: createCounter", traceId, keys.getRegistryKey(), tags);
```
andreipaduroiu: 1. The second argument should match the value you pass into the `traceEnter`, otherwise you will never know where something begins and ends.
2. I suggest removing "Metrics: " from the message. That will be logged anyway by the logger.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
```diff
@@ -111,68 +115,84 @@ public Counter call() throws Exception {
     @Override
     public void updateCounterValue(String name, long value, String... tags) {
         Exceptions.checkNotNullOrEmpty(name, "name");
+        long traceId = LoggerHelpers.traceEnter(log, "Metrics: updateCounterValue", name, value, tags);
```
andreipaduroiu: here too

andreipaduroiu: and everywhere below

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
kevinhan88: fixed

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
kevinhan88: fixed

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: 1: It seems overkill to have traceEnter/traceLeave pair for every single line metrics operation. Those operations are one line call, and unlikely to fail (because most of them are 3rd library calls). 
Here we're interested in knowing whether the operation happened, instead of whether the operation succeeded.
2. Fixed.

shared/metrics/src/main/java/io/pravega/shared/metrics/DynamicLoggerImpl.java
kevinhan88: fixed

PR #3939, opened 2019-06-21
Issue 3938: Ensure ServerConnection#close() sends all messages in the buffer before closing the channel

### LINKED ISSUE #3938: ServerConnection.close() should ensure all messages in buffer are sent before closing the netty channel.
**Problem description**
Consider the following code snippet inside PravegaRequestProcessor
```
...if (u instanceof AuthenticationException) {
            log.warn(requestId, "Authentication error during '{}'.", operation);
            invokeSafely(connection::send, new AuthTokenCheckFailed(requestId, clientReplyStackTrace), failureHandler);
            connection.close();
        } 
```
In this case it can so happen that the AuthTokenCheckFailed message is never received by the client.

**Problem location**
ServerConnectionInboundHandler

Fix the issue.

shrids:
**Change log description**  
Ensure `ServerConnection#close() ` sends all messages in the buffer before closing the channel.

**Purpose of the change**  
Fixes #3938 

**What the code does**  
This PR ensures the channel is closed after all the message up to now are flushed and sent over the wire.
This ensures that the usage pattern shown below does not cause missing messages.
```
io.pravega.segmentstore.server.host.handler.ServerConnection connection = ...
connection.send(wirecommand);
connection.close(); 
```
**How to verify it**  


kmgowda:
actually we need append close ;  flushing the pending events is already taken care in this PR : https://github.com/pravega/pravega/pull/3953

shrids:
Thanks @kmgowda however this PR is related to changes on the server side and is not about Pravega client's handling of pending events.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/AppendProcessorTest.java
```diff
@@ -348,6 +364,36 @@ public void testAppendFails() throws Exception {
...
+
+        CompletableFuture<Void> result = new CompletableFuture<>();
+        result.completeExceptionally(new RuntimeException("Fake exception for testing"));
+        when(store.append(streamSegmentName, data, updateEventNumber(clientId, data.length), AppendProcessor.TIMEOUT))
+                .thenReturn(result);
+
+        @Cleanup
+        EmbeddedChannel channel = createChannel(store);
+
+        // Send a setup append WireCommand.
```
fpj: I'm not being able to map the pattern in the description:

```
connection.send(wirecommand);
connection.close();
```

to here. Could you explain how we are validating the pattern in the PR description?

shrids: yes, this test verifies only if `connection.close()` is invoked and the netty channel is infact closed.
Will add a new Junit to simulate the exact pattern.

shrids: This issue was observed when `AuthTokenCheckFailed` wirecommand was not received by the client. Now PR#3887 fixes it. 
image

Observed that a similar scenario with `BadAttributeUpdateException` handling in AppendProcessor. I have added a junit to validate that.

PR #3947, opened 2019-06-21
Issue 3856: Add timeouts for Rawclient

### LINKED ISSUE #3856: Add timeout to requests in RawClient
**Problem description**
In the RawClient when a request is sent it is helpful to have a reply timeout so that an application does not have to block forever, and it can close the connection and retry if the server is slow.

**Problem location**
RawClient

Add a timeout parameter, which closes the connection. 

It's worth noting that we should set the default timeout differently for testing and production. We want the timeout to be infinite for our system tests so that we don't end up accidentally masking bugs.

Tristan1900:
**Change log description**  
This PR adds a timeout when sending requests using Rawclient. If the reply is not finished within timeout, a `TimeoutException` will be thrown. 

**Purpose of the change**  
Fix #3856 

**What the code does**  
The code uses another thread in `sendRequest` method as a timer and returns the `CompletableFuture<Reply>` immediately. When the timer fires it will call `reply.completeExceptionally(e)` and pass it a `TimeoutException`.

**How to verify it**  
Tests should pass

Signed-off-by: wenqimou wenqimou@gmail.com


client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -84,12 +87,14 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
 
     public RawClient(PravegaNodeUri uri, ConnectionFactory connectionFactory) {
         this.segmentId = null;
+        this.connectionFactory = connectionFactory;
```
andreipaduroiu: Check for null. I suggest adding `@NonNull` as an annotation (mind the casing; don't confuse with `@Nonnull`)

andreipaduroiu: Same below

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -145,6 +163,7 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
                     reply.completeExceptionally(cfe);
                     closeConnection(cfe);
                 }
+                timer.completeExceptionally(new InterruptedException());
```
andreipaduroiu: I would think `cancel(true)` would be more appropriate?

controller/src/test/java/io/pravega/controller/server/SegmentHelperTest.java
```diff
@@ -713,7 +715,10 @@ public Host getHostForTableSegment(String table) {
 
         @Override
         public ScheduledExecutorService getInternalExecutor() {
-            return null;
+            if (executor == null) {
```
andreipaduroiu: I don't recommend using this pattern. Please set the executor either in a constructor or when you declare the field.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
Tristan1900: Got it, thanks!

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
Tristan1900: You are right, thanks.

controller/src/test/java/io/pravega/controller/server/SegmentHelperTest.java
Tristan1900: Got it.

client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
```diff
@@ -115,7 +115,8 @@ RawClient getConnection() {
         log.debug("Getting segment info for segment: {}", segmentId);
         RawClient connection = getConnection();
         long requestId = connection.getFlow().getNextSequenceNumber();
-        return connection.sendRequest(requestId, new GetStreamSegmentInfo(requestId, segmentId.getScopedName(), delegationToken))
+        return connection.sendRequest(requestId,
+                                      new GetStreamSegmentInfo(requestId, segmentId.getScopedName(), delegationToken))
```
tkaitchuck: I don't see any functional change in this file.

Tristan1900: Yes, that's right. I was doing some functional changes in that file at first but decide to not use that change in the end. The modified line was a little long so I make it shorter. I think it might be useful so I keep it there.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -131,12 +137,25 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
...
+                                                                                  Duration timeout) {
         return connection.thenCompose(c -> {
             log.debug("Sending request: {}", request);
             CompletableFuture<Reply> reply = new CompletableFuture<>();
             synchronized (lock) {
                 requests.put(requestId, reply);
             }
+            CompletableFuture<Void> timer = Futures.futureWithTimeout(timeout, connectionFactory.getInternalExecutor());
+            Futures.onTimeout(timer, ex -> {
+                synchronized (lock) {
```
shrids: can we log the fact that we are completing the future exceptionally due to a timeout? (this will help to debug the exact the request which failed with a timeout)

andreipaduroiu:
@Tristan1900 @fpj  What's the status of this PR? There's been no activity in 4 months.

If we don't plan on moving forward with this, we should close it.

Tristan1900:
@andreipaduroiu Sorry that I totally forget about this PR. As I remember I was transferred to do something else urgent back in a few months. 
I will finish it soon.

andreipaduroiu:
@fpj Can you please have a look at this PR again please? It's waiting on you.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
Tristan1900: got it!

client/src/main/java/io/pravega/client/ClientConfig.java
```diff
@@ -108,6 +109,13 @@
      */
     private final MetricListener metricListener;
 
+    /**
+     * If the flag is set, RawClient will timeout the future if it doesn't receive response from Segment store
+     * in a given time period. Notice that we don't want this feature enabled in testing since it might mask some
+     * bugs.
+     */
+    private final Duration rawclientTimeout;
```
tkaitchuck, line 117: This is a public api, so it needs to make sense to a user. RawClient is not a public class, so this explication makes no sense. It would be better if this was "serverRequestTimeout" or something similar.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -131,12 +140,26 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
...
+
+    public <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(long requestId, T request,
+                                                                                  Duration timeout) {
         return connection.thenCompose(c -> {
             log.debug("Sending request: {}", request);
             CompletableFuture<Reply> reply = new CompletableFuture<>();
             synchronized (lock) {
                 requests.put(requestId, reply);
             }
+            CompletableFuture<Void> timer = Futures.futureWithTimeout(timeout, connectionFactory.getInternalExecutor());
```
tkaitchuck, line 154: This pattern of creating a future with a timeout and scheduling a job after a timeout, would work better if combined into a single method similar to this one: https://github.com/pravega/pravega/pull/4698/files#diff-0bcfd93b4074110157f7e0729d9c1f11R250

It would be good to add that to Futures and update anywhere else using those two calls in succession to use the new method.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -158,6 +159,7 @@ public static ClientConfig buildClientConfig(URI controllerUri) {
                                // auth
                                .credentials(new DefaultCredentials("1111_aaaa", "admin"))
                                .controllerURI(controllerUri)
+                               .rawclientTimeout(Duration.ofSeconds(3600))
```
tkaitchuck, line 162: Why so long? It seems like for tests we'd want it rather short so we catch any stalls before the test times out.

controller/src/test/java/io/pravega/controller/server/ZKBackedControllerServiceStarterTest.java
```diff
@@ -110,6 +112,7 @@ public void tearDown() throws Exception {
         private ReplyProcessor rp;
         private ClientConnection connection;
         private final ScheduledExecutorService executorService = Executors.newScheduledThreadPool(2);
+        private ClientConfig clientConfig = ClientConfig.builder().rawclientTimeout(Duration.ofSeconds(3600)).build();
```
tkaitchuck, line 115: see below

controller/src/test/java/io/pravega/controller/server/SegmentStoreConnectionManagerTest.java
```diff
@@ -351,6 +353,7 @@ public void tableEntriesRead(WireCommands.TableEntriesRead tableEntriesRead) {
     private class MockConnectionFactory implements ConnectionFactory {
         @Getter
         private ReplyProcessor rp;
+        private ClientConfig clientConfig = ClientConfig.builder().rawclientTimeout(Duration.ofSeconds(3600)).build();
```
tkaitchuck, line 356: see below

controller/src/test/java/io/pravega/controller/server/SegmentHelperTest.java
```diff
@@ -788,6 +792,12 @@ public Host getHostForTableSegment(String table) {
         @Getter
         private ReplyProcessor rp;
         private ClientConnection connection;
+        private ScheduledExecutorService executor;
+        private ClientConfig clientConfig = ClientConfig.builder().rawclientTimeout(Duration.ofSeconds(3600)).build();
```
tkaitchuck, line 796: see below

client/src/test/java/io/pravega/client/segment/impl/SegmentMetadataClientTest.java
```diff
@@ -300,9 +303,11 @@ public void testExceptionOnSend() {
         PravegaNodeUri endpoint = new PravegaNodeUri("localhost", 0);
         @Cleanup("shutdown")
         InlineExecutor executor = new InlineExecutor();
+        ClientConfig config = ClientConfig.builder().rawclientTimeout(Duration.ofSeconds(3600)).build();
```
tkaitchuck, line 306: see below

client/src/main/java/io/pravega/client/ClientConfig.java
Tristan1900: Got it! Makes sense.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
Tristan1900: Got it

Tristan1900:
@andreipaduroiu Sure, test has been added, thanks for the notice! If it passes I will ask Tom to take a look at it again.

andreipaduroiu:
@fpj This PR is waiting on you. Can you have a look at it when you have a chance please?

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -131,12 +140,27 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
...
         return connection.thenCompose(c -> {
             log.debug("Sending request: {}", request);
             CompletableFuture<Reply> reply = new CompletableFuture<>();
             synchronized (lock) {
                 requests.put(requestId, reply);
             }
+
+            Futures.onTimeout(reply, timeout, connectionFactory.getInternalExecutor(), ex -> {
+                synchronized (lock) {
+                    log.debug("Request {} times out after {}", requestId, timeout.toString());
```
shrids, line 157: Can we log it @info/@warn level since we can indicate that server took greater 30 seconds?

andreipaduroiu:
@fpj Are you going to revisit this PR?

andreipaduroiu:
@fpj ping

client/src/main/java/io/pravega/client/ClientConfig.java
```diff
@@ -108,6 +109,13 @@
      */
     private final MetricListener metricListener;
 
+    /**
+     * If the flag is set, a server request will time out if it doesn't receive response from Segment store
+     * in a given time period. Notice that we don't want this feature enabled in testing since it might mask some
+     * bugs.
+     */
+    private final Duration serverRequestTimeout;
```
fpj, line 117: From this comment, not setting it disables the timeout, but we are giving it a default value right below. Could you clarify how we are disabling the timeout for tests?

Tristan1900: Thanks for pointing that out! Sorry it is not accurate, I have rephrased the comment. It was meant for boolean type but later it was changed to a value, so right now we just pass it a big value(3600s in system test) to disable the timeout.

fpj: I see that the comment has changed, but I still don't understand what kind of behavior this is trying to implement. My interpretation is that:

- Optional means that the application does not need to change the default value.
- The default of 30s is a compromise between a good value for the application and a value that would make our tests fail.

Choosing an ideal value is hard, in particular, if we take into account system tests.

Tristan1900: Oh I see what you mean. yes we don't need to change the default value for application, however, 30s is not a compromise and it is only meant to work for application not for tests.

In system test(and some unit tests) we need to change that value and I have set it to be 3600s. See [here](https://github.com/pravega/pravega/pull/3947/files#diff-1a32a0fd61346692a52aacb1db352f8cR162). If we need to set it in other test such as longevity test, then the longevity code needs to be modifed.

fpj: Ok, so elaborate on what you are trying to achieve then. The ideal behavior is:

- The application doesn't need to set anything and out of the box it has some reasonable timeout. How have you come up with 30s? Think that a pravega operation in general takes a few milliseconds, so 30s waiting is a long time.
- Ideally, this timeout does not mask bugs in tests, so the timeout should be for practical purposes disabled.
- Ideally, we don't have to remember to set this for new tests we build, otherwise I'm sure we will forget.

I'm not sure whether this is possible, but we could think of some sort of test flag that disables the timeout for testing and disables it for all cases in a run when the flag is set (timeout disabled). it is easier to set a flag once.

client/src/main/java/io/pravega/client/ClientConfig.java
```diff
@@ -108,6 +108,15 @@
...
     private final MetricListener metricListener;
 
+    /**
+     * An optional property representing whether the current run is for testing purpose. By default this property is
+     * not enabled.
+     *
+     * If this property is set, the request timeout to segmentstore is set to a very large value(3600s) in order to not
+     * mask any bugs. By default the request timeout in real application is set to 30s.
+     */
+    private final boolean enableTesting;
```
eolivelli, line 118: This property is too vague.
Now it means that we will use a longer timeout, but the user cannot know what this property will mean in the future.

It is better to name it 'request timeout'  or something like that.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
```diff
@@ -131,12 +140,31 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
...
+    public <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(long requestId, T request,
+                                                                                  Duration timeout) {
         return connection.thenCompose(c -> {
             log.debug("Sending request: {}", request);
             CompletableFuture<Reply> reply = new CompletableFuture<>();
             synchronized (lock) {
                 requests.put(requestId, reply);
             }
+
+            Futures.onTimeout(reply, timeout, connectionFactory.getInternalExecutor(), ex -> {
```
eolivelli, line 159: This is very costly, we are adding lot of work and a few memory allocations on the hottest path.

What about saving just a deadline timestamp for the request in a map and periodically check the deadline and fail the request?

Tristan1900: Thanks. Yes this is kind of costly and I have discussed with Tom about this issue, it seems that rawclient is not on THE hottest path, but I think your way is more efficient.

client/src/main/java/io/pravega/client/ClientConfig.java
Tristan1900: 1. yes, 30s is kind of long in terms of an operation. But I'm thinking that this timeout involves network latency and server latency, there might be a temporary network partition or the server pod has restarted and stuff like that which may takes several seconds. Also here https://github.com/pravega/pravega/blob/7685d8d2bf73b7265d09e2788d3b9751c6115c4c/client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java#L39 is also using 30s. 

2. Yes in test mode this timeout is disabled. I have already set that in system tests.

3. For newly added system test, no we don't need to set that manually again because I have set it in the test framework that works for all system tests. 

Not exactly sure but I kind of think that the current flag is what you mean by the test flag since we only need to set once in the `clientconfig`.

client/src/main/java/io/pravega/client/netty/impl/RawClient.java
andreipaduroiu: RawClient may not be on the "hottest" path, but it is used a lot for non-streaming purposes. It is used for KV Tables, for the Controller Metadata requests, etc. So we do owe it to ourselves to keep it as efficient as possible.

andreipaduroiu: You can 1) do that (store all requests in a map - there already is one) and periodically loop through it and verify or 2) consider using `Futures.futureWithTimeout` (set `CompletableFuture<Reply> reply = Futures.futureWithTimeout(duration, executor)`).

You should weigh the pros and cons of each of them. The thing I dislike about option 1 is that you will periodically have to acquire the lock to go through that map and inspect all outstanding requests. This could be introducing "cleanup pauses" in the client, which may be undesirable.

Tristan1900: @eolivelli After discussing with Andrei, we are a little uncertain about the performance cost you mentioned, Andrei has used this kind of technique extensively in segment store and this method doesn't show on the flamegraph. Perhaps we missed something and it would great if you could clarify the performance cost. 
Thanks

eolivelli: If this is not one of the hottest paths than maybe there is no need to make it more complex.
I was just talking about saving memory allocations but if your analysis doesn't show that this is an hotspot then let's keep it simple

Tristan1900:
close this as change has outdated, open a new one if necessary

PR #3952, opened 2019-06-24
Issue 3951: Ensure connection pooling is configurable for Event writers.

### LINKED ISSUE #3951: Ensure connection pooling is configurable for Event Writers
**Problem description**
Connection pooling should ~not~ be enabled by default for event writers.

**Problem location**
writers.

Ensure connection pooling is configurable for event writers.

shrids:
**Change log description**  
Ensure connection pooling is configurable for Event writers.

**Purpose of the change**  
Fixes #3951 

**What the code does**  
The code ensures the user can enable/disable connection pooling for event writers. By default connection pooling is enabled for event writers.

The usage pattern to disable connection pooling is as follows
```
// Configuration to disable connection pooling for writers.
EventWriterConfig writerConfig = EventWriterConfig.builder().enableConnectionPooling(false).build();
@Cleanup
// create a writer with the configuration.
EventStreamWriter<String> writer = clientFactory.createEventWriter(SCOPE, serializer, writerConfig);
writer.writeEvent(key, event);
```
The PR also ensures segment writers created by `RevisionedStreamClient` use connection pooling by default.

**How to verify it**  
Validated the changes with a system test run.


kmgowda:
this is really nice.. I want to do this.. but you came up with PR :-)

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -94,6 +94,10 @@ public void closeFlow(ClientConnection clientConnection) {
         int flow = clientConnectionImpl.getFlowId();
         log.info("Closing Flow {} for endpoint {}", flow, clientConnectionImpl.getConnectionName());
         flowIdReplyProcessorMap.remove(flow);
+        if (flow == FLOW_DISABLED) {
```
tkaitchuck: Should we close the connection if all the flows close if there is pooling?

shrids: yes, we should prune idle connections incase pooling is enabled and this should happen after a predefined idle timeout duration. Created an issue https://github.com/pravega/pravega/issues/3961 to track the same.

shrids:
System tests are passing `https://asd-nautilus-jenkins/job/test-pravega-continuous/2168/`

PR #3953, opened 2019-06-24
Issue 3878: Avoid sending keepAlive during Segment Writer Flush

### LINKED ISSUE #3878: Segment Store is flooded with KeepAlive messages
**Problem description**
We have realized that a single Segment Store instance may receive a really high number of `KeepAlive` messages. For instance, in a system test build, we have analyzed a Segment Store instance logs and counting the for number of `Received KeepAlive` message occurrences within a second each 5 minute interval:

keepalive_requests
 
As can be observed, most of the time we observe +200 `KeepAlive` messages per second, reaching up to `480` at the end of the test. Given that the `KeepAlive` message is assumed to be sent by the client every `10` seconds, either i) there is some problem in the client/controller that justifies this amount of `KeepAlive` message, or ii) we are using `KeepAlive` messages in the IO communication between clients and the Segment Store (e.g., [flush method](https://github.com/pravega/pravega/blob/master/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java#L498)). In either case, it may be worth to revisit the use of this kind of messages. 

**Problem location**
Client/Controller

Revisit if the high rate of `KeepAlive` messages against the Segment Store is legitimate and expected.

kmgowda:
**Change log description**  
Avoid sending keepAlive on wire during data flush. But, ensure that pending writer data is flushed on wire.

**Purpose of the change**  
Fixes #3878 

**What the code does**  
A new Flush message is defined. Segment writer flush sends the flush message to CommandEncoder.
The CommandEncoder flushes only the pending data of the given writerid and avoids sending the Keepalive on wire.

**How to verify it**  
Pravega  is tested with unit tests and integration tests.
Pravega is tested with bare-metal deployment with 1 controller and 3 segment nodes on Lagaunitas (white cluster).
Benchmark mark tool : https://github.com/pravega/pravega-benchmark is used for performance bench-marking.

Signed-off-by: Keshava Munegowda keshava.munegowda@dell.com


client/src/main/java/io/pravega/client/netty/impl/ClientConnection.java
```diff
@@ -37,7 +38,16 @@
      * @throws ConnectionFailedException The connection has died, and can no longer be used.
      */
     void send(Append append) throws ConnectionFailedException;
-    
+
+    /**
+     * flushes the pending events.
```
andreipaduroiu: Capital "F"

client/src/main/java/io/pravega/client/netty/impl/ClientConnection.java
```diff
@@ -37,7 +38,16 @@
...
+     *
+     * @param flush flush the pending message.
```
andreipaduroiu: The flush command to send.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -139,6 +139,11 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
         } else if (msg instanceof WireCommand) {
             breakFromAppend(out);
             writeMessage((WireCommand) msg, out);
+        } else if (msg instanceof  Flush) {
```
andreipaduroiu: too many consecutive spaces

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/Flush.java
```diff
@@ -0,0 +1,22 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.shared.protocol.netty;
+
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
+import java.util.UUID;
+
```
andreipaduroiu: Too many empty lines

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/Flush.java
```diff
@@ -0,0 +1,22 @@
...
+
+@Data
+@RequiredArgsConstructor
```
andreipaduroiu: Javadoc

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -139,6 +139,11 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
         } else if (msg instanceof WireCommand) {
             breakFromAppend(out);
```
shrids: The `KeepAlive` message would cause the invocation of `breakFromAppend(out)` which is similar to the what is being done with new msg `Flush`.
Could you please describe the advantage of creating a new `WireCommand v/s using `KeepAlive`

client/src/main/java/io/pravega/client/netty/impl/ClientConnection.java
kmgowda: Tom , advised to move this flush as WireComand in the file WirCommands.java file. All your review comments are indirectly taken care now.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/Flush.java
kmgowda: done

client/src/main/java/io/pravega/client/netty/impl/ClientConnection.java
kmgowda: removed this interface API

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/Flush.java
kmgowda: removed this file

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: This new wireCommand flush ; will flush the pending writes ; but it does not write any additional messages to wire .. This means no keep alive to wire during flush.. This reduces the keepalives to segment store.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2081,4 +2081,15 @@ public boolean isFailure() {
         }
     }
 
+    @Data
+    public static final class Flush implements WireCommand {
```
andreipaduroiu: Can we make sure this doesn't go over the wire. Should we make `WireCommandType.FLUSH_SEGMENT` a negative value.

If this makes it over the wire by accident, there won't be a processor for it on the server side and we'll start logging errors.

For redundancy, I'd also override `mustLog()` and return `false`, which will guarantee that even if it does somehow make it on the server side, it won't pollute the logs with it.

kmgowda: ok, I will make this WireCommandType.FLUSH_SEGMENT as negative value.
But, writefields of the flush throws an IOExcpetion : https://github.com/pravega/pravega/pull/3953/files#diff-599b4d30bd4eb53895a9c32146254bb0R2092

so, there is no way that it goes on wire. so, it never be on server side.

and readFrom is null see here : https://github.com/pravega/pravega/pull/3953/files#diff-a3019538349210904685160f5be806d5R107;  AppenDecoder will throw error in case  (imaginary) it goes to server.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -136,6 +137,11 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
             if (currentBlockSize == timeoutMsg.ifStillBlockSize) {
                 breakFromAppend(out);
             }
+        } else if (msg instanceof Flush) {
+            Flush flush = (Flush) msg;
+            if (flush.getSegment().equals(segmentBeingAppendedTo) && flush.getWriterId().equals(writerIdPerformingAppends)) {
```
shrids: Flush from other segment writers will not cause the client to to break from the current append.
We can simply this check to `if (flush.getWriterId().equals(writerIdPerformingAppends)) ` right?

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/WireCommandsTest.java
```diff
@@ -621,6 +621,15 @@ public void testKeepAlive() throws IOException {
         testCommand(new WireCommands.KeepAlive());
     }
 
+    @Test(expected = IOException.class)
+    public void testFlush() throws IOException {
```
fpj: All this test is doing is validating that the new command is constructed correctly. Even though such a test should exist, I expect tests that validate the behavior as well. Do such tests exist already?

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -496,7 +496,7 @@ public void flush() throws SegmentSealedException {
         if (numInflight != 0) {
             try {
                 ClientConnection connection = Futures.getThrowingException(getConnection());
-                connection.send(new KeepAlive());
+                connection.send(new Flush(writerId, segmentName));
```
fpj: If I understand it right, this flush command is invoking `breakFromAppend` in `CommandEncoder` rather than send a command over the wire. Is there a way of invoking `breakFromAppend` under the same condition without having to create a new command?

Getting rid of  `connection.send(new KeepAlive())` in line 499 sounds right.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
kmgowda: @andreipaduroiu , I have changed the WireCommandType.FLUSH_SEGMENT as negative value.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: actually yes, but the session table maintained by the commandEncoder if tuple of (writerid , segmentName) , so i am maintaining the same here

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/WireCommandsTest.java
kmgowda: we already wave testCommand to test the write-fields (serialize to wire) and readFrom - deserialize ; but this new command is should not write anything to wire. hence , a new test to just check its constructed properly and ensures that writeFields (serialize) will throw an IO error.

kmgowda:
> In addition to the comments, I'd like a confirmation that this change is backwards compatible.

Yes, its completely backward compatible ;  Because, this new command does not write anything on wire. It just ensure current segment writer data is flushed on to the wire.
As part of testing, I have tested new client (have thing change) and old segment store(with out change) and new segment store(with this change) ; its works fine.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
kmgowda: No, we dont have any such command ; there is Event Command , but we need a command which encapsulates the segment and writer , so that we dont flush other writers data.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommandType.java
```diff
@@ -22,6 +22,7 @@
  */
 public enum WireCommandType {
     HELLO(-127, WireCommands.Hello::readFrom),
+    FLUSH_SEGMENT(-126, null),
```
tkaitchuck: Let's make this -2

kmgowda: @tkaitchuck  , -2 is already in use by PARTIAL_EVENT; so , I have assigned -3 for PARTIAL_EVENT

kmgowda:
The System tests are passing; The Pravega branch builder build number is : 1610

kmgowda:
Successfully completed the system tests ; pravega branch builder , build : 1747

kmgowda:
Synced with master; 
Successfully completed the system tests ; pravega branch builder , build : 1748

kmgowda:
Hi @andreipaduroiu 
All systems tests are completed and synced master.
Requesting you to merge this PR with master.

andreipaduroiu:
This is blocked by @fpj. I have already approved it.

kmgowda:
Thanks @andreipaduroiu 
@fpj please let me know if you have any review comments; otherwise please merge this PR to master

RaulGracia:
@fpj what is your take on this PR? Can we merge it or do you expect additional changes?

kmgowda:
Hi @andreipaduroiu and @fpj 
I request please dont close this PR and merge it.

Issue #3967, opened 2019-06-27
Add test cases for compatibilty


anishakj:
**Problem description**
Create few test cases that will work with latest version of Pravega components with previous version of Pravega client.

**Problem location**
standalone tests

**Suggestions for an improvement**

PR #3968, opened 2019-06-27
Issue 3967: Added testcases to standalone

anishakj:
Signed-off-by: anishakj <anisha.kj@emc.com>
Work in progress
**Change log description**  
Added 2 new tests to standalone, to check the pravega clients are compatible.
**Purpose of the change**  
#3967

**What the code does**  
We were facing issues like Pravega server code is not backward compatible with Pravega client,
The tests added will deploy pravega components in latest version and use the one version older of pravega client. Currently 2 tests are added.

First test will create a stream, read and write events
Second test will create a transaction, read and write events from the transaction.
More tests can be added to compatibility later, if required

**How to verify it**  


anishakj:
After discussion with @shrids and @Raul , decided to continue with new approach. So closing this PR.

PR #3970, opened 2019-06-27
Issue 3969: Performance Improvement with connection pooling by serializing writers 

### LINKED ISSUE #3969: Improve the Performance with Connection Pooling
**Problem description**
The Connection Pooling has reduced the writer performance.
Another  issue #3774 gives quantifies the performance degradation.

**Problem location**
Pravega Client / Append Processor

its observed that reducing the connection is causing multiplexing the netty channel and performance degradation.
So, if one writer is using channel , the buffer the next writer data and then serialize would improve the performance

kmgowda:
**Change log description**  
If the current netty channel is in use for one append path; other append paths are buffered and transmitted upon timeout or append completion.

**Purpose of the change**  
Fixes #3969 

**What the code does**  
The Command Encoder en-queues writer data if the current netty channel is already in appending state.
when timeout happens or  append complete ; all the pending writers are flushed to channel.
The Append Decoder can process the Append Block End independently, this eliminates the need for preceding the append block.

**How to verify it**  
All unit tests are passing
Conducted the performance bench marking on lagunitas and observed performance improvements.


kmgowda:
image

kmgowda:
image

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,28 +95,23 @@ public Request processCommand(WireCommand command) throws Exception {
...
-                throw new InvalidMessageException("AppendBlockEnd for wrong connection.");
-            }
             segment = getSegment(writerId);
             if (blockEnd.getLastEventNumber() < segment.lastEventNumber) {
                 throw new InvalidMessageException("Last event number went backwards.");
             }
             int sizeOfWholeEventsInBlock = blockEnd.getSizeOfWholeEvents();
-            if (sizeOfWholeEventsInBlock > currentBlock.getData().readableBytes() || sizeOfWholeEventsInBlock < 0) {
+            if ((currentBlock != null) && (sizeOfWholeEventsInBlock > currentBlock.getData().readableBytes() || sizeOfWholeEventsInBlock < 0)) {
                 throw new InvalidMessageException("Invalid SizeOfWholeEvents in block");
```
andreipaduroiu: It would be useful in debugging these if you also included the values that lead you to this determination (`sizeOfWholeEventsInBlock`, etc.).

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -130,30 +125,34 @@ public Request processCommand(WireCommand command) throws Exception {
...
-            }
-            if (cmd.getType() == WireCommandType.PARTIAL_EVENT) {
-                // Work around a bug in netty:
-                // See https://github.com/netty/netty/issues/5597
-                if (appendDataBuf.readableBytes() == 0) {
-                    appendDataBuf.release();
-                    return wrappedBuffer(((WireCommands.PartialEvent) cmd).getData(), blockEnd.getData());
-                } else {
-                    return wrappedBuffer(appendDataBuf, ((WireCommands.PartialEvent) cmd).getData(), blockEnd.getData());
+        if (currentBlock != null) {
```
andreipaduroiu: Formatting suggestion:
Replace this `currentBlock != null` with `currentBlock == null` and bail out early in that case (your current `else` block). Then, without writing an explicit `else` block, keep the original code. That will save you from looking like you changed this entire method.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +75,57 @@
     private UUID writerIdPerformingAppends;
     private int currentBlockSize;
     private int bytesLeftInBlock;
+    private final List<Session> pendingWrites = new LinkedList<>();
```
andreipaduroiu: I think `ArrayList` would be more efficient than a `LinkedList`

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -181,7 +244,7 @@ private void breakFromAppend(ByteBuf data, ByteBuf pendingData, ByteBuf out) {
     }
 
     @SneakyThrows(IOException.class)
-    private void writeMessage(AppendBlock block, ByteBuf out) {
+    private void writeMessage(AppendBlock block, int blkSize, ByteBuf out) {
```
andreipaduroiu: blkSize???

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -65,20 +68,67 @@
...
+        private void recordAppend(Append append) {
+            lastEventNumber = append.getEventNumber();
+            eventCount += append.getEventCount();
+        }
+
+        private void append(ByteBuf buffer, ByteBuf out) {
+            if (buffer != null && buffer.readableBytes() > 0) {
+                totalBytes += buffer.readableBytes();
+                if (data.readableBytes() == 0) {
+                    pendingWrites.add(this);
```
tkaitchuck: I'm not sure I understand this if block.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,62 +138,80 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
-            session.lastEventNumber = append.getEventNumber();
-            session.eventCount++;
             if (blockSizeSupplier != null) {
                 blockSizeSupplier.recordAppend(append.getEventNumber(), msgSize);
             }
-            if (bytesLeftInBlock == 0) {
-                currentBlockSize = msgSize + TYPE_PLUS_LENGTH_SIZE;
-                if (blockSizeSupplier != null) {
-                    currentBlockSize = Math.max(currentBlockSize, blockSizeSupplier.getAppendBlockSize());
+            if ((bytesLeftInBlock == 0) ||
```
tkaitchuck: This method is getting rather long. Would it make it easier to test it if we factored out several smaller helper methods?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
kmgowda: This only for sizeOfWholeEventsInBlock value of append end block ; comparision happens only if its preceded by AppendBlock.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
kmgowda: yes .. its changed now.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes .. done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: this only for AppendBlock; where approximate the number of data bytes that we determine while writing appendBlock

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,28 +95,23 @@ public Request processCommand(WireCommand command) throws Exception {
...
             currentBlock = (WireCommands.AppendBlock) command;
             result = null;
             break;
         case APPEND_BLOCK_END:
             WireCommands.AppendBlockEnd blockEnd = (WireCommands.AppendBlockEnd) command;
-            if (currentBlock == null) {
-                log.warn("Received AppendBlockEnd {} without AppendBlock", blockEnd);
-                throw new InvalidMessageException("AppendBlockEnd without AppendBlock.");
-            }
             UUID writerId = blockEnd.getWriterId();
```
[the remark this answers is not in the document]
kmgowda: yes .. its done.. and its backward compatible old versions of the  pravega client.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes.. I have done the refactoring now .. please have look

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: if channel is not free.. lets append is in progress then we start accumulating/buffer the data from other writers/sessions.

client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
```diff
@@ -131,14 +129,14 @@ public ConnectionPoolImpl(ClientConfig clientConfig) {
     }
 
     @Override
+    @Synchronized
```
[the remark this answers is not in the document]
kmgowda: This was part of Appendbatcher Size tracker bug.. ;  not part of this PR.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,29 +95,35 @@ public Request processCommand(WireCommand command) throws Exception {
...
+            }
             if (blockEnd.getLastEventNumber() < segment.lastEventNumber) {
                 throw new InvalidMessageException("Last event number went backwards.");
             }
-            int sizeOfWholeEventsInBlock = blockEnd.getSizeOfWholeEvents();
-            if (sizeOfWholeEventsInBlock > currentBlock.getData().readableBytes() || sizeOfWholeEventsInBlock < 0) {
-                throw new InvalidMessageException("Invalid SizeOfWholeEvents in block");
+            if (currentBlock != null) {
+               if (!currentBlock.getWriterId().equals(writerId)) {
+                   throw new InvalidMessageException("Writer ID mismatch between Append Block and Append block End.");
```
tkaitchuck: Include the values in the exception. It will help to debug.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,29 +95,35 @@ public Request processCommand(WireCommand command) throws Exception {
...
+               }
+               if (sizeOfWholeEventsInBlock > currentBlock.getData().readableBytes() || sizeOfWholeEventsInBlock < 0) {
+                    throw new InvalidMessageException("Invalid SizeOfWholeEvents in block");
```
tkaitchuck: Include the value in the exception. It will help to debug.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,62 +137,62 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
-                }
-            }
-            // Is there enough space for a subsequent message after this one?
-            if (bytesLeftInBlock - msgSize > TYPE_PLUS_LENGTH_SIZE) {
-                out.writeBytes(data);
-                bytesLeftInBlock -= msgSize;
+
+            if (isChannelFree()) {
+                session.flush(out);
+                session.recordAppend(append);
```
tkaitchuck: Trailing space

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,62 +137,62 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+                if (startAppend(ctx, blockSizeSupplier, append, out)) {
+                    continueAppend(data, out);
+               } else {
```
tkaitchuck: If and else are indented differently

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,62 +137,62 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
-                                            TimeUnit.MILLISECONDS);
-                }
-            }
-            // Is there enough space for a subsequent message after this one?
-            if (bytesLeftInBlock - msgSize > TYPE_PLUS_LENGTH_SIZE) {
-                out.writeBytes(data);
-                bytesLeftInBlock -= msgSize;
+
+            if (isChannelFree()) {
+                session.flush(out);
```
tkaitchuck: Is this needed? I am confused why this can be immediately before the start append below.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -154,34 +203,71 @@ private void validateAppend(Append append, Session session) {
...
-        if (bytesLeftInBlock != 0) {
+    private boolean isChannelFree() {
+        return writerIdPerformingAppends == null;
+    }
+
+    private boolean isChannelOwner(UUID writerID, String segment) {
+        return writerID.equals(writerIdPerformingAppends) && segment.equals(segmentBeingAppendedTo);
+    }
+
+    private boolean startAppend(ChannelHandlerContext ctx, AppendBatchSizeTracker blockSizeSupplier, Append append, ByteBuf out) {
```
tkaitchuck: Javadoc

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
kmgowda: yes.. it included now

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
kmgowda: yes its included now.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: I dint find a trailing space here..  please let me know anything i am missing in this code block

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: thanks .. it corrected

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes.. in case if the append completed (not break) , then channel become free and before it takes starts new session , the existing session data need be flushed.. there random failures such as Data already appended were observed without this.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: do you need javadoc method prototype description for this private method? please clarify

tkaitchuck: As a rule no. But it provides a convenient place to explain what is going on in complex methods or ones that need to be used in a particular way. I just thought this method would benefit from one.

kmgowda: Thanks @tkaitchuck 
 sure .. I will add the descriptions for the code changes.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
tkaitchuck: I am confused by this. If there is data on the session, and then the channel becomes free. it could flush all of the sessions at the time it is freeing the channel. If it does not, then why would the session need to flush before new data is written to the same session? It could simply add the new data to the session and then flush thus reducing the number of messages going over the wire.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -154,34 +202,71 @@ private void validateAppend(Append append, Session session) {
...
+    private void continueAppend(ByteBuf data, ByteBuf out) {
+        bytesLeftInBlock -= data.readableBytes();
+        out.writeBytes(data);
+    }
+
+    private void breakFromAppend(ByteBuf data, ByteBuf pendingData,  boolean isAppend, ByteBuf out) {
+        if (isChannelFree()) {
+            return;
+        }
+        if (isAppend) {
```
tkaitchuck: This boolean is always a constant passed at the call site. 
It would be better to have two methods instead. They can both share the code below the if by making that into a separate method.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -154,34 +202,71 @@ private void validateAppend(Append append, Session session) {
...
+        return false;
+    }
+
+    private void continueAppend(ByteBuf data, ByteBuf out) {
+        bytesLeftInBlock -= data.readableBytes();
+        out.writeBytes(data);
+    }
+
+    private void breakFromAppend(ByteBuf data, ByteBuf pendingData,  boolean isAppend, ByteBuf out) {
+        if (isChannelFree()) {
```
tkaitchuck: This causes an early return which causes `data` not to be written. Fortunately it is never passed in such a case. But I think it means this should be factored differently. See my next comment.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: I have added the Javadoc for most of the methods in the file CommandEncoder class.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,64 +136,64 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+                if (isChannelOwner(append.getWriterId(), append.getSegment())) {
+                    if (bytesLeftInBlock > data.readableBytes()) {
+                        continueAppend(data, out);
+                    } else {
+                        ByteBuf dataInsideBlock = data.readSlice(bytesLeftInBlock);
+                        breakFromAppend(dataInsideBlock, data, true, out);
+                        flushAll(out);
+                    }
+                } else {
+                      session.append(data, out);
```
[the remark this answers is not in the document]
kmgowda: yes; Now, I have changed the code to take care of event count rather the data size.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -65,20 +67,66 @@
...
+    private static final int MAX_DATA_SIZE = 4 * 1024 * 1024; // 4MB
     private final Function<Long, AppendBatchSizeTracker> appendTracker;
     private final Map<Map.Entry<String, UUID>, Session> setupSegments = new HashMap<>();
     private final AtomicLong tokenCounter = new AtomicLong(0);
     private String segmentBeingAppendedTo;
     private UUID writerIdPerformingAppends;
     private int currentBlockSize;
     private int bytesLeftInBlock;
+    private final List<Session> pendingWrites = new ArrayList<>();
+    private long totalBytes = 0;
```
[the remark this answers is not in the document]
kmgowda: yes .. its changed now.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, its refactored now

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, its refactored now.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -154,34 +202,71 @@ private void validateAppend(Append append, Session session) {
...
+    private boolean startAppend(ChannelHandlerContext ctx, AppendBatchSizeTracker blockSizeSupplier, Append append, ByteBuf out) {
+        final int msgSize = append.getData().readableBytes();
+        currentBlockSize = msgSize;
+        if (blockSizeSupplier != null) {
+            currentBlockSize = Math.max(currentBlockSize, blockSizeSupplier.getAppendBlockSize());
+        }
+        bytesLeftInBlock = currentBlockSize;
+        segmentBeingAppendedTo = append.segment;
+        writerIdPerformingAppends = append.writerId;
+        if (ctx != null && currentBlockSize > msgSize) {
```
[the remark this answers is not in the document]
kmgowda: it was checking earlier this PR; so, I maintained as its. i think its should be OK of this check.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, i have done this change now.

kmgowda:
@andreipaduroiu 
       your review comments covered along with Tom's review comments. please let me know your further comments.

Requesting review comments from @fpj and @shrids too

kmgowda:
Conducted the System Tests: https://asd-nautilus-jenkins.isus.emc.com/job/pravega-branch-builder/1680/
its Successfully completed.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,64 +195,69 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
             Append append = (Append) msg;
             Session session = setupSegments.get(new SimpleImmutableEntry<>(append.segment, append.getWriterId()));
```
shrids: As per my current understanding entries are not removed from  `Map<Map.Entry<String, UUID>, Session> setupSegments `.  Since  `Session`  includes a `ByteBuf` this would have a non negligible size as there can be multiple sessions on a given connection.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,64 +195,69 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+            flushAll(out);
             writeMessage((WireCommand) msg, out);
         } else {
             throw new IllegalArgumentException("Expected a wire command and found: " + msg);
         }
     }
 
     private void validateAppend(Append append, Session session) {
+        if (append.getEventCount() <= 0) {
+            throw new InvalidMessageException("Invalid eventCount : " + append.getEventCount() + " in the append");
```
shrids: The writer id should be included.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -59,26 +61,131 @@
...
     @RequiredArgsConstructor
-    private static final class Session {
+    private final class Session {
+        private static final int MAX_BLOCK_SIZE = 1024 * 1024;  // 1MB
+        private static final int MAX_EVENTS = 500;
         private final UUID id;
-        private long lastEventNumber = -1L;
-        private int eventCount;
         private final long requestId;
+        private final ByteBuf buffer = buffer();
```
shrids: We should the api  `io.netty.buffer.Unpooled#buffer(int, int)` so as to explicitly set the `maxCapacity`

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,29 +95,40 @@ public Request processCommand(WireCommand command) throws Exception {
...
-            }
             segment = getSegment(writerId);
+            int sizeOfWholeEventsInBlock = blockEnd.getSizeOfWholeEvents();
+            ByteBuf appendDataBuf;
+            if (blockEnd.numEvents <= 0) {
+                throw new InvalidMessageException("Invalid number of events in block. numEvents : " + blockEnd.numEvents);
+            }
             if (blockEnd.getLastEventNumber() < segment.lastEventNumber) {
-                throw new InvalidMessageException("Last event number went backwards.");
+                throw new InvalidMessageException("Last event number went backwards." +
```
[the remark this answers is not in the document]
kmgowda: yes .. its added

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes.. its taken care now

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: it actually takes 256 bytes minimal.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: default minimum is 256 ; maximum is 1MB which taken care by flush method. 
we limit at the buffer allocation level, then it requires additional code take care of buffer border conditions.  The allocated which more than 1MB , gets cleared after the flush or flushAll .
The flushAll gets triggered for every 100ms; so in this design, the buffer accumulation is can happen for at max 100ms only.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
shrids: The reason for my question here is based on my understanding that the underlying byte Arrays size once increased is never reduced. i.e.the underlying `io.netty.buffer.ByteBufAllocator#heapBuffer()` never reduces/frees the underlying byteArray until `io.netty.util.ReferenceCounted#release()` is invoked.  Also, `io.netty.buffer.ByteBuf#clear` simply sets the write index and read index to 0.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes.. its done now..

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, now, buffer release is invoked during flush

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,29 +95,42 @@ public Request processCommand(WireCommand command) throws Exception {
                     ca.getExpectedOffset(), ca.getRequestId());
             break;
         case APPEND_BLOCK:
-            getSegment(((WireCommands.AppendBlock) command).getWriterId());
```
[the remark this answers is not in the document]
kmgowda: its added back now.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,100 +178,186 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
-                segmentBeingAppendedTo = append.segment;
-                writerIdPerformingAppends = append.writerId;
-                writeMessage(new AppendBlock(session.id), out);
-                if (ctx != null && currentBlockSize > (msgSize + TYPE_PLUS_LENGTH_SIZE)) {
-                    ctx.executor().schedule(new BlockTimeouter(ctx.channel(), tokenCounter.incrementAndGet()),
-                                            blockSizeSupplier.getBatchTimeout(),
-                                            TimeUnit.MILLISECONDS);
-                }
+
+            if (session.isFull(msgSize, append.getEventCount())) {
```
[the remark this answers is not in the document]
kmgowda: yes, its done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+        * Write/flush session's data to network channel.
+        * @param out   Network channel buffer.
+        */
+       private void flush(ByteBuf out) {
+            if (!isFree()) {
+                pendingWrites.remove(id);
+                if (pendingBytes > 0) {
+                    writeMessage(new AppendBlock(id), pendingBytes, out);
+                    for (Iterator<ByteBuf> iterator = pendingList.iterator(); iterator.hasNext(); ) {
+                        out.writeBytes(iterator.next());
```
shrids: Should we invoke ByteBuf.release()  here ?

kmgowda: No need, the netty takes care of it.

kmgowda:
@tkaitchuck 
As per your review comments, I have made both pravega server and client backward compatible as per below table.

'''
| Pravega Server 	| Pravega Client 	| Status  	|
|----------------	|----------------	|---------	|
| 0.6            	        | 0.6            	| Working 	|
| 0.6            	        | 0.5            	| Working 	|
| 0.5            	        | 0.6            	| working 	|
| 0.5            	        | 0.5            	| working 	|
'''

kmgowda:
The system tests are successfully completed:  pravega branch builder number : 1720

kmgowda:
@tkaitchuck 
    please let me know you further review comments ; otherwise requesting your approval for this PR.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,29 +95,43 @@ public Request processCommand(WireCommand command) throws Exception {
...
-            }
             segment = getSegment(writerId);
+            int sizeOfWholeEventsInBlock = blockEnd.getSizeOfWholeEvents();
+            ByteBuf appendDataBuf;
+            if (blockEnd.numEvents <= 0) {
+                throw new InvalidMessageException("Invalid number of events in block. numEvents : " + blockEnd.numEvents);
+            }
             if (blockEnd.getLastEventNumber() < segment.lastEventNumber) {
-                throw new InvalidMessageException("Last event number went backwards.");
+                throw new InvalidMessageException("Last event number went backwards." +
```
andreipaduroiu: Nit pick: Let's use `String.format`; it makes the code easier to read.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -59,6 +62,12 @@
  * need to be parsed out of individual messages. Notably this includes the event number of the last
  * event in the block, so that it can be acknowledged.
  *
+ * if the channel is not free, then each append is enqueued to the session's pending List.
```
andreipaduroiu: `If` (capital "I")

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -59,6 +62,12 @@
...
+ * This session pending list gets flushed under following conditions
```
andreipaduroiu: Are these conditions "AND" or "OR"? In other words, must all of them be true or does at least one be true?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+        }
+
+        /**
+         * Check if session is empty or not.
+         * @return true if there are any pending events; false otherwise.
+         */
+        private boolean isFree() {
+            return eventCount == 0;
+        }
+
```
andreipaduroiu: remove one of these empty lines

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+
+        /**
+         * queue the Append data to Session' list.
```
andreipaduroiu: Capital `Q`

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+         * @param data  data bytes.
+         * @param out   Network channel buffer.
+         */
+        private void write(ByteBuf data, ByteBuf out) {
+            pendingWrites.putIfAbsent(id, this);
```
andreipaduroiu: What if this is already present (with this instance or another one)?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+            if (data.readableBytes() > 0) {
+                pendingBytes += data.readableBytes();
+                pendingList.add(data);
+            }
+            conditionalFlush(out);
+        }
+
+        /**
+         * check the overflow condition and flush if required.
```
andreipaduroiu: Begin all your sentences with a capital letter.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+       /**
+        * Write/flush session's data to network channel.
+        * @param out   Network channel buffer.
+        */
+       private void flush(ByteBuf out) {
+            if (!isFree()) {
+                pendingWrites.remove(id);
+                if (pendingBytes > 0) {
+                    writeMessage(new AppendBlock(id), pendingBytes, out);
+                    for (Iterator<ByteBuf> iterator = pendingList.iterator(); iterator.hasNext(); ) {
```
andreipaduroiu: This looks inefficient (O(n^2)). Please loop through the list, and after that invoke `.clear`. That will make it O(n).

andreipaduroiu: It can be as simple as:
```
pendingList.forEach(out::writeBytes);
pendingList.clear();
```

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+            }
+        }
+
+        /**
+         * Write/flush given data to network channel.
+         * by writing the Append block End containing the given/input data and indicating the size of previous
+         * append block data.
+         * @param out   Network channel buffer.
+         */
+        private void flush(int sizOfWholeEvents, ByteBuf data, ByteBuf out) {
```
andreipaduroiu: 1. typo: `sizeOfWholeEvents`
2. Please have Javadoc for every argument.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+            writeMessage(new AppendBlockEnd(id, sizOfWholeEvents, data, eventCount, lastEventNumber, requestId), out);
+            eventCount = 0;
+        }
+    }
+
+    /**
+     * Write/flush Multi session's buffered data to network channel buffer
```
andreipaduroiu: This comment can fit on one line.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -72,13 +81,102 @@
...
+     * by iterating over all the pending sessions .
+     * @param out   Network Channel Buffer.
+     */
+    private void flushAll(ByteBuf out) {
+        if (!pendingWrites.isEmpty()) {
+            ArrayList<Session> sessions = new ArrayList<>(pendingWrites.values());
```
andreipaduroiu: Why do you make a copy and then iterate through the copy? There's no benefit to this. Did you intend to clear the `pendingWrites` map?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,100 +186,191 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+     */
+    private boolean isChannelFree() {
+        return writerIdPerformingAppends == null;
+    }
+
+    /**
+     * Check/Confirm the Network channel Ownership.
+     * The Network channel ownership is associated with Writer ID and segment to which writes are performed.
+     *
+     * @param writerID  Writers UUID.
```
andreipaduroiu: `Writer's`

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,100 +186,191 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+     * @param segment   Segment.
+     * @return true if channel is used by the given WriterId and Segment; false otherwise.
+     */
+    private boolean isChannelOwner(UUID writerID, String segment) {
+        return writerID.equals(writerIdPerformingAppends) && segment.equals(segmentBeingAppendedTo);
+    }
+
+    /**
+     * Start the Append.
+     * if the Append batch size tracker decides the block size
```
andreipaduroiu: Please begin all your sentences with a capital letter.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,100 +186,191 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+     * Append block is written on channel.
+     *
+     * @param ctx               ChannelHandlerContext.
+     * @param blockSizeSupplier Append batch size tracker.
+     * @param append            Append data.
+     * @param out               channel Buffer.
+     */
+    private void startAppend(ChannelHandlerContext ctx, AppendBatchSizeTracker blockSizeSupplier, Append append, ByteBuf out) {
+        final int msgSize = append.getData().readableBytes();
+        int blkSize = 0;
```
andreipaduroiu: `blockSize`

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -88,100 +186,191 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
...
+     * @return true if channel is used by the given WriterId and Segment; false otherwise.
+     */
+    private boolean isChannelOwner(UUID writerID, String segment) {
+        return writerID.equals(writerIdPerformingAppends) && segment.equals(segmentBeingAppendedTo);
+    }
+
+    /**
+     * Start the Append.
+     * if the Append batch size tracker decides the block size
+     * which is more than event message size then timeout will be initiated.
```
andreipaduroiu: How is a timeout initiated and how is it handled? Does it block the current call? Does it throw an exception? The code seems to indicate it will schedule a task with a delayed start time. 

Let's be specific about the behavior in the documentation.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
andreipaduroiu: I am referring to the name of the argument.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
kmgowda: yes. its done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: any one condition; Updated documentation accordingly.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes , its done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, its done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, its removed

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: it will always one instance of session ; when the flush happens its removed; otherwise on write its checked if its already in the list or not; if not add the pendingWrites

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: Thanks you.. its done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: thank you .. its changed and updated the javadoc too.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, its done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes; the flush removes them PendingWrites;   hence ; pending sessions are collected first and for each session flush is invoked. flush removes itself from the pendingWrites Map.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, its done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, I have taken care for all the comments.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: Its uses the existing timer behavior; The append starts with timer, before the timeout if the append completes then no need to flush any data; otherwise, all pending session data is flushed. I have updated the documentation in blocktimeout run method and other places too.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes,  its done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: yes, its done.

kmgowda:
@andreipaduroiu 
I have covered your review comments and synced with master also. please let me know if you have further review comments.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -95,29 +95,45 @@ public Request processCommand(WireCommand command) throws Exception {
...
+            if (currentBlock != null) {
+               if (!currentBlock.getWriterId().equals(writerId)) {
+                   throw new InvalidMessageException(
+                           String.format("Writer ID mismatch between Append Block and Append block End, " +
+                                   "Append block Writer ID : %s, Append block End Writer ID: %s",
+                                   currentBlock.getWriterId(), writerId));
+               }
+               if (sizeOfWholeEventsInBlock > currentBlock.getData().readableBytes() || sizeOfWholeEventsInBlock < 0) {
+                    throw new InvalidMessageException(
+                            String.format("nvalid SizeOfWholeEvents in block : %d, Append block data bytes : %d",
```
andreipaduroiu: "Invalid"

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -164,6 +180,7 @@ private ByteBuf getAppendDataBuf(WireCommands.AppendBlockEnd blockEnd, int sizeO
         return result;
     }
 
+
```
andreipaduroiu: Remove this extra line

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
kmgowda: Thanks .. its corrected now.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
kmgowda: yes, its done

kmgowda:
Successfully completed one more round of System tests , pravega branch builder build numbered: 1733

PR #3976, opened 2019-06-28
Issue 3974: Assign Resources to Kubernetes Pods in System Tests

### LINKED ISSUE #3974: Assign Resources to Kubernetes Pods in System Tests
**Problem description**
We need to assign a sufficient amount of resources to Pravega pods executing system tests. Currently, they have the default resources assigned by the system.

**Problem location**
Kubernetes System Test Framework.

Extract the resource-related changes from PR #3922 in a separate PR.

SrishT:
**Change log description**  
Currently, while executing System Tests, Pravega pods get only the default resources assigned by the system. Instead, we need to to assign a sufficient amount of resources to Pravega pods.

**Purpose of the change**  
Resolves #3974

**What the code does**  
The code assigns sufficient resources to the Segment Store and the Controller during the execution of System Tests.

**How to verify it**  
System Tests should pass


build.gradle
```diff
@@ -834,7 +834,7 @@ project('test:system') {
         systemProperty "pravegaImageName", System.getProperty("pravegaImageName", "pravega") // pravega image name
         systemProperty "bookkeeperImageName", System.getProperty("bookkeeperImageName", "bookkeeper") // bookkeeper image name
         systemProperty "imageVersion", System.getProperty("imageVersion") // pravega version.
-        // bookkeeper version defaults to pravega version.
+	// bookkeeper version defaults to pravega version.
```
RaulGracia: Minor: missing space to fix indentation.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -137,7 +137,8 @@ public String getID() {
                                                                                       .put("segmentStoreReplicas", segmentStoreCount)
                                                                                       .put("debugLogging", true)
                                                                                       .put("cacheVolumeClaimTemplate", pravegaPersistentVolumeSpec)
-                                                                                      //.put("options", props.getProperties())
+                                                                                      .put("controllerResources", getResources("2000m", "3Gi", "1000m", "1Gi"))
+                                                                                      .put("segmentStoreResources", getResources("2000m", "5Gi", "1000m", "3Gi"))
```
shrids: Couple of questions: 
- Checked the pravega-operator and observe that https://github.com/pravega/pravega-operator/blob/0.3.2/pkg/apis/pravega/v1alpha1/pravega.go#L71
`DefaultSegmentStoreLimitMemory` is `2Gi`, this is the maximum memory that will be allotted to the pod/container. Now we are changing this to `5Gi`. 
The unintended side affect of this change would be that the pod will not be scheduled on the cluster if the request resource is not available on the cluster.

- The operator uses `-XUseCGroupMemoryLimitForHeap` and does not specify -Xmx. Will this cause issues with the rocksdb memory management which is outside the jvm process?

- Should we add memory constraints for the bookie or is the default enough for system tests?

RaulGracia: @shrids regarding the first point, I think that Kubernetes will schedule the pods based on the `requests`, not the `limits`, right? In this case, the request is `3Gi`, which is what should be available in a node so Kubernetes can spawn the pod.

Regarding the second point, the memory consumed by RocksDB is is classified as `OffHeap` memory of the Java process. To control the memory usage of RocksDB, we need to set right values to `rocksdb.writeBufferSizeMB` and `rocksdb.readCacheSizeMB` in `config-properties`, according to the memory provisioned to the SegmentStore process. I don't see any change in the default values of these properties for the Segment Store, so they are using very small default values (`writeBufferSizeMB=64`, `rocksdb.readCacheSizeMB=8`) that are not going to make the process to tun out of memory. 

From the last point, maybe we need to set more realistic configuration values to Pravega, so we test the system with a similar configuration compared to the one used by end-users.

RaulGracia:
@SrishT apart from the resource provisioning for pods, could you please add these `options` to the configuration of Pravega? The intention is to set more realistic settings to the Segment Store and Controller, similar to the ones being used in QE team and proposed in the Provisioning model:
```
      bookkeeper.bkWriteQuorumSize: "3"
      bookkeeper.bkAckQuorumSize: "2"
      pravegaservice.writeBufferSizeMB: "512"
      pravegaservice.readCacheSizeMB: "2048"
      hdfs.blockSize: "67108864"
      pravegaservice.containerCount: "32"
      controller.containerCount: "32"
      controller.retention.bucketCount: "10"
      controller.service.asyncTaskPoolSize: "20"
      controller.retention.threadCount: "4"
```
@shiveshr @andreipaduroiu are you ok with these settings for our system tests?

shrids:
bookkeeper.bkAckQuorumSize: "2" would caused the `BookieFailoverTest` to fail.

RaulGracia:
@shrids fixed, thanks!

PR #3979, opened 2019-06-29
Issue 3975: Unify usage of RawClient and fix issues in existing RawClient calls from controller to segment store

### LINKED ISSUE #3975: Issues detected in implementation of RawClient usage in Controller's communication with SegmentStore
**Problem description**
There are multiple issues introduced in SegmentHelper in the implementation of RawClient:
1. RawClient can fail a request with ConnectionClosedException which should be handled in SegmentHelper. Any exception thrown from raw client should be handled as a retryable exception arising from unexpected connection drop 

2. In transformReply method, it transforms the exception for replies like `NoSuchSegment` and `TableSegmentNotEmpty` responses for segment store irrespective of the request type. 
So for example if segment store unexpectedly responded with `NoSuchSegment` for CreateSegment request, this should result in catastrophic failure indicating `Illegal response`. Instead the method currently translates it into WireCommandFailedException.SegmentNotFound which is a retryable and will result in retries from controller. 

3. Some methods in segmentHelper use RawClient while others use older way of using a connectionManager. We should consolidate all methods to use a single mechanism for communicating with segment stores. 

**Problem location**
SegmentHelper.java

Fix the above issues

shiveshr:
**Change log description**  
This PR fixes two issues in RawClient's usage in SegmentHelper. 
It also unifies a single way of making calls into SegmentStore. 

**Purpose of the change**  
Fixes #3975 

**What the code does**  
SegmentHelper had two ways of contacting segmentstore - some of the APIs used the old way of sending wireCommands and some apis used the RawClient. 
This PR unifies the usage of RawClient across all apis including table apis and transaction apis which were left out in earlier implementation. 

It also fixes two issues in the previous implementation-
1. RawClient can fail a request with ConnectionClosedException, ConnectionFailedException or AuthenticationException. Instead of translating these exceptions to controller specific exceptions that are retryable, these exceptions are getting thrown directly. Any exception thrown from raw client should be handled as appropriate retryable exception.

2. In `transformReply` method, it transforms the exception for replies like NoSuchSegment and TableSegmentNotEmpty responses for segment store irrespective of the request type.
So for example if segment store unexpectedly responded with NoSuchSegment for CreateSegment request, this should result in catastrophic failure indicating Illegal response. Instead the method currently translates it into WireCommandFailedException.SegmentNotFound which is a retryable and will result in retries from controller. 
So we now have two sets of expected responses that we check for -- EXPECTED_SUCCESS_REPLIES and EXPECTED_FAILING_REPLIES. 
If a response received is not present in either, then a generic non-retryable ConnectionFailedException (with an error message indicating that the response was of unexpected type) is thrown that will result in failure of processing

**How to verify it**  
unit tests updated


controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -173,83 +186,38 @@ public SegmentHelper(final ConnectionFactory clientCF, HostControllerStore hostS
...
                                                      final String stream,
                                                      final long segmentId,
                                                      final UUID txId,
                                                      String delegationToken) {
         final Controller.NodeUri uri = getSegmentUri(scope, stream, segmentId);
         final String transactionName = getTransactionName(scope, stream, segmentId, txId);
-
-        final CompletableFuture<UUID> result = new CompletableFuture<>();
         final WireCommandType type = WireCommandType.CREATE_SEGMENT;
-        final FailingReplyProcessor replyProcessor = new FailingReplyProcessor() {
```
RaulGracia: By deleting all the FailingReplyProcessor` per method, we simplify greatly this class; this is good. However, in the deleted code there were log messages that helped us to see the method executed (e.g., `createTransaction`), as well as the reason for the failure (e.g., `connectionDropped`). Are we loosing this kind of detailed info in our logs with this PR?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -271,60 +239,29 @@ private String getTransactionName(String scope, String stream, long segmentId, U
...
-            }
-        };
+        return sendRequest(connection, requestId, request)
+                .thenApply(r -> {
+                    handleReply(r, connection, transactionName, WireCommands.MergeSegments.class, type);
+                    if (r instanceof WireCommands.NoSuchSegment) {
+                        if (((WireCommands.NoSuchSegment) r).getSegment().equals(transactionName)) {
+                            return TxnStatus.newBuilder().setStatus(TxnStatus.Status.SUCCESS).build();
+                        } else {
+                            log.error("RequestId={} Commit Transaction: Source segment {} not found.", requestId, ((WireCommands.NoSuchSegment) r).getSegment());
```
RaulGracia: Please, use `TagLogger` here too.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1087,6 +629,23 @@ private void closeConnection(Reply reply, RawClient client) {
         }
     }
 
+    private <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(RawClient connection, long requestId, T request) {
+        return connection.sendRequest(requestId, request)
+                .exceptionally(e -> {
+                    Throwable unwrap = Exceptions.unwrap(e);
+                    if (unwrap instanceof ConnectionFailedException || unwrap instanceof ConnectionClosedException) {
+                        log.warn("RequestId={} Connection dropped", requestId);
```
RaulGracia: Use `TagLogger` here too.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1087,6 +629,23 @@ private void closeConnection(Reply reply, RawClient client) {
...
+                        throw new WireCommandFailedException(request.getType(), WireCommandFailedException.Reason.ConnectionFailed);
+                    } else if (unwrap instanceof AuthenticationException) {
+                        log.warn("RequestId={} Connection dropped", requestId);
+                        throw new WireCommandFailedException(request.getType(), WireCommandFailedException.Reason.AuthFailed);
+                    } else {
+                        log.error("RequestId={} request failed", e);
```
RaulGracia: Use `TagLogger` here too, please.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1097,23 +656,32 @@ private void closeConnection(Reply reply, RawClient client) {
...
+        Set<Class<? extends Reply>> expectedReplies = EXPECTED_SUCCESS_REPLIES.get(requestType);
+        Set<Class<? extends Reply>> expectedFailingReplies = EXPECTED_FAILING_REPLIES.get(requestType);
         if (expectedReplies != null && expectedReplies.contains(reply.getClass())) {
             log.info(reply.getRequestId(), "{} {} {}.", requestType.getSimpleName(), qualifiedStreamSegmentName,
                     reply.getClass().getSimpleName());
-            return true;
-        } else if (reply instanceof WireCommands.NoSuchSegment) {
-            throw new WireCommandFailedException(type, WireCommandFailedException.Reason.SegmentDoesNotExist);
+        } else if (expectedFailingReplies != null && expectedFailingReplies.contains(reply.getClass())) {
+            if (reply instanceof WireCommands.NoSuchSegment) {
```
RaulGracia: This method is important for us to log properly, as it handles the failures got from the segment store. Could you please add here necessary log messages to make it easier for us to find problems related to `SegmentHelper`?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: The logs are present in handleReply method. 
I have added more logs for expected failing cases as well. 
Thanks

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: Done

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -271,60 +239,29 @@ private String getTransactionName(String scope, String stream, long segmentId, U
...
-                        new WireCommandFailedException(new AuthenticationException(authTokenCheckFailed.toString()),
-                                type, WireCommandFailedException.Reason.AuthFailed)
-                );
-            }
-        };
+        return sendRequest(connection, requestId, request)
+                .thenApply(r -> {
+                    handleReply(requestId, r, connection, transactionName, WireCommands.MergeSegments.class, type);
+                    if (r instanceof WireCommands.NoSuchSegment) {
+                        if (((WireCommands.NoSuchSegment) r).getSegment().equals(transactionName)) {
```
andreipaduroiu: I suggest putting `(WireCommands.NoSuchSegment) r` into a local variable since you use it more than once.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -1087,6 +629,23 @@ private void closeConnection(Reply reply, RawClient client) {
...
 
+    private <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(RawClient connection, long requestId, T request) {
+        return connection.sendRequest(requestId, request)
+                .exceptionally(e -> {
+                    Throwable unwrap = Exceptions.unwrap(e);
+                    if (unwrap instanceof ConnectionFailedException || unwrap instanceof ConnectionClosedException) {
+                        log.warn(requestId, "Connection dropped");
+                        throw new WireCommandFailedException(request.getType(), WireCommandFailedException.Reason.ConnectionFailed);
+                    } else if (unwrap instanceof AuthenticationException) {
+                        log.warn(requestId, "Connection dropped");
```
RaulGracia: This message seems unrelated with the above exception.

shiveshr: fixed

RaulGracia:
@andreipaduroiu are you ok with merging this PR?

PR #3991, opened 2019-07-02
Issue 3989: Fix license headers

### LINKED ISSUE #3989: Some classes have old License Header. Clean up and use new license headers in them.
**Problem description**
Some classes continue to have an old license header. We should clean up and use a common license header across all classes. 

**Problem location**
Pravega

Use consistent license header across classes.

fpj:

**Change log description**  
* Fixes license headers that accidentally included the last paragraph of the ALv2, which was excluded by our legal.

**Purpose of the change**  
Fixes #3989 

**What the code does**  
There is no code change, only changes to license headers.

**How to verify it**  
It should build correctly as usual.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/index/HostIndex.java
+++ b/controller/src/main/java/io/pravega/controller/store/index/HostIndex.java
@@ -1,17 +1,11 @@
 /**
- * Copyright (c) 2017 Dell Inc., or its subsidiaries.
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
  *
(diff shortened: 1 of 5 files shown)


controller/src/main/java/io/pravega/controller/store/index/HostIndex.java
```diff
@@ -1,17 +1,11 @@
 /**
- * Copyright (c) 2017 Dell Inc., or its subsidiaries.
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
```
RaulGracia: Why in some files are we still keeping the year and in others we are removing it?

Issue #3899, opened 2019-06-11
Use external DNS name as segment store advertised endpoint


EronWright:
**Problem description**
For TLS to work end-to-end with hostname verification enabled, the segment store's advertised endpoint must be a DNS name.   The advertised value is determined at segment store startup (`init_kubernetes.sh`) and has a problem; the external IP is advertised, rather than the external DNS name as required for hostname verification.

**Problem location**
`docker/pravega/scripts/init_kubernetes.sh`

**Suggestions for an improvement**
1. The external DNS name is available as an annotation on the `Service` resource.   Use that value in lieu of the IP address that is contained in the `status` section.  The annotation is:
`external-dns.alpha.kubernetes.io/hostname`
Note that the value may have a period as a suffix; if it does, strip it.  [More info about the annotation](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/faq.md#how-do-i-specify-a-dns-name-for-my-kubernetes-objects).

kevinhan88:
Add tag 0.5.0 into this issue as this is required for nautilus 1.0

fpj:
@kevinhan88 we have marked this for `0.6.0`, and at this point we are only looking at critical issues for `0.5.0`.

EronWright:
@pbelgundi note that #4004 altered the logic to incorporate some hostname information.  I would suggest we continue with #3899 as planned, but the overall priority order of the available info should be (high- to low-priority):
1. `external-dns.alpha.kubernetes.io/hostname` annotation
2. `.status.loadBalancer.ingress[0].hostname` field
3. `.status.loadBalancer.ingress[0].ip` field

fpj:
I'm wondering about the traffic of clients internal to the cluster. Please help me understand what happens for the case that we have one or more clients running inside the kubernetes cluster.

If the external address is used, then it will be resolved to an external address. Consequently, is it the case that all traffic from the clients inside the cluster will be routed outside and then back in? Or is it only the resolution that happens externally?

I'm sure we have discussed this before, but I can't find the issue covering it. At the time, I believe the suggestion was for the segment store to advertise both an internal and an external endpoint. In the case that this change forces inside traffic to go outside, should this be done in conjunction with the change that advertises both endpoints? Are there other options?

EronWright:
[...]

PR #4004, opened 2019-07-04
Issue 4002: Use load balancer's external endpoint hostname if IP isn't available

### LINKED ISSUE #4002: pravega segmentstore could not obtain service endpoint in AWS environment
**Problem description**
When external connectivity is enabled in AWS environment provega-segmentstore is stuck in getting LoadBalancer external endpoint.

build 0.5.0-2291

**Problem location**
https://github.com/pravega/pravega/blob/master/docker/pravega/scripts/init_kubernetes.sh#L57

script should look for hostname from the service status block. ingress.ip or ingress.hostname
status:
  loadBalancer:
    ingress:
    - hostname: aws-elb-hostname  <-- elb name redacted. it was assigned

ravisharda:
**Change log description**  
Retrieve load balancer's external endpoint by host-name if IP is not available. 

**Purpose of the change**  
Fixes #4002 . 

**What the code does**  
If the load balancer's external endpoint's IP isn't available, retrieves the endpoint's hostname. IP address of the load balancer is available in PKS. For AWS, the endpoint address is accessed via hostname. See issue #4002 for additional details. 

**How to verify it**  
Verify that a Kubernetes cluster with external access enabled works for both PKS and AWS.

### DIFF
--- a/docker/pravega/scripts/init_kubernetes.sh
+++ b/docker/pravega/scripts/init_kubernetes.sh
@@ -55,6 +55,9 @@ init_kubernetes() {
                 echo "Trying to obtain LoadBalancer external endpoint..."
                 sleep 10
                 export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[0].ip" )
+                if [ -z "${PUBLISHED_ADDRESS}" ]; then
+                    export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[0].hostname" )
+                fi
                 export PUBLISHED_PORT=$( k8 "${ns}" "services" "${podname}" ".spec.ports[].port" )
             done
         elif [ "${service_type}" == "NodePort" ]; then


docker/pravega/scripts/init_kubernetes.sh
```diff
@@ -55,6 +55,9 @@ init_kubernetes() {
                 echo "Trying to obtain LoadBalancer external endpoint..."
                 sleep 10
                 export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[0].ip" )
+                if [ -z "${PUBLISHED_ADDRESS}" ]; then
```
fpj: Based on this comment from @Tristan1900 , is this change going to work on GKE? Does it work on PKS?

ravisharda: Yes, it'll work both for: 
- PKS and GKE where the IP address is available
- AWS where the endpoint is accessible using the hostname

This is because if the IP address is available, that is set as the published address (line 57). So, for PKS and GKE the IP address shall be set as published address. 

If the IP address is not available, the new `if` condition will set it to the hostname in lines 58-60 (which will work in AWS): 
https://github.com/pravega/pravega/blob/445f8c583380aeaa10c5e5e36457bcaaaccad93e/docker/pravega/scripts/init_kubernetes.sh#L58-L60

As for your second question, I've tested it on Azure Kubernetes Service (AKS) where the IP address is available, like for PKS and GKE. I built a new image and set up a new AKS cluster and deployed Pravega with external access enabled. I verified that a client application that writes events to the cluster and then reads them back works successfully. Subba had already tested the code with AWS.

gaddamas:
I have tested in nautilus cluster on AWS with PKS.

docker/pravega/scripts/init_kubernetes.sh
gaddamas: yes, I see it working in AWS.

EronWright:
In retrospect shouldn't the hostname be preferred to the IP address?   Imagine that a certain cloud were to populate both the IP and the hostname; the latter should be used.   My rationale is, an IP address will not work with SSL.

This PR would have been an ideal time to tackle #3899.   Since it has already been merged, I will make some notes there.

Issue #3900, opened 2019-06-12
Observed all readers failure for longevity test writes events on a few routing keys with auto scaling enabled


Ranjan-Padhi:
Run longevity test (3 writer & 6 Reader) IO after deploy pravega:0.5.0-2269.6f8a820 release

After running successfully for 10 hours reader group stopped and move to (0/6) after throwing `io.pravega.client.stream.TruncatedDataException` in IO manager

NOTE :- we have verified multiple times (more than 5 times) with same IO test and every time it failed after running for 6-10 hours

As per controller log Segment `longevity/smallKey/123.#epoch.71` has been deleted by controller `2019-06-11 17:50:20,688 25897868 [epollEventLoopGroup-2-2] INFO  i.p.controller.server.SegmentHelper - [requestId=2688449318144850290] deleteSegment longevity/smallKey/123.#epoch.71 SegmentDeleted` 

due to this segmentstore unable to get segment `2019-06-11 20:52:02,211 36682778 [core-12] WARN  i.p.s.s.h.h.PravegaRequestProcessor - [requestId=5415953760256] Segment 'longevity/smallKey/123.#epoch.71' does not exist and cannot perform operation 'readSegment'.` for which worker IO showing exception

segmentstore log
```
2019-06-11 20:52:00,193 36680760 [core-12] INFO  i.p.s.s.i.bookkeeper.BookKeeperLog - Log[5]: Truncated up to Sequence = 34359747088, LedgerId = 14719, EntryId = 8720.
2019-06-11 20:52:02,211 36682778 [core-12] WARN  i.p.s.s.h.h.PravegaRequestProcessor - [requestId=5415953760256] Segment 'longevity/smallKey/123.#epoch.71' does not exist and cannot perform operation 'readSegment'.
2019-06-11 20:52:06,289 36686856 [core-12] WARN  i.p.s.s.h.h.PravegaRequestProcessor - [requestId=12691628359680] Segment 'longevity/smallKey/123.#epoch.71' does not exist and cannot perform operation 'readSegment'.
```

IO Manager Log
```
172.24.80.2 - - [11/Jun/2019:20:51:48 +0000] "POST /v1/longevity/performance HTTP/1.1" 200 2298 "-" "Jersey/2.25.1 (HttpUrlConnection 1.8.0_131)" 2
INFO  [2019-06-11 20:52:03,384] io.pravega.longevity.testmanager.manager.TestPerformanceCollector: Received Error From task taskID : null
[...]

PR #4010, opened 2019-07-09
Draft PR for experiment, Issue 3900: Disable connection pooling

tkaitchuck:

**Change log description**  
Disable connection pooling.

**Purpose of the change**  
Experiment to see if pooling is related to #3900 

**What the code does**  
Disables pooling

**How to verify it**  
Run tests. 

DO NOT MERGE


shrids:
This disables connection pooling only for the SegmentMetaClient and StateSynchronizer.
Connection pooling is still enabled for append and read paths.

andreipaduroiu:
@tkaitchuck Is this PR still needed? I believe we've resolved issue 3900.

fpj:
@andreipaduroiu this is actually not for #3900, but to address the performance penalty that connection pooling has introduced. I'd like to have it so that we can turn it off in the case the performance hit due to connection pooling becomes a problem.

andreipaduroiu:
Ok, understood. Can we change the title to something that is more appropriate and that doesn't reference isuse 3900?

fpj:
I got confused with PR #3952 , we can close this one.

PR #4012, opened 2019-07-10
Issue 417: Verify stream exists before using it 

### LINKED ISSUE #417: Verify stream exists before using it
**Problem description**
If a stream does not exist, it fails when the client attempts to contact it. We could present a much cleaner error message if the client verified its existence first.

**Problem location**
ClkientFactoryImpl

Contact the controller to verify the stream exists before using it.

huide9:
**Change log description**  
In current ClientFactoryImpl, we create stream client without checking if the stream creation has been done. The PR verify the stream through newly added gRPC interface

**Purpose of the change**  
Fixes #417

**What the code does**  
* adding server side handler - ControllerServiceImpl
* adding client side interface - Controller
* call the newly added interface for verification, throw exception if stream not created 

**How to verify it**  
UT and integration test: StreamRecreationTest should pass without problem.


controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -129,6 +130,15 @@ public ControllerService(StreamMetadataStore streamStore, StreamMetadataTasks st
                   }, executor);
     }
 
+    public CompletableFuture<Boolean> checkStreamCreated(final String scope,
```
tkaitchuck: We are also going to want to have `checkStreamWritable` etc. I think the most general way to do this would be to have an API that sends the State back to the client. (Maybe simplified). Then the client can check that the stream is writable by checking that it is not `unknown` or `sealed`

Additionally for writers, I'm not sure we should fail-fast on `creating`. That is sortof a valid state, in that we can delay until it is complete transparently to the application.

huide9: thank you Tom for the comments; 
the previous impl is to reuse the `checkStreamExists` which already provided in controller; after discussion I think that: a stream is fully created indicated by stream stat is not UNKNOWN, CREATING  (list enum below);  `checkStreamExists` is insufficient for such verification; so the question is: " at what stream state(s), client can continue with the setup process" ?  the process contains the following operation: 
```
ConditionalOutputStream cond = condFactory.createConditionalOutputStream(segment, delegationToken, config.getEventWriterConfig());
SegmentMetadataClient meta = metaFactory.createSegmentMetadataClient(segment, delegationToken);
return new RevisionedStreamClientImpl<>(segment, in, out, cond, meta, serializer);
```

all the states:
```
public enum State {
    UNKNOWN,
    CREATING,
    ACTIVE,
    UPDATING,
    SCALING,
    COMMITTING_TXN,
    TRUNCATING,
    SEALING,
    SEALED;
  ...
```

shiveshr: What @tkaitchuck  is suggesting is to make available additional information about the stream apart from just whether it was created successfully or not. 
This will help us determine if the stream is sealed or open. 

Instead of exposing all the internal states of a stream to the client, my suggestion is as follows:
have two separate apis - `checkStreamExists` and `checkStreamIsWritable`. 
Both of these apis return boolean. And their implementation on controller simply fetches the state and based on the state determines if the stream is created and not sealed.

huide9: thank you @shivesh @Tom i'm going to remove the current API which returns Boolean value, instead, a new API will return (limited) stream state  - CREATING/ACTIVE/SEALING/SEALED/UNKNOWN,  API caller can then decide to continue/not create the client based on the returning state.

-- the above is based on discussion in slack; please comment for any further concerns.

huide9:
> Please add tests for your changes.

thank you Andrei, it will be added right after the changes made.

client/src/main/java/io/pravega/client/stream/Stream.java
```diff
@@ -24,6 +24,14 @@
  * Scalable (The rate of events in a stream can greatly exceed the capacity of any single host)
  */
 public interface Stream {
+    enum State {
```
tkaitchuck: I'm not sure this belongs here. This is a public interface exposed to our customers. It is also not used anywhere in this interface, so it would be confusing.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -178,15 +179,32 @@ public ClientFactoryImpl(String scope, Controller controller, ConnectionFactory
...
+    private String checkStreamReturnDelegation(String streamName, Segment segment) {
+        CompletableFuture<Stream.State>  streamState = controller.getStreamState(scope, streamName);
+        CompletableFuture<String> delegationToken = controller.getOrRefreshDelegationTokenFor(segment.getScope(), segment.getStreamName());
+        return streamState
+                .thenCombine(delegationToken,
+                        (state, delTok) -> {
+                            if (state != Stream.State.UNKNOWN && state != Stream.State.SEALED) {
+                                return delTok;
+                            } else {
+                                log.warn("Stream does not exist.");
```
tkaitchuck: This log message and exception would confuse someone who's stream is simply sealed. 
An appropriate message should be logged or thrown. (Maybe include the stream name and the attempted operation)

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -178,15 +179,32 @@ public ClientFactoryImpl(String scope, Controller controller, ConnectionFactory
...
+        String delegationToken = checkStreamReturnDelegation(streamName, segment);
+
         SegmentOutputStream out = outFactory.createOutputStreamForSegment(segment, segmentSealedCallBack,
                 config.getEventWriterConfig(), delegationToken);
         ConditionalOutputStream cond = condFactory.createConditionalOutputStream(segment, delegationToken, config.getEventWriterConfig());
         SegmentMetadataClient meta = metaFactory.createSegmentMetadataClient(segment, delegationToken);
         return new RevisionedStreamClientImpl<>(segment, in, out, cond, meta, serializer);
     }
 
+    private String checkStreamReturnDelegation(String streamName, Segment segment) {
```
tkaitchuck: It's not obvious from the name that this throws in the event the segment is sealed. Please add a Javadoc and include this information.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
huide9: removed

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
huide9: javadoc added

client/src/main/java/io/pravega/client/stream/Stream.java
huide9: this has been moved out to impl/StreamState

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -178,15 +179,38 @@ public ClientFactoryImpl(String scope, Controller controller, ConnectionFactory
...
+     * @param segment    Segment
+     * @return A delegation string used for RevisionedStreamClientImpl creation or throw exception if the Stream is not ready.
+     */
+    private String checkStreamReturnDelegation(String streamName, Segment segment) {
+        CompletableFuture<StreamState>  streamState = controller.getStreamState(scope, streamName);
+        CompletableFuture<String> delegationToken = controller.getOrRefreshDelegationTokenFor(segment.getScope(), segment.getStreamName());
+        return streamState
+                .thenCombine(delegationToken,
+                        (state, delTok) -> {
+                            if (state != StreamState.UNKNOWN && state != StreamState.SEALED) {
```
shiveshr: why are we checking for stream sealed?
delegation token could be requested for reads which can be performed on sealed streams as well.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -393,6 +394,14 @@ private SegmentRange convert(final String scope,
                 .thenApplyAsync(res -> TxnState.newBuilder().setState(TxnState.State.valueOf(res.name())).build(), executor);
     }
 
+    public CompletableFuture<StreamStatus> getStreamState(final String scope,
+                                                         final String stream) {
+        Exceptions.checkNotNullOrEmpty(scope, "scope");
+        Exceptions.checkNotNullOrEmpty(stream, "stream");
+        return streamStore.getState(scope, stream, true, null, executor)
```
shiveshr: this can give you states like SCALING etc. You need to perform a translation of such transient workflow induced states into ACTIVE

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
huide9: please see the first comment given by @tkaitchuck above.  
here the stream state test is for client creation, not for fetching the delegation token, the reason to put them in .thenCombine is to avoid have two Future condition check running serially.

tkaitchuck: I think we need to have multiple functions here.
Creating a writer for example should fail on a sealed stream. But a reader should not. 
A RevisionedStreamClient should also allow a sealed stream as it can still be read.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -178,15 +179,38 @@ public ClientFactoryImpl(String scope, Controller controller, ConnectionFactory
...
+                                return delTok;
+                            } else {
+                                throw new IllegalStateException("Stream is not ready: " + streamName);
```
tkaitchuck: This is a public facing error message. Let's make it as clear as possible. For example we should probably throw `InvalidStreamException` and include the stream name and why it is invalid. (For example "Does not exist in the scope: ")

controller/src/main/java/io/pravega/controller/server/ControllerService.java
huide9: thanks @shiveshr,  this has been addressed in latest commit

controller/src/main/java/io/pravega/controller/server/ControllerService.java
huide9: `getStreamState` added to replace previous `checkStreamExists`, so from client side we can get the stream state in detail.   "StreamState" has been added to describe the stream state.

huide9:
working on UT

PR #4016, opened 2019-07-10
Issue 4011: (SegmentStore) Making Commit Backlog Throttler more aggressive

### LINKED ISSUE #4011: Improve throttling due to memory commit backlog
**Problem description**
The Operation Processor's throttler uses multiple inputs to calculate the processing delay to introduce: the cache size, the tier-1 write stats and the memory commit queue size.

Issue #4008 described an OOM in the Segment Store triggered by the fact that the memory commit queue became too large (500-900 items) and it could not keep up with the incoming rate of appends.

The existing throttling for commit queue works like this:
- Triggers once the commit queue exceeds 300 items (one item is one completed Tier 1 write, and may contain one or more operations).
- Adds 4ms of delay for each operation above 300 in this queue.
    - For a queue of 900, this amounts to a mere 2.4s delay, which was not enough to throttle down requests.

**Problem location**
Segment Store/Throttler Calculator.

1. Increase the maximum throttle delay.
2. Make Throttler Calculator more aggressive

andreipaduroiu:
**Change log description**  
- Increased total allowed throttle delay from 10s to 25s
- Changed the Commit Backlog throttler to be more aggressive:
    - Kicks in after the commit queue exceeds 100 items (down from 300)
    - Adds 50ms of delay for each item count above the threshold (up from 4ms)
- Logging the reason for throttling
- Fixed the Segment Store Operation Processor metrics to properly record the memory commit time (also renamed it).

**Purpose of the change**  
Fixes #4011.

**What the code does**  
See Description.
Projected effect on throttling:
- Batching throttler will be unaffected because it is capped at 50ms
- Cache throttler is now free to throttle up to 25s (up from 10s) if the unevictable portion of the cache exceeds 100% of capacity.
- Commit backlog throttler
    - Size=100: Before 0s, Current 0s.
    - Size=300: Before 0s, Current 10s.
    - Size=500: Before 0.8s, Current 20s.
    - Size=800: Before 2s, Current 25s (capped at 25s).
    - Note that in order to reach 300, 500, etc, it would have to go through several iterations (there is one item per completed Tier1/BK write), and at each of those it would be subject to its own throttling delay and after each of those the commit queue would have to increase even more.
        - So to get to 800, presumably it would get to 50, 100, 150, 200, ..., 700, 750, 800 (just a sample; the steps would vary based on load and other conditions). This would imply a cumulative delay of over 6 minutes, so the cluster would need to be in a pretty bad shape to get there.

**How to verify it**  
Tests must pass.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/OperationProcessor.java
```diff
@@ -535,17 +537,15 @@ void commit(DataFrameBuilder.CommitArgs commitArgs) {
...
                     // Commit metadata updates.
-                    int updateTxnCommitCount = OperationProcessor.this.metadataUpdater.commit(commitArgs.getMetadataTransactionId());
+                    OperationProcessor.this.metadataUpdater.commit(commitArgs.getMetadataTransactionId());
 
-                    // Commit operations to memory. Note that this will block synchronously if the Commit Queue is full (until it clears up).
+                    // Queue operations for memory commit, which will be done asynchronously.
                     toAck.forEach(OperationProcessor.this.commitQueue::add);
 
                     this.highestCommittedDataFrame = addressSequence;
-                    metrics.memoryCommit(updateTxnCommitCount, memoryCommitTimer.getElapsed());
```
fpj: A couple of questions here:

1- Are you moving this to metrics update to line 419? What's the purpose of the change?
2- Can we test that the metrics are being computed correctly?

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -136,7 +136,7 @@
     public static final String OPERATION_PROCESSOR_DELAY_MILLIS = PREFIX + "segmentstore.container.operation_processor.delay_ms";            // Per-container Histogram
     public static final String OPERATION_COMMIT_LATENCY = PREFIX + "segmentstore.container.operation_commit.latency_ms";                     // Per-container Histogram
     public static final String OPERATION_LATENCY = PREFIX + "segmentstore.container.operation.latency_ms";                                   // Per-container Histogram
-    public static final String OPERATION_COMMIT_METADATA_TXN_COUNT = PREFIX + "segmentstore.container.operation_commit.metadata_txn_count";  // Per-container Histogram
+    public static final String OPERATION_COMMIT_MEMORY_COUNT = PREFIX + "segmentstore.container.operation_commit.memory_count";              // Per-container Histogram
```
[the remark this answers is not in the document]
andreipaduroiu: The UI is not listening for this metric. It is published for debugging purposes only and AFAIK the only way to get it is by manually querying the metrics DB.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/OperationProcessor.java
andreipaduroiu: 1. This was measuring the correct thing until we moved the cache writes into an async background loop. At this time, this metric will always record 0 since the only thing done here is to commit metadata to memory, which is not very interesting.
2. I can't verify that this works "correctly" because there's no way I can measure the commit times from the outside, nor can I figure out the write batches that go through.
I moved this metric recording to the commit queue processor to measure (once again) what it was originally meant to record.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ThrottlerCalculator.java
```diff
@@ -50,13 +50,13 @@
...
     @VisibleForTesting
-    static final int COMMIT_BACKLOG_COUNT_THRESHOLD = 300;
+    static final int COMMIT_BACKLOG_COUNT_THRESHOLD = 100;
 
     /**
      * Amount of time (millis) to increase throttling by for each incremental increase of the Commit Queue, over the threshold.
      */
     @VisibleForTesting
-    static final int THROTTLING_MILLIS_PER_COMMIT_OVER_LIMIT = 4;
+    static final int THROTTLING_MILLIS_PER_COMMIT_OVER_LIMIT = 50;
```
tkaitchuck: 50ms of delay seems really aggressive. Can we get away with 20?

andreipaduroiu: If you look at this PR's description, do those sample throttling delays look too aggressive? I think that if we are able to write 500 writes to BK but not able to cache those 500 writes, waiting 20 seconds is acceptable given the dire situation we might be in.

PR #4019, opened 2019-07-11
Issue 4020: Expose additional stream metadata records for internal records on StreamMetadataStore interface 

### LINKED ISSUE #4020: Expose additional stream metadata records for internal records on StreamMetadataStore interface
**Problem description**
StreamMetadataStore interface exposes apis corresponding to queries relevant for business logic. However, the underlying metadata that is stored has some additional records that are not exposed on the interface. 
For purposes of troubleshooter tool for metadata, we need to expose all such records so that the tool can check for consistency. 

**Problem location**
Controller Metadata Store

We do not want to unnecessarily expose them over StreamMetadataStore directly though. 
So create a new interface that extends StreamMetadataStore and has additional apis for exposing the new records.

anirudhkovuru:
**Change log description**  

- There are metadata records per stream that are not exposed via StreamMetadataStore interface as they are relevant to the metadata schema only. However, for purposes of troubleshooter tool for metadata, we need to expose all such records so that the tool can check for consistency. So this PR makes records such as the `HistoryTimeSeries`, `HistoryTimeSeriesRecord` and `SealedSegmentRecord` accessible on an extended interface.

**Purpose of the change**  
Fixes #4020 

**What the code does**  
- A new store interface called `ExtendedStreamMetadataStore` which extends the `StreamMetadataStore` interface has been created in the `controller.store.stream` package.
- The `AbstractStreamMetadataStore` class now implements the `ExtendedStreamMetadataStore` instead of the `StreamMetadataStore`.
- The three types of metadata stores; `InMemoryStreamMetadataStore`, `PravegaTablesStreamMetadataStore` and the `ZKStreamMetadataStore`; have implementations for the methods described in the `ExtendedStreamMetadataStore` interface.
- A new class `StreamStoreFactoryExtended` which extends the `StreamStoreFactory` has also been created.
- Allows for access to the most recent chunk of the `HistoryTimeSeries` and `HistoryTimeSeriesRecord` given the epoch number.
- Can do a check on whether a given segment is sealed or not.

**How to verify it**  

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
@@ -63,7 +63,7 @@ import java.util.stream.Collectors;
  * Implementation of create and update queries are delegated to the specific implementations of this abstract class.
  */
 @Slf4j
-public abstract class AbstractStreamMetadataStore implements StreamMetadataStore {
+public abstract class AbstractStreamMetadataStore implements ExtendedStreamMetadataStore {
     public static final Predicate<Throwable> DATA_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException;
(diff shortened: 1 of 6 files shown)


andreipaduroiu:
@shiveshr What is the status of this PR? It looks abandoned.

PR #4021, opened 2019-07-11
Issue 4017: (SegmentStore) Using ByteBuf for Appends

### LINKED ISSUE #4017: Segment Store should accept ByteBuf instead of byte[] for appends
**Problem description**
`StreamSegmentStore.append` requires a full `byte[]` object to be passed to it. This causes the `AppendProcessor` to make a copy of its `ByteBuf`and pass it to the Segment Store, which causes all in-flight appends to be stored twice in the Java heap (while processing). This can take anywhere from a few milliseconds to a few seconds, depending on how fast we can write to BookKeeper and RocksDB.

**Problem location**
Segment Store.

Make `StreamSegmentStore` and its implementations accept a `ByteBuf` (or thin wrapper) which does not require data duplication.

andreipaduroiu:
**Change log description**  
- Changed `StreamSegmentStore.append` to require a `BufferView` instead of `byte[]` for the append's payload.

**Purpose of the change**  
Fixes #4017.

**What the code does**  
- `BufferView`: defines read-only operations that can be performed on a memory buffer. `ArrayView` (and implicity `ByteArraySegment`) has been modified to extend it.
- `ByteBufWrapper` (in `shared:protocol`) implements `BufferView` and wraps around a Netty `ByteBuf`
    - Creates a read-only wrapper whose state is separate from the underlying `ByteBuf`
    - Exposes `retain` and `release` which directly call the underlying `ByteBuf` methods with the same name. See below why these are needed.
- `StreamSegmentAppendOperation`
    - Uses a `BufferView` instead of `byte[]` for its payload. During recovery this is set to a `ByteArraySegment` (since the data is already in the heap) and from `AppendProcessor` it is a `ByteBufWrapper` (around a Netty `ByteBuf`).
    - When created, invokes the `BufferView.retain`.
    - It now exposes a `close` method that invokes `release`. This is needed because the Segment Store may process this operation even after it has acknowledged it to the upstream caller, so we want to make sure the buffer is still available for this (i.e., we want to write it to the cache). If this hadn't been done, the `AppendProcessor` would immediately free the buffer once the append future completes.
- Other changes:
    - `RevisionDataOutputStream`: supports writing a `BufferView` to a stream
    - Plumbing code required due to the method signature change. This mostly affects test code, but some non-test code had to be touched as well.
    - `AppendProcessorTest`: require major changes since the `verify` method was attempting to compare to a freed ByteBuf (it was done well beyond when the append was done), so a different technique had to be used.

**How to verify it**  
Unit, integration and system tests must pass.
System test run 2164 passed.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/MemoryStateUpdater.java
```diff
@@ -135,15 +135,30 @@ void process(Operation operation) throws DataCorruptionException {
...
                     } else {
                         throw new DataCorruptionException(String.format("Unable to create a CachedStreamSegmentAppendOperation from operation '%s'.", operation), ex);
                     }
                 }
+
+                // Release the memory occupied by this StreamSegmentAppendOperation's BufferView - it has been processed
+                // and is no longer needed.
+                try {
+                    appendOp.close();
+                } catch (Throwable ex) {
```
tkaitchuck: This method doesn't actually have a throws in it's signature.

common/src/main/java/io/pravega/common/io/serialization/RevisionDataOutputStream.java
```diff
@@ -297,6 +299,27 @@ public void writeArray(byte[] array, int offset, int length) throws IOException
...
 
+    @Override
+    public void writeBuffer(BufferView buf) throws IOException {
+        if (buf instanceof ArrayView) {
+            // Invoke the appropriate method for this type, which is more optimized for arrays.
+            writeArray((ArrayView) buf);
+            return;
+        }
+
+        if (buf == null) {
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

common/src/main/java/io/pravega/common/util/BufferView.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param target The {@link OutputStream} to write to.
+     * @throws IOException If an exception occurred.
+     */
+    void copyTo(OutputStream target) throws IOException;
+
+    /**
+     * When implemented in a derived class, notifies any wrapped buffer that this {@link BufferView} has a need for it.
+     * Use {@link #release()} to do the opposite.
+     */
+    default void retain() {
```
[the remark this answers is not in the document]
andreipaduroiu: I actually had it like that before, but it's not so simple, and let me explain why.

On the append path in the Segment Store:
1. We create a `BufferView/ByteBufWrapper` in `AppendProcessor` which we pass down to the Segment Store.
2. The Segment Store may reject the request synchronously, or asynchronously (for various reasons). In each case it would throw an exception or return a failed CompletableFuture.
3. The Segment Store may accept the request, write it to Tier 1, ack it back to AppendProcessor and the asynchronously add it to the Cache.

We need to handle all these 3 cases. If we have `ByteBufWrapper/BufferView` implement AutoCloseable, then we could add some code to `AppendProcessor` (which creates it initially) to close it once the returned Future from the Store completes. Actually this is what `AppendProcessor` does anyway: it invokes `release` on the underlying `ByteBuf` after the future completes (which is correct, since `ByteBuf` sets the refCnt to 1 when it is created).

But that may be premature! In case 3 above, the BufferView may still be needed so it can be written to the cache when `AppendProcessor` invokes `release`, which means we won't be able to use it anymore.

As such, I decided the best way to handle this is the following:
- Have `retain` and `release`. Make them optional. Simply creating a `ByteBufWrapper` will not touch the underlying `ByteBuf` refCnt.
- If we end up in cases 1 or 2, all is good, since we haven't increased the refCnt.
- In case 3, when we create a `StreamSegmentAppendOperation` (which has a payload - `BufferView`), that class' constructor will invoke `retain` and I made that class `AutoCloseable`, in which case it will release the buffer.
    - The Segment Store will invoke `close` on that operation when it is completely done with it, which will ensure the buffer is released. It will also call `close` if it has already created it but had to abort the append for whatever reason.

common/src/main/java/io/pravega/common/io/serialization/RevisionDataOutputStream.java
```diff
@@ -297,6 +299,27 @@ public void writeArray(byte[] array, int offset, int length) throws IOException
         write(array, offset, length);
     }
 
+    @Override
+    public void writeBuffer(BufferView buf) throws IOException {
+        if (buf instanceof ArrayView) {
```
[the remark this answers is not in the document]
andreipaduroiu: Actually that's a good observation. `ByteArraySegment` does do the same in `copyTo`. I've removed `writeArray(ArrayView)` from this class and its interface and updated its callsites to use `writeBuffer`, which should have the same effect.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/MemoryStateUpdater.java
andreipaduroiu: Neither does `ByteBuf.release`, but it does throw an `IllegalStateException` if you release an already freed object. The point of this method is to clearly log cases (and label them appropriately) if we cannot release the memory we held for this class.

common/src/main/java/io/pravega/common/util/BufferView.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * Defines a generic read-only view of a readable memory buffer with a known length.
+ * For array-backed Buffers, see {@link ArrayView}.
+ */
+public interface BufferView {
```
tkaitchuck: Explain the ref counting and how one is expected to use this class in the javadoc above.

common/src/main/java/io/pravega/common/util/BufferView.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    default void retain() {
+        // Default implementation intentionally left blank. Any derived class may implement if needed.
+    }
+
+    /**
+     * When implemented in a derived class, notifies any wrapped buffer that this {@link BufferView} no longer has a
+     * need for it.
+     */
+    default void release() {
```
tkaitchuck: The semantics need to be better documented. Such as it is invalid to use the object after calling release.

common/src/main/java/io/pravega/common/util/BufferView.java
andreipaduroiu: I added more explanation there.

common/src/main/java/io/pravega/common/util/BufferView.java
andreipaduroiu: Same here.

tkaitchuck:
@fpj @andreipaduroiu 
This change (as is) interacts badly with connection pooling.
As mentioned previously connection pooling causes a large increase in the amount of data sent over the wire.
This change makes the client hold onto the raw network buffer without copying it for the entire time the data is in memory on the server.
The problem with this is that because it does not make a copy, it will keep a reference to the original buffer that came in over the wire. Which if because of connection pooling on the append path is ~7x larger than it needs to be, this is going to increase the server's memory requirements 7 fold as well.

andreipaduroiu:
@fpj @tkaitchuck I have resolved #4069 with PR #4085. As long both that and this PR go in, we should be good on this front.

andreipaduroiu:
#4069/#4085 has been merged in and I've updated this branch.

common/src/main/java/io/pravega/common/util/BufferView.java
```diff
@@ -0,0 +1,77 @@
+/**
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
fpj: Your IDE has apparently reformatted the license header:

```
Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.

Licensed under the Apache License, Version 2.0 (the "License");
you may not use this file except in compliance with the License.
You may obtain a copy of the License at

    http://www.apache.org/licenses/LICENSE-2.0
```

common/src/main/java/io/pravega/common/io/serialization/RevisionDataOutput.java
```diff
@@ -275,6 +260,19 @@ default void writeArray(ByteArraySegment segment) throws IOException {
...
+     * followed by {@link BufferView#getLength()} bytes representing the data from the {@link BufferView}.
+     *
+     * This can be read back using {@link RevisionDataInput#readArray()} as this method serializes the data in the same
+     * ways as {@link #writeArray}.
+     *
+     * @param buffer The {@link BufferView} to serialize. Can be null (in which case an Empty array will be deserialized
+     *               by {@link RevisionDataInput#readArray})).
+     * @throws IOException If an IO Exception occurred.
+     */
+    void writeBuffer(BufferView buffer) throws IOException;
```
fpj: We previously had a default implementation of `writeArray`, but `writeBuffer` doesn't. What's the reason for superseding and not providing a default implementation?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/ByteBufWrapper.java
```diff
@@ -0,0 +1,106 @@
+/**
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
fpj: Same point about the format about the license header.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/ByteBufWrapperTests.java
```diff
@@ -0,0 +1,130 @@
+/**
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
fpj: License header format.

common/src/main/java/io/pravega/common/util/BufferView.java
andreipaduroiu: OK. I'm pretty sure this is the cause why: #4083.

common/src/main/java/io/pravega/common/io/serialization/RevisionDataOutput.java
andreipaduroiu: `writeArray(ArrayView)` was straightforwad to implement (invoke `writeArray(segment.array(), segment.arrayOffset(), segment.getLength())`.

However `writeBuffer(BufferView)` is not, because `BufferView` may not wrap an array. It may wrap direct memory or multiple buffers, so the implementation may not be that simple.

Due to their striking similarities, I have decided to consolidate `writeArray(ArrayView)` and the new `writeBuffer(BufferView)` into the latter, since `ArrayView` extends `BufferView`. I didn't want to have 2 methods that are very much the same.

Also, moving the implementation into the actual class made testing easier. I don't think the `writeArray(ArrayView)` had any explicit tests - it was tested indirectly in the (only) implementing class' unit tests.

common/src/main/java/io/pravega/common/util/BufferView.java
TisonKun: @andreipaduroiu actually, we even have a checkstyle rule

```xml
<module name="RegexpHeader">
  <property name="header" value="/\*\*\n * Copyright \(c\)"/>
</module>
```

we can change the `value` to `/\*\n * Copyright \(c\)` and I volunteer to follow up the effort.

TisonKun: I just noticed that one is not blocked to send a pull request before assigned to the issue. I will give it a try later today.

PR #4028, opened 2019-07-15
Issue 4027: Add Metric for Container Recovery Time

### LINKED ISSUE #4027: Add Metric for Container Recovery Time
**Problem description**
Container recovery in the Segment Store may take a variable amount of time to complete. In some situations, container recoveries may take long time to complete, which could indicate a problem in the Pravega deployment (e.g., slow tier 2 storage). We need a metric to expose the time that container recoveries take, to identify and alert if there is some problem.

**Problem location**
Segment Store metrics.

Add new metric publishing container recovery times.

RaulGracia:
**Change log description**  
Added a new metric to keep track of container recovery times.

**Purpose of the change**  
Fixes #4027.

**What the code does**  
This PR adds a new metric (`segmentstore.container.recovery_time`) to track the time it takes for a Segment Store to perform a container recovery. This will be useful to create alerts if a recovery takes too long to complete.

**How to verify it**  
The new metric should be reported. All other tests must pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
@@ -9,7 +9,6 @@
  */
 package io.pravega.segmentstore.server;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import io.pravega.common.concurrent.ExecutorServiceHelpers;
 import io.pravega.segmentstore.server.logs.operations.CompletableOperation;
@@ -25,9 +24,6 @@ import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
-import lombok.AccessLevel;
-import lombok.Getter;
 
 import static io.pravega.shared.MetricsTags.containerTag;
 
@@ -43,17 +39,6 @@ public final class SegmentStoreMetrics {
      * until it is completed.
      */
     private static final OpStatsLogger GLOBAL_OPERATION_LATENCY = STATS_LOGGER.createStats(MetricsNames.OPERATION_LATENCY);
-
-    /**
-     * Global metric to publish the time elapsed for the recovery of Segment Store containers.
-     */
-    @VisibleForTesting
-    @Getter(AccessLevel.PACKAGE)
-    private final static AtomicReference<OpStatsLogger> RECOVERY_TIMES = new AtomicReference<>();
-
-    static {
-        RECOVERY_TIMES.set(STATS_LOGGER.createStats(MetricsNames.CONTAINER_RECOVERY_TIME));
-    }
 
     //region CacheManager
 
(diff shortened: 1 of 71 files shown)


kevinhan88:
Please add unit tests for this new metric.
How about we add some tags into the metric? e.g. container epoc? will it be beneficial for the purpose?

RaulGracia:
@kevinhan88 thanks for the comment. Actually, we want to how long are recoveries taking to spot any problem either in the Segment Stores or in the storage. The only tag that comes to my mind that could be interesting to add would the "host", to know where the Segment Store instance that did the recovery is placed.l But I'm not sure that such info is available at the point where we are reporting this metric.

kevinhan88:
> @kevinhan88 thanks for the comment. Actually, we want to how long are recoveries taking to spot any problem either in the Segment Stores or in the storage. The only tag that comes to my mind that could be interesting to add would the "host", to know where the Segment Store instance that did the recovery is placed.l But I'm not sure that such info is available at the point where we are reporting this metric.

Cool, the host is the default tag applied to all the metrics, so seems you're ready to go

segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
```diff
@@ -39,6 +43,17 @@
      */
     private static final OpStatsLogger GLOBAL_OPERATION_LATENCY = STATS_LOGGER.createStats(MetricsNames.OPERATION_LATENCY);
 
+    /**
+     * Global metric to publish the time elapsed for the recovery of Segment Store containers.
+     */
+    @VisibleForTesting
```
andreipaduroiu: I believe the only reason you made this an `AtomicReference` is because of testing. There is another way you can do this:
1. Keep this as a `static final` field.
2. Create a `protected` getter for it; make sure you only use this getter in your code.
3. In your test, create a subclass where you override the getter with whatever you want to supply there.


You need to optimize for production code, not for tests. It's OK if your test has a subclass that exposes some fields - it's much better than exposing them unnecessarily or adding unneeded complexity for all cases.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
```diff
@@ -313,4 +328,17 @@ public void truncate() {
...
     //endregion
+
+    //region RecoveryProcessor
+
+    /**
+     * RecoveryProcessor metrics.
+     *
+     * @param duration Time taken for a Segment Store instance to perform the recovery of containers.
+     */
+    public static void recoveryCompleted(long duration) {
```
andreipaduroiu: This should report the recovery per container. Some containers may have longer recovery times since they may have many more operations to recover. For example, in some of our past system tests, 2 containers had a 50,000 operation backlog, while the other 2 had less than 1000. Their recovery times were proportional to this. If we report this as a global metric, the actual durations may get smoothened out (especially if we have a lot of containers) so we won't be able to detect any spikes).

andreipaduroiu:
@RaulGracia Any updates on this PR?

RaulGracia:
@andreipaduroiu I have been doing other pressing issues, but it is in my mind. Will try to do some progress this week, thanks!

RaulGracia:
@andreipaduroiu I have retaken this PR. I have changed the type of metric, as I see a "gauge per container" a better way to expose the recovery times of containers (i.e., it is clearer than a histogram, given that recovery events are expected to be rare, conversely to event writes, for instance, for which a histogram makes sense). Apart from that, I have addressed your comments and tested the code of this PR in a real cluster. As you can see, the new metric is being correctly reported:

recovery_metric

I wrote data to a Stream with 12 segments, and as can be seen, 12 containers present much higher container recovery times after deleting a Segment store instance (all the containers are recovered).

PR #4030, opened 2019-07-16
Issue 4029: Enable RocksDB logging

### LINKED ISSUE #4029: Enable RocksDB logging
**Problem description**

The RocksDB log is not redirected to the Segmenstore log right now, it is nice to have RocksDB log displayed as well.

**Problem location**

RocksDBCache.java

Enable the RocksDB log

Tristan1900:
**Change log description**  
This PR enables RocksDB logging. User can specify different logging level for RocksDB, e.g. DEBUG, INFO, WARN, ERROR, etc. The default level is ERROR, which is suggested in production. 

**Purpose of the change**  
Fix #4029 

**What the code does**  
The code adds an additional field in `RocksDBConfig.java` for the RocksDB log level. In `RocksDBCache.java`, the code gets the log level from RocksDBConfig and pass that to the RocksDB configuration when initializing.  

**How to verify it**  
Specify `rocksdb.rocksDBLogLevel: "DEBUG"` in the `pravega: option` of the pravega manifest and deploy it. We should be able to see RocksDB log when `kubectl logs pravega-segmentstore`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -283,6 +283,8 @@ project('segmentstore:storage:impl') {
         compile group: 'org.rocksdb', name: 'rocksdbjni', version: rocksdbjniVersion
         testCompile project(':test:testcommon')
         testCompile project(path:':segmentstore:storage', configuration:'testRuntime')
+        testCompile group: 'org.slf4j', name: 'log4j-over-slf4j', version: slf4jApiVersion
+        testCompile group: 'ch.qos.logback', name: 'logback-classic', version: qosLogbackVersion
     }
     javadoc {
         dependencies {
--- a/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBCache.java
+++ b/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBCache.java
@@ -259,6 +259,7 @@ class RocksDBCache implements Cache {
                 .setTableFormatConfig(tableFormatConfig)
                 .setOptimizeFiltersForHits(true)
                 .setUseDirectReads(true);
+
         InfoLogLevel logLevel = translateRocksDBLogLevel(rocksDBLogLevel);
         Options logOptions = new Options().
                 setInfoLogLevel(logLevel).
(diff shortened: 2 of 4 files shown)


segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBCache.java
```diff
@@ -255,6 +259,36 @@ private Options createDatabaseOptions() {
...
+        Options logOptions = new Options().
+                setInfoLogLevel(logLevel).
+                setCreateIfMissing(true);
+
+        Logger logger = new Logger(logOptions) {
+            @Override
+            protected void log(InfoLogLevel infoLogLevel, String logMsg) {
+                switch (infoLogLevel) {
+                    case DEBUG_LEVEL:
+                        log.debug(logMsg);
```
andreipaduroiu: This is good, but I want more precision in these messages. This will log a generic message coming from `RocksDBCache`, however we have multiple instances of this class (one per Segment Container). Each such class has an `id` (which contains `Cache_<container-id>`). It would be very helpful in debugging to figure out where this came from.

Please change these `log.xyz` calls to something like `log.xyz("{}: {}", this.id, logMsg)` (`xyz` is the method you call on `log`).

build.gradle
```diff
@@ -283,6 +283,8 @@ project('segmentstore:storage:impl') {
         compile group: 'org.rocksdb', name: 'rocksdbjni', version: rocksdbjniVersion
         testCompile project(':test:testcommon')
         testCompile project(path:':segmentstore:storage', configuration:'testRuntime')
+        testCompile group: 'org.slf4j', name: 'log4j-over-slf4j', version: slf4jApiVersion
```
andreipaduroiu: I don't think these two are needed.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBCache.java
Tristan1900: Understood! will modify that.

build.gradle
Tristan1900: We can remove that, but I think this is the reason why there is no log printed when we run unit test on IDE yesterday. What do you think?

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBCache.java
```diff
@@ -95,6 +98,7 @@
         this.writeBufferSizeMB = config.getWriteBufferSizeMB() / MAX_WRITE_BUFFER_NUMBER;
         this.readCacheSizeMB = config.getReadCacheSizeMB();
         this.cacheBlockSizeKB = config.getCacheBlockSizeKB();
+        this.rocksDBLogLevel = config.getRocksDBLogLevel();
```
RaulGracia: I would suggest to add a `Statistics` attribute here and pass it to the db options, so we can access to the statistics of RocksDB (latency histograms, hits/misses, etc.). In debug mode, having this info periodically printed in the logs may be useful (e.g., every 5 minutes). See https://github.com/facebook/rocksdb/wiki/Statistics.

andreipaduroiu:
I see this PR is against 0.5. Let's remember to have one against master as well. I want to keep this new feature going forward.

Tristan1900:
Regarding the SIGSEGV issue that we discussed in the slack. Here is a [reference](https://github.com/facebook/rocksdb/issues/5595).

andreipaduroiu:
I think we have found enough information from this. IMO, there is no need to merge this into 0.5.

If we decide to add this into a subsequent release, let's reopen a new PR against master to add it.

Tristan1900:
@andreipaduroiu This will require to change the base image of segmentstore, which is also kind of risky to merge to 0.5 without some testing.

PR #4037, opened 2019-07-19
Issue 4022: Fix Controller gRPC list streams operation to return authorized streams

### LINKED ISSUE #4022: Controller gRPC 'list streams in scope' operation does not list authorized streams
**Problem description**
When auth is enabled, invoking the list streams operation of the Controller gRPC API returns an empty list even if the user is authorized to list the streams in the scope and has access to each of the streams in the scope. 

Note the following:
1. The Controller REST API lists streams properly. This issue is limited to the gRPC API of the controller. 
2. This issue does not occur if auth is disabled. 
3. This issue occurs even if the user has admin privileges (resource: `*`, permission: `READ_UPDATE`)
4. [PR 3941](https://github.com/pravega/pravega/pull/3941) fixed a related issue of improper auth resource string construction and was verified using tests that directly exercised the Controller gRPC API implementation on the server side. This issue remains and was reproduced when invoking the API via the client. 

**Problem location**
`io.pravega.controller.server.rpc.grpc.v1.ControllerServiceImpl.listStreamsInScope(...)`

ravisharda:
**Change log description**  
Fix the issue that prevents Controller gRPC list streams operation from returning authorized streams.

**Purpose of the change**  
Fixes #4022.

**What the code does**  
1. Caches the current gRPC context during list streams operation and reuses it during filtering.
2. Adds a new test integration test class that verifies Controller list streams operation via clients. (`io.pravega.test.integration.ControllerGrpcListStreamsTest`). Several new tests were added here. 
3. Adds new custom auth handler, custom credentials and custom principal classes for testing purposes. 
4. Renames some auth-related classes, refactors some existing code, and adds a new class in order to simplify the auth handling code and make it easier to understand and reason about its workings. There were some changes required in that set of classes for the current issue and refactoring the code a bit made it easier to make those changes.
5. Add additional logging statements to make it easier to debug auth issues.

**How to verify it**  
* Run tests in `io.pravega.test.integration.ControllerGrpcListStreamsTest` class.
* All other tests should also pass.


controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Initiates a new instance using the current {@link io.grpc.Context}.
+     *u
+     * @return an instance containing {@link Principal} and {@link AuthHandler} from the current {@link io.grpc.Context}
+     */
+    public static AuthContext current() {
+        Principal principal = AuthInterceptor.AUTH_CONTEXT_TOKEN.get();
+        AuthInterceptor serverAuthInterceptor = AuthInterceptor.INTERCEPTOR_OBJECT.get();
```
shiveshr: AuthInterceptor.AUTH_CONTEXT_TOKEN is static.
If there are multiple concurrent grpc requests, wouldnt we get the same auth context token?

ravisharda:
> I have one clarification question on the approach. Some explanation about what auth token token is etc in the description would be helpful.

@shiveshr As suggested, I've now added some explanation. 
* Added some comments on the `AuthInterceptor` class to explain the keys used to index the objects stored in the current `io.grpc.Context` object. 
* Added some comments `AuthInterceptor.current(...)` method about the intent of the `Context.Key.get()` calls.
I hope that helps.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
ravisharda: It represents only the key used to index the `Principal` object stored in the current gRPC context, which remains the same, and so it is static. The value - the `Principal` object - is a different object for each gRPC call, so concurrent requests don't share the values.

I've renamed it and the second key as well, and have now described them through comments as shown in the following fragment. Hopefully, the comments reveal the intent better.

https://github.com/pravega/pravega/blob/ec4427f74dc71e46414628d68208561d5ad9b76b/controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthInterceptor.java#L38-L48

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthInterceptor.java
```diff
@@ -21,25 +21,36 @@
...
+public class AuthInterceptor implements ServerInterceptor {
+
+    private static final String PRINCIPAL_KEY_NAME = "PravegaContext";
+    private static final String INTERCEPTOR_KEY_NAME = "InterceptorContext";
+
+    /**
+     * Represents the key used for indexing the {@link java.security.Principal} object stored in the current
+     * {@link io.grpc.Context}.
+     */
+    static final Context.Key<Principal> PRINCIPAL_OBJECT_KEY = Context.key(PRINCIPAL_KEY_NAME);
```
shiveshr: Ok i think i understand this now.
The comment says this is the `Key` for indexing the principal. 
So the key is static and constant. But from the context we can get the value against this key. 
So if i understand this correctly, we store principal and interceptor into the context against some known keys and then we can get the values from `AuthContext`.`key`.get().

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -398,32 +404,61 @@ public void createScope(ScopeInfo request, StreamObserver<CreateScopeStatus> res
...
     public void listStreamsInScope(Controller.StreamsInScopeRequest request, StreamObserver<Controller.StreamsInScopeResponse> responseObserver) {
         String scopeName = request.getScope().getScope();
-        RequestTag requestTag = requestTracker.initializeAndTrackRequestTag(requestIdGenerator.get(), "listStream", scopeName);
+        RequestTag requestTag = requestTracker.initializeAndTrackRequestTag(requestIdGenerator.get(),
+                "listStream", scopeName);
         log.info(requestTag.getRequestId(), "listStream called for scope {}.", scopeName);
+
+        final AuthContext ctx;
+        if (this.grpcAuthHelper.isAuthEnabled()) {
+            ctx = AuthContext.current();
```
shiveshr: will AuthContext.current return the last authContext?
What if there were multiple interceptors. Will there be multiple authContexts?
If so, this AuthContext.current will not retrieve all the contexts, right?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -398,32 +404,61 @@ public void createScope(ScopeInfo request, StreamObserver<CreateScopeStatus> res
...
+            ctx = null;
+        }
+
         authenticateExecuteAndProcessResults(
-                () -> this.authHelper.checkAuthorization(AuthResourceRepresentation.ofScope(scopeName),
-                        AuthHandler.Permissions.READ),
-                delegationToken -> controllerService.listStreams(
-                        scopeName, request.getContinuationToken().getToken(), listStreamsInScopeLimit)
+                () -> {
+                        String result = this.grpcAuthHelper.checkAuthorization(
```
shiveshr: why cant we simply call checkAuthorization as it internally takes the same context as it is computed here --> AuthContext.current()

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
```diff
@@ -0,0 +1,52 @@
...
+    private AuthHandler authHandler;
+
+    private AuthContext(Principal principal, AuthHandler authHandler) {
+        this.principal = principal;
+        this.authHandler = authHandler;
+    }
+
+    /**
+     * Initiates a new instance using the current {@link io.grpc.Context}.
+     *u
```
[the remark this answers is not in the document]
ravisharda: Fixed.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
```diff
@@ -0,0 +1,52 @@
...
+     * @return an instance containing {@link Principal} and {@link AuthHandler} from the current {@link io.grpc.Context}
+     */
+    public static AuthContext current() {
+        // Obtains the Principal object stored in the current gRPC Context
+        Principal principal = AuthInterceptor.PRINCIPAL_OBJECT_KEY.get();
```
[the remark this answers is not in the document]
ravisharda: The gRPC Context API is a bit awkward in this respect. While the key is static, when we invoke the `get()` on the key, the value is fetched from the current context in the gRPC scope, so the value changes for each gRPC request/scope.

https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.Key.html

I've also modified `AuthContext` to update the comments to explain this better.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHandlerManager.java
```diff
@@ -136,7 +136,7 @@ public boolean authorize(String resource, Principal principal, String credential
      * to inject and register custom auth handlers. Also, this method is idempotent.
      *
      * @param authHandler the {@code AuthHandler} implementation to register
-     * @Throws NullPointerException {@code authHandler} is null
+     * @throws NullPointerException {@code authHandler} is null
```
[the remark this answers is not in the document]
ravisharda: Personally, I am of the opinion (in agreement with your comment) that an `IllegalArgumentException` - and not `NullPointerException` - should be thrown when an argument is illegal (including the case where `null` is not allowed). `NullPointerException` on the other typically indicates that a variable inappropriately turned out to be a `null`, such as when a `null` is dereferenced. 

There are many who strongly advocate using a NullPointerException instead, for asserting that 'null' isn't allowed as an argument. Joshua Bloch in his famous book "Effective Java" states the following: 
> Arguably, every erroneous method invocation boils down to an illegal argument or state, but other exceptions are standardly used for certain kinds of illegal arguments and states. If a caller passes null in some parameter for which null values are prohibited, convention dictates that NullPointerException be thrown rather than IllegalArgumentException. Similarly, if a caller passes an out-of-range value in a parameter representing an index into a sequence, IndexOutOfBoundsException should be thrown rather than IllegalArgumentException.

A lot of code in our codebase throws NullPointerException to indicate that `null` is not allowed. For example, io.pravega.common.Exceptions.checkNotNullOrEmpty(), which is widely used, throws a NullPointerException if the argument is null. Similarly, `com.google.common.base.checkNotNull() throws a NullPointerException to reject a null input.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthHandlerManager.java
```diff
@@ -146,6 +146,7 @@ public synchronized void registerHandler(AuthHandler authHandler) {
     /**
      * Loads the custom implementations of the AuthHandler interface dynamically. Registers the interceptors with grpc.
      * Stores the implementation in a local map for routing the REST auth request.
+     *
```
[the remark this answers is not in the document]
ravisharda: Done.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthInterceptor.java
ravisharda: Yes, exactly! The gRPC Context API is a bit awkward in this respect. 

(I've explained the same here: https://github.com/pravega/pravega/pull/4037#discussion_r308529649)

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: The gRPC Context isn't available at this stage of the operation and therefore the `Principal` and `AuthInterceptor` (erstwhile `PravegaInterceptor`) objects cannot be retrieved from the gRPC Context at this point. Since those values were null, we were defaulting to assuming the user doesn't have access to the stream. This was the root cause for the list streams filtering not returning any streams in scope with auth enabled. 

To avoid the issue, we now cache the relevant `Principal` and `AuthHandler` objects in the `AuthContext` object during the commencement of the operation when the gRPC Context is available. We then use the cached `AuthContext` for filtering later in the same operation.

Hope that clarifies.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+        // Obtains the AuthInterceptor object stored in the current gRPC Context
+        AuthInterceptor serverAuthInterceptor = AuthInterceptor.AUTH_INTERCEPTOR_OBJECT_KEY.get();
```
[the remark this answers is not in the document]
ravisharda: https://github.com/pravega/pravega/pull/4037#discussion_r308529649 addresses this comment too.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: The logic here works in the presence of multiple auth interceptors. Let me explain this in detail. 

`AuthContext.current()` will return an object containing the `Principal` and `AuthHandler` objects. The `Principal` is directly retrieved from the current gRPC Context. The `AuthHandler` is retrieved from the `AuthInterceptor` instance present in the current gRPC Context. 

There may be multiple `AuthInterceptor` instances registered in the system. One `AuthInterceptor` is registered per `AuthHandler` implementation found in the classpath. This happens in `io.pravega.controller.server.rpc.auth.registerInterceptors()` when the gRPC Server is constructed.  

At runtime, for each gRPC call, the `Principal` and `AuthInterceptor` in turn are set by one of the multiple `AuthInterceptors`  that might be registered in the system. The one that will do it is the one that has an `Authhandler` that matches the auth method (`Bearer`, `Basic`, etc.) specified by the client. Please see the code below: 
https://github.com/pravega/pravega/blob/7a923ed9dac4688ff1eed13a11d876292f6f90c2/controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthInterceptor.java#L59

Hope this clarifies. 

Also, I had added a new test to verify a scenario involving multiple AuthHandler plugins. It is here: https://github.com/pravega/pravega/blob/7a923ed9dac4688ff1eed13a11d876292f6f90c2/test/integration/src/test/java/io/pravega/test/integration/ControllerGrpcListStreamsTest.java#L127

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
```diff
@@ -0,0 +1,54 @@
...
+import java.security.Principal;
+
+/**
+ * A gateway to objects held in gRPC {@link io.grpc.Context} by the {@link AuthInterceptor}, for the current gRPC
+ * request.
+ */
+public class AuthContext {
+
+    @Getter
+    private Principal principal;
```
andreipaduroiu: `final`

and below too.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Getter
+    private AuthHandler authHandler;
+
+    private AuthContext(Principal principal, AuthHandler authHandler) {
```
andreipaduroiu: `@NonNull` for both arguments or explain in your JavaDoc which arguments can be null.

controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
```diff
@@ -13,51 +13,66 @@
...
+    public String checkAuthorization(String resource, AuthHandler.Permissions expectedLevel, AuthContext ctx) {
+        if (isAuthorized(resource, expectedLevel, ctx)) {
             return "";
         } else {
-            throw new RuntimeException(new AuthorizationException("Access not allowed"));
+            if (ctx == null) {
+                log.warn("AuthContext is null");
+            }
+            String message = String.format("Principal [%s] not allowed [%s] access for resource [%s]",
+                    ctx != null ? ctx.getPrincipal() : null, expectedLevel.toString(), resource);
```
andreipaduroiu: no need to call `.toString` here

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
```diff
@@ -573,4 +614,50 @@ private static File createAuthFile() {
         private final static String SCOPE1_STREAM2_READ = "authSc1Str2readonly";
         private final static String SCOPE1_STREAM1_LIST_READ = "scope1stream1lr";
     }
+
+    static class ResultObserver<T> implements StreamObserver<T> {
+        private T result = null;
+        private Throwable error;
+        private final AtomicBoolean completed = new AtomicBoolean(false);
```
andreipaduroiu: You only update this inside a `synchronized` block. Either remove the `synchronized` or make this a simple boolean.

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
```diff
@@ -573,4 +614,50 @@ private static File createAuthFile() {
...
+                this.notifyAll();
+            }
+        }
+
+        @SneakyThrows
+        public T get() {
+            synchronized (this) {
+                while (!completed.get()) {
+                    try {
+                        this.wait();
```
andreipaduroiu: It's bad practice to block while holding a lock.

Similarly above, you invoke `notifyAll` while holding the lock.

test/integration/src/main/resources/META-INF/services/io.pravega.auth.AuthHandler
```diff
@@ -0,0 +1,10 @@
+#
```
andreipaduroiu: Are these files needed?

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
fpj: What about making it a future and completing the future upon the call to `onCompleted`? This way we can block on the call to `get()` (Line 645), which would simplify the implementation of that method and of others as the `this.wait` call is no longer necessary.

test/integration/src/main/java/io/pravega/test/integration/auth/customplugin/CustomAuthHandler.java
```diff
@@ -0,0 +1,62 @@
+/**
+ * Copyright (c) 2019 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.test.integration.auth.customplugin;
```
fpj: If this is for test purposes, what about calling it `TestAuthHandler` rather than `CustomAuthHandler`?

test/integration/src/main/java/io/pravega/test/integration/auth/customplugin/CustomPrincipal.java
```diff
@@ -0,0 +1,35 @@
...
+package io.pravega.test.integration.auth.customplugin;
+
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+import java.security.Principal;
+
+@Slf4j
+@ToString
+public class CustomPrincipal implements Principal {
```
fpj: Same comment about changing `Custom*`  to `Test*`.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
ravisharda: Done.

controller/src/main/java/io/pravega/controller/server/rpc/auth/AuthContext.java
ravisharda: Nulls are allowed for both the params. I've added JavaDocs comments to express that, as suggested.

controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
ravisharda: Fixed.

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
ravisharda: I liked the idea of using the future for the purpose and have modified the code to use it instead.  That change addresses both the comments. Thanks for the suggestion - the code looks much more simpler now!

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
ravisharda: The modified code doesn't use these.

test/integration/src/main/resources/META-INF/services/io.pravega.auth.AuthHandler
ravisharda: Yes, these files are necessary for the custom test auth handler to get loaded via the service loader.

test/integration/src/main/java/io/pravega/test/integration/auth/customplugin/CustomAuthHandler.java
```diff
@@ -0,0 +1,62 @@
+/**
+ * Copyright (c) 2019 Dell Inc., or its subsidiaries. All Rights Reserved.
```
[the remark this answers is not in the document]
ravisharda: Done.

test/integration/src/main/java/io/pravega/test/integration/auth/customplugin/CustomAuthHandler.java
ravisharda: Sure, I've modified the class names accordingly.

test/integration/src/main/java/io/pravega/test/integration/auth/customplugin/CustomPrincipal.java
ravisharda: Done.

controller/src/test/java/io/pravega/controller/server/rpc/auth/ControllerGrpcAuthFocusedTest.java
```diff
@@ -627,37 +625,16 @@ public void onNext(T value) {
...
-                    }
-                }
-            }
-            if (error != null) {
-                throw error;
-            } else {
-                return result;
-            }
+        public T get() throws ExecutionException, InterruptedException {
+            return future.get();
```
fpj: Would it make sense to use `join` instead of `get` so that we don't have to declare that it throws?

ravisharda: Yes, that's a good idea. Using `join()` instead of `get()` would make the code simpler here. I've modified it as suggested.

PR #4042, opened 2019-07-19
Issue 3645: Make direct reads configurable in RocksDB

### LINKED ISSUE #3645: Make direct reads configurable in RocksDB
**Problem description**
According to RocksDB [documentation](https://github.com/facebook/rocksdb/wiki/Direct-IO), enabling direct reads is desirable because i) it avoids using the OS page cache, which requires extra copies of the data, and ii) RocksDB should have a better knowledge of the logical semantics of the data than OS (e.g., replacement policies). For this reason, `RocksDBCache.java` has by default `setUseDirectReads(true)`. However, direct reads are not always available for all OS. In this case, we should expose this parameter in the configuration to disable direct reads when this option is not available in the host OS. 

**Problem location**
`RocksDBCache` and config.

Make RocksDB direct reads configurable.

### LINKED ISSUE #4041: Cherry pick PR 3645 to r0.5
**Problem description**
We need to port #3645 to `0.5` to be able to configure direct reads in RocksDB. However, we need to keep the default value to true, so the configuration for `0.5` remains the same.

**Problem location**
RocksDB config.

Cherry pick the PR and set default value `rocksdb.directReads=true`.

RaulGracia:
**Change log description**  
* Enables RocksDB direct reads to be configurable via rocksdb.directReads
* Sets the default value rocksdb.directReads=true

**Purpose of the change**  
Fixes #3645 

**What the code does**  
Makes RocksDB direct reads configurable, as not all OS/environments support this option. See also PR #3647 .

**How to verify it**  
No code changes, all tests should be passing as before.

### DIFF
--- a/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBConfig.java
+++ b/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBConfig.java
@@ -25,7 +25,7 @@ public class RocksDBConfig {
     public static final Property<Integer> WRITE_BUFFER_SIZE_MB = Property.named("writeBufferSizeMB", 64);
     public static final Property<Integer> READ_CACHE_SIZE_MB = Property.named("readCacheSizeMB", 8);
     public static final Property<Integer> CACHE_BLOCK_SIZE_KB = Property.named("cacheBlockSizeKB", 32);
-    public static final Property<Boolean> DIRECT_READS = Property.named("directReads", false);
+    public static final Property<Boolean> DIRECT_READS = Property.named("directReads", true);
     private static final String COMPONENT_CODE = "rocksdb";
 
     //endregion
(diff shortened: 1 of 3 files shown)


segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBConfig.java
```diff
@@ -25,6 +25,7 @@
     public static final Property<Integer> WRITE_BUFFER_SIZE_MB = Property.named("writeBufferSizeMB", 64);
     public static final Property<Integer> READ_CACHE_SIZE_MB = Property.named("readCacheSizeMB", 8);
     public static final Property<Integer> CACHE_BLOCK_SIZE_KB = Property.named("cacheBlockSizeKB", 32);
+    public static final Property<Boolean> DIRECT_READS = Property.named("directReads", true);
```
andreipaduroiu: This is different than `master`. We should make them the same.

https://github.com/pravega/pravega/blob/master/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBConfig.java#L28

fpj: We want to keep the behavior the same as currently have in `0.5`, that's the reason we changing the default, so that we match the current behavior.

andreipaduroiu: @fpj told me offline he requested this. So I withdraw my comment.

fpj: For completeness, I have requested it this way because we did a good amount of testing with this behavior, and I'd rather not change at this point. For `master`, we have a different default value.

PR #4043, opened 2019-07-19
Issue 3858: Handle WriteConflict or StoreException during workflow submission

### LINKED ISSUE #3858: Unwrap real exception from RetriesExhaustedException in Pravega Table Store Helper
**Problem description**
PravegaTableStoreHelper retries calls to segment store in case of store exceptions and can throw RetriesExhaustedException. We should unwrap retries exhausted and throw real exception. 

Also, during retries, an update may have been retried while the previous attempt may have succeeded whereby we could get WriteConflict thrown back from the server. This is not a problem for background workflows as write conflict is retryable and the whole workflow is retried. 

However, updates also happen during submission of a workflow. If there is an error there, we will not post the event corresponding to workflow submission. Though the caller will get the failure and can retry. Also, the index is added so during failover the event is posted. But if neither is performed then no other workflow of the requested type could be submitted. 

**Problem location**
PravegaTableStoreHelper

Fix the issue

shiveshr:
**Change log description**  
1. Unwrap real exception from RetriesExhaustedException thrown from PravegaStoreHelper. 
2. While performing a table update, there could be retries performed by the store helper. This can result in WriteConflict being thrown if the update had succeeded on the server. Or we can also get StoreConnectionException and we wouldnt know what happened on the server. These should be handled by callers - during workflow submission, we should submit event for processing in case we encounter above failures. 

**Purpose of the change**  
Fixes #3858 

**What the code does**  
 PravegaTableStoreHelper retries calls to segment store in case of store exceptions and can throw RetriesExhaustedException. We should unwrap retries exhausted and throw real exception.

Also, during retries, an update may have been retried while the previous attempt may have succeeded whereby we could get WriteConflict thrown back from the server. This is not a problem for background workflows as write conflict is retryable and the whole workflow is retried.

However, updates also happen during submission of a workflow. If there is an error there, we will not post the event corresponding to workflow submission. Though the caller will get the failure and can retry. Also, the index is added so during failover the event is posted. But if neither is performed then no other workflow of the requested type could be submitted.

**How to verify it**  
Unit tests added. Also added test for all of pravega table store helper methods.


controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -632,11 +632,29 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                                   e -> {
+                                       toThrow.set(e);
+                                       return Exceptions.unwrap(e) instanceof StoreException.StoreConnectionException ||
+                                               Exceptions.unwrap(e) instanceof StoreException.WriteConflictException;
+                                   }, null))
                            .thenCompose(t -> RetryHelper.withIndefiniteRetriesAsync(() -> writeEvent(event), e -> { }, executor)
                                                         .thenCompose(v -> streamMetadataStore.removeTaskFromIndex(context.getHostId(), id))
-                                                        .thenApply(v -> t));
+                                                        .thenApply(v -> {
+                                                            if (toThrow.get() != null) {
```
shrids: if `streamMetadataStore.removeTaskFromIndex(...)` throws an exception we would never know that `futureSupplier.get()` faced an exception right?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStoreHelper.java
```diff
@@ -529,6 +538,14 @@ public void invalidateCache(String table, String key) {
                 e -> {
                     Throwable unwrap = Exceptions.unwrap(e);
                     return unwrap instanceof StoreException.StoreConnectionException;
-                }, NUM_OF_RETRIES, executor);
+                }, numOfRetries, executor)
+                .exceptionally(e -> {
```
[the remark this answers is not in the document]
shiveshr: No, exceptionally gives you a callback to handle(or throw back) the exception from previous completion stage

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -632,11 +632,29 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+        // (which can also occur if we had retried on a store exception), we will still post the event because we dont know 
+        // if our update succeeded or not. Posting the event is harmless. If the update had succeeded, then the event will 
+        // be used for processing. If the update had failed, the event will be discarded. We will throw the exception 
+        // that we thrown from running futureSupplier.
+        AtomicReference<Throwable> toThrow = new AtomicReference<>();
         return streamMetadataStore.addRequestToIndex(context.getHostId(), id, event)
-                           .thenCompose(v -> futureSupplier.get())
+                           .thenCompose(v -> Futures.exceptionallyExpecting(futureSupplier.get(), 
+                                   e -> {
+                                       toThrow.set(e);
```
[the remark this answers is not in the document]
shiveshr: It doesn’t matter. We will unwrap it wherever it is handled

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: I don’t follow. It is inside a completionStage callback so our future will complete exceptionally. What is your concern exactly?

shrids: ok, got it.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -632,11 +632,30 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
     @VisibleForTesting
     <T> CompletableFuture<T> addIndexAndSubmitTask(ControllerEvent event, Supplier<CompletableFuture<T>> futureSupplier) {
         String id = UUID.randomUUID().toString();
+        // We first add index and then call the metadata update.
+        // In case, while trying to perform a metadata update, if we get connection exception or write conflict exception
```
fpj: Here is slight rephrase of this comment:

```
//  While trying to perform a metadata update, upon getting a connection exception or a write conflict exception 
// (which can also occur if we had retried on a store exception), we will still post the event because we
//  don't know whether our update succeeded. Posting the event is harmless, though. If the update
// has succeeded, then the event will be used for processing. If the update had failed, then the event
// will be discarded. We will throw the exception that we thrown from running futureSupplier.

```

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -632,11 +632,30 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                                   }, null))
+                           .thenCompose(t -> RetryHelper.withIndefiniteRetriesAsync(() -> writeEvent(event),
+                                   e -> log.warn("writing event failed with {}", e.getMessage()), executor)
                                                         .thenCompose(v -> streamMetadataStore.removeTaskFromIndex(context.getHostId(), id))
-                                                        .thenApply(v -> t));
+                                                        .thenApply(v -> {
+                                                            if (toThrow.get() != null) {
+                                                                throw new CompletionException(toThrow.get());
+                                                            } else {
+                                                                return t;
```
fpj: I don't understand why we are returning `t` here.

controller/src/test/java/io/pravega/controller/store/stream/PravegaTablesStoreHelperTest.java
```diff
@@ -0,0 +1,183 @@
...
+    private SegmentHelper segmentHelper;
+    private AuthHelper authHelper;
+    private PravegaTablesStoreHelper storeHelper;
+
+    @Before
+    public void setup() throws Exception {
+        executor = Executors.newScheduledThreadPool(5);
+        segmentHelper = SegmentHelperMock.getSegmentHelperMockForTables(executor);
+        authHelper = AuthHelper.getDisabledAuthHelper();
+        storeHelper = new PravegaTablesStoreHelper(segmentHelper, authHelper, executor);
```
fpj: Does it need to be closed during tear down?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: The method (`addIndexAndSubmitTask`) returns the result of executing futureSupplier which is `t`. 
The original code was also ultimately returning `t` after writing the event. 
With this change, we are "handling" exceptions thrown from executing futureSupplier and as a safety we are writing event in case of StoreException or WriteConflictException thrown from futureSupplier. And in successful case we return its result (which is `t`) and in failure case we throw the exception.

controller/src/test/java/io/pravega/controller/store/stream/PravegaTablesStoreHelperTest.java
shiveshr: no, store helper is not closable. It has no resources that need to be closed.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
fpj: Could we refactor this part of the code and use say `exceptionallyComposeExpecting` to avoid having to pass an atomic reference for a throwable outside the completion stage flow?

shiveshr: We want to throw the exception even after performing a `thenCompose` on expected failures. This is to indicate to the caller that the requested operation "may" have failed since we received an exception from the store.

fpj: Can we use `thenCombine` to combine the result of `addRequestToIndex` and the retry loop to remove tasks from index?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -658,11 +658,36 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+        // will be discarded. We will throw the exception that we received from running futureSupplier or return the
+        // successful value
+
+        Supplier<CompletableFuture<Void>> postExecuteFutureSupplier = 
+                () -> RetryHelper.withIndefiniteRetriesAsync(() -> writeEvent(event),
+                e -> log.warn("writing event failed with {}", e.getMessage()), executor)
+                   .thenCompose(v -> streamMetadataStore.removeTaskFromIndex(context.getHostId(), id));
+
+        return streamMetadataStore.addRequestToIndex(context.getHostId(), id, event) 
+        .thenCompose(v -> Futures.handleCompose(futureSupplier.get(),
```
fpj: can you fix the indentation, please?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -658,11 +658,36 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
     @VisibleForTesting
     <T> CompletableFuture<T> addIndexAndSubmitTask(ControllerEvent event, Supplier<CompletableFuture<T>> futureSupplier) {
         String id = UUID.randomUUID().toString();
-        return streamMetadataStore.addRequestToIndex(context.getHostId(), id, event)
-                           .thenCompose(v -> futureSupplier.get())
-                           .thenCompose(t -> RetryHelper.withIndefiniteRetriesAsync(() -> writeEvent(event), e -> { }, executor)
-                                                        .thenCompose(v -> streamMetadataStore.removeTaskFromIndex(context.getHostId(), id))
-                                                        .thenApply(v -> t));
+        // We first add index and then call the metadata update.
```
fpj: Can you fix this comment as it is saying that we first add to the index, but then the next line of code is defining a supplier? Either we move the comment or rephrase to explain the flow better.

common/src/main/java/io/pravega/common/concurrent/Futures.java
```diff
@@ -259,7 +260,7 @@
             }
         });
     }
-
+
```
fpj: Please remove the whitespaces or tab in this line.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: yes, we still call the addIndex first.
The supplier is defined, which is used after addIndex (we call supplier.get and thats when the future is returned to us). 
I have now moved it inside the handleCompose where it is used so as to avoid this confusion..

PR #4047, opened 2019-07-23
Issue 4046: (SegmentStore) Segment Container Recovery bug with partial operations

### LINKED ISSUE #4046: Recovery fails with DataCorruptionException if an operation was only partially written to Tier 1
**Problem description**
Consider this scenario:
1. We need to write a large operation that does not fit into a single Tier 1 write.
2. The operation is split across two `DataFrame`s (or more)
3. The first `DataFrame` makes it to Tier 1, but not the second one (because Tier 1 crashed or the host process crashed or whatever other reason)
4. The Container recovers - all is OK; the above operation is ignored.
5. More operations are written to the log in this container.
6. The Container fails over again.
7. Subsequent recovery fails with `DataCorruptionException` on the partially written operation (which should have been ignored).
8. The Container enters 'Offline` mode (due to a perceived corruption during recovery) and becomes unusable - manual intervention is needed to bootstrap it.

**Problem location**
Somewhere in `DurableLog`, `RecoveryProcessor` or beneath.

Fix whatever recovery bug is there.

andreipaduroiu:
**Change log description**  
- Fixed a bug in `DataFrameInputStream` where it was possible to skip over (parts of) a valid operation if the previous operation was only partially committed to Tier 1.

**Purpose of the change**  
Fixes #4046.

**What the code does**  
- The `DataFrameInputStream` provides an InputStream-like interface on top of `DataFrame Entries`  that make up the Tier 1 `DurableDataLog`. Every call to `DataFrameInputStream.read` may either read from the currently loaded `DataFrameEntry` or fetch another one (via `fetchNextEntry`).
- However, in certain exceptional cases (such as when an Operation has been split but only one part was successfully written to Tier 1 (BK or Segment Store crashed), we may have had to request more DataFrameEntries in order to figure out the situation (which means we may have also read part of the next (valid) operation). When this happens, we notify the upstream code (via a `RecordResetException`) and set ourselves in a state where we can only proceed once that upstream code has recovered from this situation and is ready to begin reading the item (via  `beginRecord`).
 - The upstream code is `VersionedSerializer`, which wraps this in a `BoundedInputStream` (bounded for each `DataFrameRecord` (i.e., Operation)) with a bound set to its initial serialization length. It further wraps it in `RevisionDataInputStream` (one such per revision of the code).
    - In normal conditions, it is ok for the `RevisionDataInputStream` to request a `skip` of a number of bytes - this happens in case of forward compatibility (older code reads data serialized in newer format), so `skip` must be a valid operation.
    - However, in an exceptional case, such as when `RecordResetException` is thrown, the `BoundedInputStream` skips over to its "perceived" end of the record. However, for a partially commited operation, that "end of the record" is way past the last written byte of that operation (since we haven't committed the second part); it may be in the middle of the next operation or even beyond.
    - This would cause the `DataFrameInputStream` to erroneously jump to the middle of another record or even past it, which is a sure cause for a `DataCorruptionException`.
- As such, we need a way to "freeze" the `DataFrameInputStream` until the upstream code can tell us it is ready to proceed. Such a state is already codified in the `prefetchedEntry`, so all we need to do is block reads and skips (skips by default invoke `read`) if this flag is set to true.

**How to verify it**  
Unit tests updated to verify behavior.
New unit test for `DurableLog` that verifies the scenario used to reproduce the issue initially.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/DataFrameInputStream.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/DataFrameInputStream.java
@@ -42,6 +42,22 @@ public class DataFrameInputStream extends InputStream {
     @Getter
     private boolean closed;
     private boolean hasReadAnyData;
+    /**
+     * The {@link DataFrameInputStream} provides an {@link InputStream}-like interface on top of
+     * {@link DataFrame.DataFrameEntry} instances that make up the {@link DurableDataLog}. Every call to {@link #read} may
+     * either read from the currently loaded {@link DataFrame.DataFrameEntry} or fetch another one ({@link #fetchNextEntry()}.
+     *
+     * However, in certain exceptional cases (such as when an Operation has been split but only one part was successfully
+     * written to the {@link DurableDataLog}), we may have had to request more {@link DataFrame.DataFrameEntry} instances
+     * in order to figure our the situation (which means we may have also read part of the next (valid) operation). When
+     * this happens, we need to notify the upstream code (via a {@link RecordResetException}) and set ourselves in a state
+     * where we can only proceed once that upstream code has recovered from this situation and is ready to begin reading
+     * the next item ({@link #beginRecord}.
+     *
+     * Upstream code may wrap this instance in other types of {@link InputStream}s which may invoke {@link #skip} when
+     * closed (for exceptional cases). We need to make sure we don't skip over bytes that may actually be needed, while
+     * at the same time do support {@link #skip} for other, valid cases.
+     */
     private boolean prefetchedEntry;
 
     //endregion
@@ -77,6 +93,8 @@ public class DataFrameInputStream extends InputStream {
     @Override
     @SneakyThrows(DurableDataLogException.class)
     public int read() throws IOException {
+        Preconditions.checkState(!this.prefetchedEntry,
+                "Cannot read or skip from a prefetched entry without prior calling beginRecord().");
         while (!this.closed) {
             int r = this.currentEntry.getData().read();
             if (r >= 0) {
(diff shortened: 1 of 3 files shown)


segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/DurableLogTests.java
```diff
@@ -1033,6 +1034,68 @@ public void testRecoveryWithMetadataCleanup() throws Exception {
...
+        @Cleanup
+        ContainerSetup setup = new ContainerSetup(executorService());
+        @Cleanup
+        DurableLog dl1 = setup.createDurableLog();
+        dl1.startAsync().awaitRunning();
+        Assert.assertNotNull("Internal error: could not grab a pointer to the created TestDurableDataLog.", setup.dataLog.get());
+        val segmentId = createStreamSegmentsWithOperations(1, dl1).stream().findFirst().orElse(-1L);
+
+        // Part of this operation should fail.
+        ErrorInjector<Exception> aSyncErrorInjector = new ErrorInjector<>(
```
fpj: `aSyncErrorInjector` -> `asyncErrorInjector`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/DataFrameInputStream.java
```diff
@@ -77,6 +93,8 @@ public void close() {
     @Override
     @SneakyThrows(DurableDataLogException.class)
     public int read() throws IOException {
+        Preconditions.checkState(!this.prefetchedEntry,
+                "Cannot read or skip from a prefetched entry without prior calling beginRecord().");
```
fpj: Please replace with:

> Must call beginRecord() before reading or skipping from a prefetched entry.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/DataFrameInputStream.java
```diff
@@ -96,6 +114,8 @@ public int read() throws IOException {
     @Override
     @SneakyThrows(DurableDataLogException.class)
     public int read(byte[] buffer, int index, int length) throws IOException {
+        Preconditions.checkState(!this.prefetchedEntry,
+                "Cannot read or skip from a prefetched entry without prior calling beginRecord().");
```
fpj: Please replace with:

> Must call beginRecord() before reading or skipping from a prefetched entry.

PR #4063, opened 2019-07-27
Issue 4014: Reduce chances of throwing Lock Failed Exception for concurrent stream creation

### LINKED ISSUE #4014: StreamManager.Create throws Locking Resource Exception
**Problem description**

_StreamManager.create_ throws an `INTERNAL: Failed locking resource` exception if called on a Stream that is currently being used.  If the stream isn't currently being used then the method behaves differently and returns silently if the stream already exists.

The problem is that there is currently no way of knowing if the stream already exists since StreamManager.listStreams also does not work.

```
2019-06-27 13:03:05,799 WARN  io.pravega.client.stream.impl.ControllerImpl                  - gRPC call for createStream with trace id -1116438983471604940 failed with server error.

io.pravega.shaded.io.grpc.StatusRuntimeException: INTERNAL: Failed locking resource flink-longevity-1/longevity-test.
        at io.pravega.shaded.io.grpc.Status.asRuntimeException(Status.java:530)
        at io.pravega.shaded.io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:434)
        at io.pravega.shaded.io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
        at io.pravega.shaded.io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
        at io.pravega.shaded.io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
        at io.pravega.shaded.io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
        at io.pravega.shaded.io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
        at io.pravega.shaded.io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
        at io.pravega.shaded.io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:694)
```

**Problem location**
StreamManager.Create

StreamManager.create should behave in the same manor regardless of if the stream is currently in use or not.  Really it should always return silently if the requested stream already exists.

shiveshr:

**Change log description**  
LockFailedException is thrown upon multiple concurrent calls to createStream. This is retryable and should be retried on controller rather than throwing it back to the client. 

**Purpose of the change**  
Fixes #4014 

**What the code does**  
1. ControllerService.createStream should only attempt to acquire lock and create stream if the stream is not created yet. For already created streams immediately return stream exists.
2. StreamMetadataTasks.createStream has a new wrapper method called createStreamWithRetries which retries if the exception is of type LockFailedException

**How to verify it**  
Unit test added/updated


controller/src/test/java/io/pravega/controller/server/ControllerServiceWithStreamTest.java
```diff
@@ -65,6 +65,10 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.*;
```
shrids: Nit: we should remove * in import ref: https://github.com/pravega/pravega/blob/master/checkstyle/checkstyle.xml#L78

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -118,15 +120,24 @@ public ControllerService(StreamMetadataStore streamStore, StreamMetadataTasks st
...
-                                                createTimestamp)
-                  .thenApplyAsync(status -> {
-                       reportCreateStreamMetrics(scope, stream, streamConfig.getScalingPolicy().getMinNumSegments(), status,
-                                timer.getElapsed());
-                       return CreateStreamStatus.newBuilder().setStatus(status).build();
-                  }, executor);
+        return Futures.exceptionallyExpecting(streamStore.getState(scope, stream, true, null, executor), 
+            e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException, State.UNKNOWN) 
+        .thenCompose(state -> {
+            if (state.equals(State.UNKNOWN) || state.equals(State.CREATING)) {
```
fpj: I don't understand why we need to try to create again in the case that the stream is `State.CREATING`? Doesn't it mean that it is in the process of being created?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -143,6 +144,20 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+     *
+     * @param scope           scope.
+     * @param stream          stream name.
+     * @param config          stream configuration.
+     * @param createTimestamp creation timestamp.
+     * @return CompletableFuture which when completed will have creation status for the stream.
+     */
+    public CompletableFuture<CreateStreamStatus.Status> createStreamWithReries(String scope, String stream, StreamConfiguration config, long createTimestamp) {
+        return RetryHelper.withRetriesAsync(() ->  createStream(scope, stream, config, createTimestamp), 
+                e -> Exceptions.unwrap(e) instanceof LockFailedException, 10, executor);
```
fpj: In the case that the number of retries exhausts, what's the error that the application will see? Is it `RetriesExhaustedException`? I'd rather have the result of the last attempt, which should be more informative.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -118,15 +120,24 @@ public ControllerService(StreamMetadataStore streamStore, StreamMetadataTasks st
...
-                                                streamConfig,
-                                                createTimestamp)
-                  .thenApplyAsync(status -> {
-                       reportCreateStreamMetrics(scope, stream, streamConfig.getScalingPolicy().getMinNumSegments(), status,
-                                timer.getElapsed());
-                       return CreateStreamStatus.newBuilder().setStatus(status).build();
-                  }, executor);
+        return Futures.exceptionallyExpecting(streamStore.getState(scope, stream, true, null, executor), 
+            e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException, State.UNKNOWN) 
+        .thenCompose(state -> {
```
fpj: The indentation here seems wrong.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -143,6 +144,20 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+    /**
+     * Method to create stream with retries for lock failed exception.
+     *
+     * @param scope           scope.
+     * @param stream          stream name.
+     * @param config          stream configuration.
+     * @param createTimestamp creation timestamp.
+     * @return CompletableFuture which when completed will have creation status for the stream.
+     */
+    public CompletableFuture<CreateStreamStatus.Status> createStreamWithReries(String scope, String stream, StreamConfiguration config, long createTimestamp) {
```
fpj: I don't understand why we need a new call `createStreamWithReries`. Can't we make the `createStream` call retryable directly? Why do you want to have one retryable and one non-retryable?

controller/src/test/java/io/pravega/controller/server/v1/ControllerServiceImplTest.java
```diff
@@ -303,33 +302,7 @@ public void createStreamTests() {
         status = result6.get();
         assertEquals(status.getStatus(), CreateStreamStatus.Status.SUCCESS);
     }
-
-    @Test
-    public void testCreateStreamThrowsLockFailed() {
```
fpj: If you are removing this test case, then you should consider removing all other auxiliary code that came with the test case (follow `blockCriticalSection`).

controller/src/main/java/io/pravega/controller/server/ControllerService.java
shiveshr: A previous attempt to create a stream could have failed and left the stream in CREATING state.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: CreateStream has annotations which is used by task framework. LockException is thrown by task framework which relies on those annotations as the method with annotations is called. So the retry block has to be on the caller of createstream method. hence the wrapper method.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: Currently it would be RetriesExhausted that wraps LockFailedException. 
I will unwrap it.

controller/src/test/java/io/pravega/controller/server/v1/ControllerServiceImplTest.java
shiveshr: I have removed blockCriticalSection but retained the class TaskStoreFactoryForTests as that could be useful in authoring more tests in future.

PR #4064, opened 2019-07-29
Issue 4044: (Segment Store) Cache Improvements [WIP]

### LINKED ISSUE #4044: Direct Memory Cache
**Problem description**
Our current cache implementation is based on RocksDB and has several drawbacks:
- It's a Key-Value store. We do not need a KVS; it's sufficient to insert data and get an address back that can later be used to access that data.
- It has overhead due to being a Key-Value Store. This includes indexing and compaction work.
- It has inflexible APIs that aren't appropriate for Streaming. For example, the only way (using the JNI) to append to an existing entry is to read it, copy it to a new byte array, add the new data, then re-insert the updated array. For this reason we chose to have an entry for each append we receive in the Segment Store.
- It has disk-spilling abilities which aren't as useful as we anticipated. While it would be nice to be able to have a cache larger than available RAM, doing one without specialized hardware (SSD or NVM) will introduce a significant performance bottleneck once memory buffers are full.

**Problem location**
Segment Store.

**Requirements**
We need a cache storage that meets the following requirements:
- Does not store data in Java heap and hence is not subject to GC governance.
- Has performance on par or better than using a Java HashMap (excluding GC pauses).
- Does not suffer from memory fragmentation or frequent allocation/deallocation problems.
- Has minimal heap and non-heap overhead.
- Supports appending to existing entries.
- Minimizes the number of memory copies required when inserting into or reading from it.
- Has a clear maximum bound and optionally notifies (whomever is listening) when the capacity is reached and/or exceeded.

A solution that meets all the above requirements would implement a pooled block-based cache that works along the following lines:
- Splits the cache into large direct memory buffers of equal size (i.e., 2MB each)
- Each buffer is split into equally-sized blocks (i.e., 512 4KB blocks)
- Block 0 in each buffer is reserved for metadata, and the rest are used for storing data.
    - Each block's metadata indicates whether it's used, its successor or predecessor and its length. For efficiency, empty blocks may also be chained together to make writing faster.
- Inserting would involve finding the first available buffer, writing as much data to it as possible, then locating additional buffers and repeating until the entire data is stored. A pointer to the last block (codified as an Int32/64) will be returned and serves as the entry's address.
- Deleting would involve walking all the blocks reachable from an address and marking them as unused.
- Reading would involve walking all the blocks reachable from an address and combining them into a composite buffer that can be returned to the caller.
- Appending would append as much data as possible to the last block (if non-full) and optionally create new blocks.

Disadvantages:
- Rather complex compared to simply allocating direct memory via `ByteBuffer.allocateDirect`
    - Using `ByteBuffers` would require us to keep pointers to each one individually (extra heap overhead) and would be subject to memory fragmentation, so the complexity is warranted.
- Internal fragmentation. An entry will occupy a number of blocks, and the last block will be fully reserved, even if only 1 byte is required. This has the potential of allocating a lot of unused memory, however that's where `append` comes into play. We can append to an existing entry until we fill up its last block, and then we can insert a new entry. This plays well into the Streaming world, where we append by default so we are guaranteed to always have new data to add at the end.
- Reserved memory for metadata. There is one block reserved in each buffer for metadata, which could constitute memory that could otherwise be used for storage. However this overhead is minimal. For a 4KB block/2MB buffer, the reserved space is 0.2% of the total capacity. So a 10GB cache would only need 20MB of reserved space for metadata.

Benchmarking:
- See comment below.

andreipaduroiu:
**WIP: Please look away.**

**Change log description**  
(2-3 concise points about the changes in this PR. When committing this PR, the committer is expected to copy the content of this section to the merge description box)

**Purpose of the change**  
Fixes #4044.

**What the code does**  

**How to verify it**  


andreipaduroiu:
Will resubmit as a rebased branch.

PR #4065, opened 2019-07-30
Issue 3967 : Add test cases for compatibility job

anishakj:
**Change log description**  
As a prerequisite for this test,  pravega should be running in background. 
Jenkins  job will checkout the previous commit of pravega client and run these tests with master version. Currently 2 tests related to transaction is added.

**Purpose of the change**  
#3967

**What the code does**  
Created a new project `compatibility` in gradle. Tests are excluded from normal build of pravega. added a task `startCompatibilityTests` to start the tests. Test assumes pravega will be running in background.

**How to verify it**  
Verify compatibility tests are passing.


shrids:
The reason for the travis build failure is because the `src/main/java` is missing in the PR. This would cause the `java` gradle plugin to be not applied to the sub project `test:compatibility`.
https://github.com/pravega/pravega/blob/0a97e376a7e11ee9721d9321e74d46aa9e6a4897/build.gradle#L46-L47

build.gradle
```diff
@@ -536,6 +536,30 @@ project('test:integration') {
...
+    dependencies {
+        testCompile group: 'ch.qos.logback', name: 'logback-classic', version: qosLogbackVersion
+        testCompile group: 'org.apache.commons', name: 'commons-csv', version: apacheCommonsCsvVersion
+        testCompile project(path:':common', configuration:'testRuntime')
+        testCompile project(path:':shared:protocol', configuration:'testRuntime')
+        testCompile project(path:':segmentstore:server', configuration:'testRuntime')
+        testCompile project(path:':segmentstore:server:host', configuration:'testRuntime')
+        testCompile project(path:':shared:metrics', configuration:'testRuntime')
+   }
+    //disable the default test task.
```
fpj: I don't understand what is being disabled here.

test/compatibility/src/main/resources/logback-test.xml
```diff
@@ -0,0 +1,23 @@
...
+  Copyright (c) 2017 Dell Inc., or its subsidiaries.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+-->
+<configuration scan="true">
```
fpj: We have no code in `src/main` , so I'm wondering if we need this logback configuration file. Also, why is it calle `logback-test.xml`?

fpj:
Thinking some more about this PR, this addition is odd. If you are a user of Pravega and you see a folder under tests that says "compatibility/", then you'd expect the tests there to perform compatibility validation. But, this is not what's happening. This is code to enable an internal jenkins job to validate, the tests themselves don't do it. Consequently, these test cases need to be in a location that enables others to use it too and in a meaningful manner.

Please propose a different location for these tests.

fpj:
My understanding is that we no longer need this PR, so I'm closing it.

PR #4068, opened 2019-07-31
Issue 4067: Avoid loop for Pending bytes calculation

### LINKED ISSUE #4067: Avoid loop for pending bytes calculations
**Problem description**
Currently Append processor maintains the list of pending append for each of the writers.
upon every requests, pending bytes are calculated with scanning all the append requests for every writer to suspend/resume netty connection.
This calculation is is complexity O(N) for every new message with append per netty connection.  N is the number of pending appends 

**Problem location**
Append Processor / Segment Store

Calculation of the pending bytes per netty connection can be reduced to from O(N) to O(1).
This reduction will be useful in future to place the pending bytes as one pravega metric

kmgowda:
**Change log description**  
The loop for pending bytes calculation per netty connection is removed by maintaining the long variable per netty connection for pending bytes increase/decrease.

**Purpose of the change**  
Fixes #4067 

**What the code does**  
Currently the Append processor of segment store scans pending appends list for every writer to calculate the pending bytes to suspend or resume the netty connection.  This loop causes  complexity of O(N) for every new message on netty connection. This loop is removed now.
The Append Processor maintains the long variable pendingBytes per netty connection. This variable is incremented as the append request comes in and decremented as the segment store process the appends and educed the complexity of pending bytes calculation from O(N) to O(1). where N is total number of pending appends per writer id.

**How to verify it**  
1. All Unit tests and Integration tests are passing.
2. Pravega bench-marking is conducted with larger data set and duration on bare metal.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -301,6 +303,9 @@ private void handleAppendResult(final Append append, Throwable exception, Timer
                     latestEventNumbers.put(Pair.of(append.getSegment(), append.getWriterId()), append.getEventNumber());
                 } else {
                     if (!conditionalFailed) {
+                        List<Append> appends = waitingAppends.get(append.getWriterId());
+                        pendingBytes -= appends.stream().filter(a -> !a.isConditional())
+                                .mapToInt(a -> a.getData().readableBytes()).sum();
                         waitingAppends.removeAll(append.getWriterId());
```
andreipaduroiu: `.removeAll` will return all the items that are removed. You do not need to call `.get` and `.removeAll`; coalesce them into a single call.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -301,6 +303,9 @@ private void handleAppendResult(final Append append, Throwable exception, Timer
                     latestEventNumbers.put(Pair.of(append.getSegment(), append.getWriterId()), append.getEventNumber());
                 } else {
                     if (!conditionalFailed) {
+                        List<Append> appends = waitingAppends.get(append.getWriterId());
+                        pendingBytes -= appends.stream().filter(a -> !a.isConditional())
+                                .mapToInt(a -> a.getData().readableBytes()).sum();
```
andreipaduroiu: I am not sure if this will give you the results you want. If someone has read all or part of `a.getData()` then `.readableBytes()` will return less than what it was before, so your calculation will be inaccurate.

Do you have unit tests for this feature?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
kmgowda: I have removed this change; because, before calling this method, the getNextAppend is clearing all the data for the give writerid.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
kmgowda: I have removed this change.. because ,  this method is always called after getNextAppend .
The getNextAppend is clearing the pending bytes.
The existing unit tests covers this code diff.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -301,7 +304,8 @@ private void handleAppendResult(final Append append, Throwable exception, Timer
                     latestEventNumbers.put(Pair.of(append.getSegment(), append.getWriterId()), append.getEventNumber());
                 } else {
                     if (!conditionalFailed) {
-                        waitingAppends.removeAll(append.getWriterId());
+                        pendingBytes -= waitingAppends.removeAll(append.getWriterId())
```
andreipaduroiu: You still haven't answered my original question. What does `.readableBytes()` return there? The buffer will have already been read, so it's likely this is 0.

There is no unit test that verifies this value. You need to write one.

kmgowda: @andreipaduroiu 
yes, right now, it always returns 0 only; but , however, if the conditional append exists in between appends and the conditional failed, then it should be removed. hence code in 307 is required.
The existing tests in the AppendProcessorTest.java covering the 100% code l but I could not able to add any test case to induce the conditional append in between the data appends, because  Append processor append method invokes performNextWrite and performNextWrite invokes getNextAppend which reads the appends and all these calls are synchronous. 

one more thing about this PR is not just avoiding the loop; I want to this pending bytes to metrics in near future as a separate PR.

tkaitchuck: For the test you can add an additional append in the middle by creating two append processors that share the same store.

kmgowda: @tkaitchuck 
   I tried to do this; but , problem is this flow : performNextWrite -> getNextAppend  which always clears the pending bytes. please give me some guide lines here.
However , now, I have added additional code ensure that pendingBytes values never reaches negative value.

andreipaduroiu: Here's an idea:
- Make a protected method in `AppendDecoder` which sets the value of `pendingBytes`
- Subclass `AppendDecoder` and override that method
- In the subclass, record all values assigned to this field in an ordered list
- In your test, verify that the values recorded in this list are as you expected.

kmgowda: Hi @andreipaduroiu 
      AppendDecoder? but the 'pendingBytes" are local to "AppendProcessor".
I have added an private method getPendingBytes;  this can made as protected; however, this values is always turnns out 0 because Appendprocessor's Append method calls performNextWrite -> getNextAppend always clears this to 0.

andreipaduroiu: My bad. I was probably thinking of another PR when I wrote that comment. I meant `AppendProcessor`.

To answer your question, the way you would clear that value is by setting it to 0 (or subtracting until you reach 0). What I was suggesting was that, instead of directly assigning a new value to that field, we create a protected method that sets the field to whatever value we pass as an argument. We change all callsites to invoke this method instead of directly assigning it. Then in your unit test, you can subclass AppendProcessor and override the method so that you can record every single value it had.

kmgowda: Thanks a lot @andreipaduroiu 
yes,  I will post the unit test as per your above review comments.

kmgowda: Hi @andreipaduroiu 
    I have added the unit tests ; please let me know your further review comments.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -78,6 +79,8 @@
     private static final String EMPTY_STACK_TRACE = "";
     private final StreamSegmentStore store;
     private final ServerConnection connection;
+    private AtomicLong pendingBytes = new AtomicLong(0);
```
andreipaduroiu: Make this `final`

kmgowda: yes , its done now

kmgowda:
Thanks @andreipaduroiu
 
@tkaitchuck 
     please let me know your further review comments.

andreipaduroiu:
@kmgowda I'm waiting for the build to finish.

kmgowda:
@andreipaduroiu 
The build : https://travis-ci.org/pravega/pravega/builds/575177730?utm_source=github_status&utm_medium=notification
is successfully completed; but dont know why its not fetching the data.

kmgowda:
@andreipaduroiu 
     build and tests are completed successfully.

PR #4070, opened 2019-08-01
Issue 4066: List Stream should throw meaningful exception for non-existent scope

### LINKED ISSUE #4066: Pravega: List Streams for a Scope
**Problem description**
Scope not found exception should be thrown if scope is not present and the call should not be retriable. Also, there should be a way to know if there are streams in the scope. hasNext() does not work on the iterator received from the call.

**Problem location**
io.pravega.client.admin.StreamManager
Iterator<Stream> listStreams(String scopeName);

[exception.log](https://github.com/pravega/pravega/files/3449859/exception.log)

shiveshr:

**Change log description**  
Throw meaningful exception (ScopeNotFoundException) when getStreams is called for non-existent scope. 

**Purpose of the change**  
Fixes #4066

**What the code does**  
1. New exception type called `ScopeDoesNotExistExeption` added on client
2. List stream grpc response updated to include a status which can have stream not found status
3. List stream controller service grpc implementation to catch DataNotFoundException and set ScopeNotFound status in response.  
4. zk scope to throw DataNotFound if scope doesnt exist while doing getChildren
5. InMemoryscope to throw DataNotFound if scope does not exist and listStream is called. 
6. Standalone to use PravegaTables as the default store implementation. 

**How to verify it**  
unit test added


client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -243,9 +244,18 @@ public ControllerImpl(ManagedChannelBuilder<?> channelBuilder, final ControllerI
                                                                   .newBuilder().setScope(scopeInfo).setContinuationToken(token).build(), callback);
                         return callback.getFuture()
                                        .thenApply(x -> {
-                                           List<Stream> result = x.getStreamsList().stream()
-                                                                  .map(y -> new StreamImpl(y.getScope(), y.getStream())).collect(Collectors.toList());
-                                           return new AbstractMap.SimpleEntry<>(x.getContinuationToken(), result);
+                                           switch (x.getStatus()) {
+                                               case SCOPE_NOT_FOUND:
```
shrids: We can add a log similar to the other apis with the pattern `log.warn(requestId, "Scope {} does not exist", scopeName);`
and a INFO log with `requestId` when it succeeds.

shiveshr: the caller who made the exception will get the error and already logs all exceptions. Do we need the log here?

shrids: I see this pattern for other apis.
https://github.com/pravega/pravega/blob/b21decf445312ff3cc6d89dbf7563bf80019b3eb/client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java#L206-L217

client/src/main/java/io/pravega/client/stream/ScopeDoesNotExistException.java
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
```
fpj: Please remove year from copyright statement.

client/src/main/java/io/pravega/client/stream/ScopeDoesNotExistException.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.stream;
+
+public class ScopeDoesNotExistException extends RuntimeException {
```
fpj: Is `NoSuchScopeException` better?

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -232,20 +233,31 @@ public ControllerImpl(ManagedChannelBuilder<?> channelBuilder, final ControllerI
...
+                                                   log.warn(requestId, "Scope not found: {}", scopeName);
+                                                   throw new ScopeDoesNotExistException();
+                                               case FAILURE:
+                                                   log.warn(requestId, "Internal Server Error while trying to list streams in scope: {}", scopeName);
+                                                   throw new RuntimeException("Failure while trying to list streams");
```
fpj: Is there anything else we can get to indicate the cause of the failure?

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -232,20 +233,31 @@ public ControllerImpl(ManagedChannelBuilder<?> channelBuilder, final ControllerI
...
+                                               case SUCCESS: 
+                                               default: // we will treat all other case as success for backward 
+                                                   // compatibility reasons
```
fpj: Please put the comment block on top of `default:`:
```
// we will treat all other case as success for backward
// compatibility reasons
default:
```

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -232,20 +233,31 @@ public ControllerImpl(ManagedChannelBuilder<?> channelBuilder, final ControllerI
     public AsyncIterator<Stream> listStreams(String scopeName) {
         Exceptions.checkNotClosed(closed.get(), this);
         long traceId = LoggerHelpers.traceEnter(log, "listStreams", scopeName);
```
fpj: Do we still need this `traceId`?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -403,19 +405,26 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request, StreamO
...
-                                     .collect(Collectors.toList());
-                             return Controller.StreamsInScopeResponse.newBuilder().addAllStreams(streams)
-                                     .setContinuationToken(Controller.ContinuationToken.newBuilder().setToken(
-                                             response.getValue()).build()).build();
-                        }),
+                delegationToken -> Futures.exceptionallyExpecting(controllerService.listStreams(
+                        scopeName, request.getContinuationToken().getToken(), listStreamsInScopeLimit),
+                        e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException, null)
+                                          .thenApply(response -> {
+                                              if (response == null) {
```
fpj: Is it possible that `response` is `null` and not because of a data not found exception? What if there is no stream in the scope, what's the response?

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
```diff
@@ -298,13 +298,13 @@ protected void declareVersions() {
             private void read00(RevisionDataInput revisionDataInput, Token.TokenBuilder builder) throws IOException {
                 builder.msb(revisionDataInput.readCompactInt())
                        .middle(revisionDataInput.readCompactInt())
-                       .lsb(revisionDataInput.readCompactInt());
+                       .lsb(revisionDataInput.readInt());
```
fpj: I don't understand this change.

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
```diff
@@ -298,13 +298,13 @@ protected void declareVersions() {
...
             }
 
             private void write00(Token token, RevisionDataOutput revisionDataOutput) throws IOException {
                 revisionDataOutput.writeCompactInt(token.msb);
                 revisionDataOutput.writeCompactInt(token.middle);
-                revisionDataOutput.writeCompactInt(token.lsb);
+                revisionDataOutput.writeInt(token.lsb);
```
fpj: I don't understand this change either.

controller/src/test/java/io/pravega/controller/server/v1/PravegaTableControllerServiceImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.apache.curator.test.TestingServer;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * Zookeeper stream store configuration.
+ */
+public class PravegaTableControllerServiceImplTest extends ControllerServiceImplTest {
```
fpj: Why is this in this PR? It seems unrelated.

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
```diff
@@ -350,7 +350,7 @@ private ControllerServiceMain startLocalController(int controllerId) {
                 .trustStorePasswordPath(keyPasswordFile)
                 .build();
 
-        StoreClientConfig storeClientConfig = StoreClientConfigImpl.withZKClient(zkClientConfig);
+        StoreClientConfig storeClientConfig = StoreClientConfigImpl.withPravegaTablesClient(zkClientConfig);
```
fpj: This change is unrelated to this PR.

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
```diff
@@ -177,13 +177,17 @@ public ZKStoreHelper(final CuratorFramework cf, Executor executor) {
     }
 
     CompletableFuture<List<String>> getChildren(final String path) {
+        return getChildren(path, true);
+    }
+
+    CompletableFuture<List<String>> getChildren(final String path, boolean ignoreDataNotFound) {
```
fpj: I don't understand the purpose of this flag. Why would we want to ignore data not found sometimes?

shared/controller-api/src/main/proto/Controller.proto
```diff
@@ -154,7 +154,13 @@ message ContinuationToken {
  message StreamsInScopeResponse {
     repeated StreamInfo streams = 1;
     ContinuationToken continuationToken = 2;
-}
+    enum Status {
+        SUCCESS = 0;
+        FAILURE = 1;
+        SCOPE_NOT_FOUND = 2;
```
fpj: Instead of creating another return value, could we keep it at success/failure and include the failure message the reason? The advantage of throwing a explicit exception is that the application can react to it and create the scope. Is it what we are trying to achieve?

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
shiveshr: This is placeholder enumeration for generic exceptions where we will explicitly fail the request. 
Currently for this api, that response code is never actually sent from the server

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
shiveshr: That is tracing the method entry and will match will corresponding method exit

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
shiveshr: No, if there are no children we get empty set .

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
shiveshr: Compact int takes only non negative values. token.EMPTY has long.min value as lab.. which is cannot be serialised using compact int. Hence changed it to int

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
shiveshr: Same as before.. we write and read int instead of compact int

controller/src/test/java/io/pravega/controller/server/v1/PravegaTableControllerServiceImplTest.java
shiveshr: It’s not unrelated. ControllerserviceimplTest was not being tested with pravega tables. Only in memory and Zk.
So this effective adds all tests for APIs in controllerservice with Pravega tables too which includes the test for liststreams

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
shiveshr: Yes, I mentioned in the description that we are hanging the default store used in standalone to be Pravega tables. So all standalone tests exercise our default options

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
shiveshr: Earlier, for getChildren dataNotFound was translated into empty list by default. 
Everywhere else in zookeeperstreammetadatastore we did getChildren and expected empty list if there are no children or if the parent didn’t exist. However that was detrimental in determining if scope didn’t exist while doing list stream on scope for zookeeper based store. Hence I added a new api which did not translate the data not found to empty list

shared/controller-api/src/main/proto/Controller.proto
shiveshr: Since we are throwing explicit new exception so having a specific error code rather than a message is better

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
fpj: I understand that when listing streams you want the `StoreException.DataNotFoundException` to be thrown, but I don't understand why we need to allow two behaviors, that's confusing. What prevents us from always propagating `StoreException.DataNotFoundException` instead of having a dual behavior?

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
fpj: Can't we use a different encoding for empty rather than replacing compact int? Using compact encoding as much as we can is definitely preferable.

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
fpj: Piggybacking changes makes it difficult to track changes, would you mind changing this in a separate issue?

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
shiveshr: So getChildren is used in two places - get txns in epoch and get streams in scope. 
We had implemented the getChildren method to expect data not found and return empty list. 
We can have a single behaviour for getChildren as long as that is throwing data not found exception. 

And if we do that, we will have to handle that exception in transactionsInEpoch call to retain existing behaviour of that api. 
I thought instead of disturbing existing code, I would rather add a new getChildren api that threw the desired exception

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
shiveshr: I will handle it in a separate issue. I included it here because I was testing the standalone with liststreams and realised that we were using zookeeper based store as default in standalone

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
fpj: I suspect that the change in `transactionsInEpoch` to handle the exception should be minimal, but you tell me. If my suspicion is right, then I'd rather make that change and make the behavior consistent. Another option is to merge this one as is and create another issue to remove the flag and make the behavior consistent, but I'd want to do it right away to avoid the risk of releasing with this flag.

Given the size of this PR and the necessary changes to make it consistent, we might as well do it here.

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
fpj: Can you create the issue and post the number here, please?

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
fpj: I don't understand, you 👍 it, but I don't see a change.

controller/src/main/java/io/pravega/controller/store/stream/ZKStoreHelper.java
shiveshr: Actually i checked where all `getChildren` is called:
1. ZkOrderer : 4 different places
2. ZKScope: listStream
3. ZkStream: 5 different methods - all related to transactions
4. ZkStreamMetadataStore:  getCompletedTxn, listScopes
5. ZkBucketStore: getStreamsInBucket

So it will be a substantial change

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
shiveshr: https://github.com/pravega/pravega/issues/4092

controller/src/main/java/io/pravega/controller/store/stream/ZKScope.java
shiveshr: Sure. I can encode msb with a value greater than that number of top level znodes (which is 100). 
following is the tree structure:
100 -> 10k -> 10k
So encoding MSB with anything greater than 100 can be one "hacky" way. 
I didnt quite like that. 
These continuation tokens are only sent over the wire and 2 additional bytes for non-critical query api call (`listStreams`) does not seem too expensive.

client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
shiveshr: Actually we will keep adding specific error codes to the enum just like we have SCOPE_NOT_FOUND currently. 
So FAILURE is generic. 
And as i said earlier, it isnt sent from the server as of now. So its only a placeholder.

PR #4074, opened 2019-08-05
Issue 4044: (Segment Store) Direct Memory Cache

### LINKED ISSUE #4044: Direct Memory Cache

andreipaduroiu:
**Change log description**  
- Created a block-based, direct memory cache implementation and integrated into the Segment Store components.
- Removed old cache implementations (RocksDB and InMemoryCache).

**Purpose of the change**  
Fixes #4044.

**What the code does**  
- This is a rather big PR. However, most of the changes in this PR are due to the fact that we removed the old cache implementations and replaced with `CacheStorage`, which required cascading API changes. 
    - The most important changes in this PR are concentrated in `DirectMemoryCache`, `DirectMemoryBuffer`, `StreamSegmentReadIndex`, `SegmentKeyCache`, `SegmentAttributeBTreeIndex` and `CacheManager`.
- Direct Memory Cache implementation
    - Please refer to parent issue (#4044) for overview and explanation as to why this approach makes sense over others.
    - `CacheLayout` defines how the cache is structured. The following is based on the default cache layout implemented.
        - `DefaultCacheLayout` has the current implementation for this. Future versions of Pravega may choose to alter this setup, in which case they will need to implement a new layout.
        - Data is chained in a singly linked list (each block has a pointer the the previous block for the entry). The address returned to the caller points to the last block. This makes appends O(1) (no lookups) and makes inserts more efficient with no penalty on reads.
    - `DirectMemoryCache`
        - Splits the cache space into 2MB direct memory buffers (wrapped and managed by `DirectMemoryBuffer`). Its goal is to locate the appropriate buffers for all operations and to delegate requests to them.
        - It manages the set of unallocated buffers and available buffers with the goal of performing all operations in O(n), where `n` is the size of the data transferred. Buffer lookup is always O(1) so the overhead is minimal.
    - `DirectMemoryBuffer`
        - Wrapps a Netty `ByteBuf` which is only allocated on first access. This buffer is never deallocated (unless the owning `DirectMemoryCache` is closed). This is intentional and was done for performance reasons.
        - Splits the buffer into 4KB blocks. The first block is reserved for metadata, and the rest are data blocks. Each of the 511 data blocks has 8 bytes of metadata, containing information such as whether it is in use, its length, predecessor address and other info. `CacheLayout` is used to construct this metadata.
        - It keeps a linked list (in the metadata) of all free blocks. This makes inserts more efficient as there is no time spent on looking up where to insert. This linked list is updated upon every insert or delete.
- `CacheManager` owns the `DirectMemoryCache` instance and provides it to whomever needs it.
    - Its goal is to keep the `UsedBytes` under a certain threshold as indicated by its `CachePolicy`.
    - It provides a callback to the `DirectMemoryCache` which allows it to be notified if the cache has reached maximum capacity; if so it can force a cleanup.
- Integrated `DirectMemoryCache` into `StreamSegmentReadIndex`, `SegmentAttributeBTreeIndex` and `SegmentKeyCache`.
    - Since the new cache is not a KVS anymore, this allowed for the simplification of the cache access by simply storing the address of the cached entry.
    - `StreamSegmentReadIndex`
        - Updated `append()` to make use of `DirectMemoryCache.append()` if possible (so that it may fill up partial last blocks of existing entries, and then insert the rest.
- Deleted all references to old cache implementations and removed RocksDB maven dependency.

**How to verify it**  
All existing tests must pass.
New tests must pass.
System tests must pass (had 7/7 successful runs so far including different Tier 2 choices)


sachin-j-joshi:
Have we actually decided to remove old cache implementations (RocksDB and InMemoryCache)?

andreipaduroiu:
@sachin-j-joshi If we merge this PR in we will automatically retire the old caches (it is included in this PR). They are incompatible with the new cache storage and have vastly different (and more complex) APIs.

config/config.properties
```diff
@@ -102,13 +102,30 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
...
-# RocksDB) a larger value will be appropriate, as that cache implementation should be able to offload extra memory to disk
-# and fetch it back when needed.
-#pravegaservice.cacheMaxSize=17179869184
+# Recommended values: Multiples of 1GB, but less than XX:MaxDirectMemorySize. Choosing a lower size will conserve memory
+# at the expense of a lot of cache misses (resulting in Tier2 reads and possibly high latencies). Choosing a higher size
+# will keep data in the cache for longer, but at the expense of memory. The Segment Store uses a Direct Memory Cache backed
+# by direct ByteBuffers, so if XX:MaxDirectMemorySize is not set to a value higher than this one, the process will
+# eventually crash with an OutOfMemoryError.
+#pravegaservice.cacheMaxSize=4294967296
+
```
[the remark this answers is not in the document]
andreipaduroiu: Yes, it would be. However I would prefer to change that in a different PR since there's a number of other things that set this externally (such as the K8s operator).

andreipaduroiu:
> I'm worried about high levels of fragmentation that is bound to happen over the period of time. I might have missed it but we don't seem to concat consecutive blobs of free memory into bigger blobs.

That's the beauty of this approach. There is no memory fragmentation whatsoever. We allocate memory buffers once (in 2MB increments) and never deallocate them. Once allocated, that memory is fixed and since we never deallocate we will not be subject to that kind of fragmentation.

If you refer to fragmentation inside the cache, then yes, you are correct. We break each entry into 4KB blocks and try to write them in the same buffer, but it is possible that each such block is actually located in a different buffer, so in a different part of memory. Since we are not paging memory (yet), there is little benefit to trying to put things together vs just writing it wherever we can. I purposefully did not do any sort of defrag or consolidation as the effort (runtime) to do that would be disruptive to the whole cache performance.

> Over the time the we will have more smaller fragments and subsequently more metadata for each fragment.

The fragments cannot get smaller than 4KB and the metadata is constant. There is 8 bytes of metadata for each 4KB block (0.2% of the total cache capacity) which doesn't change if the block is not used, partially used or fully used (with or without adjacent neighbors).

> Why not just hold on to the buffers allocated by netty and use them in cache directly? Let the native memory manager do the hard work of alloc/dealloc.

We are holding on to those netty buffers. The only buffers that hold the memory are the 2MB ones in the cache itself. Any ByteBuf that is sliced from there is just a view into those buffers. 

If you are suggesting that we simply allocate and deallocate memory every time we insert/evict an entry to/from the cache, well that will incur memory allocation costs and will be subject to severe fragmentation over time. The sizes of the cache entries are highly unpredictable and vary greatly. Allocating and deallocating such random memory segments will cause us to use more memory than we want due to the inability to relocate existing memory pointers (this is native memory, not managed, so not subject to GC moving it). In addition, since we cannot do `append` we will be reverting back to the old ways of using the cache, where each event will have its own cache entry - leading to an explosion of cache entries, where our metadata just to keep track of them will be significant.

> Also can we store metadata in "normal" java data structures instead of storing it in buffers? Eg. if the metadata is of fixed size then allocate a separate array of fixed size

yes that is an option. However I preferred to keep this in the metadata itself so that 1) all memory required for it is bound by the Cache Max Size, 2) We do not incur memory overhead due to Java objects, 3) we do not bother the GC with all the arrays that we need to hold, and 4) it's faster to do bitwise manipulation than mutating Java objects.

RaulGracia:
@andreipaduroiu in this branch, I found the following problem when changing the value of `pravegaservice.cacheMaxSize` via the operator. Maybe is a general issue of the configuration, but I haven't see that before:

```
(CONFIGURATION)
...
      bookkeeper.bkAckQuorumSize: "3"
      # While these are low setting for RocksDB in terms of space/memory, we have verified that they make the Segment Store stable in the long run.
      pravegaservice.cacheMaxSize: "3221225472‬"
      pravegaservice.cacheMaxTimeSeconds: "600"
      rocksdb.writeBufferSizeMB: "64"
...
```

``` 
(LOGS IN SEGMENTSTORE)
2019-10-13 16:01:36,153 320  [main] ERROR i.p.s.server.host.ServiceStarter - Could not create a Service with default config, Aborting.
io.pravega.common.util.InvalidPropertyValueException: Value '3221225472‬' is invalid for property 'pravegaservice.cacheMaxSize'.
        at io.pravega.common.util.TypedProperties.tryGet(TypedProperties.java:144)
        at io.pravega.common.util.TypedProperties.getLong(TypedProperties.java:92)
        at io.pravega.segmentstore.server.store.ServiceConfig.<init>(ServiceConfig.java:320)
        at io.pravega.common.util.ConfigBuilder.build(ConfigBuilder.java:116)
        at io.pravega.segmentstore.server.store.ServiceBuilderConfig.getConfig(ServiceBuilderConfig.java:69)
        at io.pravega.segmentstore.server.host.ServiceStarter.<init>(ServiceStarter.java:59)
        at io.pravega.segmentstore.server.host.ServiceStarter.main(ServiceStarter.java:218)
Caused by: java.lang.NumberFormatException: For input string: "3221225472‬"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Long.parseLong(Long.java:589)
        at java.lang.Long.parseLong(Long.java:631)
        at io.pravega.common.util.TypedProperties.tryGet(TypedProperties.java:142)
        ... 6 common frames omitted
```

gradle.properties
```diff
@@ -49,8 +49,7 @@ nettyBoringSSLVersion=2.0.17.Final
 protobufGradlePlugin=0.8.9
 protobufProtocVersion=3.5.1
 qosLogbackVersion=1.2.3
-rocksdbjniVersion=5.8.6
-swaggerJersey2JaxrsVersion=1.5.22
+swaggerJersey2JaxrsVersion=1.5.16
```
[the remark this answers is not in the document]
andreipaduroiu: Automated merging ... I removed it.

common/src/main/java/io/pravega/common/util/ByteArraySegment.java
```diff
@@ -237,6 +247,7 @@ public OutputStream getWriter() {
      * @throws ArrayIndexOutOfBoundsException If offset or length are invalid.
      */
     public ByteArraySegment subSegment(int offset, int length) {
+        // TODO: drop this in favor of slice(). Also rename the one with `readOnly`
```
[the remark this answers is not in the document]
andreipaduroiu: I didn't want to do it as part of this PR since there's a good number of callsites, which would have made this PR unnecessarily bigger. #4318.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheLayout.java
```diff
@@ -0,0 +1,347 @@
...
+
+    //endregion
+
+    //region Constructor
+
+    /**
+     * Creates a new instance of the {@link CacheLayout} class and performs any necessary sanity checks.
+     */
+    CacheLayout() {
+        Preconditions.checkState(MAX_TOTAL_SIZE % bufferSize() == 0,
```
tkaitchuck: This could cause a problem. This is an abstract class but it is calling `bufferSize()` in its constructor which is an abstract method. Classes that extend this one are **Required** to call `super()` as the first line of their constructor, before any member variables are assigned. So if a subclass relies on setting bufferSize by a value passed in at construction time, it will not be set yet, so the value here could be -1.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheLayout.java
```diff
@@ -0,0 +1,347 @@
...
+
+    //region Default Layout
+
+    /**
+     * {@link CacheLayout} with the following characteristics:
+     * - 2MB Buffers
+     * - 4KB block size
+     * - 511 usable Buffer-Blocks per Buffer.
+     *
+     * Address layout (26 bits used out of 32)
```
tkaitchuck: It seems like somewhere the fact that this allows a max size of 256gb should be defined as a constant.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Performs an atomic compare-and-appends of data at the end of an entry. This operation will only take effect if
+     * the current length of the entry matches the argument `expectedLength` and both the entry's length and contents
+     * will be updated atomically.
+     *
+     * @param address        An integer representing the address to append to.
+     * @param expectedLength The currently known length of this entry.
+     * @param data           A {@link BufferView} representing the data to append.
+     * @return An integer representing the number of bytes to append. If 0, then either the address points to an invalid
```
tkaitchuck: This is phrased strangely.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,121 @@
...
+     * entry (does not exist) or the entry at this address may not be appended to (i.e., it may already be fully aligned
+     * to a block).
+     * @throws IncorrectCacheEntryLengthException If `expectedLength` does not match the current length of the entry.
```
tkaitchuck: I find it odd that the address not existing at all is not an exception but is represented in the return code, but the compare and set failing is an exception. It seems like this should be reversed.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,121 @@
...
+     * address.
+     */
+    BufferView get(int address);
+
+    /**
+     * Returns a {@link CacheSnapshot} representing the current state of the {@link CacheStorage}.
+     *
+     * @return A new {@link CacheSnapshot} instance.
+     */
+    CacheSnapshot getSnapshot();
```
tkaitchuck: Why is this needed?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
```diff
@@ -0,0 +1,439 @@
...
+import java.util.function.Supplier;
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.ThreadSafe;
+import lombok.NonNull;
+import lombok.val;
+
+/**
+ * Represents a block-based {@link CacheStorage} that uses direct memory buffers to store data.
+ *
+ * Data Organization (actual values based on {@link CacheLayout.DefaultLayout}).
```
tkaitchuck: A lot of this logic including breaking up memory into buffers and buffers into blocks of uniform size and managing them by reusing the same allocation is already implemented in Netty which this code is depending on anyway to allocate the actual buffers. `PooledByteBufAllocator` provides much of what this class implements, but has multiple pools for different size allocations, and has an optimized algorithm based on jemalloc to reclaim and reuse buffers. 

We should just use this class and pass in parameters to its constructor to get it to do what you want.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/cache/DirectMemoryCacheTests.java
```diff
@@ -0,0 +1,407 @@
...
+import lombok.val;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Unit tests for the {@link DirectMemoryCache} class.
+ */
+public class DirectMemoryCacheTests {
+    private static final CacheLayout LAYOUT = new CacheLayout.DefaultLayout();
+    private static final long REQUESTED_MAX_SIZE = 11 * 1024 * 1024; // 1MB
```
tkaitchuck: Comment says 1 value is 11

config/config.properties
```diff
@@ -117,13 +117,30 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
...
+# Percentage (of pravegaservice.cacheMaxSize) that defines the maximum usable size of the Local Shared Cache. The Segment
+# Store will apply full throttling if reaching or exceeding this value.
+# Valid values: 1 to 100 (inclusive).
+# Recommended values: Between pravegaservice.cacheTargetUtilizationPercent + 10 and 99. Setting a value too close to
+# pravegaservice.cacheTargetUtilizationPercent will cause the Segment Store to apply maximum throttling as soon as that
+# limit is reached, which may cause it to have sudden, multi-second stalls in the ingestion pipeline. It is not recommended
+# to set this value ever to 100, since that would mean the cache is already full when max throttling kicks in, and the
+# Segment Store will likely shut down Segment Containers in order to alleviate cache pressure.
+#pravegaservice.cacheMaxUtilizationPercent=98
```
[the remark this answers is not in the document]
andreipaduroiu: Good point. This was a carryover from the RocksDB setup when we could go over the limit. I was actually thinking of lowering this value a bit based on recent stress tests, but I will change them right now.

Setting target utilization to 75% (will try to keep cache at this level if possible), and full throttle at 90%. This should give enough buffer for sudden bursts or Tier 2 cache reads.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheLayout.java
andreipaduroiu: That is a valid concern and I agree with that. However I couldn't think of any other way to do this. I wanted `CacheLayout` to define the structure of any layout and leave it to a derived class to actually define proper values. However I do want some basic validation that the values the derived class is providing won't cause issues. 

While in general I want to stay away from such patterns, in this case I thin it's OK since derived classes will be close to this one and any issues will be quickly spotted by running a unit test trying to instantiate this.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheLayout.java
andreipaduroiu: This is: `CacheLayout.MAX_TOTAL_SIZE`.  I have just updated it to be 256GB (it had an smaller value, which was from a previous choice of layout).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: Rephrased.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: Good catch. I fixed this so it throws an `IllegalArgumentException` if the address is invalid.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: the `CacheManager` needs it so it can determine the Cache Utilization, so that it may make proper eviction and/or throttling adjustments.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
andreipaduroiu: I had a similar thought when I originally wrote this cache and I tried to use Netty's `PooledByteBufAllocator`. However I decided against it for the following reasons:
1. I would be tightly coupled with Netty. I don't want to depend on Netty and want to choose my own implementation at will. Currently I do use Netty's `ByteBuf` in the `DirectMemoryBuffer` class, but that could very easily be changed to Java's `ByteBuffer` (or whatever other implementation.
2. Netty has a bunch of logic to keep track of used and not-used blocks, but it does more than what I need. I wanted to keep the management of these buffers to a bare minimum to reduce the CPU needed to access them. In addition, I never deallocate the buffers and I want to implement my own policies as to how to place data (close to each other, etc).
3. I want to have the freedom to expand this into a disk-spillable cache. If I manage my own buffers it will make it easier for me to add such a functionality in the future. Tangling with Netty's allocator may actually make things harder.

Even without the three points above, the logic that deals with allocating the memory and managing it is so small that I did not find it worth it the effort to use Netty just for that. Most of the code in these 2 classes deal with organizing the data inside these buffers and providing access to them.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/cache/DirectMemoryCacheTests.java
andreipaduroiu: Typo. Fixed.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
tkaitchuck: I misunderstood how blocks worked. I didn't realize they were not shared across segments but were shared by appends within a segment.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryBuffer.java
```diff
@@ -0,0 +1,493 @@
...
+        blockLength += data.getLength();
+
+        // Update metadata.
+        blockMetadata = this.layout.setLength(blockMetadata, blockLength);
+        metadataBuf.setLong(bufIndex, blockMetadata);
+        return data.getLength();
+    }
+
+    /**
+     * Reads from the buffer starting at the given Buffer-Block Id and moving backwards (following
```
tkaitchuck: This is a very odd way to describe the API it is "moving backwars" but returns the items in reverse order. Would it be still accurate to say, it reads all of the data and returns it in order?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
```diff
@@ -0,0 +1,437 @@
...
+            throw ex;
+        }
+
+        CacheMetrics.insert(data.getLength());
+        return lastBlockAddress;
+    }
+
+    @Override
+    public int replace(int address, BufferView data) {
+        // The simplest and safest way to do a replace is to first insert the new data and then remove the old one. This
```
tkaitchuck: Given this implementation, is this API really providing value? Couldn't a caller just do that.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
```diff
@@ -0,0 +1,437 @@
...
+
+            // Keep track if this was full before we removed anything from it.
+            boolean wasFull = !b.hasCapacity();
+
+            // Delete whatever we can from it and remember the successor.
+            DirectMemoryBuffer.DeleteResult result = b.delete(blockId);
+            address = result.getPredecessorAddress();
+            deletedLength += result.getDeletedLength();
+            if (wasFull) {
+                synchronized (this.availableBufferIds) {
```
tkaitchuck: Shouldn't this be inside the if?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
```diff
@@ -368,13 +354,57 @@ void append(long offset, BufferView data) {
...
         long length = this.metadata.getLength();
-        long endOffset = offset + dataLength;
+        long endOffset = offset + data.getLength();
         Exceptions.checkArgument(endOffset <= length, "offset", "The given range of bytes (%d-%d) is beyond the StreamSegment Length (%d).", offset, endOffset, length);
 
         // Then append an entry for it in the ReadIndex. It's ok to insert into the cache outside of the lock here,
         // since there is no chance of competing with another write request for the same offset at the same time.
-        this.cache.insert(new CacheKey(this.metadata.getId(), offset), data);
-        appendEntry(new CacheIndexEntry(offset, dataLength));
+        log.debug("{}: Append (Offset = {}, Length = {}).", this.traceObjectId, offset, data.getLength());
```
tkaitchuck: This method is getting huge. Can parts be factored into private methods?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
```diff
@@ -492,61 +522,80 @@ void completeMerge(SegmentMetadata sourceMetadata) {
...
         ReadIndexEntry oldEntry;
-        synchronized (this.lock) {
-            this.cache.insert(getCacheKey(entry), data);
-            oldEntry = addToIndex(entry);
+        try {
+            newEntry = new CacheIndexEntry(offset, data.getLength(), dataAddress);
+            synchronized (this.lock) {
+                oldEntry = addToIndex(newEntry);
+            }
+        } catch (Throwable ex) {
```
tkaitchuck: Is this a possible/expected case?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheLayout.java
andreipaduroiu: Good idea. I made that change.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: Renamed to `CacheState`.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryBuffer.java
```diff
@@ -0,0 +1,493 @@
...
+        }
+
+        ByteBuf metadataBuf = getMetadataBlock();
+        long blockMetadata = metadataBuf.getLong(0);
+        int blockId = this.layout.getNextFreeBlockId(blockMetadata);
+        assert blockId != CacheLayout.NO_BLOCK_ID;
+        final int firstBlockId = blockId;
+        int dataOffset = 0;
+        try {
+            // Use a do-while loop since we want to properly handle the (valid) case when data.getLength() == 0.
```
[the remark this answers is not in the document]
andreipaduroiu: Well, we can very well insert a zero-length entry in the Cache Storage. That is a bad idea indeed, but the calling code may want to do this and the Cache Storage needs to support this case. The only way to make it work is to allocate a whole block for it so that we know it's registered. 

As of now there is no case in the Segment Store where it would insert a zero-length entry, but this is a possibility so we need to handle it.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryBuffer.java
```diff
@@ -0,0 +1,493 @@
...
+        metadataBuf.setLong(bufIndex, blockMetadata);
+        return data.getLength();
+    }
+
+    /**
+     * Reads from the buffer starting at the given Buffer-Block Id and moving backwards (following
+     * {@link CacheLayout#getPredecessorAddress} for each encountered block.
+     *
+     * @param blockId     The id of the Buffer-Block to begin reading from.
+     * @param readBuffers A list of {@link ByteBuf} to add read data to. The {@link ByteBuf}s will be added in reverse
```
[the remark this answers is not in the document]
andreipaduroiu: Rephrased.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryBuffer.java
andreipaduroiu: Rephrased.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
andreipaduroiu: Yes, given the **current** implementation, the caller can just do the same. However this API has a drawback in that it requires extra (temporary) cache storage space. Both the old and the new entry must be stored in the cache prior to deleting the old one.

In the future, we may want to optimize this to replace the entry block-by block, in which case it won't require these 2 steps. That may also provide a small perf gain.

I want to keep this so that we may implement any optimizations we want inside the cache and keep track of those callsites that use replacement.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
andreipaduroiu: This could go either way. 

The initial thinking here is that this needs to be synchronized with `getNextAvailableBuffer`. I prefer to have the buffer capacity check inside the synchronized block since otherwise it may get full again prior to adding it to the `availableBufferIds`.

However I now realize that I have code in `getNextAvailableBuffer` to handle full buffers and auto-remove them from there, so really this doesn't matter that much. I can make this change.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
andreipaduroiu: No, not expected, but there can always be a random `RuntimException` being thrown from one of the sub-methods called from there. In this case, I want to make sure that the data we have just inserted into the CacheStorage will be removed, otherwise we'll start leaking.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
andreipaduroiu: It will be hard. Each block within this method uses and modifies several variables, so it's not easy to extract them into meaningful methods.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Deletes the entry at the given address. If this address does not currently map to an entry, this operation will have
+     * no effect.
+     *
+     * @param address An integer representing the address to delete.
+     */
+    void delete(int address);
+
+    /**
+     * Retrieves the contents of an entry with the given address.
```
fpj: I'm not clear on the expected behavior for this call, and specifically, my question is the following. Say that the address passed corresponds to an old view of the entry, and it maps to a different position in the last block or even to a different block altogether. What is `get` expected to return? A stale view of the entry corresponding to the passed address? The current view of the entry? An error?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,122 @@
...
+     * @param address        An integer representing the address to append to.
+     * @param expectedLength The currently known length of this entry.
+     * @param data           A {@link BufferView} representing the data to append.
+     * @return An integer representing the number of bytes that were appended. If 0, then the entry at the given address
+     * is already at its max capacity (its length equals {@link #getMaxEntryLength()}.
+     * @throws IncorrectCacheEntryLengthException If `expectedLength` does not match the current length of the entry.
+     * @throws IllegalArgumentException If address does not point to a valid entry or the given {@link BufferView} is
+     * too long to fit into the given entry (use {@link #getAppendableLength} to determine maximum length).
+     */
+    int append(int address, int expectedLength, @NonNull BufferView data);
```
fpj: No unconditional append?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
```diff
@@ -0,0 +1,437 @@
...
+        int lastBlockLength = currentLength % this.layout.blockSize();
+        return currentLength == 0 ? this.layout.blockSize() : (lastBlockLength == 0 ? 0 : this.layout.blockSize() - lastBlockLength);
+    }
+
+    @Override
+    public int append(int address, int expectedLength, BufferView data) {
+        Exceptions.checkNotClosed(this.closed.get(), this);
+        Preconditions.checkArgument(address != CacheLayout.NO_ADDRESS, "Invalid address.");
+
+        // We can only append to fill the last block. For anything else a new write will be needed.
```
fpj: According to this comment and the code, if the append data overflows the last block, then a new write will be needed. I'm not sure what this implies, is it that the whole entry needs to be reinserted?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,122 @@
...
+     * @return The block alignment, in bytes.
+     */
+    int getBlockAlignment();
+
+    /**
+     * Gets a value representing the maximum size of an entry that can be stored in this {@link CacheStorage}.
+     *
+     * @return The maximum size, in bytes.
+     */
+    int getMaxEntryLength();
```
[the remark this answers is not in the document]
andreipaduroiu: Internally the metadata for each Cache Block is bounded (64 bits). This metadata needs to contain a number of different things, including how long is the entry beginning (or ending) at that block. Since I only have so many bits available,  I need to upper-bound this.

Additionally, inserting a very large entry into the cache is not every efficient. It will take longer to do and its chances of failing (due to the cache being full) are increased. Default is 256MB, which should be plenty for our needs.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,122 @@
...
+     * Performs an atomic compare-and-appends of data at the end of an entry. This operation will only take effect if
+     * the current length of the entry matches the argument `expectedLength` and both the entry's length and contents
+     * will be updated atomically.
+     *
+     * @param address        An integer representing the address to append to.
+     * @param expectedLength The currently known length of this entry.
+     * @param data           A {@link BufferView} representing the data to append.
+     * @return An integer representing the number of bytes that were appended. If 0, then the entry at the given address
+     * is already at its max capacity (its length equals {@link #getMaxEntryLength()}.
+     * @throws IncorrectCacheEntryLengthException If `expectedLength` does not match the current length of the entry.
```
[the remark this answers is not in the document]
andreipaduroiu: Yes. This was not added strictly because of concurrency, although this would handle it too. This is similar to an offset-conditional append, where we want to make sure the caller knows what they're doing. It's a way to double-check the integrity of the data, since `append` is the only way to modify what would otherwise be an immutable entry.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
```diff
@@ -0,0 +1,122 @@
...
+     */
+    CacheState getState();
+
+    /**
+     * Sets a callback that will be invoked during {@link #insert} if there is insufficient capacity to add more entries.
+     *
+     * @param cacheFullCallback The callback to invoke. This should return `true` if a cache cleanup was performed, and
+     *                          `false` otherwise.
+     */
+    void setCacheFullCallback(@NonNull Supplier<Boolean> cacheFullCallback);
```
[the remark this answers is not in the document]
andreipaduroiu: The eviction policy is implemented by `CacheManager` in conjunction with all the other information that the `ReadIndex` and other cache clients have. This is purely a Cache Storage: it stores and provides access to data, but does not bother with eviction as it has insufficient information to decide what can be evicted.

In other cases, we could write a small module that can manage this Cache Storage and perform eviction based on policies.

The point of this method is to notify the `CacheManager` that the cache is full and that it should try to execute an eviction if possible, otherwise it will reject new inserts with `CacheFullException`.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: The address always points to a block and never to an offset within a block. No two entries may share a block.

Once inserted, an entry will never change its address. `append` is the only thing modifying an entry, but its purpose is to fill up the last block, but not to add more blocks to the entry. So there is no such thing as an "old view of the entry".

If the entry has been deleted, the address will either point to a "free" block (which will result in an error), or to some other entry, in which case that entry's contents will be returned. You may say that this is incorrect, but this is a low-level storage (just like disk blocks), and it's up to the upstream code (`ReadIndex`) to manage the entries and make sure that when `delete` is invoked it forgets about that address.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: see https://github.com/pravega/pravega/pull/4074#discussion_r353981028

The point of `append` is to fill up the last entry in a block. If the upstream code wants more, then it can begin a new entry.

The reason I added this is because blocks are fixed size, so if an entry's length is not a multiple of that size (4KB), then we'll have extra unused memory. Append provides a very fast way of filling up that last block without the need to find new blocks or perform index changes. Appends are 5x faster than inserts as measured by the benchmark I added with this PR.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
andreipaduroiu: No. See https://github.com/pravega/pravega/pull/4074#discussion_r353985267.

Appends have been added especially to support streaming appends (1 byte, 10 bytes, 100 bytes, etc). They provide a very fast way of quickly adding that data to an existing entry without the need to allocate new blocks.

If we need to append more than 4KB (or whatever is available), the `ReadIndex` will fill up the last block and then create a new entry for the remaining data (for which it will also add a pointer in its index).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
fpj: My question is why you are adding only a conditional append operation to the API rather than adding both a conditional and an unconditional one.

Since we are on the topic of appends to fill up blocks, I'm wondering how effective that can be in our case. A block is fairly small (4KB), so I'm wondering how often we are expected to have small enough appends to be able to fill up a block.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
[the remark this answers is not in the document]
fpj: I think you are confirming that a policy is triggered by such a callback. When the cache is full, it needs to evict due to reaching its capacity, and some policy needs to make a call on what to evict. You are saying that the storage part  notifies via this callback and whatever policy is implemented, if any at all, reacts to such a notification. In the current implementation, this is the cache manager.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: There is no need for unconditional at this point. It can be extended.

The append situation is not something new. Even in our perf tests we discovered that, for many clients sending many small events (100 bytes), we have a hard(er) time putting all of that into the cache. The same is true with internal Table Segments (for metadata) which store about 100-500 bytes at a time). 

To solve that we either need to create a new entry or retrieve and reinsert an updated entry. Each of these is inefficient for their own reasons.

The other reason we need appends is exactly the 4KB block size. If we only use 200 bytes of it, the remaining 3896 bytes are wasted memory. If we create a new entry for each append, we'd be wasting a lot of it. So append was added also to help solve this (note that Tier 2 reads do not suffer from this problem as they are extended to be multiples of 4KB).

The final reason is that it's much faster to append than to insert. Insert requires locating new Cache Blocks, while Appends do no such thing - they just write to the last block which is conveniently pointed to by the entry's address.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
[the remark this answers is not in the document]
andreipaduroiu: Yes that is correct. The Cache Storage has insufficient information about what is stored and what should be evicted. The Cache Manager (external) has all the info it needs (what data is in Tier 2, for example) to make such decisions. The Cache Manager is also wired into the Throttler to slow ingestion down if it cannot evict fast enough.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
fpj: The client is expected to batch appends, in which case, even if the individual events are small, the append to the cache is likely to be larger than 4KB. But, the client might not be able to batch and send small append blocks. If that's correct, then the use of appends is optimizing for the latency of low throughout scenarios. For high-throughput scenarios, the appends blocks should be large enough to require new entries.

Is it a fair analysis?

andreipaduroiu: Yes, your understanding is correct.

But the memory optimization part plays a heavier weight. Not only does this align the cache entry sizes to the OS's page file size, but it also helps the `ReadIndex` by effectively putting a high bound on the total number of entries (while previously we could stash up to 4096 1 byte entries in 4KB (worst case append scenario), now all of that is represented by a single cache entry).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/CacheIndexEntry.java
```diff
@@ -10,34 +10,56 @@
...
+        this.length += delta;
+    }
+
     @Override
     boolean isDataEntry() {
         return true;
     }
+
+    @Override
+    public synchronized String toString() {
```
fpj: Why does this need to be synchronized? `cacheAddress` has no `@GuardedBy` annotation.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
```diff
@@ -42,20 +43,21 @@
...
- * a larger amount of data being cached (since it is backed by the process-wide {@link Cache}). The Tail Section, while
+ * a larger amount of data being cached (since it is backed by the process-wide {@link CacheStorage}). The Tail Section, while
  * dynamic, is not expected to grow too large due to the Table Segment being continuously indexed in the background, which
  * causes the Last Indexed Offset to be updated frequently.
  */
 @ThreadSafe
 @RequiredArgsConstructor
 class SegmentKeyCache {
     //region Members
+    private static final int HASH_GROUP_COUNT = 1024;
```
fpj: What's the real impact of making this constant larger or smaller?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
```diff
@@ -505,12 +449,50 @@ synchronized void update(UUID keyHash, long segmentOffset, int currentGeneration
...
-            SegmentKeyCache.this.cache.insert(key, entryData);
-            this.size = entryData.length;
+            storeInCache(new ByteArraySegment(entryData));
             this.generation = currentGeneration;
             this.highestOffset = Math.max(this.highestOffset, segmentOffset);
         }
 
+        /**
+         * Removes the contents of this entry from the cache, if anything was stored there in the first place. Invoking
+         * this will cause {@link #storeInCache} to throw an {@link IllegalStateException} going forward.
```
fpj: I don't understand what "this" is referring to in `Invoking this will cause`. What's "this"?

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/cache/BenchmarkTests.java
```diff
@@ -0,0 +1,223 @@
...
+ * of various changes in the implementation. Since the results may vary significantly based on the hardware used, outcomes
+ * are not comparable across environments. The way to use this is to establish a benchmark using the base code and then
+ * re-run it with the modifications already implemented and tested.
+ *
+ * The number of iterations {@link #ITERATION_COUNT} should always be greater than 1 since the first iteration will incur
+ * the costs of the initial memory allocation, while the remaining ones will benefit from the memory already being allocated.
+ *
+ * This is marked as @Ignore since these are not real unit tests (no correctness checking) and they take a long time to execute.
+ */
+@Ignore
```
fpj: Given that we are ignoring it when running junit, I'm wondering whether this is the right place for this benchmark. Is there are reason you want this file here? Could it be in `main` or even in another repository altogether like pravega-benchmark?

One concern I have with it being embedded deep down into the code repository is that only you will remember that this is here, while we might want to give some more visibility to it.

test/system/src/test/resources/pravega.properties
```diff
@@ -14,4 +14,8 @@ bookkeeper.bkAckQuorumSize=3
 controller.transaction.maxLeaseValue=120000
 controller.retention.frequencyMinutes=2
 log.level=DEBUG
-hdfs.replaceDataNodesOnFailure=false
\ No newline at end of file
+hdfs.replaceDataNodesOnFailure=false
+#4GB of cache.
+pravegaservice.cacheMaxSize=4294967296
+#4GB of cache + a buffer for whatever else Netty might need it for.
+io.netty.maxDirectMemory=5368709120
```
fpj: How do we estimate how much direct memory might need? How did you do it here?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheStorage.java
andreipaduroiu: Yes, if you still have an old address around, you can go read some garbage data from another entry (if it had been written again).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
```diff
@@ -0,0 +1,437 @@
...
+ * -- Are split into equal sized Blocks ({@link CacheLayout#blockSize()}.
+ * -- The Blocks form a singly-linked list with every block pointing to the previous block in the chain.
+ * -- The address returned by {@link #insert} or {@link #replace} points to the last block in this chain.
+ * -- No two Entries may share the same block. If the entry's size is not a multiple of {@link CacheLayout#blockSize()},
+ * then the last block will not be fully used; it will still be {@link CacheLayout#blockSize()} in length but no other
+ * entry may be added there.
+ * - The last block can be filled with {@link #append} which is optimized to add data to only the last block in an entry
+ * until it is full. It cannot be used to add new blocks to an entry (use {@link #getAppendableLength} to determine how
+ * much can be appended).
+ * - Fragmentation:
```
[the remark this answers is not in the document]
andreipaduroiu: We can create a new issue with optimizations to how entries are organized. We can choose an allocation algorithm that tries its best to keep entries in as few Cache Buffers as possible or some other method. At the moment, the benchmarks (included with this PR) show that there is little benefit from any quick " optimization" in that area (the cost of extra management outweighs any locality benefits).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheLayout.java
```diff
@@ -0,0 +1,351 @@
...
+/**
+ * Base class that defines the memory layout for a block-based {@link CacheStorage}.
+ */
+abstract class CacheLayout {
+    //region Members
+
+    /**
+     * The maximum number of bytes that can be stored in a {@link CacheStorage} using this type of layout.
+     */
+    static final long MAX_TOTAL_SIZE = 256 * 1024 * 1024 * 1024L;
```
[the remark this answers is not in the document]
andreipaduroiu: The reason I chose this "small" max value is because of the number of bits available in the `address` of each entry. There is room for more data in that 32-bit address, but at the moment I didn't have a need for all of them. For Pravega's current needs 256GB (per process) is good enough. If we run in a situation where we need more, we can always change this to support future needs. (there is no need to be too preemptive here as this is not persistent data so we need not worry about any versioning, etc.).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheMetrics.java
```diff
@@ -0,0 +1,37 @@
...
+package io.pravega.segmentstore.storage.cache;
+
+import io.pravega.shared.MetricsNames;
+import io.pravega.shared.metrics.DynamicLogger;
+import io.pravega.shared.metrics.MetricsProvider;
+
+/**
+ * Metrics for {@link CacheStorage}
+ */
+final class CacheMetrics {
```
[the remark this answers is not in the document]
andreipaduroiu: #4434

segmentstore/server/src/test/java/io/pravega/segmentstore/server/TestCacheManager.java
```diff
@@ -9,14 +9,15 @@
  */
 package io.pravega.segmentstore.server;
 
+import io.pravega.segmentstore.storage.cache.CacheStorage;
 import java.util.concurrent.ScheduledExecutorService;
 
 /**
  * Exposes the applyCachePolicy method in the CacheManager.
  */
 public class TestCacheManager extends CacheManager {
```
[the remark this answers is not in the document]
andreipaduroiu: `CacheManagerTests` is unit tests for `CacheManager`
`TestCacheManager` is an extension for `CacheManager` that exposes the `applyCachePolicy` method, so that other unit tests (for other components) that want to verify their cache eviction code can do so in a controlled manner (without having to rely on a timer to do so).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CachePolicy.java
```diff
@@ -22,11 +22,11 @@
     /**
      * Unless specified via the constructor, defines the default value for {@link #getMaxUtilization()}.
      */
-    public static final double DEFAULT_TARGET_UTILIZATION = 0.85;
+    public static final double DEFAULT_TARGET_UTILIZATION = 0.75;
```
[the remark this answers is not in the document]
andreipaduroiu: With the old implementation it really wasn't a big deal if we exceeded the maximum cache size. With this one we have a hard upper bound on how much data we can store in the cache - it was one of its design principles (to improve our stability).

After talking to Tom (in some of the comments on this PR) we decided it was safer to reduce these targets so that we can safely react to sudden burst in ingestion rates. For example, suppose we are sitting somewhere at the throttling limit, and we get a sudden 256MB influx of data. For a 3GB cache, 256MB is 8%, which takes us dangerously close to 100% if we kept the old values around (throttle begins at 90%, max at 98%). This would have been exacerbated if during that same time we had to perform a container recovery which would put even more pressure on the cache.

Lowering them reduces the likelihood that either one or a combination of 2 or more factors could lead to `CacheFullException`. We have tweaked this a bit in various perf tests that we've run recently and noticed a good improvement in the stability of the cluster by lowering them (stability: `CacheFullException` causes at least one Segment Container to shut down (gets evicted); the Controller will restart it in place, which may or may not fix the problem depending on how much Cache is available at that time).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
```diff
@@ -219,12 +227,60 @@ public void unregister(Client client) {
...
+     *
+     * Under normal operating conditions, the only method invoking this is {@link #runOneIteration()} which is guaranteed
+     * to execute only once at a time. Concurrent invocations come from {@link #cacheFullCallback()} which are triggered
+     * by the {@link CacheStorage} becoming full while inserting into it, which can come (at the same time) from different
+     * requesting threads.
+     *
+     * @return True if anything changed, false otherwise.
+     */
+    @VisibleForTesting
     protected boolean applyCachePolicy() {
```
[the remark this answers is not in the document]
andreipaduroiu: A few milliseconds most likely, from what I've seen in logs. It has to loop through all registered cache clients and tell them to evict oldest data (if possible). This is much faster now that we don't use RocksDB, but still we shouldn't expect it to be measured just in microseconds.

common/src/main/java/io/pravega/common/util/BufferView.java
```diff
@@ -49,13 +69,22 @@
     byte[] getCopy();
 
     /**
-     * Copies the contents of this {@link BufferView} to the given {@link OutputStream} using a 4KB copy buffer.
+     * Copies the contents of this {@link BufferView} to the given {@link OutputStream}.
      *
      * @param target The {@link OutputStream} to write to.
      * @throws IOException If an exception occurred.
```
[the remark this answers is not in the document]
andreipaduroiu: fixed

common/src/main/java/io/pravega/common/util/ByteArraySegment.java
```diff
@@ -140,6 +145,11 @@ public void copyTo(byte[] target, int targetOffset, int length) {
         System.arraycopy(this.array, this.startOffset, target, targetOffset, length);
     }
 
+    @Override
+    public void copyTo(ByteBuffer target) {
+        target.put(this.array, this.startOffset, Math.min(this.length, target.remaining()));
```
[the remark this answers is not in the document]
andreipaduroiu: That makes sense. I changed this to return an int.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
```diff
@@ -45,7 +45,7 @@
     public static final Property<DataLogType> DATALOG_IMPLEMENTATION = Property.named("dataLogImplementation", DataLogType.INMEMORY);
     public static final Property<StorageType> STORAGE_IMPLEMENTATION = Property.named("storageImplementation", StorageType.HDFS);
     public static final Property<Boolean> READONLY_SEGMENT_STORE = Property.named("readOnlySegmentStore", false);
-    public static final Property<Long> CACHE_POLICY_MAX_SIZE = Property.named("cacheMaxSize", 16L * 1024 * 1024 * 1024);
+    public static final Property<Long> CACHE_POLICY_MAX_SIZE = Property.named("cacheMaxSize", 4L * 1024 * 1024 * 1024);
```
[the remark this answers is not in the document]
andreipaduroiu: This is the config value passed via `config.properties`. We set it via the manifest and this decides the max cache size.

The one in `CacheLayout` is the maximum supported cache size for that particular layout. If somehow someone decides to exceed that value via the config, they will learn about it upon the system startup as it will throw an exception.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/CacheIndexEntry.java
andreipaduroiu: No it doesn't. I don't know how it ended up there.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
andreipaduroiu: This is the number of Cache Entries that the Segment Key Cache will use. Making it larger will increase the number of Cache Entries used, but they will be shorter on average, while making them smaller will decrease the number of cache entries used but storing more in each. I decided to decrease this because it is now much cheaper to read/write cache entries (even if we have more than one entry), but we'll make better use of the cache space with bigger entries due to the 4KB block size.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
andreipaduroiu: the "method"

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/cache/BenchmarkTests.java
andreipaduroiu: It will have even less visibility if it's in some obfuscated `main` class. I like it as unit test because it's easy to run it from the IDE and execute the IDE's code coverage on it. 

This is a pattern I've used before (for Segment Attributes and for the `AVLTreeIndex`).

test/system/src/test/resources/pravega.properties
andreipaduroiu: Cache Size + 1GB. 

Whatever is in here doesn't matter. All of these are overridable via the manifests we deploy with, which have much more tailored values for the HW config we're deploying to.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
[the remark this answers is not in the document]
fpj: Thanks for creating the issue.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
```diff
@@ -505,12 +449,50 @@ synchronized void update(UUID keyHash, long segmentOffset, int currentGeneration
...
+            if (address >= 0) {
+                SegmentKeyCache.this.cacheStorage.delete(address);
+                return true;
+            }
+
+            return false;
+        }
+
+        @GuardedBy("this")
+        private byte[] getFromCache() {
```
fpj: `getFromCache` in line 400 is not being invoked under a lock.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CachePolicy.java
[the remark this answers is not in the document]
fpj: One potential, non-trivial thing we could do to avoid such discussions around the target utilization of the cache is to treat it as a real cache with say an LRU policy (or whatever other policy, I'm not stuck on LRU) and if it fills up and we have to evict data that isn't in Tier 2 yet, then we get it from BK.

Such a change implies that we would read from BK in a case that isn't only recovery, and would indeed require adding such a code path. This path would in principle not be triggered very often if the cache is dimensioned properly,  and potentially leads to a more efficient utilization of the cache space.

If you think this is worth exploring, then let's capture it in an issue.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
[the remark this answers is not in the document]
fpj: Do I understand correctly that the lock guarding `applyCachePolicyInternal` only guards the state of the cache manager (e.g., set of clients, generations), but operations against cache storage can continue?

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/cache/BenchmarkTests.java
fpj: Location matters if you want others to use it too. To clarify, when I said `main`, I was not referring to a `main` class, but instead the `src/java/main` directory instead of `src/java/test` so that it is not confused with ignored test code.

Feel free to resolve this thread once you've read this, I only wanted to clarify the point.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
andreipaduroiu: Good catch. I changed this method to be `synchronized`.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/cache/BenchmarkTests.java
andreipaduroiu: This is good info. Thanks!

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CachePolicy.java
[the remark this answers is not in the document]
andreipaduroiu: #4438

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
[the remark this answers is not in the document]
andreipaduroiu: Yes. This is just to ensure that no two runs of the `CacheManager` will overlap. The `CacheStorage` itself is thread-safe and has much more fine-grained internal locks - it does not require external locking when operated on.

PR #4080, opened 2019-08-08
Issue 3684: A controller start/restart could stall in event processor start up if zk session expires

### LINKED ISSUE #3684: Zk session expiration during controller's start/restart can cause controller start to stall indefinitely
**Problem description**
In ControllerServiceMain, we start the controller starter followed by waiting on session expiration future. 
If the session expires before controller's modules are started, then there is one case where controller's start could stall indefinitely. 
This is because we have an infinite loop for retrying creation of internal streams. This loop retries unconditionally. 
A stream creation could fail because of zk session expiration, which can lead to a stale hostContainer mapping with the controller. 
The stream creation is retried indefinitely, whereby stalling the start of controller modules that depend on presence of internal streams. So controllerServiceStarter's startup will never complete in such case. And controller Service main will be stuck waiting on controllerServiceStarter to start successfully. 

**Problem location**
ControllerServiceMain and ControllerServiceStarter.

Stop controllerServiceStarter's start up attempt if zk session expires as startup is being attempted. await termination and then attempt internal restart loop in controllerServiceMain.

shiveshr:
**Change log description**  
Fixes an issue in ControllerServiceStarter where starter.start could stall indefinitely because it awaits start of eventProcessor. 

**Purpose of the change**  
Fixes #3684 

**What the code does**  
EventProcessors unconditionally and indefinitely retry CreateStream as part of its bootstrap and start up. If zk session expires during this time, the event processor startup is stalled whereby stalling ControllerServiceStarter and in turn ControllerServiceMain.

1. ControllerServiceMain already adds a listener on zk session expiration. if session expires as it is attempting to start the controllerServiceStarter, it notifies the starter about the same. 
ControllerServiceStarter maintains a future called `StoreClientFailureFuture` which is completed exceptionally with SessionExpirationException in the aforesaid notify method.

2. The controllerServiceStarter had two bugs: If zk session expires as create scope/stream is attempted indefinitely OR if eventProcessors.bootstrap threw exception, then controllerServiceStarter would get stuck waiting on `eventProcessor.awaitRunning`. 

As a fix we perform eventprocessor.awaitRunning by chaining it to eventProcessor.bootstrap future. 
And we wait on this future along with sessionExpirationFuture. If either of it completes (successfully or exceptionally) the future is completed. 
So if bootstrap succeeded, then startAsync and awaitRunning will be performed as expected. 
However, if bootstrap was stuck retrying indefinitely (and zk-session expired) or threw an exception in some step, then the composite future of `storeClientFailureFuture` and eventProcessor.bootstrap->startAsync->awaitRunning would have failed with one or the other. 

ControllerServiceStarter is also modified to make it more testable where we supply connectionFactory and streamMetadataStore in new constructors that are package visible for testing purposes. 

**How to verify it**  
Unit tests added


controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
```diff
@@ -286,14 +305,23 @@ protected void startUp() {
             // Wait for controller event processors to start.
             if (serviceConfig.getEventProcessorConfig().isPresent()) {
                 log.info("Awaiting start of controller event processors");
-                controllerEventProcessors.awaitRunning();
+                // if store client has failed because of session expiration, there are two possibilities where 
+                // controllerEventProcessors.awaitRunning may be stuck forever -
+                // 1. stream creation is retried indefinitely and cannot complete because of zk session expiration
+                // 2. event writer after stream creation throws exception. 
+                // In both of above cases controllerEventProcessors.startAsync may not get called. 
+                CompletableFuture.anyOf(storeClientFailureFuture,
```
shrids: We can have a scenario where `storeClientFailureFuture` completes (exceptionally) and `controllerEventProcessors.awaitRunning()` is invoked right?

shiveshr: yes, it is possible that we have the awaitRunning blocked while storeClientFailureFuture is failed. This will fail the ControllerServiceStarter's startup whereby throwing an exception. 
This will cause ControllerServiceMain to thorw the exception and exit the process.

controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
```diff
@@ -233,6 +251,7 @@ protected void startUp() {
             setController(new LocalController(controllerService, serviceConfig.getGRPCServerConfig().get().isAuthorizationEnabled(),
                     serviceConfig.getGRPCServerConfig().get().getTokenSigningKey()));
 
+            CompletableFuture<Void> eventProcessorFuture = CompletableFuture.completedFuture(null);
```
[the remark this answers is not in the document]
shiveshr: no, completed future is correct. If eventProcessor is not enabled then this is set. Otherwise eventProcessorFuture is set to future returned by `EventProcessors.bootStrap`

controller/src/main/java/io/pravega/controller/server/ControllerServiceMain.java
```diff
@@ -120,6 +120,9 @@ protected void run() throws Exception {
                 notifyServiceStateChange(ServiceState.STARTING);
                 starter.startAsync();
 
+                sessionExpiryFuture.whenComplete((r, e) -> {
+                     starter.notifySessionExpiration();
```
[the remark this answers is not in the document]
shiveshr: short answer, no we cannot move it to inside that `if block` because `ControllerServiceStarter` object is not present at that time. 

this is a callback registered on the session expiration future. and in the callback we send a notification to the starter object.

controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
fpj: If the zk session expires on startup, then `storeClientFailureFuture` completes exceptionally and unblocks this call. However, if case (2) happens (event writer after stream creation throws exception), then `storeClientFailureFuture` does not complete and from the comment `controllerEventProcessors.startAsync` may not get called. Are we stuck in this case?

controller/src/main/java/io/pravega/controller/server/ControllerServiceMain.java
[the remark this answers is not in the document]
shiveshr: I want to understand what the benefit of that suggestion would be? It would only be cosmetic. 

Following is what we do in current code:
1. we create curator client 
2. then we register for session expiration callback first thing before even attempting to perform any queries on the zookeeper session. In the callback we complete sessionExpirationFuture. 
3. we create the `starter` object using the storeClient which is a wrapper over the curator client. 
4. we register a callback on `sessionExpirationFuture` to call starter.notifySessionExpiration. 
5. we start the starter. 

we can do `3` before `2` and merge `4` into `2`. But what will we achieve with that change?

controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
shiveshr: No.
we are doing the following:
```
eventProcFuture = eventProcessors.bootStrap().thenAccept(v -> eventProcessors.startAsync());
...

CompletableFuture.anyOf(soreClientFailureFuture, eventProcFuture);
```

bootstrap:
1. IndefiniteRetry(createScope)
2. IndefiniteRetry(createStream)
3. CreateWriters

If `3` throws any exception then the future returned by bootStrap completes exceptionally. 
So `eventProcFuture` fails exceptionally with the same exception. 
And CompletableFuture.anyOf fails causing the startup to fail.

controller/src/main/java/io/pravega/controller/server/SessionExpirationException.java
```diff
@@ -0,0 +1,19 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.controller.server;
+
+/**
+ * Base class for exceptions thrown from controller server code base.
+ */
+public class SessionExpirationException extends Exception {
```
RaulGracia: If this related to Zookeeper sessions, maybe `ZKSessionExpirationException` could be a bit more specific.

controller/src/main/java/io/pravega/controller/server/SessionExpirationException.java
```diff
@@ -0,0 +1,19 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.controller.server;
+
+/**
+ * Base class for exceptions thrown from controller server code base.
```
RaulGracia: Is this comment correct? Seems a bit too broad given the exception at hand, which is specific to session expiration.

controller/src/main/java/io/pravega/controller/server/ControllerServiceMain.java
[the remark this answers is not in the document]
fpj: > I want to understand what the benefit of that suggestion would be?

It's code readability. We complete `sessionExpiryFuture` in line 112, it makes sense that we notify the starter right after completing that future. When I checked this code change the first time, I didn't immediately understand why we had a `sessionExpiryFuture.whenComplete` in the middle of the flow of that method.

PR #4081, opened 2019-08-08
Issue 3344: Reader side of watermarking implementation 

tkaitchuck:
**Change log description**  
This adds the changes to the reader side of the client per PDP-33. See PDP for details.

**Purpose of the change**  
Allow delivery of time information to reading applications

**What the code does**  
Reads data from the mark segments and uses this to compute a window based on the lastReadPosition of the reader group (which is now tracked via the reader group state) This tracks where the readers are not with respect to a checkpoint but with respect to the latest read (so it can go backwards in the event of a host failure). This is used to compute a cluster wide stream cut which is compared against the marks in the mark segment to determine what is and is not inflight. Finally this is used to obtain a time window which is made available to a client by calling the getTimeWindow() method on the reader.

The updates to this are piggybacked on an existing updates of the 'lag' which measures the distance to the tail of the stream for the current reader. (This is not now, nor was it previously accurately computed. Rather it is a very rough approximation used for balancing load between instances.)


client/src/main/java/io/pravega/client/stream/TimeWindow.java
```diff
@@ -14,18 +14,34 @@
...
  * 
- * The upper time bound is a timestamp which is greater than or equal to any that were provided by
- * any writer via the {@link EventStreamWriter#noteTime(long)} API prior to the current location in
- * the stream.
- * 
- * Similarly the lower time bound is a timestamp which is less than or equal to the most recent
+ * The lower time bound is a timestamp which is less than or equal to the most recent
  * value provided via the {@link EventStreamWriter#noteTime(long)} API for by any writer using that
- * API at the current location in the stream.
+ * API at the current location in the stream. If the reader is near the beginning of the truncation
```
shiveshr: Even if the stream is truncated, the timewindow with a position before truncation point can still be present in the watermark-segment. 
The computation of watermark is done after a writer submits a position and truncation can happen after it but the watermarks are all preserved in the watermark segment for each stream.
So even though timewindows can have a lower bound on time and position, but a reader can only really read from the truncation point. 

So i am not sure if the description here correctly reflects that.

client/src/main/java/io/pravega/client/state/SynchronizerConfig.java
```diff
@@ -21,10 +21,12 @@
 @Builder
 public class SynchronizerConfig implements Serializable {
```
shiveshr: should we move this to pravega's custom serialization? or is the synchronizerConfig class expected to be serialized and deserialized by user/application code too?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupState.java
```diff
@@ -191,6 +202,18 @@ public String getScopedStreamName() {
         return result;
     }
     
+    @Synchronized
+    Map<SegmentWithRange, Long> getLastReadpositions(Stream stream) {
```
shiveshr: nit: getLastRead`P`ositions

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
```diff
@@ -366,4 +368,9 @@ public static boolean isTableSegment(String qualifiedName) {
         return tokens[1].equals(TABLES);
     }
     // endregion
+    
+    // region watermark
+    public static String getMarkForStream(String stream) {
```
shiveshr: why do we need this additional wrapper method?

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
```diff
@@ -366,4 +368,9 @@ public static boolean isTableSegment(String qualifiedName) {
...
+        return NameUtils.getMarkStreamForStream(stream);
+    }
```
shiveshr: nit: region non closed

client/src/main/java/io/pravega/client/state/SynchronizerConfig.java
tkaitchuck: Yes: https://github.com/pravega/pravega/issues/4097

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
tkaitchuck: I don't think we do. What are your thoughts on unifying those two files? (NameUtils and StreamSegmentNameUtils)

shiveshr: We should definitely consolidate them. 
However, even if we do not unify them as part of this PR, StreamSegmentNameUtils is already visible publically. So we dont need to add this additional wrapper method here and directly access the method in StreamSegmentNameUtils.

tkaitchuck:
I updated the description. UpdateDistanceToTail and LagUpdateTimer aren't actually new concepts they are just being used to add the position to the reader group state. This makes sense as these are used to track reader's positions for the purposes of load balancing, so it makes sense for these to be in sync.

client/src/main/java/io/pravega/client/stream/TimeWindow.java
tkaitchuck: That is true. Though I'm not sure it is relevant to an end user. They can only read from the truncation point forward and the time will reflect their position in the stream, so the fact that marks exist below this may or may not allow for more accurate time information if they are at the truncation point, but as this cannot be guaranteed, it is being disclaimed here.

client/src/test/java/io/pravega/client/stream/impl/SerializationTest.java
```diff
@@ -155,7 +156,7 @@ public void testReaderGroupUpdates() throws Exception {
         verify(serializer, new RemoveReader(createString(), createSegmentToLongMap()));
         verify(serializer, new ReleaseSegment(createString(), createSegment(), r.nextLong()));
         verify(serializer, new AcquireSegment(createString(), createSegment()));
-        verify(serializer, new UpdateDistanceToTail(createString(), r.nextLong()));
+        verify(serializer, new UpdateDistanceToTail(createString(), r.nextLong(), createSegmentRangeMap()));
```
shrids: we should add tests for both the revisions right?

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -173,7 +175,13 @@ public ClientFactoryImpl(String scope, Controller controller, ConnectionFactory
                 synchronizerConfig);
         ReaderGroupStateManager stateManager = new ReaderGroupStateManager(readerId, sync, controller, nanoTime);
         stateManager.initializeReader(config.getInitialAllocationDelay());
-        return new EventStreamReaderImpl<T>(inFactory, metaFactory, s, stateManager, new Orderer(), milliTime, config);
+        Builder<Stream, WatermarkReaderImpl> watermarkReaders = ImmutableMap.builder();
+        if (!config.isDisableTimeWindows()) {
+            for (Stream stream : stateManager.getStreams()) {
+                watermarkReaders.put(stream, new WatermarkReaderImpl(stream, this, connectionFactory.getInternalExecutor()));
```
shrids: We can create `RevisionedStreamClient` here and pass it to the WatermarkReader here.

client/src/main/java/io/pravega/client/stream/impl/WatermarkReaderImpl.java
```diff
@@ -0,0 +1,209 @@
...
+        this.processor = new LatestItemSequentialProcessor<>(processFunction, Preconditions.checkNotNull(executor));     
+    }
+    
+    /**
+     * Advances the position in the watermark segment based on the provided position.
+     * 
+     * @param readerGroupPosition - The latest read positions of the open segments in the reader group.
+     */
+    public void advanceTo(Map<SegmentWithRange, Long> readerGroupPosition) {
+        processor.updateItem(readerGroupPosition);
```
shrids: If `WatermarkReaderImpl.advanceTo(...)` is invoked when  `processor.updateItem(..)` is being processed then it will be ignored. Could you explain the scenario when that would happen ?

common/src/test/java/io/pravega/common/concurrent/LatestItemSequentialProcessorTest.java
```diff
@@ -0,0 +1,81 @@
...
+        processor.updateItem("c");
+        startedLatch.await();
+        latch.release();
+        pool.shutdown();
+        pool.awaitTermination(5, TimeUnit.SECONDS);
+        assertEquals(ImmutableList.of("a", "c"), processed);
+    }
+    
+    @Test
+    public void testNotCalledInParrallel() throws InterruptedException {
```
shrids: nit: typo.

client/src/main/java/io/pravega/client/stream/impl/WatermarkReaderImpl.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private final Object lock = new Object();
+    @GuardedBy("lock")
+    private final ArrayDeque<Watermark> inflight = new ArrayDeque<>();
+    @GuardedBy("lock")
+    private Long passedTimestamp = null;
+    
+    @RequiredArgsConstructor
+    private static class WatermarkFetcher {
+        private final RevisionedStreamClient<Watermark> client;
```
shrids: the client needs to be closed to release the connection.

client/src/main/java/io/pravega/client/stream/impl/WatermarkReaderImpl.java
tkaitchuck: If multiple calls to advanceTo happen in rapid succession this logic ensures that the last one will get processed but not all of the intermediate updates will necessarily be processed depending on how long it takes to do the processing. This is OK because a subsequent update is going to be a strictly greater position as the readers only move forward. It's a good idea because it ensures that the logic to track the marks can't fall behind, and doesn't end up wasting cycles trying to 'catch up' when it can just skip ahead.

client/src/main/java/io/pravega/client/stream/impl/WatermarkReaderImpl.java
tkaitchuck: Done

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
tkaitchuck: Done

client/src/main/java/io/pravega/client/segment/impl/SegmentInputStreamFactoryImpl.java
```diff
@@ -44,6 +45,7 @@ private EventSegmentReader getEventSegmentReader(Segment segment, long endOffset
                                                                 RuntimeException::new);
         AsyncSegmentInputStreamImpl async = new AsyncSegmentInputStreamImpl(controller, cf, segment, delegationToken);
         async.getConnection();
+        bufferSize = MathHelpers.minMax(bufferSize, 1024, 10 * 1024 * 1024); //Sanity enforcement
```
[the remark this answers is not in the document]
tkaitchuck: done.

client/src/test/java/io/pravega/client/stream/impl/SerializationTest.java
tkaitchuck: Done

tkaitchuck:
Missing code coverage is coming from three places:
https://codecov.io/gh/pravega/pravega/compare/a579a07e286c0f54ca2f25c487e58210f0f40f02...82014ef9a32a55c4f76fcdebc266d67943da3208/diff#D6-594  -- Which is lombok code.
https://codecov.io/gh/pravega/pravega/compare/a579a07e286c0f54ca2f25c487e58210f0f40f02...82014ef9a32a55c4f76fcdebc266d67943da3208/src/client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java#L206  -- Which is just adding more info to controller error messages.
and 
https://codecov.io/gh/pravega/pravega/compare/a579a07e286c0f54ca2f25c487e58210f0f40f02...82014ef9a32a55c4f76fcdebc266d67943da3208/src/client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java#L181 -- Which is part of the public interface and is covered by integration tests not unit tests in the same package.

PR #4085, opened 2019-08-09
Issue 4069: (AppendDecoder) Trimming excess allocated memory in decoded appends.

### LINKED ISSUE #4069: Avoid retaining padding in ram on server.
**Problem description**
Following #4021 and connection pooling the server is forced to hold onto padding that is sent over the wire. This could be eliminated if in AppendDecoder the block was copied into the Append object rather than simply being wrapped.

**Problem location**
AppendDecoder

Invoke `capacity()` on the byte buff before returning it to shrink it to size if needed.

andreipaduroiu:
**Change log description**  
- Invoking `ByteBuf.copy()` so that we eliminate excess padding around appends.

**Purpose of the change**  
Fixes #4069 

**What the code does**  
- An append on the Segment Store side may be received from multiple append blocks. With connection pooling, the actual usable data in these blocks may be significantly less than the allocated size of the buffers (with the excess being padding).
- What we pass to the downstream Segment Store is a ByteBuf that is either sliced off a single Append Block or a composite of (slices) of multiple Append Block. As such, we can end up with an Append ByteBuf of length N which actually holds pointers to ByteBufs allocating significantly more memory than N. 
- This can be a problem if the Append ByteBuf is required for a significant amount of time and we have incoming append pressure. For example, if writing to Tier 1 or the Cache slows down for whatever reason, we could accumulate a large number of such buffers in memory that are pending to be written or committed; these buffers' lifetimes could span seconds or dozens of seconds, which would be enough to cause the process to run out of heap memory prematurely.
- A solution to this involves invoking `ByteBuf.copy()` on the bytebuf prior to sending it to the downstream Segment Store. This will create a new HeapByteBuf which uses only the memory it needs. This way we can release the memory from the original Append Blocks and reduce the memory pressure in case of Tier 1/Cache slowdowns.

**How to verify it**  
All tests must pass.

### DIFF
--- a/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
+++ b/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
@@ -147,13 +147,19 @@ public class AppendDecoder extends MessageToMessageDecoder<WireCommand> {
                 // See https://github.com/netty/netty/issues/5597
                 if (appendDataBuf.readableBytes() == 0) {
                     appendDataBuf.release();
-                    return wrappedBuffer(((WireCommands.PartialEvent) cmd).getData(), blockEnd.getData());
+                    appendDataBuf = wrappedBuffer(((WireCommands.PartialEvent) cmd).getData(), blockEnd.getData());
                 } else {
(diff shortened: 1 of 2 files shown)


andreipaduroiu:
I closed this and reopened as #4086, however the github build issue remains.

PR #4096, opened 2019-08-12
Issue 4095: Integration and System tests for watermarking 

### LINKED ISSUE #4095: Add integration and system test for Watermarking
**Problem description**
Add end to end integration and system tests to verify that watermarking works as expected. 

**Problem location**
Tests

Add tests

shiveshr:
**Change log description**  
System Test and Integration Test added for watermarking. 

**Purpose of the change**  
Fixes #4095 

**What the code does**  
Integration test that creates 2 writers and scales the stream multiple times as writers are writing. 
Writers write time into the events. 
Writers also periodically report time of the last event (every 10 seconds) by calling `noteTimestamp` method. 
Waits until one watermark is emitted and uses that as the lower bound for a new readergroup that it instantiates. 
It validates that all events read have times ahead of `watermark.lowerTimeBound`

SystemTest does exactly all of the above in a real deployment with multiple controllers. 
It waits for multiple watermarks to be emitted and uses these as both lower and upper bounds for readergroup. 
We also reduce controller instance and verify that watermarking continues to function correctly. 

**How to verify it**  
Tests only


test/integration/src/test/java/io/pravega/test/integration/WatermarkingTest.java
```diff
@@ -0,0 +1,245 @@
...
+            event = reader.readNextEvent(10000L);
+        }
+    }
+
+    private void scale(Controller controller, Stream streamObj, StreamConfiguration configuration) throws InterruptedException {
+        // perform several scales
+        int numOfSegments = configuration.getScalingPolicy().getMinNumSegments();
+        double delta = 1.0 / numOfSegments;
+        for (long segmentNumber = 0; segmentNumber < numOfSegments - 1; segmentNumber++) {
+            Thread.sleep(5000L);
```
tkaitchuck: Theses sorts of long delays are appropriate for a System test but not an integration test. This is going to run on every commit and the whole of the `gradle check` must run in under 1 hour on travis. So we want to keep them relatively quick.

test/integration/src/test/java/io/pravega/test/integration/WatermarkingTest.java
```diff
@@ -0,0 +1,245 @@
...
+        StreamConfiguration markConfig = StreamConfiguration.builder().scalingPolicy(ScalingPolicy.fixed(1)).build();
+        controller.createStream(scope, NameUtils.getMarkStreamForStream(stream), markConfig).join();
+
+        Stream streamObj = Stream.of(scope, stream);
+
+        // create 2 writers
+        ClientConfig clientConfig = ClientConfig.builder().controllerURI(controllerUri).build();
+        @Cleanup
+        EventStreamClientFactory clientFactory = EventStreamClientFactory.withScope(scope, clientConfig);
+        JavaSerializer<String> javaSerializer = new JavaSerializer<>();
```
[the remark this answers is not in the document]
shiveshr: changed to JavaSerializer<Long> and reading and writing long. No need to convert it to String as i was doing earlier.

test/integration/src/test/java/io/pravega/test/integration/WatermarkingTest.java
```diff
@@ -0,0 +1,244 @@
...
+                                                                           .startingStreamCuts(start).build());
+
+        // create reader on the stream
+        @Cleanup
+        final EventStreamReader<Long> reader = clientFactory.createReader("myreader",
+                readerGroup,
+                javaSerializer,
+                ReaderConfig.builder().build());
+        
+        // read events from the reader.
```
tkaitchuck: It would be good to assert that the reader receives a timeWindow from the `getTimeWindow` that has and that after all the events are read the low watermark is the largest of the low watermarks in the mark segment.

tkaitchuck: It once it is passed the noted time it should eventually return a watermark passed it. This can be asserted using assertEventuallyEquals with a readNextEvent() call in there to give it a chance to update the group state.

test/system/src/test/java/io/pravega/test/system/WatermarkingTest.java
```diff
@@ -0,0 +1,287 @@
...
+        Watermark watermark1 = watermarks.take();
+        Watermark watermark2 = watermarks.take();
+        Watermark watermark3 = watermarks.take();
+        
+        assertTrue(watermark0.getLowerTimeBound() < watermark0.getUpperTimeBound());
+        assertTrue(watermark1.getLowerTimeBound() < watermark1.getUpperTimeBound());
+        assertTrue(watermark2.getLowerTimeBound() < watermark2.getUpperTimeBound());
+        assertTrue(watermark3.getLowerTimeBound() < watermark3.getUpperTimeBound());
+
+        // verify that watermarks are increasing in time.
```
tkaitchuck: The upper time bounds should also go up.

test/integration/src/test/java/io/pravega/test/integration/WatermarkingTest.java
```diff
@@ -0,0 +1,244 @@
...
+        
+        stopFlag.set(true);
+
+        // read events from the stream
+        @Cleanup
+        ReaderGroupManager readerGroupManager = new ReaderGroupManagerImpl(scope, controller, syncClientFactory, connectionFactory);
+        String readerGroup = "rg";
+
+        long timeLow = watermark.getLowerTimeBound();
+        Map<Segment, Long> positionMap0 = watermark.getStreamCut()
```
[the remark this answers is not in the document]
shiveshr: with a scaling stream, what we can verify is that entire range is covered.

test/system/src/test/java/io/pravega/test/system/WatermarkingTest.java
shiveshr: that may not be true necessarily, at least in the current implementation. 
if a writer had reported a higher time in previous watermark computation and is no longer participating in the subsequent watermarking computation then the upper bound (which is upper bound on all reported times) could be lower than a previously reported high time. 

We could set it to max(previous.upperbound, currentUpperBound) which will guarantee that its always increasing.

tkaitchuck: ok

PR #4102, opened 2019-08-13
Issue 4101: Prevent users from canceling checkpoint completion

### LINKED ISSUE #2258: Cancellation support for `Futures::loop`
**Problem description**
The future that is returned by `Futures::loop` should support cancellation by terminating the loop.   For example, the Flink connector cancels the future produced by `ReaderGroup::initiateCheckpoint`  after some timeout.

**Problem location**
`io.pravega.common.concurrent.Futures`

### LINKED ISSUE #4101: Prevent callers from canceling futures returned from initiate checkpoint.
**Problem description**
Applications calling initiateCheckpoint on the reader group and then erroneously calling cancel on the future under the belief that it will cancel the computation (see #2258) will result in the the number of outstanding checkpoints not decreasing. If this is done repeatedly the application will be unable to create any more checkpoints once it hits the limit and be stuck in that state permanently.

**Problem location**
ReaderGroupImpl

Wrap the returned future so that cancel won't affect the removal of the completed future from the readerGroupState.

tkaitchuck:

**Change log description**  
Prevent canceling of future returned from initiateCheckpoint() from impacting the checkpoint.

**Purpose of the change**  
Fixes #4101, Closes #2258

**What the code does**  
Adds an additional stage to the returned future to prevent the call to cancel from influencing the computation.

**How to verify it**  
Added a test to verify the effectiveness and also fixed a test code bug introduced in #3844 which prevented verification.

### DIFF
--- a/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
+++ b/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
@@ -140,7 +140,8 @@ public class ReaderGroupImpl implements ReaderGroup, ReaderGroupMetrics {
         }
 
         return waitForCheckpointComplete(checkpointName, backgroundExecutor)
-                .thenApply(v -> completeCheckpoint(checkpointName));
+                .thenApply(v -> completeCheckpoint(checkpointName))
+                .thenApply(checkpoint -> checkpoint); //Added to prevent users from canceling completeCheckpoint
     }
 
     /**
--- a/client/src/test/java/io/pravega/client/stream/impl/ReaderGroupImplTest.java
+++ b/client/src/test/java/io/pravega/client/stream/impl/ReaderGroupImplTest.java
@@ -25,6 +25,8 @@ import io.pravega.client.stream.ReaderGroupConfig;
 import io.pravega.client.stream.Serializer;
 import io.pravega.client.stream.Stream;
 import io.pravega.client.stream.StreamCut;
+import io.pravega.client.stream.impl.ReaderGroupState.ClearCheckpointsBefore;
+import io.pravega.test.common.AssertExtensions;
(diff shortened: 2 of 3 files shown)


client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -140,7 +140,8 @@ public void readerOffline(String readerId, Position lastPosition) {
         }
 
         return waitForCheckpointComplete(checkpointName, backgroundExecutor)
-                .thenApply(v -> completeCheckpoint(checkpointName));
+                .thenApply(v -> completeCheckpoint(checkpointName))
+                .thenApply(checkpoint -> checkpoint); //Added to prevent users from canceling completeCheckpoint
```
maddisondavid: Just to confirm that it will still be safe to call `cancel` on the `CompletableFuture` (as the connector does)

fpj: This change prevents a cancel of the future to skip the `completeCheckpoint`computation, which is the logic that causes the number of outstanding checkpoints to drop. Long story short, yes, this makes it safe to cancel the future and the additional stage will only the `checkpoint` object from being returned. See the test case below for a validation of the expected behavior.

EronWright: May I suggest that a helper method be used to encapsulate this technique, .e.g `Futures.cancellable(...)`.   Would make it easier to identify the locations where cancellation is explicitly supported, and provide overrides in support of taking an action upon cancellation (such as cancelling an upstream future).

Regarding this specific instance, I wonder if the counter should be decremented in exceptional cases too.   Imagine that the caller were to close the `backgroundExecutor`, it would again defeat `completeCheckpoint`.

Issue #3923, opened 2019-06-18
JVM and controller.host  metrics have colliding `host` tag


jkhalack:
**Problem description**
Both JVM metrics and [Controller Host metrics](https://github.com/pravega/pravega/blob/master/shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java#L182-L184) have a `host` tag.
The JVM metrics report host name as a `host` tag (something like `pravega-segmentstore-0.pravega-segmentstore.pravega.svc.cluster.local` for a Kubernetes deployment), while controller host metrics report host IP (something like `1.2.3.4:12345:`).
This create a confusion when creating automated queries for JVM metrics (half availabe  `host` tag values do not have any JVM metrics for them).

**Problem location**
HostContainerMetrics.java ?

**Suggestions for an improvement**
It would be great if the Controller host metrics reported a `host` tag that is a kubernetes hostname instead of IP.

kevinhan88:
This issue is also closed by https://github.com/pravega/pravega/pull/4103

PR #4103, opened 2019-08-13
Issue 4054: Make common host tag consistent and shorter across controller and segmentstore

### LINKED ISSUE #4054: Inconsistent `host` values for controller and segmentstore metrics
**Problem description**

Controller and segmentstore Micrometer metrics have their `host` tags set up in a different way for the Kubernetes deployment.

When performing InfluxDB query on the metrics reported by Pravega,
`SHOW TAG VALUES FROM "jvm_memory_max" WITH KEY="host"` results in controller replicas represented by pid name, while segmentstore replicas being reported by their service names:
```
pravega-controller-5dccfbb7f7-g4xq9
pravega-segmentstore-0.pravega-segmentstore.default.svc.cluster.local
pravega-segmentstore-1.pravega-segmentstore.default.svc.cluster.local
pravega-segmentstore-2.pravega-segmentstore.default.svc.cluster.local
```

**Problem location**
Pravega docker scripts?
Segmentstore metrics hostTag() function implementation?

It would be much nicer if segmentstore host tags were reported with shorter strings, as pod names for example (like `pravega-segmentstore-0`).

kevinhan88:

**Change log description**  
This change checks if "HOSTNAME" is defined as system property (in the case of docker deployment it is) and use the defined hostname for the common host tag, in order to:
- have consistent host tag across controllers and segmentstores, and across pravega metrics and system metrics. This will make the metrics dashboard queries consistent and simpler.
- be flexible for the set of hostname. In this particular case, shorter hostname without the repetitive tailing segments.

**Purpose of the change**  
Fixes #4054, #3923 

**What the code does**  
When metrics is being initialized, instead of go straight to call InetAddress.getLocalHost().getHostName(), the code checks if "HOSTNAME" is defined and use it if so.

**How to verify it**  
Additional unit test added.
On metrics dashboard, consistent (and shorter) hostname should show up as host tag.
If we want to change the hostname for some reason, we can do so by changing system var, and then the metrics should reflect the change.


shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
```diff
@@ -139,6 +142,13 @@ private static String getSegmentBaseName(String segmentQualifiedName) {
      */
     public static String[] createHostTag() {
         String[] hostTag = {MetricsTags.TAG_HOST, null};
+
+        //Always take property "HOSTNAME" if it's defined.
+        hostTag[1] = System.getProperty(HOSTNAME_PROPERTY_NAME);
+        if (StringUtils.isNotBlank(hostTag[1])) {
```
andreipaduroiu: Can we use `Strings.isEmpty` here? I prefer not to add a dependency from this class to micrometer just for this method.

kevinhan88: Yes, changed to Strings.

PR #4105, opened 2019-08-14
Issue 4106: Enable checkstyle rule InvalidJavadocPosition

TisonKun:
**Change log description**  

Enable checkstyle rule `InvalidJavadocPosition` and fix conflicts.

**Purpose of the change**  

Mainly get rid of dangling Javadoc comments.

**What the code does**  

Enable checkstyle rule `InvalidJavadocPosition` and fix conflicts.

**How to verify it**  

`./gradlew checkstyleMain checkstyleTest` until no errors report.


TisonKun:
@fpj Thanks for your review and suggestion. I will open an issue on this thread.

After investigation I can find a rule[1] of checkstyle plugin can figure out all these instances. However, it requires upgrade checkstyle plugin to 8.23. A compatibility issue is raised as companion which I think I can handle it.

[1] https://checkstyle.sourceforge.io/config_javadoc.html#InvalidJavadocPosition

PR #2417, opened 2018-03-30
Issue 2390: Evicting Metrics cache based on time

Prabha-Veerubhotla:
Signed-off-by: PrabhaVeerubhotla <vvlprabha@gmail.com>

**Change log description**
The metrics `segment_read_bytes` are reported very frequently( for every 10 s) for non-active segments, due to which we are missing metrics from active segments.
We do not need to report metrics which are not accessed for a long period of time.
Cache eviction policy needs to be modified.

**Purpose of the change**
Fixes #2390 

**What the code does**
* Adds a new property `DYNAMIC_CACHE_EVICTION_DURATION` to Metrics Config with a default value of `30*60s`.
* Enables Metrics Cache eviction to be done based on time also

**How to verify it**
Metrics that are not being accessed for a period of time ( cache eviction duration) will be evicted and no longer reported.

Issue #886, opened 2017-03-29
Read Bytes Counter is going down


glounder:
**Problem description**
We are seeing the total read bytes for our stream going up and down over time.  The expectation is that this value always goes up over time.

**Problem location**
Metrics

**Suggestions for an improvement**
Ensure that the read/write bytes always go up over time.

![screen shot 2017-03-29 at 9 40 31 am](https://cloud.githubusercontent.com/assets/9335655/24455103/c7fea7d0-1464-11e7-94d3-02ba7e9ba051.png)

tkaitchuck:
From the discussion, it sounds like the best thing to do would be to simply replace this metric with a rate. As an absolute count is both hard to deal with on the receive side, and makes eviction hard on the server side.

PR #4117, opened 2019-08-19
Issue 4149: Reduce metrics cache eviction to 3 minutes

### LINKED ISSUE #4149: Reduce Metrics Cache Eviction Timeout
**Problem description**
Before the micrometer migration, there was a problem for the traditional counter, that if the counter object was evicted from the cache, then the counter was "reset", causing the "going up and down" effect on the metrics dashboard.
To fix the problem, we had to extend the cache eviction timeout to very long, by default 30 minutes. A lot of memory is occupied by the inactive metrics objects.

Refer https://github.com/pravega/pravega/issues/886 for the discussion.

With the micrometer migration, we started using StepRegistry, and in particular, StepCounter, which nailed down the above problem completely - without changing the Counter interface to application, the StepCounter implementation is actually a Rate.
Actually @tkaitchuck already pointed it out in https://github.com/pravega/pravega/issues/886 that we should change the counter to rate. Micrometer also states there shouldn't be counter at all and they changed the underlying implementation of counter to rate, solved the problem silently and elegantly.

How exactly the problem got solved? With StepCounter, each time the counter just report the delta, without concerning the previous value. Whether the previous counter object is there or not, is no longer important.
Now It is up to the back-end to determine how to aggregate the value for the counter:  to get the traditional view of counter, just aggregate for the period "infinite past to current".
To be honest, this absolute counter value is often useless, and not practical for a streaming system (imaging the counter of total read bytes for 2 years); for streaming we usually care a particular time window or period, which the rate is designed for!

With push mode, by default we export metrics every 10 seconds. The 30 mins cache eviction is now too long and unnecessary.

**Problem location**
MetricsConfig.java - metrics.dynamicCacheEvictionDurationMinutes

Reduce the default metrics.dynamicCacheEvictionDurationMinutes down to 3 minutes from 30 minutes.

kevinhan88:

**Change log description**  
Reduced the metrics cache eviction timeout to 3 minutes (from 30 minutes)

**Purpose of the change**  
Fixes #4149 

**What the code does**  
Changed the default cache eviction time from 30 minutes to 3 min inside MetricsConfig.java

**How to verify it**  
No functional changes.
The memory usage should be lower than before (with sufficient load) because the metrics cache now should hold less objects.


fpj:
See this issue for context on why we have initially removed the TTL altogether:

     https://github.com/pravega/pravega/issues/886

and later brought it back with a much longer TTL:

    https://github.com/pravega/pravega/pull/2417

kevinhan88:
To make things clear and straightforward, I created https://github.com/pravega/pravega/issues/4149 and assigned this pull request to that issue.

RaulGracia:
@fpj do you mind to have another look to this PR?

andreipaduroiu:
@kevinhan88 @fpj What's the status with this PR? Are we still pursuing it?

kevinhan88:
@andreipaduroiu I'll have a demo tomorrow to show the difference of two approaches.

PR #4119, opened 2019-08-20
Issue 4062: Fix ReaderGroup recreation failures.

### LINKED ISSUE #4062: Reader Group Recreation failures
**Problem description**
here is  the flow :
createReaderGroup -> deleteReaderGroup -> createReaderGroup  [Error] 

The 2nd invocation of createReaderGroup  fails with error

io.pravega.client.stream.InvalidStreamException: Segment does not exist: Scope/_RGtopic-t1-104RdGrp/0.#epoch.0
        at io.pravega.client.stream.impl.ClientFactoryImpl.createStateSynchronizer(ClientFactoryImpl.java:200)
        at io.pravega.client.admin.impl.ReaderGroupManagerImpl.createReaderGroup(ReaderGroupManagerImpl.java:95)
        at io.pravega.perf.PravegaStreamHandler.createReaderGroup(PravegaStreamHandler.java:153)
        at io.pravega.perf.PravegaPerfTest$PravegaTest.<init>(PravegaPerfTest.java:415)
        at io.pravega.perf.PravegaPerfTest.createTest(PravegaPerfTest.java:155)
        at io.pravega.perf.PravegaPerfTest.main(PravegaPerfTest.java:100)

**Problem location**
Controller/Client

I could see that "createStateSynchronizer"  always uses the segment ID  set to 0.
If the readergroup is deleted and recreated , the segment id would differ with standard segment id 0. hence this bug.
Needs further investigation.

shrids:
**Change log description**  
- Remove hardcoding of segment 0 for `StateSynchronizer` and `RevisionedStreamClient`. 
- Fix incorrect SealedSegment handling for `RevisionedStreamClient`.

**Purpose of the change**  
Fixes #4062

**What the code does**  
At present the client assumes segment 0 to be present for methods `io.pravega.client.stream.impl.ClientFactoryImpl#createStateSynchronizer` and `io.pravega.client.stream.impl.ClientFactoryImpl#createRevisionedStreamClient`. This is not valid for  scenarios where a ReaderGroup has been deleted and recreated. This PR ensures we query the controller to fetch the current segment.

This PR also ensure RevisionedStreamClient correctly handles a SealedSegment response from the segment store.

**How to verify it**  


client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -422,6 +423,24 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
+    public SegmentOutputStreamImpl(String segmentName, boolean useConnectionPooling, Controller controller,
+                            ConnectionFactory connectionFactory, UUID writerId, RetryWithBackoff retrySchedule,
+                            String delegationToken) {
+        this.segmentName = segmentName;
+        this.useConnectionPooling = useConnectionPooling;
+        this.controller = controller;
+        this.connectionFactory = connectionFactory;
+        this.writerId = writerId;
+        this.resendToSuccessorsCallback = segment -> {
+            List<PendingEvent> r = this.getUnackedEventsOnSeal();
```
tkaitchuck: This can be done externally to the class. So I think it would be better to put this in the revision stream client, than have two constructors that end up creating classes that behave differently. If it needs to be here for some reason, then it should at the very least should be spelled out in the Javadoc.

shrids: This change ensures that the default behavior of `SegmentOutputStreamImpl` is to change the state of writer to `Sealed` and ensure all the write futures are completed exceptionally. This change ensures all the users of ` SegmentOutputStream createOutputStreamForSegment(Segment segment, EventWriterConfig config, String delegationToken) ` have the same behavior.   (this is used by RevisionedStreamClient and ByteStreamClient).  Let me know your thoughts on this.

I have updated the java doc for the constructor documenting the behaviour of the default callback.

shrids: Thanks for the feedback, I have refactored the code and removed the extra constructor.

shrids:
System test run with this change-set was successful. ref: `test-pravega-continuous/2249/`

PR #4121, opened 2019-08-20
Issue 4120 : Close Append Implementation

### LINKED ISSUE #4120: Implement CLOSE_APPEND wire command
**Problem description**
Currently the pravega segment writers issues the SETUP_APPEND before doing any writes; 
Due to connection pooling, there will be many writers sharing same connection ; The command Encoder and Append Decoder maintains the hash maps for the writer ids;  The entries are remaining permanently even after the writer close is invoked. Due to this, the hash tables in command encoders and Append Decoders always grows and they never shrinks.

**Problem location**
Pravega Client / Segment output stream/ Command Encoder
Segment Store / Append Decoder.

Similar to SETUP_APPEND,  while closing the writer , its should cleanup the all meta data related to writer id and segment from the client and it should issue CLOSE_APPEND wire command so that append decoder cleans up any Hash table entries related to given writer id.

kmgowda:
**Change log description**  
A new Wire Command "CLOSE_APPEND" is implemented. During writer close, the pravega client issue this command to segment store. 

**Purpose of the change**  
Fixes #4120 

**What the code does**  
whenever the Pravega writer id invokes close;  The Segment output stream sends the new Wire command "CLOSE_APPEND"; The Command Encoder flushes all pending data of this writer and issues this command to segment store and cleanups the data of the writer id from the segments hash map of the Command Encoder. The Append Decoder of the segment store receives this command and clears the writer id and associated data.

**How to verify it**  
All Unit tests are passing
Testing is done by using pravega benchmark tool also.


kmgowda:
This PR is re-based  on top of PR: https://github.com/pravega/pravega/pull/3953

kmgowda:
@fpj 
1- What happens if the client does not close gracefully (e.g., it crashes)? I suppose we still have the same issue?
if the client crashes , then then netty connection closes at segment store and complete segment/writer id table of append decoder gets erased/removed.

2- Is this addition backwards compatible?
No, this is not backward compatible.

andreipaduroiu:
Just out of curiosity, why isn't this backwards compatible? An older client can talk to a newer server. I guess a newer client can't talk to an older server; can we fix this by bumping wire protocol versions?

kmgowda:
Thanks @andreipaduroiu 
Yes you are right.

@fpj 
Here is compatibility matrix
Pravega Server	Pravega Client	Compatibilty
New	old	Yes
New	New	Yes
old	new	No

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -480,8 +481,8 @@ public void close() throws SegmentSealedException {
             return;
         }
         log.debug("Closing writer: {}", writerId);
-        // Wait until all the inflight events are written
-        flush();
+        // Wait until all the in-flight events are written
+        clearInFlight(new CloseAppend(writerId, segmentName));
```
tkaitchuck: The name 'clear' in this context is misleading.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -179,14 +181,15 @@ private void flushAll(ByteBuf out) {
             ArrayList<Session> sessions = new ArrayList<>(pendingWrites.values());
             sessions.forEach(session -> session.flush(out));
         }
+        tokenCounter.incrementAndGet();
```
tkaitchuck: I don't understand why this token counter logic is being changed.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -230,12 +233,27 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
             SetupAppend setup = (SetupAppend) msg;
             setupSegments.put(new SimpleImmutableEntry<>(setup.getSegment(), setup.getWriterId()),
                               new Session(setup.getWriterId(), setup.getRequestId()));
+        } else if (msg instanceof CloseAppend) {
```
tkaitchuck: There are some commonalities between CloseAppend, BlockTimeout, and Flush. In all cases all appends for a particular writerId. Let's make a private method for that.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -519,6 +519,27 @@ public static WireCommand readFrom(DataInput in, int length) throws IOException
         }
     }
 
+    @Data
+    public static final class CloseAppend implements WireCommand {
+        final WireCommandType type = WireCommandType.CLOSE_APPEND;
+        final UUID writerId;
+        final String segment;
```
tkaitchuck: Do we actually need to pass the segment here?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2081,4 +2102,15 @@ public boolean isFailure() {
         }
     }
 
+    @Data
+    public static final class Flush implements WireCommand {
+        final WireCommandType type = WireCommandType.FLUSH_SEGMENT;
+        final UUID writerId;
+        final String segment;
```
tkaitchuck: Same

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
kmgowda: ok, I have changed to flushInFlight

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: in case of flush and close append; we flush the session of give writerid and segment. hence no need of increment the token counter for flush (one session ) ; its required for flushAll (all pending sessions)  triggered by blocktimeout or any wire command.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
kmgowda: yes, see the setupSegments table;  Entry is a pair of <writer id, segment>.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
kmgowda: yes, see the setupSegments table;  Entry is a pair of <writer id, segment>.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
kmgowda: Not exactly  100%; However,  I have reordered them now..
1) close append
       - break the existing append
        - flush the pending session data
       - remove the entry from setupSegments.
       - send the wire command
2)  flush command
       - if the session has any data ( if the give session of writer id and segment does not have any data, then no need to break the append)  then break the append
       - flush the pending data
                 NOTE : no wire command to segment store
3) block timeout
        -  break the append
        -  write all pending data
             NOTE : no wire command to segment store.

you have mentioned " In all cases all appends for a particular writerId" 
   for this we need type cast to  some message type ..
please let me know your thoughts

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
tkaitchuck: What about the path on line 209? That does not increment the counter, so the block timeout will still occur and end up flushing some unrelated write that happens later unnecessarily.

kmgowda: yes;  block timeout is required; theoretically , there may be pending data for other sessions.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
kmgowda: @tkaitchuck 
yes  you are right..
I have done the changes setupSegments -> setupWriters with writer id as key;
No, this enable us to remove the segment name from flush and close wire command.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
tkaitchuck: Ok, so there is a rather complex invariant we are trying to maintain here:
When appending there is only one session with a block open, and that sessions data goes out while others' accumulates. 
And if data has been waiting for more than a certain timeout, it gets flushed.

I am not exactly sure how best to do it. But it would be good if we could restructure this so that those are made more explicit. Perhaps checking against the session that owns the channel rather than a global token count? (With some sort of block id...)

kmgowda: even I tried couple of experiments; but having separate timeouts each of writer ids sessions may load the client with timers. having single block timeout for multiple session would be less in load.

tkaitchuck: Ok. Let's keeps it as a block. Below I think that should be:
```
            if (isChannelFree()) {
                if (session.isFree()) {
                    session.record(append);
                    startAppend(ctx, blockSizeSupplier, append, out);
                    continueAppend(data, out);
                    if (bytesLeftInBlock == 0) {
                        completeAppend(null, out);
                        tokenCounter.incrementAndGet();   // <--- added
                    }
                } else {
                  //... Is this branch reachable? 
                  //... Maybe this should throw an IllegalStateException?
                }
             }
```

kmgowda: '''
 if (isChannelFree()) {
                if (session.isFree()) {
                    session.record(append);
                    startAppend(ctx, blockSizeSupplier, append, out);
                    continueAppend(data, out);
                    if (bytesLeftInBlock == 0) {
                        completeAppend(null, out);
                        tokenCounter.incrementAndGet();   // <--- added
 
here, while active session is writing on wire; other sessions accumulates in queue (pendingWrites).  The block timeout is required for pendingWrites

                    }
                } else {


                  //... Is this branch reachable? 
                  //... Maybe this should throw an IllegalStateException?

Yes, this branch is reachable and its valid case;  if the append completes and channel becomes free ; then if the write comes on a session in which there is already accumulated data exists , at this time , new data is append to accumulated data and flushed to wire.

                }
             }
'''

tkaitchuck: Ok. lets add coverage for this case.

kmgowda: @tkaitchuck 
Yes, I have added the test coverage now.. please let me know if you have any further comments.

kmgowda:
Hi @andreipaduroiu and @fpj 
I request please dont close this PR and merge it.

Issue #4083, opened 2019-08-09
License header should not be doc comment


tisonkun:
**Problem description**

License header is not a kind of document comments indeed. Besides, it would confuse some IDEs like IDEA when reformatting the file. Specifically,

```
diff --git a/common/src/main/java/io/pravega/common/auth/ZKTLSUtils.java b/common/src/main/java/io/pravega/common/auth/ZKTLSUtils.java
index 293ba81c3..bebde8375 100644
--- a/common/src/main/java/io/pravega/common/auth/ZKTLSUtils.java
+++ b/common/src/main/java/io/pravega/common/auth/ZKTLSUtils.java
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
  */
 package io.pravega.common.auth;
```

change it into

```
/*
 * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
 * You may obtain a copy of the License at
 *
 *     http://www.apache.org/licenses/LICENSE-2.0
 */
```

could solve the problem.

**Problem location**

Nearly all of our java files.

**Suggestions for an improvement**

It doesn't require a dedicated pull request or commit for the change, but any new pull requests/commits tough a new file could do this handily.

tisonkun:
not a big deal.

closing....

PR #4130, opened 2019-08-21
Issue 4083: License Header is multi-line comment

TisonKun:
**Change log description**  

See also this [comment](https://github.com/pravega/pravega/pull/4021#discussion_r314118878) and description #4083 .

License header is not a kind of document comments indeed. Besides, it would confuse some IDEs like IDEA when reformatting the file. Change it into multi-line comment could solve the problem.

However, since we want to prevent style-only commit as well as the impact changing all license headers possibly non-functionally conflicts with a lot of ongoing efforts, we reach a consensus that just update `checkstyle/checkstyle.xml` for allowing multi-line comment style license header. It is hopefully we change license headers on the fly with other pull requests and we eventually adopt a multi-line style only codebase.

**Purpose of the change**  

Allow license header to be multi-line comment.

**What the code does**  

1. Replace `/\*\*\n \* Copyright \(c\)` with `/\*\*?\n * Copyright \(c\)` in `checkstyle/checkstyle.xml`

**How to verify it**  

trivial work.

ref #4083 

cc @andreipaduroiu @fpj


TisonKun:
> A nice side effect is that if different contributors have different formatting settings, we won't be seeing a slew of formatting-only changes in our commit history - this helps reduce noise in PRs.

I can see your concerns. Also I locally use a workaround the same as yours :-)

If we decide not to rush in a huge commit then I would propose to change our checkstyle rules from `/\*\*\n \* Copyright \(c\)` to `/\*\(\*?)n \* Copyright \(c\)` and generally replace it on touched intentionally among further commits. What do you think about this approach?

TisonKun:
Hi @andreipaduroiu I've updated the pull request. What do you think with the last proposal? Shall we keep context of this topic or just close it for a reason?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -277,4 +277,4 @@ public void close() {
             tlsCertFileModificationMonitor.stopMonitoring();
         }
     }
-}
+}
```
fpj: This change is unrelated to this PR, please undo it.

checkstyle/checkstyle.xml
```diff
@@ -9,7 +9,7 @@
 
   <!-- header -->
   <module name="RegexpHeader">
-    <property name="header" value="/\*\*\n * Copyright \(c\)"/>
+    <property name="header" value="/\*\*?\n * Copyright \(c\)"/>
```
fpj: The PR description no longer reflects this change. Could you please update, @TisonKun ?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
TisonKun: Done.

checkstyle/checkstyle.xml
TisonKun: Updated

TisonKun:
Thanks for your review @fpj . Addressed comment.

TisonKun:
Any update?

fpj:
hi @TisonKun , I'd rather not make this change. you are right that we could and perhaps even should have formatted the license header differently, but I'd rather not mix formats for the license header. we should keep this in mind for the future in the case we move the project under a different umbrella (_e.g._, ASF).

let me also say that I don't find it good practice to reformat existing files as you risk introducing unrelated formatting changes. pull requests, at least for this project, should not contain unrelated formatting changes. formatting changes should be in a separate pull request.

TisonKun:
Thanks for your information.

PR #4131, opened 2019-08-21
Issue 4129: Ensure that a failure to connect propagates upward

### LINKED ISSUE #4129: Connect hangs in connectionPoolImpl if the connection fails on hello.
**Problem description**
In `ConnectionPoolImpl` if establishConnection is called for a new connection, and inside of `FlowHandler.channelRegistered` the send of the `Hello` fails, then `channelUnregistered` will be invoked. However this happens before the `ChannelFutureListener` inside of `establishConnection` so in the final `CompletableFutures.allOf` at the end of the method it will create a future that can never return, because the future passed to `completeWhenRegistered` is not going to be run.

**Problem location**
`ConnectionPoolImpl.establishConnection` 

Switch the logic from an `allOf` to a `thenCompose` so that a failure of the first future will fail the result without waiting for the second future.

tkaitchuck:

**Change log description**  
Ensure that a failure to connect propagates upward 

**Purpose of the change**  
Fixes #4129 

**What the code does**  
Inside of `ConnectionPool.establishConnection()` two futures are created, and the output is dependent on `allOf` them. The problem with this is the method explicitly blocks until both complete even in the exception case. So if the first future fails, the returned future will block forever as the second future will never be completed. This changes the logic so that the second future depends on the first.

**How to verify it**  
System tests should be run against this branch to verify the problem does not reoccur.


client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
```diff
@@ -204,11 +204,11 @@ public int getActiveChannelCount() {
         } catch (Exception e) {
             connectionComplete.completeExceptionally(new ConnectionFailedException(e));
         }
-
-        final CompletableFuture<Void> channelRegisteredFuture = new CompletableFuture<>(); //to track channel registration.
-        handler.completeWhenRegistered(channelRegisteredFuture);
-
-        return CompletableFuture.allOf(connectionComplete, channelRegisteredFuture);
+        return connectionComplete.thenCompose(v ->  {
```
shrids: +1, we should not wait for the `channelRegistredFuture` to be completed (either exceptionally or successfully) if the CompletableFuture `connectionComplete` completes exceptionally.

client/src/main/java/io/pravega/client/netty/impl/FlowHandler.java
```diff
@@ -214,6 +213,7 @@ public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
                 log.warn("Encountered exception invoking ReplyProcessor for flow id {}", flowId, e);
             }
         });
+        registeredFutureLatch.releaseExceptionally(new ConnectionClosedException());
```
shrids: https://github.com/pravega/pravega/blob/177eed2f187d7b88a7084619d3ad5bef162ff282/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java#L171-L178

This would prevent segment readers from retrying, right?

tkaitchuck: That getConnection call on line 180 is going to end up causing the retry to happen on a new connection. This one connection won't be re-used once it is unregistered, the client will simply create a new one. I believe we don't currently have any code paths that attempt to re-use an unregistered connection. Please let me know if I am wrong about that.

shrids: Took a second look at the code flow, yes we do not have any code path that attempts to re-use an unregistered connection.

shrids:
System test run was successful, ref: `job/test-pravega-continuous-1/372/`

PR #4136, opened 2019-08-22
Issue 4135: Additional logging in Controller for easier debugging of missing delegation token issues

### LINKED ISSUE #4135: Logging statements to enable debugging of auth/delegation token issues
**Problem description**
An issue has been encountered where the delegation token was found empty causing the Segment Store to reject requests. The cause for the delegation token being empty is not immediately apparent from the current logs. Additional logging should be done to enable debugging of issues like this. 

**Problem location**

Add additional log statements.

ravisharda:
**Change log description**  
Adds logging statements in the Controller to aid debugging of failure scenarios where delegation token might be null or empty. 

**Purpose of the change**  
Resolved #4135 

**What the code does**  
Adds logging statements that will emit log entries under these circumstances:
1. When an auth handler plugin throws `RuntimeException` upon invocation of authorization operation.
2.  When auth is enabled, but the delegation token being generated is observed as null or empty, at some key points in the Controller gRPC service implementation class. 

The new logging statements being added log at `WARN` level, so it can be alarming for the admins/operators if the logging is unnecessary and excessive. The conditions mentioned above should be few if any, and therefore the new log statements should in turn occur under rare circumstances. 

**How to verify it**  
NA

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
@@ -75,7 +75,13 @@ public class GrpcAuthHelper {
 
     public String checkAuthorizationAndCreateToken(String resource, AuthHandler.Permissions expectedLevel) {
         if (isAuthEnabled) {
-            checkAuthorization(resource, expectedLevel);
+            try {
+                checkAuthorization(resource, expectedLevel);
+            } catch (RuntimeException e) {
+                // An auth handler plugin loaded in the system may throw this exception.
+                log.warn("Authorization failed", e);
+                throw e;
+            }
             return createDelegationToken(resource, expectedLevel, tokenSigningKey);
         }
         return "";
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -497,18 +497,9 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
 
     private void logIfEmpty(String delegationToken, String requestName, String scopeName,
                             String streamName) {
-        if (isAuthEnabled()) {
-            String message = String.format("Delegation token for request [%s] with scope [%s] and stream [%s], is: [%s]",
+        if (isAuthEnabled() && (delegationToken == null || delegationToken.equals(""))) {
+            log.warn("Delegation token for request [{}] with scope [{}] and stream [{}], is: [{}]",
                     requestName, scopeName, streamName, delegationToken);
-            logIfEmpty(delegationToken, message);
-        }
-    }
-
-    private void logIfEmpty(String delegationToken, String message) {
-        if (isAuthEnabled()) {
-            if (delegationToken == null || delegationToken.equals("")) {
-                log.warn(message);
-            }
         }
     }
 
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -196,19 +196,23 @@ public class ControllerServiceImpl extends ControllerServiceGrpc.ControllerServi
                 AuthResourceRepresentation.ofStreamInScope(request.getStreamInfo().getScope(),
                         request.getStreamInfo().getStream()),
                 AuthHandler.Permissions.READ_UPDATE),
-                delegationToken -> controllerService.getSegmentsAtHead(request.getStreamInfo().getScope(),
-                        request.getStreamInfo().getStream())
-                                       .thenApply(segments -> {
-                                           SegmentsAtTime.Builder builder = SegmentsAtTime.newBuilder()
-                                                                                          .setDelegationToken(delegationToken);
-                                           for (Entry<SegmentId, Long> entry : segments.entrySet()) {
-                                               builder.addSegments(SegmentLocation.newBuilder()
-                                                                                  .setSegmentId(entry.getKey())
-                                                                                  .setOffset(entry.getValue())
-                                                                                  .build());
-                                           }
-                                           return builder.build();
-                                       }),
+                delegationToken -> {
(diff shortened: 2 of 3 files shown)


controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
```diff
@@ -75,7 +75,13 @@ public String checkAuthorization(String resource, AuthHandler.Permissions expect
 
     public String checkAuthorizationAndCreateToken(String resource, AuthHandler.Permissions expectedLevel) {
         if (isAuthEnabled) {
-            checkAuthorization(resource, expectedLevel);
+            try {
+                checkAuthorization(resource, expectedLevel);
+            } catch (RuntimeException e) {
+                // An auth handler plugin loaded in the system may throw this exception.
+                log.warn("Authorization failed", e);
```
fpj: Is this actionable? The main difference between warn and error is that warn is something that the user should know, but it is not necessarily actionable, while an error typically indicates an unrecoverable situation that requires action from the user.

I see that you used warn for all new messages. Are they really warnings according to the definition above?

andreipaduroiu:
Please change the title of this PR to meet our PR naming pattern.

controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
ravisharda: It is not actionable at this point. A custom auth handler may throw an exception and we want to have a log of it. This may be a temporary issue if the custom Auth Handler talks to an Identity and Access Management system. Accordingly, I think `WARN` level is appropriate in this case.

The same is the case for other log statements I've added. I have done checks for whether auth is enabled, and if it is enabled, a null or empty delegation token hints to a potential problem.

ravisharda:
> e `WARN` logs making them hard to parse. Please change

@andreipaduroiu I think the new log statements should get triggered rarely since they occur either when the Auth Handler throws an unexpected RuntimeException or when Auth is enabled, but delegation token generated is null/empty. Am I missing something?

ravisharda:
> t our PR naming patt

Sorry, I somehow forgot the guideline when specifying the title for the PR. @fpj thanks for updating the title.

andreipaduroiu:
@ravisharda That is the kind of explanation I would have expected in the description of this PR. I do not want to pollute the logs unnecessarily. If you could update the description of the PR with your analysis/estimate of how often (and under what conditions) these would be logged, I would be more inclined to accept them as WARN (instead of DEBUG).

ravisharda:
> @ravisharda That is the kind of explanation I would have expected in the description of this PR. I do not want to pollute the logs unnecessarily. If you could update the description of the PR with your analysis/estimate of how often (and under what conditions) these would be logged, I would be more inclined to accept them as WARN (instead of DEBUG).

Sure @andreipaduroiu, that sounds good. I fully understand your concern. I've updated the description, as per your suggestion. Thanks.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -477,19 +485,38 @@ public void getDelegationToken(StreamInfo request, StreamObserver<DelegationToke
...
+                            .setDelegationToken(delegationToken)
+                            .build());
+                },
                 responseObserver);
     }
 
+    private void logIfEmpty(String delegationToken, String requestName, String scopeName,
+                            String streamName) {
+        if (isAuthEnabled()) {
+            String message = String.format("Delegation token for request [%s] with scope [%s] and stream [%s], is: [%s]",
```
fpj: This string should only be constructed if we are sure we are going to output it. Otherwise, we will be constructing these strings unnecessarily. Please reorganize this logic across these two methods to log if empty to avoid constructing these strings unnecessarily.

ravisharda: Sure, that makes sense. I've modified the code and pushed the changes.

controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
```diff
@@ -75,7 +75,13 @@ public String checkAuthorization(String resource, AuthHandler.Permissions expect
 
     public String checkAuthorizationAndCreateToken(String resource, AuthHandler.Permissions expectedLevel) {
         if (isAuthEnabled) {
-            checkAuthorization(resource, expectedLevel);
+            try {
+                checkAuthorization(resource, expectedLevel);
+            } catch (RuntimeException e) {
```
shiveshr: Would it have to be runtime or can we narrow the scope to a specific base exception?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -477,19 +485,29 @@ public void getDelegationToken(StreamInfo request, StreamObserver<DelegationToke
...
+                            .newBuilder()
+                            .setDelegationToken(delegationToken)
+                            .build());
+                },
                 responseObserver);
     }
 
+    private void logIfEmpty(String delegationToken, String requestName, String scopeName,
+                            String streamName) {
+        if (isAuthEnabled() && (delegationToken == null || delegationToken.equals(""))) {
```
shiveshr: we can use guava's Strings.isNullOrEmpty

controller/src/main/java/io/pravega/controller/server/rpc/auth/GrpcAuthHelper.java
ravisharda: The Auth Handler implementation may be an externally supplied plugin and it might throw a RuntimeException (we don't wrap the handler at this point). Here, we log the exception, just in case it does that, to aid debugging.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: Sure, that sounds good. I've modified the code and pushed the change.

Issue #173, opened 2016-11-16
Work out security principals for different calls


tkaitchuck:
Streams have reading events, writing events, getting metrics, config etc,
We need to work out how these fit together with the larger authorization roles.

PR #4141, opened 2019-08-26
Issue 3854 : K8 system tests fail with operator 0.4.0

### LINKED ISSUE #3854: K8 system tests fail with operator 0.4.0

SrishT:
**Change log description**  
After the 0.4.0 release of the operator, Kubernetes-based system tests started to fail. The reason for these failures was that the operator was unable to create the new admission webhook component due to lack of permissions. This PR solves that issue by assigning the requisite permissions.

**Purpose of the change**  
Resolves #3854

**What the code does**  
Here's what the code in this PR does :

1. It adds the `ClusterRole` and `ClusterRoleBinding` into the test framework.
2. It changes the custom resource definition provided in the test file, incorporating the changes introduced in the `0.4.0` release of the operator.
3. Patches the `PravegaCluster` spec in case of scale tests instead of rewriting the entire spec.

**How to verify it**  
System Tests should pass


test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,38 +127,82 @@ public String getID() {
...
+                    .put("resources", getResources("2000m", "5Gi", "1000m", "3Gi"))
+                    .put("storage", ImmutableMap.builder()
+                            .put("indexVolumeClaimTemplate", bkPersistentVolumeSpec)
+                            .put("ledgerVolumeClaimTemplate", bkPersistentVolumeSpec)
+                            .put("journalVolumeClaimTemplate", bkPersistentVolumeSpec)
+                            .build())
+                    .put("autoRecovery", true)
+                    .build();
+
+            final Map<String, Object> pravegaSpec = ImmutableMap.<String, Object>builder().put("controllerReplicas", controllerCount)
```
pbelgundi: To avoid code repetition and improve modularity you could have spec builder classes for different operator versions (>=0.4.0 vs older) and depending on the builder used the right specs would get generated and populated here.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
```diff
@@ -97,7 +97,18 @@ public boolean isRunning() {
                             log.debug("Current instance counts : Bookkeeper {} Controller {} SegmentStore {}.", currentBookkeeperCount,
                                       currentControllerCount, currentSegmentStoreCount);
                             if (currentBookkeeperCount != newInstanceCount) {
-                                return deployPravegaUsingOperator(zkUri, currentControllerCount, currentSegmentStoreCount, newInstanceCount, properties)
+                                final Map<String, Object> bookieSpec = ImmutableMap.<String, Object>builder()
```
pbelgundi: Please create bookie spec in a separate method and call it here. Keeps code more readable.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/PravegaControllerK8sService.java
```diff
@@ -100,7 +100,18 @@ public boolean isRunning() {
                            log.debug("Current instance counts : Bookkeeper {} Controller {} SegmentStore {}.", currentBookkeeperCount,
                                      currentControllerCount, currentSegmentStoreCount);
                            if (currentControllerCount != newInstanceCount) {
-                               return deployPravegaUsingOperator(zkUri, newInstanceCount, currentSegmentStoreCount, currentBookkeeperCount, properties)
+                               final Map<String, Object> pSpec = ImmutableMap.<String, Object>builder()
```
pbelgundi: Same as bookkeeper service

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/PravegaSegmentStoreK8sService.java
```diff
@@ -97,7 +97,18 @@ public boolean isRunning() {
                            log.debug("Current instance counts : Bookkeeper {} Controller {} SegmentStore {}.", currentBookkeeperCount,
                                      currentControllerCount, currentSegmentStoreCount);
                            if (currentSegmentStoreCount != newInstanceCount) {
-                               return deployPravegaUsingOperator(zkUri, currentControllerCount, newInstanceCount, currentBookkeeperCount, properties)
+                               final Map<String, Object> pSpec = ImmutableMap.<String, Object>builder()
```
pbelgundi: Same as bookkeeper service

pbelgundi:
Travis build seems to be failing because of some test failure.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,40 +127,74 @@ public String getID() {
...
+        final Map<String, Object> pravegaImgSpec;
+
+        if (isLaterOperator) {
+
+            bookkeeperImgSpec = ImmutableMap.of("repository", bookkeeperImg);
+            pravegaImgSpec = ImmutableMap.of("repository", pravegaImg);
+
+        } else {
+
+            bookkeeperImgSpec = getImageSpec(bookkeeperImg, PRAVEGA_BOOKKEEPER_VERSION);
```
pbelgundi: Do we really need getImageSpec()? we can directly create the ImmutableMap in the else block. That would keep code more readable.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,40 +127,74 @@ public String getID() {
...
+            pravegaImgSpec = getImageSpec(pravegaImg, PRAVEGA_VERSION);
+
+        }
+
+        final Map<String, Object> bookkeeperSpec = ImmutableMap.<String, Object>builder().put("image", bookkeeperImgSpec)
+                .put("replicas", bookieCount)
+                .put("resources", getResources("2000m", "5Gi", "1000m", "3Gi"))
```
pbelgundi: With this, I guess resources would get added when using operator 0.3.2 as well?
Lets make sure this is supported by earlier operator versions.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,40 +127,74 @@ public String getID() {
...
                 .put("metadata", ImmutableMap.of("name", PRAVEGA_ID, "namespace", NAMESPACE))
-                .put("spec", ImmutableMap.builder().put("zookeeperUri", zkLocation)
-                                         .put("bookkeeper", bookkeeperSpec)
-                                         .put("pravega", pravegaSpec)
-                                         .build())
+                .put("spec", pravegaClusterSpecBuilder(zkLocation, bookkeeperSpec, pravegaSpec))
                 .build();
     }
 
+    protected Map<String, Object> pravegaClusterSpecBuilder(String zkLocation, Map<String, Object> bookkeeperSpec, Map<String, Object> pravegaSpec) {
```
pbelgundi: Use verbs for methods. - buildPravegaSpec()

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,40 +127,74 @@ public String getID() {
...
+        if (isLaterOperator) {
+            return ImmutableMap.<String, Object>builder()
```
pbelgundi: You could have 1 map with common entries like:
```
ImmutableMap<String, String> commonEntries = ImmutableMap.<String, Object>builder()
                    .put("zookeeperUri", zkLocation)
                    .put("bookkeeper", bookkeeperSpec)
                    .put("pravega", pravegaSpec)
                    .build();
```
and then for 0.4.0 spec, just do: 
```
ImmutableMap<String, String> immutableMap = ImmutableMap.<String, String>builder()
  .putAll(commonEntries)
  .put("version", PRAVEGA_VERSION)
  .build();
```

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -130,77 +130,67 @@ public String getID() {
...
+                .build();
+    }
+
+    protected Map<String, Object> pravegaClusterSpecBuilder(String zkLocation, Map<String, Object> bookkeeperSpec, Map<String, Object> pravegaSpec) {
+        if (isLaterOperator) {
+            return ImmutableMap.<String, Object>builder()
+                    .put("zookeeperUri", zkLocation)
+                    .put("bookkeeper", bookkeeperSpec)
+                    .put("pravega", pravegaSpec)
+                    .put("version", PRAVEGA_VERSION)
```
shrids: If adding the version is the only additional step then we can refactor the code further.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -195,8 +236,30 @@ public String getID() {
...
+                .put("requests", ImmutableMap.builder()
+                        .put("cpu", requestsCpu)
+                        .put("memory", requestsMem)
+                        .build())
+                .build();
+    }
+
     private V1beta1CustomResourceDefinition getPravegaCRD() {
 
+        String pravegaOperatorTag = PRAVEGA_OPERATOR_IMAGE.substring(PRAVEGA_OPERATOR_IMAGE.lastIndexOf(":") + 1);
```
pbelgundi: Why are we doing this inside getPravegaCRD?
Please create a separate method from this code and invoke it @ln 95, where isLaterOperator is defined. That way isLaterOperator can be final.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
```diff
@@ -81,6 +81,22 @@ public boolean isRunning() {
...
     }
 
+    protected Map<String, Object> patchSpecBuilder(int newInstanceCount) {
+
+        final Map<String, Object> bookkeeperSpec = ImmutableMap.<String, Object>builder()
+                .put("replicas", newInstanceCount)
+                .build();
+
+        return ImmutableMap.<String, Object>builder()
+                .put("apiVersion", "pravega.pravega.io/v1alpha1")
```
shrids: The same code is getting duplicated across BookkeeperK8sService, PravegaControllerK8sService and PravegaSegmentStoreK8sService. Could you refactor it?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
```diff
@@ -81,6 +81,22 @@ public boolean isRunning() {
...
+                .put("kind", CUSTOM_RESOURCE_KIND_PRAVEGA)
+                .put("metadata", ImmutableMap.of("name", PRAVEGA_ID, "namespace", NAMESPACE))
+                .put("spec", ImmutableMap.builder()
```
shrids: Can you elaborate on what has changed in the lastest `scaleService()` logic here ?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
```diff
@@ -81,6 +81,22 @@ public boolean isRunning() {
                                               t -> new TestFrameworkException(RequestFailed, "Failed to fetch ServiceDetails for bookkeeper", t));
     }
 
+    protected Map<String, Object> patchSpecBuilder(int newInstanceCount) {
+
+        final Map<String, Object> bookkeeperSpec = ImmutableMap.<String, Object>builder()
+                .put("replicas", newInstanceCount)
+                .build();
+
+        return ImmutableMap.<String, Object>builder()
```
pbelgundi: Create a new utility method (Utils.java) that takes pravegaSpec as parameter and adds the rest of the values in this code. It can be invoked here as well as in ln #173

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -163,6 +164,22 @@ public static ClientConfig buildClientConfig(URI controllerUri) {
 
     }
 
+    public static Map<String, Object> buildPatchedPravegaClusterSpec(String replicas, int replicaCount, String component, String namespace, String name, String kind) {
```
RaulGracia: Please, add comments describing this public method.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -85,6 +89,7 @@
     private static final String BOOKKEEPER_IMAGE_NAME = System.getProperty("bookkeeperImageName", "bookkeeper");
     private static final String TIER2_NFS = "nfs";
     private static final String TIER2_TYPE = System.getProperty("tier2Type", TIER2_NFS);
+    private static final boolean LATER_OPERATOR = getOperatorVersion();
```
RaulGracia: Should this field be named as `OPERATOR_VERSION`?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
RaulGracia: In fact, this looks to be the same change as for this open PR, right? https://github.com/pravega/pravega/pull/3980

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,45 +127,85 @@ public String getID() {
...
-        return ImmutableMap.<String, Object>builder().put("repository", imageName)
-                                                     .put("tag", tag)
-                                                     .put("pullPolicy", IMAGE_PULL_POLICY)
-                                                     .build();
+    private static boolean getOperatorVersion() {
+        String pravegaOperatorTag = PRAVEGA_OPERATOR_IMAGE.substring(PRAVEGA_OPERATOR_IMAGE.lastIndexOf(":") + 1);
+        if (!pravegaOperatorTag.equals("latest")) {
+            String version = pravegaOperatorTag.substring(0, pravegaOperatorTag.lastIndexOf("."));
+            float v = Float.parseFloat(version);
+            if (v < 0.4) {
```
RaulGracia: This method is named as `getOperatorVersion`, so why is it returning a Boolean? It should return the operator version so you can deal with the different cases based on the version number returned by this method.

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
```diff
@@ -51,7 +51,7 @@
     private static final String SCOPE = "testBookieFailoverScope" + RandomFactory.create().nextInt(Integer.MAX_VALUE);
     private static final int NUM_WRITERS = 5;
     private static final int NUM_READERS = 5;
-    private static final int BOOKIE_FAILOVER_WAIT_MILLIS = 20 * 1000;
+    private static final int BOOKIE_FAILOVER_WAIT_MILLIS = 15 * 1000;
```
RaulGracia: Is this change related with this PR?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,45 +127,85 @@ public String getID() {
...
-                                                                                      //.put("options", props.getProperties())
-                                                                                      .put("options", props)
-                                                                                      .put("image",
-                                                                                       getImageSpec(DOCKER_REGISTRY + PREFIX + "/" + PRAVEGA_IMAGE_NAME, PRAVEGA_VERSION))
-                                                                                      .put("tier2", tier2Spec())
-                                                                                      .build();
+                .put("segmentStoreReplicas", segmentStoreCount)
+                .put("debugLogging", true)
+                .put("cacheVolumeClaimTemplate", pravegaPersistentVolumeSpec)
+                .put("controllerResources", getResources("2000m", "3Gi", "1000m", "1Gi"))
```
RaulGracia: Maybe it is advisable to merge https://github.com/pravega/pravega/pull/3980 before merging this PR, so we only focus on the compatibility issue in this PR.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
pbelgundi: Thats a boolean field indicating if the operator version is >= 0.4.0 or lower than that.
Yes, we could do with a better name here. 
Perhaps `IS_OPERATOR_VERSION_ABOVE_0_4_0`

RaulGracia:
In a discussion with @SrishT, I expressed my point of view regarding the scope of this PR: now, the code has very narrow scope, as it only deals with operator versions before and after `0.4`. However, if I'm not mistaken, in future versions of the operator we may also need to introduce changes here. My point is basically to embrace a code structure that makes it easier to deal with future changes in the test framework required by the operator (e.g., `OPERATOR_VERSION` needs to be a number instead of a boolean, treat `OPERATOR_VERSION` ranges in separate blocks, encapsulate re-usable logic to make the class more readable, etc.).

pbelgundi:
> In a discussion with @SrishT, I expressed my point of view regarding the scope of this PR: now, the code has very narrow scope, as it only deals with operator versions before and after `0.4`. However, if I'm not mistaken, in future versions of the operator we may also need to introduce changes here. My point is basically to embrace a code structure that makes it easier to deal with future changes in the test framework required by the operator (e.g., `OPERATOR_VERSION` needs to be a number instead of a boolean, treat `OPERATOR_VERSION` ranges in separate blocks, encapsulate re-usable logic to make the class more readable, etc.).

Note that every new operator version need not be non-backward compatible as was the case with 0.3.2 and 0.4.0. We don;t expect this to happen very often. 
I don;t see the need to complicate the code too much at this point by providing for explicit version range checks or other mechanisms, which may not be useful for many future operator versions to come. A simple if else block should be good for now.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
pbelgundi: Since #3980 is not merged so far and resource changes are tested with operator 0.3.0 and operator 0.4.0 with this PR, I would prefer we merge this and close that PR.

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
SrishT: Yes, as the the wait time is stalling the test for a longer time due to which the BookieFailoverTest is failing for operator `0.4.0`

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: Both the issues can be handled from this PR itself hence I closed that PR

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: Verified that this change is compatible with operator 0.3.2 as well

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: Yes as it is being called from other places as well

RaulGracia:
@pbelgundi the changes I have discussed with @SrishT are not different than the `if/else` logic that is already there; I just suggested a couple of small changes to make it clearer and easier for anyone looking at the code. I understand that the kind of changes in the operator that motivated this PR are not going to happen often, but this does not mean that they won't happen never again. I think that just changing a bit the current code to make more explicit the relationship between the versions of the operator and the different logic executed is not a big deal, makes the code more understandable for people not very familiar to the operator, and proposes a basic structure to further extensions of the same kind in that class.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/PravegaSegmentStoreK8sService.java
```diff
@@ -81,6 +81,22 @@ public boolean isRunning() {
                                               t -> new TestFrameworkException(RequestFailed, "Failed to fetch ServiceDetails for pravega-segmentstore", t));
     }
 
+    protected Map<String, Object> patchSpecBuilder(int newInstanceCount) {
+
+        Map<String, Object> pravegaSpec = ImmutableMap.<String, Object>builder()
+                .put("segmentStoreReplicas", newInstanceCount)
+                .build();
+
+        return ImmutableMap.<String, Object>builder()
```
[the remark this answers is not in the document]
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

SrishT:
> It looks much better now.
> There are some comments around coding conventions and reducing code repetition.
> Please address these and publish test results for 0.4.0 operator as well as 0.3.2 operator after your changes.

Publishing test results for both operator versions
- With Operator 0.3.0 : https://asd-nautilus-jenkins.isus.emc.com/job/systest-srishti/21/testReport/
- With Operator 0.4.0 : https://asd-nautilus-jenkins.isus.emc.com/view/Pravega/job/test-pravega-continuous-1/356/testReport/

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -163,6 +164,22 @@ public static ClientConfig buildClientConfig(URI controllerUri) {
...
+
+        final Map<String, Object> componentSpec = ImmutableMap.<String, Object>builder()
+                .put(replicas, replicaCount)
+                .build();
+
+        return ImmutableMap.<String, Object>builder()
+                .put("apiVersion", "pravega.pravega.io/v1alpha1")
```
shrids: `kubernetes.AbstractService#CUSTOM_RESOURCE_GROUP_PRAVEGA` and `kubernetes.AbstractService#CUSTOM_RESOURCE_VERSION_PRAVEGA` refer to the above string, we should reuse it for easier maintainability.
We could move this to `io.pravega.test.system.framework.services.kubernetes.AbstractService`.

SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,45 +128,120 @@ public String getID() {
...
         // generate Pravega Spec.
         final Map<String, Object> pravegaPersistentVolumeSpec = getPersistentVolumeClaimSpec("20Gi", "standard");
+        
+        final String bookkeeperImg = DOCKER_REGISTRY + PREFIX + "/" + BOOKKEEPER_IMAGE_NAME;
+        final String pravegaImg = DOCKER_REGISTRY + PREFIX + "/" + PRAVEGA_IMAGE_NAME;
+        final Map<String, Object> bookkeeperImgSpec;
+        final Map<String, Object> pravegaImgSpec;
+
+        if (IS_OPERATOR_VERSION_ABOVE_040) {
+
```
RaulGracia: Minor: remove extra blank lines.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,45 +128,120 @@ public String getID() {
...
                 .build();
     }
 
-    protected Map<String, Object> getImageSpec(String imageName, String tag) {
-        return ImmutableMap.<String, Object>builder().put("repository", imageName)
-                                                     .put("tag", tag)
-                                                     .put("pullPolicy", IMAGE_PULL_POLICY)
-                                                     .build();
+    /**
+     * As upgrade has been included in operator 0.4.0
```
RaulGracia: Minor: use a fixed line width for comments (e.g., 80 chars). Otherwise, it looks odd.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,45 +128,120 @@ public String getID() {
...
+     * a few changes have been included in the Pravega Cluster Spec
+     * like the image tag field for pravega and bookeeper has been replaced by a
+     * single version field which indicates the version for both the pravega and bookeeper components
```
RaulGracia: Minor: please, use `Pravega` and `Bookkeeper` in capitals, to make it consistent with other places in the documentation/comments.

RaulGracia: Also, there is a missing dot at the end of the sentence.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -120,45 +128,120 @@ public String getID() {
...
+     * Hence determining whether the operator version to be used for running the System Tests is below
+     * 0.4.0 or not is of crucial importance
```
RaulGracia: Minor: missing dot at the end of the sentence.

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
RaulGracia: I do not understand why this sleep makes the test to fail with a different operator version. Could you please provide more details here?

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/BookkeeperK8sService.java
SrishT: Earlier in `scaleService()` all components in the spec were rewritten. As the test pod didn’t contain the arguments passed while running the test, the operator redeployed the components with the default values.
Now only those fields are mentioned in the spec whose values need to be patched, other values remain same as were specified at the time of deployment.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -85,6 +90,7 @@
     private static final String BOOKKEEPER_IMAGE_NAME = System.getProperty("bookkeeperImageName", "bookkeeper");
     private static final String TIER2_NFS = "nfs";
     private static final String TIER2_TYPE = System.getProperty("tier2Type", TIER2_NFS);
+    private static final boolean IS_OPERATOR_VERSION_ABOVE_040 = isOperatorVersionAbove040();
```
RaulGracia: Again, the change is only focused on the incompatibility before/post `0.4`. So if there is a new change in the operator that requires additional management here (e.g., between operator versions `0.9` and `1.X`), then we will need to refactor this class and then start working with "version ranges". I don't see the problem about starting to work in that way right from the beginning.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: As most of the code is the same in both cases with only one or two changes required for each, I don't think that would be necessary.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
pbelgundi: Could you please explain why this change cannot be introduced when it is needed?

I don't want to hold off this PR, for some future requirements that may not be needed for a very long time to come and don't look very compelling to me at this point.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
SrishT: done

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
SrishT: In the BookieFailover test, after bringing down  one bookie replica, we wait for twice the wait time before a new bookie instance is spawned again. In the meanwhile neither writes or reads will be performed. When the new bookie instance comes up, it again takes some time before the Segment Store is able to connect to this bookie and the reads and writes are able to resume only after this. If the wait time is kept long, the test times out before the reads and writes are able to complete.

RaulGracia: This does not answer my question. This test has been working for a long time without problems. Why changing the operator version requires changing this timeout? If this change is not required, then is better to remove it; if the change is actually needed, we need at least to understand why with the new operator this test requires a shorter timeout.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
pbelgundi: @SrishT It looks like you haven;t understood what I;m suggesting above. 
I'm suggesting builder classes based on version range, in the same lines as what Raul had suggested recently. It is definitely a better and more clean way of doing things, through not absolutely critical to this fix.

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
SrishT: An `InitialDelay` was added to the `Readiness Probe` of the bookie after `0.3.0` as can be seen here : https://github.com/pravega/pravega-operator/commit/3280f8b07609d18d1d2f5f834edbd5cbecf21001#diff-4620fbe56580da27f4724ef7207b8d07R126
This might be reason that bookie pods take a few extra seconds with operator `0.4.0`

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
SrishT: done

PR #4154, opened 2019-09-05
Issue 4153: Link Mark Stream's lifecycle explicitly with underlying stream. 

### LINKED ISSUE #4153: Delete Mark Stream explicitly whenever the stream is deleted.
**Problem description**
We create mark stream under user scope. This mark stream should have its lifecycle tied to the parent stream and should be sealed and deleted when parent stream is sealed and deleted. 

**Problem location**
Watermarking. 

tie lifecycle of mark stream to the underlying stream it is created for.

shiveshr:

**Change log description**  
As part of create stream workflow, we will also create mark stream associated with underlying stream. 
As part of delete stream workflow, we will also delete mark stream associated with the underlying stream. 

**Purpose of the change**  
Fixes #4153 

**What the code does**  

1. Create Mark stream proactively as opposed to lazily along with underlying stream's creation. 
So the changes are primarily in `PeriodicWatermarking.java` and `StreamMetadataTasks.java`.
CreateMarkStream method is removed from PeriodicWatermarking class. 
CreateMarkStream method is added to StreamMetadataTasks and it is invoked from within the `createStream` workflow where we create mark stream after creating the base stream but before setting it to active. 

2. As part of delete workflow, before deleting the stream, delete the mark stream as well. 
The changes are primarily in DeleteStreamTask where we delete mark stream as part of delete workflow for the base stream. 
Mark stream is deleted before deleting segments and metadata for the base stream

Since mark stream is now created along with base stream so all tests that perform listStreamsInScope had to be updated to expect presence of mark stream too. 

**How to verify it**  
Unit test and integration added


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteStreamTask.java
```diff
@@ -69,7 +70,8 @@ public DeleteStreamTask(final StreamMetadataTasks streamMetadataTasks,
                         log.warn(requestId, "{}/{} stream not sealed", scope, stream);
                         return Futures.failedFuture(new RuntimeException("Stream not sealed"));
                     }
-                    return notifyAndDelete(context, scope, stream, requestId);
+                    return deleteAssociatedStreams(scope, stream, requestId)
+                        .thenCompose(v -> notifyAndDelete(context, scope, stream, requestId));
```
tkaitchuck: I am concerned about this order. If this fails or gets delayed in the middle it looks like it leaves the associated streams deleted by the main stream still exists. This would appear as an invalid state on the client. If it were the other way around, the client would just see the stream as deleted and the lingering associated streams would be of no concern to it. This of course assumes they would get cleaned up eventually, which we would have to also ensure.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteStreamTask.java
```diff
@@ -81,6 +83,18 @@ public DeleteStreamTask(final StreamMetadataTasks streamMetadataTasks,
                 });
     }
 
+    private CompletableFuture<Void> deleteAssociatedStreams(String scope, String stream, long requestId) {
+        String markStream = NameUtils.getMarkStreamForStream(stream);
+        OperationContext context = streamMetadataStore.createContext(scope, markStream);
+        return streamMetadataStore.checkStreamExists(scope, markStream)
+            .thenCompose(exists -> {
+                if (exists) {
+                    return notifyAndDelete(context, scope, markStream, requestId);
```
tkaitchuck: Can notifyAndDelete be made idempotent so this exists check is not needed?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteStreamTask.java
shiveshr: once a delete has been issued on a stream, it will eventually get deleted and any client accessing the stream after its delete has been invoked can find the stream to be in a state where some resources for the stream are cleaned up while others are in the process of being cleaned up.
for example, we delete the segments followed by metadata.
so a client accessing the stream while a delete is ongoing will not be able to work with stream in a meaningful way anyway and that is the expectation. 
so same thing extends to mark streams.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteStreamTask.java
shiveshr: its actually idempotent, except it can throw DataNotFoundException if metadata is not found. 
i am catching that now instead of doing checkExists. 
Also, that uncovered a bug in InMemoryStream where for one case we do not throw DataNotFound which can cause NPE. Fixed that too.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteStreamTask.java
tkaitchuck: Deleting the segments followed by the metadata makes sense from a client POV. The segment goes away and it gets an exception which it can recognise and send upwards. With the mark segment it's weird, because not having a mark segment a valid state, it's a stream without marks. So we could end up in the situation where the error does not cleanly manifest on the client instead it looks like things are working correctly, but it is in fact returning data which is wrong.

shiveshr: The point i am making is - once delete request for a stream is accepted, we make no guarantees about client behavior if it tries to access any resources for the said stream. It can and will fail in different ways depending on resources that we have been able to delete at the time when client was accessing the stream or its associated resources. 

Anyway, if we want to delete associated resources after having deleted the stream, then we can do that too. Its not a problem per se. So let me make that change and update the PR.

tkaitchuck: > once delete request for a stream is accepted, we make no guarantees about client behavior if it tries to access any resources for the said stream. It can and will fail in different ways depending on resources that we have been able to delete at the time when client was accessing the stream or its associated resources.

This is not acceptable. We need to to fail with a clear error exception that users can understand.

shiveshr: I disagree. 
If someone has issued delete, the delete workflow is performed asynchronously and different resources are deleted non-atomically. 
so depending on which resource client is attempting to access, and where we are in the delete workflow, the client could get a different failure..
for example, as we delete segments for the stream, some segments may be deleted while others may not yet be deleted. So readers could read from some of them while they find others deleted. They would translate this to SegmentTruncatedException. 

Similarly, associated streams also fall in the same category as one of the resources for the stream.
if stream is being deleted, its associated streams may be deleted while stream metadata, which is last entity to be deleted, is still intact. 

It is important to note that the stream's deletion could complete at any moment so its not like failure to find associated streams will be catastrophic for the client.

> This also sounds like a problem. We should fix this, as deleting and then recreating streams is something users will reasonably do.

Yes, that is why deleting associated streams before deleting the underlying stream is the correct approach.

PR #4155, opened 2019-09-05
Issue 4161: Improve password handling based on static code analysis report.

### LINKED ISSUE #4161: Improve password handling inside PasswordAuthHandler
**Problem description**
Improve the password handling inside `PasswordAuthHandler` by zeroing out the password

**Problem location**
PasswordAuthHandler.java

Fix the issue to prevent leak via heap inspection.

shrids:
**Change log description**  
- Improve password handling inside Pravega.

**Purpose of the change**  
Fixes #4161 

**What the code does**  
Improve the password handling inside `PasswordAuthHandler` by zeroing out the password

**How to verify it**  

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
@@ -79,17 +79,19 @@ public class PasswordAuthHandler implements AuthHandler {
     public Principal authenticate(String token) throws AuthException {
         String[] parts = parseToken(token);
         String userName = parts[0];
-        String password = parts[1];
+        char[] password = parts[1].toCharArray();
 
         try {
             if (userMap.containsKey(userName) && encryptor.checkPassword(password, userMap.get(userName).encryptedPassword)) {
                 return new UserPrincipal(userName);
             }
+            throw new AuthenticationException("User authentication exception");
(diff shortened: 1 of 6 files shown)


controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java
```diff
@@ -79,17 +79,19 @@ public String getHandlerName() {
     public Principal authenticate(String token) throws AuthException {
         String[] parts = parseToken(token);
         String userName = parts[0];
-        String password = parts[1];
+        char[] password = parts[1].toCharArray();
```
ravisharda: Should there be a null check here, in case `parts[1]` is null? Or, maybe some sort of an input validation to ensure input is valid?

shrids: https://github.com/pravega/pravega/blob/7f76baa4674f1e572cc0571bb22f505ae0583076/controller/src/main/java/io/pravega/controller/server/rpc/auth/PasswordAuthHandler.java#L122-L126
No since `io.pravega.controller.server.rpc.auth.PasswordAuthHandler#parseToken` already validates if the `parts.length ==2` and `java.lang.String#split(java.lang.String, int)` ensures no null string is returned.

ravisharda: Sounds good.

PR #4166, opened 2019-09-12
Issue 4164: (SegmentStore) Bugfix for possible ingestion pipeline starvation due to throttling

### LINKED ISSUE #4164: Incorrect throttling-induced starvation of the ingestion pipeline
**Problem description**
In cases when the cache size is rather small in comparison to the ingestion throughput (i.e., 3GB cache with 150MB+ throughput) it is possible to fill up the cache pretty quicky. Moreover, we can "jump" from a situation with no throttling (cache utilization is just under the threshold of 85% for cleanup) to partial or full throttling in a mater of seconds (1 second at 150MBs is 5% of the 3GB cache).

In such situations, it is possible that we induce a lengthy throttle (10 seconds, for example), however that data was quickly moved to Tier 2 (sometimes in less than 1 second). In such cases, the ingestion pipeline is halted (due to the large throttle delay) but no data is flowing to Tier 2 either, because it's fully caught up.

The throttler works fine for situations where the cache utilization **gradually** increases, but it does not handles situations where it jumps quickly too gracefully. This problem is exacerbated by cases where we have a large number of Segment Containers on the same Segment Store instance.

**Problem location**
OperationProcessor.

The following improvement should fix the situation:
1. Change the Cache Manager eviction frequency from 5 seconds to 1 second (this will be our new granularity).
2. Allow each Operation Processor to register itself in the Cache Manager as a listener to all Cache Eviction events where something actually got evicted.
3. If such an event is raised and the Operation Processor is currently executing a cache-triggered throttle, cancel that throttler and re-evaluate the situation based on the current state of the cache. We may:
    - Proceed without any further throttling OR
    - Determine that the cache still exceeds throttling utilization threshold, at which point we should continue throttling (some adjustments to the delay may be needed here).

andreipaduroiu:
**Change log description**  
- Changed `CacheManager` to run every 1 second (down from 5). 
    - This will cause more frequent cache evictions and with higher granularity.
- Added the ability to cancel a throttling delay, if that delay was due to the cache being over-utilized and if the cache utilization has since dropped below the throttling threshold.
    - This will enable us to more quickly resume the ingestion pipeline if the data in the cache has been moved to Tier 2 before the throttling delay expired.

**Purpose of the change**  
Fixes #4164.

**What the code does**  
- `CacheManager` will now accept `CacheCleanupListeners`; anyone interested in being notified when the cache evicts at least one item can register itself here. The `CacheManager` will notify these listeners after every scheduled run that resulted in at least one entry being evicted.
- Moved throttling logic out of `OperationProcessor` into its own class to ease testing.
- Differentiating between interruptible throttle delays (currently only Cache) and non-interruptible. The latter will be carried over to their completion. The former can be cancelled if conditions permit, prior to their expiration.
    - Every throttle due to a Cache reason is recorded.
    - If the `Throttler` is notified by the `CacheManager` (via the `CacheCleanupListener` interface) that the cache has been cleaned up, then it will 
        - Cancel any throttling that was caused due to Caching (leaves any others in place).
        - Either allow operations to proceed (no more throttling) or throttling with a smaller delay than before. Both of these would be an effect of improved conditions (cache utilization was significantly reduced than before so no need to wait that long anymore). 

A few examples of how this might play out:
1. Initial delay is 20s. After 2s, the CacheManager evicts a lot of data. The Throttler determines there is no need to throttle anymore and cancels the initial throttling (saving 18s).
2. Initial delay is 20s. After 2s, the CacheManager evicts some data. The throttle determines that the new delay (based on current utilization) is now 10s; it sets the new delay to be 10s, thus saving 8s in total (20s - 2s - min(10s, 18s)).
3. Initial delay is 5s. After 2s, the CacheManager evicts a little bit of data, but some other actor added a lot more, thus increasing the utilization compared to before. The new throttle delay would be 20s, but the Throttler will only wait 3s (what was left of the initial delay).
    - This sounds counter intuitive, but it is necessary in order to avoid an infinite starvation situation (where each time we check, we simply extend the throttling by some amount). If the situation does get "worse" after the initial expiration, then whenever this pipeline resumes it will add a bit of data and then halt based on the new situation (which could be a maximum delay).

**How to verify it**  
New unit tests added.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
@@ -26,6 +26,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 import lombok.Getter;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
 /**
@@ -55,6 +56,8 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
     private final CachePolicy policy;
     private final AtomicBoolean closed;
     private final SegmentStoreMetrics.CacheManager metrics;
+    @GuardedBy("cleanupListeners")
+    private final HashSet<CleanupListener> cleanupListeners;
 
     //endregion
 
@@ -78,6 +81,7 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
         this.executorService = executorService;
         this.closed = new AtomicBoolean();
         this.metrics = new SegmentStoreMetrics.CacheManager();
+        this.cleanupListeners = new HashSet<>();
     }
 
     //endregion
@@ -117,7 +121,10 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
         }
 
         try {
-            applyCachePolicy();
+            boolean anythingEvicted = applyCachePolicy();
+            if (anythingEvicted) {
+                notifyCleanupListeners();
+            }
         } catch (Throwable ex) {
             if (Exceptions.mustRethrow(ex)) {
                 throw ex;
@@ -152,6 +159,19 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
     @Override
     public double getCacheMaxUtilization() {
         return this.policy.getMaxUtilization();
+    }
+
+    @Override
+    public void registerCleanupListener(@NonNull CleanupListener listener) {
+        if (listener.isClosed()) {
+            log.warn("{} Attempted to register a closed Cleanup Listener ({}).", TRACE_OBJECT_ID, listener);
+            System.out.println("registered closed listener");
+            return;
+        }
+
+        synchronized (this.cleanupListeners) {
+            this.cleanupListeners.add(listener); // This is a Set, so we won't be adding the same listener twice.
+        }
     }
 
     //endregion
@@ -199,13 +219,13 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
 
     //region Helpers
 
-    protected void applyCachePolicy() {
+    protected boolean applyCachePolicy() {
         // Run through all the active clients and gather status.
         CacheStatus currentStatus = collectStatus();
         if (currentStatus == null || currentStatus.getSize() == 0) {
             // This indicates we have no clients or those clients have no data.
             this.cacheSize.set(0);
-            return;
+            return false;
         }
 
         // Increment current generation (if needed).
@@ -216,22 +236,25 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
 
         if (!currentChanged && !oldestChanged) {
             // Nothing changed, nothing to do.
-            return;
+            return false;
         }
 
         // Notify clients that something changed (if any of the above got changed). Run in a loop, until either we can't
         // adjust the oldest anymore or we are unable to trigger any changes to the clients.
         long sizeReduction;
+        boolean reducedOverall = false;
         do {
             sizeReduction = updateClients();
             if (sizeReduction > 0) {
                 currentStatus = currentStatus.withUpdatedSize(-sizeReduction);
                 logCurrentStatus(currentStatus);
                 oldestChanged = adjustOldestGeneration(currentStatus);
+                reducedOverall = true;
             }
         } while (sizeReduction > 0 && oldestChanged);
         this.cacheSize.set(currentStatus.getSize());
         this.metrics.report(currentStatus.getSize(), currentStatus.getNewestGeneration() - currentStatus.getOldestGeneration());
+        return reducedOverall;
     }
 
     private CacheStatus collectStatus() {
@@ -360,6 +383,34 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
                 this.oldestGeneration,
                 size,
                 status.getSize() / 1048576);
+    }
+
+    private void notifyCleanupListeners() {
+        ArrayList<CacheUtilizationProvider.CleanupListener> toNotify = new ArrayList<>();
+        ArrayList<CacheUtilizationProvider.CleanupListener> toRemove = new ArrayList<>();
+        synchronized (this.cleanupListeners) {
+            for (CacheUtilizationProvider.CleanupListener l : this.cleanupListeners) {
+                if (l.isClosed()) {
+                    toRemove.add(l);
+                } else {
+                    toNotify.add(l);
+                }
+            }
+
+            this.cleanupListeners.removeAll(toRemove);
+        }
+
+        for (CacheUtilizationProvider.CleanupListener l : toNotify) {
+            try {
+                l.cacheCleanupComplete();
+            } catch (Throwable ex) {
+                if (Exceptions.mustRethrow(ex)) {
+                    throw ex;
+                }
+
+                log.error("{}: Error while notifying cleanup listener {}.", TRACE_OBJECT_ID, l, ex);
+            }
+        }
     }
 
     //endregion
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
@@ -165,7 +165,6 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
     public void registerCleanupListener(@NonNull CleanupListener listener) {
         if (listener.isClosed()) {
             log.warn("{} Attempted to register a closed Cleanup Listener ({}).", TRACE_OBJECT_ID, listener);
-            System.out.println("registered closed listener");
             return;
         }
 
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
@@ -9,6 +9,7 @@
  */
 package io.pravega.segmentstore.server.logs;
 
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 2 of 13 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
```diff
@@ -49,7 +49,7 @@
     public static final Property<Integer> CACHE_POLICY_TARGET_UTILIZATION = Property.named("cacheTargetUtilizationPercent", (int) (100 * CachePolicy.DEFAULT_TARGET_UTILIZATION));
     public static final Property<Integer> CACHE_POLICY_MAX_UTILIZATION = Property.named("cacheMaxUtilizationPercent", (int) (100 * CachePolicy.DEFAULT_MAX_UTILIZATION));
     public static final Property<Integer> CACHE_POLICY_MAX_TIME = Property.named("cacheMaxTimeSeconds", 30 * 60);
-    public static final Property<Integer> CACHE_POLICY_GENERATION_TIME = Property.named("cacheGenerationTimeSeconds", 5);
+    public static final Property<Integer> CACHE_POLICY_GENERATION_TIME = Property.named("cacheGenerationTimeSeconds", 1);
```
fpj: I don't fully understand the additional overhead that introduces when we drop from 5s to 1s. I'm also wondering whether this value isn't related to the cache size (i.e., larger cache can have a larger cache policy generation time while a smaller cache requires a shorter cache policy generation time), in which case, perhaps we could derive it from the cache size.

Could you elaborate, please?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @return A CompletableFuture that is either already completed (if no throttling is required) or will complete at
+     * some time in the future based on the throttling delay value returned by {@link ThrottlerCalculator}.
+     */
+    CompletableFuture<Void> throttle() {
+        val delay = new AtomicReference<ThrottlerCalculator.DelayResult>(this.throttlerCalculator.getThrottlingDelay());
+        if (!delay.get().isMaximum()) {
+            // We are not delaying the maximum amount. We only need to do this once.
+            val existingDelay = this.currentDelay.get();
+            if (existingDelay != null && existingDelay.source == delay.get().getThrottlerName()) {
```
sachin-j-joshi: I am not sure I understood this condition. Is it possible that delay has larger value than existing delay yet different source?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
andreipaduroiu: The overhead I'm speaking of consists of the work that the Cache Manager is doing. At each iteration it does the following steps:
1. Query all its registered clients of their state (how much data they have in the cache)
2. Determine if it needs to cleanup.
3. Instruct all registered clients to evict data progressively. This is done by incrementing the "oldest generation" that is allowed for cache entries until the clients report that enough data has been evicted to bring the utilization below the threshold.

A cache client here is anything that stores data in the cache (nothing to do with the Pravega Client):
1. Each Segment has a Cache Client in its Read Index (used for caching segment data)
2. Each Segment that uses Attributes has an additional Cache Client for its BTree nodes.
3. Each Table Segment has an additional Cache Client for storing hot keys.

The biggest and most frequently used chunk of the cache is used by 1), followed by 2) and then 3). 

The Cache Manager needs to manage all cache clients for all active Segments on the Segment store Instance (from all the containers). 

In the worst case, if we have 8 containers with max allowed segments (25K) we may have 200K segments so up to 400K clients. This is an extreme case and in this case each such segment will not have much data in the cache anyway, so the overhead will be in looping through all these clients.

The other extreme is where we have a few hot segments (such as our focused benchmark tests). We may only have a handful of clients, but the Segment Read Index and Attribute Index Caches are very active. The Segment Read Index will likely have hundreds of thousands of entries (assuming a 32KB average size within a 3GB cache); in this case querying is not expensive (we keep stats separately) but even evicting 1% of these items may include thousands of modifications to the index (and removals from the cache), all of which have to be done while holding a lock on the index. The Index is an optimized AVL tree so it's not the bottleneck, but the cache is.

My concern was that by increasing the frequency of cache cleanups, we may be spending more time maintaining these caches. However after further thinking, it is highly plausible that the increased frequency will result in smaller evictions so we are actually going to make the system better - instead of a large eviction every 5 seconds, we have much smaller ones every 1s.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
andreipaduroiu: That's a good point. The goal was not to mistakenly cancel the original throttle delay because of a different reason, but it is entirely possible that a delay of a different reason (i.e., Backlog vs the original Caching) indicate that the Caching delay reason has completely disappeared. This would have prevented us from (correctly) cancelling the delay in those cases.

I just relaxed this constraint.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
fpj: Let's talk specifically about these points:

> In the worst case, if we have 8 containers with max allowed segments (25K) we may have 200K segments so up to 400K clients.

> The other extreme is where we have a few hot segments (such as our focused benchmark tests). We may only have a handful of clients, but the Segment Read Index and Attribute Index Caches are very active.

For the first case, I understand that the time to "visit" each client is short, but I don't have a sense of how short. How long does it take to loop through 400k clients in the worst case? Is it more than 1s?

For the second case, the visit to each client is more costly, but I still don't understand how long each visit takes, roughly, I'm not necessarily asking about precise numbers.

andreipaduroiu: > How long does it take to loop through 400k clients in the worst case? Is it more than 1s?

No, we are not talking about seconds. All clients are kept in a list so we need to iterate through the list. A 400K iteration shouldn't take more than 1ms.

> For the second case, the visit to each client is more costly, but I still don't understand how long each visit takes, roughly, I'm not necessarily asking about precise numbers.

This varies by client. Let's consider each of them in turn:
- Segment Read Index. 
    - FetchStatus loops through all the generations available (the spread can be as small as 1 or as large as 2000 (one generation == one iteration of the Cache Manager; so max generation spread should be the value of cache.maxCacheTime (30 mins)). So a pretty quick one for each.    
    - UpdateGenerations (apply policy) is a bit tricker. It acquires the lock (one lock per Segment read index) and then loops through every single entry in the index (AVLTree in-order walk) and verifies if eligible for removal. If so, they are added to a removal list (we can't remove while iterating) and then they are removed from the index and from the cache. Everything is done under the lock.
        - UpdateGenerations is applied one generation at a time (the Cache Manager increments the generation, applies it, and then re-queries to see if it needs to more. So it will not hold the lock for too long. The biggest time will be spent removing items from the cache, but fortunately when we do #4044 this will be done in <50ns and there is no need to hold the lock while removing from the cache anymore.
- AttributeIndex and TableKeyCache (they are much simpler and smaller)
    - Both GetStatus and UpdateGenerations loop through all the cache entries and sum up their values or evict them, respectively. There are locks involved in both of them.

RaulGracia:
@andreipaduroiu We have tested Pravega `0.6` + PR 4166 + PR 4044 (Streaming Cache) to stress the current changes with the highest throughput we can achieve so far and a relatively small cache (3GB). It has been able to sustain a throughput of 500 MB/s for various minutes. Looking at the results, I think that the changes proposed are leading well with the issue addressed.

stream_cache

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
```diff
@@ -362,6 +384,34 @@ private void logCurrentStatus(CacheStatus status) {
...
+
+        for (CacheUtilizationProvider.CleanupListener l : toNotify) {
+            try {
+                l.cacheCleanupComplete();
+            } catch (Throwable ex) {
+                if (Exceptions.mustRethrow(ex)) {
+                    throw ex;
+                }
+
+                log.error("{}: Error while notifying cleanup listener {}.", TRACE_OBJECT_ID, l, ex);
```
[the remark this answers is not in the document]
andreipaduroiu: No it's not unrecoverable and no action is required. I put try-catch here so that if one listener decides to throw an unexpected exception, we don't prevent other listeners from being notified. If there is an error, we should log it so that someone can take a look at why it got thrown in the first place.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
```diff
@@ -55,6 +56,8 @@
     private final CachePolicy policy;
     private final AtomicBoolean closed;
     private final SegmentStoreMetrics.CacheManager metrics;
+    @GuardedBy("cleanupListeners")
+    private final HashSet<CleanupListener> cleanupListeners;
```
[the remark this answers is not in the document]
andreipaduroiu: Yes. The listener is the `Throttler` object inside OperationProcessor which manages throttling. There's one per OP, and there's one OP per Segment Container. Hence a small number of listeners per process.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
andreipaduroiu: No, they are Java `synchronized` blocks. I found that read-write locks are performing worse than `synchronized` blocks when locking for short amounts of time; they are intended for longer operations, such as when doing some sort of IO.

I don't have any monitoring for cycle times. We can add this as a metric, but I'll do it with a different issue. #4186.

PR #4173, opened 2019-09-16
Issue 4158: Potential starvation of Tier 2 writes with unthrottled clients

### LINKED ISSUE #4158: Potential starvation of Tier 2 writes with unthrottled clients
**Problem description**
Executing benchmarks in various Pravega deployments, we have found a common issue: Pravega performs much worse and exhibits higher instability when _clients write as fast as possible (unthrottled)_. Next, we describe a reproduction of this issue that can shed some light to understand the nature of the problem.

_Deployment_: The following experiment has been executed with Pravega 0.6 + RocksDB in memory (this removes one IO volume from the equation and provides better performance). The cluster deployed consists of 1 Segment Store and 3 Bookies (PKS, 4 large worker nodes, on-premise). The manifest used is the main one proposed internally, plus the new RocksDB flag to use in-memory option. 

_Benchmark_: We used a single pod to execute Pravega benchmark (v0.5). The benchmark executed for the first experiment is: `pravega-benchmark -controller tcp://xxx:9090 -stream perfeval1seg -segments 1 -producers 5 -time 600 -size 1000  -throughput 30`. The second experiment just removes the `-throughput` option, so clients are unthrottled.

_Observations_: In the following, we compare the behavior of the Pravega executing the benchmark with and without `throughput` option.
anchorsteam_vs_frost_throttling_client

- When the client is throttled to 30MBps, the system is stable and the ingest rate of Pravega (`Segment Store Write Throughput`) and the writes to Tier 2 are balanced.
- With an unthrottled client, we observe that the ingest rate at the beginning of the experiment is higher than in the throttled case (35MBps-40MBps) for a short period of time. However, at the same time, the write rate to Tier 2 drops dramatically up to 10MBps.
- The imbalance between write throughput to Tier 1 and Tier 2 makes the cache to get filled progressively. Once it reaches the limit (3GB), an aggressive throttling activity comes into play, making the throughput perceived from the client low and very variable.
- Note that the latency of writes to Bookkeeper and Tier 2 are the same in both scenarios, so this effect cannot be attributed to any of these external components. Moreover, RocksDB in memory is faster and avoids any potential interference related to the cache storage volume. 

**Problem location**
Segment Store (storage writer, cache manager?).

One possibility is that, when unthrottled clients write data to the Segment Store, Tier 1 components consume most of the resources (e.g., CPU) and make the Storage Writer to starve, thus reducing the throughput of writes to Tier 2 (i.e., throttling clients may give opportunities to the Storage Writer to get sufficient resources to keep balance). This hypothesis is supported by the above plot, as once the unthrottled writes to Tier 1 stop, there is a significant increase in the writes to Tier 2 that allow the cache to get emptied.

RaulGracia:
**Change log description**  
This PR improves RocksDB configuration, specially to prevent compaction to be a bottleneck. Moreover, we have added few metrics to better understand the behavior of the cache.

**Purpose of the change**  
Fixes #4158. 

**What the code does**  
This PR adds some settings to RocksDB configuration to improve its performance in various aspects, but specially regarding data compaction. In the following, we specify the motivation behind each of the changes in configuration:

_Improvement of reads/read amplification_:
- `setBlockCache(new LRUCache(readCacheSizeMB * 1024L * 1024L, 4))`: In our current configuration, we are missing the to instantiate an actual cache. While if we leave this unset RocksDB internally uses an 8MB cache, it is better to expose it as we can have control over it.
- `setIndexType(IndexType.kHashSearch)`: According to the documentation, the hash index is designed to improve the CPU efficiency of point lookup (we basically work with K/V).
- `optimizeForPointLookup(readCacheSizeMB * 1024L * 1024L)`: We basically use RocksDB as a KV store, so we can use this option. According to the documentation, we can use this as we don't need to keep the data sorted (i.e. we don't use an iterator, only `Put()` and `Get()` API calls).
- `setFilter(new BloomFilter())`: When doing point lookups it is encouraged to turn bloom filters on. We use bloom filters to avoid unnecessary disk reads.

_Improvement of internal parallelism & compaction throughput_:
- `setIncreaseParallelism(INTERNAL_ROCKSDB_PARALLELISM)`: The documentation states that by default, RocksDB uses only one background thread for flush and compaction. We want to increase this number to allow more internal parallelism in the DB.
- `setCompactionStyle(CompactionStyle.LEVEL)`: Specify the compaction scheme we are using.
- `optimizeLevelStyleCompaction()`: Optimize the compaction settings for large datasets.
- `setMaxBackgroundCompactions(INTERNAL_ROCKSDB_PARALLELISM)` & `setMaxBackgroundJobs(INTERNAL_ROCKSDB_PARALLELISM)`: Sets the maximum number of concurrent background compaction jobs.
- `setLevelCompactionDynamicLevelBytes(true)`: This optimization appears to be useful to limit worse case space amplification.
- `setMaxSubcompactions(INTERNAL_ROCKSDB_PARALLELISM)`: Allow multiple parallel compactions from L0 to L1.

_Miscellaneous:_
- `setWalBytesPerSync(0)` & `setWalSizeLimitMB(0)`: Be sure that all WAL settings are turned off.
- `setSkipStatsUpdateOnDbOpen(true)`: No need to update database stats, as we are not using them.

Apart from the RocksDB configuration, we have added some metrics that have been useful during the investigation of this issue to better understand the behavior of the cache:
- Size of cache operations (`segmentstore.cache.insert_size`, `segmentstore.cache.get_size`).
- Counter of cache operations (`segmentstore.cache.insert_count`, `segmentstore.cache.get_count`, `segmentstore.cache.delete_count`).
- Latency of delete operations (`segmentstore.cache.delete_latency_ms`).

**How to verify it**  
Before this change, Pravega with RocksDB on disk reported a 20MBps throughput with a lot of variability just few seconds after an unthrottled client started to write as fast as possible. With these changes, we have executed 3 unthrottled clients against 3 Segment Stores and Pravega sustained 150MBps for 1 hour. Note that more tests are still in progress.
All other tests should be passing as before.

### DIFF
--- a/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBCache.java
+++ b/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/rocksdb/RocksDBCache.java
@@ -56,6 +56,11 @@ class RocksDBCache implements Cache {
      */
     private static final int MIN_WRITE_BUFFER_NUMBER_TO_MERGE = 2;
 
+    /**
+     * Number of internal parallel threads that RocksDB will use for a specific task.
+     */
+    private static final int INTERNAL_ROCKSDB_PARALLELISM = 8;
+
     @Getter
     private final String id;
     private final Options databaseOptions;
@@ -255,6 +260,8 @@ class RocksDBCache implements Cache {
                 .setCreateIfMissing(true)
                 .setDbLogDir(Paths.get(this.dbDir, DB_LOG_DIR).toString())
                 .setWalTtlSeconds(0)
+                .setWalBytesPerSync(0)
+                .setWalSizeLimitMB(0)
                 .setWriteBufferSize(writeBufferSizeMB * 1024L * 1024L)
                 .setMaxWriteBufferNumber(MAX_WRITE_BUFFER_NUMBER)
                 .setMinWriteBufferNumberToMerge(MIN_WRITE_BUFFER_NUMBER_TO_MERGE)
@@ -262,15 +269,12 @@ class RocksDBCache implements Cache {
                 .setOptimizeFiltersForHits(true)
                 .setUseDirectReads(this.directReads)
                 .setSkipStatsUpdateOnDbOpen(true)
-                .setWalBytesPerSync(0)
-                .setWalSizeLimitMB(0)
-                .setIncreaseParallelism(4)
                 .optimizeForPointLookup(readCacheSizeMB * 1024L * 1024L)
+                .setIncreaseParallelism(INTERNAL_ROCKSDB_PARALLELISM)
+                .setMaxBackgroundJobs(INTERNAL_ROCKSDB_PARALLELISM)
                 .setCompactionStyle(CompactionStyle.LEVEL)
-                .setMaxBackgroundCompactions(8)
-                .setMaxBackgroundJobs(8)
-                .setCompactionReadaheadSize(1024L * 1024L)
                 .optimizeLevelStyleCompaction()
+                .setMaxBackgroundCompactions(INTERNAL_ROCKSDB_PARALLELISM)
                 .setLevelCompactionDynamicLevelBytes(true);
 
         if (this.memoryOnly) {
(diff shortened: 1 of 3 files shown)


timjbutler:
Performance testing showed expected throughput for unthrottled client. So the PR is good from performance perspective.

RaulGracia:
@andreipaduroiu @fpj it seems that @timjbutler has verified that this change improves the problem they were having with RocksDB on disk. Could you please have a look to this PR if you want any further changes?

PR #4179, opened 2019-09-18
Issue 1546: Remove thread from ControllerResolverFactory

### LINKED ISSUE #1546: ControllerResolverFactory should re-use external executor service
**Problem description**
The ControllerResolverFactory used by our Controller gRPC client, creates a dedicated singlethreaded executor for internal use. We should be using an external supplied executor to optimize the application's thread pool requirement.
Also we are making blocking calls to controller and to DNS server in this thread. If we are going to use an external executor we need to make all calls non-blocking to avoid holding onto the external thread for too long. Need to use async version of all apis to achieve this.

**Problem location**
ControllerResolverFactory 

Using existing executor service instead of creating a new one for dedicated use to optimize the thread resources.

tkaitchuck:
**Change log description**  
This removes the dedicated thread from ControllerResolverFactory.

**Purpose of the change**  
Fixes #1546

**What the code does**  
Moves the thread onto the standard thread pool

**How to verify it**  
This should be covered by existing tests.


client/src/main/java/io/pravega/client/stream/impl/ControllerResolverFactory.java
```diff
@@ -172,32 +171,38 @@ public void start(Listener listener) {
...
+                    } else {
+                        scheduleDiscovery = true;
+                    }
+                }
+            } else {
+                scheduleDiscovery = true;
+            }
+            if (scheduleDiscovery) {
+                // Schedule the first discovery immediately.
+                this.scheduledFuture = this.scheduledExecutor.schedule(this::getControllers, 0L, TimeUnit.SECONDS);
```
shrids: - If the `bootstrap servers` do not contain an IP address then it is scheduled for discovery/resolution. The `getControllers()` method uses `java.net.InetSocketAddress` to resolve it. Now the default value of system property `networkaddress.cache.ttl` is -1. i.e. all addresses once resolved will be cached forever. This would defeat the goal of getControllers() being invoked periodically.
https://docs.oracle.com/javase/8/docs/technotes/guides/net/properties.html#nct

- If the dns resolution takes more time when we are using the `io.pravega.client.netty.impl.ConnectionFactory#getInternalExecutor()` to run the resolver task it would affect the other segment readers and writers right ?

tkaitchuck: 1) Yes it would... I am not sure what to do about that though. Because this is a library we can't go changing system properties which have global effects.
2) Yes.

shrids: Point 2 will be problematic to debug.
As a next step let me trigger a round of system tests to verify it.

shrids:
Triggered a branch build against this change ref: `pravega-branch-builder/2008/`

shrids:
System test run with this changeset was successful `job/test-pravega-continuous/2506/`

PR #4181, opened 2019-09-19
Issue 3344: Watermarking

### LINKED ISSUE #3344: Implement watermarks

tkaitchuck:
**Change log description**  
* Add support for watermarking as described in PDP-33

**Purpose of the change**  
Fixes #3344. 

**What the code does**  
See https://github.com/pravega/pravega/wiki/PDP-33:-Watermarking for full details. In short:

A new api `noteTime` is added on the writer side, and another `getTimeWindow` is added on the reader. If the writers of a stream provide timestamps by calling `noteTime` these are combined with the writer's position information and sent to the controller. The controller takes the maximum position across the writers and records this information as a stream cut along with the range of times supplied by the writers. This is written to a special `mark` stream which is a shadow stream for the main stream with a single segment written using a RevisionedStreamClient. Within the reader the EventStreamReader creates a background process where this information is read and compared against the positions of the local reader and the others in the reader group. It combines these positions to create a stream cut which is compared with the stream cuts written to the `mark` stream. This allows the reader to calculate where the group is with respect to the window of times provided by the writers at the time the events were written. This is then made available to the reading application via the `getTimeWindow` api.

**How to verify it**  
New unit, system, and integration tests have been added. These need to consistently pass.


segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessorTest.java
```diff
@@ -300,7 +301,7 @@ public void testCreateSegment() throws Exception {
         // so createSegmentStats may record 1 or 2 createSegment operation here.
     }
 
-    @Test(timeout = 20000)
+    @Test//(timeout = 20000)
```
fpj: Remove comment.

client/src/main/java/io/pravega/client/segment/impl/SegmentAttribute.java
```diff
@@ -30,4 +30,6 @@
         HashHelper hash = HashHelper.seededWith("SegmentAttribute");
         value = hash.toUUID(this.name());
     }
+
```
fpj: These two new lines are unnecessary and without them, this file would not be changed at all. Please revert these.

client/src/main/java/io/pravega/client/segment/impl/SegmentInputStreamImpl.java
```diff
@@ -33,7 +33,9 @@
 @Slf4j
 @ToString
 class SegmentInputStreamImpl implements SegmentInputStream {
+    static final int MIN_BUFFER_SIZE = 1024;
```
fpj: Could you explain why these new constants are needed?

client/src/main/java/io/pravega/client/stream/TimeWindow.java
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright (c) 2018 Dell Inc., or its subsidiaries. All Rights Reserved.
```
fpj: Could you remove the year from the copyright line to comply with the accepted header across all new files?

```
Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.

Licensed under the Apache License, Version 2.0 (the "License");
you may not use this file except in compliance with the License.
You may obtain a copy of the License at

    http://www.apache.org/licenses/LICENSE-2.0
```

https://github.com/pravega/pravega/blob/master/HEADER

client/src/test/java/io/pravega/client/netty/impl/FlowHandlerTest.java
```diff
@@ -280,8 +280,8 @@ public void testProcessWithErrorReplyProcessor() throws Exception {
         doAnswer((Answer<Void>) invocation -> {
             throw new RuntimeException("ReplyProcessorError");
         }).when(processor).process(any(Reply.class));
-
-        WireCommands.DataAppended msg = new WireCommands.DataAppended(flow.asLong(), UUID.randomUUID(), 2, 1);
+
```
fpj: Please remove unnecessary leading spaces.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -186,21 +187,25 @@ private Checkpoint completeCheckpoint(String checkpointName) {
     public void resetReadersToCheckpoint(Checkpoint checkpoint) {
         synchronizer.updateState((state, updates) -> {
             ReaderGroupConfig config = state.getConfig();
-            Map<Segment, Long> positions = new HashMap<>();
+            Map<SegmentWithRange, Long> positions = new HashMap<>();
             for (StreamCut cut : checkpoint.asImpl().getPositions().values()) {
-                positions.putAll(cut.asImpl().getPositions());
+                for (Entry<Segment, Long> e : cut.asImpl().getPositions().entrySet()) {
+                    //TODO watermarking: deal with empty range information.
```
shrids: we should create a tracking issue for this.

client/src/main/java/io/pravega/client/segment/impl/SegmentInputStreamImpl.java
tkaitchuck: These are just sanity bounds, because we have made the buffer size configurable.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
tkaitchuck: https://github.com/pravega/pravega/issues/4194

PR #4199, opened 2019-09-23
Issue 4195: Use StreamConfiguration.timestampAggregationTimeout in timing out writers

### LINKED ISSUE #4195: StreamConfiguration.timestampAggregationTimeout is not honoured by controller
**Problem description**
Controller currently does not make use of StreamConfiguration.timestampAggregationTimeout to timeout writers. Timeout writers that are inactive and remove them from watermarking computation. 

**Problem location**
Watermarking workflow on controller.

Fix the issue

shiveshr:

**Change log description**  
Enforce streamConfiguration policy for writer timestamp timeout. 

**Purpose of the change**  
Fixes #4195 

**What the code does**  
1. While storing stream configuration in metadata store, serialize and deserialize the TimestampAggregationTimeout.
2. Use this timeout in PeriodicWatermarking workflow. 
The workflow now starts tracking a writer when it found inactive. It adds an entry into a map for writerId to the wall clock time at which it is added. 
In subsequent cycles, if the writer continues to be inactive and if `current wall clock time` - `writer's recorded time` > config.timestampAggregationTimeout then a writer is declared timedout and becomes a candidate for removal. 

**How to verify it**  
unit tests added


controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
```diff
@@ -368,6 +371,14 @@ private SegmentWithRange transform(StreamSegmentRecord segment) {
         private AtomicInteger windowStart;
         private final int windowSize;
 
+        /**
+         * Map to track inactive writers for their timeouts. This map records wall clock time when
+         * this writer was found to be inactive. If in subsequent cycles
+         * it continues to be inactive and the time elapsed is greater than
+         * {@link StreamConfiguration#timestampAggregationTimeout}, then it is declared timedout.
+         */
+        private ConcurrentHashMap<String, Long> inactiveWriters;
```
andreipaduroiu: `final`

PR #4208, opened 2019-09-25
Issue 4084: (SegmentStore) Max Rolling Segment Size

### LINKED ISSUE #4084: Rolling storage feature for data segments
**Problem description**

Rolling storage feature that enables us to avoid issues with underlying file size limitation (e.g 4 TB on isilon). Currently, we do not allow rolling the segment file for data segments. We do it only for attribute segments, which is where we store extended attributes for a segment.

 

We need to support rolling the segment file. See attached email.
**Problem location**
Rolling Storage

andreipaduroiu:
**Change log description**  
- Added the ability to specify the absolute maximum size of any segment chunk in Tier 2 via a setting (`writer.maxRolloverSizeBytes`).

**Purpose of the change**  
Fixes #4084.

**What the code does**  
- Added a new setting (`writer.maxRolloverSizeBytes`) which defines the absolute maximum size of any segment chunk in Tier 2.
- When creating segments in Tier 2, the rolling size passed to `RollingStorage` will the the minimum between this setting and whatever is configured in the Segment's `Attributes.ROLLOVER_SIZE` attribute (which is set by the Controller when it creates the segment). 

**How to verify it**  
New unit test added.


config/config.properties
```diff
@@ -578,4 +578,11 @@ extendeds3.url=localhost:9020
 # Valid values: Positive integer.
 #writer.shutdownTimeoutMillis=10000
 
+# The maximum size of a single Segment Chunk in Storage. While this can be configured on a per-segment basis via the
+# Segment Store API, this setting establishes a ceiling to that value. No Segment Chunk may exceed this length regardless
+# of how it was configured via the API (if at all). Use this setting if you need to meet some hardware limitation with
+# the underlying Storage implementation (i.e., maximum file size).
+# Valid values: Positive number.
+#writer.maxRolloverSizeBytes=9223372036854775807
```
sachin-j-joshi: The size in KB/MB might be more useful and less error prone.

PR #4210, opened 2019-09-26
Issue 4209: Bump versions of dependencies

### LINKED ISSUE #4209: Bump versions of dependencies
**Problem description**
For the 0.6 release we should upgrade versions of our dependencies.

**Problem location**
Gradle.properties

Update all versions to the latest ones that continue to have all tests pass.

tkaitchuck:
**Change log description**  
* Update versions of dependencies.

**Purpose of the change**  
Fixes #4209

**What the code does**  
Increases the versions of all of our current dependencies without breaking any code/tests.

**How to verify it**  
We should run all tests against this branch.


build.gradle
```diff
@@ -564,6 +568,12 @@ project('shared:controller-api') {
 }
 
 project('controller') {
+    
+    //TODO: Work around deprication warning in generated code caused by https://github.com/FasterXML/jackson-databind/issues/1786
```
ravisharda: `deprecation`

ravisharda: The content referenced by the link does not seem to explain why this is needed. I guess we don't want the build to fail on warnings that indicate deprecated code, here. Rewording the comment should help.

tkaitchuck: I have changed the code to suppress only the one file.

controller/src/main/java/io/pravega/controller/server/rest/generated/api/RFC3339DateFormat.java
```diff
@@ -1,17 +1,15 @@
...
 
-public class RFC3339DateFormat extends ISO8601DateFormat {
+@SuppressWarnings("deprecation") //See https://github.com/FasterXML/jackson-databind/issues/1786
+public class RFC3339DateFormat extends com.fasterxml.jackson.databind.util.ISO8601DateFormat {
 
     // Same as ISO8601DateFormat but serializing milliseconds.
     @Override
     public StringBuffer format(Date date, StringBuffer toAppendTo, FieldPosition fieldPosition) {
-        String value = ISO8601Utils.format(date, true);
+        String value = com.fasterxml.jackson.databind.util.ISO8601Utils.format(date, true);
```
shiveshr: why are we updating generated file?

tkaitchuck: Otherwise the build fails with a compilation error.
There is no functional change here, this just suppresses the warnings from importing a deprecated class. Which would normally be a warning, but our build is set to treat warnings as errors and fail the build. Originally I just changed that behaviour but based on feedback, it seemed better to remove that safety net for the sake of preserving one generated file in it's unmodified state.

Once there is an update to the code generator to use non-deprecated code, we should be able to re-generate the file and leave it unmodified.

fpj: Given the constraints in front of us, I'd say that we modify the file and create a separate issue explaining the problem and pointing to the jackson issue. The new issue should also say that once the jackson issue is resolved, we need to upgrade and regenerate the file.

shiveshr: @tkaitchuck can you please open an issue for the same.

tkaitchuck: Created https://github.com/pravega/pravega/issues/4224

tkaitchuck:
Merging based on conversation with @shiveshr .

tkaitchuck:
Coverage is at 0% because changes are in the gradle file, and the service starter which is only used by integration tests. (Which don't count towards the coverage metric)

PR #4212, opened 2019-09-26
Issue 4211: Fix backward compatibility bug allowing unread fields

### LINKED ISSUE #4211: Fix compatibility issue caused by additional field.
**Problem description**
If there is an additional field in a WireCommand the protocol is supposed to ignore it to preserve compatibility. Currently the code ignores the wirecommand, but as the data is un-read netty interprets it as the next command and attempts to parse that, which generally fails killing the connection.

**Problem location**
CommandDecoder

Advance the number of bytes read to reflect the length of the data included in the command even if it was not all used.

tkaitchuck:

**Change log description**  
* In CommandDecoder: advance readerIndex to reflect data read.

**Purpose of the change**  
Fixes #4211

**What the code does**  
Ensures that is there is an extra field in a wire command from the server to the client that the client does not attempt to interpret these extra bytes as a separate command.


RaulGracia:
@tkaitchuck is this passing the compatibility tests build?

shrids:
The compatibility test passed successfully with this change ref: `job/test-pravega-compatibility/29/`

PR #4226, opened 2019-09-28
Issue 4225: Fix Controller.WatermarkClient's to consider only previous watermark for tracking active writers

### LINKED ISSUE #4225: Bugs: Controller.WatermarkClient's sliding window logic incorrectly assumes RevisionStreamClient.readFromRevision to read the event at the `revision` position which can prevent watermark from progressing
**Problem description**
PeriodicWatermarking.WatermarkClient maintains the runtime state for per stream watermark computation where it keeps list of known writers and previously emitted watermark. It tracks writers and identifies them as active participating or inactive writers and is responsible for reading from and writing to `mark` stream. 
This class implements a complex sliding and dynamically adjusting window of "active watermarks" and writers that were behind start window of "active watermarks" were considered inactive and removed. 

There is a bug that is present in current sliding window computation: 
RevisionStreamClient.readFrom(revision) reads "after" the revision. And `event, revision` tuples that are returned have revision which identifies end offset for the event. However, watermark client assumed it include the `event` if it called readFrom(`revision`). This can cause duplicate watermarks to be emitted. 

This can cause multiple issues ->
1. A stale writer may stall progress of watermarks because it could be considered active if it has reported time ahead of older watermark in the sliding window. And it is not timedout because it is considered active. 
2. In a corner case, two successive watermarks could have same lowerbound on time while advancing the position. This could happen because we read after previously checkpointed revision and if the window has been shrunk to after previous watermark, then new watermark could be computed with same or lower time than previous watermark.  

**Problem location**
PeriodicWatermarking.WatermarkClient

We had the complex sliding window logic of identifying writers for omission because Watermark workflow did not consider streamConfiguration.timestampAggregationTimeout. This was fixed as part of #4195 but it regressed and caused `1`. 

Since we are now using StreamConfiguration.timestampAggregationTimeout we dont have a need for any complexity in tracking active writers. 
We can simplify the logic for activeWriters by comparing it with only `previousWatermark.lowerboundOnTIme`. 
And if a writer has not been shutdown or have timed out then its active. And its tracked for timeout only if its behind the previous watermark. 
This means a writer can be identified as non participating in one cycle and removed after `timestampAggregationTimeout`.
 
In each periodic cycle, inactive writers are pruned, while a watermark is computed only if all active writers are also participating. 
By simply deciding on active writer based on last watermark, we can simplify the existing complex logic greatly and easily argue correctness of the code.

shiveshr:
**Change log description**  
1. Simplify Controller.WatermarkClient that is used for tracking active writers, previous watermarks and writing marks into mark stream.
2. Fix issue in code where for `(event, revision)` tuple returned by RevisionedStreamClient, watermarkclinet incorrectly expects to find `event` if it reads from `revision`. 

**Purpose of the change**  
Fixes #4225 

**What the code does**  
1. Active writer tracking changes:
PeriodicWatermarking.WatermarkClient is the class which is responsible for maintaining the state and tracking inactive writers and timing them out while also reading from and writing to `mark` stream. 
Since we had not used the StreamConfiguration.timestampAggregationTimeout earlier, so this class had a complex logic of sliding window of previous watermarks to determine which writers were inactive and time them out. 
However we fixed writer's timeout management with #4195 to use the timeout specified in StreamConfiguration. So tracking lifecycle of writers can be greatly simplified to rely only on most recent watermark. 
So a writer can be considered to be in one of three possible states - inactive, active and participating.
Inactive writers are those that have not reported a time greater than previousWatermark.lowerboundOnTIme. And they have either been shutdown or have timed out. 
Active writers are same as inactive except that they are still alive and have not been timedout. 
Participating writers are those which have their `time` greater than previous watermark.lowerBoundOnTime. 

In each periodic cycle, inactive writers are pruned, while a watermark is computed only if all active writers are also participating. If there are non participating active writers, they are added for tracking for timeout. 

By simply deciding on active writer based on last watermark, we can simplify the existing complex logic greatly and easily argue correctness of the code. 

2. RevisionStreamClient.readFrom(revision) reads "after" the revision. And `event, revision` tuples that are returned have revision which identifies end offset for the event. However, watermark client assumed it include the `event` if it called readFrom(`revision`). This can cause duplicate watermarks to be emitted. 

**How to verify it**  
Unit tests updated

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
+++ b/controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
@@ -42,7 +42,6 @@ import lombok.Synchronized;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.ParametersAreNonnullByDefault;
-import javax.annotation.concurrent.GuardedBy;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -57,7 +56,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -226,7 +224,7 @@ public class PeriodicWatermarking {
                                  .collect(Collectors.toMap(y -> getSegmentWithRange(scope, streamName, context, y.getKey()),
                                          Entry::getValue)));
                     }).collect(Collectors.toList()));
-            
+            log.debug("Emitting watermark for stream {}/{} with time {}", scope, streamName, lowerBoundOnTime);
             return positionsFuture.thenAccept(listOfPositions -> listOfPositions.forEach(position -> {
                 // add writer positions to upperBound map. 
                 addToUpperBound(position, upperBound);
@@ -362,14 +360,10 @@ public class PeriodicWatermarking {
     }
 
     static class WatermarkClient {
-        private static final int WINDOW_SIZE = 2;
         private final RevisionedStreamClient<Watermark> client;
         
-        @GuardedBy("$Lock")
-        private List<Map.Entry<Revision, Watermark>> entries;
+        private AtomicReference<Map.Entry<Revision, Watermark>> previousWatermark = new AtomicReference<>();
         private final AtomicReference<Revision> markRevision = new AtomicReference<>();
-        private AtomicInteger windowStart;
-        private final int windowSize;
 
         /**
          * Map to track inactive writers for their timeouts. This map records wall clock time when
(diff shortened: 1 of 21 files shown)


controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
```diff
@@ -362,14 +360,10 @@ private SegmentWithRange transform(StreamSegmentRecord segment) {
     }
 
     static class WatermarkClient {
-        private static final int WINDOW_SIZE = 2;
         private final RevisionedStreamClient<Watermark> client;
         
-        @GuardedBy("$Lock")
-        private List<Map.Entry<Revision, Watermark>> entries;
+        private AtomicReference<Map.Entry<Revision, Watermark>> previousWatermark = new AtomicReference<>();
```
tkaitchuck: final

controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
```diff
@@ -388,139 +382,68 @@ private SegmentWithRange transform(StreamSegmentRecord segment) {
...
-         * If distance between active window watermark and latest window watermark is less than 2, 
-         * active window is not progressed if a new watermark is emitted. 
          */
         @Synchronized
         void reinitialize() {
             Revision revision = client.getMark();
             // revision can be null if no window has been set yet. 
             if (revision == null) {
                 markRevision.set(client.fetchOldestRevision());
+                client.compareAndSetMark(null, markRevision.get());
```
tkaitchuck: You may be interest in `NewestReference` in common.

PR #4082, opened 2019-08-08
Issue 4056: Reduce creation of new CompletionStage while trying to fill read buffer.

shrids:
(Longevity tests are still in progress)

**Change log description**  
Reduce the creation of new `java.util.concurrent.CompletionStage` inside `SegmentInputStreamImpl#fillBuffer()`

**Purpose of the change**  
Fixes #4056 

**What the code does**  
If a reader attempts to read from a Segment which has no data then a new `java.util.concurrent.CompletionStage` is created for every execution of `SegmentInputStream#fillBuffer`.  These new `CompletionStage`s will never be garbage collected and can lead to increased usage of heap memory in a long running application.

**How to verify it**  
All the existing tests should continue to pass. 
Also, we should verify that the memory consumption by the readers is reduced in the case of segments with no new data. e.g: this scenario can arise when events written to an auto scale enabled stream with a small set of keys.

PR #4227, opened 2019-09-28
Issue 3395: Mitigate high read latency on streams with idle segments

### LINKED ISSUE #2609: High read latency when there is no backlog and idle segments
**Problem description**
When reading a stream that has numerous segments, and a subset of those segments are idle (has no data), the `EventStreamReaderImpl` spends considerable time blocking on the idle segment(s), leading to poor overall read performance.

Below is a trace that demonstrates the issue.  This example involves a stream with two fixed segments, and works by creating a feedback loop involving a single routing key.   The program is writing to the same stream that it is reading from, therefore we expect to read back a given event almost immediately after writing it.  By using a single routing key, one of the two segments is always idle.

The poor performance stems from the fact that once the orderer selects a given segment, the reader blocks on that segment for up to one second (`TIME_UNIT`), even if data becomes available sooner on another segment.  In the example, the orderer examines all segments, realizes that none are readable without blocking, and then makes a round-robin selection.   Just afterward, data becomes available on the segment that wasn't selected, but the reader has already committed to a 1-second nap.

```
16:22:36,668 TRACE io.pravega.client.segment.impl.SegmentOutputStreamImpl        - Sending append request: Append(segment=examples/ourboros-bf24bb45-b31e-4879-afa6-fcbe4d892aea/0, writerId=51f7dbfc-f1a0-402f-bc52-6c9e436134e3, eventNumber=11, eventCount=1, data=UnpooledHeapByteBuf(ridx: 0, widx: 19, cap: 19/19), expectedLength=null)
16:22:36,669 TRACE io.pravega.client.stream.impl.EventStreamReaderImpl           - ENTER readNextEvent@1326077567568736 [].

  16:22:36,669 TRACE io.pravega.client.stream.impl.Orderer                         - ENTER nextSegment@1326077567732493 [].
    16:22:36,669 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - canReadWithoutBlocking false
    16:22:36,669 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@992bc0f, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@5a669c29, headerReadingBuffer=java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], offset=0, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@2890521c[Not completed])
    16:22:36,669 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - canReadWithoutBlocking false
    16:22:36,669 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@4b131692, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@3a0901b8, headerReadingBuffer=java.nio.HeapByteBuffer[pos=8 lim=8 cap=8], offset=288, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@498fb16d[Not completed])
  16:22:36,669 TRACE io.pravega.client.stream.impl.Orderer                         - LEAVE nextSegment@1326077567732493 (elapsed=166us).

  16:22:36,669 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - ENTER read@1326077567917282 [].
    16:22:36,669 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Read called at offset 0
    16:22:36,669 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@992bc0f, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@5a669c29, headerReadingBuffer=java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], offset=0, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@2890521c[Not completed])
    16:22:36,670 TRACE io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl    - Received read result WireCommands.SegmentRead(type=SEGMENT_READ, segment=examples/ourboros-bf24bb45-b31e-4879-afa6-fcbe4d892aea/0, offset=288, atTail=false, endOfSegment=false, data=java.nio.HeapByteBuffer[pos=0 lim=27 cap=27])
  16:22:37,675 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - LEAVE read@1326077567917282 (elapsed=1005457us).

  16:22:37,677 TRACE io.pravega.client.stream.impl.Orderer                         - ENTER nextSegment@1326078575627035 [].
    16:22:37,677 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - canReadWithoutBlocking true
  16:22:37,677 TRACE io.pravega.client.stream.impl.Orderer                         - LEAVE nextSegment@1326078575627035 (elapsed=92us).

  16:22:37,677 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - ENTER read@1326078575737608 [].
    16:22:37,677 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Read called at offset 288
    16:22:37,677 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@4b131692, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@3a0901b8, headerReadingBuffer=java.nio.HeapByteBuffer[pos=8 lim=8 cap=8], offset=288, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@498fb16d[Completed normally])
  16:22:37,677 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - LEAVE read@1326078575737608 (elapsed=212us).

  16:22:37,677 TRACE io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl    - Sending read request WireCommands.ReadSegment(type=READ_SEGMENT, segment=examples/ourboros-bf24bb45-b31e-4879-afa6-fcbe4d892aea/0, offset=315, suggestedLength=65536, delegationToken=)

16:22:37,678 TRACE io.pravega.client.stream.impl.EventStreamReaderImpl           - LEAVE readNextEvent@1326077567568736 (elapsed=1009001us).

...

16:22:37,679 TRACE io.pravega.client.stream.impl.EventStreamReaderImpl           - ENTER readNextEvent@1326078577418736 [].
16:22:37,679 TRACE io.pravega.client.stream.impl.Orderer                         - ENTER nextSegment@1326078577605328 [].
16:22:37,679 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - canReadWithoutBlocking false
16:22:37,679 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@992bc0f, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@5a669c29, headerReadingBuffer=java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], offset=0, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@2890521c[Not completed])
16:22:37,679 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - canReadWithoutBlocking false
16:22:37,679 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@4b131692, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@3a0901b8, headerReadingBuffer=java.nio.HeapByteBuffer[pos=8 lim=8 cap=8], offset=315, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@1783163f[Not completed])
16:22:37,679 TRACE io.pravega.client.stream.impl.Orderer                         - LEAVE nextSegment@1326078577605328 (elapsed=167us).
16:22:37,679 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - ENTER read@1326078577788314 [].
16:22:37,679 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Read called at offset 0
16:22:37,679 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@992bc0f, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@5a669c29, headerReadingBuffer=java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], offset=0, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@2890521c[Not completed])
16:22:37,680 TRACE io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl    - Received read result WireCommands.SegmentRead(type=SEGMENT_READ, segment=examples/ourboros-bf24bb45-b31e-4879-afa6-fcbe4d892aea/0, offset=315, atTail=false, endOfSegment=false, data=java.nio.HeapByteBuffer[pos=0 lim=27 cap=27])
16:22:38,681 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - LEAVE read@1326078577788314 (elapsed=1001515us).
16:22:38,681 TRACE io.pravega.client.stream.impl.Orderer                         - ENTER nextSegment@1326079579545161 [].
16:22:38,681 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - canReadWithoutBlocking true
16:22:38,681 TRACE io.pravega.client.stream.impl.Orderer                         - LEAVE nextSegment@1326079579545161 (elapsed=63us).
16:22:38,681 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - ENTER read@1326079579643715 [].
16:22:38,681 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Read called at offset 315
16:22:38,681 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - Filling buffer SegmentInputStreamImpl(asyncInput=io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl@4b131692, readLength=65536, buffer=io.pravega.common.util.CircularBuffer@3a0901b8, headerReadingBuffer=java.nio.HeapByteBuffer[pos=8 lim=8 cap=8], offset=315, endOffset=9223372036854775807, receivedEndOfSegment=false, receivedTruncated=false, outstandingRequest=java.util.concurrent.CompletableFuture@1783163f[Completed normally])
16:22:38,681 TRACE io.pravega.client.segment.impl.SegmentInputStreamImpl         - LEAVE read@1326079579643715 (elapsed=150us).
16:22:38,681 TRACE io.pravega.client.segment.impl.AsyncSegmentInputStreamImpl    - Sending read request WireCommands.ReadSegment(type=READ_SEGMENT, segment=examples/ourboros-bf24bb45-b31e-4879-afa6-fcbe4d892aea/0, offset=342, suggestedLength=65536, delegationToken=)
16:22:38,682 TRACE io.pravega.client.stream.impl.EventStreamReaderImpl           - LEAVE readNextEvent@1326078577418736 (elapsed=1002903us).
```

For reference: [EventStreamReaderImpl.java](https://github.com/pravega/pravega/blob/6e338b4a20072e834b51f62351716d7769a6725a/client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java#L83)

**Problem location**

- Rework `EventStreamReaderImpl` to use async segment reads, using `CompletableFuture::anyOf` to wait for the next event from any assigned segment.
- Factor out the framing logic within `SegmentInputStreamImpl` to a new class, e.g. `AsyncFrameDecoderImpl`, that would decode in an async way to support the above.
- Reconcile the semantic irregularity between `SegmentInputStream` and `AsyncSegmentInputStream`; the latter produces raw segment data, the former deals with event framing.

### LINKED ISSUE #3395: Lower read latency on periodically idle stream
We can improve read latency at the expense of order by changing Orderer to be pathologically unfair and use the futures from the `fillBuffer` call to select which segment to read from. Because this negativity impacts fairness it should not be done before PDP-33 is complete. 

This could be implemented as follows: 
1. Change the `CompletableFuture<Void> fillBuffer()` call on SegmentInputStreamImpl to return `CompletableFuture<Segment>` where it returns it's own segment. (This can be made into a constant completed future to avoid overhead)
2. In EventStreamReaderImpl.readNextEventInternal move the sleep following the call to nextSegment into that call. So Orderer.nextSegment will need to be modified to take a timeout.
3. Inside of Orderer.nextSegment in the else block for isSegmentReady collect the returned futures into an array called `readyFutures`.
4. At the end in the unhappy case instead of `segments.get(MathHelpers.abs(counter.incrementAndGet()) % segments.size())` it can return `Futures.anyOf(readyFutures).get(timeout)`

RaulGracia:
**Change log description**  
Set a shorter reader timeout on idle segments.

**Purpose of the change**  
Partially fixes #3395.

**What the code does**  
As described in #3395, in Streams with idle segments, clients observe very high read latency. The reason is that, in the current implementation of the client, if there is a situation where segments have no new data (i.e., idle), readers are instructed to wait for a given period of time on a segment until data is available. Currently, this waiting time is set to _1 second_.

Naturally, while a reader is blocked for 1 second, new data may arrive to another segment and it will remain un-read until the 1 second blocking period completes for that reader. In this PR, we just set a lower max blocking period of readers on idle segments to mitigate the high read latency observed. This new blocking time is based on a base time (i.e., 5ms, similar to the server write latency) and the number of readers in the reader group. The reason is that, setting a low blocking interval and considering many readers on the same machine, the CPU consumption could be excessive. We have also verified that the lower blocking period does not have a significant impact on CPU usage for a growing number of readers.

**How to verify it**  
In the following, we provide the end-to-end latency (1 writer/1 reader) of the execution of Pravega Benchmark with different reader blocking periods: 1 second (current), 10ms, and 5ms. The objective is to observe the improvements in read latency, as well as any potential increase in CPU usage from using shorter blocking periods (higher polling frequency):
wait_time_comparison

Observations:
- A reader blocking period of 5ms can improve latency by 3x-36x (depending on the percentile), compared to the current 1s blocking period.
- As visible in the CPU usage plots, the benchmark tool exhibits a similar CPU usage level in the 3 cases (20%-40%), meaning that the improvement in latency does not induce a high increase of compute resources consumption.

No code logic changes, all the tests should work as before.


client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -110,7 +113,8 @@
     }
     
     private EventRead<Type> readNextEventInternal(long timeout) throws ReaderNotInReaderGroupException, TruncatedDataException {
-        long waitTime = Math.min(timeout, ReaderGroupStateManager.TIME_UNIT.toMillis());
+        int numReaders = Math.max(1, groupState.getNumberOfReaders());
```
shrids: `groupState.getNumberOfReaders()` cannot return less 1.

RaulGracia: Thanks, this was a kind of defensive setting. If your statement is correct, then maybe I can just use the result of `groupState.getNumberOfReaders()`.

tkaitchuck: I don't really understand why the number of readers should effect this calculation.

RaulGracia:
@tkaitchuck thanks, will rework the PR to attempt the change you are proposing.

client/src/main/java/io/pravega/client/stream/impl/Orderer.java
```diff
@@ -48,18 +53,39 @@
...
         }
+        List<CompletableFuture<T>> idleSegments = new ArrayList<>();
         for (int i = 0; i < segments.size(); i++) {
             T inputStream = segments.get(MathHelpers.abs(counter.incrementAndGet()) % segments.size());
             if (inputStream.isSegmentReady()) {
                 log.trace("Selecting segment: " + inputStream.getSegmentId());
                 return inputStream;
             } else {
-                inputStream.fillBuffer();
+                idleSegments.add(inputStream.fillBuffer().thenApply(b -> inputStream));
```
tkaitchuck: This would effectively undo: https://github.com/pravega/pravega/pull/4082
That proved to be very significant optimization, and the change prevented the conditions needed to create #3900. 
I don't think we should do this.

tkaitchuck: Perhaps a way around this, would be to add the stream to the idle segment list and then use the result of the read call to get the string of the segment and compare that to a map that you construct after the loop using the segmentid from the reader. 

That's not a great solution either though because it requires changing interfaces and makes the highest level code dependent on the wire protocol, but it also doesn't truly solve the problem in the event that the whole stream is idle. In that case we'll end up adding one completion stage per invocation.

tkaitchuck: Maybe if we had something like a 'fillBufferAndSetListener' call which filled the buffer and set a listener to notify when it was completed instead of using a future. The trouble with this is I'm not sure how to make the set of the listener atomic with respect to the completion of the future. It's also not a great interface and it might cause problems with the threading model.

RaulGracia:
@tkaitchuck @shrids given that the async implementation may induce problems with long chains or read futures, I will revert the PR to its original form.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -110,7 +113,8 @@
     }
     
     private EventRead<Type> readNextEventInternal(long timeout) throws ReaderNotInReaderGroupException, TruncatedDataException {
-        long waitTime = Math.min(timeout, ReaderGroupStateManager.TIME_UNIT.toMillis());
+        long waitTime = Math.min(timeout, Math.min(ReaderGroupStateManager.TIME_UNIT.toMillis(),
+                BASE_READER_WAITING_TIME_MS + groupState.getNumberOfReaders()));
```
shrids: Thank you @RaulGracia .
Observing that we have now changed the logic for wait time from 
`READER_WAITING_TIME_MS * numReaders` to `BASE_READER_WAITING_TIME_MS  + numReader` i.e. the wait time increases by a 1ms per new reader added.
I have a couple of question regarding this
- How does the CPU utilization and latency numbers vary between using `BASE_READER_WAITING_TIME_MS  + numReader` v/s a constant value ?
- The perf numbers in the PR description are based on the previous scheme right?

RaulGracia: Thanks @shrids, yes, you are right, let me reply to your comments:
- I have executed Pravega Benchmark in my laptop (not in a k8 pod) with 1 producer (writing 1 e/s) and 50 consumers (50-segment stream). With the current approach, the CPU utilization of the process (that includes the writer and the read) is generally below 2-3%. Note that in this idleness situation is where the timeout defined is exercised the most.
- Based on your question, I tried the constant value. For the same experiment as executed before, I see that the CPU consumption for 50 readers case is similar to the previous case. The only differential factor is that during the creation of the reader group it takes high CPU consumption 90%-95% for a longer time compared to the previous case (in which this also happens, but for shorter period of time).
- Yes, the numbers are based on the previous scheme. Current numbers are the same for 1 reader and better as the number of reader increase (as the timeout grows less than before based on the number of readers).

Based on that, and given that in real cases multiple readers will be distributed across multiple machines, I will simplify the PR and leave a constant value. This will improve performance for low event rates until we find a safe async way to do the same.

PR #4242, opened 2019-10-03
Issue 4241: Enable Client metrics

### LINKED ISSUE #4241: Enable Client Metrics
**Problem description**
Enable Client Metrics. The user should be able to specify a `MetricListener` implementation as part of ClientConfig. 

**Problem location**
Client

Ensure the client does not use EventLoop threads to update the client metrics.

shrids:
**Change log description**  
* Enable Client metrics. This PR enables the user to pass an implementation of `MetricListener` as part of `io.pravega.client.ClientConfig`

**Purpose of the change**  
Fixes #4241 

**What the code does**  
This PR enables Client metrics. It enables a new configuration entry in `io.pravega.client.ClientConfig` where the user can pass an implementation of `io.pravega.shared.metrics.MetricListener`. 
At present metrics of the `AppendBLockSize` and amount of time taken to write and flush an Event on the netty Channel have been enabled.

**How to verify it**  


client/src/main/java/io/pravega/client/ClientConfig.java
```diff
@@ -101,6 +103,21 @@
...
+     * A optional listener which can be used to get performance metrics from the client. The user
+     * can implement this interface to obtain performance metrics of the client.
+     */
+    private final MetricListener metricListener;
+
+    /**
+     * A method which returns the optional metric listener {@link MetricListener} .
+     * @return MetricListener.
+     */
+    public Optional<MetricListener> getMetricListener() {
```
andreipaduroiu: What's the difference between a plain getter that can return `null` and returning an `Optional` which you can check with null? IMO, the plain getter is simpler to implement and to use, especially for nullable values. Plus you don't have to carry an `Optional` everywhere you pass this (constructors, etc.)

common/src/main/java/io/pravega/common/concurrent/MultiKeyLatestItemSequentialProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+    public void updateItem(KeyType key, ItemType newItem) {
+        Preconditions.checkNotNull(newItem);
+        Preconditions.checkNotNull(key);
+        if (toProcessKVP.put(key, newItem) == null) {
+            executor.execute(() -> {
+                ItemType item = newItem;
+                processFunction.accept(key, item);
+                while (!toProcessKVP.remove(key, item)) {
+                    item = toProcessKVP.get(key);
+                    processFunction.accept(key, item);
```
andreipaduroiu: You need to account for the fact that `processFunction` may throw. If it does, you will exit from this loop and never process the last item, thus leaving the processing in limbo.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricNames.java
```diff
@@ -0,0 +1,48 @@
...
+    public static String metricKey(String metric, String... tags) {
+
+        if (tags == null || tags.length == 0) {  //if no tags supplied, the original metric name is used for both cache key and registry key.
+            return metric;
+        } else { //if tag is supplied, append tag value to form cache key; original metric name is registry key.
+            StringBuilder sb = new StringBuilder(metric);
+            Preconditions.checkArgument((tags.length % 2) == 0, "Tags is a set of key/value pair so the size must be even: %s", tags.length);
+            for (int i = 0; i < tags.length; i += 2) {
+                Preconditions.checkArgument(!Strings.isNullOrEmpty(tags[i]) || !Strings.isNullOrEmpty(tags[i + 1]), "Tag name or value cannot be empty or null");
+                sb.append('.').append(tags[i + 1]);
```
andreipaduroiu: Just to confirm: you are intentionally not appending `tags[i]` here, correct?

common/src/main/java/io/pravega/common/concurrent/MultiKeyLatestItemSequentialProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Provides a way to run a function on Key Value Pairs but guarantees that the function is only invoked for one item at a time per key.
+ * If multiple 'updates' to a particular key  are provided while the processor is running, only the most recent is passed to the
+ * function.
+ * This allows it to 'skip' updates for keys that are not the most recent value.
+ *
+ * @param <KeyType> The type of the key to be processed.
+ * @param <ItemType> The type of item to be processed.
+ */
+public class MultiKeyLatestItemSequentialProcessor<KeyType, ItemType> {
```
andreipaduroiu: I strongly recommend you make this class implement `AutoCloseable` and add a `close` method that would cancel any ongoing executions and prevent new items from being added. Helps a lot with graceful shutdowns.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+import io.pravega.common.concurrent.ExecutorServiceHelpers;
+import io.pravega.common.concurrent.MultiKeyLatestItemSequentialProcessor;
+import java.util.concurrent.ScheduledExecutorService;
+
+
+public class ClientMetricUpdater implements MetricNotifier {
+    private final MultiKeyLatestItemSequentialProcessor<String, Long> successProcessor;
+    private final MultiKeyLatestItemSequentialProcessor<String, Long> failureProcessor;
+
+    private final ScheduledExecutorService clientMetricExecutor = ExecutorServiceHelpers.newScheduledThreadPool(2, "clientMetrics");
```
andreipaduroiu: Would one thread be sufficient?

shared/protocol/src/main/java/io/pravega/shared/metrics/MetricNotifier.java
```diff
@@ -0,0 +1,34 @@
...
+        @Override
+        public void updateFailureMetric(String metricName, long value) {
+        }
+
+        @Override
+        public void close() {
+        }
+    };
+
+    void updateSuccessMetric(String metricName, long value);
```
andreipaduroiu: Could we have this interface extend from `MetricListener` so we don't define this seemingly similar methods again?

common/src/main/java/io/pravega/common/concurrent/MultiKeyLatestItemSequentialProcessor.java
andreipaduroiu: You will also never know it threw as there's nobody listening on this thread. Let's add a try-catch and log the error in the catch block.

client/src/main/java/io/pravega/client/netty/impl/ClientConnectionImpl.java
```diff
@@ -50,9 +55,12 @@ public void send(WireCommand cmd) throws ConnectionFailedException {
 
     @Override
     public void send(Append append) throws ConnectionFailedException {
+        Timer timer = new Timer();
         checkClientConnectionClosed();
         nettyHandler.setRecentMessage();
         Futures.getAndHandleExceptions(nettyHandler.getChannel().writeAndFlush(append), ConnectionFailedException::new);
+        nettyHandler.getUpdateMetric().updateSuccessMetric(metricKey(ClientMetricNames.CLIENT_APPEND_LATENCY, segmentTags(append.getSegment())),
```
tkaitchuck: Perhaps we can pass the tags into the function so that any string concatenation happens inside ClientMetricUpdater in the background thread.

common/src/test/java/io/pravega/common/concurrent/MultiKeyLatestItemSequentialProcessorTest.java
```diff
@@ -0,0 +1,88 @@
...
+        processor.updateItem("k", "b");
+        processor.updateItem("k", "c");
+        startedLatch.await();
+        latch.release();
+        pool.shutdown();
+        pool.awaitTermination(5, TimeUnit.SECONDS);
+        assertEquals(ImmutableList.of("a", "c"), processed);
+    }
+
+    // TODO: write test with multiple keys
```
tkaitchuck: Please add

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricTags.java
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * @param qualifiedSegmentName fully qualified segment name.
+     * @return string array as segment tag of metric.
+     */
+    public static String[] segmentTags(String qualifiedSegmentName) {
+        Preconditions.checkNotNull(qualifiedSegmentName);
+        String[] tags = {TAG_SCOPE, null, TAG_STREAM, null, TAG_SEGMENT, null, TAG_EPOCH, null};
+
+        String segmentBaseName = getSegmentBaseName(qualifiedSegmentName);
+        String[] tokens = segmentBaseName.split("[/]");
```
tkaitchuck: This should probably rely on the same split code as in Stream or Segment so that it's guaranteed to be consistent. We also created NameUtils and SegmentNameUtils. Maybe something could be refactored into one of those.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricTags.java
```diff
@@ -0,0 +1,74 @@
...
+        return tags;
+    }
+
+    /**
+     * Get base name of segment with the potential transaction delimiter removed.
+     *
+     * @param segmentQualifiedName fully qualified segment name.
+     * @return the base name of segment.
+     */
+    private static String getSegmentBaseName(String segmentQualifiedName) {
```
tkaitchuck: See above

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricUpdater.java
tkaitchuck: Yes, let's use just one. That way an implementation of a users code does not have to deal with multiple threads.

common/src/main/java/io/pravega/common/concurrent/MultiKeyLatestItemSequentialProcessor.java
shrids: At present, the lifecycle of the executor is managed by the `io.pravega.shared.metrics.ClientMetricUpdater` which instantiates an instance of `MultiKeyLatestItemSequentialProcessor`. 
Also, `io.pravega.shared.metrics.MetricNotifier` implements `AutoCloseable` .

shared/protocol/src/main/java/io/pravega/shared/metrics/MetricNotifier.java
shrids: refactored the api based on review comments.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricTags.java
shrids: done.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricNames.java
shrids: yes.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricKeys.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.common.base.Strings;
+import lombok.Getter;
+
+/**
+ * The various metrics that can be generated by the client.
+ */
+public enum ClientMetricKeys {
+
+    CLIENT_APPEND_LATENCY("client.segment.append_latency_ms"),
+    CLIENT_APPEND_BLOCK_SIZE("client.segment.append_block_size");
```
RaulGracia: To have a complete understanding of the performance of the client, we are missing one metric: the `append_response_latency_ms`. That is, `append_latency_ms` gives use the time to write the event to the server, but as I posted in some plots, this differs from the total latency an application sees waiting for the write ack to complete. I think we can easily add this metric in `FlowHandler.channelRead` when handling `WireCommands.DataAppended` events, right?

kevinhan88:
Lots of code duplication due to invisible metrics dependency - we need to adjust the dependency relationship, either before or after this PR.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricKeys.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @VisibleForTesting
+    @Getter
+    private final String metricKey;
+
+    ClientMetricKeys(String metricKey) {
+        StringBuilder sb = new StringBuilder("pravega");
+        sb.append( ".");
```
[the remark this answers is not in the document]
shrids: fixed.

shared/protocol/src/main/java/io/pravega/shared/metrics/ClientMetricKeys.java
shrids: Thanks @RaulGracia . Looked into enabling metric `append_response_latency_ms`, time taken to receive an ack for a given event. Do not see a way to enable it without an additional penalty on the client-side. We could track this in a different issue.
Based on your comment did the following changes
- Ensured the segment writer id is part of the metric, since there can be multiple writers to the same segment.
- Added a new metric `CLIENT_OUTSTANDING_APPEND_COUNT`. This metric represents the number of appends for which we have not yet received acks from the segment store.

PR #4243, opened 2019-10-03
Issue 4247:  Remove lombok and netty-all from client's dependency list.

### LINKED ISSUE #4247: Remove lombok as a runtime dependency for client.
**Problem description**
`:client:dependencies` tasks indicates that lombok is a runtime dependency.

**Problem location**
build.gradle

Remove lombok runtime dependency for the client and prune its dependency list.

shrids:
**Change log description**  
- Ensure lombok is not part of the client's dependency list.
- Remove redundant netty-all dependency from `:shared:controller-api`

**Purpose of the change**  
Reduces dependencies. Fixes #4247.

**What the code does**  
- Removing `lombok` and `netty-all` reduces the dependencies of the client by 5+MB.
- Ensure `javadoc` gradle task runs on the delombok'ed code.
- Update gradle lombok plugin to `3.2.0`

**How to verify it**  
Ensure java docs of lombok annotated code are generated and ensure `:client: dependencies` does not list lombok and netty-all as part of its `runtimeOnly` dependencies.


gradle/java.gradle
```diff
@@ -37,6 +37,7 @@ plugins.withId('java') {
     }
     artifacts { archives sourceJar }
 
+
```
andreipaduroiu: Let's revert this change.

gradle.properties
```diff
@@ -26,7 +26,7 @@ spotbugsAnnotationsVersion=3.1.12
 spotbugsPluginVersion=1.6.9
 jcipAnnotationsVersion=1.0
 gradleDockerPlugin=3.1.0
-gradleLombokPluginVersion=1.14
+gradleLombokPluginVersion=3.2.0
```
andreipaduroiu: This is quite a big version increment. Are we sure we won't introduce unwanted problems by this?

build.gradle
```diff
@@ -123,14 +136,9 @@ project('common') {
 
     javadoc {
         title = "Pravega Common Libraries"
+        dependsOn delombok
```
andreipaduroiu: Isn't there a way in gradle to apply these to all projects? Can we use the `allprojects` section at the top of this file?

gradle.properties
shrids: Reviewed the release notes and do not observe any issues. Ran the javadocs gradle task and do not observe issues.

build.gradle
shrids: Tried a couple of approaches, do not see a way to ensure tasktype `javadoc` to depend on delombok by default.

PR #4248, opened 2019-10-04
Issue 4206: Add support for multi ledgers and journals

### LINKED ISSUE #4206: Pass multiple journal&ledger directories to Bookkeeper
**Problem description**
The current script does not support passing multiple journal&ledger directories to the Bookkeeper server. We need to modify it to support such requests.

**Problem location**
Bookkeeper docker script

Tristan1900:
Signed-off-by: wenqimou <452787782@qq.com>

**Change log description**  
This PR will enable multi directories feature for ledgers and journals. By specifying the name of the directoies in the Pravega manifest like below, the entryscript of Bookkeeper will be able to create related directories and give them appropriate ownership.
```
spec:
  bookkeeper:
    options:
      journalDirectories: "/bk/journal/j0,/bk/journal/j1,/bk/journal/j2,/bk/journal/j3"
      ledgerDirectories: "/bk/ledgers/l0,/bk/ledgers/l1,/bk/ledgers/l2,/bk/ledgers/l3"
```

**Purpose of the change**  
Fix #4206 

**What the code does**  
Parse the comma-separated directory names and create related directories and grant give right ownership to them.

**How to verify it**  
Deployment should work


docker/bookkeeper/entrypoint.sh
```diff
@@ -20,11 +32,14 @@ BK_CLUSTER_NAME=${BK_CLUSTER_NAME:-"bookkeeper"}
...
 
+create_dir "${BK_journalDirectories}"
+create_dir "${BK_ledgerDirectories}"
+
 export BOOKIE_PORT=${BOOKIE_PORT}
 export BK_zkServers=${BK_zkServers}
 export BK_metadataServiceUri=zk://${ZK_URL}${BK_LEDGERS_PATH}
-export BK_journalDirectory=${BK_DIR}/journal
-export BK_ledgerDirectories=${BK_DIR}/ledgers
+export BK_journalDirectories=${BK_journalDirectories:-${BK_DIR}/journal}
```
RaulGracia: We have found the issue with this change in which, if `journalDirectories` is not specified in the Pravega manifest, the deprecated `journalDirectory` (and `ledgerDirectory`) parameter is taken as the path to store Bookkeeper data. Note that default value for these deprecated parameters is currently `/data/xxx`, instead of `/bk/xxx`. The consequence of this is that the Bookie can end up storing data in the local storage of the pod, instead of using the volumes defined for that purpose. We may need to override the default values for `journalDirectory` and `ledgerDirectory` to avoid this problem if someone forgets to specify `journalDirectories`/`ledgerDirectories`.

Tristan1900: Thank Raul, that's a good point. However, I think here it will provide the default value `${BK_DIR}/journal`,which is `/bk/journal`, to the Bookkeeper if `BK_journalDirectories` is not specified. The reason that we are getting `/data/bk` is because `/data/bk` is the default value in official Bookkeeper scirpt and I didn't provide our default value in our experiment image(sorry about that!).

PR #4262, opened 2019-10-10
Issue 4263: Ensure Epoll transport is used by Pravega client.

### LINKED ISSUE #4263: Ensure epoll native transport is used by the pravega client on linux x86_64
**Problem description**
Observing that NIO is being used by Pravega client on linux boxes even though Pravega client has a runtime dependency on `netty-transport-native-epoll`.

**Problem location**
Build.gradle

Ensure the correct classifier is used. ref: https://netty.io/wiki/native-transports.html

shrids:
**Change log description**  
Ensure ePoll native transport is used by Pravega client on linux_x86_64.  

**Purpose of the change**  
Fixes #4263

**What the code does**  
The Pravega client already has a dependency on `netty-transport-native-epoll`, which provides platform-specific transports i.e the jar already contains the shared library files. Also, the client dynamically switches back to NIO if EPOLL transport is not available. 
The PR ensures the `classifier ` is set `x86_64` to ensure epoll is used on linux based x86_64 systems.

**How to verify it**  
All the tests should continue to pass. Inspect the pravega-client logs to ensure ePoll is infact used.
e.g: log snippets
```
2019-10-10 01:50:21,812 2273 [epollEventLoopGroup-4-1] INFO  i.p.s.p.netty.FailingReplyProcessor - Received hello: WireCommands.Hello(type=HELLO, highVersion=8, lowVersion=5)

```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -236,7 +236,9 @@ project('client') {
         compile project(':shared:authplugin')
         compile project(':shared:protocol')
         compile project(":shared:controller-api")
-        compile group: 'io.netty', name: 'netty-transport-native-epoll', version: nettyVersion
+        // Ensure epoll native transport is used for linux-x86_64 architecture systems.
+        // The client fallsback to NIO based transport if epoll is not available.
+        compile group: 'io.netty', name: 'netty-transport-native-epoll', version: nettyVersion, classifier: 'linux-x86_64'
         compile group: 'com.google.guava', name: 'guava', version: guavaVersion
         testCompile project(':test:testcommon')
         testCompile group: 'org.slf4j', name: 'log4j-over-slf4j', version: slf4jApiVersion


shrids:
A system test run has been triggered /job/test-pravega-continuous-1/382/ with these changes.

andreipaduroiu:
@shrids I'm a bit confused. You are forcing a compile-time dependency, but epoll/nio is chosen at runtime. How does this work?

shrids:
Netty provides JNI transports only for Linux (since 4.0.16) and MacOS/BSD (since 4.1.11)
ref: https://netty.io/wiki/native-transports.html
Now the document indicates that we should use the `os-maven-plugin` to detect the build and ensure the respective epoll library is used i.e. https://repo1.maven.org/maven2/io/netty/netty-transport-native-epoll/4.1.30.Final/netty-transport-native-epoll-4.1.30.Final-linux-x86_64.jar
But this implies the build system's version is used and not the target systems version !

Here we have hardcoded the dependency to `linux-x86_64` 

Now this netty transport epoll is part of the runtime classpath and netty checks if itis able to get it working ref: 
https://github.com/netty/netty/blob/3a9ac829d53e1da7365db2d8f544d7d112147321/transport-native-epoll/src/main/java/io/netty/channel/epoll/Epoll.java#L37-L69

andreipaduroiu:
@shrids That makes sense. Thanks for the clarification.

PR #4266, opened 2019-10-10
Issue 4246: Delete Partially created streams

### LINKED ISSUE #4246: Controller does not allow deletion of partially created streams because it expects stream to be sealed
**Problem description**
Controller does not allow streams to be deleted unless they are sealed. 
This means any stream whose creation has failed and is in partially created state with State being set to CREATING. It cannot be deleted. 
This prevents scope to be deleted as well because of presence of partially created stream which cannot be deleted. 

**Problem location**
StreamMetadataTasks

Allow deletion of stream for non-active states as well. (for states = creating or unknown).

shiveshr:

**Change log description**  
- Only sealed streams were allowed to be deleted. Now Delete stream even if the stream is in CREATING state. 

**Purpose of the change**  
Fixes #4246 

**What the code does**  
Delete Stream workflow expected the stream to be sealed. Now we allow deletion of a stream which is only partially created. 
Since a stream's creation could have failed at different stages, the delete stream workflow, which assumed presence of certain records may not find them. 

Following are the changes done to handle this:
1. StreamMetadataTasks checks for the stream state before deleting it. 
If its sealed, then nothing changes. 
If its creating, then also we submit an event to delete the stream in background. 
If a stream's state is not set, then it only has some metadata records that are deleted directly by calling store.deletestream. 

2. DeleteStreamTask is modified to process the event if the stream is either CREATING or SEALED. It also does not assume presence of segments metadata to be present in the stream's metadata record. 

3. AbstractStreamMetadataStore stores last known segment number for enabling recreation of a stream. This basically assumed that active segment record was created. This assumption is fixed and DataNotFound is ignored as we proceed to delete stream metadata records. 

4. PravegaTableStreams also expected epoch table to have been created which may not be the case. 

**How to verify it**  
Unit tests added.


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteStreamTask.java
```diff
@@ -93,14 +95,14 @@ public DeleteStreamTask(final StreamMetadataTasks streamMetadataTasks,
 
     private CompletableFuture<Void> notifyAndDelete(OperationContext context, String scope, String stream, long requestId) {
         log.info(requestId, "{}/{} deleting segments", scope, stream);
-        return streamMetadataStore.getAllSegmentIds(scope, stream, context, executor)
-                .thenComposeAsync(allSegments -> {
-                    return streamMetadataTasks.notifyDeleteSegments(scope, stream, allSegments, streamMetadataTasks.retrieveDelegationToken(), requestId)
+        return Futures.exceptionallyExpecting(streamMetadataStore.getAllSegmentIds(scope, stream, context, executor)
+                .thenComposeAsync(allSegments -> 
+                    streamMetadataTasks.notifyDeleteSegments(scope, stream, allSegments, streamMetadataTasks.retrieveDelegationToken(), requestId)),
+                            e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException, null)
```
shrids: We are already handling StoreException.DataNoFoundException in line 80. Hence this check is not required right ?

shiveshr: No, this is required because we want to delete stream metadata even if getAllSegments threw DataNotFoundException. 

If trying to delete a stream throws DNF then that is treated as an idempotent case which is handled on line 80.

shrids: thanks @shiveshr

PR #4269, opened 2019-10-10
Issue 4267: (SegmentStore) Made StorageWriter Tier 1 Truncation Async

### LINKED ISSUE #4267: StorageWriter can deadlock with Throttler
**Problem description**
There is a possibility that the Storage Writer may be blocked from making progress because its call to truncate Tier 1 is throttled because of excessive cache, which cannot be resolved because no data is eligible for cache eviction.

Consider this scenario:
1. A client continuously appends data (to one or more segments).
2. The StorageWriter processes it and writes it to Tier 2 (normally, but slower than Tier 1).
3. The Cache fills up and its utilization exceeds the eviction and throttling threshold, as such all new operations are subject to throttling. This throttling is interrupted only when more data is evicted from the cache (data can be evicted once it's moved to Tier 2).
4. The StorageWriter needs to acknowledge/truncate Tier 1, but this requires queuing a small operation into Tier 1.
5. Due to Tier 1 being throttled, the ack cannot be processed (it's delayed because of the throttle), and the Storage Writer cannot write more data to Tier 2 (as it's blocked).

So we enter a situation where:
1. An almost full cache cannot be evicted until the storage writer moves more data to Tier 2.
2. The Storage Writer  cannot move data to Tier 2 until the ack/truncation from the previous iteration completes.
3. The ack/truncation cannot complete because new operations are throttled/delayed
4. New operations are throttled because the cache is full.

1 depends on 2; 2 depends on 3; 3 depends on 4 and 4 depends on 1: deadlock!

**Problem location**
StorageWriter.

Make Tier 1 ack/truncation an async task. There is no need for it to be synchronous.

andreipaduroiu:
**Change log description**  
- The StorageWriter no longer waits for Tier 1 ack/truncation to complete before moving on.

**Purpose of the change**  
Fixes #4267. 

**What the code does**  
See #4267 for the problem this is trying to fix.
- By making Tier 1 ack/truncation async, the StorageWriter is no longer subject to any Tier 1 throttling delays and is free to continue moving data to Tier 2, which will make more cache entries eligible for eviction, and thus unblock throttling.
- However the Tier 1 ack must be executed sequentially (we don't want two or more acks to overlap) since they check and update local state. Having the possibility of concurrent execution would potentially cause unwanted side effects or false errors (such as "already truncated at this point).

**How to verify it**  
All unit tests and system tests must pass. No new tests are needed since functionality hasn't changed and existing tests do verify that ack-ing/truncation is performed.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
@@ -18,6 +18,7 @@ import io.pravega.common.ObjectClosedException;
 import io.pravega.common.Timer;
 import io.pravega.common.concurrent.AbstractThreadPoolService;
 import io.pravega.common.concurrent.Futures;
+import io.pravega.common.concurrent.SequentialProcessor;
 import io.pravega.segmentstore.server.DataCorruptionException;
 import io.pravega.segmentstore.server.SegmentOperation;
 import io.pravega.segmentstore.server.UpdateableSegmentMetadata;
@@ -60,6 +61,7 @@ class StorageWriter extends AbstractThreadPoolService implements Writer {
     private final Timer timer;
     private final AckCalculator ackCalculator;
     private final WriterFactory.CreateProcessors createProcessors;
+    private final SequentialProcessor ackProcessor;
 
     //endregion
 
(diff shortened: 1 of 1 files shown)


andreipaduroiu:
Reopened in #4271

PR #4293, opened 2019-10-17
Issue 4092: Change default store in InProcPravegaCluster to pravega tables

### LINKED ISSUE #4092: Pravega Standalone uses Zookeeper as Stream Metadata Store.
**Problem description**
Pravega Standalone uses Zookeeper as stream metadata store. 
Since standalone, upon restart should have a clean slate, so we should not use any durable store where any stream specific information should be present. 
This is because the segment store uses in memory for standalone. 
So use in memory or pravega tables for storing metadata for streams.

**Problem location**
InProcPravegaCluster.java

Use PravegaTables instead as that is the default store.

shiveshr:
**Change log description**  
Change default store in InProcPravegaCluster to pravega tables from Zookeeper

**Purpose of the change**  
Fixes #4092

**What the code does**  
Changes default store in InProcPravegaCluster to pravega tables.

**How to verify it**  
All tests should pass.


client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -951,6 +951,8 @@ private TxnSegments convert(CreateTxnResponse response) {
                                                       .setTxnId(ModelHelper.decode(txId));
             if (timestamp != null) {
                 txnRequest.setTimestamp(timestamp);
+            } else {
+                txnRequest.setTimestamp(Long.MIN_VALUE);
```
andreipaduroiu: How is this related to this PR?

shiveshr: Commit from other PR got here as i accidentally built this branch on top of that one. 
Fixed it now.

PR #4301, opened 2019-10-21
Issue 4265: Ensure backward compatability for ClientFactoryImpl#createTransactionalEventWriter api

### LINKED ISSUE #4265: createTransactionalEventWriter() method under EventStreamClientFactory interface has incompatibility issue
**Problem description**

The **createTransactionalEventWriter()** under **EventStreamClientFactory<>** does NOT support **Transaction event writer creation** when writer id is not passed for its call, and hence is creating **incompatibility issues** with traditional transaction calls ( for example, as used by pravega-puncher tool ).

**Problem location**

Pravega version: `0.6.0-2391.97363b3`
File location: `io.pravega.client.EventStreamClientFactory.createTransactionalEventWriter()`

The **createTransactionalEventWriter()** method can be **overloaded** to support transactional event writer creation even when **writer id** is not passed in its method call.

shrids:
**Change log description**  
Ensure backward compatability for ClientFactoryImpl#createTransactionalEventWriter api

**Purpose of the change**  
Fixes #4265 

**What the code does**  
Ensure backward compatability for ClientFactoryImpl#createTransactionalEventWriter api when writerId is not passed.

**How to verify it**  


client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -145,23 +145,27 @@ public ClientFactoryImpl(String scope, Controller controller, ConnectionFactory
...
                                                                                 EventWriterConfig config) {
-        log.info("Creating transactional writer for stream: {} with configuration: {}", streamName, config);
+        log.info("Creating transactional writer:{} for stream: {} with configuration: {}", writerId, streamName, config);
         Stream stream = new StreamImpl(scope, streamName);
         return new TransactionalEventStreamWriterImpl<T>(stream, writerId, controller, outFactory, s, config, connectionFactory.getInternalExecutor());
     }
 
+    @Override
+    public <T> TransactionalEventStreamWriter<T> createTransactionalEventWriter(String streamName, Serializer<T> s, EventWriterConfig config) {
+        return createTransactionalEventWriter(UUID.randomUUID().toString(), streamName, s, config);
```
tkaitchuck: My only concern here is that `StreamConfiguration.timestampAggregationTimeout` needs to be set to a reasonable value here, otherwise an application unaware might endup holding up watermarks.

I suppose given that the default is sane, it's ok to let users configure something badly. But I think we may want to make a note in the Javadoc, perhaps recommending they set `automaticallyNoteTime(true)` on the `EventWriterConfig`

shrids: Updated the java doc to indicate that writer id is autogenerated.
(io.pravega.client.stream.Transaction#commit(long) will be used to note the time and the default value of `timestampAggregationTimeout` is 0)

PR #4302, opened 2019-10-22
Issue 4260: No-Op tier-2 Adapter

### LINKED ISSUE #4260: NoOp Tier 2 Adapter for Testing Purposes
**Problem description**
We want to be able to run Pravega Benchmark without interference from Tier 2 sync. As such, for testing purposes only, we want to have a setting that turns off all Tier 2 sync.

**Problem location**
Segment Store

- Create a configuration setting (default=false; add a warning that it's for test only) that enables no-op tier2
- Implement a `NoOpStorage` class that wraps an existing `Storage` interface which
    - Does nothing when requested to modify a segment. 
        - Ideally we'd have another setting that defines how long should such a no-op operation last. This would be useful for testing different scenarios where we vary the transfer rate to Tier 2. This can be easily done using a `Futures.delayedFuture` call.
    - Throws `UnsupportedOperationException` for any reads
    - Others as needed
- Subclass `StreamSegmentStoreServiceTests` to do an end-to-end test with this setting turned on. We might need to add some additional flags to this test to not expect any data from Tier 2.
- To enable successful container failovers
    - The NoOpStorage should not ignore operations any `_system` segments (anything that is prefixed with `_system`). Those are all metadata tables and we **must** preserve them between container failovers. As such, for these segments, we must pass-through the request to the underlying storage adapter.
    - The NoOpStorage may need access to the Segment Container Metadata so that it returns appropriate values for `exists` and `getSegmentInfo`.
    - Other things may be needed. The E2E test should reveal most of them.

kevinhan88:

**Change log description**  
. Created a thin No-Op (of payload) layer on top of SyncStorage, which doesn't change any existing functionality but skips the tier 2 payload writing (system segments still go through)
. Introduced two new config properties for the No-Op layer
. StorageLoader now detects the new config and load the new No-Op factory if necessary
. Created a new unit test to prove that existing functionalities still work, and the new No-Op layer works as expected.

**Purpose of the change**  
Closes #4260

**What the code does**  
. Added NoOpStorage and NoOpStorageFactory - this is a layer on top of SyncStorage, which ignores non-system writing to tier-2
. Added default method createSyncStorage() into interface StorageFactory; this is to expose the syncStorage for No-Op factory to encapsulate.
. Changes HDFSStorageFactory, FileSystemStorageFactory, ExtendedS3StorageFactory and InMemoryStorageFactory to adopt the above interface change.
. Added storageextra.storageNoOpMode into config.properties with default value false. If this switch is set to true, NoOp mode will be turned on.
. Added storageextra.storageWriteNoOpLatencyMilli into config.properties with default value 20. When No-Op mode is turned on, this latency is applied to make the no-op operation real as possible.
. StorageExtraConfig created to accommodate the above two new config properties.
. StorageLoader is revised, when it loads storage implementation, it also checks storageextra.storageNoOpMode; if it's true, then the loader wraps the original implementation factory with the NoOp factory.
. StreamSegmentServiceNoOpStoreTests is added to test the No-Op layer. In order to verify no-operation, internally the test creates an internal in-memory storage to accommodate the segment which should be ignored. With this mechanism, the end-to-end test inside StreamSegmentStoreTestBase should pass as usually, without noticing any difference.

**How to verify it**  
All existing tests, particularly those extend StreamSegmentStoreTestBase, should work without any change. 
The new unit test, StreamSegmentServiceNoOpStoreTests should pass.


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
```diff
@@ -0,0 +1,265 @@
...
+            }
+        }
+    }
+
+    @Override
+    public int read(SegmentHandle handle, long offset, byte[] buffer, int bufferOffset, int length) throws StreamSegmentException {
+        if (isSystemSegment(handle.getSegmentName())) {
+            return baseStorage.read(handle, offset, buffer, bufferOffset, length);
+        } else {
+            if (noopInMemoryStorage != null) {
```
tkaitchuck: This if statement is in every single method in this class. Perhaps you should just make two classes that have the same interface.

config/config.properties
```diff
@@ -91,6 +91,22 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
...
 
+# Storage NO-OP Mode: segment payload write is no-oped; payload reading is not supported
+# This mode is used to avoid tier 2 interference in testing while keep the system functioning as usual.
+#
+# NOTE: pravegaservice.storageImplementation is still used to store metadata.
+# WARNING: Do NOT set NO-OP Mode to true in production environment!!! Data loss will happen!!!
+#
+# Valid values: true, false
+# Default value: false
+#storageextra.storageNoOpMode=false
```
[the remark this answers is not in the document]
andreipaduroiu: Because the No-Op implementation cannot replace the Tier 2 adapter. It needs a Tier 2 adapter as pass-through so that `_system` segments can function normally (i.e., they should not be affected). If we also ignore all operations on `_system` segments then the Controller may not be able to retrieve its metadata and the cluster would become unusable after the first Segment Container failover & recovery. Passing-through the `_system` segments has no impact on measuring performance (they hold a combined total of a few MBs at most) but are required for the good functioning of Pravega.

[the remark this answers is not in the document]
kevinhan88: Put this property into a separate config component also improves the isolation - in case Storage is not in No-Op mode, which is in most cases, then the entire NoOpStorage, including relevant config properties, are completely off the scene.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/StorageExtraConfig.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Extra Configuration for Storage Component.
+ */
+@Slf4j
+public class StorageExtraConfig {
+
+    public static final Property<Boolean> STORAGE_NO_OP_MODE = Property.named("storageNoOpMode", false);
+    public static final Property<Integer> STORAGE_WRITE_NO_OP_LATENCY = Property.named("storageWriteNoOpLatencyMilli", 20);
+    private static final String COMPONENT_CODE = "storageextra";
+
```
andreipaduroiu: Too many consecutive empty lines.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
```diff
@@ -47,6 +47,8 @@
     protected static final String APPEND_FORMAT = "Segment_%s_Append_%d";
     private static final int SEGMENT_COUNT = 4;
 
+    protected boolean isTestingSystemSegment = false;
```
andreipaduroiu: It is bad practice to create protected non-final fields. Please make this private and provide whatever access you need to it via protected methods.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
```diff
@@ -497,6 +499,14 @@ protected void createSegment(String name, Storage s) {
         s.create(name, null).join();
     }
 
+    protected String createSegmentName(String originName) {
+        if (isTestingSystemSegment) {
+            return "_" + originName;
```
andreipaduroiu: I'm pretty sure system segments are prefixed with `_system`. Are there other system segments that are not prefixed with that?

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/noop/NoOpStorageSystemDataWriteReadTests.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    protected Storage createStorage() {
+        return new AsyncStorageWrapper(new NoOpStorage(config, systemStorage, null), executorService());
+    }
+
+    @Override
+    public void testFencing() {
+    }
```
andreipaduroiu: Did you purposefully leave this empty? If so, you should write "This method intentionally left blank", otherwise someone might think this is incomplete.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
```diff
@@ -0,0 +1,267 @@
...
+import io.pravega.segmentstore.contracts.SegmentProperties;
+import io.pravega.segmentstore.contracts.StreamSegmentException;
+import io.pravega.segmentstore.storage.SegmentHandle;
+import io.pravega.segmentstore.storage.SyncStorage;
+import java.io.InputStream;
+import java.util.concurrent.TimeUnit;
+
+import lombok.extern.slf4j.Slf4j;
+
+import static io.pravega.shared.NameUtils.INTERNAL_NAME_PREFIX;
```
andreipaduroiu: We should use `INTERNAL_SCOPE_NAME`.

While you're at it:
- Also use this in your unit test (I left a comment way below about it)
- Please modify `StreamSegmentNameUtils.java#64` to reference `INTERNAL_SCOPENAME` as a prefix instead of `_system`.

andreipaduroiu: While looking at this class, I realized that there may be more metadata tables out there that need to be included in the pass-through. If you look at `StreamSegmentNameUtils.getQualifiedTableName`, which is used by the Controller to generate names of its metadata tables, it stores this in a non-system segment which contains `_tables` (StreamSegmentNameUtils.TABLES) as the second token (position 1). I think we should modify `NoOpStorage` to allow this as pass-through as well (just check it if contains this substring. If we don't, we risk the Controller not being able to read its own Scope/Stream metadata which could quickly lead to weird behavior.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorageFactory.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Factory for No-Op mode Storage adapters.
+ */
+public class NoOpStorageFactory implements StorageFactory {
+    private final StorageExtraConfig config;
+    private final Executor executor;
+    private final StorageFactory systemStorageFactory;
+    private final StorageFactory userStorageFactory;
+
+    public NoOpStorageFactory(StorageExtraConfig config, Executor executor, StorageFactory systemStorageFactory, StorageFactory userStorageFactory) {
```
andreipaduroiu: When is `userStorageFactory` ever non-null? I see in `StorageLoader` that you pass `null` as this argument.

config/config.properties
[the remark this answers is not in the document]
kevinhan88: The prefix is the name space for the properties. Similarly for FileSystemStorage, S3Storage, HDFSStorage, their specific properties are all put under different name space. Otherwise the pravegaservice namespace is going to be very crowded; but more importantly, under the same name space, we may have to still differentiate properties by names such as "latencyForS3", "latencyForHDFS" for those common properties across storage.

config/config.properties
```diff
@@ -91,6 +91,22 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
 # Default value: HDFS
 # pravegaservice.storageImplementation=HDFS
 
+# Storage NO-OP Mode: segment payload write is no-oped; payload reading is not supported
+# This mode is used to avoid tier 2 interference in testing while keep the system functioning as usual.
```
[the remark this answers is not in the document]
kevinhan88: fixed

config/config.properties
```diff
@@ -91,6 +91,22 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
...
+#
+# NOTE: pravegaservice.storageImplementation is still used to store metadata.
```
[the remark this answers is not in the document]
kevinhan88: fixed

config/config.properties
```diff
@@ -91,6 +91,22 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
...
+#
+# Valid values: true, false
+# Default value: false
+#storageextra.storageNoOpMode=false
+
+# Only if Storage NO-OP Mode is true, storageWriteNoOpModeLatencyMilli is applied to write operation.
+# This is used to compensate the supposed latency if the tier2 storage is not in no-op mode.
+# Valid values: non-negative integer
+# Default vale: 20 (milliseconds)
+#storageextra.storageWriteNoOpLatencyMilli=20
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/StorageLoader.java
```diff
@@ -26,10 +29,17 @@
...
         ServiceLoader<StorageFactoryCreator> loader = ServiceLoader.load(StorageFactoryCreator.class);
-        for (StorageFactoryCreator factory : loader) {
-            log.info("Loading {}, trying {}", storageImplementation, factory.getName());
-            if (factory.getName().equals(storageImplementation)) {
-                return factory.createFactory(setup, executor);
+        for (StorageFactoryCreator factoryCreator : loader) {
+            log.info("Loading {}, trying {}", storageImplementation, factoryCreator.getName());
+            if (factoryCreator.getName().equals(storageImplementation)) {
+                StorageFactory factory = factoryCreator.createFactory(setup, executor);
+                StorageExtraConfig noOpConfig = setup.getConfig(StorageExtraConfig::builder);
```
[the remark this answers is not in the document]
kevinhan88: fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/StorageLoader.java
```diff
@@ -26,10 +29,17 @@
...
+                if (!noOpConfig.isStorageNoOpMode()) {
+                    return factory;
+                } else { //No-Op mode is set to true
+                    log.info("WARNING !!! {} is in NO-OP MODE: make sure this is by full intention for testing purpose!", storageImplementation);
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/server/src/test/java/io/pravega/segmentstore/server/store/StreamSegmentServiceNoOpWriteOnlyTests.java
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
```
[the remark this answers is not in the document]
kevinhan88: fixed.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/store/StreamSegmentServiceNoOpWriteOnlyTests.java
```diff
@@ -0,0 +1,102 @@
...
+                .withDataLogFactory(setup -> this.durableDataLogFactory);
+    }
+
+    /**
+     * Ignore testEndToEnd because it includes segment content verification,
+     * which is not possible when write is no-oped.
+     */
+    @Override
+    @Ignore
+    public void testEndToEnd() throws Exception {
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * Storage adapter for testing scenario without interference from tier 2.
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * . When this storage is used, all system segments are written to system storage.
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
```diff
@@ -0,0 +1,267 @@
...
+ * . If userStorage is present, user data segments are written to it; otherwise no-op.
+ * . If userStorage is present, user data segments are read from userStorage; otherwise UnsupportedOperationException thrown.
+ * . For no-oped write operation, latency is applied in order to make the no-op as real as possible.
+ */
+@Slf4j
+class NoOpStorage implements SyncStorage {
+
+    // Storage for system segments; must not null.
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
```diff
@@ -0,0 +1,267 @@
...
+        if (userStorage != null) {
+            userStorage.close();
+        }
+    }
+
+    @Override
+    public void initialize(long epoch) {
+        systemStorage.initialize(epoch);
+        if (userStorage != null) {
+            userStorage.initialize(1);
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
kevinhan88: fixed

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
kevinhan88: Yes in production code when fully qualified segment names are used, it start from "_system"; however in test code, too many shortcut system segment names start with "_".
It seems not practical to start using all fully qualified segment name for all scenarios.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
kevinhan88: Seems only checking the leading "_" just works for this scenario too.
As I explained above, in test codes we used too many short cut segment name instead of fully qualified names. But so far the bottom line ("_" is for internal segments) is protected.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorageFactory.java
kevinhan88: In the unit test of NoOpStorage itself, the supposedly no-oped segments are stored in this storage.
It is essential to verify the no-oped segment to ensure the no-op storage works as expected. For example, there is a rollingStorage layer inbetween NoOpStorage and underlying base storage, rolling layer may seal segment when things got all "rolled" up. If we don't verify the No-Oped segment, then it's impossible to test whether the rolling layer is doing things correctly.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/noop/NoOpStorageSystemDataWriteReadTests.java
kevinhan88: Fixed. The method must show up as it's defined as abstract method in super

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
kevinhan88: Added table segment as pass-through in NoOpStorage.

As for the INTERNAL_SCOPENAME, as I explained above, there are many test codes not using fully qualified segment names. If we think it's necessary now to enforce it, it worth a separate effort to convert all the disqualified segment into qualified in all test code.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+
+    @Override
+    public SegmentProperties getStreamSegmentInfo(String streamSegmentName) throws StreamSegmentException {
+        if (isSystemSegment(streamSegmentName)) {
```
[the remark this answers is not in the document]
kevinhan88: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/noop/NoOpStorage.java
kevinhan88: fixed

config/config.properties
```diff
@@ -91,6 +91,21 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
 # Default value: HDFS
 # pravegaservice.storageImplementation=HDFS
 
+# Storage NO-OP Mode: in No-Op mode, user stream segment writing is no-oped; user stream segment reading is not supported.
```
RaulGracia: This sentence is not obvious: `user stream segment writing is no-oped`. Maybe something like this one (or similar can be clearer for a user): `in No-Op mode, Pravega does not write the payloads of events to Tier 2 storage. Therefore, reads involving data from Tier 2 storage are not supported`.

config/config.properties
```diff
@@ -91,6 +91,21 @@ pravegaservice.dataLogImplementation=BOOKKEEPER
...
+# This mode is used to avoid storage interference in testing while still keep the system functioning as usual.
```
RaulGracia: Suggestion for rephrase: `This mode is useful to test the performance of Pravega against Tier 1 (e.g., Bookkeeper) without interferences related to moving data to Tier 2.`

PR #4294, opened 2019-10-18
Issue 4303: Avoid Netty Flush from the Client

tkaitchuck:
Signed-off-by: Tom Kaitchuck <tom.kaitchuck@emc.com>

**Change log description**  
* Avoids calling netty's channel.flush()
* Avoids blocking on the futures returned from write() unless it is needed.

**Purpose of the change**  
This is an experiment to find out if this improves performance on small writes.

**What the code does**  
The code creates an alternative to the simple `writeAndFlush` in ClientConnectionImpl which works by cooperating with CommandEncoder to delay the flush call until after the encoding has been performed. (To work around the way netty works this needs to be done via a subclass). 

**How to verify it**  
This needs a LOT of additional testing as this is a high risk change.

PR #4304, opened 2019-10-22
Non-blocking event writing

### LINKED ISSUE #4303: WriteEvent call is blocked for approx 90% of the time
**Problem description**
The client performance is expected to be degrading due to producer calls blocking.

**Problem location**
`io.pravega.client.segment.impl.SegmentOutputStreamImpl#write`
method calls `send` method which blocks

Use something like `io.pravega.client.netty.impl.ClientConnection#sendAsync(java.util.List<io.pravega.shared.protocol.netty.Append>, io.pravega.client.netty.impl.ClientConnection.CompletedCallback)`
instead

akurilov:
**Change log description**  
- Add `void sendAsync(Append append, CompletedCallback callback)` method for `ClientConnection`
- Replace the blocking `send` call with the new `sendAsync` call in `io.pravega.client.segment.impl.SegmentOutputStreamImpl#write`

**Purpose of the change**  
Fixes #4303

**What the code does**  
Send the append asynchronously. The failure handling is moved to the callback.

**How to verify it**  
Producer shouldn't be blocked for the majority of the time, the client performance is expected to become better.


tkaitchuck:
https://github.com/pravega/pravega/pull/4294 is already attempting to do this, but it retains backpressure so that the client won't OOM if the server is slower than the client is writing.
Feel free to run experiments on that branch.

Issue #4305, opened 2019-10-22
The unlimited pending events count causes OOM


akurilov:
**Problem description**
The `writeEvent` call adds all pending events into the `inflight` map. The map is not limited by size so this may cause client OOM.

**Problem location**
https://github.com/pravega/pravega/blob/43c5a16536736945cb0f862c5a920b72c6781c23/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java#L110

**Suggestions for an improvement**
Experiments show that for the client application heap limit of 4GB up to several hundreds of thousands of pending events may be in flight simultaneously. So the count limit should be less than 1,000,000. For example, 100,000.

tkaitchuck:
Existing back pressure should prevent this. This is why we don't observe the OOM in master that was introduced in the Draft PR: https://github.com/pravega/pravega/pull/4304

PR #4315, opened 2019-10-23
Non-blocking event writing, limit the pending events count

akurilov:
Signed-off-by: Andrey Kurilov <andrey.kurilov@emc.com>

**Change log description**  

* Replace the blocking send call with the new sendAsync call in `io.pravega.client.segment.impl.SegmentOutputStreamImpl#write`
* Make inflight map limited by size (100K)

**Purpose of the change**  
Fixes #4304, #4305

**What the code does**  
* Allow the client to write events asynchronously
* Protect the client from OOM

**How to verify it**  
 The client performance is expected to become better.


akurilov:
Looks like I have merged something unwanted, I'll recreate the PR

PR #4331, opened 2019-10-31
Issue 4239: AppendProcessor pipelining

### LINKED ISSUE #4329: Optimize AppendProcessor
**Problem description**
AppendProcessor does not make use of append pipelining in the underlying store.

**Problem location**
AppendProcessor.java

Have AppendProcessor make use of append pipelining in the underlying store.

andreipaduroiu:
**Change log description**  
- Changed AppendProcessor to pipeline requests through the underlying SegmentStore.
- Changed the AppendProcessor's TCP back-throttling logic to consider all connections when deciding which connection to pause or resume.

**Purpose of the change**  
Fixes #4329.

**What the code does**  
- `AppendProcessor` immediately sends the append requests as to the underlying Segment store, in the same order as it receives them. It no longer performs any aggregations at its level.
    - It sends the acks asynchronously as soon as each Store request is done. 
    - Multiple synchronization blocks have been removed (unnecessary), yet there still had to be a per-writer sync block, since the Client expects acks to arrive in order, without gaps. It is possible that the acks coming from the Segment store may be out of order or due to multi-threading the acks sent to the Client may get out-of-order, hence the new per-writer `synchronized`block in `handleAppendResult`
    - The connection pause/resume logic has been moved out of this class and into `ConnectionTracker`. `AppendProcessor` simply reports the size of the appends as they come in or out and keeps track of the per-connection outstanding byte count.
- `ConnectionTracker`
    - Shared object (across all `AppendProcessor` instances owned by a single `PravegaRequestListener`) that keeps track of the total outstanding byte count (across all connections).
    - Pauses all connections if the cumulative total exceeds a specified threshold and pauses individual connections if they exceeded the per-connection max limit (which may vary dynamically based on what other connections are doing).

**How to verify it**  
All unit tests must pass.
System tests must pass.
Perf tests must show improved throughput/latency for single-segment tests, with no negative repercussions for multi-segment tests.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -385,58 +317,70 @@ private void logError(String segment, Throwable u) {
...
 
-    /**
-     * return the remaining bytes to be consumed.
-     */
-    protected long getPendingBytes() {
-        return pendingBytes.get();
-    }
+        /**
+         * Invoked when a conditional append has failed due to {@link BadOffsetException}. If no more appends are in the
+         * pipeline, then the Last Stored Event Number is reverted to the Last (Successfully) Acked Event Number.
```
tkaitchuck: Shouldn't this be done unconditionally? After all won't all the intermediate writes fail with a badAttributeException?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -385,58 +317,70 @@ private void logError(String segment, Throwable u) {
...
-    protected void setPendingBytes(long val) {
-        pendingBytes.set(Math.max(val, 0));
-    }
+        /**
+         * Invoked when an append has been successfully stored.
+         *
+         * @param eventNumber The Append's Event Number.
+         * @return The last successful Append's Event Number (prior to this one).
+         */
+        synchronized long appendSuccessful(long eventNumber) {
```
tkaitchuck: This is subject to some very complicated race conditions. It looks like the Max takes care of this in conjunction with how it is called. But this should be called out in the javadoc, and a justification for correctness can be provided.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -263,56 +211,40 @@ private Append getNextAppend() {
...
             if (success) {
-                final DataAppended dataAppendedAck = new DataAppended(append.getRequestId(), append.getWriterId(), append.getEventNumber(),
-                        previousEventNumber, newWriteOffset);
-                log.trace("Sending DataAppended : {}", dataAppendedAck);
-                connection.send(dataAppendedAck);
+                Preconditions.checkState(state != null, "Synchronization error while processing append: %s. Unable to send ack.", append);
+                synchronized (state.getAckLock()) {
+                    // Acks must be sent in order. The only way to do this is by using a lock.
+                    long previousLastAcked = state.appendSuccessful(append.getEventNumber());
+                    if (previousLastAcked < append.getEventNumber()) {
```
tkaitchuck: This `if` performs similar logic to the `max` inside of the appendSuccessful and neither works without the other. 
As these are split in two locations it makes it hard to reason about. It would be better to have all the logic either here or in one private function so that it's obvious previousLastAsked is correct.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -189,7 +191,7 @@ public void initChannel(SocketChannel ch) {
                      p.addLast(TLSHelper.TLS_HANDLER_NAME, sslHandler);
                  }
 
-                 ServerConnectionInboundHandler lsh = new ServerConnectionInboundHandler();
+                 ServerConnectionInboundHandler lsh = new ServerConnectionInboundHandler(connectionTracker);
```
tkaitchuck: This could just as easily go into the append processor.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
andreipaduroiu: This is to handle offset-conditional appends, which do not close the connection. If an offset-conditional append fails due to bad offset, the WriterState's lastStoredEventNumber will have already been updated to a now-bad value; subsequent calls, even if they have the correct offset, will automatically trigger a BadAttributeUpdate and close the connection unnecessarily. This is only invoked for BadOffsetException to revert to the last known good state, which doesn't have that side effect.

This wasn't a problem before since all append were sequential and the "lastStoredEventNumber" was only updated after a successful append. This can no longer be done with pipelining; we need to update it before we make the append and deal with failure later.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
andreipaduroiu: Actually this one does a slightly different job than `appendSuccessful`.
- `appendSuccessful` records the fact that an append with an EventNumber is complete, and updates the state. It is meant to record the last Event Number that was ack-ed.
- This piece of code deals with actually sending the Ack. I do not understand why the client needs to know both the previously acked event number and the current one, when the current one would have been enough. It also requires them to be sent in the correct order (i.e., 10-15 must be before 15-19, and not vice-versa).  If they aren't, then it will throw an exception and close the connection. 
- I chose it to have two separate `synchronized` blocks because I did not  want sync-ing for acks to interfere with updating the `lastStoredEventNumber`. Updating 2 fields is faster than sending the ack over the wire so I didn't want to block any new appends on that completing.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
andreipaduroiu: Ok

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
andreipaduroiu: This makes testing a lot simpler. Instead of adding complicated checks in `AppendProcessorTest` to figure out when to pause or resume a connection, it's much simpler to just verify that the `adjustOutstandingBytes` method is invoked on this mock with a predetermined value.

Then in `ServerConnectionInboundHandler`, it's much easier to reason about pausing or resuming connections when passing in specific values to that method.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/ServerConnection.java
```diff
@@ -32,9 +32,11 @@
...
     void setRequestProcessor(RequestProcessor cp);
 
-    void pauseReading();
-
-    void resumeReading();
+    /**
+     * Adjusts the number of outstanding bytes by the given amount.
+     * @param delta The amount to adjust by. May be negative.
+     */
+    void adjustOutstandingBytes(int delta);
```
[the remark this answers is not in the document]
andreipaduroiu: https://github.com/pravega/pravega/pull/4331#discussion_r341412402

andreipaduroiu:
https://github.com/pravega/pravega/pull/4331#pullrequestreview-310220437

I added a unit test that verifies this.

andreipaduroiu:
2 System test runs passed: 2388, 2390.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
tkaitchuck: Ah. What if the ConnectionTracker method instead of returning a boolean took in a Connection that is paused or resumed?

andreipaduroiu: I assume you don't imply that ConnectionTracker has an internal collection of all registered connections. That would make cleaning up a very important task, as we could easily introduce leaks.

Do you mean that `ConnectionTracker.adjustOutstandingBytes` takes in a `ServerConnection` object that it itself will pause or resume?

If I understand your suggestion properly:
- We keep `ServerConnection` as it is (in master) with a `pauseConnection` and `resumeConnection`
- In `AppendProcessor`, which already has a reference to the `ServerConnection`, invokes `ConnectionTracker.pauseOrResumeConnection(int deltaBytes, int connectionOutstandingBytes, ServerConnection c)` ?
- Modify `ConnectionTracker` accordingly.

This would mean the following:
- Add yet another arg to the 2 AppendProcessor's constructors, and plumb it in all the tests that reference it directly.
- It may make `AppendProcessorTest` a bit tricky since I can't estimate the `connectionOutstandingBytes` properly due to the processing being async internally.


This could be done. Is your goal here to not have `ServerConnection` keep track of its outstanding bytes and leave that to `AppendProcessor`?

tkaitchuck: yes

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -192,66 +168,35 @@ public void setupAppend(SetupAppend setupAppend) {
...
+        Preconditions.checkState(state != null, "Data from unexpected connection: %s.", id);
+        long previousEventNumber = state.beginAppend(append.getEventNumber());
+        int appendLength = append.getData().readableBytes();
+        adjustOutstandingBytes(appendLength);
         Timer timer = new Timer();
-        storeAppend(append)
-                .whenComplete((length, e) -> {
-                    handleAppendResult(append, length, e, timer);
-                    LoggerHelpers.traceLeave(log, "storeAppend", traceId, append, e);
+        storeAppend(append, previousEventNumber)
```
tkaitchuck: To confirm the obvious: the store guarantees that the order of calls to append will be guaranteed in the event that another append is passed before the last one has completed.

andreipaduroiu: Yes. There is a great deal of effort in the `StreamSegmentContainer.java` and `MetadataStore.java` that ensures that, even if intermediate async calls are ack-ed out of order.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -263,56 +208,40 @@ private Append getNextAppend() {
         }
     }
 
-    private void handleAppendResult(final Append append, Long newWriteOffset, Throwable exception, Timer elapsedTimer) {
+    private void handleAppendResult(final Append append, Long newWriteOffset, Long previousEventNumber, Throwable exception, Timer elapsedTimer) {
```
shrids: We can use the primitive `long` here instead of `Long` right ?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -263,56 +208,40 @@ private Append getNextAppend() {
...
-                connection.send(dataAppendedAck);
+                Preconditions.checkState(state != null, "Synchronization error while processing append: %s. Unable to send ack.", append);
+                synchronized (state.getAckLock()) {
+                    // Acks must be sent in order. The only way to do this is by using a lock.
+                    long previousLastAcked = state.appendSuccessful(append.getEventNumber());
+                    if (previousLastAcked < append.getEventNumber()) {
+                        final DataAppended dataAppendedAck = new DataAppended(append.getRequestId(), append.getWriterId(), append.getEventNumber(),
+                                previousLastAcked, newWriteOffset);
+                        log.trace("Sending DataAppended : {}", dataAppendedAck);
+                        connection.send(dataAppendedAck);
```
shrids: `connection.send(...)` does a `channel.writeAndFlush`. Did not understand the reason why we should we do this within the synchronized lock 🤔

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
shrids: `previousEventNumber` is never used in this method and can be removed.

andreipaduroiu:
System tests 2395, 2396 and 389 passed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
andreipaduroiu: The client expects acks to arrive in the correct order (i.e., 1-5, 6-7,8-100). If they arrive out of order or have gaps it will freak out and close the connection.

This method is a callback from the `StreamSegmentStore.append`. While the Segment Store does issue acks in order, due to thread scheduling issues this callback may be invoked in arbitrary order. The code in `WriterState` does account for that, though.

The problem is that this callback is also subject to extreme race conditions. Since appends are pipelined to the store, the acks will not be serialized, but be issued in parallel. As such, more than one thread could be in this method. Thread A can invoke `writerState.appendSuccessful` and get ack range N1-N2 and Thread B can get range N2-N3. However Thread B could get to `connection.send` before thread A and, as such, send the acks out of order (N2-N3, followed by N1-N2).

The point of this sync block is to ensure that the acks are sent in order. 

If the client can be modified to handle acks out of order, given the pipelining guarantees of the store, we could remove that constraint from the client and simplify this code.

PR #4332, opened 2019-11-03
Issue 3745: Moved Security Constants

### LINKED ISSUE #3745: Move security constants to :shared:test

Prabhaker24:
Change log description
moved the security constants from ```:test:common``` to ```:shared:security```

Purpose of the change
fixes #3745 

What the code does
Moved SecurityConfigDefaults from ```:test:common``` to ```:shared:security``` as these defaults are specific to pravega and shouldn't be inside the ```:test:common```

How to verify it
Running all the tests with tls enabled


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Sun Nov 03 17:36:54 IST 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
```
andreipaduroiu: Revert this file.

PR #4337, opened 2019-11-06
Issue 3745: Moved security constants

### LINKED ISSUE #3745: Move security constants to :shared:test

Prabhaker24:
**Change log description**  
moved the security constants from ```:test:common``` to ```:shared:security```

**Purpose of the change**  
fixes #3745 

**What the code does**  
Moved SecurityConfigDefaults from ```:test:common``` to ```:shared:security``` as these defaults are specific to pravega and shouldn't be inside the ```:test:common```, many packages had test dependency on the ```shared: security``` i.e the tests in many packages depended on tests of ```shared: security``` and the best way of including such dependency is to create a jarTest for ```shared:security``` jarTest task builds the jar from the test outputs adding a ‘test’ to the jar name then I have created a configuration which basically is a group of dependencies and finally added the jar artifact to the configuration I have passed the configuration as parameter to the project that depends on it , thus whenever someone(project which is dependent on it) ask for testOutput, gradle will run the jarTest task

**How to verify it**  
Running all the tests with tls enabled


shared/security/src/test/java/io/pravega/shared/security/utils/SecurityConfigDefaults.java
```diff
@@ -7,7 +7,7 @@
  *
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
-package io.pravega.test.common;
+package io.pravega.shared.security.utils;
```
ravisharda: It isn't clear looking at the package name or the class name (`io.pravega.shared.security.utils.SecurityConfigDefaults`) that this class is for use in test classes only, even though the location of the source file clarifies it a bit. I'd suggest that it should be made explicit by relocating this class to: `io.pravega.shared.security.test.SecurityConfigDefaults`.

build.gradle
```diff
@@ -169,6 +169,19 @@ project('shared:authplugin') {
 }
 
 project('shared:security') {
+    task jarTest (type: Jar) {
```
ravisharda: Can you explain why these changes are needed in the PR `description > what the code does?`?

build.gradle
```diff
@@ -249,6 +262,7 @@ project('client') {
         // The client fallsback to NIO based transport if epoll is not available.
         compile group: 'io.netty', name: 'netty-transport-native-epoll', version: nettyVersion, classifier: 'linux-x86_64'
         compile group: 'com.google.guava', name: 'guava', version: guavaVersion
+        testCompile project(path: ':shared:security', configuration: 'testOutput')
```
ravisharda: Why was this needed?

build.gradle
```diff
@@ -469,6 +484,7 @@ project('test:integration') {
         compile project(':test:testcommon')
         compile group: 'junit', name:'junit', version: junitVersion
         compile group: 'org.apache.curator', name: 'curator-test', version: apacheCuratorVersion
+        compile project(path: ':shared:security', configuration: 'testOutput')
```
ravisharda: Why is this needed, and what is it doing?

shared/security/src/test/java/io/pravega/shared/security/utils/SecurityConfigDefaults.java
Prabhaker24: I have changed the package name

build.gradle
Prabhaker24: many packages had test dependency on the ```shared: security``` i.e the tests in many packages depended on tests of ```shared: security``` and the best way of including such dependency is to create a jarTest for ```shared:security``` jarTest task builds the jar from the test outputs adding a ‘test’ to the jar name then I have created a configuration which basically is a group of dependencies and finally added the jar artifact to the configuration I have passed the configuration as parameter to the project that depends on it , thus whenever someone(project which is dependent on it) ask for testOutput, gradle will run the jarTest task

build.gradle
Prabhaker24: this line is needed as project client has test dependency over project security, when this is run gradle will run the jarTest task in the security project in the build.gradle.

build.gradle
Prabhaker24: This line is needed as the project integration has compile dependency over the test present in the project security so when this is run gradle will run the jarTest task in the security project in the build.gradle.

andreipaduroiu:
@Prabhaker24 Are we still pursuing this PR?

Prabhaker24:
> @Prabhaker24 Are we still pursuing this PR?

@ravisharda can you please take a look at the pr again?

andreipaduroiu:
@ravisharda @Prabhaker24 Please let me know your intentions with this PR. If I don't see any response or activity by March 31, I will close it as being abandoned.

andreipaduroiu:
Closed for inactivity. Please reopen if work is to resume on this task.

PR #4215, opened 2019-09-26
Issue 4107: Automatic delegation token renewal

ravisharda:
**Change log description**  
Introduces automatic token renewal. 

**Purpose of the change**  
Resolves #4107. 

**What the code does**  
Here's what the code does, at a high level:
1. Adds a `DelegationTokenProvider` (as well as related factory and implementation classes) that is used as a proxy for a delegation token in places where a string delegation token was used. 

   For JWT tokens, when the token is retrieved via the `retrieveToken()` method, the provider checks if the delegation token is nearing expiry. If the token is nearing expiry, it refreshes the token, i.e., fetches a new token and returns that one. Otherwise, it returns an existing token. 
   
2. Adds a token handling strategy for tackling these different scenarios: 
    * A JWT token
    * An empty token which occurs when auth is disabled. This strategy is also used in many test cases. 
    * A non-JWT token, which is used in some of the tests. 
    * A null token for cases where a delegation token is not populated in advance and is fetched lazily when needed. 

**How to verify it**  
All tests should pass.

I used the following temporary test suite to verify the changes: 

```java
package io.pravega.test.integration;

import io.pravega.client.security.auth.DelegationTokenProviderFactoryTest;
import io.pravega.client.security.auth.JwtBodyTest;
import io.pravega.client.security.auth.JwtTokenProviderImplTest;
import io.pravega.client.segment.impl.AsyncSegmentInputStreamTest;
import io.pravega.client.segment.impl.ConditionalOutputStreamTest;
import io.pravega.client.segment.impl.SegmentMetadataClientTest;
import io.pravega.client.state.impl.RevisionedStreamClientTest;
import io.pravega.client.stream.impl.EventStreamReaderTest;
import io.pravega.client.stream.impl.SegmentSelectorTest;
import io.pravega.segmentstore.server.host.handler.PravegaRequestProcessorAuthFailedTest;
import io.pravega.shared.protocol.netty.WireCommandsTest;
import org.junit.runner.RunWith;
import org.junit.runners.Suite;

@RunWith(Suite.class)
@Suite.SuiteClasses({
        // New unit test classes focused on the enhancement
        JwtTokenProviderImplTest.class,
        DelegationTokenProviderFactoryTest.class,
        JwtBodyTest.class,

        // Some of the relevant existing unit test classes
        EventStreamReaderTest.class,
        RevisionedStreamClientTest.class,
        AppendTest.class,
        AppendReconnectTest.class,
        ReadTest.class,
        SegmentSelectorTest.class,
[...]

PR #4356, opened 2019-11-12
Issue 4352: Turn retrieve token operation of DelegationTokenProvider asynchronous

### LINKED ISSUE #4352: AutoScaleProcessor scale events get blocked due to a deadlock
**Problem description**
In a recent test, we have noticed that scaling signals (indicating whether a segment is hot/cold) posted by the `AutoScaleProcessor` class in `SegmentStore` on the internal request stream (`_scope/_requeststream`) did not reach the Controller. 

The signals didn't reach the Controller, because the writer used in the `AutoScaleProcessor` isn't able to write them to the stream. 

Upon investigation, we found that the root cause is as described below (source: @shiveshr). 

`SegmentOutputStreamImpl` uses `connectionFactory.getInternalExecutor()` to send the setupAppend command, which uses the same executor also given to the Controller client for making all calls to the Controller. The call to `JwtTokenProviderImpl.retrieveToken()` while composing `SetupAppend` request is also made on the same internal executor.

https://github.com/pravega/pravega/blob/e81f6abd3e65d0e86ce85cfc6fc313fdfcabb043/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java#L578-L606

The `JwtTokenProviderImpl.retrieveToken()` makes a call to the controller and blocks on a future to complete. 

https://github.com/pravega/pravega/blob/c87f62340fcd89f322854f62edea165ec2a1061e/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java#L236

And, on this server there is exactly one internal thread in that threadpool: `clientInternal-1-1`

In the thread dump we can see that this thread is blocked: 

```
clientInternal-1-1" #60 daemon prio=5 os_prio=0 tid=0x0000555f8c93a000 nid=0x80 waiting on condition [0x00007f716aeeb000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000000db216dd0> (a java.util.concurrent.CompletableFuture$Signaller)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1693)
	at java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3323)
	at java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1729)
	at java.util.concurrent.CompletableFuture.join(CompletableFuture.java:1934)
	at io.pravega.client.security.auth.JwtTokenProviderImpl.refreshToken(JwtTokenProviderImpl.java:236)
	at io.pravega.client.security.auth.JwtTokenProviderImpl.retrieveToken(JwtTokenProviderImpl.java:191)
	at io.pravega.client.segment.impl.SegmentOutputStreamImpl.lambda$null$3(SegmentOutputStreamImpl.java:576)
	at io.pravega.client.segment.impl.SegmentOutputStreamImpl$$Lambda$644/1564253710.apply(Unknown Source)
	at java.util.concurrent.CompletableFuture.uniCompose(CompletableFuture.java:952)
	at java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:926)
	at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:442)
	at java.util.concurr...
...
"auto-scaler-6" #59 daemon prio=5 os_prio=0 tid=0x0000555f8c929000 nid=0x76 waiting on condition [0x00007f716afec000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000000db267de0> (a java.util.concurrent.CompletableFuture$Signaller)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1693)
	at java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3323)
	at java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1729)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1895)
	at io.pravega.common.concurrent.Futures.getThrowingException(Futures.java:153)
	at io.pravega.client.segment.impl.SegmentOutputStreamImpl.write(SegmentOutputStreamImpl.java:459)
	- locked <0x00000000d7517a50> (a java.lang.Object)
	at io.pravega.client.stream.impl.EventStreamWriterImpl.writeEventInternal(EventStreamWriterImpl.java:136)
	- locked <0x00000000d6d0b590> (a java.lang.Object)
	- locked <0x00000000d6d0b580> (a java.lang.Object)
```
So the thread is not available to make call to controller as it is blocked on waiting for call to controller to complete.

**Problem location**
`JwtTokenProviderImpl.java`, `SegmentOutputStreamImpl.java`

Make all calls in `JwtTokenProviderImpl.java` asynchronous and not block on the same thread.

ravisharda:
**Change log description**  
Make the `DelegationTokenProvider` interface operation `retrieveToken()` asynchronous and make its invocations in client non-blocking, in order to avoid client-side deadlocks. 

**Purpose of the change**  
Fixes #4352 

**What the code does**  
* Makes the `retrieveToken()` operation of the `DelegationTokenProvider` interface asynchronous, and modifies the following implementation classes accordingly: 
  * `JwtTokenProviderImpl` 
  * `EmptyTokenProviderImpl` 
  * `StringTokenProviderImpl`
* Modifies the following client classes to avoid blocking, by leveraging `CompletableFuture`'s `CompletionStage` tasks: 
  * `AsyncSegmentInputStreamImpl`
  * `ConditionalOutputStreamImpl.java`
  * `SegmentMetadataClientImpl.java`
  * `SegmentOutputStreamImpl.java`
* Modifies existing tests to account for the above items. 

* Modifies existing code 

**How to verify it**  
All existing tests - unit, integration and system tests - should pass. `Auth` enabled autoscale system tests should pass.  

Here's a subset of unit and integration tests I use to quickly verify the changes: 
```java
package io.pravega.test.integration;

import io.pravega.client.security.auth.DelegationTokenProviderFactoryTest;
import io.pravega.client.security.auth.JwtBodyTest;
import io.pravega.client.security.auth.JwtTokenProviderImplTest;
import io.pravega.client.segment.impl.AsyncSegmentInputStreamTest;
import io.pravega.client.segment.impl.ConditionalOutputStreamTest;
import io.pravega.client.segment.impl.SegmentMetadataClientTest;
import io.pravega.client.segment.impl.SegmentOutputStreamTest;
import io.pravega.client.state.impl.RevisionedStreamClientTest;
import io.pravega.client.stream.impl.EventStreamReaderTest;
import io.pravega.client.stream.impl.SegmentSelectorTest;
import io.pravega.segmentstore.server.host.handler.PravegaRequestProcessorAuthFailedTest;
import io.pravega.shared.protocol.netty.WireCommandsTest;

@RunWith(Suite.class)
@Suite.SuiteClasses({
        JwtTokenProviderImplTest.class,
        DelegationTokenProviderFactoryTest.class,
        JwtBodyTest.class,
        EventStreamReaderTest.class,
        RevisionedStreamClientTest.class,
        AppendTest.class,
        AppendReconnectTest.class,
        ReadTest.class,
        SegmentSelectorTest.class,
        ConditionalOutputStreamTest.class,
        PravegaRequestProcessorAuthFailedTest.class,
        WireCommandsTest.class,
        AsyncSegmentInputStreamTest.class,
        SegmentMetadataClientTest.class,
        SegmentOutputStreamTest.class,
        DelegationTokenTest.class
})

public class DelegationTokenTestSuite {
}
```

### DIFF
--- a/client/src/main/java/io/pravega/client/security/auth/DelegationTokenProvider.java
+++ b/client/src/main/java/io/pravega/client/security/auth/DelegationTokenProvider.java
@@ -22,7 +22,7 @@ public interface DelegationTokenProvider {
     /**
      * Retrieve delegation token.
      *
-     * @return a delegation token
+     * @return a CompletableFuture that, when completed, will return the retrieved delegation token
      */
     CompletableFuture<String> retrieveToken();
 
--- a/client/src/main/java/io/pravega/client/security/auth/DelegationTokenProvider.java
+++ b/client/src/main/java/io/pravega/client/security/auth/DelegationTokenProvider.java
@@ -9,6 +9,8 @@
  */
 package io.pravega.client.security.auth;
 
+import java.util.concurrent.CompletableFuture;
+
 /**
  * A client-side proxy for obtaining a delegation token from the server.
  *
@@ -22,7 +24,7 @@ public interface DelegationTokenProvider {
      *
      * @return a delegation token
      */
-    String retrieveToken();
+    CompletableFuture<String> retrieveToken();
 
     /**
      * Populates the object with the specified delegation token.
--- a/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
+++ b/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
@@ -176,7 +176,7 @@ public class JwtTokenProviderImpl implements DelegationTokenProvider {
      * Returns the delegation token. It returns existing delegation token if it is not close to expiry. If the token
      * is close to expiry, it obtains a new delegation token and returns that one instead.
      *
-     * @return String the delegation token JWT compact value
+     * @return a CompletableFuture that, when completed, will return the delegation token JWT compact value
      */
     @Override
     public CompletableFuture<String> retrieveToken() {
--- a/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
+++ b/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
@@ -179,18 +179,18 @@ public class JwtTokenProviderImpl implements DelegationTokenProvider {
      * @return String the delegation token JWT compact value
      */
     @Override
-    public String retrieveToken() {
+    public CompletableFuture<String> retrieveToken() {
         DelegationToken currentToken = this.delegationToken.get();
 
         if (currentToken == null) {
             return this.refreshToken();
         } else if (currentToken.getExpiryTime() == null) {
-            return currentToken.getValue();
+            return CompletableFuture.completedFuture(currentToken.getValue());
         } else if (isTokenNearingExpiry(currentToken)) {
             log.debug("Token is nearing expiry for scope/stream {}/{}", this.scopeName, this.streamName);
             return refreshToken();
         } else {
-            return currentToken.getValue();
+            return CompletableFuture.completedFuture(currentToken.getValue());
         }
     }
 
@@ -222,7 +222,7 @@ public class JwtTokenProviderImpl implements DelegationTokenProvider {
     }
 
     @VisibleForTesting
-    String refreshToken() {
+    CompletableFuture<String> refreshToken() {
         long traceEnterId = LoggerHelpers.traceEnter(log, "refreshToken", this.scopeName, this.streamName);
         CompletableFuture<Void> currentRefreshFuture = tokenRefreshFuture.get();
         if (currentRefreshFuture == null) {
@@ -232,13 +232,14 @@ public class JwtTokenProviderImpl implements DelegationTokenProvider {
         } else {
             log.debug("Token is already under refresh for scope {} and stream {}", this.scopeName, this.streamName);
         }
-        try {
-            currentRefreshFuture.join(); // Block until the token is refreshed
-        } finally {
-            this.tokenRefreshFuture.compareAndSet(currentRefreshFuture, null); // Token is already refreshed, so resetting the future to null.
-        }
-        LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
-        return delegationToken.get().getValue();
+
+        final CompletableFuture<Void> handleToCurrentRefreshFuture  = currentRefreshFuture;
+        return currentRefreshFuture.thenApply(v -> {
+                    // Token is already refreshed, so resetting the future to null.
+                    this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);
+                    LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
+                    return delegationToken.get().getValue();
+                });
     }
 
     private CompletableFuture<Void> recreateToken() {
--- a/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
@@ -168,7 +168,7 @@ class AsyncSegmentInputStreamImpl extends AsyncSegmentInputStream {
     public CompletableFuture<SegmentRead> read(long offset, int length) {
         Exceptions.checkNotClosed(closed.get(), this);
 
-        return this.tokenProvider.retrieveToken().thenCompose(token -> {
+        return this.tokenProvider.retrieveToken().thenComposeAsync(token -> {
             WireCommands.ReadSegment request = new WireCommands.ReadSegment(segmentId.getScopedName(), offset, length,
                     token, requestId);
             return backoffSchedule.retryWhen(t -> {
--- a/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
@@ -195,7 +195,7 @@ class AsyncSegmentInputStreamImpl extends AsyncSegmentInputStream {
                                 })
                         );
             }, connectionFactory.getInternalExecutor());
-        });
+        }, connectionFactory.getInternalExecutor());
     }
         
     private CompletableFuture<SegmentRead> sendRequestOverConnection(WireCommands.ReadSegment request, ClientConnection c) {
--- a/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
@@ -167,32 +167,35 @@ class AsyncSegmentInputStreamImpl extends AsyncSegmentInputStream {
     @Override
     public CompletableFuture<SegmentRead> read(long offset, int length) {
         Exceptions.checkNotClosed(closed.get(), this);
-        WireCommands.ReadSegment request = new WireCommands.ReadSegment(segmentId.getScopedName(), offset, length,
-                                                                        this.tokenProvider.retrieveToken(), requestId);
-        return backoffSchedule.retryWhen(t -> {
-            Throwable ex = Exceptions.unwrap(t);
-            if (closed.get()) {
-                log.debug("Exception while reading from Segment : {}", segmentId, ex);
-            } else {
-                log.warn("Exception while reading from Segment : {}", segmentId, ex);
-            }
-            return ex instanceof Exception && !(ex instanceof ConnectionClosedException) && !(ex instanceof SegmentTruncatedException);
-        }).runAsync(() -> {
-            return getConnection()
-                    .whenComplete((connection, ex) -> {
-                        if (ex != null) {
-                            log.warn("Exception while establishing connection with Pravega node", ex);
-                            closeConnection(new ConnectionFailedException(ex));
-                        }
-                    }).thenCompose(c -> sendRequestOverConnection(request, c)
-                            .whenComplete((reply, ex) -> {
-                                if (ex instanceof ConnectionFailedException) {
-                                    log.debug("ConnectionFailedException observed when sending request {}", request, ex);
-                                    closeConnection((ConnectionFailedException) ex);
-                                }
-                            })
-                    );
-        }, connectionFactory.getInternalExecutor());
+
+        return this.tokenProvider.retrieveToken().thenCompose(token -> {
+            WireCommands.ReadSegment request = new WireCommands.ReadSegment(segmentId.getScopedName(), offset, length,
+                    token, requestId);
+            return backoffSchedule.retryWhen(t -> {
+                Throwable ex = Exceptions.unwrap(t);
+                if (closed.get()) {
+                    log.debug("Exception while reading from Segment : {}", segmentId, ex);
+                } else {
+                    log.warn("Exception while reading from Segment : {}", segmentId, ex);
+                }
+                return ex instanceof Exception && !(ex instanceof ConnectionClosedException) && !(ex instanceof SegmentTruncatedException);
+            }).runAsync(() -> {
+                return getConnection()
+                        .whenComplete((connection, ex) -> {
+                            if (ex != null) {
+                                log.warn("Exception while establishing connection with Pravega node", ex);
+                                closeConnection(new ConnectionFailedException(ex));
+                            }
+                        }).thenCompose(c -> sendRequestOverConnection(request, c)
+                                .whenComplete((reply, ex) -> {
+                                    if (ex instanceof ConnectionFailedException) {
+                                        log.debug("ConnectionFailedException observed when sending request {}", request, ex);
+                                        closeConnection((ConnectionFailedException) ex);
+                                    }
+                                })
+                        );
+            }, connectionFactory.getInternalExecutor());
+        });
     }
         
     private CompletableFuture<SegmentRead> sendRequestOverConnection(WireCommands.ReadSegment request, ClientConnection c) {
--- a/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
@@ -576,7 +576,6 @@ class SegmentOutputStreamImpl implements SegmentOutputStream {
                      log.info("Fetching endpoint for segment {}, writer {}", segmentName, writerId);
 
                      return controller.getEndpointForSegment(segmentName)
-
                          // Establish and return a connection to segment store
                          .thenComposeAsync((PravegaNodeUri uri) -> {
                              log.info("Establishing connection to {} for {}, writerID: {}", uri, segmentName, writerId);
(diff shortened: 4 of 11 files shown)


client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
```diff
@@ -167,32 +167,35 @@ public boolean isClosed() {
...
-                            .whenComplete((reply, ex) -> {
-                                if (ex instanceof ConnectionFailedException) {
-                                    log.debug("ConnectionFailedException observed when sending request {}", request, ex);
-                                    closeConnection((ConnectionFailedException) ex);
-                                }
-                            })
-                    );
-        }, connectionFactory.getInternalExecutor());
+
+        return this.tokenProvider.retrieveToken().thenCompose(token -> {
```
shiveshr: please use thenComposeAsync here to ensure this is executed on internal executor itself

ravisharda: Sure, done.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -222,7 +222,7 @@ boolean isWithinRefreshThreshold(Instant currentInstant, Instant expiration) {
     }
 
     @VisibleForTesting
-    String refreshToken() {
+    CompletableFuture<String> refreshToken() {
         long traceEnterId = LoggerHelpers.traceEnter(log, "refreshToken", this.scopeName, this.streamName);
         CompletableFuture<Void> currentRefreshFuture = tokenRefreshFuture.get();
         if (currentRefreshFuture == null) {
```
shrids: We can use the following snippet here. This eliminates the unnecessary recreateToken() invocation.
```
   final CompletableFuture<Void> currentRefreshFuture = tokenRefreshFuture.updateAndGet(fut -> {
            final CompletableFuture<Void> tokenFuture = null;
            if (fut == null) {
                log.debug("Initiated token refresh for scope {} and stream {}", this.scopeName, this.streamName);
                return( this.recreateToken());
            } else {
                log.debug("Token is already under refresh for scope {} and stream {}", this.scopeName, this.streamName);
                return fut;
            }
        });
```

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -232,13 +232,14 @@ String refreshToken() {
...
         }
-        try {
-            currentRefreshFuture.join(); // Block until the token is refreshed
-        } finally {
-            this.tokenRefreshFuture.compareAndSet(currentRefreshFuture, null); // Token is already refreshed, so resetting the future to null.
-        }
-        LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
-        return delegationToken.get().getValue();
+
+        final CompletableFuture<Void> handleToCurrentRefreshFuture  = currentRefreshFuture;
```
shrids: no need of this variable if the above approach is used.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
shiveshr: I notice a problem in the above suggestion (and its there with the original code too) -- its not optimizing on making the call to controller.
If two concurrent threads call refreshToken, we could have two calls made to controller to get delegation token and then one of them will race to win and set the token. 
we could do better than that by synchronizing here.

ravisharda: We wanted to avoid the use of explicit locking here, as discussed somewhere in [the PR](https://github.com/pravega/pravega/pull/4215) that implemented this. While we could optimize for avoiding multiple calls, which in turn can occur under rare circumstances, explicit locking perhaps would induce additional costs. Besides the suggested optimization is not necessary for fixing the specific issue at hand. I am thinking, I should leave it as it right now.

ravisharda: Using the `updateAndGet` operation would simplify the code but the unary operator would return the existing `fut` in the else condition, and the atomic update shall occur even when it is not needed.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
ravisharda: Please see my comment above.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
ravisharda: I've created a new enhancement issue to track/implement these suggestions: https://github.com/pravega/pravega/issues/4358, FYI.

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
```diff
@@ -115,6 +118,28 @@ public void testConnectAndSend() throws SegmentSealedException, ConnectionFailed
...
+                true, signal);
+
+        ClientConnection connection = mock(ClientConnection.class);
+        cf.provideConnection(uri, connection);
+        SegmentOutputStreamImpl output = new SegmentOutputStreamImpl(SEGMENT, true, controller, cf, cid, segmentSealedCallback,
+                RETRY_SCHEDULE, DelegationTokenProviderFactory.create(controller, "scope", "stream"));
+        output.reconnect();
+
+        signal.join();
+        assertEquals(1, controller.countOfDelegationTokenCalls.get());
```
shiveshr: you can use `mockito.verify(mock, times(k))` to count number of invocations.

ravisharda: Sure, fixed. :)

client/src/main/java/io/pravega/client/security/auth/DelegationTokenProvider.java
```diff
@@ -22,7 +24,7 @@
      *
      * @return a delegation token
      */
-    String retrieveToken();
+    CompletableFuture<String> retrieveToken();
```
andreipaduroiu: Update Javadoc `@return` tag

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -179,18 +179,18 @@ static Long parseExpirationTime(String jwtBody) {
      * @return String the delegation token JWT compact value
      */
     @Override
-    public String retrieveToken() {
+    public CompletableFuture<String> retrieveToken() {
```
andreipaduroiu: update javadoc

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -232,13 +232,14 @@ String refreshToken() {
...
-        } finally {
-            this.tokenRefreshFuture.compareAndSet(currentRefreshFuture, null); // Token is already refreshed, so resetting the future to null.
-        }
-        LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
-        return delegationToken.get().getValue();
+
+        final CompletableFuture<Void> handleToCurrentRefreshFuture  = currentRefreshFuture;
+        return currentRefreshFuture.thenApply(v -> {
+                    // Token is already refreshed, so resetting the future to null.
+                    this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);
```
andreipaduroiu: What happens if `compareAndSet` returns false? Your update will not have succeeded.

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
```diff
@@ -167,32 +167,35 @@ public boolean isClosed() {
...
-                            .whenComplete((reply, ex) -> {
-                                if (ex instanceof ConnectionFailedException) {
-                                    log.debug("ConnectionFailedException observed when sending request {}", request, ex);
-                                    closeConnection((ConnectionFailedException) ex);
-                                }
-                            })
-                    );
-        }, connectionFactory.getInternalExecutor());
+
+        return this.tokenProvider.retrieveToken().thenComposeAsync(token -> {
```
andreipaduroiu: Pass the executor to this method. Otherwise it will run on `ForkJoinPool`

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
```diff
@@ -167,32 +167,35 @@ public boolean isClosed() {
...
+            WireCommands.ReadSegment request = new WireCommands.ReadSegment(segmentId.getScopedName(), offset, length,
+                    token, requestId);
+            return backoffSchedule.retryWhen(t -> {
+                Throwable ex = Exceptions.unwrap(t);
+                if (closed.get()) {
+                    log.debug("Exception while reading from Segment : {}", segmentId, ex);
+                } else {
+                    log.warn("Exception while reading from Segment : {}", segmentId, ex);
+                }
+                return ex instanceof Exception && !(ex instanceof ConnectionClosedException) && !(ex instanceof SegmentTruncatedException);
```
andreipaduroiu: `ex instanceof Exception` is redundant

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -566,43 +567,57 @@ void reconnect() {
...
-                             if (exception instanceof SegmentSealedException) {
-                                 log.info("Ending reconnect attempts on writer {} to {} because segment is sealed", writerId, segmentName);
-                                 return null;
-                             }
-                             if (exception instanceof NoSuchSegmentException) {
-                                 log.info("Ending reconnect attempts on writer {} to {} because segment is truncated", writerId, segmentName);
-                                 return null;
+
+                     return controller.getEndpointForSegment(segmentName)
+
```
andreipaduroiu: remove empty line

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -232,13 +232,14 @@ String refreshToken() {
...
-        try {
-            currentRefreshFuture.join(); // Block until the token is refreshed
-        } finally {
-            this.tokenRefreshFuture.compareAndSet(currentRefreshFuture, null); // Token is already refreshed, so resetting the future to null.
-        }
-        LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
-        return delegationToken.get().getValue();
+
+        final CompletableFuture<Void> handleToCurrentRefreshFuture  = currentRefreshFuture;
+        return currentRefreshFuture.thenApply(v -> {
```
tkaitchuck: This block could be moved up to inside the if on line 231 so that it is only called once.

client/src/main/java/io/pravega/client/security/auth/DelegationTokenProvider.java
ravisharda: Done.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
ravisharda: Sure, done.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
ravisharda: Yes, that is as intended. We want to update it to null only if it is still the same object.

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
ravisharda: I've made the change in the latest push.

client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
ravisharda: This is an existing code and I don't want to modify it right now. Maybe the author intended to ensure it wasn't an error.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
ravisharda: Done.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
ravisharda: We do want to run this block in either case (`if`/`else`). So, I can't move it inside the `if`. Am I missing something?

tkaitchuck: I don't think we do.
When a token expires there may be multiple concurrent invocations of `refreshToken`. In this case the first one will reset the token future on line 231. Subsequent calls will take the else branch and end up chaining off the same future. This is correct.
But consider in this case how many times line 239 is executed. It will be run once for each call waiting on the future. Which could be many times. However as this is doing a compare an set, it only makes sense to call this once. Similarly the log line on 240, it only makes sense for it to happen once, when the token is refreshed, not once for every task that was blocked on the future. The only other line is 241 which is mapping the void to the token itself. This is needlessly indirect. The future returned from recreateToken can just be made to return the token buy adding a `return token;` on 247.

ravisharda: Since the `whenComplete` takes a `BiConsumer` as input, I cannot have it return the `DelegationToken` object. I can do something like the following instead, but it'll change the existing logic a great deal (the implications of which needs to be studied) in `refreshToken() and elsewhere in the class`. 

```
private CompletableFuture<String> recreateToken() {
        return controllerClient.getOrRefreshDelegationTokenFor(scopeName, streamName)
                .whenComplete((token, ex) -> this.delegationToken.set(new DelegationToken(token, extractExpirationTime(token))));
    }
```
Regarding your other points: 
* The exit trace log is actually intended for every call to `refreshToken`, whether it returns a new delegation token created by the current thread or one created by another thread. 
* Similarly, the block below the if/else condition will return a new delegation token whether it was created by the current thread or created by another thread. 

In any case, I have an [enhancement issue](https://github.com/pravega/pravega/issues/4358) to look at optimizations in this code. We can look at improvements holistically as part of that issue. What do you say?

PR #4373, opened 2019-11-15
Issue 4369: (SegmentStore) Fix memory leak with Future Reads

### LINKED ISSUE #4369: FutureReadResultEntry cancellation does not clean it up from the ReadIndex
**Problem description**
1. We invoke `ReadIndex.read(...)` which gives us `StreamSegmentReadResult` `rr`.
2. We invoke `rr.next()` which gives us a `FutureReadResultEntry` `fre` (it could return other types, but we only care about this one now).
    - As a side effect, the `ReadIndex` has registered the `FutureReadResultEntry` internally so that when we get a new append, it is automatically completed.
3. We invoke `rr.close()`, which has an effect of cancelling `fre`. 

At this point the caller code releases all references to both `rr` and `fre`. However, the `ReadIndex` still has a pointer to `fre` internally which would only be cleaned up upon an append. If no such append arrives, we end up keeping this reference for a long time, which can lead to memory leaks and HeapOOM.

**Problem location**
ReadIndex/StreamSegmentReadResult.

Desired outcome is to clean up `FutureReadResultEntries` from the `ReadIndex` when they are cancelled. This should be done in a way that won't affect performance or have other side effects (`ReadIndex` is on the hot path for both reads and writes).

andreipaduroiu:
**Change log description**  
- Fixed a memory leak that can be triggered by abandoning Future Reads without any segment write activity.

**Purpose of the change**  
Fixes #4369.

**What the code does**  
- Every `FutureReadResultEntry` gets assigned a callback to invoke when it is completed (successfully or not). This is assigned when being registered into the `FutureReadResultEntryCollection` (of the Read Index) and unassigned when being polled from that collection.
    - This callback, when invoked, will cause that entry to be unregistered from the collection.
    - The callback should be unregistered when polled out of the collection because the next step that the Read Index would take is to complete it, which means it would have caused the collection to unnecessarily search for it.
- This has been wired up into the `StreamSegmentReadResult` to be invoked when the result itself is closed and thus abandoning any `FutureReadResultEntry` that it might be sitting on. 

**How to verify it**  
All unit tests must pass. New unit tests added.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadResult.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadResult.java
@@ -184,7 +184,11 @@ class StreamSegmentReadResult implements ReadResult {
                 this.canRead = false;
             } else {
                 // After the previous entry is done, update the consumedLength value.
-                entry.setCompletionCallback(length -> this.consumedLength += length);
+                entry.setCompletionCallback(length -> {
+                    synchronized (StreamSegmentReadResult.this) {
+                        this.consumedLength += length;
+                    }
+                });
                 this.lastEntry = entry;
 
                 // Check, again, if we are closed. It is possible that this Result was closed after the last check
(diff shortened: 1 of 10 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadResult.java
```diff
@@ -29,7 +28,7 @@
     private final long streamSegmentStartOffset;
     private final int maxResultLength;
     private final NextEntrySupplier getNextItem;
-    private CompletableFuture<ReadResultEntryContents> lastEntryFuture;
+    private CompletableReadResultEntry lastEntry;
```
tkaitchuck: What's going on with thread safety here? It should at least be documented

andreipaduroiu: Fixed the whole class to be thread safe.

PR #4377, opened 2019-11-18
Issue 4374: Optimize the updateSegmentTags method

### LINKED ISSUE #4374: updateSegmentTags() method takes a lot of time for the small events write
**Problem description**
The `updateSegmentTags` method takes a lot of time and its implementation is not efficient.
screenshot

**Problem location**
io.pravega.shared.segment.StreamSegmentNameUtils#updateSegmentTags

Avoid using `String.split()` method, use the precompiled regexp instead.

akurilov:
Signed-off-by: Andrey Kurilov <andrey.kurilov@emc.com>

**Change log description**  
 Optimize the `updateSegmentTags()` method for better small events write performance.

**Purpose of the change**  
 Avoid using `String.split` in `updateSegmentTags`, use the precompiled regexp instead.

**What the code does**  
 Fixes #4374

**How to verify it**  
Basic testing shows that the updateSegmentTags takes twice less CPU time. The resulting small (1 byte) events rate is higher by 7%.

### DIFF
--- a/shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
+++ b/shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
@@ -9,6 +9,7 @@
  */
 package io.pravega.shared.segment;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import io.pravega.common.Exceptions;
@@ -89,6 +90,7 @@ public final class StreamSegmentNameUtils {
     /**
      * This is used to parse the segment name.
      */
+    @VisibleForTesting
     static final Pattern SEGMENT_TAGS_PATTERN =
         Pattern.compile("((([\\w\\-\\\\.]+)/)?(([\\w\\-\\\\.]+)/))?(\\w+)(" + Pattern.quote(EPOCH_DELIMITER) + "(\\d+))?");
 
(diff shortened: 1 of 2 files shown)


RaulGracia:
Thanks @akurilov. I'm curious, what is the performance difference between the 2 implementation of `updateSegmentTags`? Could you execute a micro-benchmark and post the results here so we can compare them?

andreipaduroiu:
Even though the pre-compiled regex approach seems faster, this could be sped up even more by a linear search in the string (each time you locate the next token and extract the data between the current position and that). That should give you O(n) perf without maybe up to a 2x multiplier, which is much better than a regex can do.

akurilov:
> Thanks @akurilov. I'm curious, what is the performance difference between the 2 implementation of `updateSegmentTags`? Could you execute a micro-benchmark and post the results here so we can compare them?

@RaulGracia 
The benchmark was to parse 100 million random inputs generated as
```java
System.nanoTime() + "/" + System.nanoTime() + "/" + System.nanoTime() + ".#epoch." +
                    System.nanoTime()
```
Results:
- Baseline (split): 200.917 s
- Fixed (precompiled regex): 96.087 s

shrids:
`MetricsTags.segmentTags()` could have a similar optimization.

shared/protocol/src/main/java/io/pravega/shared/segment/StreamSegmentNameUtils.java
```diff
@@ -84,6 +86,13 @@
     private static final String TABLES = "_tables";
     private static final String MARK = "_MARK";
 
+    /**
+     * This is used to parse the segment name.
+     */
+    static final Pattern SEGMENT_TAGS_PATTERN =
```
andreipaduroiu: Please add `@VisibleForTesting` annotation since the only reason for this not being private is because it's used in unit tests.

PR #4378, opened 2019-11-19
Issue 4357: Call noteWriterMark once per writer after committing a batch of transactions

### LINKED ISSUE #4357: Chances of getting WriteConflictException (retryable) during transaction commits if multiple transactions are from same writer
**Problem description**
After completing a batch of transaction commits, we note time corresponding to each transaction.
This has two problems -
1. we make multiple metadata update calls for same writer if it has multiple transactions in the batch 
2. since all these calls are made in parallel, so there is a good chance that some of them may fail because of write conflicts and are retried. Which is unnecessary. 

**Problem location**
Controller.StreamMetadataStore

Collate all transactions from same writer and make a single call to the store per writer.

shiveshr:

**Change log description**  
* Call noteWriterMark once per writer after committing a batch of transactions. 

**Purpose of the change**  
Fixes #4357 

**What the code does**  
We run commit workflow for multiple tranasctions in a single batch at the end of which we used to call noteWriterMark corresponding to each transaction. 
This is suboptimal for two reasons - 
1. we make mutliple calls for the same writer overwriting the just written previous value
2. since all the marks for same writer was updated concurrently, there was a possibility of WriteConflict which would be unnecessarily retried. 

The fix we have performed is as follows:
1. group transactions by writer ids
2. take the latest transaction for each writer and use the time there to note writer's mark. 

**How to verify it**  
Unit test added.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
@@ -1223,24 +1223,37 @@ public abstract class PersistentStreamBase implements Stream {
      * transaction record for which a writer with time and position information is available. 
      */
     CompletableFuture<Void> generateMarksForTransactions(CommittingTransactionsRecord committingTransactionsRecord) {
-        return Futures.allOf(committingTransactionsRecord.getTransactionsToCommit().stream().map(txId -> {
+        // Ignore data not found exceptions. DataNotFound Exceptions can be thrown because transaction record no longer 
+        // exists and this is an idempotent case. DataNotFound can also be thrown because writer's mark was deleted 
+        // as we attempted to update an existing record. Note: Delete can be triggered by writer explicitly calling
+        // removeWriter api. 
+        val getTransactionsFuture = Futures.allOfWithResults(committingTransactionsRecord.getTransactionsToCommit().stream().map(txId -> {
             int epoch = RecordHelper.getTransactionEpoch(txId);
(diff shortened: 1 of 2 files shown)


controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1223,24 +1223,37 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
...
+                                    x.getObject().getCommitTime() >= 0L && !x.getObject().getCommitOffsets().isEmpty())
+                                                     .collect(Collectors.groupingBy(x -> x.getObject().getWriterId()));
+
+                    // For each writerId we will take the transaction with the time and position pair (which is to take
+                    // max of all transactions for the said writer). 
+                    // Note: if multiple transactions from same writer have same time, we will take any one arbitrarily and
+                    // use its position for watermarks. Other positions and times would be ignored. 
+                    val noteTimeFutures = groupedByWriters.entrySet().stream().map(groupEntry -> {
+                        ActiveTxnRecord latest = groupEntry.getValue().stream().max(Comparator.comparingLong(x -> x.getObject().getCommitTime()))
+                                                           .get().getObject();
```
shrids: `.max()` returns an Optional.empty() if there are no elements in the stream and doing a `.get()` would cause a `NoSuchElementException`. (This is possible if `txnRecords ` is a list of nulls say for instance...)

shiveshr: Actually this method never gets empty transactions. So max will always return a value

Issue #3821, opened 2019-05-22
Create describeStream REST call in Controller


RaulGracia:
**Problem description**
Currently, Pravega CLI offers a `describeStream` command that outputs relevant information about a Stream (e.g., state, open transactions, etc.). However, at the moment this command uses low level Controller abstractions to manually get this information from the metadata store, which differs from the rest of controller-related commands that resort to the REST API. We may want to unify the implementation of Pravega CLI calls to the Controller via REST API, which reduces implementation/configuration complexity on the CLI and gives full control of the actual implementation to the server.

**Problem location**
Controller REST API.

**Suggestions for an improvement**
Create new `describeStream` REST API call.

PR #4380, opened 2019-11-19
Issue 4020: Expose additional apis on StreamMetadataStore interface for Pravega Cli

### LINKED ISSUE #4020: Expose additional stream metadata records for internal records on StreamMetadataStore interface

shiveshr:
**Change log description**  
To expose additional apis on StreamMetadataStore interface to get all relevant metadata

**Purpose of the change**  
Fixes #4020 

**What the code does**  
There are three metadata record types that are part of internal metadata schema but are not exposed over StreamMetadataStore interface. 
However, for purposes of pravega cli, we need to expose all such records so that they can be accessed by the admin tool. 

**How to verify it**  
Unit test added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
@@ -25,7 +25,9 @@ import io.pravega.controller.store.stream.records.ActiveTxnRecord;
 import io.pravega.controller.store.stream.records.CommittingTransactionsRecord;
 import io.pravega.controller.store.stream.records.EpochRecord;
 import io.pravega.controller.store.stream.records.EpochTransitionRecord;
+import io.pravega.controller.store.stream.records.HistoryTimeSeries;
 import io.pravega.controller.store.stream.records.RetentionSet;
+import io.pravega.controller.store.stream.records.SealedSegmentsMapShard;
 import io.pravega.controller.store.stream.records.StreamConfigurationRecord;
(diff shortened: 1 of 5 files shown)


RaulGracia:
@shiveshr while it is good to have this information available, in issue #3821 I proposed to add a new API that summarizes all the relevant information for a stream (e.g., `describStream`). Currently, the Pravega CLI does that by querying all the metadata available for a Stream: https://github.com/pravega/pravega-tools/blob/master/pravega-cli/src/main/java/io/pravega/tools/pravegacli/commands/controller/ControllerDescribeStreamCommand.java
But, maybe a call that provides a summary of the relevant information for a Stream may be really useful, as it would simplify the Pravega CLI.

shiveshr:
@RaulGracia I agree that a describe stream will be a useful API. However, that is orthogonal to this change. 
The purpose of this change is to allow pravega CLI to directly make use of stream metadata store interface to read the metadata directly from the cluster without querying a controller service for the same. 
This enables us to look at internal metadata schema that may not be exposed over the view of the stream on our public endpoints like rest or grpc.

Issue #4386, opened 2019-11-20
Unify NameUtils classes


tkaitchuck:
**Problem description**
There are currently two utility classes for managing names of segments and streams. 

**Problem location**
pravega-shared-protocol

**Suggestions for an improvement**
These should be combined into one class.

RaulGracia:
Closed via #4387.

PR #4387, opened 2019-11-20
Issue 4385: Limit characters that can appear in a readerId or a writerId

### LINKED ISSUE #4385: Restrict characters allowed in ReaderIds and WriterIds
**Problem description**
When the controller handles ReaderIds and WriterIds these are often sent to ZK which interprets certain special characters, notably "/". 

**Problem location**
Client and Controller

We should limit the allowed characters in ReaderIds and WriterIds the same way we do for streams and segments to prevent problems.

tkaitchuck:
**Change log description**  
* Unifies NameUtils into a single utility class
* Limits characters that can appear in a readerId or a writerId to the same set as are allowed in stream names.

**Purpose of the change**  
Fixes #4385, #4386

**What the code does**  
When a reader or writer is created it is checked against a regex for allowed characters in the readerId or writerId. (The same one as for streams is used.) 
(The remainder of the code changes are pure refactoring and non-functional)


tkaitchuck:
Merged. 
The only classes with changes of any substance are NameUtils and StreamSegmentNameUtils

tkaitchuck:
Merged.
The only classes with changes of any substance are NameUtils and StreamSegmentNameUtils

shiveshr:
LGTM however, there are conflicts with master causing build failure that need to be fixed.

PR #4391, opened 2019-11-23
Issue 4390: Ensure retries by client when delegation token expires

### LINKED ISSUE #4390: TokenExpiryException is propagated to the client application in some cases
**Problem description**
First, a bit of background. PR #4215 implemented automatic token renewal, which transparently renews the delegation token when the delegation token is nearing expiry. 

However, there are cases where the token might not be nearing expiry at the time when that condition is being checked, but it actually expires by the time it reaches the segment store. It was intended that under such conditions, the client's automatic retry mechanism shall ensure that the client-to-server call is retried and upon the retry, the client-side logic for checking if the delegation token is nearing expiry will automatically fetch a new token. 

The problem is that we have encountered two places where the client doesn't actually retry the request and throws an exception to the client application when encountering a token expiry error from the server (the segment store). 

**Problem location**
`SegmentOutputStreamImpl` and `SegmentMetadataClientImpl`

ravisharda:
**Change log description**  
Ensure that the client retries the request to segment store if the segment store throws an error indicating token expiry. Doing so will ensure that a new delegation token will be obtained after reauthentication and reauthorization with the Controller and used in the retried request. 

**Purpose of the change**  
Resolves #4390. 

**What the code does**  
- Modifies `io.pravega.client.segment.impl.SegmentMetadataClientImpl` class to throw `ConnectionFailedException` upon receiving a response from the segment store indicating delegation token expiry. This will ensure that the request is retried. 
- Modifies `io.pravega.client.segment.impl.SegmentOutputStreamImpl` class to not release the connection upon encountering token expiry indicator from the segment store. 
- Makes `TokenException` abstract in order to prevent any code from throwing that exception. It was originally intended to be a base class for all token related exceptions, and since it wasn't abstract some of the code - including what was initially developed for this PR - ended up throwing it (instead of the subclasses that represent specific conditions) inadvertently. 

**How to verify it**  
All existing tests should verify. 

I also used this test suite temporarily to quickly verify changes. 

```java
package io.pravega.test.integration;

import io.pravega.client.security.auth.DelegationTokenProviderFactoryTest;
import io.pravega.client.security.auth.JwtBodyTest;
import io.pravega.client.security.auth.JwtTokenProviderImplTest;
import io.pravega.client.segment.impl.AsyncSegmentInputStreamTest;
import io.pravega.client.segment.impl.ConditionalOutputStreamTest;
import io.pravega.client.segment.impl.SegmentMetadataClientTest;
import io.pravega.client.segment.impl.SegmentOutputStreamTest;
import io.pravega.client.state.impl.RevisionedStreamClientTest;
import io.pravega.client.stream.impl.EventStreamReaderTest;
import io.pravega.client.stream.impl.SegmentSelectorTest;
import io.pravega.segmentstore.server.host.handler.PravegaRequestProcessorAuthFailedTest;
import io.pravega.shared.protocol.netty.WireCommandsTest;

@RunWith(Suite.class)
@Suite.SuiteClasses({
        JwtTokenProviderImplTest.class,
        DelegationTokenProviderFactoryTest.class,
        JwtBodyTest.class,
        EventStreamReaderTest.class,
        RevisionedStreamClientTest.class,
        AppendTest.class,
        AppendReconnectTest.class,
        ReadTest.class,
        SegmentSelectorTest.class,
        ConditionalOutputStreamTest.class,
        PravegaRequestProcessorAuthFailedTest.class,
        WireCommandsTest.class,
        AsyncSegmentInputStreamTest.class,
        SegmentMetadataClientTest.class,
        SegmentOutputStreamTest.class,
        DelegationTokenTest.class,
})

public class DelegationTokenTestSuite {
}
```

### DIFF
--- a/client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
@@ -10,6 +10,7 @@
 package io.pravega.client.segment.impl;
 
 import com.google.common.annotations.VisibleForTesting;
+import io.pravega.auth.InvalidTokenException;
 import io.pravega.auth.TokenException;
 import io.pravega.auth.TokenExpiredException;
 import io.pravega.client.netty.impl.ConnectionFactory;
@@ -127,7 +128,7 @@ class SegmentMetadataClientImpl implements SegmentMetadataClient {
                 throw new ConnectionFailedException(new TokenExpiredException(authTokenCheckReply.toString()));
             } else {
                 log.info("Delegation token invalid");
-                throw new TokenException(authTokenCheckReply.toString());
+                throw new InvalidTokenException(authTokenCheckReply.toString());
             }
         } else {
             throw new ConnectionFailedException("Unexpected reply of " + reply + " when expecting a "
--- a/client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
@@ -9,6 +9,7 @@
  */
 package io.pravega.client.segment.impl;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.pravega.auth.TokenException;
 import io.pravega.auth.TokenExpiredException;
 import io.pravega.client.netty.impl.ConnectionFactory;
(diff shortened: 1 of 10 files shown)


ravisharda:
> The changes look good, could you add test cases for the token expiration scenario.

Sure @shrids, I've added some tests as suggested.

shrids:
@ravisharda should we also modify `io.pravega.segmentstore.server.host.delegationtoken.DelegationTokenVerifier#verifyToken` now that TokenException is abstract. ?

ravisharda:
> @ravisharda should we also modify `io.pravega.segmentstore.server.host.delegationtoken.DelegationTokenVerifier#verifyToken` now that TokenException is abstract. ?

@shrids I wanted to leave the possibility open for throwing a new type of exception (that extends TokenException) in case we implement another kind of token verifier. What do you say?

PR #4394, opened 2019-11-24
Issue 4531: Use new BookKeeper API and Upgrade to BookKeeper 4.9.2

### LINKED ISSUE #4531: Use new BookKeeper API and Upgrade to BookKeeper 4.9.2
**Problem description**
Since version 4.8 BookKeeper introduced a new client API (org.apache.bookkeeper.client.api).

Pravega is still using old API, LedgerHandler, that is going to be deprecated officially in next upcoming release (4.11).

Other nodes:
The new API is totally based on CompletableFuture facility of the JDK.
The new API allows access to the internal refcounted Netty buffer used by BK client.

- Switch from using LedgerHandle to WriteHandle and ReadHandle.
- Upgrade a newer version of BookKeeper, 4.9.2 or 4.10.0

eolivelli:
**Change log description**  
BookKeeper 4.8 introduced a new API (org.apache.bookkeeper.client.api).
The old API, LedgerHandler, is going to be deprecated officially in next upcoming release (4.11).
The new API is based on CompletableFuture facility of the JDK.
The new API allows access to the internal refcounted Netty buffer used by BK client.

**Purpose of the change**  
Fixes #4531.

Switch from using LedgerHandle to WriteHandle and ReadHandle.
I am changing BK version to 4.9.2.

**What the code does**  
Upgrade BookKeeper to 4.9.2 in Java code (not docker images)
Replace  usages of LedgerHandle with WriteHandle and ReadHandle.
Use the new CompletableFuture based API.

**How to verify it**  
Tests should run without errors.


RaulGracia:
Thanks @eolivelli for the contribution. I realized that the build is not passing, and it would be great to have it green before proceeding to review the PR. Thanks!

eolivelli:
I guess the build is not passing due to other needed changes to upgrade to 4.9.x.

Do you have plans?
IIRC I saw messages on BK mailing list about little problems with the upgrade.

I can help with the upgrade.
If you want I can continue this work and make the build pass

eolivelli:
@RaulGracia tests are failing because in segmentstore:impl:test we are stopping and restarting one single bookie and it seems that with BK 4.8.x this way of restating the bookie does not work anymore.

I will investigate more

eolivelli:
@RaulGracia is there any way to run a single method from o.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLogTests\$RegularBookKeeperLogTests

Currently I am running:
./gradlew :segmentstore:storage:impl:test --tests io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLogTests\$RegularBookKeeperLogTests --info

but it is very uncomfortable, I have to \@Ignore every test I don't want to run

RaulGracia:
@eolivelli thanks for the work. The times I needed to run a specific test like this, I did that either as you have done or directly via the IDE debugger. Regarding the full upgrade, I think that is an interesting option and actually the Bookkeeper version we are using (4.7.3) is becoming a bit old. But let's see the input from others: @fpj @andreipaduroiu @tkaitchuck, what is your take on upgrading BK to `4.9.2` on Pravega `0.7` (or later)?

tkaitchuck:
@RaulGracia Yes. Lets do this as soon as possible.

andreipaduroiu:
Agreed. Let's get the build to pass and a system test run complete successfully before moving on too.

eolivelli:
Good.
I will complete the work.

I have to finish the reader part and some other little stuff.

The real time consuming task is to understand why sometimes after restarting the bookie the BK client doesn't see it up and so writes/create ledger fail.

build.gradle
```diff
@@ -321,6 +321,8 @@ project('segmentstore:storage:impl') {
         compile group: 'org.rocksdb', name: 'rocksdbjni', version: rocksdbjniVersion
         testCompile project(':test:testcommon')
         testCompile project(path:':segmentstore:storage', configuration:'testRuntime')
+        testCompile group: 'io.netty', name: 'netty-transport-native-epoll', version: nettyVersion, classifier: 'linux-x86_64'
```
fpj: Why do we need to add these test compile dependencies?

gradle.properties
```diff
@@ -16,7 +16,7 @@ apacheCommonsCompressVersion=1.18
 apacheCuratorVersion=4.0.1
 apacheZookeeperVersion=3.5.5
 checkstyleToolVersion=8.23
-bookKeeperVersion=4.7.3
+bookKeeperVersion=4.9.2
```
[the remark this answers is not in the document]
fpj: I'm not sure what you mean with this comment, @eolivelli .

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
```diff
@@ -451,7 +451,19 @@ private boolean executeWrites(List<Write> toExecute) {
...
+                        .whenComplete((entryId, error) -> {
+                            int rc = BKException.Code.OK;
+                            if (error != null) {
+                                if (error instanceof BKException) {
+                                    rc = ((BKException) error).getCode();
+                                } else {
+                                    rc = BKException.Code.WriteException;
+                                }
+                            }
+                            addCallback(rc, w.getWriteLedger().ledger, entryId, w);
```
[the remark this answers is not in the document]
fpj: Could you create an issue in the case we are not fixing it in this PR?

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -42,18 +48,19 @@
...
+                            .withEnsembleSize(config.getBkEnsembleSize())
+                            .withWriteQuorumSize(config.getBkWriteQuorumSize())
+                            .withAckQuorumSize(config.getBkAckQuorumSize())
+                            .withDigestType(LEDGER_DIGEST_TYPE)
+                            .withPassword(config.getBKPassword())
+                            .execute()));
+        } catch (BKNotEnoughBookiesException bkEx) {
             throw new DataLogNotAvailableException("Unable to create new BookKeeper Ledger.", bkEx);
-        } catch (BKException bkEx) {
+        } catch (Exception bkEx) {
```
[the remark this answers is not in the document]
fpj: Could you elaborate?

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -67,11 +74,17 @@ static LedgerHandle create(BookKeeper bookKeeper, BookKeeperConfig config) throw
      * @return A LedgerHandle for the newly opened ledger.
      * @throws DurableDataLogException If an exception occurred. The causing exception is wrapped inside it.
      */
-    static LedgerHandle openFence(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
+    static ReadHandle openFence(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
         try {
             return Exceptions.handleInterruptedCall(
-                    () -> bookKeeper.openLedger(ledgerId, LEDGER_DIGEST_TYPE, config.getBKPassword()));
-        } catch (BKException bkEx) {
+                    () -> FutureUtils.result(bookKeeper
```
fpj: Do we still need to handle interrupted here?

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -85,11 +98,17 @@ static LedgerHandle openFence(long ledgerId, BookKeeper bookKeeper, BookKeeperCo
      * @return A LedgerHandle for the newly opened ledger.
      * @throws DurableDataLogException If an exception occurred. The causing exception is wrapped inside it.
      */
-    static LedgerHandle openRead(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
+    static ReadHandle openRead(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
         try {
             return Exceptions.handleInterruptedCall(
-                    () -> bookKeeper.openLedgerNoRecovery(ledgerId, LEDGER_DIGEST_TYPE, config.getBKPassword()));
-        } catch (BKException bkEx) {
+                    () -> FutureUtils.result(bookKeeper
```
fpj: Same question here as above.

fpj:
Does anyone know why there is no `4.9.2` take for bookkeeper on docker hub?

https://hub.docker.com/r/apache/bookkeeper/tags

gradle.properties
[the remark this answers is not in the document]
eolivelli: @fpj sorry for late reply
I was saying that the upgrade of BK should be performed in a separate patch.
In this patch I am doing two things: upgrade BK and move to the new API.
We could upgrade BK and then move to the new API in a follow up work

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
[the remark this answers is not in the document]
eolivelli: I don't want to leave 'catch(Exception)'
I will fix it

build.gradle
eolivelli: I thought it was a problem of not having 'epoll' available, I can drop this line if it is not needed (once I see all tests passing)

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
eolivelli: yes I will do

RaulGracia:
@eolivelli it looks like there is a problem with the DCO plugin. We will need to get it passing to merge this PR eventually.

eolivelli:
@RaulGracia I will sign my commit once the build passes

eolivelli:
Once build is green with the upgrade of BK I have to finalize this patch:
- fix BookKeeper reader API usage
- fix exception handling
- use new API for booting BooKeeper client

tkaitchuck:
I set DCO to pass. That should work for now. (Feel free to push updated comments though)

eolivelli:
> I set DCO to pass. That should work for now. (Feel free to push updated comments though)

thanks @tkaitchuck

eolivelli:
Please do not force update this branch using github button, I have a bunch of changes locally.

eolivelli:
@RaulGracia it seems to me that you can run a single test method using this syntax:

./gradlew :segmentstore:storage:impl:test --tests io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLogTests\$RegularBookKeeperLogTests.testAutoCloseOnBookieFailure --info

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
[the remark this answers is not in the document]
eolivelli: I have fixed this method. no need to open an issue.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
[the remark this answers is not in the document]
eolivelli: I realized I can't do better. I am leaving catch (Exception).

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
eolivelli: same answer as above, we are bloking for the operation to complete and we should handle InterruptedException

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -141,10 +143,10 @@ private void openNextLedger(LedgerAddress address) throws DurableDataLogExceptio
 
         ReadLedger previousLedger;
         try {
-            val reader = Exceptions.handleInterruptedCall(
-                    () -> ledger.readEntries(address.getEntryId(), lastEntryId));
+            LedgerEntries reader = Exceptions.handleInterruptedCall(
```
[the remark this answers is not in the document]
eolivelli: done

build.gradle
eolivelli: it is not needed. dropped

RaulGracia:
Thanks @eolivelli, but I see the Travis build failing due to a compilation error:
```
> Task :segmentstore:storage:impl:compileJava FAILED
/home/travis/build/pravega/pravega/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java:174: error: exception IOException is never thrown in body of corresponding try statement
        @SneakyThrows(IOException.class)
        ^
1 error
FAILURE: Build failed with an exception.
```

eolivelli:
@RaulGracia I have fixed the build problem.
All checks are green.

The docker image for BookKeeper 4.9.2 is ready as well

gradle.properties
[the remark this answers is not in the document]
andreipaduroiu: PR #4420. That should go in before this one.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
```diff
@@ -636,32 +638,41 @@ private void handleWriteException(Throwable ex) {
...
+                        // Write-related failure or current Ledger closed. This is a retryable exception.
+                        ex = new WriteFailureException("Unable to write to active Ledger.", ex);
+                        break;
+                    case Code.ClientClosedException:
+                        // The BookKeeper client was closed externally. We cannot restart it here. We should close.
+                        ex = new ObjectClosedException(this, ex);
+                        break;
+                    default:
+                        // All the other kind of exceptions go in the same bucket.
+                        ex = new DurableDataLogException("General exception while accessing BookKeeper.", ex);
```
andreipaduroiu: We should find a way to unify this line with 647. 
Maybe we can move this while `switch` into its own method which will return `null` if `ex` is not `BKException` of if we get in `default`. Then, back in this method, if the new method returns null, we just convert `ex` to a `DurableDataLogException`.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
andreipaduroiu: Agreed. We should file an issue to convert the methods in this class to use `CompletableFuture` now that the underlying APIs support that. We shouldn't do that as part of this effort as there'd be a lot of upstream code to update.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -168,11 +169,10 @@ private void openNextLedger(LedgerAddress address) throws DurableDataLogExceptio
         @Getter
         private final LedgerAddress address;
 
-        @SneakyThrows(IOException.class)
-        ReadItem(LedgerEntry entry, LedgerMetadata ledgerMetadata) {
-            this.address = new LedgerAddress(ledgerMetadata, entry.getEntryId());
-            this.payload = entry.getEntryInputStream();
-            this.length = this.payload.available();
+        ReadItem(long entryId, byte[] entry, LedgerMetadata ledgerMetadata) {
```
andreipaduroiu: This will allocate a new Java byte array from the internal `ByteBuffer`. The reason it used `getEntryInputStream` was precisely to avoid that.

Let's figure out a way to get an `InputStream` directly from the underlying buffer. You can use `ByteBufInputStream`, but I am sad to see this API gone.

andreipaduroiu: Ok I see below the reason why you did it this way. Then we should take care to release this buffer when `getNext` is invoked (for the previously returned item) or when we close the `LogReader` class.

In any case I want to avoid having to allocate too many Java byte arrays, which will make the recovery process unnecessarily slow.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -185,14 +185,53 @@ public String toString() {
...
+         * Holder for the list of entries.
+         * A null value means that this ledger is empty.
+         */
+        final Iterator<EntryHolder> entries;
+
+        public ReadLedger(LedgerMetadata metadata, ReadHandle handle, LedgerEntries ledgerEntries) {
+            this.metadata = metadata;
+            this.handle = handle;
+            if (ledgerEntries != null) {
+                List<EntryHolder> entriesHolder = new ArrayList<>();
```
andreipaduroiu: Why do you buffer all the entries from the ledger? This, in combination with copying all the data to Java byte arrays and a max ledger size of 1GB has the potential for HeapOOMs.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
eolivelli: @andreipaduroio
I am performing the copy because I am not sure about the contract of ReadItem.

Will getInputStream be called once and closed once?

We can release the buffer when the caller Invokes InputStream#close and close all unconsumed entries in LogReader#close.

If I understand correctly you are telling me that this approach is doable and we won't introduce memory leaks.

I will try to update the patch soon.

I have to check more but as the prev BK API did not provide a close() method the copy was performed inside BK

eolivelli: @andreipaduroiu done

tkaitchuck:
@andreipaduroiu @eolivelli 
Some tests are failing with NPEs:
```
io.pravega.segmentstore.server.host.FileSystemIntegrationTest > testEndToEnd FAILED
    java.util.concurrent.CompletionException: java.lang.NullPointerException
        Caused by:
io.pravega.segmentstore.server.host.FileSystemIntegrationTest > testEndToEndWithFencing FAILED
    java.util.concurrent.CompletionException: java.lang.NullPointerException
        Caused by:
io.pravega.segmentstore.server.host.ServiceStarterTest > testCuratorClientCreation PASSED
io.pravega.segmentstore.server.host.ExtendedS3IntegrationTest > testEndToEnd FAILED
    java.util.concurrent.CompletionException: java.lang.NullPointerException
        Caused by:
io.pravega.segmentstore.server.host.ExtendedS3IntegrationTest > testEndToEndWithFencing FAILED
    java.util.concurrent.CompletionException: java.lang.NullPointerException
        Caused by:
```

eolivelli:
@tkaitchuck  
Sorry, I did push my local changes but the work was not complete.
I will push again when I am ready

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
eolivelli: @andreipaduroiu done. I have taken another approch, please see if it is looks good to your

eolivelli:
@andreipaduroiu I have addressed your comments.
Please take a look

eolivelli:
CI is green

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
```diff
@@ -477,7 +478,11 @@ private boolean executeWrites(List<Write> toExecute) {
                 }
 
                 // Invoke the BookKeeper write.
-                w.getWriteLedger().ledger.asyncAddEntry(w.data.array(), w.data.arrayOffset(), w.data.getLength(), this::addCallback, w);
+                final WriteHandle handle = w.getWriteLedger().ledger;
```
andreipaduroiu: You don't actually need to store this in a variable as you're only using it once. You can just invoke `appendAsync` from `w.getWriteLedger().ledger` directly.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -42,18 +46,19 @@
      *                                      inside it.
      * @throws DurableDataLogException      If another exception occurred. The causing exception is wrapped inside it.
      */
-    static LedgerHandle create(BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
+    static WriteHandle create(BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
```
andreipaduroiu: Update `@return` tag in javadoc to reflect the changed return type.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -67,11 +72,17 @@ static LedgerHandle create(BookKeeper bookKeeper, BookKeeperConfig config) throw
      * @return A LedgerHandle for the newly opened ledger.
      * @throws DurableDataLogException If an exception occurred. The causing exception is wrapped inside it.
      */
-    static LedgerHandle openFence(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
+    static ReadHandle openFence(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
```
andreipaduroiu: update `@return` javadoc

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
andreipaduroiu: #4433

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -85,11 +96,17 @@ static LedgerHandle openFence(long ledgerId, BookKeeper bookKeeper, BookKeeperCo
      * @return A LedgerHandle for the newly opened ledger.
      * @throws DurableDataLogException If an exception occurred. The causing exception is wrapped inside it.
      */
-    static LedgerHandle openRead(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
+    static ReadHandle openRead(long ledgerId, BookKeeper bookKeeper, BookKeeperConfig config) throws DurableDataLogException {
```
andreipaduroiu: javadoc

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -141,11 +144,12 @@ private void openNextLedger(LedgerAddress address) throws DurableDataLogExceptio
...
         try {
-            val reader = Exceptions.handleInterruptedCall(
-                    () -> ledger.readEntries(address.getEntryId(), lastEntryId));
+            LedgerEntries entries = Exceptions.handleInterruptedCall(
+                    () -> ledger.read(address.getEntryId(), lastEntryId));
             previousLedger = this.currentLedger;
-            this.currentLedger = new ReadLedger(metadata, ledger, reader);
+            this.currentLedger = new ReadLedger(metadata, ledger, entries);
             if (previousLedger != null) {
+                previousLedger.release();
```
andreipaduroiu: Here you are invoking `release` before `Ledgers.close`, but everywhere else you did it after. Is this a problem?

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -185,14 +188,72 @@ public String toString() {
...
+         */
+        final Iterator<EntryHolder> entryIterator;
+        final AtomicBoolean closed = new AtomicBoolean(false);
+
+        public ReadLedger(LedgerMetadata metadata, ReadHandle handle, LedgerEntries ledgerEntries) {
+            this.metadata = metadata;
+            this.handle = handle;
+            this.ledgerEntries = ledgerEntries;
+            if (ledgerEntries != null) {
+                List<EntryHolder> entries = new ArrayList<>();
```
andreipaduroiu: I still don't understand why we need to loop through all the entries here, convert them to `EntryHolder` instances, then in `nextItem` we just loop through this array. We are still holding a pointer to `ledgerEntries`.

Can't we just have `nextItem` use the `ledgerEntries.iterator()` and return from there? That way we won't have a need for `EntryHolder` nor will we need to double-loop through the entries.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
eolivelli: It shouldn't be a problem, as *release* is only about releasing Netty memory.
I will move this line after *close* in order to have a consistent pattern.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
eolivelli: Good idea!
Current code is the result of the first implementation with a the memcopy, now this nee ArrayList is a waste.
I will update the patch

eolivelli: Good idea!
Current code is the result of the first implementation with a the memcopy, now this nee ArrayList is a waste.
I will update the patch

eolivelli:
@andreipaduroiu Thanks for your suggestions.
I will apply the new changes and clean up javadocs

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
eolivelli: okay. There is no difference at runtime.
It is a matter of taste.
changed

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
eolivelli: done

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
eolivelli: done

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
eolivelli: done

eolivelli:
I have to fix checkstyle issues, sorry

eolivelli:
@andreipaduroiu patch is ready now, CI is green

andreipaduroiu:
All other PRs that were blocking this have been merged. @fpj and @RaulGracia please take another look at this PR and let's get it in.

andreipaduroiu:
@eolivelli We are ready to move on with this PR. Would you mind updating your branch and resolving the conflicts we have on it? Thx.

eolivelli:
Sure. I will do as soon as possible (not sure today)

eolivelli:
Travis failed with this error, not related to this patch:
io.pravega.test.integration.ControllerMetricsTest > streamMetricsTest FAILED

    java.lang.IllegalStateException: The requested reader: 3 cannot be added to the group because it is already in the group. Perhaps close() was not called?

        at io.pravega.client.stream.impl.ReaderGroupStateManager.initializeReader(ReaderGroupStateManager.java:117)

        at io.pravega.client.stream.impl.ClientFactoryImpl.createReader(ClientFactoryImpl.java:180)

        at io.pravega.client.stream.impl.ClientFactoryImpl.createReader(ClientFactoryImpl.java:166)

        at io.pravega.test.integration.ReadWriteUtils.readEvents(ReadWriteUtils.java:40)

        at io.pravega.test.integration.ReadWriteUtils.readEvents(ReadWriteUtils.java:51)

        at io.pravega.test.integration.ControllerMetricsTest.streamMetricsTest(ControllerMetricsTest.java:185)

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -201,15 +207,40 @@ public String toString() {
...
+                } catch (DurableDataLogException bkEx) {
+                    log.error("Unable to close ReadHandle for Ledger {}.", handle.getId(), bkEx);
+                }
+            }
         }
 
         @SneakyThrows(BKException.class)
         private Iterator<LedgerEntry> readRange(long fromEntryId, long toEntryId) {
-            return Iterators.forEnumeration(Exceptions.handleInterruptedCall(() -> this.handle.readEntries(fromEntryId, toEntryId)));
+            if (currentLedgerEntries != null) {
```
eolivelli: Unfortunately we must keep a reference to the current **LedgerEntries** in order to release memory

eolivelli:
@andreipaduroiu I have merged this patch to master and signed my commit.
Tests are passing, the patch is ready for a final review.

@fpj @RaulGracia you already commented on this patch, please take a look if needed

RaulGracia:
Thanks a lot @eolivelli! I think that first we need to merge #4420 so we have the server side ready and then merge this one, if I'm not mistaken. PR #4420 is very close to be merged, so we expect to merge this one very soon.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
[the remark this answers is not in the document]
fpj: Is it an option that we change the call to `result` to one that takes an exception handler and provide one?

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -64,14 +69,20 @@ static LedgerHandle create(BookKeeper bookKeeper, BookKeeperConfig config) throw
...
-                    () -> bookKeeper.openLedger(ledgerId, LEDGER_DIGEST_TYPE, config.getBKPassword()));
-        } catch (BKException bkEx) {
+                    () -> FutureUtils.result(bookKeeper
+                            .newOpenLedgerOp()
+                    .withLedgerId(ledgerId)
+                    .withDigestType(LEDGER_DIGEST_TYPE)
+                    .withPassword(config.getBKPassword())
+                    .withRecovery(true)
+                    .execute()));
+        } catch (Exception bkEx) {
```
fpj: Same question here about changing the `result` call to one that takes an exception handler.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -82,14 +93,20 @@ static LedgerHandle openFence(long ledgerId, BookKeeper bookKeeper, BookKeeperCo
...
-                    () -> bookKeeper.openLedgerNoRecovery(ledgerId, LEDGER_DIGEST_TYPE, config.getBKPassword()));
-        } catch (BKException bkEx) {
+                    () -> FutureUtils.result(bookKeeper
+                            .newOpenLedgerOp()
+                    .withLedgerId(ledgerId)
+                    .withDigestType(LEDGER_DIGEST_TYPE)
+                    .withPassword(config.getBKPassword())
+                    .withRecovery(false)
+                    .execute()));
+        } catch (Exception bkEx) {
```
fpj: Same question about the exception handler.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -183,16 +178,27 @@ public String toString() {
...
     }
 
+    private static DurableDataLog.ReadItem wrapItem(LedgerEntry entry, LedgerMetadata metadata) {
+        ByteBuf content = entry.getEntryBuffer();
+        return new LogReader.ReadItem(entry.getEntryId(),
+               new ByteBufInputStream(content, false /*relaseOnClose*/),
+               content.readableBytes(), metadata);
+    }
+
+
```
fpj: Remove new line.

gradle.properties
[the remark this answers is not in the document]
eolivelli: @fpj I think the plan is to go with 4.9.2, as in #4519 
Honestly I think there is no value in staying in 4.8.x.
Hopefully we will release 4.11 in April and 4.8 will be far off the current active branches.
 
I can try to revert to 4.8 but the new API really started to be in good shape in 4.8 but it had some problem and since 4.9 it is the recommended one (in my personal opinion).

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -17,43 +17,48 @@
...
+import org.apache.bookkeeper.client.api.WriteHandle;
+import org.apache.bookkeeper.common.concurrent.FutureUtils;
 
 /**
  * General utilities pertaining to BookKeeper Ledgers.
  */
 @Slf4j
 final class Ledgers {
-    static final long NO_ENTRY_ID = LedgerHandle.INVALID_ENTRY_ID;
+    static final long NO_ENTRY_ID = -1;
```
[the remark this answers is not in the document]
eolivelli: @fpj there is no more such constant "INVALID_ENTY_ID" in the new API.
I wanted to drop every reference to the old class LedgerHandle (in org.apache.bookkeeper.client package), now we are using only the `org.apache.bookkeeper.client.api` classes.

We can think about bringing back that constant into some utility class inside  `org.apache.bookkeeper.client.api`

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
eolivelli: done

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -201,15 +207,40 @@ public String toString() {
...
+        boolean isEmpty() {
+            return this.reader == null;
+        }
 
+        private void close() {
+            // Release memory held by BookKeeper internals.
+            // we have to prevent a double free
+            if (closed.compareAndSet(false, true)) {
+
```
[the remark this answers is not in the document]
eolivelli: done

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
```diff
@@ -572,18 +577,15 @@ private long fetchLastAddConfirmed(WriteLedger writeLedger, Map<Long, Long> last
...
-        Write write = (Write) ctx;
+    private void addCallback(Long entryId, Throwable error, Write write) {
         try {
-            assert handle.getId() == write.getWriteLedger().ledger.getId()
-                    : "Handle.Id mismatch: " + write.getWriteLedger().ledger.getId() + " vs " + handle.getId();
-            write.setEntryId(entryId);
-            if (rc == 0) {
+            if (error == null) {
+                assert entryId != null;
+                write.setEntryId(entryId);
```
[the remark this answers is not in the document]
eolivelli: here we are in the success case branch, so we are setting the entryId so that the caller knows it, it is the same behaviour as before.
The `if` was `rc==0` that means success, now the condition is `error==null`

Or maybe I misunderstood your comment

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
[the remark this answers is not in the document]
eolivelli: very good idea ! I have updated the code

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
eolivelli: updated with BK_EXCEPTION_HANDLER

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
eolivelli: Updated with BK_EXCEPTION_HANDLER

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -15,43 +15,65 @@
...
         try {
             return Exceptions.handleInterruptedCall(() ->
-                    bookKeeper.createLedger(
-                            config.getBkEnsembleSize(),
-                            config.getBkWriteQuorumSize(),
-                            config.getBkAckQuorumSize(),
-                            LEDGER_DIGEST_TYPE,
-                            config.getBKPassword()));
-        } catch (BKException.BKNotEnoughBookiesException bkEx) {
+                    FutureUtils.result(bookKeeper.newCreateLedgerOp()
```
eolivelli: please note that FutureUtils#result already handles InterruptedException in the same way Exception#handleInterruptedCall does

https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/concurrent/FutureUtils.java#L73

I prefer leaving Pravega Exceptions#handleIntettuptedCall in place in order to make it clear that we are dealing with InterruptedException, but actually it is not needed

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/LogReader.java
```diff
@@ -71,12 +72,7 @@
     public void close() {
         if (!this.closed.getAndSet(true)) {
             if (this.currentLedger != null) {
-                try {
-                    Ledgers.close(this.currentLedger.handle);
-                } catch (DurableDataLogException bkEx) {
-                    log.error("Unable to close LedgerHandle for Ledger {}.", this.currentLedger.handle.getId(), bkEx);
-                }
-
+                this.currentLedger.close();
```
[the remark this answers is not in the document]
eolivelli: Closing a ReadHandle is actually a no-op

RaulGracia:
@eolivelli I have executed this branch in our Jenkins server, and the build failed because a Bookkeeper-related test:
```
io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLogTests$SecureBookKeeperLogTests.testAppendTransientBookieFailure
Error Message
java.util.concurrent.CompletionException: io.pravega.common.util.RetriesExhaustedException: io.pravega.segmentstore.storage.WriteFailureException: Active Ledger is closed.
Stacktrace
java.util.concurrent.CompletionException: io.pravega.common.util.RetriesExhaustedException: io.pravega.segmentstore.storage.WriteFailureException: Active Ledger is closed.
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
	at java.util.concurrent.CompletableFuture.biRelay(CompletableFuture.java:1284)
	at java.util.concurrent.CompletableFuture$BiRelay.tryFire(CompletableFuture.java:1270)
	at java.util.concurrent.CompletableFuture$CoCompletion.tryFire(CompletableFuture.java:1020)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1977)
	at io.pravega.segmentstore.storage.impl.bookkeeper.Write.fail(Write.java:183)
	at io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLog.executeWrites(BookKeeperLog.java:493)
	at io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLog.processPendingWrites(BookKeeperLog.java:426)
	at io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLog.processWritesSync(BookKeeperLog.java:386)
	at java.util.concurrent.CompletableFuture.uniRun(CompletableFuture.java:705)
	at java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:687)
	at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:442)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
Caused by: io.pravega.common.util.RetriesExhaustedException: io.pravega.segmentstore.storage.WriteFailureException: Active Ledger is closed.
	at io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLog.executeWrites(BookKeeperLog.java:477)
	... 12 more
Caused by: io.pravega.segmentstore.storage.WriteFailureException: Active Ledger is closed.
	at io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLog.handleWriteException(BookKeeperLog.java:663)
	at io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLog.addCallback(BookKeeperLog.java:598)
	at io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLog.lambda$executeWrites$5(BookKeeperLog.java:484)
	at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:760)
	at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:736)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1977)
	at org.apache.bookkeeper.client.SyncCallbackUtils.finish(SyncCallbackUtils.java:83)
	at org.apache.bookkeeper.client.SyncCallbackUtils$SyncAddCallback.addComplete(SyncCallbackUtils.java:251)
	at org.apache.bookkeeper.client.AsyncCallback$AddCallback.addCompleteWithLatency(AsyncCallback.java:91)
	at org.apache.bookkeeper.client.LedgerHandle$9.safeRun(LedgerHandle.java:1312)
	at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	... 1 more
Caused by: org.apache.bookkeeper.client.BKException$BKLedgerClosedException: Attempt to write to a closed ledger
	... 9 more
```
Here I post the rest of log available in Jenkins: [log.txt](https://github.com/pravega/pravega/files/4117840/log.txt)

It happens sometimes that a build works well in one environment (e.g.m, Travis) and then we spot an issue on another environment (e.g., Jenkins). I will re-execute the build to verify whether this is a consistent problem or not.

RaulGracia:
Please @fpj, do not merge this PR until we verify whether the failure observed is consistent or not.

eolivelli:
@RaulGracia 

> @eolivelli I have executed this branch in our Jenkins server, and the build failed because a Bookkeeper-related test

Do you have news ? Did you re-run the tests ?

fpj:
In the meanwhile, @eolivelli what's the issue number associated to this PR?

RaulGracia:
@eolivelli looks like the failure was one off, I have executed 2 builds on a row without issues.

eolivelli:
> In the meanwhile, @eolivelli what's the issue number associated to this PR? 

Here it is #4531

eolivelli:
> could you look into the codecov report to see how we can increase coverage, please?

I can't find any relation between this change and the changes on code coverage reported by codecov
https://codecov.io/gh/pravega/pravega/compare/85f4f8bcd9f1ea10e948c28a72747488184b19a0...2461f09354cb0580abe764fbf71781a597bc7cbf/changes

I am new to codecov but it seems that the diff it is reporting from the "Details button" is not related to this patch

https://codecov.io/gh/pravega/pravega/compare/85f4f8bcd9f1ea10e948c28a72747488184b19a0...2461f09354cb0580abe764fbf71781a597bc7cbf/diff

https://codecov.io/gh/pravega/pravega/compare/85f4f8bcd9f1ea10e948c28a72747488184b19a0...2461f09354cb0580abe764fbf71781a597bc7cbf/changes

gradle.properties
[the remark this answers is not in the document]
andreipaduroiu: @fpj We are upgrading to 4.9.2. There is nothing to revert here.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLog.java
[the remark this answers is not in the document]
andreipaduroiu: This is pre-existing code. The entry id is used by upstream code to calculate an address in the BookKeeper Log (not ledger) which is used for truncation purposes.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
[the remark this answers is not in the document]
andreipaduroiu: Suggestion: add this constant in our `Ledgers.java` class.

test/integration/src/main/java/io/pravega/test/integration/selftest/adapters/BookKeeperAdapter.java
```diff
@@ -168,24 +174,18 @@ protected void shutDown() {
...
-            if (rc == BKException.Code.OK) {
-                result.complete(null);
-            } else {
-                result.completeExceptionally(BKException.create(rc));
-            }
-        };
-        lh.asyncAddEntry(s.array(), s.arrayOffset(), s.getLength(), addCallback, null);
-        return result;
+        // appendAsync returns CompletableFuture<Long> 
+        return (CompletableFuture) lh
```
andreipaduroiu: Use `Futures.toVoid`. I do not like how this casting works here; this will also eliminate the need for `@SuppressWarnings`.

andreipaduroiu:
@fpj It looks like CodeCov went nuts with this PR. It doesn't show any of @eolivelli's changes but it complains about totally unrelated code. IMO, its results for this should be ignored.

test/integration/src/main/java/io/pravega/test/integration/selftest/adapters/BookKeeperAdapter.java
eolivelli: @andreipaduroiu done

eolivelli:
@andreipaduroiu 
> Suggestion: add this constant in our Ledgers.java class.
I can't understand this comment, in that line (I can't add an inline note, I don't know why) I am already adding a constant in Ledgers.java

andreipaduroiu:
@eolivelli My bad. For some reason I missed the context of that line and I didn't see what class it was in. What you're doing is OK, no worries.

RaulGracia:
@eolivelli we are executing performance tests on this PR compared to the `master` version without it. We want to be on the safe side regarding any unintentional impact on performance that this change could induce. Will let you know once we complete the experiments.

eolivelli:
Okay

eolivelli:
> @eolivelli we are executing performance tests on this PR compared to the `master` version without it. We want to be on the safe side regarding any unintentional impact on performance that this change could induce. Will let you know once we complete the experiments.

@RaulGracia do you have good news ?

RaulGracia:
Hi @eolivelli, according to our observations there are a couple of points from this PR that may require extra effort:

1) This PR fails the build very often because of `BookKeeperLogTests$RegularBookKeeperLogTests > testAppendTransientBookieFailure FAILED`. We have tested in another PR that the very fact of just upgrading the client to `4.9.2` induces this test to fail. This is something we need to address.
2) In our benchmark tests, we executed the branch of this PR and we have observed more Segment Store restarts that are not observed with `master` (same configuration, same hardware).
3) On the successful experiments, we have observed a small decrease in performance (10%-15%) compared to `master`. This does not make much sense, given that the only change is the API used, but is what we have observed so far. Maybe the instability is impacting on the performance results.

@andreipaduroiu any additional comments?

eolivelli:
`given that the only change is the API used`
@RaulGracia this is partially not true, we are also upgrading BK Client runtime.
As far as I know there are no big changes from 4.7 to 4.9, but I must say that in my experience I have always followed upgrades of BK one version at a time, so maybe there is some new knob to tune that I don't remember.

IIRC some time ago in BK we introduced server side throttling of requests. There were also improvements and bug fixes around "group commit" on Journal (but maybe they are in 4.10)

eolivelli:
@RaulGracia  can you please elaborate more about the performance problems ?

What is "performance" in this case ? Do you have a way to isolate:
- client writer performance kpis (latency, throughput, memory usage, CPU?)
- reader performance kpis  (latency, throughput, memory usage, CPU?)
- bookie side kpis (memory usage, CPU....)

Do you have logs about Bookies failures ?

eolivelli:
@RaulGracia @andreipaduroiu is it worth to rebase this patch ?
Ping me when you want to resume this work.
I feel it would be better to move to 4.10.x, that's the baseline I am running now in production.

I am curious about the problems you found, I left a question above to @RaulGracia

RaulGracia:
Thanks @eolivelli, I will provide information about the information you requested. Still, it would be great to execute again the experiments with the recent changes we introduced in the Segment Store (specially the change in the digest type and the new way of allocating memory for data frames).
Regarding rebasing, I think that it makes sense, but let's wait until we create the release candidate for Pravega 0.7 (this week). And thanks for your patience!

eolivelli:
@RaulGracia I am updating the patch to latest master.

I suggest to try to move to 4.10.0, as if we make new releases in BK it is better to do them in 4.10  branch other than 4.9.x. As far as I know only Pravega project is still on that old version (< 4.10)

I can update gradle.properties to 4.10.0 if you want

cc @fpj @andreipaduroiu

eolivelli:
Travis passed.
I have signed and force pushed.

I will wait for news

eolivelli:
Codecov is not working properly here.

RaulGracia:
Thanks a lot @eolivelli, now that we have released Pravega 0.7 will wil lretake the work on this PRs. I think that at the moment we can stick upgrading to 4.9.2 at the moment. The next step from our side is to give this PR another round of benchmarking to verify how it compares with current `master`.

RaulGracia:
Just FYI, found this test failing:
```
io.pravega.segmentstore.storage.impl.bookkeeper.BookKeeperLogTests$RegularBookKeeperLogTests > testAppendTransientBookieFailure FAILED
    java.util.concurrent.CompletionException: io.pravega.common.util.RetriesExhaustedException: io.pravega.segmentstore.storage.WriteFailureException: Active Ledger is closed.
        Caused by:
        io.pravega.common.util.RetriesExhaustedException: io.pravega.segmentstore.storage.WriteFailureException: Active Ledger is closed.
            Caused by:
            io.pravega.segmentstore.storage.WriteFailureException: Active Ledger is closed.
                Caused by:
                org.apache.bookkeeper.client.BKException$BKLedgerClosedException: Attempt to write to a closed ledger
```

andreipaduroiu:
Since we first observed that unit test fail (on this branch), I have been monitoring all the master branch builds and I have not seen that one fail (at all or with that message). So there must be some change on the BK/BK Client side that either doesn't work well or we need to adjust.

From my initial observation (a while back) is that some change has been made to the BK client that will immediately mark the Ledger as "Closed" when we want to close it (i.e., for rollovers). This might be tripping the upstream code in a way that it doesn't expect it.

eolivelli:
@andreipaduroiu the major improvement in 4.9 compared to 4.8 is a general refactor that made "immutable" the LedgerMedata internal variable in order to prevent race conditions and to have a clearer management of the status of the writer.
Maybe we are hitting some race here.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -15,43 +15,67 @@
...
 @Slf4j
 final class Ledgers {
-    static final long NO_ENTRY_ID = LedgerHandle.INVALID_ENTRY_ID;
+    static final long NO_ENTRY_ID = -1;
     /**
      * How many ledgers to fence out (from the end of the list) when acquiring lock.
      */
     static final int MIN_FENCE_LEDGER_COUNT = 2;
 
+    private static final DigestType LEDGER_DIGEST_TYPE = DigestType.MAC;
```
RaulGracia: Is `LEDGER_DIGEST_TYPE` even used? We did a change in PR #4559 in which the digest is passed via config. Actually, in this file I see that you are using `config.getDigestType()`, which is good. So, is this still necessary?

eolivelli:
@RaulGracia @andreipaduroiu 
do you think we can start again on this topic ?

RaulGracia:
@eolivelli sure, we have released Pravega 0.7.1 and it would be great to include this change for 0.8.0 :)

eolivelli:
@RaulGracia @andreipaduroiu
I will rebase the patch.

Is is better to move to 4.10.0 client ?
In BK community we are going to release 4.11 and 4.9.2 is quite far from current master
Btw there are no API changes, it is only a matter of changing the version in the gradle configuration file

eolivelli:
I have rebased the patch.
I will wait for CI in order to report errors on tests.

I think it is better to evaluate to move to BK 4.10.
Probably now one in the active known users of BK is using 4.9 release line.

Btw this patch does most of the work, as it adapts Pravega to the new client side API.
Moving to 4.10 will be a matter of  changing the version, checking dependencies and perform the needed performance and regression tests

RaulGracia:
@eolivelli thanks, I see that the build has passed, which is great. Could you please fix the DCO check? (perhaps you missed to sign-off the last commits).

eolivelli:
I had problems with merge commits.
Maybe I can create a new patch from scratch when review passed.
For other PRs Tom or someone else had the ability to turn off DCO check.

RaulGracia:
@eolivelli no worries, let's do whatever changes are remaining in this PR and you can create a new PR when you prefer. I have started system tests no this branch, will let you know the result.

RaulGracia:
@eolivelli good news: this branch has passed 2/2 system tests builds, which is good sign of stability. Please, could you fix the DCO issue or re-create this PR so it can get all the necessary checks to be merged?

eolivelli:
@ravisharda the new version of this patch is #4798
I am keeping this PR open in order to keep all of the review work

RaulGracia:
Closing this PR in favor of #4798.

PR #4402, opened 2019-11-27
Issue 4399: BookKeeper Write Throttling

### LINKED ISSUE #4399: Tier 1 Throttling
**Problem description**
TBD

**Problem location**

andreipaduroiu:
**Change log description**  
- Changed configuration defaults: `bookkeeper.bkWriteTimeoutMillis=60000`, `bookkeeper.readTimeoutMillis=30000`
- Added another throttler to the Segment Store that manages the in-flight data sent to BookKeeper.

**Purpose of the change**  
Fixes #4399.

**What the code does**  
- Increased the defaults for BK Write and Read timeouts to larger values. This helps reduce instability in those deployments where BK is very slow. A write timeout would cause a Ledger to be closed, thus triggering an avalanche of Ledger Rollovers and retries, further overwhelming the system.
- Added a new Throttler in the Segment Store's `OperationProcessor` that works along the following lines
    - Throttling is triggered only if the recent BK Write Latencies exceed 10% of `bookkeeper.bkWriteTimeoutMillis`. The 10% is hardcoded as a constant.
    - If triggered, the throttle delay is a gradual measurement as a function of the amount of data outstanding in the BK queue
    - The maximum delay is applied when the outstanding data exceeds `bookkeeper.maxOutstandingBytes`.
    - There is no delay if the outstanding data is below 10% (same constant as above) of `bookkeeper.maxOutstandingBytes`.
    - The delay increases gradually between these values, as a function of the amount of outstanding data.
- A bit of refactoring to consolidate similar patterns that relate to throttling interruption by using a unified `ThrottleSourceListener`.

**How to verify it**  
All unit tests must pass.
System tests must pass.

### DIFF
--- a/config/config.properties
+++ b/config/config.properties
@@ -328,12 +328,22 @@ metrics.enableStatistics=false
 #bookkeeper.bkWriteQuorumSize=3
 
 # Write Timeout, in milliseconds.
+# This value is also used for throttling purposes. Once BookKeeper Write Latencies exceed 10% of this value, the Segment
+# Store will begin throttling in order to manage the BookKeeper write backlog and reduce the chance of write timeouts.
 # Note: BookKeeper only allows multiples of 1 second (1000 millis). This value will be rounded up to the nearest second.
-#bookkeeper.bkWriteTimeoutMillis=5000
+#bookkeeper.bkWriteTimeoutMillis=60000
 
 # Read Timeout, in milliseconds.
 # Note: BookKeeper only allows multiples of 1 second (1000 millis). This value will be rounded up to the nearest second.
-#bookkeeper.bkReadTimeoutMillis=5000
+#bookkeeper.bkReadTimeoutMillis=30000
+
+# Maximum number of bytes that can be outstanding per BookKeeperLog at any given time. This value is used for throttling
+# purposes. This value is not set on the BookKeeper Client Configuration, rather it is used internally by the Segment
+# Store throttler to manage the BookKeeper write backlog and reduce the chance of write timeouts.
+# Recommended Value: 256MB. A smaller value will make throttling more aggressive but increase the overall stability of
+# the system. A larger value will increase the likelihood of BookKeeper write timeouts if BookKeeper is unable to keep
+# up with the load the Segment Store sends its way.
+#bookkeeper.maxOutstandingBytes=268435456
 
 # Maximum Ledger size (bytes) in BookKeeper. Once a Ledger reaches this size, it will be closed and another one open.
 # Note that ledgers will not be cut off at this size, rather them reaching this size will trigger a rollover; in-flight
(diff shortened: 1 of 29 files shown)


config/config.properties
```diff
@@ -328,12 +328,22 @@ metrics.enableStatistics=false
...
-#bookkeeper.bkReadTimeoutMillis=5000
+#bookkeeper.bkReadTimeoutMillis=30000
+
+# Maximum number of bytes that can be outstanding per BookKeeperLog at any given time. This value is used for throttling
+# purposes. This value is not set on the BookKeeper Client Configuration, rather it is used internally by the Segment
+# Store throttler to manage the BookKeeper write backlog and reduce the chance of write timeouts.
+# Recommended Value: 256MB. A smaller value will make throttling more aggressive but increase the overall stability of
+# the system. A larger value will increase the likelihood of BookKeeper write timeouts if BookKeeper is unable to keep
+# up with the load the Segment Store sends its way.
+#bookkeeper.maxOutstandingBytes=268435456
```
RaulGracia: As a conservative default, is better to set this to 32MB.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ThrottlerCalculator.java
```diff
@@ -366,7 +437,11 @@ public String toString() {
         /**
          * Throttling is required due to excessive size of the Commit (Memory) Backlog Queue.
          */
-        CommitBacklog
+        CommitBacklog,
+        /**
+         * Throttlign is required due to excessive size of DurableDataLog's in-flight queue.
```
eolivelli: Typo

RaulGracia: Thanks @eolivelli, yes we tried configuring the `setThrottleValue()` parameter to low numbers (e.g., 100, 10), but it is not preventing the system to be unstable in cases of "unthrottled" clients and low IO performance Tier 1 storage (i.e., where Bookkeeper writes to). I'm validating this PR, and it looks like it is making the system stable even in non-favorable scenarios as the one mentioned. But note that this does not prevent Pravega from benefiting from the throttling capabilities of newer versions of Bookkeeper (if I'm not wrong, in `4.8` there was a server-side throttling mechanism implemented).

andreipaduroiu: @eolivelli Thank you for your suggestion. Like @RaulGracia said, that was one of the first things we tried. However I noticed that it does little to prevent backpressure on the application itself and simply guards how many entries it has "in-flight" to the bookies. At least based on the latest code in the BK repository, this uses Guava's `RateLimiter` which blocks synchronously until permits are available. In Pravega's situation that would quickly exhaust all of our threads and lead to a lot of problems, so we need an async solution.

Pravega's SegmentStore already has an async throttler that can be configured with various inputs (triggers and "cancelers") which implement a gradual throttling scheme based on how busy the system is (two thresholds: min threshold is when we throttle a little bit, and max threshold is when we apply a full stop). The advantage of this async throttler is that it doesn't block any threads in doing so, and those threads can be used by other components of the system to clear out backlog or perform other necessary functions.

eolivelli:
Yes I see.
Thanks for you asnwers

PR #4410, opened 2019-12-01
Issue 4409: Reset tokenRefreshFuture upon exception in obtaining a new delegation token

### LINKED ISSUE #4409: JWTTokenProvider does not handle errors
**Problem description**
When the JWTTokenProvider has a DNS issue during refresh the `currentRefreshFuture` never gets cleared which means subsequent `refreshToken` calls skip over the `recreateToken` call.  This leads to the TokenProvider getting locked as it never recovers even when the DNS fault has been resolved.

**Problem location**
https://github.com/pravega/pravega/blob/813b0babd4cbf99db3f4c2fb44273f9d25c555e7/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java#L225

The `currentRefreshFuture.thenApply` has no provision for an exceptionally completed future which means the `this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);` ever gets called to clear the `tokenRefreshFuture`.

ravisharda:
**Change log description**  
Fix a bug that prevents `tokenRefreshFuture` from getting reset, preventing the `DelegationTokenProvider` from working properly after errors such as a failed attempt to obtain a new delegation token.  The `tokenRefreshFuture` remained in `iscompletedExceptionally` state forever, and therefore the `JwtTokenProviderImpl` failed to retrieve token post that event. 

**Purpose of the change**  
Fixes #4409 

**What the code does**  
Resets `tokenRefreshFuture` shared variable to `null` if there is an exception in obtaining a new delegation token, so that the logic to refresh the token will be executed again on next attempt. 

**How to verify it**  
All existing and new tests must pass. 
System tests must pass, with and without `auth` enabled.

### DIFF
--- a/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
+++ b/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
@@ -22,6 +22,7 @@ import lombok.extern.slf4j.Slf4j;
 import java.time.Instant;
 import java.util.Base64;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -239,6 +240,12 @@ public class JwtTokenProviderImpl implements DelegationTokenProvider {
                     this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);
                     LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
                     return delegationToken.get().getValue();
+                })
+                .exceptionally(ex -> {
+                    this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);
+                    log.warn("Encountered an exception in refreshToken", ex);
+                    LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
+                    throw ex instanceof CompletionException ? (CompletionException)ex: new CompletionException(ex);
                 });
     }
 
--- a/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
+++ b/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
@@ -245,7 +245,7 @@ public class JwtTokenProviderImpl implements DelegationTokenProvider {
                     this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);
                     log.warn("Encountered an exception in refreshToken", ex);
                     LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
-                    throw ex instanceof CompletionException ? (CompletionException)ex: new CompletionException(ex);
+                    throw ex instanceof CompletionException ? (CompletionException) ex : new CompletionException(ex);
                 });
     }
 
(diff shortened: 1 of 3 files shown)


client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -239,6 +240,12 @@ boolean isWithinRefreshThreshold(Instant currentInstant, Instant expiration) {
                     this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);
                     LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
                     return delegationToken.get().getValue();
+                })
+                .exceptionally(ex -> {
```
shrids: We can use `.whenComplete((v, t) -> { }`. It would simplify the code here.

shrids: we an use `whenComplete((token, throwable) -> {}` if `this.recreateToken()` returns a CompletableFuture<DelegationToken>. This would simplify the code right ?

ravisharda: The [JavaDocs ](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) for CompletableFuture says that `whenComplete` _"returns a new CompletionStage with the same result or exception as this stage, that executes the given action when this stage completes"_.

In this piece of code, we are not returning the result of the previous CompletionStage (except for the exception itself), but are rather returning an object extracted inside this stage: `delegationToken.get().getValue()`. 

Therefore, I believe that `whenComplete` is not a good choice here.

shrids: I think if we modify `this.recreateToken()` returns a CompletableFuture<String> or CompletableFuture<DelegationToken> then this would simply the code here.

ravisharda: As discussed, I have modified the code to remove duplication via use of the `handle()` method.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -226,20 +229,25 @@ boolean isWithinRefreshThreshold(Instant currentInstant, Instant expiration) {
...
-                    LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
-                    return delegationToken.get().getValue();
-                });
+        return currentRefreshFuture.handle((v, ex) -> {
+            this.tokenRefreshFuture.compareAndSet(handleToCurrentRefreshFuture, null);
+            LoggerHelpers.traceLeave(log, "refreshToken", traceEnterId, this.scopeName, this.streamName);
+            if (ex != null) {
+                log.warn("Encountered an exception in when refreshing token for scope {} and stream {}",
+                        this.scopeName, this.streamName, Exceptions.unwrap(ex));
+                throw ex instanceof CompletionException ? (CompletionException) ex : new CompletionException(ex);
```
shrids: we can simply throw `ex` wrapped inside a CompletationException.

ravisharda: The intention of throwing the existing exception is to avoid creating a new exception object unnecessarily.

shiveshr: if you dont want to "handle" the returned value/exception, then you can use `whenComplete`

shrids: ok. `controllerClient.getOrRefreshDelegationTokenFor(scopeName, streamName)` will not throw an exception of type CompletationException.

PR #4413, opened 2019-12-02
Issue 4412: Fix bug in ExtendedS3StorageConfig to correctly handle missing trailing `/` in root.

### LINKED ISSUE #4412: ExtendedS3StorageConfig does not handle missing trailing `/` in root
**Problem description**
Current logic in ExtendedS3StorageConfig incorrectly assumes that root has a trailing '/' character. If this is not the case then the S3 storage is wrongly initialized and results in invalid paths/urls.

**Problem location**
ExtendedS3StorageConfig.java constructor where we initialize values from TypedProperties.

Correctly handle the case, by adding the / char when it is missing.

sachin-j-joshi:
Current logic in ExtendedS3StorageConfig incorrectly assumes that root has a trailing '/' character. If this is not the case then the S3 storage is wrongly initialized and results in invalid paths/urls.
This change fixes the bug by adding the / char when it is missing.

**Change log description**  
Fix bug in ExtendedS3StorageConfig to correctly handle missing trailing `/` in root.

**Purpose of the change**  
Fixes #4412 

**What the code does**  
This change fixes it by adding the / char when it is missing.

**How to verify it**  
All tests should pass.

### DIFF
--- a/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
+++ b/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
@@ -33,6 +33,7 @@ public class ExtendedS3StorageConfig {
     public static final Property<Boolean> USENONEMATCH = Property.named("useNoneMatch", false);
 
     private static final String COMPONENT_CODE = "extendeds3";
+    private static final String PATH_SEPARATOR = "/";
 
     //endregion
 
@@ -91,7 +92,8 @@ public class ExtendedS3StorageConfig {
      * @param properties The TypedProperties object to read Properties from.
      */
     private ExtendedS3StorageConfig(TypedProperties properties) throws ConfigurationException {
-        this.root = properties.get(ROOT);
+        String givenRoot = properties.get(ROOT);
+        this.root = givenRoot.endsWith(PATH_SEPARATOR) ? givenRoot : givenRoot + PATH_SEPARATOR;
         this.accessKey = properties.get(ACCESS_KEY_ID);
         this.secretKey = properties.get(SECRET_KEY);
         this.url = java.net.URI.create(properties.get(URI));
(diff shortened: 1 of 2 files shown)


bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3StorageTest.java
```diff
@@ -80,6 +84,23 @@ public void testCreateIfNoneMatch() {
     }
     //endregion
 
+    @Test
```
kevinhan88: Suggest to add case to ensure the tailing space could also be trimmed.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
```diff
@@ -33,6 +33,7 @@
     public static final Property<Boolean> USENONEMATCH = Property.named("useNoneMatch", false);
 
     private static final String COMPONENT_CODE = "extendeds3";
+    private static final String PATH_SEPARATOR = "/";
```
andreipaduroiu: This should go into `StreamSegmentNameUtils.java` and you should update all references to the string `"/"` in there too.

sachin-j-joshi: StreamSegmentNameUtils is in ':shared:protocol' project which has mostly protocol related stuff. Today 'bindings' project does not depend on it and it doesn't make sense to bring in that dependency just for this change.

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3StorageTest.java
sachin-j-joshi: Added test

Issue #4416, opened 2019-12-04
Remove deprecated API EventStreamWriter.beginTxn()


tkaitchuck:
**Problem description**
Before the introduction to TransactionalEventStreamWriter, EventStreamWriter had the transaction methods on in so transactional writes and non-transactional writes were performed via the same interface. This was deprecated in https://github.com/pravega/pravega/pull/3126

**Problem location**
EventStreamWriter

**Suggestions for an improvement**
Remove the deprecated method and update all the code still using it.

Issue #1949, opened 2017-10-07
PDP 22: Apache BookKeeper for Tier 2


fpj:
**Problem description**
Pravega Design Proposal to add Apache BookKeeper as an option for Tier 2.

See https://github.com/pravega/pravega/wiki/PDP-22:-BK-based-Storage-(Tier-2)-for-Pravega

**Problem location**
Segment store, Tier 2.

**Suggestions for an improvement**
See the PDP for detail.

andreipaduroiu:
> Shall we improve the existing BookkeeperLog to implement a generic storage OR implement a different storage ledger for tier-2 purpose.

andreipaduroiu:
> We need to build an algorithm to map offset to entry-id.

arvindkandhare:
> I would strongly prefer keeping them separate. BookKeeperLog is (and should be) optimized for DurableDataLog and writing to a log. Storage has vastly different requirements and access patterns.

Thanks for the inputs. I am leaning in the same direction as well. Will wait for a couple of days for other inputs on this and then close having independent implementation as a plan.

tkaitchuck:
In terms of tracking offsets, we could use EntryId as the offset. It is already a long, and BK supports using it to position reads.

eolivelli:
I am very interested in this feature, as it will allow to deploy Pravega without HDFS or NFS.
I have already Bookies but I am using Kafka for "event streams"; Using BK as Storage for Pravega will make me able to switch from using Kafka to Pravega, and dropping the Kafka servers in favor of more "light" Pravega daemons.

Maybe I can give some help on this topic.
I am recently working at a project which will use BookKeeper as Binary Large Object Storage, so I am working on improvements on BK in that direction (see https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-14+Relax+durability)

Some ideas:
Using createLedgerAdv will let you have full control about entryIds and even on ledgerIds. But there are a few constrains:
- you cannot create gaps in the sequence of entry ids
- the write for entry id X won't be acknowledged until every entry from 0 to X-1 has been acknowledged even if your are sending entries in sparse order

We have also to take into account that an entry as maximum size (by default 1MB) so every chunk must be split into smaller chunks.
if you can pre-calculate the size of the chunk (and this is the case), you can pre-allocate the range of entry ids which will cover the chunk, so you can store as coordinates of the chunk the triple (ledgerid,first_entry_id,last_entry_id)

[...]

PR #4417, opened 2019-12-04
Issue 4416: Remove deprecated client APIs

tkaitchuck:
**Change log description**  
* Removes `ClientFactory` (Replaced by EventStreamClientFactory)
* Removes `beginTxn()` and `getTxn()` methods from the EventStreamWriter. (Replaced by methods on TransactionalEventStreamWriter)
* Removes the ByteStreamClient interface. (Replaced by ByteStreamClientFactory)

**Purpose of the change**  
Fixes https://github.com/pravega/pravega/issues/4416

**What the code does**  
Removes APIs deprecated in https://github.com/pravega/pravega/pull/3126
No other functional changes.


tkaitchuck:
@shrids  Build #1949 went through.

client/src/main/java/io/pravega/client/stream/ReaderGroup.java
```diff
@@ -10,9 +10,7 @@
 package io.pravega.client.stream;
 
 import com.google.common.annotations.Beta;
```
RaulGracia: Can we also remove those API calls tagged with "Beta"? For instance `generateStreamCuts` has been there for the last 2 major versions at least. The same for the Byte API related methods (I also remember some Beta annotation there). I think that we can remove these `Beta` annotations at this point as many of these calls have been tested for quite a while so far. Also, I still see a deprecated call in this clasee: `resetReadersToCheckpoint`. Can we get rid of this one also in this PR to help users to use the right methods from 0.7 onwards?

tkaitchuck:
> System test run `/test-pravega-continuous-1/442/` was successful with this change set.
> 
> Can we remove the the following deprecated apis too in this PR?
> 
> * `io.pravega.client.batch.BatchClient`
> * `io.pravega.client.batch.StreamInfo`
> * `io.pravega.client.stream.StreamConfiguration.StreamConfigurationBuilder#scope`
> * `io.pravega.client.stream.StreamConfiguration.StreamConfigurationBuilder#streamName`
> * `io.pravega.client.stream.ReaderGroup#resetReadersToCheckpoint`
> 
> Also, we can remove redundant @SuppressWarnings("deprecation") in `client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java`

Done.

client/src/main/java/io/pravega/client/stream/ReaderGroup.java
tkaitchuck: We should do that in a different PR.

RaulGracia:
Once merged, we need to focus on at least (but not only):
- https://github.com/pravega/flink-connectors/issues/304
- https://github.com/pravega/pravega-samples/pull/231
- https://github.com/pravega/video-samples/issues/10
- https://github.com/pravega/hadoop-connectors/issues/106

PR #4418, opened 2019-12-04
Issue 4393: Fixing error sequencing in AppendProcessor

### LINKED ISSUE #4393: Events read out of sequence
**Problem description**
We have detected events read out of sequence in out automated longevity experiments executed against master. Concretely, the workload used (`small-key`) writes data to a very small subset of keys (4), which induces a lot of stream scale operations (+100) within 12 hours. We have seen the test failing twice in a row due to the same problem:
```
22:57:31  INFO  [2019-11-22 21:56:42,674] io.pravega.longevity.utils.PerformanceUtils: Writers (3/3): events:346,834,138, events/sec:8,039, KB/sec:6164.0195
22:57:31  INFO  [2019-11-22 21:56:42,674] io.pravega.longevity.utils.PerformanceUtils:  Transactions: started: 0, perSec:0.0, committed:0, failed:0, aborted:0, aborted but read: 0
22:57:31  INFO  [2019-11-22 21:56:42,674] io.pravega.longevity.utils.PerformanceUtils:  Blocked Waiting For ACK: times blocks: 7,991,034, average block time (ms): 6
22:57:31  INFO  [2019-11-22 21:56:42,674] io.pravega.longevity.utils.PerformanceUtils: Readers (6/6): events:346,840,230, events/sec:8,039, KB/sec:6164.128
22:57:31  172.28.14.3 - - [22/Nov/2019:21:56:42 +0000] "POST /v1/longevity/performance HTTP/1.1" 200 2278 "-" "Jersey/2.25.1 (HttpUrlConnection 1.8.0_131)" 1
...
22:58:03  eventsInSequence:
22:58:03  succeeded : false
22:58:03  expected : 0, got 2813
```
```
15:46:17  INFO  [2019-11-23 14:45:29,059] io.pravega.longevity.utils.PerformanceUtils: Writers (3/3): events:405,584,230, events/sec:9,401, KB/sec:7208.1494
15:46:17  INFO  [2019-11-23 14:45:29,059] io.pravega.longevity.utils.PerformanceUtils:  Transactions: started: 0, perSec:0.0, committed:0, failed:0, aborted:0, aborted but read: 0
15:46:17  INFO  [2019-11-23 14:45:29,059] io.pravega.longevity.utils.PerformanceUtils:  Blocked Waiting For ACK: times blocks: 14,035,742, average block time (ms): 5
15:46:17  INFO  [2019-11-23 14:45:29,059] io.pravega.longevity.utils.PerformanceUtils: Readers (6/6): events:405,591,766, events/sec:9,401, KB/sec:7208.283
15:46:17  172.28.14.3 - - [23/Nov/2019:14:45:29 +0000] "POST /v1/longevity/performance HTTP/1.1" 200 2278 "-" "Jersey/2.25.1 (HttpUrlConnection 1.8.0_131)" 
...
15:46:36  eventsInSequence:
15:46:36  succeeded : false
15:46:36  expected : 0, got 3733
```
That is, we got around 3000-4000 events out of sequence out of 340-400 millions.

Note that the last successful longevity run we had was before PR #4331. We are re-executing this experiment again to double check that this data point is actually right.

**Problem location**
Client/Append Processor interactions.

Ensure that events are sent in order for a given routing key.

andreipaduroiu:
**Change log description**  
- Changed the AppendProcessor to send error messages back to the client only after all appends initiated prior to that error complete.

**Purpose of the change**  
Fixes #4393.

**What the code does**  
- Moved `WriterState` out of `AppendProcessor` in order to write unit tests against it.
- Upon encountering a failed append, the `AppendProcessor` will take note of the current `lastStoredEventNumber` and delay handling the exception until after all Appends with Event Numbers up to that one are completed (successfully or not). 
    - The `AppendProcessor` will record all failed appends and send error message for each such failure. 
- The `AppendProcessor` will keep a `WriterState` registered until there are no more in-flight requests or errors to be handled. This ensures that any retries that the Client may attempt due to a previous error will not be met with a generic `IllegalStateException` or cause a new `setupAppend` to be initiated, which could cause event duplication,

**How to verify it**  
Unit, system and longevity tests must pass. 
- Longevity runs 200, 201 and 202 passed. 
- System test run 2460 passed.

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -175,7 +175,7 @@ public class AppendProcessor extends DelegatingRequestProcessor {
         long traceId = LoggerHelpers.traceEnter(log, "append", append);
         UUID id = append.getWriterId();
         WriterState state = this.writerStates.get(Pair.of(append.getSegment(), id));
-        Preconditions.checkState(state != null, "Data from unexpected connection: %s.", id);
+        Preconditions.checkState(state != null, "Data from unexpected connection: Segment=%s, WriterId=%s.", append.getSegment(), id);
         long previousEventNumber = state.beginAppend(append.getEventNumber());
         int appendLength = append.getData().readableBytes();
         adjustOutstandingBytes(appendLength);
@@ -230,23 +230,26 @@ public class AppendProcessor extends DelegatingRequestProcessor {
                     // Revert the state to the last known good one. This is needed because we do not close the connection
                     // for offset-conditional append failures, hence we must revert the effects of the failed append.
                     log.debug("Conditional append failed due to incorrect offset: {}, {}", append, exception.getMessage());
-                    state.conditionalAppendFailed();
-                    connection.send(new ConditionalCheckFailed(append.getWriterId(), append.getEventNumber(), append.getRequestId()));
+                    synchronized (state.getAckLock()) {
+                        state.conditionalAppendFailed();
+                        connection.send(new ConditionalCheckFailed(append.getWriterId(), append.getEventNumber(), append.getRequestId()));
+                    }
                 } else {
-                    // Clear the state in case of error.
-                    this.writerStates.remove(Pair.of(append.getSegment(), append.getWriterId()));
-
                     // Record the exception handling into the Writer State. It will be executed once all the in-flight
                     // appends are drained.
-                    state.appendFailed(() ->
-                            handleException(append.getWriterId(), append.getRequestId(), append.getSegment(), append.getEventNumber(),
-                                    "appending data", exception));
+                    state.appendFailed(() -> {
+                        handleException(append.getWriterId(), append.getRequestId(), append.getSegment(), append.getEventNumber(),
+                                "appending data", exception);
+
+                        // Clear the state in case of error.
+                        this.writerStates.remove(Pair.of(append.getSegment(), append.getWriterId()));
+                    });
                 }
             }
 
             // After every append completes, check if we are done with this WriterState, and if we were asked to execute
             // something after that happens.
-            Runnable runWhenAllComplete = state.getRunWhenAllComplete();
+            Runnable runWhenAllComplete = state.fetchRunWhenAllComplete();
             if (runWhenAllComplete != null) {
                 synchronized (state.getAckLock()) {
                     runWhenAllComplete.run();
@@ -339,10 +342,13 @@ public class AppendProcessor extends DelegatingRequestProcessor {
         @GuardedBy("this")
         private long inFlightCount;
         @GuardedBy("this")
+        private boolean closed;
+        @GuardedBy("this")
         private Runnable whenAllComplete;
 
         WriterState(long initialEventNumber) {
             this.inFlightCount = 0;
+            this.closed = false;
             this.lastStoredEventNumber = initialEventNumber;
             this.lastAckedEventNumber = initialEventNumber;
         }
@@ -354,7 +360,7 @@ public class AppendProcessor extends DelegatingRequestProcessor {
          * @return The previously attempted Event Number.
          */
         synchronized long beginAppend(long eventNumber) {
-            Preconditions.checkState(this.whenAllComplete == null, "Connection closed. Cannot begin new appends.");
+            Preconditions.checkState(!isClosed(), "Connection closed. Cannot begin new appends.");
             long previousEventNumber = this.lastStoredEventNumber;
             Preconditions.checkState(eventNumber >= previousEventNumber, "Event was already appended.");
             this.lastStoredEventNumber = eventNumber;
@@ -398,9 +404,19 @@ public class AppendProcessor extends DelegatingRequestProcessor {
             return previousLastAcked;
         }
 
-        Runnable getRunWhenAllComplete() {
+        Runnable fetchRunWhenAllComplete() {
             synchronized (this) {
-                return this.inFlightCount == 0 ? this.whenAllComplete : null;
+                if (this.inFlightCount == 0) {
+                    Runnable result = this.whenAllComplete;
+                    this.whenAllComplete = null;
+                    if (result != null) {
+                        this.closed = true;
+                    }
+
+                    return result;
+                } else {
+                    return null;
+                }
             }
         }
 
@@ -412,6 +428,11 @@ public class AppendProcessor extends DelegatingRequestProcessor {
         synchronized void appendFailed(Runnable runWhenAllComplete) {
             this.inFlightCount--;
             this.whenAllComplete = runWhenAllComplete;
+        }
+
+        @GuardedBy("this")
+        private boolean isClosed() {
+            return this.closed || this.whenAllComplete != null;
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -229,19 +226,23 @@ private void handleAppendResult(final Append append, Long newWriteOffset, Throwa
...
+                        state.conditionalAppendFailed(append.getEventNumber());
+                        connection.send(new ConditionalCheckFailed(append.getWriterId(), append.getEventNumber(), append.getRequestId()));
                     }
-                    connection.send(new ConditionalCheckFailed(append.getWriterId(), append.getEventNumber(), append.getRequestId()));
                 } else {
-                    // Clear the state in case of error.
-                    this.writerStates.remove(Pair.of(append.getSegment(), append.getWriterId()));
-                    handleException(append.getWriterId(), append.getRequestId(), append.getSegment(), append.getEventNumber(),
-                            "appending data", exception);
+                    // Record the exception handling into the Writer State. It will be executed  once all the current
```
tkaitchuck: I have a concern with this model: It allows for the possibility that an append could fail but that one pipelined after it could succeed. Obviously this should not be possible and there is a check to try to prevent this. But if that logic for whatever reason had an edge case were it allowed that to occur, then this code would cause the ack of the later event to be sent to the client first effectively acking the failed append also. Can we incorporate some sort of assertion to prevent this possibility form occurring and hopefully aid in debugging it if it were to trigger?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/WriterState.java
```diff
@@ -0,0 +1,292 @@
...
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.ThreadSafe;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Maintains state for a Writer in the {@link AppendProcessor}.
+ * This class is tightly-coupled with the {@link AppendProcessor} behavior and should not be used for any other purpose.
```
tkaitchuck: Indeed it is. Perhaps we could separate concerns a little bit: The main complication that is making things difficult is the possibility of threads reordering things. What if we used this class to order the responses of all of the replies successful or not. This could be done similar to how it is here with the list of runnables. If however each were associated with an event number we could make sure they are run in the correct order.
Obviously the two classes would still be fairly closely tied because of locking etc. But it might make it easier to reason about the correctness.

andreipaduroiu: I'm answering both this and https://github.com/pravega/pravega/pull/4418#discussion_r354090715.

Your solution will indeed solve the correctness of the general case. However, the downside is the need for an array/list/queue for each `WriterState` where we'll keep all the in-flight append event numbers. Assuming an average array size of 256 entries (array expands, but doesn't contract), we will need 256 x 8 bytes x 10000 connections = 20MB of heap just to store this.

Taking a step back, this problem (with a successful append acking a possible failed one) exists even now in master. This PR doesn't make the problem worse. The problem was introduced with the pipelining.

However let's think practically. In reality we'll only have SegmentSealed and BadAttribute.

If we get SegmentSeal then any append after that will also be failed. None will succeed.
For BadAttribute it's the same story. The client isn't going to send a random bad one followed by a good one. It will send a flurry of bad ones.

The main goal of this PR was to make sure that any error messages are sent back to the client only after all previous appends have been acked, otherwise we risk duplicating them Inna segment's successors. 

If that had been enough this PR would have been simpler. However for some reason I don't fully understand, some integration tests expect that I send all error messages. I did a dump of what exceptions are generated in those tests and it was always SegmentIsSealed. 

So while I agree with your assessment that this method is not correct in the general case, when factoring in the behavior of the client and the Appendprocessor, it will work as expected.

andreipaduroiu: We have discussed this offline and agreed that, barring the Segment Store works incorrectly, this should not be an issue.

I've added a `log.warn` to indicate if we get any successful appends after a failed one via a Connection/Writer.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -226,22 +223,33 @@ private void handleAppendResult(final Append append, Long newWriteOffset, Throwa
                         connection.send(dataAppendedAck);
                     }
                 }
+
+                if (append.getEventNumber() > state.getLowestFailedEventNumber()) {
```
tkaitchuck: Should this be called while still under the ackLock above?

andreipaduroiu: `append` is immutable and `state.getLowestFailedEventNumber` can only change to a lower value. So whether we do it earlier (in the lock) or now, it will still fire under the same conditions. 

`WriterState` is thread safe so there is no need to sync on `getAckLock` (which is used only to make sure the acks are sent to the client in the correct order).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/WriterState.java
```diff
@@ -0,0 +1,311 @@
...
+        this.smallestFailedEventNumber = Math.min(eventNumber, this.smallestFailedEventNumber);
+        if (this.errorContexts == null) {
+            this.errorContexts = new ArrayList<>();
+        }
+
+        // Update any existing ErrorContexts that an append has completed.
+        updateErrorContexts(eventNumber);
+
+        // Record a new ErrorContext.
+        this.errorContexts.add(new ErrorContext(this.lastStoredEventNumber, this.inFlightCount, delayedErrorHandler));
```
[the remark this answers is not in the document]
andreipaduroiu: Good point. I decided to sort `errorContexts` when adding new items to it.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/WriterState.java
```diff
@@ -0,0 +1,311 @@
...
+     */
+    @GuardedBy("this")
+    private long lastStoredEventNumber;
+    /**
+     * The Event Number of the last Event that was Acked to the Client.
+     */
+    @GuardedBy("this")
+    private long lastAckedEventNumber;
+    /**
+     * Number of Events that have been appended to the Store but not yet acked back from the Store.
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -253,6 +261,36 @@ private void handleAppendResult(final Append append, Long newWriteOffset, Throwa
         }
     }
 
+    /**
+     * Inquires the {@link WriterState} for any eligible {@link WriterState.DelayedErrorHandler} that can be executed
+     * right now. If so, invokes all eligible handlers synchronously. If there are no more handlers remaining after this,
+     * the {@link WriterState} is unregistered, which would essentially force the client to reinvoke {@link #setupAppend}.
+     *
+     * @param state  The {@link WriterState} to query.
+     * @param append The {@link Append} that triggered this.
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

Issue #4270, opened 2019-10-11
java.io.FileNotFoundException inside Bookkeeper post successful upgrade.


Richa-Srivastava:
**Once Pravega was upgraded from `0.6.0-2391.97363b3` to `0.6.0-2378.3ae8db3`, the upgrade was successful but Readers stalled from IO which was running in background**
```
INFO  [2019-10-11 07:59:12,840] io.pravega.longevity.utils.PerformanceUtils: Readers (4/6): events:949,769,714, events/sec:5,355, KB/sec:4106.326
INFO  [2019-10-11 07:59:12,840] io.pravega.longevity.utils.PerformanceUtils: Writers (3/3): events:949,786,677, events/sec:5,356, KB/sec:4107.0938
INFO  [2019-10-11 07:59:12,840] io.pravega.longevity.utils.PerformanceUtils: Readers (4/6): events:949,769,714, events/sec:5,356, KB/sec:4107.0205
INFO  [2019-10-11 07:59:42,839] io.pravega.longevity.utils.PerformanceUtils: Writers (3/3): events:949,937,540, events/sec:5,355, KB/sec:4106.3564
INFO  [2019-10-11 07:59:42,839] io.pravega.longevity.utils.PerformanceUtils: Readers (2/6): events:949,934,912, events/sec:5,355, KB/sec:4106.3457
INFO  [2019-10-11 07:59:42,839] io.pravega.longevity.utils.PerformanceUtils: Writers (3/3): events:949,937,540, events/sec:5,355, KB/sec:4106.3564
INFO  [2019-10-11 07:59:42,839] io.pravega.longevity.utils.PerformanceUtils: Readers (2/6): events:949,934,912, events/sec:5,355, KB/sec:4106.3457
INFO  [2019-10-11 08:00:12,838] io.pravega.longevity.utils.PerformanceUtils: Writers (3/3): events:950,103,001, events/sec:5,356, KB/sec:4106.377
INFO  [2019-10-11 08:00:12,838] io.pravega.longevity.utils.PerformanceUtils: Readers (1/6): events:950,100,107, events/sec:5,355, KB/sec:4106.365
```

Following is the upgrade status:
```
Status:
  Conditions:
    Last Transition Time:  2019-10-11T07:59:51Z
    Last Update Time:      2019-10-11T07:59:51Z
    Status:                False
    Type:                  Upgrading
    Last Transition Time:  2019-10-11T08:00:21Z
[...]

PR #4420, opened 2019-12-04
Issue 3696: Upgrade Bookeeper from 4.7.3 to 4.8.2

### LINKED ISSUE #4459: Pravega Bookkeeper docker image has an old version of Bookkeeper server
**Problem description**
Pravega Bookkeeper image currently uses an old version of Bookkeeper: `4.7.3`. Many defects were fixed in later versions of Bookeeper, including but not limited to the `FileNotFoundException` reported in issue #4270 .

**Problem location**
https://github.com/pravega/pravega/blob/5d0176839755e82f64329938466099e25088c3ad/docker/bookkeeper/Dockerfile#L12

Upgrade the Bookkeeper version

ravisharda:
**Change log description**  
Upgrade the Bookeeper server from `4.7.3` to `4.8.2`. 

**Purpose of the change**  
Resolves #4459 and #4270.  

**What the code does**  

Modifies the Dockerfile and entrypoint to work with Bookeeper `4.8.2`. It does the following (as described very well by @shrids elsewhere): 

1. The default 4.8.2 bookkeeper start up scripts uses `ZooKeeperMain` to create/modify nodes in ZK. This throws 
```
/opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 3
```
Due to this, we ended up in using `zk-shell` on similar lines as Apache Bookeeper [PR 1666](https://github.com/apache/bookkeeper/pull/1666). 

2. The `4.8.2` script exits if `BK_STREAM_STORAGE_ROOT_PATH` is not set. Also, a zk node in this path is expected to be present if a bookie is up and running. (This is not the case with the default bk 4.8.2 image)

3. `/opt/bookkeeper/bin/bkctl --service-uri` this option used in Bookkeeper image did not work. `java.lang.IllegalArgumentException: No service URI` is is observed. It is similar to issues observed in https://github.com/apache/bookkeeper/pull/2112.

4. It has a workaround for creating ephemeral node using `zk-shell`. The exit status returned by the `/opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain` used in the base image is always 0, even in the case the node exists, causing unnecessary bookie restart.

**How to verify it**  
* A successful system test run
* A successful longevity test run
* Manual testing of at least the following scenarios:
  * Upgrade
  * Bookkeeper metrics in Grafana
  * Failure of Bookkeeper instances in a cluster


docker/bookkeeper/entrypoint.sh
```diff
@@ -32,7 +20,11 @@ BK_CLUSTER_NAME=${BK_CLUSTER_NAME:-"bookkeeper"}
 BK_LEDGERS_PATH="/${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME}/${BK_CLUSTER_NAME}/ledgers"
```
sarlaccpit: Do you know where PRAVEGA_CLUSTER_NAME set ?

ravisharda: It is here: 

https://github.com/pravega/pravega/blob/e2dfc4681721b4576a090e33e09f79d24040ddc2/docker/bookkeeper/entrypoint.sh#L18. The default value is `pravega-cluster`.

docker/bookkeeper/entrypoint.sh
```diff
@@ -8,31 +8,32 @@
...
-  IFS=',' read -ra directories <<< $1
-  for i in "${directories[@]}"
-  do
-      mkdir -p $i
-      if [ "$(id -u)" = '0' ]; then
-          chown -R "${BK_USER}:${BK_USER}" $i
-      fi
-  done
-}
+# set -e
```
pbelgundi: Lets uncomment this line and things should work after the `sysctl` is properly fixed

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +49,136 @@ export BK_tlsTrustStoreType=JKS
...
+      sed -i "s|.*useHostNameAsBookieID=.*\$|useHostNameAsBookieID=${BK_useHostNameAsBookieID}|" ${BK_HOME}/conf/bk_server.conf
+    fi
+}
+
+disable_bk_ipv6_check() {
+   # This check done in "common.sh" in the Bookeeper image fails in environments where IPV6 is not enabled.
+   # We are disabling it TEMPORARILY. This needs to be reverted back/tackled in a better way, by modifying the
+   # check instead to also make sure that the file "/proc/sys/net/ipv6/bindv6only" exists, before it executes the
+   # "/sbin/sysctl -n net.ipv6.bindv6only" command.
+   sed -i "s|/sbin/sysctl|/sbin/sysctl2|" ${BK_HOME}/bin/common.sh
```
pbelgundi: Yes we need to add this check to fix it the right way....

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +49,136 @@ export BK_tlsTrustStoreType=JKS
...
+
+# Init the cluster if required znodes not exist in Zookeeper.
+# Use ephemeral zk node as lock to keep initialize atomic.
+function init_cluster() {
+    zk-shell --run-once "ls ${BK_zkLedgersRootPath}/available/readonly" ${BK_zkServers}
+    if [ $? -eq 0 ]; then
+        echo "Metadata of cluster already exists, no need format"
+    else
+        echo "Sleep for a random time"
+        sleep $[ ( $RANDOM % 10 )  + 1 ]s
```
pbelgundi: Is this random sleep needed? Does removing it cause issues with creation of the ephemeral node.

docker/bookkeeper/entrypoint.sh
```diff
@@ -8,31 +8,32 @@
...
 BK_DIR="/bk"
+BK_zkLedgersRootPath=${BK_LEDGERS_PATH}
+
+BK_HOME=/opt/bookkeeper
+BINDIR=${BK_HOME}/bin
+BOOKKEEPER=${BINDIR}/bookkeeper
+SCRIPTS_DIR=${BK_HOME}/scripts
 
+export PATH=$PATH:/opt/bookkeeper/bin
+export JAVA_HOME=/usr/lib/jvm/jre-1.8.0
```
andreipaduroiu: Are we sure these two paths will be there on the deployment server?

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +49,136 @@ export BK_tlsTrustStoreType=JKS
...
+    else
+        echo "Sleep for a random time"
+        sleep $[ ( $RANDOM % 10 )  + 1 ]s
+        # create ephemeral zk node bkInitLock, initiator who this node, then do init; other initiators will wait.
+        lock=`zk-shell --run-once "create ${BK_CLUSTER_ROOT_PATH}/bkInitLock1 '' true false false" ${BK_zkServers}`
+        echo "output of zk-shell create bkInitLock"
+        echo $lock
+        if [ -z "$lock" ]; then
+            # bkInitLock created success, this is the successor to do znode init
+            echo "Bookkeeper znodes not exist in Zookeeper, do the init to create them."
```
andreipaduroiu: The number of typos and grammatical errors in this file's strings made me think this is copied from somewhere. And indeed, this is from `init_bookie.sh` from the BK repository.

- Can we not make use of that file?
- If you do copy-paste all this code, please check out the copyright on the original file and under what conditions you are allowed (if at all) to copy the contents of your file for your own use.

NOTICE
```diff
@@ -1 +1,6 @@
 Copyright (c) 2017-2019 Dell Inc., or its subsidiaries. All Rights Reserved.
+
+This software contains source code from Apache BookKeeper, distributed under the Apache License Version 2.0, and
```
fpj: This line is overflowing, could you break it at 80 characters, please?

docker/bookkeeper/Dockerfile
```diff
@@ -1,12 +1,17 @@
...
-# Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+# Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
+# This file contains source code copied from the Apache BookKeeper project (http://bookkeeper.apache.org).
```
fpj: Could you separate this comment block from the license header block? Also, break the line earlier, like I mentioned in the previous comment.

docker/bookkeeper/entrypoint.sh
```diff
@@ -8,6 +8,10 @@
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
+# This file contains source code copied from the Apache BookKeeper project (http://bookkeeper.apache.org).
```
fpj: Same as above.

docker/bookkeeper/Dockerfile
```diff
@@ -7,31 +7,35 @@
...
+FROM apache/bookkeeper:4.9.2
 
-ARG BK_VERSION=4.7.3
+ARG BK_VERSION=4.9.2
 ARG DISTRO_NAME=bookkeeper-all-${BK_VERSION}-bin
-ARG GPG_KEY=FD74402C
 
 RUN set -x \
-    && yum install -y iproute \
+    && yum install -y iproute wget \
```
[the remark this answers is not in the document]
ravisharda: This change was not anything to do with security vulnerabilities, but rather to use the same tool that is used in the corresponding Bookkeeper image.

docker/bookkeeper/entrypoint.sh
ravisharda: Yes, that's a good point. I've pushed a commit `033931a`, which addresses this issue.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +53,131 @@ export BK_tlsTrustStoreType=JKS
...
+    if [ $? -eq 0 ]; then
+        echo "Cluster metadata already exists"
+    else
+        echo "Sleeping for a random time"
+        sleep $[ ( $RANDOM % 10 )  + 1 ]s
+
+        # Create an ephemeral zk node `bkInitLock` for use as a lock.
+
+        , initiator who this node, then do init; other initiators will wait.
+        lock=`zk-shell --run-once "create ${BK_CLUSTER_ROOT_PATH}/bkInitLock1 '' true false false" ${BK_zkServers}`
```
shrids: rename `bkInitLock1` to `bkInitLock`

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +53,131 @@ export BK_tlsTrustStoreType=JKS
...
+        if [ -z "$lock" ]; then
+            echo "Bookkeeper znodes do not exist in Zookeeper. Initializing a new Bookeekeper cluster."
+            /opt/bookkeeper/bin/bookkeeper shell initnewcluster
+            if [ $? -eq 0 ]; then
+                echo "initNewCluster operation succeeded"
+            else
+                echo "initNewCluster operation failed. Please check the reason."
+                echo "Exit status of initnewcluster"
+                echo $?
+                sleep 1000
```
shrids: we can remove this sleep.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +53,131 @@ export BK_tlsTrustStoreType=JKS
...
+                exit
+            fi
+        else
+            echo "Other containers may be initializing at the same time."
+            tenSeconds=1
+            while [ ${tenSeconds} -lt 1000 ]
```
shrids: 1000 can be reduced to 100.

docker/bookkeeper/Dockerfile
```diff
@@ -40,7 +44,4 @@ RUN chmod +x -R /opt/bookkeeper/scripts/
 ENTRYPOINT [ "/bin/bash", "/opt/bookkeeper/scripts/pravega_entrypoint.sh" ]
 CMD ["bookie"]
 
-# BookKeeper healthcheck was broken in 4.7.3 and was not fixed until 4.9.0
-# This overrides the healthcheck to the default BK healthcheck method
-# https://github.com/apache/bookkeeper/issues/1687
-HEALTHCHECK --interval=10s --timeout=60s CMD /opt/bookkeeper/bin/bookkeeper shell bookiesanity
+# HEALTHCHECK --interval=10s --timeout=60s CMD /bin/bash /opt/bookkeeper/scripts/healthcheck.sh
```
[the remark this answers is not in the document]
ravisharda: Yes, this change was a temporary change, which was meant to be reverted back. It is done in the latest commit that I have just pushed.

docker/bookkeeper/entrypoint.sh
ravisharda: Actually, it is not needed. It was among the temporary changes done for debugging purposes and has been removed in the latest commit that was just pushed.

docker/bookkeeper/Dockerfile
ravisharda: Sure, done.

docker/bookkeeper/entrypoint.sh
ravisharda: Done.

docker/bookkeeper/entrypoint.sh
ravisharda: Fixed.

docker/bookkeeper/entrypoint.sh
ravisharda: Sure, have done that in the latest commit I've just pushed.

docker/bookkeeper/entrypoint.sh
ravisharda: Sure, done.

docker/bookkeeper/entrypoint.sh
ravisharda: This change was a temporary one. It is resolved in one of the later commits that were pushed.

docker/bookkeeper/entrypoint.sh
ravisharda: The latest commit has the change that fixes this code.

docker/bookkeeper/entrypoint.sh
ravisharda: Yes, these paths are available in the container as of the v`4.9.2` of the Apache/Bookeeper image we use as the base image.

docker/bookkeeper/entrypoint.sh
```diff
@@ -1,38 +1,46 @@
...
+#       source code copied from file `init_bookie.sh`, revision `3bfecd4`:
+#  github.com/apache/bookkeeper/blob/branch-4.7/docker/scripts/init_bookie.sh.
+#
+
+# Exit the script in case of an error
+set -e
 
 BOOKIE_PORT=${bookiePort:-${BOOKIE_PORT}}
 BOOKIE_PORT=${BOOKIE_PORT:-3181}
+BOOKIE_HTTP_PORT=${BOOKIE_HTTP_PORT:-9090}
```
shrids: Should we turn this off using BK_httpServerEnabled config?

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +56,143 @@ export BK_tlsTrustStoreType=JKS
...
+                    echo "Waited $tenSeconds * 10 seconds. Successfully listed ''${BK_zkLedgersRootPath}/available/readonly'"
+                    break
+                else
+                    echo "Waited $tenSeconds * 10 seconds. Continue waiting."
+                    (( tenSeconds++ ))
+                    continue
+                fi
+            done
+
+            if [ ${tenSeconds} -eq 1000 ]; then
```
shrids: this needs to be changed to 100.

docker/bookkeeper/entrypoint.sh
ravisharda: I've modified the default port to the one used in Bookkeeper image.

docker/bookkeeper/entrypoint.sh
ravisharda: I've fixed the same in the latest commit that I just pushed.

fpj:
The issue associated with this PR says that it targets `r0.6`, but the base branch for the PR is set to `master`. Also, there is this comment by @andreipaduroiu that warns about merging to `master`:

https://github.com/pravega/pravega/pull/4420#pullrequestreview-326894567

could you please clarify, @ravisharda ?

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +56,143 @@ export BK_tlsTrustStoreType=JKS
...
+                echo "Waited 100 seconds for bookkeeper cluster to initialize, but to no avail. Something is wrong, please check."
+                exit
+            fi
+        fi
+    fi
+    set -e
+}
+
+format_bookie_data_and_metadata() {
+    if [ `find $BK_journalDirectory $BK_ledgerDirectories $BK_indexDirectories -type f 2> /dev/null | wc -l` -gt 0 ]; then
```
fpj: The comments in this block say "probably" a few times. It doesn't give a lot of confidence to the action being taken. Is there anything we can do to reduce the uncertainty?

We need to at least create issues to track these cases, assuming there are corner cases that can be problematic.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +56,143 @@ export BK_tlsTrustStoreType=JKS
...
+    else
+      # The container does not contain any BK data, it is probably a new
+      # bookie. We will format any pre-existent data and metadata before starting
+      # the bookie to avoid potential conflicts.
+      echo "Formatting bookie data and metadata"
+      /opt/bookkeeper/bin/bookkeeper shell bookieformat -nonInteractive -force -deleteCookie || true
+    fi
+}
+
+echo "Creating directories for Bookkeeper journal and ledgers"
```
fpj: In BookKeeper, we also have the option of setting the path for index directories. Is this a feature we might want to use in production, writing the ledger indexes to a separate directory/drive? If so, we need to provision for it.

docker/bookkeeper/entrypoint.sh
ravisharda: These are an existing piece of code and comments which were refactored into a function for modularity.

docker/bookkeeper/entrypoint.sh
ravisharda: I am not sure about this, but I suppose it might be useful to support locating ledger indexes to a specified directory. However, I want to scope of this PR limited to upgrade to Bookkeeper `4.9.2`. This is also one of the preexisting logic. Hope that is OK.

ravisharda:
> The issue associated with this PR says that it targets `r0.6`, but the base branch for the PR is set to `master`. Also, there is this comment by @andreipaduroiu that warns about merging to `master`:
> 
> [#4420 (review)](https://github.com/pravega/pravega/pull/4420#pullrequestreview-326894567)
> 
> could you please clarify, @ravisharda ?

@fpj Thanks for spotting the problem with the linked issue. I've now created a new issue #4459 that is targeted by this PR. Hope that helps. 

About the comment from @andreipaduroiu about merging to master, I believe at the time the PR was still a work-in-progress. Even now until we test it out fully, it shouldn't be merged. Especially, as we potentially might need to upgrade Bookkeeper client-side code as well, which is being tracked separately under PR #4394 .

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +56,143 @@ export BK_tlsTrustStoreType=JKS
...
+   #    `sysctl: cannot stat /proc/sys/net/ipv6/bindv6only: No such file or directory`
+   #
+   # If the `common.sh` file is sourced in a script that sets "exit on error" as true (set -e), the script shall return
+   # an error code `255`. Since, the `bookkeeper/bin/bookeeper` shell script sources the `common.sh` file, and it sets
+   # "exit on error" as true, we encounter this error when it is invoked. To avoid that error, here we modify the
+   # `common.sh` file to execute `/sbin/sysctl -n net.ipv6.bindv6only` only if file `/proc/sys/net/ipv6/bindv6only`
+   # is available.
+
+   # Replace the 22nd line with "if [ -f /sbin/sysctl ] && [ -f /proc/sys/net/ipv6/bindv6only ]; then".
+   sed -i "22s|.*|if [ -f /sbin/sysctl ] \&\& [ -f /proc/sys/net/ipv6/bindv6only ]; then|" ${BK_HOME}/bin/common.sh
```
pbelgundi: The purpose of check `/sbin/sysctl -n net.ipv6.bindv6only` in common.sh, is to error-out if IPV6 is enabled - indicated by `sysctl -n net.ipv6.bindv6only` giving ouput `net.ipv6.bindv6only=1`.
In our case this command fails because the key "net.ipv6.bindv6only" is not set in sysctl. IMO, instead of checking for existence of file `/proc/sys/net/ipv6/bindv6only` which is an implementation detail of `sysctl` command, it may be better to skip failing the script if output of `/sbin/sysctl -n net.ipv6.bindv6only` is non zero, indicate command did not succeed.
This can be done by adding "set -e" @ ln 23 and "set +e" at line 25 in [common.sh](https://github.com/apache/bookkeeper/blob/release-4.9.2/bin/common.sh#L25), so that the script does not exit when `/sbin/sysctl -n net.ipv6.bindv6only` gives non-zero exit code.

docker/bookkeeper/Dockerfile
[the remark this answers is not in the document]
pbelgundi: ok

docker/bookkeeper/entrypoint.sh
ravisharda: Turning off exit-on-error flag (`set +e`) in the `common.sh` temporarily before executing [this ](https://github.com/apache/bookkeeper/blob/release-4.9.2/bin/common.sh#L25) line in is only a workaround. So is true with what is being done in the code referenced above (where we are replacing the if condition to include check for `/proc/sys/net/ipv6/bindv6only`. 

IMO, the latter is better due to the following: 
* It modifies the specific line that has the problem and therefore the workaround is more explicit about what the problem is.
* It is closer to what the ultimate fix may be. Witness [this ](https://github.com/apache/bookkeeper/blob/a7855d994ee70b22d12223dc14da602da9b26196/bin/common.sh#L21) code in a Bookkeeper PR that I've raised.  Note that ultimately, the `set -e` is done in other Bookkeeper scripts that invoke the `common.sh` (and not in the `common.sh` file itself), so adding these flags in the `common.sh` file would be odd.

pbelgundi: I do not agree that adding "set -e" is a "workaround" and there is nothing "odd" about using it.
This the standard way to ignore failures in commands that are likely to fail for reasons unrelated to the script.
Infact checking for  "/proc/sys/net/ipv6/bindv6only" is a **hack** as the location of that file may be different on different platforms.

ravisharda: It is workaround in the sense that we let the error occur and then ignore it. The change in condition, on the other hand, prevents the error from happening in the first place, and is still a workaround in the sense that it is compensating for the bug in `common.sh`. 

Are you sure that the path `/proc/sys/net/ipv6/bindv6only` is different on different platforms? And, can this path change with the Bookkeeper container running on those platforms, given that it is based of centos (https://github.com/apache/bookkeeper/blob/44a1a4e986e22d87ba340069a7b5619feb30f0dc/docker/Dockerfile#L20)?

docker/bookkeeper/entrypoint.sh
Tristan1900: Sorry for the late review, I remember this issue can be traced back several months ago when Adrian found that if we don't format bookie when bookie starts, the system will have InvalidCookieException if there are some metadata of the previous deployment left in the zookkeeper. However, we don't want to format the bookie when the bookkeeper container is just restarted or upgraded, otherwise it will cause data loss.

The main purpose here is basically to use `bookieformat -deleteCookie ` before a new deployment to prevent InvalidCookieException.

docker/bookkeeper/entrypoint.sh
Tristan1900: The reason to create customized journal and ledger files here is to support multi ledger/journal for better write performance as we find that by increasing the parallasim of writing to thoses files, the performance will be increased. However, during the experiment I find that index dir have a very low write throughput so it shouldn't have any negative effect to the performance by just using the default settings.

docker/bookkeeper/entrypoint.sh
ravisharda: Thanks for the clarification, @Tristan1900. @fpj I've updated the comments in the latest commit. Please see if that resolves your concern here.

docker/bookkeeper/entrypoint.sh
ravisharda: Thanks @Tristan1900. @fpj I've added comments to explain the context in the code. Hope that helps.

NOTICE
```diff
@@ -1 +1,6 @@
 Copyright (c) 2017-2019 Dell Inc., or its subsidiaries. All Rights Reserved.
+
+This software contains source code from Apache BookKeeper, distributed under
```
fpj: Please do not touch the NOTICE file, this is for legal compliance and we shouldn't be changing it as part of this PR.

docker/bookkeeper/Dockerfile
```diff
@@ -1,37 +1,51 @@
...
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
-FROM apache/bookkeeper:4.7.3
 
-ARG BK_VERSION=4.7.3
+#
+# Note: This file contains source code copied from the Apache BookKeeper
```
fpj: I assume that the `4.9` revision still applies to this PR. For consistency with the `4.8.2` version we are changing to, I wonder if it is best to refer to the `4.8` branch.

docker/bookkeeper/Dockerfile
```diff
@@ -1,37 +1,51 @@
...
-ARG GPG_KEY=FD74402C
 
 RUN set -x \
-    && yum install -y iproute \
+    && yum install -y iproute wget \
     && cd /opt \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.asc" \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.sha512" \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
```
fpj: The change from `curl` to `wget` seems gratuitous in this PR. Is it necessary for the upgrade or just convenient?

docker/bookkeeper/entrypoint.sh
fpj: I'm in general skeptical of ignoring errors because it is easy to overlook some important error and get bitten at some point. If we can preserve exit-on-error while ignoring only the specific error we are trying to, then I'd say that's a safer option.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +56,151 @@ export BK_tlsTrustStoreType=JKS
...
+
+initialize_cluster() {
+    set +e
+
+    zk-shell --run-once "ls ${BK_zkLedgersRootPath}/available/readonly" ${BK_zkServers}
+    if [ $? -eq 0 ]; then
+        echo "Cluster metadata already exists"
+    else
+        # Create an ephemeral zk node `bkInitLock` for use as a lock.
+        lock=`zk-shell --run-once "create ${BK_CLUSTER_ROOT_PATH}/bkInitLock '' true false false" ${BK_zkServers}`
```
fpj: I don't understand this. If the script is set to run once, then I interpret it as it is not going to hold a zk session. If it is not holding a bk session, then all ephemerals will be deleted upon the script completing its execution and there will be no lock whatsoever. This smells like a bug.

Also, where is unlock being implemented? For an ephemeral lock znode, either it is deleted explicitly or it is gone when the session closes. Which one is happening here?

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +56,151 @@ export BK_tlsTrustStoreType=JKS
...
+                exit
+            fi
+        fi
+    fi
+    set -e
+}
+
+format_bookie_data_and_metadata() {
+    if [ `find $BK_journalDirectory $BK_ledgerDirectories $BK_indexDirectories -type f 2> /dev/null | wc -l` -gt 0 ]; then
+      # The container already contains data in BK directories. Examples of when this can happen include:
```
fpj: Say there is some kind of partial loss and some of the directories contain data but others lost it. What is the expected behavior in this situation?

fpj:
The description still has some references to `4.9`, e.g., :

> The 4.9.2 script exits if...

NOTICE
ravisharda: @fpj My understanding was that we needed this statement in the `NOTICE` file to address copyright concerns. No?

> ...
> Add a note in the NOTICE file saying that this software includes software copyrighted to the ASF from Apache BookKeeper.
> ...

docker/bookkeeper/Dockerfile
ravisharda: Yes, it'd have been nice to have a branch name that represented the now modified target version `4.8.2`. But modifying the remote branch (using steps such as those described [here](https://linuxize.com/post/how-to-rename-local-and-remote-git-branch/) seems disruptive to this PR and its comments/history. What do you say?

docker/bookkeeper/Dockerfile
ravisharda: It is just convenient. The code in the `apache/bookkeeper` image [uses](https://github.com/apache/bookkeeper/blob/branch-4.8/docker/Dockerfile) `wget` as well, so it is consistent with what's used there and makes it easier to see the changes in our image on top of `Apache's`.

docker/bookkeeper/entrypoint.sh
ravisharda: The idea here is that if there are multiple containers coming up simultaneously, one of them reaches here, holds the lock and performs bookkeeper init cluster, while the rest of the containers wait for the initialization to get completed. 

To answer your question: the latter is happening. 

To provide a bit more background: The same mechanism is present in `apache/bookkeeper:4.7.3`: https://github.com/apache/bookkeeper/blob/f307a36acf9d2acf5c3ff315e1b3bfd7873e3cce/docker/scripts/init_bookie.sh#L43. 

In `apache/bookkeeper: 4.8.2`, the code uses `bookkeeper org.apache.zookeeper.ZooKeeperMain` command, but that command doesn't work, so we had to bring in `zk-shell` back in our implementation. 
https://github.com/apache/bookkeeper/blob/4fa14617a855df236a902e76c8c68a21952bb48a/docker/scripts/init_bookie.sh#L50

docker/bookkeeper/entrypoint.sh
ravisharda: In a scenario where a subset of these directories has lost data, based on the code, the `else` block is not executed. So, 
* The current bookie's server contents are not deleted
* The current bookie's cookie is not deleted in Zookeeper

The expected behavior would be that the bookie might fail to start-up unless the directory that was lost is in a recoverable state. 

It should also be noted that this is a preexisting code that was extracted into a function to make it easier to read.

docker/bookkeeper/Dockerfile
ravisharda: Upon reading your comment again, I realized that your statement was really about the `apache/bookkeeper` source code reference in `Dockerfile`. 

Yes, the `4.9` revision still applies, as we have borrowed from that version of the code over here.

docker/bookkeeper/entrypoint.sh
ravisharda: @pbelgundi I've found that the procfs path is really a pseudo file-system that exists only on memory. See https://wiki.archlinux.org/index.php/Procfs. So, we can indeed depend on the presence on `/proc/sys/`. 

@fpj I agree with your last statement and believe that the current code is in line with it. I suppose you aren't proposing any change.

docker/bookkeeper/Dockerfile
```diff
@@ -40,7 +54,4 @@ RUN chmod +x -R /opt/bookkeeper/scripts/
 ENTRYPOINT [ "/bin/bash", "/opt/bookkeeper/scripts/pravega_entrypoint.sh" ]
 CMD ["bookie"]
 
-# BookKeeper healthcheck was broken in 4.7.3 and was not fixed until 4.9.0
-# This overrides the healthcheck to the default BK healthcheck method
-# https://github.com/apache/bookkeeper/issues/1687
-HEALTHCHECK --interval=10s --timeout=60s CMD /opt/bookkeeper/bin/bookkeeper shell bookiesanity
+HEALTHCHECK --interval=10s --timeout=60s CMD /bin/bash /opt/bookkeeper/scripts/healthcheck.sh
```
RaulGracia: According to the previous comment here, the health check was not fixed until `4.9.0`. But this PR is now targeting 4.8.2. Are you sure that this works in its current form?

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +56,151 @@ export BK_tlsTrustStoreType=JKS
...
+                    break
+                else
+                    echo "Waited $tenSeconds * 10 seconds. Continue waiting."
+                    (( tenSeconds++ ))
+                    continue
+                fi
+            done
+
+            if [ ${tenSeconds} -eq 100 ]; then
+                echo "Waited 100 seconds for bookkeeper cluster to initialize, but to no avail. Something is wrong, please check."
```
RaulGracia: Maybe this is just a mistake in the message, but if the script reaches this point it means that it waited 1000 seconds, not 100 (each iteration waits for 10 seconds, so if `tenSeconds==100`, this is 1000 seconds). Could it be too much?

docker/bookkeeper/Dockerfile
ravisharda: @RaulGracia yes you are right. This was an oversight. I've fixed it now.

docker/bookkeeper/entrypoint.sh
ravisharda: Good catch! :) Thanks for spotting it. I've reduced both the wait time and the echo message.

RaulGracia:
With commit `ff8f7bc`, this PR has passed 2 system tests builds (with and without security). Please @pbelgundi @fpj, could you do one more pass to see if there is anything else to address? Thanks!

ravisharda:
With commit `a769ddd` too, a system test build (with auth enabled) passed successfully. FYI.

docker/bookkeeper/entrypoint.sh
pbelgundi: @fpj we're not  **ignoring** an error here, but avoiding the script from exiting because a certain command fails in the absence of `IPV6`, which is a very valid thing to do....
I still think adding "set -e"  and "set +e" is the right way to solve the problem.

However, it isn't so severe that I would want to block the approval of this PR for the same.
@ravisharda I would suggest you think through this and decide.

NOTICE
fpj: I didn't read the addition carefully, you are right.

docker/bookkeeper/entrypoint.sh
fpj: > The idea here is that if there are multiple containers coming up simultaneously, one of them reaches here, holds the lock and performs bookkeeper init cluster, while the rest of the containers wait for the initialization to get completed.

That's not what I'm asking, please read my comment again.

> To answer your question: the latter is happening.

The latter what? If the shell is running in non-interactive mode (`run-once`), then it is not holding a session, which means that ephemerals are gone and the lock is not locking anything.

It seems broken to me.

docker/bookkeeper/entrypoint.sh
fpj: > The expected behavior would be that the bookie might fail to start-up unless the directory that was lost is in a recoverable state.

That's right, see the comment in the BK code:

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java#L59

This is not new code, so let me create an issue to investigate this.

docker/bookkeeper/entrypoint.sh
ravisharda: I ran a quick experiment to test whether the logic works or not. The minor changes I did for debugging were:
* Added "set -x" to the script so that all commands are printed in the logs
* Added an additional `echo` statement to print the lock (as shown below)
* Added an additional `zk-shell ls` command (as shown below) to check if the `bkInitLock` is seen by nodes which don't create it.

```
...
 # Create an ephemeral zk node `bkInitLock` for use as a lock.
 lock=`zk-shell --run-once "create ${BK_CLUSTER_ROOT_PATH}/bkInitLock '' true false false" ${BK_zkServers}`
 echo "Lock: $lock" <--- Added this line 
 zk-shell --run-once "ls ${BK_CLUSTER_ROOT_PATH}/" ${BK_zkServers} <--- Added this line
...
```
The behavior I found upon looking at the logs seems correct: one node  (BK 2) creates the `/pravega/pravega-cluster/bookkeeper/bkInitLock` path and initializes the cluster, and the others (BK 1 and BK 3) find that it is already present and don't initialize the cluster. 

```
root@pvgcompose:~/bk_upgrade_pr# grep "Lock:" *.log
bk1_23-13-15--2020-01-21.log:Lock: Path /pravega/pravega-cluster/bookkeeper/bkInitLock already exists
bk2_23-13-15--2020-01-21.log:Lock:
bk3_23-13-15--2020-01-21.log:Lock: Path /pravega/pravega-cluster/bookkeeper/bkInitLock already exists
```

The node that creates the lock has the following in the logs: 
```
Lock:
bkInitLock
Bookkeeper znodes do not exist in Zookeeper. Initializing a new Bookeekeper cluster.
initnewcluster operation succeeded
```
The nodes that find the lock is already established have this in the logs: 
```
Lock: Path /pravega/pravega-cluster/bookkeeper/bkInitLock already exists
bkInitLock
Another instance might be initializing the cluster at the same time.
```

So, the logic seems to be working fine here, from what I can see. This piece of code is same as what you can see in `apache/bookkeeper` code here: 
https://github.com/apache/bookkeeper/blob/2fb7f3a80561eb70dc063c89763d41519fb45f1e/docker/scripts/init_bookie.sh#L43

fpj: Even though this is used in the BK code, I'm not convinced this is entirely correct. One additional point to support what I'm saying is that neither the `bkInitLock` znode is deleted nor a session is closed explicitly.

To recap the semantics of ephemerals, a znode exists for as long as its session is considered alive. It is possible that the session within which the znode was created was not closed and the ephemeral is not deleted while it is alive, but it will eventually timeout, and if it times out before we complete the operation, then the lock will fail. If for some reason we have to acquire the lock again within a short period, then the creation will fail incorrectly.

We need to be able to acquire and release explicitly, and we need to make sure that we hold the session for the ephemeral while the lock is acquired. I'm not up-to-speed with the syntax and semantics of `zk-shell`, perhaps we can have a look to see how to implement this correctly?

ravisharda: @fpj would it be OK to explore what you have suggested as part of a separate issue, especially given the fact that this code has been working for us via the `apache/bookkeeper` `4.7.3` image code previously and now in this `pravega/bookkeeper` image as seen via manual tests and automated system tests? I haven't been able to find good documentation on `zk-shell` and believe it'll require some experimentation to understand the behavior of its commands precisely. What do you say?

ravisharda: I don't have a definitive answer right now and am digging further. I've found that the `bkInitLock` is indeed an ephemeral znode that goes away shortly after the node is created. So, there is certainly a potential for a race condition causing issues here. Copying the result of the experiment here (for reference): 

```
# Command format: create <path> <value> [ephemeral] [sequence] [recursive] [async]
# All [..] are false by default
$ zk-shell --run-once 'create /test1 '\'''\'' false false false' zookeeper:2181
$ zk-shell --run-once 'ls /' zookeeper:2181
test1
$ zk-shell --run-once 'ls /test1' zookeeper:2181
[empty]
$ zk-shell --run-once 'create /test1/bkInitLock '\'''\'' true false false' zookeeper:2181
$ zk-shell --run-once 'ls /test1/' zookeeper:2181
bkInitLock
$ zk-shell --run-once 'ls /test1/' zookeeper:2181
bkInitLock
$ zk-shell --run-once 'ls /test1/' zookeeper:2181
bkInitLock
$ zk-shell --run-once 'ls /test1/' zookeeper:2181
bkInitLock
$ zk-shell --run-once 'ls /test1/' zookeeper:2181
[empty]
```

ravisharda: Confirming that system tests have passed with these changes. The run was: `test-pravega-continuous/2589/` and it comprised of auth-enabled Pravega cluster.

docker/bookkeeper/entrypoint.sh
```diff
@@ -122,10 +124,14 @@ initialize_cluster() {
             /opt/bookkeeper/bin/bookkeeper shell initnewcluster
             if [ $? -eq 0 ]; then
                 echo "initnewcluster operation succeeded"
+                zk-shell --run-once "rm ${BK_CLUSTER_ROOT_PATH}/bkInitLock" ${BK_zkServers}
```
fpj: Alright, this looks better, but what happens if the process running this script crashes between creating the lock znode and deleting it? Are we going to be in a situation that the lock is held forever by a non-existing process?

ravisharda: If that happens, the persistent node will be left behind in Zookeeper. 
A remedy would be to manually deleting that znode from Zookeeper cluster.

fpj: Yeah, I don't think this is going to work well. Here is a different proposal:

1- Change it back to ephemeral
2- Run `zk-shell` without `--run-once` and in a background process. Make sure to get the PID of the background process right after creating it.
3- Once you are done with the bk initialization, kill PID.

With this proposal, if the container crashes, then all processes die and the session is eventually closed removing the ephemeral. In the absence of crashes, it removes the session after being done with the initialization. The only missing case that is not covered is the one of the parent process dying. In this case, the child process becomes orphaned and the session is never closed. If we can find a way to have the child process exiting with the parent, then this could be a nice solution.

fpj: Ok, I think this SO thread gives a solution to the last point by using `trap`:

https://stackoverflow.com/questions/360201/how-do-i-kill-background-processes-jobs-when-my-shell-script-exits/2173421

See if this new proposal works, please.

ravisharda: @fpj I think this change would be a significant one (especially w.r.t. testing negative scenarios), and we should do this as part of a separate enhancement issue. 

I can create an enhancement issue if you agree. Better still, we can try this strategy out separately in the Apache/Bookkeeper image improvement PR that is currently in progress: https://github.com/apache/bookkeeper/pull/2219.

fpj: I don't know why that's a significant change, it is a four-line change!

1- Change it back to ephemeral -> Only changes a parameter in the existing command
2- Run zk-shell without --run-once and in a background process. Make sure to get the PID of the background process right after creating it. -> Changes the command and you have to add `&` to the end of the command. Get the PID of the background process with `$!`, which another line but only an assignment.. 
3- Once you are done with the bk initialization, kill PID. -> Replace the deletes with kill PID.
4- Add a trap call to kill the child in the case of exiting prematurely.

Have you even tried?

ravisharda: I meant it is a significant change in the sense of target behavior and the impact on testing. And, yes I have been trying to make that change - not done yet.

ravisharda: When trying to test out the suggestion, I found this issue. We need the output of the `zk-shell` command (the exit status is always `0`) to verify if the lock was created successfully, which cannot be done using `trap` AFAIK. Am I missing something?

https://github.com/pravega/pravega/blob/6da3c0ed1627f6be9c44a37023b8774ba2efcc4b/docker/bookkeeper/entrypoint.sh#L116

Besides, taking a step back and reconsidering the discussion here, I realize that the original code that created the ephemeral node inline was just fine. Here's why: 

1. **Scenario 1:** All nodes attempt to come up simultaneously. All see `${BK_zkLedgersRootPath}/available/readonly` znode missing (as the cluster is not initialized yet). One node successfully creates `${BK_CLUSTER_ROOT_PATH}/bkInitLock` znode and initializes the cluster, others see that `${BK_CLUSTER_ROOT_PATH}/bkInitLock`znode is already present, so they don't attempt to initialize the cluster. Business as usual, and all nodes come up successfully.

2. **Scenario 2:** One node comes up first, and creates the `${BK_CLUSTER_ROOT_PATH}/bkInitLock` znode, successfully initilizes the cluster (which in turn successfully creates `${BK_zkLedgersRootPath}/available/readonly`). Others come up later, they see `${BK_zkLedgersRootPath}/available/readonly`, so they don't attempt to initialize the cluster and run successfully. 

3. **Scenario 3:** All nodes see `${BK_zkLedgersRootPath}/available/readonly` znode missing (so the cluster is not initialized yet). All of them even find `${BK_CLUSTER_ROOT_PATH}/bkInitLock` znode missing, so each one tries to create the znode. Only one succeeds in creating the znode and it also initializes the cluster, the others get a non-empty output when attempting to create the znode, so they exit. The orchestration engine (Kubernetes/Docker Swarm) brings the instance back up and they start running successfully, as they find `${BK_zkLedgersRootPath}/available/readonly` znode is present and don't attempt to initialize the cluster this time. 

4. **Scenario 4:** All nodes see `${BK_zkLedgersRootPath}/available/readonly` znode missing (so the cluster is not initialized yet). All of them even find `${BK_CLUSTER_ROOT_PATH}/bkInitLock` znode missing, so each one tries to create the znode. Only one succeeds in creating the znode and it also initializes the cluster. By the others wake up from slumber, the ephemeral `${BK_CLUSTER_ROOT_PATH}/bkInitLock` is gone, and therefore they try to initialize the cluster. Init cluster operation fails and they exit. Like in the previous case the orchestration engine brings the instance back up and they start running successfully, as they find `${BK_zkLedgersRootPath}/available/readonly` znode is present and don't attempt to initialize the cluster this time.

fpj: If what you say above is true, in particular for scenario 4, then we don't need a lock at all. If we do need a lock, then reverting the change will give you a broken lock implementation. The question we need to answer is whether the lock is strictly needed or not.

ravisharda: Without the `${BK_CLUSTER_ROOT_PATH}/bkInitLock`, in the scenarios where all instances didn't find that the cluster was already initialized, n-1 instances will always unnecessarily attempt to initialize the cluster, wait for the operation to fail, and will come up only upon restart. By use of the lock, we reduce (but not eliminate) the chances of that happening. The downside of the current lock implementation is that it doesn't strictly prevent such situations, but it does reduce such situations. 

Nonetheless, I see merit in strengthening the lock. I suggest we do that as part of a separate issue.

fpj: Is it the case that if two instances try to initialize concurrently, then only one will succeed? If that's the case, then mutual exclusion is already implemented, and we don't strictly need a lock. Could you please confirm that this interpretation is correct? 

If it is correct, then the "lock" is not really a lock, it is only a convenient signal to avoid unnecessary restarts, and as such, it can remain as is.

ravisharda: Yes, that is right - if two instances attempt to initialize concurrently, only one will succeed. The others will exit (and will be restarted by the orchestrator). 

https://github.com/pravega/pravega/blob/6da3c0ed1627f6be9c44a37023b8774ba2efcc4b/docker/bookkeeper/entrypoint.sh#L133

Sure, it sounds more like a signal. :)

I'll revert it back to ephemeral znode as before.

ravisharda: Done.

docker/bookkeeper/entrypoint.sh
ravisharda: This conversation continued here: https://github.com/pravega/pravega/pull/4420#discussion_r370048900.

docker/bookkeeper/entrypoint.sh
fpj: > Yes, that is right - if two instances attempt to initialize concurrently, only one will succeed. 

What evidence do you have that this is correct and that concurrent attempts won't lead to some form of unexpected corruption due to a race?

ravisharda: I don't have hard evidence that anything wrong cannot happen. What I said is based on the logic in the code, as well as casual observations made from looking at the logs from various runs during this effort of fixing the BK image.

fpj: This is the code that initializes a new cluster:

https://github.com/apache/bookkeeper/blob/branch-4.8/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java#L419

There are two safety guards. First, it checks whether the ledger root path exists. It could happen that two instances cross that check concurrently, though. Second, the operations that need to be executed against zookeeper are executed in a multi transaction (all or nothing). If one fails, then none is applied and the call throws, which will happen in the case of concurrent attempts.

This code is safe without the `bkInitLock` that is being discussed here.

PR #4421, opened 2019-12-06
Issue 4403: Add deadline to all grpc calls to controller

### LINKED ISSUE #4403: StreamManager.sealStream() never times out
**Problem description**
A call to `sealStream` seems to block indefinitely (never times out), for an unknown reason.

**Problem location**

Ensure that the `StreamManager` respects a configured RPC timeout.

shiveshr:

**Change log description**  
Add deadline to all Grpc calls on client. 

**Purpose of the change**  
Fixes #4403 

**What the code does**  
In ControllerImpl we now set deadling on grpc client stub.
The deadline is set in Controller Client's Config. 

**How to verify it**  
Unit test added.


client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java
```diff
@@ -130,7 +130,8 @@
 
     // Generate random numbers for request ids.
     private final Supplier<Long> requestIdGenerator = RandomFactory.create()::nextLong;
-
+    
+    private final long deadline;
```
andreipaduroiu: You should call this timeout and add the unit of measure suffix (`timeoutMillis`). Deadline is not really a commonly used term.

PR #1534, opened 2017-07-05
Issue 1531: Controller gRPC client to retry DNS resolution on failures

skrishnappa:
**Change log description**
Updated the gRPC nameresolver to retry DNS resolution on network failures. Also attempting DNS resolution at regular intervals to detect change in address (when dynamic DNS is used)

**Purpose of the change**
Fixes #1531 

**What the code does**
Attempts DNS resolution on gRPC network failures and also at regular intervals (retained the earlier 2 mins scheduler used for controller discovery)

**How to verify it**
Existing ControllerImplLBTest.java covers network failure scenarios. Manually tested failure handling due to DNS propagation delays. Will explore if I can add unit tests for this.

PR #4423, opened 2019-12-06
NPS-5581 Patch bookkeeper image for issue 1534

gaddamas:
**Change log description**  
cherry pick bookkeeper bug fix #1534

**Purpose of the change**  
to include in pravega build

**What the code does**  
builds pravega with bookkeeper patch

**How to verify it**  
exec into bookkeeper pod and verify md5sum
md5sum lib/org.apache.bookkeeper-bookkeeper-server-4.7.3.jar
8f312a041ad6250e9bc8faeeacb2eb3a lib/org.apache.bookkeeper-bookkeeper-server-4.7.3.jar


andreipaduroiu:
Also your DCO is missing.

RaulGracia:
@gaddamas I think that @andreipaduroiu is right, this belongs to an internal repo, as the general public does not have access to the patched 4.7.3 version of Bookkeeper. We probably will update Bookkeeper in Pravega `0.6.1` to Bookkeeper 4.8.2 to get the same fix.

Issue #3503, opened 2019-03-21
Intermittent failure in io.pravega.segmentstore.server.host.HDFSIntegrationTest


sachin-j-joshi:
**Problem description**
Intermittent failure in io.pravega.segmentstore.server.host.HDFSIntegrationTest.
This test fails occasionally.

           io.pravega.segmentstore.server.host.HDFSIntegrationTest > testEndToEnd FAILED
     java.util.concurrent.CompletionException: 
           io.pravega.segmentstore.contracts.StreamSegmentNotExistsException: [Segment 'Segment_0$offset.0'] The StreamSegment does not exist.
 
                  Caused by:
                  io.pravega.segmentstore.contracts.StreamSegmentNotExistsException: [Segment 'Segment_0$offset.0'] The StreamSegment does not exist.
 
                      Caused by:
                      java.io.FileNotFoundException: Segment_0$offset.0

**Problem location**
io.pravega.segmentstore.server.host.HDFSIntegrationTest

**Suggestions for an improvement**
Need to repro this by continuously running the test in loop until it fails.

RaulGracia:
While the situation with this test has improved recently with #3294, it is a recurrent issue: #2944, #2359, #2250, #1757.

RaulGracia:
Closed via #4408

PR #4427, opened 2019-12-09
Issue 3503: Disable HDFSIntegrationTest.

sachin-j-joshi:
Issue 3503: Disable HDFSIntegrationTest.

**Change log description**  
The HDFSIntegrationTest has been unstable and fails with intermittent failures. This causes builds with changes unrelated to HDFS to fail. 
Disabling HDFSIntegrationTest for a short while until it gets completely rewritten/replaced as part of PDP-34.

**Purpose of the change**  
Short term mitigation for #3503 until HDFSIntegrationTest gets completely rewritten/replaced as part of PDP-34

**What the code does**  
Disable HDFSIntegrationTest.

**How to verify it**  
HDFSIntegrationTest should not be invoked.

### DIFF
--- a/segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/HDFSIntegrationTest.java
+++ b/segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/HDFSIntegrationTest.java
@@ -22,10 +22,12 @@ import lombok.val;
 import org.apache.hadoop.hdfs.MiniDFSCluster;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 
 /**
  * End-to-end tests for SegmentStore, with integrated Storage and DurableDataLog.
  */
+@Ignore ("Short term mitigation for #3503 until HDFSIntegrationTest gets completely rewritten/replaced as part of PDP-34.")
 public class HDFSIntegrationTest extends BookKeeperIntegrationTestBase {
     //region Test Configuration and Setup
 


RaulGracia:
@sachin-j-joshi if this test is going to be removed with PDP-34, why don't we wait until those changes? We have been for a long time with this intermittent issue and we have never decided to remove the test.  Why do we need to do this now that the test is going to be removed soon? Just wanted to understand the rationale and timing for this change to come.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/HDFSIntegrationTest.java
```diff
@@ -22,10 +22,12 @@
 import org.apache.hadoop.hdfs.MiniDFSCluster;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 
 /**
  * End-to-end tests for SegmentStore, with integrated Storage and DurableDataLog.
  */
+@Ignore
```
kevinhan88: Please add some background information to make it easier to bring this test back in the future, if at all.

sachin-j-joshi:
@RaulGracia It was a team decision to disable it for now. Rationale being that it is unnecessarily causing build failures for changes totally unrelated to segment store or HDFS.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/HDFSIntegrationTest.java
sachin-j-joshi: The PR and commit message already provides background.

kevinhan88: Then why not just add the explanation below into comment? 
In the future when people get back to this test for whatever reason, they will wonder what to do with it - bring it back? keep ignoring? improve it? or simply delete it?
PR and commit message don't stay with code always.

sachin-j-joshi: Kevin I've already fixed it as per your suggestion!

PR #4431, opened 2019-12-10
Issue 3992: (DR) Reconcile BookKeeperLog with BookKeeper Ledgers

### LINKED ISSUE #3992: Reconcile BookKeeperLog Metadata
**Problem description**
Handle the following scenarios in terms of BookKeeper data recovery:
1. BookKeeperLog Metadata and all BookKeeper Ledgers are lost.
2. BookKeeperLog Metadata is lost, but BookKeeper Ledgers are still available.
3. BookKeeperLog metadata is available, but lost one or more (or all) BookKeeper Ledgers

NOTE: the scope of this is constrained to the BookKeeperLog. This does not include any upstream changes that are needed to handle data loss coming from such a case.

**Problem location**
BookKeeperLog.

1. Nothing that can be done.
2. BookKeeper Ledgers support custom metadata. Add a field that encodes the BookKeeperLog id.
3. Remove references to unrecoverable ledgers from the metadata.

andreipaduroiu:
**Change log description**  
- Added an API to `BookKeeperLog` (`DebugLogWrapper`) to allow reconstructing the `BookKeeperLog` metadata (stored in ZK) in case it got wiped out for whatever reason.

**Purpose of the change**  
Fixes #3992.

**What the code does**  
- Every BK ledger gets a piece of metadata that stores the BookKeeperLog ID that it belongs to.
- If a reconciliation/reconstruction is needed, the API will look for all ledgers that match the BookKeeperLog ID to reconstruct and rebuild the latter with information from it.
    - Ledgers are added in order of their IDs
    - Empty ledgers are ignored.

**What this doesn't do**
- Does not "open up" ledgers to validate or repair their contents.
- Does not work if the BK Ledger metadata is missing the information
- Does not account for missing ledgers in the BookKeeperLog chain
    - This would be an extra piece of work that should be tracked in a different issue.
- Does not update the Pravega CLI to make use of this.
    - The CLI will need to add support for this.

**How to verify it**  
Unit test added to verify feature.


segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -187,4 +198,28 @@ static void delete(long ledgerId, BookKeeper bookKeeper) throws DurableDataLogEx
...
+
+    /**
+     * Creates a new {@link Map} that can be set as a Ledger's Custom Metadata.
+     *
+     * @param logId The {@link BookKeeperLog} Id to encode.
+     * @return An immutable {@link Map} containing the encoded Ledger's Custom Metadata.
+     */
+    private static Map<String, byte[]> createLedgerCustomMetadata(int logId) {
+        byte[] logIdSerialized = new byte[Integer.BYTES];
+        BitConverter.writeInt(logIdSerialized, 0, logId);
```
eolivelli: What about writing this as a string?
The tools I know that manage custom metadata actually assume that the values of the map are utf8 encoded strings.
It is not a big waste to use strings because this info is read/written out of the hot read/write paths

andreipaduroiu:
@RaulGracia I am not comfortable doing anything in an automated fashion. While doing something like this automatically would not even register on the logs as an error, what I am concerned is about deleting some valid ledger or ledger pointer which can lead to real data loss.

For now, I am backing that all DR efforts be built into Pravega, but they would have to be triggered by a human after they double-check the state of the system. We can think about adding more DR automation as we get more experience with such scenarios.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
andreipaduroiu: Interesting idea. Let me see what I can do there. That would also make it easier to edit the data directly in ZK, without the use of this tool.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/DebugLogWrapper.java
```diff
@@ -128,6 +152,98 @@ public void disable() throws DurableDataLogException {
...
+                // No information.
+                highestLedgerId = Ledgers.NO_LEDGER_ID;
+            }
+        } else {
+            // No metadata.
+            highestLedgerId = Ledgers.NO_LEDGER_ID;
+        }
+
+        // First, we filter out any Ledger that does not reference this Log as their owner or that are empty.
+        candidateLedgers = candidateLedgers
```
sachin-j-joshi, line 193: What happens with upgrade here?

cc: @RaulGracia

andreipaduroiu: Nothing too exciting. This will not work on ledgers that are not tagged, which is why this is important to go in well before any attempt to fix any issues so that each ledger in the system is properly tagged.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -187,4 +198,40 @@ static void delete(long ledgerId, BookKeeper bookKeeper) throws DurableDataLogEx
...
+        if (logIdSerialized == null) {
+            log.warn("No property '{}' found on Ledger {}. This is OK if this ledger was created prior to Pravega 0.7.1.",
+                    PROPERTY_LOG_ID, handle.getId());
+            return NO_LOG_ID;
+        }
+
+        String deserialized = null;
+        try {
+            deserialized = new String(logIdSerialized, Charsets.US_ASCII);
+            return Integer.parseInt(deserialized);
```
eolivelli, line 220: Good

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
```diff
@@ -187,4 +198,40 @@ static void delete(long ledgerId, BookKeeper bookKeeper) throws DurableDataLogEx
...
+    }
+
+    /**
+     * Creates a new {@link Map} that can be set as a Ledger's Custom Metadata.
+     *
+     * @param logId The {@link BookKeeperLog} Id to encode.
+     * @return An immutable {@link Map} containing the encoded Ledger's Custom Metadata.
+     */
+    private static Map<String, byte[]> createLedgerCustomMetadata(int logId) {
+        return Collections.singletonMap(PROPERTY_LOG_ID, Integer.toString(logId).getBytes(Charsets.US_ASCII));
```
eolivelli, line 235: Can we also add  the stream name and a constant 'application' = 'pravega' ?

This will ease understanding the meaning of the ledger.
Especially while using a shared BK cluster
or with tools like
https://github.com/diennea/bookkeeper-visual-manager that allow to inspect a BK cluster

andreipaduroiu: A Stream is not necessarily tied to a single BK Ledger/BKLog. It is made up of many segments, and those segments are created, sealed, deleted, etc. throughout a Stream's lifecycle. We only keep BK ledgers for a short amount of time, so it doesn't make much sense from our POV to tag Stream names.

We can add Pravega as a tag too. That's not a bad idea.

PR #4437, opened 2019-12-11
Issue 4426: Transaction performance metrics

### LINKED ISSUE #4426: Better understanding of Transaction performance
**Problem description**
Currently, we have several metrics that indicate the latency of Transaction operations (i.e., `controller.stream.created_latency_ms`, `controller.stream.committed_latency_ms`, `controller.stream.aborted_latency_ms`). However, a Transaction operation is a complex workflow that is composed of both metadata and segment operations. To better understand the performance of transactions, we would need to have metrics of the specific parts of a transactions to help us understand what could be the bottleneck.

**Problem location**
Transaction metrics. 

Due to the complexity of Transaction operations, we could just measure the time it takes to manage the Segments against the Segment Store. If we have this metric, we can deduct it from the total Transaction time (already there) and infer the time devoted to metadata operations.

RaulGracia:
**Change log description**  
Added and refactored transaction-related metrics to understand better transaction performance.

**Purpose of the change**  
Fixes #4426.

**What the code does**  
Currently, the performance of Transactions is measured only at the `ControllerService` level. This is an insufficient approach to measure the actual performance of Transactions. This PR is aimed at improving the current metrics we export of Transactions in the following ways:
- Distinguish between `COMMITTING` and `COMMITTED` Transactions: Currently, the only notion of latency we have of a Transaction is the time it takes to get from `OPEN` to `COMMITTING` or `ABORTING` (`CREATED` is fine as it is as sync operation). These metrics can be easily understood as the time it takes for a Transaction to get committed/aborted, but this is not true. For this reason, this PR renames the metrics reported at the `ControllerService` level as `[COMMITTING|ABORTING]_TRANSACTION_LATENCY`. Now, the `ControllerService` reports the latency it takes for switching Transactions to committing/aborting states.

- Measure the actual time to commit/abort Transactions in their respective handlers: The actual processes that perform the background job of committing and aborting transactions are their respective request handlers. For this reason, the `COMMIT` and `ABORT` Transaction latency now are measured at this level.

- Separate metric for Segment management: Moreover, we have added a special kind of metrics related to the time it takes to the Controller to manage the segments against the Segment Store for a particular transaction (merge, delete, create,...). In case of low Transaction performance, this metric can tell us whether the problem is with the Segment manegament or with the metadata management of Transactions.

**How to verify it**  
We have added an integration test to check the publishing of Transactions metrics. All other tests should be passing as before. Also, we have deployed a Pravega cluster with these changes and the metrics are being reported as expected:

new-transaction-dashboard


controller/src/main/java/io/pravega/controller/metrics/TransactionMetrics.java
```diff
@@ -10,19 +10,10 @@
...
-import static io.pravega.shared.MetricsNames.ABORT_TRANSACTION_LATENCY;
-import static io.pravega.shared.MetricsNames.COMMIT_TRANSACTION;
-import static io.pravega.shared.MetricsNames.COMMIT_TRANSACTION_FAILED;
-import static io.pravega.shared.MetricsNames.COMMIT_TRANSACTION_LATENCY;
-import static io.pravega.shared.MetricsNames.CREATE_TRANSACTION;
-import static io.pravega.shared.MetricsNames.CREATE_TRANSACTION_FAILED;
-import static io.pravega.shared.MetricsNames.CREATE_TRANSACTION_LATENCY;
-import static io.pravega.shared.MetricsNames.OPEN_TRANSACTIONS;
-import static io.pravega.shared.MetricsNames.globalMetricName;
+import static io.pravega.shared.MetricsNames.*;
```
shiveshr: nit: please revert it

controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java
```diff
@@ -100,23 +101,25 @@ public ControllerEventProcessors(final String host,
                                      final ConnectionFactory connectionFactory,
                                      final StreamMetadataTasks streamMetadataTasks,
                                      final StreamTransactionMetadataTasks streamTransactionMetadataTasks,
-                                     final ScheduledExecutorService executor) {
+                                     final ScheduledExecutorService executor,
+                                     final TransactionMetrics transactionMetrics) {
```
shiveshr: Should we pass new transactionMetrics object here? Can we not do it like other metrics, where we use static provider and get specific transaction metrics?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -106,12 +113,13 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
...
                             log.error("Exception while attempting to commit transaction on stream {}/{}", scope, stream, e);
                         }
+                        transactionMetrics.commitTransactionFailed(scope, stream);
                         future.completeExceptionally(cause);
                     } else {
                         if (r >= 0) {
                             log.info("Successfully committed transactions on epoch {} on stream {}/{}", r, scope, stream);
                         } else {
-                            log.info("No transactions found in committing state on stream {}/{}", r, scope, stream);
+                            log.info("No transactions found in committing state {} on stream {}/{}", r, scope, stream);
```
shiveshr: please replace this with
`                             log.info("No transactions found in committing state on stream {}/{}", scope, stream); `

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -180,10 +189,10 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
                                                         activeEpochRecord.getReferenceEpoch() == txnEpochRecord.getReferenceEpoch()) {
                                                     // If active epoch's reference is same as transaction epoch,
                                                     // we can commit transactions immediately
-                                                    return commitTransactions(scope, stream, new ArrayList<>(activeEpochRecord.getSegmentIds()), txnList, context)
+                                                    return commitTransactions(scope, stream, new ArrayList<>(activeEpochRecord.getSegmentIds()), txnList, context, timer)
```
shiveshr: we dont need to pass timer to all methods, esp when its only exercised at one place after the last step.
 we should instead timer.getElapsed at the end of this method itself.

RaulGracia: I see two problem with that approach:
1) Granularity of transaction latency report: If we just get the time elapsed after this method, we are not measuring the time to commit individual transactions, but the whole batch. In the PR, we execute a `timer.getElapsed()` for each transaction, which looks more precise to me.
2) Completeness: I realized that the `rollTransactions()` method also invokes internally `commitTransactions()` (within `copyTxnEpochSegmentsAndCommitTxns()`). This means that we also need to cover this path to report all the transaction commit latency, otherwise we would be missing some transactions.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
RaulGracia: Fixed.

controller/src/main/java/io/pravega/controller/metrics/TransactionMetrics.java
RaulGracia: Done.

controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java
RaulGracia: I see two options:
1) We pass the metrics object as currently done, so we need to modify the constructor of all classes using it.
2) We make `TransactionMetrics` (and maybe `StreamMetrics` too) singleton objects, so we only need to invoke something like `TransactionMetrics.getInstance().xxxx()` from the places where we want to report metrics. This would not require to change the constructors for that class.

Please, let me know your opinion on that.

shiveshr: My preference would be to use state/singleton objects. So let us focus on `2` and see if it is a viable solution. 
Having static/singleton objects exposed via static getters, they can be accessed from anywhere without having to pass these objects in constructors. 
However, do remember that controller modules can be reinitialized (in case of zk session expiration, all modules restart internally).
We need to assess if singleton metrics objects could have any detrimental effect in any way. Let me elaborate - if there was any state (any metrics) in the previous iteration of metrics object, do we need to reset such states during restarts of other modules that use this metrics object?

If not, we can use singleton objects without any worry.  
However, if we need to instantiate new metrics object each time we restart, then we could have these metrics initialized in ControllerServiceStarter and reinitialized upon each restart which refreshes their state or creates a new object altogether perhaps.

controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
```diff
@@ -222,5 +239,6 @@ public void close() {
         sealStreamLatency.close();
         updateStreamLatency.close();
         truncateStreamLatency.close();
+        INSTANCE.set(null);
```
shiveshr: you are modifying a static variable in the close of an object reference. This may not be a good idea generally speaking. You will end up setting this static variable across close from multiple objects.

controller/src/main/java/io/pravega/controller/metrics/TransactionMetrics.java
```diff
@@ -31,9 +38,34 @@
...
+        commitTransactionLatency = STATS_LOGGER.createStats(COMMIT_TRANSACTION_LATENCY);
+        commitTransactionSegmentsLatency = STATS_LOGGER.createStats(COMMIT_TRANSACTION_SEGMENTS_LATENCY);
+        committingTransactionLatency = STATS_LOGGER.createStats(COMMITTING_TRANSACTION_LATENCY);
+        abortTransactionLatency = STATS_LOGGER.createStats(ABORT_TRANSACTION_LATENCY);
+        abortTransactionSegmentsLatency = STATS_LOGGER.createStats(ABORT_TRANSACTION_SEGMENTS_LATENCY);
+        abortingTransactionLatency = STATS_LOGGER.createStats(ABORTING_TRANSACTION_LATENCY);
+    }
+
+    public static synchronized TransactionMetrics getInstance() {
+        if (INSTANCE.get() == null) {
```
shiveshr: same as above

controller/src/main/java/io/pravega/controller/metrics/TransactionMetrics.java
```diff
@@ -128,7 +215,13 @@ public static void reportOpenTransactions(String scope, String streamName, int o
...
     public void close() {
         createTransactionLatency.close();
+        createTransactionSegmentsLatency.close();
         commitTransactionLatency.close();
+        commitTransactionSegmentsLatency.close();
+        committingTransactionLatency.close();
         abortTransactionLatency.close();
+        abortTransactionSegmentsLatency.close();
+        abortingTransactionLatency.close();
+        INSTANCE.set(null);
```
shiveshr: same as above

controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
shiveshr: Basically dont make this class Closable/AutoClosable. 
Dont let its users call close on it. 
They should call a new `static synchronized reset` method explicitly which takes the existing object and calls a private close on it. 
This will ensure there exists at maximum a single open object and there is never a closed object that getInstance method would return.

controller/src/main/java/io/pravega/controller/metrics/TransactionMetrics.java
```diff
@@ -29,11 +36,36 @@
...
+        commitTransactionLatency = STATS_LOGGER.createStats(COMMIT_TRANSACTION_LATENCY);
+        commitTransactionSegmentsLatency = STATS_LOGGER.createStats(COMMIT_TRANSACTION_SEGMENTS_LATENCY);
+        committingTransactionLatency = STATS_LOGGER.createStats(COMMITTING_TRANSACTION_LATENCY);
+        abortTransactionLatency = STATS_LOGGER.createStats(ABORT_TRANSACTION_LATENCY);
+        abortTransactionSegmentsLatency = STATS_LOGGER.createStats(ABORT_TRANSACTION_SEGMENTS_LATENCY);
+        abortingTransactionLatency = STATS_LOGGER.createStats(ABORTING_TRANSACTION_LATENCY);
+    }
+
+    public static synchronized TransactionMetrics getInstance() {
+        if (INSTANCE.get() == null) {
```
shrids: A `getInstance()` call post `reset()` will create a new instance of TransactionMetrics. This can be avoided.
(Using AutoCloseable interface is also desirable).

controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
```diff
@@ -38,13 +39,30 @@
 /**
  * Class to encapsulate the logic to report Controller service metrics for Streams.
  */
-public final class StreamMetrics extends AbstractControllerMetrics implements AutoCloseable {
+public final class StreamMetrics extends AbstractControllerMetrics {
+
+    private static final AtomicReference<StreamMetrics> INSTANCE = new AtomicReference<>();
```
shiveshr: minor nit: since instance is protected by lock (synchronized methods), we can make it non final and remove atomic reference which is basically double locking.

controller/src/main/java/io/pravega/controller/metrics/TransactionMetrics.java
shiveshr: the intent of reset is to destroy the previous instantiation. 

Also, auto closable will be on the object. Which would mean the object could be closed and getInstance could return a closed object. Which should not be allowed. 
The close method on the object should be effectively private (logically it is similar to what @RaulGracia has done now).

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -328,10 +312,10 @@ private SegmentRange convert(final String scope,
                     return new ImmutablePair<>(data.getId(), getSegmentRanges(segments, scope, stream));
                 }).handle((result, ex) -> {
                     if (ex != null) {
-                        transactionMetrics.createTransactionFailed(scope, stream);
+                        TransactionMetrics.getInstance().createTransactionFailed(scope, stream);
```
shiveshr: do we need to call getInstance each time? Can we call it once and store it in a field in this object and use it for the lifecycle of this object?

shiveshr: Calling getInstance which is synchronized from so many place, and from potentially multiple threads for different workflows, will result in lot of lock contention and waits which could be detrimental for performance. 
In the constructor we can call it once and set it in a local field and use it throughout this class.

controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
```diff
@@ -186,6 +184,10 @@ protected void startUp() {
             log.info("Creating the checkpoint store");
             checkpointStore = CheckpointStoreFactory.create(storeClient);
 
+            // Initialize Stream and Transaction metrics.
```
shiveshr: why do we need to do this initialization? 
Upon its first usage, it would get initialized once and reused everywhere.

shiveshr: Although i think its better for the readability of the code.

controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
RaulGracia: I observed that our code checks specifically forbid to create non-final static variables for safety. That's the reason to create an `AtomicReference` there.

controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
RaulGracia: Yes, to make the initialization explicit I invoked it there.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
RaulGracia: As far as I know, this is how the traditional singleton pattern defines to get the instance. However, if you want, I could "embed" the `getInstance()` call on each method of `TransactionMetrics` and `StreamMetrics`, so from the usage perspective, it would be virtually like calling a static method (i.e., no need to explicitly call `getInstance()`). Is that better? I think that what we want to avoid is to add extra fields to classes just to report these metrics, so created a reference field on each class would be kind of against the change, right?

controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
RaulGracia: Done.

controller/src/main/java/io/pravega/controller/metrics/TransactionMetrics.java
shrids: ok.
I assumed `reset()` is invoked when we are planning to close the instance of TransactionMetrics i.e. after reset() is invoked we would not want to invoke `getInstance()`. I was thinking more on the lines where calling `getInstance()` post the close would throw an `ObjectClosedException`.

kevinhan88:
First of all, this PR is doing two separate things: 
1. add more metrics; 
2. refactor StreamMetrics and TransactionMetrics using Singleton pattern.
I'd suggest next time we address separate concern with separate tickets, in particular the Singleton refactoring is a significant one.

Secondly, I have a few concerns regarding the singleton refactoring.
1. Don't see the need. Even in the worst case there are multiple StreamMetrics or TransactionMetrics objects got created (from existing code I didn't see the case though), it has no harm at all - because the metrics are "step" based, having more objects is like having more steps. It doesn't change functionality.
2. Then what we gain with the overhead of the singleton? Note for every metrics call, it must go through the chain of "getInstance()..."
3. Even more importantly, AutoCloseable is guaranteed by language, while the newly introduced "initialize"/"reset" mechanism is not. There is a risk of memory leaking if the method is not called.

RaulGracia:
@kevinhan88 let me address your comments:
1) The fact of also refactoring `StreamMetrics` in this PR is because is a class doing exactly the same and in the same way as `TransactionsMetrics`, so we considered fine to refactor it as well.
2) The Singleton-like refactor of these clases is not because of generating more or less object, is because i) we need a single instance of `StreamMetrics` and `TransactionsMetrics` to prevent opening multiple `OpStatsLogger` with the same name, which are conflicting among them, ii) these clases are used from multiple classes. The way we were doing this was passing an object to each class using `StreamMetrics` and/or `TransactionsMetrics` via a constructor, but this ends up impacting the signature of many constructors, which in turn also impacts testing code and is annoying. 

The approach taken here achieves addresses these points: makes `StreamMetrics` and `TransactionsMetrics` easier to use without impacting the classes where they are used. Also, the lifecycle of the single instances of these clases are managed in a single point (`ControllerServiceStarter`).

controller/src/main/java/io/pravega/controller/server/ControllerService.java
shiveshr: The concern is everytime you call getInstance you contest for lock. 

Instead, you can call getInstance once when an object of this class was created  -- e.g. ControllerServiceStarter's constructor and keep that reference in a field in ControllerServiceStarter.

So any subsequent call in ControllerServiceStarter for metrics will use the referenced object instead of calling getInstance again.

RaulGracia:
@shiveshr verified that current code works in a cluster and metrics are correctly reported even after controller restarts.

PR #4439, opened 2019-12-11
Issue 4204: (SegmentStore) Fixed excessive cache eviction bug

### LINKED ISSUE #4204: Fix excessive cache entry eviction bug
**Problem description**
The CacheManager updates its Oldest/Newest Generations based on input it receives from its registered cache clients. However it has no way of telling if these clients report correct or incorrect data so it is important that all registered clients provide accurate information to the Cache Manager so that it may make the right decision.

It turns out that while `StreamSegmentReadIndex` works correctly, `SegmentAttributeBTreeIndex` and `ContainerKeyCache`/`SegmentKeyCache` report incorrect Old/New Gen information if their respective caches are empty. Instead of reporting `Integer.MAX_VALUE` in that case, they report `0`, which is a valid value. This causes the `CacheManager` to become paranoid about evicting items, essentially cornering it into a situation where it evicts data as soon as that data becomes eligible for eviction, thus rendering the cache mostly useless.

**Problem location**
CacheManager, ContainerKeyCache/SegmentKeyCache, SegmentAttributeBTreeIndex.

Centralize computation of these values (they are all copy-pasted, which is why this bug propagated so easily) and fix it so that it reports values that make sense to the Cache Manager.

andreipaduroiu:
**Change log description**  
- Fixed a bug in `CacheManager` where it could excessively evict items from the cache as soon as they become eligible even if there was no cache pressure present.

**Purpose of the change**  
Fixes #4204.

**What the code does**  
- Unified the reporting logic of all Cache Clients (`StreamSegmentReadIndex`, `SegmentAttributeBTreeIndex` and `SegmentKeyCache`) into a common method that is now subject to unit tests.
- Updated the `Cache Manager` to
    - Better handle inconsistent input from its clients
        - This is slightly redundant with the unification logic above, but it provides a failsafe in case we add new types of clients in the future which may exhibit this bug.
    - Be more active in case the cache utilization exceeds configured thresholds.
        - Previously, the cache would not necessarily increment the `Current Generation` if there was no activity in the cache. Doing so when there is pressure creates more granularity in the cache generations, making eviction easier (evicting fewer items more times instead of a lot of items fewer times).
    - Better report empty caches in the logs

**How to verify it**  
All tests must pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import javax.annotation.concurrent.GuardedBy;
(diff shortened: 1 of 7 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
```diff
@@ -471,7 +463,8 @@ synchronized boolean evict() {
             return false;
         }
 
-        private synchronized byte[] getFromCache() {
+        @GuardedBy("this")
+        private byte[] getFromCache() {
```
tkaitchuck: This is invoked here: https://github.com/pravega/pravega/blob/344180501362fa0f12c1bd98f6e35dd583f26172/segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java#L391
which is not holding the `this` lock. (If that method is supposed to be guarded by `this` (I can't tell if it is supposed to be) it should be private, annotated, and the other use of the lock inside the method should be removed)

andreipaduroiu: This was a merge mistake on my end. Fixed now.

PR #4447, opened 2019-12-14
Issue 4444: (SegmentStore) Fine-tuning Cache Throttler

### LINKED ISSUE #4444: Make OperationProcessor Throttler more fine grained
**Problem description**
OperationProcesosr may take in too many items at once, which can have the following adverse side effects:
1. Fill up the cache despite the Throttler trying to slow things down.
2. Cause an unstable throughput by accepting a large amount of data at once, then slowing down due to throttling, then again.

**Problem location**
OperationProcessor.

Consider bounding the operations that will be processed by their (cache) size, not by their number.

andreipaduroiu:
**Change log description**  
1.  The `CacheThrottler` now also accounts for in-flight appends when calculating Cache Utilization.
2. Changed the `OperationProcessor` to fetch fewer items from its operation queue if the cache nears its utilization targets.

**Purpose of the change**  
Fixes #4444.

**What the code does**  
Change 1:
- In-flight appends are those appends that have already been sent to Tier1 and will eventually make their way into the cache (there is no way to stop them, short of crashing Tier1). As such, they should also be considered when calculating the Cache Utilization. Even though they are not yet part of the Cache, they will be very shortly.
    - If we don't do this, the Cache Throttler will be basing its decisions on information that will soon be outdated and it risks accepting more appends than it should be, which can easily lead to `CacheFullException`s being thrown afterwards.
- Changed the `CacheUtilizationProvider` to be a class. 
    - There is a single instance of it in the Segment Store and it is attached to the `CacheManager` (on one end) and registers listeners (on the other). There is little new functionality here - most of this has been moved out of the `CacheManager` into it, which also enabled new unit tests to be written.
    - This class also collects the number of outstanding bytes throughout the segment store (from all its containers) and includes that in its calculations.
    - The `OperationProcessor` has been modified to report outstanding bytes as soon as an operation is guaranteed to be processed and it will unreport those bytes when the operation has been inserted into the cache (if applicable) or cancelled (due to some other error). The `CacheManager` both logs an error and throws an assertion (for unit tests) if it detects any leaks.

Change 2:
- Changed the `OperationProcessor` to fetch a variable number of operations from its queue. 
    - Previously it would fetch a max of 1000 at a time, which could significantly increase the chances of `CacheFullException`s under the following conditions:
        - Relatively small cache size (i.e., 3GB) but at a current high utilization (close or above the target utilization (i.e., 75%)).
        - Moderate to high container count per Segment Store pod (8+)
        - Moderate to high load per each container (40-50MB+).
    - For example, 8 containers fetching 1000 items of 50KB each means 400MB will be ingested at once, which can seriously destabilize the cache.
- The new max fetch limit is calculated as follows:
    - We define `TargetUtilization` as Target Cache Utilization, `MaxCacheUtilization` as Max Allowed Cache Utilization, `Spread = MaxCacheUtilization - TargetUtilization`
        - Ex: TargetUtilization=75%, MaxCacheUtilization=90%, -> Spread=15%.
    - If cache utilization is below TargetUtilization-Spread: 1000 (unchanged)
    - If cache utilization is between [TargetUtilization - Spread, MaxCacheUtilization]: gradually decrease (from 1000 to 1)
    - If cache utilization is above MaxCacheUtilization: 1
- This helps better control the amount of data ingested when the cache utilization approaches critical levels while not affecting performance in any other cases.

**How to verify it**  
Tests must pass.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/OperationProcessor.java
```diff
@@ -230,6 +236,17 @@ private boolean isShutdownException(Throwable ex) {
...
 
+    /**
+     * Gets the maximum number of Operations to fetch from the operation queue. This is calculated based on the estimated
+     * cache insertion capacity and its goal is to reduce the number of operations we have in flight as we near the
+     * maximum configured cache capacity.
+     *
+     * @return The maximum number of Operations to fetch from the operation queue.
+     */
+    private int getFetchCount() {
+        return Math.max(1, (int) (this.cacheUtilizationProvider.getCacheInsertionCapacity() * MAX_READ_AT_ONCE));
```
RaulGracia: While I understand that the cache throttler needs to take into account in-flight data, I have a couple of questions about dynamically changing the size of the set of elements picked for processing:
1) I may be probably wrong, but my understanding is that picking elements from this in-memory queue is so fast, that it may not be so different in terms of processing time doing 10 iterations of 100 elements than 100 iterations of 10 elements... However, I see that it is useful to control the amount of in-flight data during saturation periods.
2) The items picked from that queue are aggregated anyway in a DataFrame to be written to Tier 1, right? In this sense, is this indirectly doing some kind of throttling on Tier1 as well?

andreipaduroiu: In general, picking items from the queue is very fast. However, if we start to throttle, we won't be picking items from that queue that frequently and it will start to grow. In addition, once we pick items from the queue we cannot put them back. We either have to process the operation (validate it, write it to BK, put it in cache, etc), or cancel it. We have no intention of cancelling it at this point since that will mess things up for the upstream code. In addition, once we pick them up, these operations will spend a significant amount of time waiting to be acked from BK (millis or even hundreds of millis for busy systems) - there is nothing we can do to abort these operations at this point.

The cache-based throttler used to make its decision based on the data that already existed in the cache, however it did not account for data that we have already committed to processing (in-flight) which will make it to the cache at any time. It is for this reason that I added this extra piece of information. Once the cache nears capacity, picking a large number of operations to commit (at once) risks pushing us beyond that capacity and thus into "Cache Full" territory. However, if we reduce the amount of data that we pick at once we can fine-tune this and thus make sure we won't run into that situation. 

As an analogy, think how you would fill a glass of water without spilling over and without waiting an eternity - while it's still mosty empty, you fill it using a maximum flow, but as it gets full, you reduce the flow so you can properly stop when it's full without spilling water over. In this example, the glass of water is the cache and the faucet is the Operation Processor and the flow of water is managed using throttling and picking the number of items to process at once.

> is this indirectly doing some kind of throttling on Tier1 as well

Yes. All throttling is done before Tier 1.

RaulGracia:
Looks good and we have validated this PR on IO intensive workloads for several hours. @tkaitchuck do you want to have a look?

Issue #4452, opened 2019-12-14
Unable to retrieve cache contents while sync-ing to Tier2


andreipaduroiu:
**Problem description**
This occasionally happens, which causes the container to restart. No data loss.

```
io.pravega.segmentstore.server.DataCorruptionException: Unable to retrieve CacheContents for 'AggregatedAppend: SegmentId = 1, Offsets = [249399-249626), SeqNo = 17626147'.
	at io.pravega.segmentstore.server.writer.SegmentAggregator.getFlushArgs(SegmentAggregator.java:826)
	at io.pravega.segmentstore.server.writer.SegmentAggregator.flushPendingAppends(SegmentAggregator.java:754)
	at io.pravega.segmentstore.server.writer.SegmentAggregator.lambda$flushExcess$14(SegmentAggregator.java:696)
	at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:729)
	at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:698)
	at io.pravega.common.concurrent.Futures.runOrFail(Futures.java:589)
	at io.pravega.common.concurrent.Futures$Loop.run(Futures.java:742)
```

**Problem location**
Segment Store.

**Suggestions for an improvement**
TBD

andreipaduroiu:
This has been root-caused to a concurrency between a Read Index Append and Read Index eviction.

Summary:
1. We have a Cache Entry E1 with Generation G1, and its entire contents is in Storage.
    - E1 maps to the end of the Segment.
2. We initiate an append A1, which will update the contents of E1.
    - E1 has been chosen and A1 has been written to the CacheStorage (but it's not done yet)
3. The Cache Manager executes.
    - E1 is identified for eviction and is evicted.
4. Append A1 resumes and updates E1 with its new length.
    - However E1 is no longer part of the Index, and A1 is "leaked" into the Cache Storage and impossible to retrieve without a container restart.

PR #4455, opened 2019-12-15
Issue 4454: Reading BK Ledger Entries in batches

### LINKED ISSUE #4454: BookKeeperLog.LogReader should read in batches.
**Problem description**
LogReader loads the entire ledger contents (!!!) in memory before serving it to the upstream code. This can easily lead to Heap OOM.

**Problem location**
LogReader.

Read ledger in batches.

andreipaduroiu:
**Change log description**  
- When performing a Ledger Read (Segment Container Recovery) the Ledger is now read in batches, as opposed from all at once.

**Purpose of the change**  
Fixes #4454.

**What the code does**  
- `LedgerHandle.readEntries` loads all the entries in the Java Heap before returning them as an `Enumeration`. This can cause Java Heap OOM if there is significant memory pressure and we need to load a full ledger (1GB).
- `LogReader` has been changed to read the ledger in batches (configurable via setting `bookkeeper.readBatchSize`, defaulted to 64 (entries at once)) instead of the whole thing at once. Nothing is changed from the POV of the upstream code.
- Introduced a `BufferedIterator` in `common` that enables this behavior (for purposes of reuse and proper testing).

**How to verify it**  
All tests must pass.

### DIFF
--- a/common/src/main/java/io/pravega/common/util/BufferedIterator.java
+++ b/common/src/main/java/io/pravega/common/util/BufferedIterator.java
@@ -1,0 +1,98 @@
+/**
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.common.util;
+
+import com.google.common.base.Preconditions;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.function.BiFunction;
+import javax.annotation.concurrent.NotThreadSafe;
+import lombok.NonNull;
+
+/**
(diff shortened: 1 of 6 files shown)


andreipaduroiu:
@eolivelli We identified this while running heavy load tests on Pravega 0.7, after the recent perf improvment PRs have been merged. Our main goal now is to stabilize the 0.7 branch and then we'll take on more PRs.

I understand that you want #4394 in (and so do we - believe me!), but we don't want to add more uncertainty at this point. We have a number of PRs that will improve the situation pretty quickly and have already root-caused some other ones that are bugging us (i.e., #4452 ). However with winter holidays coming up, progress will be a bit slow for a couple of weeks.

Regardless though, I tried to only change 1-2 lines in the `LogReader` class especially because I knew of your PR. Hopefully it won't be too complicated to resolve any conflicts, whichever of these PRs ends up first.

eolivelli:
No problem.
I will rebase that patch  when the time to commit it will come.
No hurry.
I am glad that soon Pravega will be using latest BK APIs

This will help a lot the deprecation (for removal) of the old APIs

PR #4461, opened 2019-12-16
Issue 4452: (SegmentStore) Fixed Read Index Append Concurrency bug

### LINKED ISSUE #4452: Unable to retrieve cache contents while sync-ing to Tier2

andreipaduroiu:
**Change log description**  
- Fixed a bug in `StreamSegmentReadIndex` where it would have been possible for an append data to get lost if the `append` method was invoked concurrently with a cache eviction (`updateGenerations`).
 
**Purpose of the change**  
Fixes #4452.

**What the code does**  
- See parent issue comment for reproduction scenario (https://github.com/pravega/pravega/issues/4452#issuecomment-566168592)
- If a cache eviction occurs during a call to `append`, it would have been possible to append to a `ReadIndexEntry` that was just evicted or about to be. This would have led to temporary data unavailability and an eventual segment container restart (which would have addressed the issue).
- The bug was caused because even though we updated the Read Index Structure under its lock, we were modifying the last `ReadIndexEntry` to have an additional length. This should have been done under the lock.
    - It would be preferable to not invoke `CacheStorage.append` under this lock, however if we update the `ReadIndexEntry`'s length prior to actually writing the data, we risk serving garbage data to a concurrent reader (which would have been worse than the current situation). As such, we have no choice but to do the whole append under this lock. This is not too bad, given that the append itself is doing a memory copy of up to 4KB.
- This was introduced with #4074.

**How to verify it**  
All tests must pass. New unit test was added to verify that appending and cache eviction cannot run concurrently.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
@@ -357,53 +357,31 @@ class StreamSegmentReadIndex implements CacheManager.Client, AutoCloseable {
         long endOffset = offset + data.getLength();
         Exceptions.checkArgument(endOffset <= length, "offset", "The given range of bytes (%d-%d) is beyond the StreamSegment Length (%d).", offset, endOffset, length);
 
-        // Then append an entry for it in the ReadIndex. It's ok to insert into the cache outside of the lock here,
-        // since there is no chance of competing with another write request for the same offset at the same time.
         log.debug("{}: Append (Offset = {}, Length = {}).", this.traceObjectId, offset, data.getLength());
         Preconditions.checkArgument(this.lastAppendedOffset.get() < 0 || offset == this.lastAppendedOffset.get() + 1,
                 "The given range of bytes (Offset=%s) does not start right after the last appended offset (%s).", offset, this.lastAppendedOffset);
-        ReadIndexEntry lastEntry;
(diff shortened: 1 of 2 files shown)


RaulGracia:
@andreipaduroiu we have executed multiple experiments with an image including this change. We need to verify that the very same issue did not reproduce, but the system seems more stable.

andreipaduroiu:
@RaulGracia Agreed. Just a note that this was a concurrency bug, so there's no clear way to reproduce it. We only ran into it on a rather busy system, which had plenty of ongoing appends and Tier 2 writes.

PR #4466, opened 2019-12-20
Issue 4465: Provide api to expose reader <-> segment-count distribution in readergroup

### LINKED ISSUE #4465: Readergroup provides reader <-> segment-count distribution metric
**Problem description**
There is benefit in application being aware of reader to segment-count distribution in a readergroup. 
This is particularly interesting for the application when automatic checkpointing is disabled, whereby automatic rebalancing of segments. 
Making such a signal/api available to applications can help them identify either disproportionately loaded or underloaded/idle readers, whereby enabling them to take application level decisions.
Note: The segment level view will still be opaque to the application and they will not be exposed to segments, merely the segment count. 

Motivation for this comes from #2183. If controller processes are aware of idle readers and a skewed distribution, they could employ a strategy to shutdown higher load readers to facilitate a new distribution among idle readers whereby moving closer to an even load distribution. 

**Problem location**
ReaderGroup

There are two options:
1. new Notification Metric that informs the listeners about reader to segment count periodically. 
2. a direct api in readergroup interface. 

IMO `2` is better than `1` as there is no update linked to a change in state. So observer pattern does not serve a purpose.

shiveshr:

**Change log description**  
Provides new api in ReaderGroup to expose reader <-> segment-count distribution in readergroup.

**Purpose of the change**  
Fixes #4465 

**What the code does**  
Adds a new api in ReaderGroup interface to share readers to segment count distribution. 
This distribution continues to keep the application opaque from segments, while exposing a map of distribution per reader in the readergroup. 
This information will be useful for applications that have disabled automatic checkpointing. It will act as an information point for them to decide if they want to start/stop any readers to trigger any redistribution of segments. There is no assumption about what the redistribution will look like as that depends on readergroups logic and how readers acquire orphaned segments. 

**How to verify it**  
Unit test added

### DIFF
--- a/client/src/main/java/io/pravega/client/stream/ReaderGroup.java
+++ b/client/src/main/java/io/pravega/client/stream/ReaderGroup.java
@@ -187,11 +187,11 @@ public interface ReaderGroup extends ReaderGroupNotificationListener, AutoClosea
 
     /**
      * Returns current distribution of number of segments assigned to each reader in the reader group. 
-     * For unassigned segments, it adds an entry to the map where readerId is empty string.
      *
-     * @return A map of readerId to segment count.   
+     * @return an instance of ReaderSegmentDistribution which describes the distribution of segments to readers 
+     * including unassigned segments.   
      */
-    Map<String, Integer> getReaderSegmentDistribution();
+    ReaderSegmentDistribution getReaderSegmentDistribution();
 
     /**
      * Closes the reader group, freeing any resources associated with it.
--- a/client/src/main/java/io/pravega/client/stream/ReaderGroup.java
+++ b/client/src/main/java/io/pravega/client/stream/ReaderGroup.java
@@ -184,7 +184,15 @@ public interface ReaderGroup extends ReaderGroupNotificationListener, AutoClosea
      */
     @Beta
     CompletableFuture<Map<Stream, StreamCut>> generateStreamCuts(ScheduledExecutorService backgroundExecutor);
-    
+
+    /**
+     * Returns current distribution of number of segments assigned to each reader in the reader group. 
+     * For unassigned segments, it adds an entry to the map where readerId is empty string.
+     *
+     * @return A map of readerId to segment count.   
+     */
+    Map<String, Integer> getReaderSegmentDistribution();
+
     /**
      * Closes the reader group, freeing any resources associated with it.
      */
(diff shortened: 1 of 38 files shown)


client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -206,6 +207,25 @@ public void resetReaderGroup(ReaderGroupConfig config) {
...
+        state.getOnlineReaders().forEach(reader -> {
+            Map<SegmentWithRange, Long> assigned = state.getAssignedSegments(reader);
+            int size = assigned != null ? assigned.size() : 0;
+            mapBuilder.put(reader, size);
+        });
+
+        // add unassigned against empty string
+        int unassigned = state.getNumberOfUnassignedSegments();
+        mapBuilder.put("", unassigned);
+        return mapBuilder.build();
```
shrids: Could we add a log statement which logs the state of the ReaderGroup's state at the point of computation of the segment distribution?

client/src/main/java/io/pravega/client/stream/ReaderGroup.java
```diff
@@ -184,7 +184,15 @@
...
     CompletableFuture<Map<Stream, StreamCut>> generateStreamCuts(ScheduledExecutorService backgroundExecutor);
-    
+
+    /**
+     * Returns current distribution of number of segments assigned to each reader in the reader group. 
+     * For unassigned segments, it adds an entry to the map where readerId is empty string.
+     *
+     * @return A map of readerId to segment count.   
+     */
+    Map<String, Integer> getReaderSegmentDistribution();
```
shrids: We can move this API to `io.pravega.client.stream.ReaderGroupMetrics`. Since this API is similar to the `io.pravega.client.stream.ReaderGroupMetrics#unreadBytes` in the sense that it provides insight into the current metric/state of the ReaderGroup.

shiveshr: I had thought about it. However, i think metrics is a single numerical value. 
I dont know if the distribution really fits into that. Hence i kept it outside of metrics.

shrids:
System test run `job/test-pravega-continuous/2507/`

PR #4472, opened 2019-12-27
Issue 2183: Rebalance readers in event processors when new controllers are added

### LINKED ISSUE #2183: Evaluate options for rebalancing readers in readergroups on internal streams used in controller
**Problem description**
Controller performs concurrent asynchronous processing of requests posted in internal streams.
Here, we apply a trick to linearize all requests for a single "key" using an in-memory queue. The key used is scoped stream name. 
So this ensures that we guarantee in-order processing of requests for a given stream. But at the same time we do not block work for unrelated streams. 

Controller also maintains a reader-position checkpoint that keeps track of highest position object such that all events till this position are processed completely. 
For example, say we have events 1...100.
We have completed processing of 2 till 99. 1 and 100 are still being processed. 
In such case, the checkpoint will be null. The moment 1's processing completes, the checkpoint will move to position object corresponding to 99. 

Also note: all processing is idempotent. 
So this works well if a controller fails and reader's segments are reassigned, the new reader will resume reading on the segments using the checkpointed position object. Which may mean reprocessing some events, but ordering guarantee is maintained. 

However, if new controller instances are added, we cannot automatically rebalance segments to new readers. 
The rebalancing in readergroup, otherwise relies on automatic checkpointing. However, auto-checkpoint is not useful in scenario of asynchronous processing as there is no way we can determine conclusively which events are processed and which are currently being processed when a rebalance happens. 

If we were to rebalance segments if new readers are added, then auto-checkpointing will erroneously assume the position on the segment to be much further than actual because auto-checkpoint takes position corresponding to last processed event for its checkpoint. But as described above, in asynchronous processing, the checkpoint computation is custom to controller's logic. 

This introduces two broad problems:
1. Since processing of event read by origin segment owner `reader` was scheduled asyncrhonously, we can have two different controller instances processing events from the same segment. 
This can break ordering guarantees for proessing events for a key (/scopedStreamName). 

2. New reader starts to process events from a future checkpoint. If original owner controller fails during rebalance, the processing for some events may not have completed while auto-checkpoints checkpoint may point to a future checkpoint and new reader would be processing event from the higher position. 

**Problem location**
Controller event processor + auto-checkpoint + rebalance

We should have a mechanism to explicitly bring reader-offline after completing all ongoing processing before we allow rebalancing of segments owned by the reader. 
Automatic checkpoint does not suffice for this purpose and we should come up with a new scheme that handles such cases.

shiveshr:

**Change log description**  
1. Adds logic in EventProcessorGroup to identify idle and overloaded readers in readergroup and perform.
2. While shutting down an overloaded reader cancel any outstanding request on its event processor cell.
3. In Concurrent event processor, complete the shutdown only after all outstanding requests have been terminated. 

**Purpose of the change**  
Fixes #2183

**What the code does**  
When a readergroup has idle as well as overloaded readers, a rebalance is triggered. 
The EventProcessor group, which owns the readergroup for the eventprocessor, periodically checks if there are idle readers in the readergroup. If there are idle readers and this instance of EventProcessorGroup owns an overloaded reader (reader with more than fair share of segments in the distribution), then it is replaced. Replacement is two step process - first add a new reader, then shutdown the overloaded reader. Addition of new reader does not automatically trigger a rebalance. But when overloaded reader is shutdown, its segments become orphaned and are redistributed among all available readers. This results in redistribution, though not necessarily fair, of segments from overloaded reader to all other readers. 
As eventprocessor, whose reader is removed, is shutdown, it waits until all outstanding requests are either cancelled or completed before completing the shutdown. 

Changes by classes:
- `EventProcessorConfig.java`
Add a new configuration to eventprocessor config for minimum rebalance period. 
- `EventProcessorGroupImpl.java`
Event processorGroup will periodically check if there are idle readers in the reader group and if it owns any overloaded reader. 
If there is an overloaded reader, a replacement reader is added to the group and then eventprocessorcell containing the overloaded reader is then shutdown. This calls the `afterStop` method in the underlying eventprocessor implementation. 
- `ConcurrentEventProcessor.java`
The ConcurrentEventProcessor's `afterStop` sets the stop flag and sends a CANCEL signal to all existing "running jobs" and waits for all jobs to either complete or cancel. 
- `SerializedRequestHandler.java`
Each job, before starting, in SerializedRequestHandler, checks if the job has been canceled or cancels the future if the job has been cancelled. 

**How to verify it**  
Unit tests added to EventProcessorTest (unit and integration) to test rebalance logic. 
Unit test added to ConcurrentEventProcessorTest and SerializedRequestHandlerTest to verify that all

### DIFF
--- a/controller/src/main/java/io/pravega/controller/eventProcessor/impl/EventProcessorGroupImpl.java
+++ b/controller/src/main/java/io/pravega/controller/eventProcessor/impl/EventProcessorGroupImpl.java
@@ -334,13 +334,13 @@ public final class EventProcessorGroupImpl<T extends ControllerEvent> extends Ab
 
         EventProcessorCell<T> cell = eventProcessorMap.get(readerId);
         log.info("Stopping event processor cell: {}", cell);
-        cell.stopAsync();
-        log.info("Awaiting termination of event processor cell: {}", cell);
         try {
+            cell.stopAsync();
+            log.info("Awaiting termination of event processor cell: {}", cell);
             cell.awaitTerminated();
             eventProcessorMap.remove(readerId);
         } catch (Exception e) {
-            log.warn("Failed terminating event processor cell {}.", cell, e);
+            log.error("Failed terminating event processor cell {}.", cell, e);
         }
     }
     
--- a/controller/src/main/java/io/pravega/controller/eventProcessor/impl/EventProcessorGroupImpl.java
+++ b/controller/src/main/java/io/pravega/controller/eventProcessor/impl/EventProcessorGroupImpl.java
@@ -10,7 +10,9 @@
 package io.pravega.controller.eventProcessor.impl;
 
 import io.pravega.client.admin.ReaderGroupManager;
+import io.pravega.client.stream.ReaderSegmentDistribution;
 import io.pravega.client.stream.Stream;
+import io.pravega.common.Exceptions;
 import io.pravega.common.LoggerHelpers;
 import io.pravega.controller.store.checkpoint.CheckpointStore;
 import io.pravega.controller.store.checkpoint.CheckpointStoreException;
@@ -26,6 +28,7 @@ import io.pravega.client.stream.ReaderGroup;
 import io.pravega.client.stream.ReaderGroupConfig;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import com.google.common.util.concurrent.AbstractIdleService;
 import lombok.AccessLevel;
 import lombok.Getter;
@@ -38,6 +41,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 @Slf4j
 public final class EventProcessorGroupImpl<T extends ControllerEvent> extends AbstractIdleService
@@ -59,6 +65,11 @@ public final class EventProcessorGroupImpl<T extends ControllerEvent> extends Ab
 
     private final CheckpointStore checkpointStore;
 
+    private final ScheduledExecutorService executorService;
+    
+    private ScheduledFuture<?> rebalanceFuture;
+    
+    private final long minRebalancePeriodMillis;
     /**
      * We use this lock for mutual exclusion between shutDown and changeEventProcessorCount methods.
      */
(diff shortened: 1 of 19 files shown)


controller/src/main/java/io/pravega/controller/eventProcessor/impl/EventProcessorGroupImpl.java
```diff
@@ -262,7 +284,66 @@ public void changeEventProcessorCount(int count) throws CheckpointStoreException
...
+                });
+            }
+        } catch (Exception e) {
+            Throwable realException = Exceptions.unwrap(e);
+            log.warn("Rebalance failed with exception {} {}", realException.getClass().getSimpleName(), e.getMessage());
+        }
+    }
 
+    private boolean isRebalanceCandidate(int assigned, int readerCount, int segmentCount) {
+        double fair = (double) segmentCount / (double) readerCount;
```
shrids: The segment count also includes `unassigned` segments. These unassigned segments would be eventually distributed across the existing readers. Shouldn't `isRebalanceCandidate()` be based on the unassigned segments so as to prevent unnecessary rebalance operations?

shiveshr: The objective is to rebalance load from any overloaded reader but to not do it too aggressively either. 
So we are rebalancing only if a reader is overloaded AND there are idle readers in the group. 
If such imbalance is present its an opportunity to rebalance. 
Idle readers and unassigned segments is an interesting case where one could argue that idle readers can be picking up unassigned segments and may no longer be idle. But we still have an overloaded reader and we treat it as an opportunity to rebalance the overloaded reader.

controller/src/main/java/io/pravega/controller/eventProcessor/impl/EventProcessorGroupImpl.java
```diff
@@ -262,7 +284,66 @@ public void changeEventProcessorCount(int count) throws CheckpointStoreException
...
+    void rebalance() {
+        try {
+            ReaderSegmentDistribution readerSegmentDistribution = readerGroup.getReaderSegmentDistribution();
+            Map<String, Integer> distribution = readerSegmentDistribution.getReaderSegmentDistribution();
+            int readerCount = distribution.size();
+            int unassigned = readerSegmentDistribution.getUnassignedSegments();
+            int segmentCount = distribution.values().stream().reduce(0, (x, y) -> x + y) + unassigned;
+            
+            // If there are idle readers (no segment assignments, then identify and replace overloaded readers). 
+            boolean idleReaders = distribution.entrySet().stream().anyMatch(x -> !Strings.isNullOrEmpty(x.getKey()) && x.getValue() == 0);
```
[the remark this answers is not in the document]
shiveshr: x is from an entryset in the map, the entry cannot be null.. in fact even the key cannot be null in a map.

shrids:
Triggered a system test run with this changeset ref: `pravega-branch-builder/2016/`.

shrids:
Sstem test run `job/test-pravega-continuous/2522/` was successful.

PR #4477, opened 2020-01-07
Issue 4460: BookieFailover Test sporadic failure

### LINKED ISSUE #4460: Sporadic failure of BookieFailoverTest
**Problem description**
We are observing frequent system test failures related to `BookieFailoverTest`. Doing a quick look at the logs, the problem manifests as follows:
1) The test is able to read/write events up to scaling Bookkeeper from 3 instances to 2:
```
2019-12-14 04:46:13,228 53918 [Time-limited test] INFO  i.p.test.system.BookieFailoverTest - Sleeping for 15 seconds.
...
2019-12-14 04:46:28,236 68926 [Time-limited test] INFO  i.p.test.system.BookieFailoverTest - Write count is 2074 after 15 seconds sleep after bookie failover.
```

2) Then, the test downscales Bookkeeper correctly and Pravega cannot make progress, given that the write quorum to Bookkeeper is not satisfied:
```
2019-12-14 04:46:43,240 83930 [Time-limited test] INFO  i.p.test.system.BookieFailoverTest - Write count is 2074 after 30 seconds sleep after bookie failover.
2019-12-14 04:46:43,240 83930 [Time-limited test] INFO  i.p.test.system.BookieFailoverTest - Writes failed when bookie is scaled down.
```

3) Finally, we scale up the service again and provide some more time to check that the test is able to read/write data against Pravega again:
```
2019-12-14 04:46:43,240 83930 [Time-limited test] INFO  i.p.t.s.f.s.k.BookkeeperK8sService - Scaling Bookkeeper service to 3 instances.
2019-12-14 04:48:02,715 163405 [Time-limited test] INFO  i.p.test.system.BookieFailoverTest - Final write count 2079.
```

However, it looks like the recovery time for the Bookies took too long (we can observe multiple `WireCommands.WrongHost`) and the readers gave up retrying, so the test assertion failed:
```2019-12-14 04:48:07,737 168427 [Time-limited test] INFO  i.p.t.system.AbstractReadWriteTest - Unable to read events:
io.pravega.common.util.RetriesExhaustedException: io.pravega.shared.protocol.netty.ConnectionFailedException: WireCommands.WrongHost(type=WRONG_HOST, requestId=549755813889, segment=testBookieFailoverScope1838653616/_RGtestBookieFailoverReaderGroup1492013574/0.#epoch.0, correctHost=, serverStackTrace=)
```

As the rest of failover tests work consistently, we need to understand why the container recovery is not working fast enough when killing a Bookie instead of a Segment Store, for instance.

**Problem location**
`BookieFailoverTest`.

Fix the issue.

RaulGracia:
**Change log description**  
Increased retry timeout in SegmentMetadataClientImpl to mitigate test failures due to slow Bookie restart and added read event assertions in BookieFailoverTest.

**Purpose of the change**  
Fixes #4460.

**What the code does**  
This PR increases the retry time for `SegmentMetadataClientImpl` to prevent system test failures in the case that the environment is slow on restarting the failed Bookie in `BookieFailoverTest`. This has been proven to remove the sporadic failures of this test in our system test environment. Apart from that, the `BookieFailoverTest` also checks the progress of read events. 

Note that in extremely slow environments, we may still see this test failing, but this is expected given that the test logic interprets any exception in the reader side as a test failure. Therefore, in such slow environments, we may still get `RetriesExhaustedException` on the reader side. But this is fine, given that it will give us a clue that the environment we are using for system tests is not suitable (e.g., lack of resources).

**How to verify it**  
We have executed this test in isolation 8 times in a row without experiencing the original problem `RetriesExhaustedException`. Before, we observed `BookieFailoverTest` failing 30%-50% of the time.

### DIFF
--- a/client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
@@ -49,7 +49,7 @@ import lombok.extern.slf4j.Slf4j;
 @RequiredArgsConstructor
 @Slf4j
 class SegmentMetadataClientImpl implements SegmentMetadataClient {
-    private static final RetryWithBackoff RETRY_SCHEDULE = Retry.withExpBackoff(1, 10, 9, 30000);
+    private static final RetryWithBackoff RETRY_SCHEDULE = Retry.withExpBackoff(1, 10, 10, 30000);
 
     private final Segment segmentId;
     private final Controller controller;
--- a/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/AsyncSegmentInputStreamImpl.java
@@ -167,35 +167,32 @@ class AsyncSegmentInputStreamImpl extends AsyncSegmentInputStream {
     @Override
     public CompletableFuture<SegmentRead> read(long offset, int length) {
         Exceptions.checkNotClosed(closed.get(), this);
-
-        return this.tokenProvider.retrieveToken().thenComposeAsync(token -> {
-            WireCommands.ReadSegment request = new WireCommands.ReadSegment(segmentId.getScopedName(), offset, length,
+        return backoffSchedule.retryWhen(t -> {
+            Throwable ex = Exceptions.unwrap(t);
+            if (closed.get()) {
+                log.debug("Exception while reading from Segment : {}", segmentId, ex);
+            } else {
+                log.warn("Exception while reading from Segment : {}", segmentId, ex);
+            }
+            return ex instanceof Exception && !(ex instanceof ConnectionClosedException) && !(ex instanceof SegmentTruncatedException);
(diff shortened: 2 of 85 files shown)


RaulGracia:
@shiveshr @shrids @eolivelli running system tests on this branch for safety.

RaulGracia:
System tests have passed (build 2034).

client/src/main/java/io/pravega/client/segment/impl/SegmentMetadataClientImpl.java
```diff
@@ -49,7 +49,7 @@
 @RequiredArgsConstructor
 @Slf4j
 class SegmentMetadataClientImpl implements SegmentMetadataClient {
-    private static final RetryWithBackoff RETRY_SCHEDULE = Retry.withExpBackoff(1, 10, 9, 30000);
+    private static final RetryWithBackoff RETRY_SCHEDULE = Retry.withExpBackoff(1, 10, 10, 30000);
```
shiveshr: should we increase this to a larger retry count so that we reduce the chances of encountering the issue even on slower systems?

PR #4483, opened 2020-01-09
Issue 4459: [r0.6] Upgrading bookkeeper version to 4.8.2

### LINKED ISSUE #4459: Pravega Bookkeeper docker image has an old version of Bookkeeper server

SrishT:

**Change log description**  
Upgrade the Bookeeper server from `4.7.3` to `4.8.2`.

**Purpose of the change**  
Resolves #4459 

**What the code does**  
Modifies the Dockerfile and entrypoint to work with Bookeeper `4.8.2`. It does the following:

1. The default 4.8.2 bookkeeper start up scripts uses `ZooKeeperMain` to create/modify nodes in ZK. This throws
```
/opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 3
```
Due to this, we ended up in using `zk-shell` on similar lines as Apache Bookeeper PR https://github.com/apache/bookkeeper/pull/1666.

2. The `4.8.2` script exits if `BK_STREAM_STORAGE_ROOT_PATH` is not set. Also, a zk node in this path is expected to be present if a bookie is up and running. (This is not the case with the default bk 4.8.2 image)

3. `/opt/bookkeeper/bin/bkctl --service-uri` this option used if Bookkeeper image did not work. `java.lang.IllegalArgumentException: No service URI` is observed. It is similar to issues observed in apache/bookkeeper#2112.

4. It has a workaround for creating ephmeral node using `zk-shell`. The exit status returned by the `/opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain` used in the base image is always 0, even in the case the node exists, causing unnecessary bookie restart.

**How to verify it**  
* A successful system test run
* A successful longevity test run
* Manual testing of at least the following scenarios:
  * Upgrade
  * Failure of Bookkeeper instances in a cluster

### DIFF
--- a/docker/bookkeeper/Dockerfile
+++ b/docker/bookkeeper/Dockerfile
@@ -7,31 +7,35 @@
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
-FROM apache/bookkeeper:4.7.3
+FROM apache/bookkeeper:4.8.2
 
-ARG BK_VERSION=4.7.3
+ARG BK_VERSION=4.8.2
 ARG DISTRO_NAME=bookkeeper-all-${BK_VERSION}-bin
-ARG GPG_KEY=FD74402C
 
 RUN set -x \
-    && yum install -y iproute \
+    && yum install -y iproute wget \
     && cd /opt \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.asc" \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.sha512" \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.asc" \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.sha512" \
     && sha512sum -c ${DISTRO_NAME}.tar.gz.sha512 \
-    && gpg --batch --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys "$GPG_KEY" || \
-       gpg --batch --keyserver hkp://ipv4.pool.sks-keyservers.net --recv-keys "$GPG_KEY" || \
-       gpg --batch --keyserver hkp://pgp.mit.edu:80 --recv-keys "$GPG_KEY" \
+    && wget https://dist.apache.org/repos/dist/release/bookkeeper/KEYS \
+    && gpg --import KEYS \
     && gpg --batch --verify "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz" \
     && tar -xzf "$DISTRO_NAME.tar.gz" \
     && cp -r bookkeeper-all-${BK_VERSION}/* /opt/bookkeeper/ \
     && rm -rf "bookkeeper-all-${BK_VERSION}" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz.sha512" \
+    # install zookeeper shell
+    && wget -q https://bootstrap.pypa.io/get-pip.py \
+    && python get-pip.py \
+    && pip install zk-shell \
+    && rm -rf get-pip.py \
     && yum clean all
 
 WORKDIR /opt/bookkeeper
 
 COPY entrypoint.sh /opt/bookkeeper/scripts/pravega_entrypoint.sh
+
 # For backwards compatibility with older operator versions
 COPY entrypoint.sh /opt/bookkeeper/entrypoint.sh
 
@@ -40,7 +44,4 @@ RUN chmod +x -R /opt/bookkeeper/scripts/
 ENTRYPOINT [ "/bin/bash", "/opt/bookkeeper/scripts/pravega_entrypoint.sh" ]
 CMD ["bookie"]
 
-# BookKeeper healthcheck was broken in 4.7.3 and was not fixed until 4.9.0
-# This overrides the healthcheck to the default BK healthcheck method
-# https://github.com/apache/bookkeeper/issues/1687
-HEALTHCHECK --interval=10s --timeout=60s CMD /opt/bookkeeper/bin/bookkeeper shell bookiesanity
+HEALTHCHECK --interval=10s --timeout=60s CMD /bin/bash /opt/bookkeeper/scripts/healthcheck.sh
--- a/docker/bookkeeper/entrypoint.sh
+++ b/docker/bookkeeper/entrypoint.sh
@@ -10,20 +10,9 @@
 #
 set -e
 
-# To create directories for multiple ledgers and journals if specified
-create_dir() {
-  IFS=',' read -ra directories <<< $1
-  for i in "${directories[@]}"
-  do
-      mkdir -p $i
-      if [ "$(id -u)" = '0' ]; then
-          chown -R "${BK_USER}:${BK_USER}" $i
-      fi
-  done
-}
-
 BOOKIE_PORT=${bookiePort:-${BOOKIE_PORT}}
 BOOKIE_PORT=${BOOKIE_PORT:-3181}
+BOOKIE_HTTP_PORT=${BOOKIE_HTTP_PORT:-9090}
 BK_zkServers=$(echo "${ZK_URL:-127.0.0.1:2181}" | sed -r 's/;/,/g')
 ZK_URL=$(echo "${ZK_URL:-127.0.0.1:2181}" | sed -r 's/,/;/g')
 PRAVEGA_PATH=${PRAVEGA_PATH:-"pravega"}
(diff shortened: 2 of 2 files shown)


eolivelli:
Why are you upgrading to 4.82 and not to 4.9.2?

docker/bookkeeper/Dockerfile
```diff
@@ -7,31 +7,35 @@
...
-ARG GPG_KEY=FD74402C
 
 RUN set -x \
-    && yum install -y iproute \
+    && yum install -y iproute wget \
     && cd /opt \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.asc" \
-    && curl -O "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.sha512" \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
```
fpj: Not a big deal, but it seems to be out of scope to change from `curl` to `wget`.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +47,143 @@ export BK_tlsTrustStoreType=JKS
...
+
+configure_bk() {
+    # We need to update the metadata endpoint and Bookie ID before attempting to delete the cookie
+    sed -i "s|.*metadataServiceUri=.*\$|metadataServiceUri=${BK_metadataServiceUri}|" /opt/bookkeeper/conf/bk_server.conf
+    if [ ! -z "$BK_useHostNameAsBookieID" ]; then
+      sed -i "s|.*useHostNameAsBookieID=.*\$|useHostNameAsBookieID=${BK_useHostNameAsBookieID}|" ${BK_HOME}/conf/bk_server.conf
+    fi
+}
+
+fix_bk_ipv6_check() {
```
fpj: This feels a bit hacky. Could we not exit on error and check the error code instead? If it is 255 with that message, then we ignore rather than exit.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +47,143 @@ export BK_tlsTrustStoreType=JKS
...
+            if [ $? -eq 0 ]; then
+                echo "initnewcluster operation succeeded"
+            else
+                echo "initnewcluster operation failed. Please check the reason."
+                echo "Exit status of initnewcluster"
+                echo $?
+                exit
+            fi
+        else
+            echo "Others may be initializing the cluster at the same time."
```
fpj: What others can be initializing the cluster at the same time?

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +47,143 @@ export BK_tlsTrustStoreType=JKS
...
+                    break
+                else
+                    echo "Waited $tenSeconds * 10 seconds. Continue waiting."
+                    (( tenSeconds++ ))
+                    continue
+                fi
+            done
+
+            if [ ${tenSeconds} -eq 1000 ]; then
+                echo "Waited 1000 seconds for bookkeeper cluster to initialize, but to no . Something is wrong, please check"
```
fpj: This error message is incomplete, please revisit it.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +47,143 @@ export BK_tlsTrustStoreType=JKS
...
+            else
+                echo "initnewcluster operation failed. Please check the reason."
+                echo "Exit status of initnewcluster"
+                echo $?
+                exit
+            fi
+        else
+            echo "Others may be initializing the cluster at the same time."
+            tenSeconds=1
+            while [ ${tenSeconds} -lt 100 ]
```
fpj: How are we choosing the max time to wait? This seems to be waiting for 16-17 minutes max.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +47,143 @@ export BK_tlsTrustStoreType=JKS
...
+
+create_zk_root() {
+    if [ "x${BK_CLUSTER_ROOT_PATH}" != "x" ]; then
+        echo "Creating the zk root dir '${BK_CLUSTER_ROOT_PATH}' at '${BK_zkServers}'"
+        zk-shell --run-once "create ${BK_CLUSTER_ROOT_PATH} '' false false true" ${BK_zkServers}
+    fi
+}
+
+configure_bk() {
+    # We need to update the metadata endpoint and Bookie ID before attempting to delete the cookie
```
fpj: Could you add to the comment to elaborate on what's really happening? Explain what a cookie is and why the metadata service endpoint needs to be updated. Otherwise, it is difficult to understand what it going on.

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +47,143 @@ export BK_tlsTrustStoreType=JKS
...
+   #    `sysctl: cannot stat /proc/sys/net/ipv6/bindv6only: No such file or directory`
+   #
+   # If the `common.sh` file is sourced in a script that sets "exit on error" as true (set -e), the script shall return
+   # an error code `255`. Since, the `bookkeeper/bin/bookeeper` shell script sources the `common.sh` file, and it sets
+   # "exit on error" as true, we encounter this error when it is invoked. To avoid that error, here we modify the
+   # `common.sh` file to execute `/sbin/sysctl -n net.ipv6.bindv6only` only if file `/proc/sys/net/ipv6/bindv6only`
+   # is available.
+
+   # Replace the 22nd line with "if [ -f /sbin/sysctl ] && [ -f /proc/sys/net/ipv6/bindv6only ]; then".
+   sed -i "22s|.*|if [ -f /sbin/sysctl ] \&\& [ -f /proc/sys/net/ipv6/bindv6only ]; then|" ${BK_HOME}/bin/common.sh
```
fpj: it looks like we can propose a change upstream to Apache BookKeeper rather than making patches like this. Are you considering doing it?

docker/bookkeeper/entrypoint.sh
```diff
@@ -48,31 +47,143 @@ export BK_tlsTrustStoreType=JKS
...
+}
+
+initialize_cluster() {
+    set +e
+
+    zk-shell --run-once "ls ${BK_zkLedgersRootPath}/available/readonly" ${BK_zkServers}
+    if [ $? -eq 0 ]; then
```
fpj: The description of the PR says:

```
The exit status returned by the zk-shell is always 0, even in the case the node exists, causing unnecessary bookie restart.)
```

If it is always zero, then I wonder if the behavior of this if-else block is correct.

docker/bookkeeper/Dockerfile
ravisharda: The base Bookkeeper image uses `wget` now, so this was modified too. 

https://github.com/apache/bookkeeper/blob/branch-4.8/docker/Dockerfile

docker/bookkeeper/entrypoint.sh
ravisharda: Yes, it is indeed hacky. The problem is that one of the internal Bookkeeper script invokes the `${SCRIPTS_DIR}/common.sh` script too, so we must modify the `common.sh` file itself. 

Eventually, this change would not be required once the [Bookkeeper PR](https://github.com/apache/bookkeeper/pull/2219) we have raised is merged and we modify our image to leverage the modified image.

docker/bookkeeper/entrypoint.sh
ravisharda: @SrishT, 
It looks like not all of the changes made in the source [PR 4420](https://github.com/pravega/pravega/pull/4420/) (for the upgrade to `4.9.2`)  have made it to this PR. This error message was modified in the source PR: https://github.com/pravega/pravega/blob/5c4c77fb7febc5e548807d2c8750bc87c368d6db/docker/bookkeeper/entrypoint.sh#L149. 

Could you please integrate the latest updates made in the source PR into this PR?

docker/bookkeeper/entrypoint.sh
ravisharda: It is other "instances" of Bookkeeper containers.

docker/bookkeeper/entrypoint.sh
ravisharda: Yes, there is an open PR for `apache/bookkeeper`: https://github.com/apache/bookkeeper/pull/2219.

docker/bookkeeper/entrypoint.sh
ravisharda: Good catch, @fpj ! :) The description was erroneous. I have updated the description in this (as well as the [source PR](https://github.com/pravega/pravega/pull/4420/)).

ravisharda:
PR #4420 now targets BookKeeper 4.8.2 instead of 4.9.2. Accordingly, this PR is now redundant. So, closing this PR.

eolivelli:
@ravisharda why are we going to 4.8.2?

Latest BK release is 4.10, latest stable is 4.9.2.
4.8 branch is almost dead

ravisharda:
> @ravisharda why are we going to 4.8.2?
> 
> Latest BK release is 4.10, latest stable is 4.9.2.
> 4.8 branch is almost dead

@eolivelli upgrading to `4.8.2` is a temporary/tactical thing, and we'll upgrade Bookkeeper to higher version shortly later after we have had the chance to verify that everything works and performs as expected.

eolivelli:
I see.
Thanks

PR #4485, opened 2020-01-11
Issue 4451: (SegmentStore) Enhancing Segment Attribute Tier 2 Sync

### LINKED ISSUE #4451: BTree Data Corruption on NFS mounts
**Problem description**
Found while running load tests:
```
io.pravega.segmentstore.server.DataCorruptionException: BTreeIndex operation failed. Index corrupted.
	at io.pravega.segmentstore.server.attributes.SegmentAttributeBTreeIndex.handleIndexOperationException(SegmentAttributeBTreeIndex.java:647)
...
Caused by: io.pravega.common.util.IllegalDataFormatException: Wrong footer information. RootPage Offset (236958) + Length (974) exceeds Footer Offset (237568).
	at io.pravega.common.util.btree.BTreeIndex.initialize(BTreeIndex.java:215)
	at io.pravega.common.util.btree.BTreeIndex.lambda$null$0(BTreeIndex.java:196)
	... 11 common frames omitted
```

**Problem location**
BTree Index.

TBD

andreipaduroiu:
**Change log description**  
- Storing Segment Attribute BTreeIndex Root Pointer as a Core Attribute on the parent segment, every time the BTreeIndex is updated.
- Split out the Tier 2 Attribute sync logic out of SegmentAggregator into its own class (AttributeAggregator).
- Allowing certain UpdateAttributeOperations to be executed after the Segment is sealed.

**Purpose of the change**  
Fixes #4451. 

**What the code does**  
- Segment Attribute BTreeIndex Root Pointer
    - Changed BTreeIndex to return the offset of the root page with every update. The upstream code (StorageWriter and sub-components) will take this value and store it in a core attribute (`Attributes.ATTRIBUTE_SEGMENT_ROOT_POINTER`).
    - Upon recovery, the BTreeIndex is initialized with both the Attribute Segment Length (as before) and the root pointer value (if available). The BTreeIndex will use the root pointer to initialize and if not available, it uses the length to infer that (for backwards compatibility).
    - This helps with the situations when a BTreeIndex update was only partially written to Tier 2, in which case a subsequent recovery would have been unable to reinitialize the index. By storing the root offset **after** the write is complete we guarantee to always have a working view/version of the BTree to use.
- Allowing certain `UpdateSegmentAttributeOperation`s to execute after the segment is sealed
    - This is necessary because we may seal the segment immediately after updating some attributes on it, so by the time those attributes are sync-ed to Tier 2, it would be impossible to update the Root Pointer in the Segment's Metadata (due to it being sealed), which would cause us to not be able to access the latest updates in that BTreeIndex.
    - Added a flag to this operation to mark it as `Internal`. If it is internal and only updates core attributes, then it is allowed to be executed after a segment is sealed. (Note that Core Attributes have a different lifecycle than extended attributes; they are always pinned to memory and, when the segment is evicted, they are stored in the Segment's Metadata, not in the BTreeIndex).
- Splitting Tier 2 data sync and Attribute Sync
    - SegmentAggregator was getting too big and complex and it was becoming hard to provide specialized handling for Tier 2 Attribute Sync.
    - Created `AttributeAggregator` which aggregates attributes and only persists them to Tier2 when certain flush conditions are met (either time-based or when a minimum number of attributes are accumulated). This should significantly reduce the number of BTreeIndex updates, thus keeping its size small(er) and improving Tier 2 data sync (previously, every Tier 2 data sync would require writing both the segment data and attribute data, which could slow things down).
    - The `AttributeAggregator` asynchronously updates the core attributes (`ATTRIBUTE_SEGMENT_ROOT_POINTER` and `ATTRIBUTE_SEGMENT_PERSIST_SEQ_NO`) to avoid a situation where it would be subject to Tier 1 throttling (which would essentially deadlock the StorageWriter and Tier 1).
    - The StorageWriter now has 3 processors for each Segment: `SegmentAggregator`, `AttributeAggregator` and `WriterTableSegmentProcessor` (the latter only for Table Segments). Each of these is passed every operation in the `DurableLog` and such an operation is acked/truncated out of DurableLog/Tier1 only after all processors have successfully sync-ed it to Tier2. (This functionality was already present and was not added as part of this change).

**How to verify it**  
All existing and new tests must pass. System and longevity tests must pass as well.


common/src/main/java/io/pravega/common/util/btree/BTreeIndex.java
```diff
@@ -790,14 +792,17 @@ private ByteArraySegment generateMinKey() {
         long rootMinOffset = lastPage.getMinOffset();
         assert rootMinOffset >= 0 : "root.MinOffset not set";
         return this.write.apply(pages, oldOffsets, rootMinOffset, timeout)
-                         .thenApply(indexLength -> setState(indexLength, rootOffset, rootLength).length);
+                .thenApply(indexLength -> {
+                    setState(indexLength, rootOffset, rootLength);
```
eolivelli: Will some sort of compare-and-set help here in preventing races?
You know the expected previous **state**

andreipaduroiu: It's not just in memory state that I care about. What I noticed is that, at least for NFS Tier 2 mounts, even if I send a 1MB data block, if we crash while writing it, part of it may have been durably stored onto whatever is behind the NFS mount. While not a problem for regular stream segments (since they're append-only and we can always resume from where we left off), for BTreeIndex it's different. It needs the file (at least the last 12 bytes of it) to be structured, otherwise it won't be able to read the BTree Pages properly. If we wanted to write 10 pages, for example, but crashed after writing 6.5 of them, then the data structure won't be able to recover afterwards.

Going back to your question. I need to persist **somewhere** the pointer to the root page of the BTree index. Previously it was the last 12 bytes of the Attribute File, but since that's not reliable, one of the changes in this PR is to also store it in the main Segment's Core Attributes (these attributes have a different lifecycle so they're not susceptible to this problem). If we crash in the middle of a write (or just after a good write but before persisting the new pointer), we'll always be able to recover from the previous BTree Index pointer, and try again.

Please wait until I finish this PR. I only submitted it so I can get some Travis builds done and get some system tests executed, but I am not ready for review. I'll add the appropriate description and explanations when I'm done, most likely in a few days.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/WriterConfig.java
```diff
@@ -52,6 +53,12 @@
     @Getter
     private final Duration flushThresholdTime;
 
+    /**
+     * The minimum number of attributes that should accumulate before flushing them into the Attribute Index.
```
tkaitchuck: It seems odd to configure a minimum in terms of a count without also providing some sort of bound. Like maybe based on time. Otherwise it could sit unflushed indefinitely.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/AttributeAggregatorTests.java
```diff
@@ -0,0 +1,784 @@
...
+/**
+ * Unit test for the {@link AttributeAggregator} class.
+ */
+public class AttributeAggregatorTests extends ThreadPooledTestSuite {
+    private static final int CONTAINER_ID = 0;
+    private static final long SEGMENT_ID = 123;
+    private static final String SEGMENT_NAME = "Segment";
+    private static final byte[] APPEND_DATA = SEGMENT_NAME.getBytes();
+    private static final UUID CORE_ATTRIBUTE_ID = Attributes.EVENT_COUNT;
+    private static final List<UUID> EXTENDED_ATTRIBUTE_IDS = IntStream.range(0, 20).mapToObj(i -> UUID.randomUUID()).collect(Collectors.toList());
```
tkaitchuck: This should be an immutable list.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentContainerTests.java
```diff
@@ -143,6 +145,11 @@
  * DurableDataLog.
  */
 public class StreamSegmentContainerTests extends ThreadPooledTestSuite {
+    /**
+     * Auto-generated attributes which are not set externally but maintained internally. To ease our testing, we will
+     * exclude these from all our checks.
+     */
+    private static final Collection<UUID> AUTO_ATTRIBUTES = Sets.newHashSet(Attributes.ATTRIBUTE_SEGMENT_ROOT_POINTER, Attributes.ATTRIBUTE_SEGMENT_PERSIST_SEQ_NO);
```
tkaitchuck: This should be an immutable set.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
```diff
@@ -331,26 +322,23 @@ public String toString() {
     @Override
     public void add(SegmentOperation operation) throws DataCorruptionException {
         ensureInitializedAndNotClosed();
+        if (!(operation instanceof StorageOperation)) {
```
tkaitchuck: Between this and the Javadoc at the top of the file it seems like maybe a refactoring is due.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/AttributeAggregator.java
```diff
@@ -0,0 +1,477 @@
...
+        Exceptions.checkNotClosed(isClosed(), this);
+        Preconditions.checkArgument(
+                operation.getStreamSegmentId() == this.metadata.getId(),
+                "Operation '%s' refers to a different Segment than this one (%s).", operation, this.metadata.getId());
+        if (isSegmentDeleted()) {
+            return;
+        }
+
+        boolean processed = false;
+        if (operation instanceof StreamSegmentSealOperation) {
```
tkaitchuck: Would an enum be better here?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/AttributeAggregator.java
```diff
@@ -0,0 +1,477 @@
...
+
+    //endregion
+
+    //region AggregatedAttributes
+
+    /**
+     * Aggregates pending Attribute Updates.
+     */
+    private static class State {
+        private final HashMap<UUID, Long> attributes;
```
tkaitchuck: The other members of this class are atomics and this is a normal hashmap which is accessed by `include` without any synchronization. This looks like a bug.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/WriterConfig.java
andreipaduroiu: There already is a time bound (see line 54 above). It's the same setting used for the regular segment transfer; I decided I didn't want another setting that would have essentially the same value.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/AttributeAggregator.java
andreipaduroiu: Good catch. Fixed.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentContainerTests.java
andreipaduroiu: Fixed.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/AttributeAggregatorTests.java
andreipaduroiu: Fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/AttributeAggregator.java
andreipaduroiu: That means I'd need to maintain an enum for the types of Operations, where each operation would have a getter that returns its enum value. That would resolve the `instanceof` but I would still need to cast them eventually to the correct type, so I am not sure we're going to get much benefit out of it. The downside of that would be the need to ensure that the enum matches the operation type; at least with `instanceof` and cast we don't need to worry about that problem.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
andreipaduroiu: Well this was a refactoring. The `SegmentAggregator` deals exclusively with transferring data within the segment and the `AttributeAggregator` handles attribute transfer.

I decided to keep segment deletion logic in here because we want to remember to delete attributes when we delete the segment, when we merge it, and one more complex situation. Instead of doing it twice, I preferred to keep that part in one place.

I don't understand what is wrong with this line though. The `SegmentAggregator` only handles `StorageOperation`s; all other types of operations are `MetadataOperation`s are not its responsibility. There are other processors for each segment within the StorageWriter, and each deals with its own types of operations. `WriterTableProcessor` only handles append operations, `AttributeAggregator` handles `AttributeUpdaterOperations` (appends & attribute updates) and this handles appends, seals, mergers, deletes, truncates, etc. In other words, anything which applies directly to Tier 2 is a `StorageOperation` and is handled by this class, while others are a bit more specialized and do not touch Tier 2 (directly).

andreipaduroiu: Let's see if I can explain. The `StorageWriter` picks Operations (all of them) from Tier1/DurableLog. The `StorageWriter` also has `WriterProcessors` for each active segment in the container. A `WriterProcessor` has a `SegmentAggregator`, an `AttributeAggregator` (new in this PR) and a `WriterTableProcessor` (for Table Segments).

The `StorageWriter` picks operation from Tier 1 (in order) and sends them to all its `WriterProcessor`s, which in turn sends them to all of the aggregators sub-processors. So from this POV, it is using the visitor pattern (in a way). The operations themselves are immutable, but each Processor/Aggregator knows which operations it needs to process and will just ignore the rest. IMO, this can be done in two ways: the current way, in which all processors see all operations and decide which ones they use, and another in which they register with the StorageWriter to send them only those operations they care about. I just happen to do the former, because even if it accepts an operation of a given type, it may end up ignoring it due to other factors.

tkaitchuck: That sounds reasonable. 
Alternatively: You could also make WriterProcessor the one to send some types to the SegmentAddregator and others to the WriterTableProcessor. (Obviously this makes it more aware of what those classes are doing, so its up to you as to whether that falls too far outside of its current responsibilities.)

andreipaduroiu: Good suggestion. Since this is outside of the scope of this change, I captured this in https://github.com/pravega/pravega/issues/4493, which also includes an additional refactoring of the StorageWriter to make it more efficient.

PR #4496, opened 2020-01-20
Issue 4495: Recording WARN and ERROR log entries as metrics.

### LINKED ISSUE #4495: Pravega Metrics for logged ERRORs and WARNs
**Problem description**
It would be nice if we could create 2 new metrics that tracked the number of errors and warning we log. This will enable us to quickly determine if we see an increase in errors (per component or overall) by looking at a metric dashboard.

**Problem location**
All components.

This can be done in a non-invasive way using the Log4j configuration:
1. Custom appender.
    - We create a custom appender  that implements `Appender<ILoggingEvent>`
    - We override the `Appender.doAppend(ILoggingEvent)` method and intercept only those events that have `getLevel()` equal to `Level.ERROR` or `Level.WARN`.
    - We create Counters (`logged_errors` and `logged_warnings`) and increment their value accordingly. For extra accuracy, we can pass `ILoggingEvent.getLoggerName()` as a custom tag so we can filter out by component.
2. Wire up custom appender using `logback.xml`:
    - Define custom appender `<appender name="Metrics" class="path-to-custom-appender-class">`
    - Link it up. Modify `<root level="...">` by adding `<appender-ref ref="Metrics"/>`

** Example implementation**

MetricsAppender.java
```
public class MetricsAppender implements ch.qos.logback.core.Appender<ILoggingEvent> {
    private static final DynamicLogger DYNAMIC_LOGGER = MetricsProvider.getDynamicLogger();

    @Override
    public String getName() {
        return "Metrics Appender";
    }

    @Override
    public void doAppend(ILoggingEvent o) throws LogbackException {
        if (o.getLevel() == Level.ERROR) {
            DYNAMIC_LOGGER.incCounterValue("logged_errors", 1, o.getLoggerName());
        } else if (o.getLevel() == Level.WARN) {
            DYNAMIC_LOGGER.incCounterValue("logged_warnings", 1, o.getLoggerName());
        }
    }

   // Removed unimplemented methods for brevity.
}
```

logback.xml
```
<configuration scan="true">
   <!-- All other appenders defined here -->

    <appender name="Metrics" class="io.pravega.segmentstore.server.host.MetricsAppender"/>
    <root level="${log.level:-DEBUG}">
        <appender-ref ref="consoleAppender"/> <!-- Preexisting appender -->
        <appender-ref ref="Metrics"/>
    </root>
</configuration>
```

andreipaduroiu:
**Change log description**  
- Recording all WARNs and ERRORs as metrics with logging class name and exception name as tags.

**Purpose of the change**  
Fixes #4495.

**What the code does**  
- Created a new Log Appender to intercept all log entries that have severity set to WARN or ERROR. 
- All these are reported into metrics via 2 distinct counters (one for WARN, one for ERROR), and using the source class/log name as a tag, and exception name (if any) as another tag.

**How to verify it**  
Unit tests updated. A longevity with metrics should also be run to verify this.

### DIFF
--- a/shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
+++ b/shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
@@ -64,6 +64,10 @@ public final class MetricsNames {
     // PREFIX can be loaded from config when config becomes dependency.
     public static final String PREFIX = "pravega" + ".";
 
+    // Metrics for all services
+    public static final String LOG_ERRORS = PREFIX + "logged_errors";     // Counter
+    public static final String LOG_WARNINGS = PREFIX + "logged_warnings"; // Counter
+
     // Metrics in Segment Store Service
     // Segment-related stats
     public static final String SEGMENT_CREATE_LATENCY = PREFIX + "segmentstore.segment.create_latency_ms";              // Histogram
--- a/build.gradle
+++ b/build.gradle
@@ -217,6 +217,7 @@ project ('shared:metrics') {
         compile group: 'io.micrometer', name: 'micrometer-registry-influx', version: micrometerVersion
         // https://mvnrepository.com/artifact/io.micrometer/micrometer-registry-influx/<micrometerVersion>
         compile project(':common')
+        compile group: 'ch.qos.logback', name: 'logback-classic', version: qosLogbackVersion
     }
 
     javadoc {
--- a/dist/conf/logback.xml
+++ b/dist/conf/logback.xml
@@ -36,8 +36,11 @@ You may obtain a copy of the License at
         </encoder>
     </appender>
 
+    <appender name="Metrics" class="io.pravega.shared.metrics.MetricsLogAppender"/>
+
     <root level="${log.level:-INFO}">
         <appender-ref ref="consoleAppender"/>
         <appender-ref ref="FILE"/>
+        <appender-ref ref="Metrics"/>
     </root>
 </configuration>
--- a/shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
+++ b/shared/metrics/src/main/java/io/pravega/shared/MetricsTags.java
@@ -28,6 +28,7 @@ public final class MetricsTags {
     public static final String TAG_SEGMENT = "segment";
     public static final String TAG_TRANSACTION = "transaction";
     public static final String TAG_EPOCH = "epoch";
+    public static final String TAG_CLASS = "class";
 
     private static final String TRANSACTION_DELIMITER = "#transaction.";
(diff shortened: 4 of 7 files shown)


RaulGracia:
@andreipaduroiu this proposal is really great. If we could get as a tag not only the class logger sending the log (to know the component originating the error/warn), but also the exception being thrown it would be great because:
1) We could see in the dashboard at the same time "which" component is experiencing problems and "what" is the problem being thrown.
2) We could create alerts based on logged exceptions (e.g., if we find `DataCorruptionException` send alert).

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -64,6 +64,10 @@
     // PREFIX can be loaded from config when config becomes dependency.
     public static final String PREFIX = "pravega" + ".";
 
+    // Metrics for all services
+    public static final String LOG_ERRORS = PREFIX + "logged_errors";     // Counter
```
kevinhan88: Why not just name as ERROR_COUNTER, WARN_COUNTER? LOG_ERRORS sounds logs more than counter.

kevinhan88:
Yes this approach to generate summary of Error/Warnings, together with another initiative to export all logs as metrics, could potentially change the way we view and use logs. I'm also eager to see how it helps.

kevinhan88:
Strong recommend we implement this as Rate (Meter) instead of Counter, thinking the eventual use case:
Counter only gives us the total of errors/warns ever since.
Rate tells us not only the total, but also the distribution of them, which is even more useful for us to spot the patterns.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
andreipaduroiu: That would be inconsistent with the naming of the other metrics defined in this class.

Plus, I want the name "Log" in it so that I can make it clear that the source of the information is the logs, and not something else.

kevinhan88:
I suggested the rewording of the original issue as below:
It would be nice if we could create 2 new metrics that track events of errors and warning ...

RaulGracia:
I have tested this PR and it works fine, this will be very useful to understand rapidly the problems of the system:
log-metrics

RaulGracia:
@andreipaduroiu one question: should we also add `CRITICAL` messages to this PR? I don't know how widely used are they in our code base, but perhaps we should move some `ERROR` messages to `CRITICAL` (in another PR, of course).

andreipaduroiu:
@RaulGracia 
We don't use CRITICAL in our code base. Just TRACE, DEBUG, INFO, WARN and ERROR. So I think we should be good for now.

PR #4499, opened 2020-01-21
Issue 4456: Consolidate get active txn calls during commit transactions for multiple transactions

### LINKED ISSUE #4456: Transaction Create and Commit Performance and back pressure
**Problem description**
We have a relatively low rate of transactions per second. Expectedly, it goes down further with larger number of segments as well. 

Instrument and identify bottlenecks and improve transaction creation and commit. 

One obvious improvement is to consolidate metadata updates for batch of transactions in Commit Workflow. (this will reduce number of calls to segment store and gain some performance for PravegaTables based store. It wont improve anything for zookeeper based store though). 

Another improvement could to be reduce number of wire calls we make to commit transactions - currently we identify a batch of transactions and then make calls to merge segments one after the other honoring transaction commit order. This can be consolidated into single merge call per stream segment which sends a list of txn segments to be merged. 

Other than the improvements, we also need a way to induce back pressures on create transactions. Otherwise if the rate of creation and faster than the rate at which transactions are committed, we can possibly have an ever growing backlog of transactions to be committed. 
The problem really comes in because we merge transactions in order which means though multiple transactions can be created concurrently, they can only be merged one transaction at a time. 
Take for example 10 writers, each creating transactions and writing data and then calling commit on them. So we have 10 transactions that were created concurrently and their commits were issued in some order. Since the transactions are committed one after another, even though transactions were marked for committing together, the commit latency of transactions will include latencies in committing all previous transactions. This can lead to an ever increasing backlog. 

One idea for backpressure would be to put `create transactions` requests serialized with commits workflows, such that they are executed on the same event processor. This way we can automatically slow down transaction creation until previous commits are performed. 
This will be detrimental to performance though as transactions lifecycle may increase even further. So we need to look at other approaches which induce some back pressures without affecting performance in regular cases. 

**Problem location**
Controller's transaction related workflows

Improve the performance

shiveshr:
**Change log description**  
Contains first set of improvements to transaction commit where we consolidate multiple (parallel) requests into pravega tables into a single request.

**Purpose of the change**  
Partially Fixes #4456

**What the code does**  
During commit workflow execution we find all transactions in committing state and batch them and perform commit on this batch of transactions one after another. 
This has one minor optimization where we earlier retrieve transaction metadata from the store by making multiple calls into the store for each transaction in lowest epoch. With this change we make a single readTableKeys call where we read multiple keys in a single request instead. 
The gains will not be too significant as we made those calls in parallel earlier, but it reduces the number of store requests we make to segment store. 

As follow up, we will add some more improvements to transaction commits to consolidate merge segment requests as well. 

**How to verify it**  
Unit tests added.


shrids:
System test run with this changeset was successful.

PR #4525, opened 2020-01-27
Issue:  Prevent readers from waiting indefinitely if there is a partial/no response from the Segment Store.

shrids:
(Under discussion)

**Change log description**  
- Ensure Revisioned Stream resends the read request if the Segment store does not respond within 30 seconds after confirmation that the write-offset is greater than the current read-offset.
- Ensure the Event Stream Reader resends the read request if the complete WireCommand frame is not received within 30 seconds.

**Purpose of the change**  
Fixes #

**What the code does**  
This PR is in continuation of PR #4430. During the longevity testing of Pravega, we have observed scenarios where part of the WireCommand is never received by the client. This is observed post a network glitch on a Pravega cluster deployment that uses SDN with load balancers. The underlying network connection was never dropped nor did the packet arrive at the client. This PR addresses this issue which can potentially cause the readers to be stuck waiting for the last part of the data.

This change ensures the following:

- RevisionedStreamClient now resends the read request if there is no response from SSS. This resend logic is invoked only after the SegmentMetaClient confirms that the write-offset > the current read-offset i.e. the SSS has the data.
- Event StreamReader now resends the read request in the case of partial WireCommand frames being received by the client. Here too we are sure that the SSS has the data.

Note: A warning message indicating the partial response will be logged by the client along with the missing bytes.

**How to verify it**  
All the existing and newly added tests should continue to work.

### DIFF
--- a/client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
@@ -29,7 +29,8 @@ import lombok.extern.slf4j.Slf4j;
 @ToString
 class EventSegmentReaderImpl implements EventSegmentReader {
 
-    // Flaky network
+    /* Partial data timeout. This timeout is the maximum amount of time the reader will wait in the case of
+     * partial data being received by the client. After this timeout the client will resend the request. */
     private static final long READ_TIMEOUT_MS = 500;
 
     @GuardedBy("$lock")
@@ -92,7 +93,7 @@ class EventSegmentReaderImpl implements EventSegmentReader {
         }
         while (headerReadingBuffer.hasRemaining()) {
             if (in.read(headerReadingBuffer, READ_TIMEOUT_MS) == 0) {
-                log.warn("Timeout out while trying to read WireCommand headers during read of segment {} at offset {}",
+                log.warn("Timeout out while trying to read WireCommand header during read of segment {} at offset {}",
                         in.getSegmentId(), in.getOffset());
                 throw new TimeoutException("Timeout while trying to read WireCommand headers");
             }
--- a/client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
@@ -73,7 +73,7 @@ class EventSegmentReaderImpl implements EventSegmentReader {
             return result;
         } catch (TimeoutException e) {
             timeout = true;
-            log.warn("Timeout observed while trying to read data from Segment store, the read request will be retransmitted. Details: {}", e.getMessage());
+            log.warn("Timeout observed while trying to read data from Segment store, the read request will be retransmitted");
             return null;
         } finally {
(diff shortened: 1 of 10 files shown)


client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
```diff
@@ -28,6 +29,10 @@
 @ToString
 class EventSegmentReaderImpl implements EventSegmentReader {
 
+    /* Partial data timeout. This timeout is the maximum amount of time the reader will wait in the case of
+     * partial data being received by the client. After this timeout the client will resend the request. */
+    static final long READ_TIMEOUT_MS = 500;
```
tkaitchuck: This seems really aggressive. I don't see any reason this should be more aggressive than or different from `RevisionedStreamClientImpl.READ_TIMEOUT_MS`. In both cases we know the server has data and we are waiting to receive it.

client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
```diff
@@ -54,30 +59,44 @@ public long getOffset() {
...
+        } catch (TimeoutException e) {
+            timeout = true;
+            log.warn("Timeout observed while trying to read data from Segment store, the read request will be retransmitted. Details: {}", e.getMessage());
+            return null;
         } finally {
-            LoggerHelpers.traceLeave(log, "read", traceId, in.getSegmentId(), originalOffset, timeout, success);
+            LoggerHelpers.traceLeave(log, "read", traceId, in.getSegmentId(), originalOffset, firstByteTimeoutMillis, success);
             if (!success) {
-                in.setOffset(originalOffset);
+                if (timeout) {
```
tkaitchuck: I think this can be simplified to: `in.setOffset(originalOffset, timeout);`

client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
```diff
@@ -54,30 +59,44 @@ public long getOffset() {
...
-        int read = in.read(headerReadingBuffer, timeout);
+        int read = in.read(headerReadingBuffer, firstByteTimeoutMillis);
         if (read == 0) {
+            // a resend will not be triggered in-case of a firstByteTimeout.
             return null;
         }
         while (headerReadingBuffer.hasRemaining()) {
-            in.read(headerReadingBuffer, Long.MAX_VALUE);
+            if (in.read(headerReadingBuffer, READ_TIMEOUT_MS) == 0) {
+                log.warn("Timeout out while trying to read WireCommand header during read of segment {} at offset {}",
```
tkaitchuck: Given three instances of this if and throw, it probably makes sense to have a private method.

client/src/main/java/io/pravega/client/segment/impl/EventSegmentReaderImpl.java
shrids: ok.Changed it to 30 seconds.

shrids:
System tests were successful with this changeset ref: `test-pravega-continuous-1/559/`

PR #4548, opened 2020-02-04
Issue 3961: Prune Idle connections

### LINKED ISSUE #3961: Prune idle connections in connection pooling
**Problem description**
Prune idle connections in a connection pool

**Problem location**
Client connection pooling.

Ensure connections are closed automatically after a timeout.

shrids:
**Change log description**  
- Prune connections which now point to an invalid IP address.
- Prune connections that are idle.

**Purpose of the change**  
Fixes #3961 

**What the code does**  
This PR periodically prunes connections from the connection pool that satisfy the following conditions

1. Connections that are idle.
2. If connections pointing to a docker container does not terminate even after the docker container has been assigned a newer IP the requests will be sent to the older IP address. This is possible in scenarios where the client is behind a proxy or load balancing layer. 
The current code attempts to do a DNS lookup and verify if the IP Address of the container has changed. If there is a change then such connections are also closed.

**How to verify it**  

### DIFF
--- a/client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java
@@ -77,7 +77,7 @@ public class ConnectionPoolImpl implements ConnectionPool {
     static {
         // The default caching policy for name lookup cache is -1 which is forever.
         // Change this configuration to 60 seconds.
-        java.security.Security.setProperty("networkaddress.cache.ttl" , "60");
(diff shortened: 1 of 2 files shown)


shrids:
Yes, the `getInetAddress would trigger DNS calls only if the cache has been invalidated. This is configured via `networkaddress.cache.ttl` configuration...

shrids:
Closing this PR since the issue is caused when DNS records are not deleted from DNS provider. This can cause the name lookup issue when a new instance of Pravega is deployed.

PR #4567, opened 2020-02-21
Issue 4565: (SegmentStore) Improving Data Frames memory allocation

### LINKED ISSUE #4565: Investigate better Data Frame memory allocation
**Problem description**
We allocate 1MB for each Data Frame, regardless of whatever size we might need. This can cause excessive memory consumption.

**Problem location**
Segment Store.

TBD.

andreipaduroiu:
**Change log description**  
Changed the underlying buffer for `DataFrame` from a `ByteArraySegment` backed by a single byte array to a memory efficient composite array that only allocates memory when needed.

**Purpose of the change**  
Fixes #4565.

**Background**
We format the BK/Tier1 writes using a class called `DataFrame`. When creating instances of this class we have no way of knowing how much data we'd end up writing (since we don't know how much of the existing queued data is valid and how much it will serialize to, and since we don't know if any other operations will arrive as we are constructing it). As such, we had two options for this class. We could either allocate a full 1MB buffer every time we create it (and fill as much as we need), or begin with a small buffer and then resize it as needed. The second approach ends up doing a lot of allocations and memory copies and the first one is less CPU intensive, but at the cost of over-allocating memory.

It turns out that in some cases we could be over-allocating memory by a lot. For example, a recent perf test indicated that even though we had about 420MB of data "in-flight" to BK, we were actually holding up about 2GB of heap, with the difference going to overallocation.

**What the code does**  
Created a `CompositeByteArraySegment` (and parent interface `CompositeByteArrayView`)
- These extend from `BufferView` so they're interchangeable with other uses of existing `ByteArraySegment` classes.
- This class creates a "virtual" index-based data structure mimicking an array, but it doesn't allocate all the bytes at once. 
- It breaks down the "virtual array" into smaller blocks (configurable). Each such block represents a contiguous range within the larger array and is only allocated when the first byte within that range is set.
    - If not allocated, upon reading all bytes within a block range will be read as 0 (same as a freshly initialized array).
- The goal of this class is to reduce the amount of memory used by `DataFrame`s by allocating memory in small increments as we populate it.

Integration with the rest of the code:
- Changed `DataFrame` to wrap a `CompositeByteArraySegment` instead of `ByteArraySegment`.
- Changed `DurableDataLog` interface to use `CompositeArrayView` instead of `ArrayView` for writes.
- Changed `BookKeeperLog` & `bookkeeper.Write` to convert a `CompositeArrayView` into a Netty `CompositeByteBuf` which is eventually passed on to BK's `asyncAddEntry`.
    - This is needed since the `CompositeArrayView` organizes data as individual arrays and there is no point in allocating a new byte array and copy them there just for this purpose.

**How to verify it**  
All tests must pass.

### DIFF
--- a/common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
+++ b/common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
@@ -179,24 +179,8 @@ public class CompositeByteArraySegment implements CompositeArrayView {
 
     @Override
     public byte[] getCopy() {
-        // This method is a special instance of collect(), however we can optimize it since we can skip over unallocated
-        // components without having to create a dummy one.
-        int startId = getArrayId(0);
-        int endId = getArrayId(this.length - 1);
         byte[] result = new byte[this.length];
-        int resultOffset = 0;
-        int arrayOffset = getArrayOffset(resultOffset); // The first array may need an offset, if this.startOffset > 0.
-        for (int i = startId; i <= endId; i++) {
-            byte[] array = getArray(i, false); // Don't allocate array if not allocated yet.
-            int copyLength = Math.min(result.length - resultOffset, this.arraySize - arrayOffset);
-            if (array != null) {
-                System.arraycopy(array, arrayOffset, result, resultOffset, copyLength);
-            }
-
-            resultOffset += copyLength;
-            arrayOffset = 0; // After processing the first array (handling this.startOffset), all other array offsets are 0.
-        }
-
+        copyTo(ByteBuffer.wrap(result));
         return result;
     }
 
--- a/common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
+++ b/common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
@@ -9,6 +9,7 @@
  */
 package io.pravega.common.util;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterators;
 import io.pravega.common.Exceptions;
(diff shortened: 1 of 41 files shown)


common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
```diff
@@ -0,0 +1,273 @@
...
+import java.util.ArrayList;
+import lombok.Getter;
+import lombok.NonNull;
+
+/**
+ * A composite, index-based array-like structure that is made up of one or more individual arrays of equal size. Each
+ * component array maps to a contiguous offset range and is only allocated when the first index within its range needs
+ * to be set (if unallocated, any index within its range will have a value of 0).
+ */
+public class CompositeByteArraySegment implements CompositeArrayView {
```
eolivelli: Why aren't we using Netty ByteBufs ? There is already lot of support for composition and they support direct memory?

eolivelli:
Did you evaluate to switch to using Netty ByteBufs all over Pravega code?
They play well with BK and GRPC, they are reference counted and they support direct memory.
Netty is very well maintained and very live

andreipaduroiu:
I don't want to put dependencies on Netty all over Pravega. 

For most scenarios where short lived buffers are needed, plain old Java byte arrays are sufficient and provide the best performance you can get. I don't need reference counting or whatever other features Netty ByteBufs provide.

There is no feature in Netty that would solve the problem I am addressing here. Please wait until I finish drafting the PR before asking clarifying questions.

RaulGracia:
@andreipaduroiu I verified that in terms of performance there is no issue with this change. @tkaitchuck would you like to have a look to this PR, please?

common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
```diff
@@ -0,0 +1,293 @@
...
+
+            length -= arrayLength;
+            arrayOffset = 0; // After processing the first array (handling this.startOffset), all other array offsets are 0.
+        }
+
+        assert length == 0 : "Collection finished but " + length + " bytes remaining";
+    }
+
+    @Override
+    public byte[] getCopy() {
```
tkaitchuck:
```
byte[] result = new byte[this.length];
copyTo(ByteBuffer.wrap(result));
return result;
```

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Write.java
```diff
@@ -60,12 +68,19 @@
         this.beginAttemptTimer = new AtomicReference<>();
     }
 
+    private ByteBuf convertData(CompositeArrayView data) {
```
tkaitchuck: Is this the best place for this method?

common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
andreipaduroiu: Done.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Write.java
andreipaduroiu: This method converts a `CompositeByteArraySegment` to a Netty ByteBuf. `CompositeByteArraySegment` is in `common` which doesn't have a dependency on Netty, so I can't do that there.

This is the main reason why I added the `collect` method in `CompositeArrayView`, so that I can essentially convert it to anything else without having to add extra dependencies in it.

PR #4579, opened 2020-02-27
Issue 4578: Early wake reader in response to server reply.

tkaitchuck:

**Change log description**  
Unblock a reader thread which is waiting in the event that data arrives from the server  on another segment.

**Purpose of the change**  
Re-Implements the end-to-end latency optimization that had to be reverted in https://github.com/pravega/pravega/pull/4082 but without using futures to avoid the problems in the Java 8 implementation.

**What the code does**  
Uses a semephore to wake the reading thread when a reply from the server is received.

**How to verify it**  
All existing tests should continue to pass, but lower end-to-end latency should be observed on multi-segment readers that are at the tail of the stream.


client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -201,20 +213,19 @@ private String updateGroupStateIfNeeded() throws ReaderNotInReaderGroupException
                     releaseSegmentsIfNeeded(position);
                     atCheckpoint = null;
                 }
-                return null;
+                checkpoint = groupState.getCheckpoint();
```
shrids: This change would cause to complete checkpointing of all the pending silent checkpoints for this reader. In the older code the application would have to invoke `readNextEvent()` again to clear this, right ?

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -59,7 +61,7 @@
 public class EventStreamReaderImpl<Type> implements EventStreamReader<Type> {
 
     // Base waiting time for a reader on an idle segment waiting for new data to be read.
-    private static final long BASE_READER_WAITING_TIME_MS = 10;
+    private static final long BASE_READER_WAITING_TIME_MS = ReaderGroupStateManager.TIME_UNIT.toMillis();
```
shrids: The waiting time is now increased from 10ms to 1000ms .Any specific reason for this ?

client/src/main/java/io/pravega/client/segment/impl/SegmentInputStreamFactory.java
```diff
@@ -48,10 +49,11 @@
      * does not exist.
      *
      * @param segment The segment to create an input for.
+     * @param hasData A semephore that will have `release` called when data is available.
```
andreipaduroiu: semaphore

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -155,6 +160,13 @@
                                    new EventPointerImpl(segment, offset, length), null);
     }
 
+    private void blockFor(long timeoutMs) {
+        Exceptions.handleInterrupted(() -> {
+            @SuppressWarnings("unused")
+            boolean aquired = segmentsWithData.tryAcquire(timeoutMs, TimeUnit.MILLISECONDS);
```
shrids: Typo

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -127,7 +131,8 @@
             }
             EventSegmentReader segmentReader = orderer.nextSegment(readers);
             if (segmentReader == null) {
-                Exceptions.handleInterrupted(() -> Thread.sleep(firstByteTimeoutMillis));
+                blockFor(firstByteTimeoutMillis);
```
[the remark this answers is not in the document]
tkaitchuck: yes

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -127,7 +131,8 @@
...
+                segmentsWithData.drainPermits();
```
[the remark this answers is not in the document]
tkaitchuck: yes

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
tkaitchuck: yes

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
tkaitchuck: Originally it was 1000ms and we used futures to wake up the blocked thread. However futures created the issue linked above, so we changed it to 10ms so that while it would not wakeup in response to data being available, it least it would check every 10ms. Now that we have a wakeup again I am restoring it to 1000ms. This is fine because it will awake in response to data. We still need to wake up eventually though, the source of the timeout shows why, even if there is no data, there might be an update to the group state, which we pull in updates for every 1000ms.

PR #4584, opened 2020-02-28
Issue 4583: (SegmentStore) StorageWriter auto-close in case of Tier 2 write stall

### LINKED ISSUE #4583: StorageWriter should shut down if an iteration exceeds a maximum time limit
**Problem description**
Not all Tier 2 implementations natively support timeouts. If a Tier 2 call hangs indefinitely, that will cause the StorageWriter to stall, which will eventually lead to other issues upstream, such as cache getting full or reaching the maximum number of active segments in a container.

**Problem location**
Storage Writer.

- Define a Storage Writer Iteration timeout and express it as a function of the Flush Timeout (a multiple of this number).
- If, during a particular Iteration, no activity has been recorded for the duration of this timeout, immediately shut down the Storage Writer, which should cause the owning Segment Container to shut down and restart (after which hopefully things will work well again).
    - A full container restart will increment the Container's Epoch. This, combined with fencing techniques implemented by StorageWriter/Tier2, will ensure that if whatever hanging operation did eventually execute and resume, it will not corrupt the data in Tier 2.
- Optionally, collect and dump the operations that were left hanging.

andreipaduroiu:
**Change log description**  
The Storage Writer now tracks all operations it performs during an iteration and shuts down if it doesn't observe any progress after a specified amount of time.

**Purpose of the change**  
Fixes #4583.

**What the code does**  
See #4583 for a scenario that triggered this.

The StorageWriter now tracks the execution of all async Tier 2 flushes it performs. This is done for each Segment and for each of that Segment's processors (AttributeAggregator, SegmentAggregator, WriterTableProcessor).  Once a flush is initiated, it is tracked until it completes, either normally or with an exception. The `TaskTracker` can calculate the amount of time each pending task has been "idle" for, and returns the maximum of that value.

Added a config value `writer.idleTimeoutMillis` which is used by the StorageWriter in the following way:
- If a pending task (tracked by `TaskTracker`) has been idle for an amount of time longer than this value, then the `StorageWriter` will auto-close. It will attempt to abort any ongoing operations.
    - It will dump the list of pending tasks (with sufficient information to determine which is which) into the logs.
- The owning Segment Container will also auto-close, and will be restarted quickly.
    - After a restart, the new `StorageWriter` instance will resume operations.
    - If whatever operation from the old `StorageWriter` instance does eventually get unstuck, it should not have any negative effect (such as corrupting data) in Tier 2 since the new Segment Container will have a higher epoch than the old one, so the fencing logic within the Tier 2 adapter will prevent it from executing.

**How to verify it**  
New unit tests added to cover new functionality.

### DIFF
--- a/config/config.properties
+++ b/config/config.properties
@@ -570,7 +570,7 @@ extendeds3.url=localhost:9020
 
 # The maximum timeout (in milliseconds) to wait for new items to come into the DurableLog (in case we reached the tail).
 # Valid values: Positive integer.
-#writer.maxReadTimeoutMillis=1800000
+#writer.maxReadTimeoutMillis=180000
 
 # The amount of time to sleep (in milliseconds) if an exception was encountered during the execution of a Writer's iteration.
 # This can be used as a way to "back off" in cases a transient error occurred and not go into a very tight loop while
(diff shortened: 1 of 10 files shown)


andreipaduroiu:
No longer needed.

PR #4588, opened 2020-03-03
Issue 4568: Key-Value Table Client Contracts

### LINKED ISSUE #4568: Key Value Table APIs
**Problem description**
Implement Key Value Table APIs.

See [PDP-39](https://github.com/pravega/pravega/wiki/PDP-39-Key-Value-Tables) for details.

**Problem location**
Client.

Implement.

andreipaduroiu:
**Change log description**  
- Created the publicly visible `KeyValueTable` interface which defines all operations that can be performed on a Key-Value Table. Created/updated ancillary interfaces and classes that are referenced by this one.
- Refactored `TableSegment` to not require generic Keys and Values but work with ByteBufs instead.
- Updated `SegmentHelper` to make use of the updated `TableSegment` ByteBuf API.

**Purpose of the change**  
Fixes #4568.

**What the code does**  
New contracts added:
- `KeyValueTable`

Moved some classes from `impl` into `tables` package:
- Exceptions: `ConditionalTableUpdateException`, `BadKeyVersionException`, `NoSuchKeyException`
- Iterators: `IteratorItem` and `IteratorState`
- `TableKey`, `KeyVersion` and `TableEntry`: made these classes instead of interfaces.

Changed `TableSegment` from requiring a generic Key and generic Value to using ByteBufs
- Added `TableSegmentKey`, `TableSegmentKeyVersion` and `TableSegmentEntry` that pertain to `TableSegments`
- These classes will be used exclusively by the `KeyValueTableImpl` (not in this PR) and `SegmentHelper` (due to it currently using them).

Refactored `SegmentHelper`
- It was using the generic `TableKey`/`TableEntry`/`KeyValue`
- Now it uses the `ByteBuf` versions of those (the internal ones, in `impl`).
- No functional changes have been made to`SegmentHelper` or upstream code. All changes are purely due to refactoring and most code changed was to maintain the ByteBuf lifecycles (See Javavdoc on `TableSegment` for details).

NOTE: there is no implementation for `TableSegment` in this PR. A subsequent PR will introduce that.

**How to verify it**  
All unit tests must pass.


andreipaduroiu:
@shrids I added you as a reviewer because I touched `SegmentHelper` and `PravegaTablesStoreHelper` because I touched their APIs. Please have a look at those two classes. Thank you.

client/src/main/java/io/pravega/client/tables/IteratorState.java
```diff
@@ -0,0 +1,54 @@
...
+        return this.token.readableBytes() == 0;
+    }
+
+    /**
+     * Deserializes the IteratorState from its serialized form obtained from calling {@link #getToken()} .
+     *
+     * @param serializedState A serialized {@link IteratorState}.
+     * @return The IteratorState object.
+     */
+    public static IteratorState fromBytes(ByteBuf serializedState) {
```
tkaitchuck: Is this intended to be a public method? If so we cannot expose ByteBuf as it is from netty. This will need to be ByteBuffer as will the token above.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,247 @@
...
+ * or more Table Segments (depending on the Key-Value Table configuration chosen when it was created). Such partitioning
+ * enables the Key-Value Table to be distributed across the Pravega cluster but also introduces some constraints for
+ * certain operations (such as multi-key/entry atomic updates). See below for details.
+ * <p>
+ * Key Families are used to group related Keys together in the same Table Partition, which allows multiple
+ * keys/entries belonging to the same Key Family to be updated/removed atomically.
+ * <ul>
+ * <li> Multiple Keys/Entries in the same Key Family can be updated or removed atomically (either all at once or none).
+ * <li> Iterating through all Keys/Entries in the same Key Family is possible.
+ * <li> The same Key may exist in multiple Key Families or even not be associated with any Key Family at all. Such keys
```
tkaitchuck: This concept is already complex enough without the additional possibility of some keys not being in a family. I think we can just make it mandatory. If users want them to be random they can just pass the same string twice.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,247 @@
...
+ * are treated as distinct keys and will not interfere with each other (i.e., if key K1 exists in Key Families F1 and F2,
+ * then F1.K1 is different from F2.K1 and both are different from K1 (no Key Family association).
+ * <li> Keys that do not belong to any Key Family will be uniformly distributed across the Key-Value Table Partitions and
+ * cannot be used for multi-key/entry atomic updates or removals or be iterated on.
+ * <li> Improper use of Key Families may result in degraded performance. If a disproportionate number of Keys/Entries
```
tkaitchuck: This is explaining that "improper" use is bad, but it does so without first explaining how the system is architected and what a 'proper' use looks like given that architecture.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,247 @@
...
+ * Table will be one where Keys are not part of any Key Families or the number of Keys in each Key Family is approximately
+ * the same. An improperly designed Key-Value Table will have all Keys part of a single Key Family which will cause a
+ * single Table Segment to bear the full storage and processing load of the entire Key-Value Table.
+ * </ul>
+ * <p>
+ * Types of Updates:
+ * <ul>
+ * <li> Unconditional Updates will insert and/or overwrite any existing values for the given Key, regardless of whether
+ * that Key previously existed or not, and regardless of what that Key's version is.
+ * <li> Conditional Updates will only overwrite an existing value if the specified version matches that Key's version.
```
tkaitchuck: With these conditional and unconditional updates, I think we should blatantly copy the verbage from concurrenthashmap, because they are standardized in the Java Collections.
`put`, `putAll`, `putIfAbsent`, `remove`, and `replace`. 
Java developers will recognise the meaning of these terms.

client/src/main/java/io/pravega/client/tables/KeyVersion.java
```diff
@@ -0,0 +1,65 @@
...
+     * update or removal being performed, conditioned on the {@link TableKey} not existing at the time of the operation.
+     * See {@link KeyValueTable} for details on conditional/unconditional updates.
+     */
+    public static final KeyVersion NOT_EXISTS = new KeyVersion(null, TableSegmentKeyVersion.NOT_EXISTS);
+
+    /**
+     * The Segment where this Key resides. May be null if this is a {@link #NOT_EXISTS} or {@link #NO_VERSION}
+     * {@link KeyVersion}.
+     */
+    @Getter
```
tkaitchuck: Why is there a getter for this? Generally we don't expose segments

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -376,27 +366,24 @@ private String getTransactionName(String scope, String stream, long segmentId, U
...
+                                                                              final long clientRequestId) {
         final Controller.NodeUri uri = getTableUri(tableName);
         final WireCommandType type = WireCommandType.UPDATE_TABLE_ENTRIES;
-        List<ByteBuf> buffersToRelease = new ArrayList<>();
         List<Map.Entry<WireCommands.TableKey, WireCommands.TableValue>> wireCommandEntries = entries.stream().map(te -> {
             final WireCommands.TableKey key = convertToWireCommand(te.getKey());
-            ByteBuf valueBuffer = wrappedBuffer(te.getValue());
-            buffersToRelease.add(key.getData());
-            buffersToRelease.add(valueBuffer);
-            final WireCommands.TableValue value = new WireCommands.TableValue(valueBuffer);
```
shrids: @andreipaduroiu Could you help undertstand how ByteBuf release is being handled here?

client/src/main/java/io/pravega/client/tables/IteratorItem.java
```diff
@@ -0,0 +1,35 @@
...
+@Data
+public class IteratorItem<T> {
+    /**
+     * Gets an {@link IteratorState} that can be used to reinvoke {@link KeyValueTable#entryIterator} or
+     * {@link KeyValueTable#keyIterator} if a previous iteration has been interrupted (by losing the pointer to the
+     * {@link AsyncIterator}), system restart, etc.
+     */
+    private final IteratorState state;
+    /**
+     * A List of items that are contained in this instance. The items in this list are not necessarily related to each
```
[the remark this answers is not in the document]
andreipaduroiu: I removed this verbiage. Having a `List` gives the most flexibility in the API in case we want to return KVPs in a specific order in the future.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
andreipaduroiu: I moved it up into `PravegaTablesStoreHelper`. It is the responsibility of whomever generates the `ByteBuf`s to release them. Now that they are no longer generated inside `SegmentHelper`, `PravegaTablesStoreHelper` is the one that both creates and releases them.

I needed to to this since the eventual `TableSegmentImpl` class deals with `ByteBuf`s that come from the eventual `KeyValueTableImpl` class, so I had to standardize the usage across both implementations.

client/src/main/java/io/pravega/client/tables/KeyVersion.java
andreipaduroiu: Ok. Until I know how these will be exposed externally (probably some opaque serialization), I have decided to remove the getter for this one and make `getSegmentVersion` package-private. We'll revisit some of these once we have implementations done.

client/src/main/java/io/pravega/client/tables/IteratorState.java
andreipaduroiu: I separated this into an interface (`IteratorState` - in this package) and an implementing class (`IteratorStateImpl` - in the `impl` package). The interface only deals with `ByteBuffers`, but the implementation was already heavily used in the `SegmentHelper`, so I had to provide both `ByteBuffer` and `ByteBuf` methods for that one. 

Once this whole feature is merged in, we should put some effort into making `SegmentHelper` use `TableSegmentImpl` (coming in next PR), which will reduce the need for all of these extra methods.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: What I am concerned is that if users absolutely do not care about Key families, they will just do something like `put("", key, value)`, which will essentially group all keys into the same TableSegment, so there's no distribution done.

I am in favor of keeping both around.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: It is explained a few lines above:

```
 Each {@link TableKey} is hashed to a Table Partition which may be represented by one or more Table Segments

...

Key Families are used to group related Keys together in the same Table Partition, which allows multiple keys/entries belonging to the same Key Family to be updated/removed atomically.
```

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: The HashMap/ConcurrentHashMap methods do not have the rich semantics the ones in the KeyValueTableAre. 
`HashMap.put` performs an unconditional update of a single key and returns the existing key.
`HashMap.putAll` performs an unconditional update of multiple key and doesn't return anything.
Similarly with `remove`.

The `KeyValueTable` methods give you more flexibility than just simple `put/putAll/remove`. For example, when you update multiple entries at once, you can set a condition on 0, one or more of those entries; all the conditions on the update batch will need to be satisified in order for the entire batch to be accepted. In addition, you can mix conditions (one key may have no condition, one key may condition on it not existing before, and another key may be conditioned on a specific version). 

HashMap does not give you that - as a programmer, you'd have to write your own code to do something similar.

So I don't see how we can differentiate this using different API names given all the possibilities it supports.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
tkaitchuck: Perhaps we can then hide this inside of the key object itself. Have a Key type that we take which has an optional family field. That keeps this API clean, and there is no risk of misuse.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
tkaitchuck: Yes, put it unconditional. `replace` is the name of the API when it is conditional on a particular value, and `putIfAbsent` is when it is conditional on the non-existence of the value. I see no reason to name these `put` which is already established as meaning a non-conditional operation.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: That will cause trouble with batch updates and removals. Having the family separate from the keys forces the user to think that the update affects a single key family. Being able to mix keys from different key families together will not be supported, and we'll need to throw a runtime exception, which I think is worse than it is now.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: Ok, so how are we going to name the batch operations, which can take multiple keys, each having a condition of a different type (version, not exists, or no condition at all)?

Shall we have `put(single entry)`, `replace(single entry)`, `putIfAbsent(single entry)`, `remove(single key)` (this one for both conditional and unconditional), and `putAll(multiple entries)` and `removeAll`?

Each of these calls will need to validate their input properly. Currently all APIs take in `TableKey` or `TableEntry` which contain various fields, including a version. We'll need to ensure that we don't pass one of these with no version to a conditional API or viceversa.

tkaitchuck: No. I am proposing the following:
Rename KeyVersion to Version and move it off of key and put in on TableEntry. (It has to do with entries no keys after all) and have TableEntry extend Map.Entry.
Then on this class have:
`CompletableFuture<TableEntry<KeyT, ValueT>> get(String keyFamily, KeyT key);`
`CompletableFuture<List<TableEntry<KeyT, ValueT>>> getAll(String keyFamily, List<KeyT> keys);`
`CompletableFuture<Version> put(String keyFamily, KeyT key, ValueT value);`
`CompletableFuture<List<Version>> putAll(String keyFamily, Iterable<Entry<KeyT, ValueT> items);`
`CompletableFuture<Version> putIfAbsent(String keyFamily, KeyT key, ValueT value);`
`CompletableFuture<Version> replace(String keyFamily, KeyT key, Version version, ValueT value);`
`CompletableFuture<List<Version>> replaceAll(String keyFamily, Iterable<TableEntry<KeyT, ValueT>> items);`
`CompletableFuture<TableEntry<KeyT, ValueT>> remove(String keyFamily, KeyT key);`
`CompletableFuture<Boolean> remove(String keyFamily, KeyT key, Version version);`
`CompletableFuture<Boolean> removeAll(String keyFamily, Iterable<TableEntry<KeyT, ValueT>> items);`
In all of the above `keyFamily` may be null.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: Ok. I rephrased the entire paragraph.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,258 @@
...
+     *                  Key Family.
+     * @param key       The Key to insert or update.
+     * @param value     The Value to be associated with the Key.
+     * @return A CompletableFuture that, when completed, will contain the {@link KeyVersion} associated with the newly
+     * inserted or updated entry.
+     */
+    CompletableFuture<KeyVersion> put(@Nullable String keyFamily, @NonNull KeyT key, @NonNull ValueT value);
+
+    /**
+     * Conditionally inserts a new Entry in the {@link KeyValueTable} if and only if the given Key is not already present.
```
fpj: I suggest to rephrase this to the following:

```
Inserts a new Entry in the {@link KeyValueTable} only if the given Key is not already present.
```

The use of conditionally is redundant and the use of iff does not sound logically correct.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,258 @@
...
+     * @param keyFamily The Key Family for the all provided {@link TableEntry} instances.
+     * @param entries   An {@link Iterable} of {@link Map.Entry} instances to insert or update.
+     * @return A CompletableFuture that, when completed, will contain a List of {@link KeyVersion} instances which
+     * represent the versions for the inserted/updated keys. The size of this list will be the same as the number of
+     * items in entries and the versions will be in the same order as the entries.
+     */
+    CompletableFuture<List<KeyVersion>> putAll(@NonNull String keyFamily, @NonNull Iterable<Map.Entry<KeyT, ValueT>> entries);
+
+    /**
+     * Conditionally updates an existing Entry in the {@link KeyValueTable} if and only if the given Key exists and its
```
fpj: Same comment about rephrasing.

client/src/main/java/io/pravega/client/tables/KeyVersion.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+package io.pravega.client.tables;
+
+import io.pravega.client.tables.impl.TableSegmentKeyVersion;
+import java.io.Serializable;
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Version of a Key in a Table.
```
fpj: I find this notion of "key version" confusing. I understand you want the comparison to be on separate value compared to making it conditional on the value itself, and I think I understand why you associate the version to the key: you the key is what is constant in a key-value pair. But, the version really refers to the KV pair, and it indicates state changes to a particular KV pair, where the state change is reflected in the change of a value.

I'm saying all this to ask whether we can disassociate version from key. If we call it just version, then it would be ideal but it is too generic maybe and might conflict with some other definition of version. Another option is to call it `KVVersion` to associate to a KV pair.

Let me know what you think.

client/src/main/java/io/pravega/client/tables/impl/TableSegmentEntry.java
```diff
@@ -0,0 +1,133 @@
...
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Entry in a {@link TableSegment}.
+ */
+@Getter
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE)
+public class TableSegmentEntry {
```
fpj: Is this supposed to be public or package protected? Do I understand right that `TableSegment` isn't a concept we are exposing, that it is an internal implementation detail?

client/src/main/java/io/pravega/client/tables/impl/TableSegmentKey.java
```diff
@@ -0,0 +1,101 @@
...
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Key for a {@link TableSegment}.
+ */
+@Getter
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE)
+public class TableSegmentKey {
```
fpj: Same question about the visibility of this class.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: Fixed.

client/src/main/java/io/pravega/client/tables/BadKeyVersionException.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.tables;
+
+/**
+ * Exception that is thrown whenever a conditional {@link KeyValueTable} update failed due to the provided key version
+ * mismatching. This is different from {@link NoSuchKeyException}.
+ */
+public class BadKeyVersionException extends ConditionalTableUpdateException {
```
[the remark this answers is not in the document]
andreipaduroiu: `NoSuchKeyException` is another type of `ConditionalTableUpdateException` that indicates the Key does not exist when it was expected to exist. In some cases it is useful to the caller to know that the key does not exist instead of just "BadVersion". The Controller already uses this differentiation.

client/src/main/java/io/pravega/client/tables/IteratorState.java
```diff
@@ -0,0 +1,47 @@
...
+package io.pravega.client.tables;
+
+import io.netty.buffer.Unpooled;
+import io.pravega.client.tables.impl.IteratorStateImpl;
+import java.nio.ByteBuffer;
+
+/**
+ * Represents the state of a resumable iterator. Such an iterator can be executed asynchronously and continued after an
+ * interruption. Each iteration will result in a new request to the server (which is stateless). The entire state of
+ * the iterator is encoded in this object and is non-transferable between different types of iterations. The server will
```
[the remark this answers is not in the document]
andreipaduroiu: I am saying you may not use the state:
1. From one KVT to another KVT
2. From a KVT Key Iterator to the same KVT Entry iterator.

In other words, use the state only on the same method and on the same KVT.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: removed `iff`

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: fixed

client/src/main/java/io/pravega/client/tables/impl/TableSegmentEntry.java
andreipaduroiu: Because the Controller **already** uses these classes I have no option but to make them public,

client/src/main/java/io/pravega/client/tables/impl/TableSegmentKey.java
andreipaduroiu: Same answer.

client/src/main/java/io/pravega/client/tables/KeyVersion.java
andreipaduroiu: The KVT is a collection of KV pairs, then each KV pair is essentially immutable. As such it cannot have "versions". 

The KVT can also be viewed as a collection of Keys that have values associated, and every KVT mutation modifies the Value associated with a Key. As such, every such mutation creates a new KVP for that Key, which means the version of the Key needs to change. Hence the association of Versions to Keys (i.e., `KeyVersion`).

client/src/main/java/io/pravega/client/tables/IteratorState.java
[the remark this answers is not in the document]
andreipaduroiu: Reworded the paragraph to clarify.

client/src/main/java/io/pravega/client/tables/BadKeyVersionException.java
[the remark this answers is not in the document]
andreipaduroiu: Check out here (branch forked from this that implements the Table Segment Client) on how Wire commands are converted to such exceptions:
https://github.com/andreipaduroiu/pravega/blob/issue-4333-tables-segment-client/client/src/main/java/io/pravega/client/tables/impl/TableSegmentImpl.java#L227

client/src/main/java/io/pravega/client/tables/impl/TableSegmentEntry.java
fpj: Let's create an issue ahead of time so that we don't forget, please.

client/src/main/java/io/pravega/client/tables/KeyVersion.java
fpj: In the general case of a table, a triple {row, column, version} gives the value of a cell, e.g., as in hbase:

https://hbase.apache.org/book.html#versions

if we treat KV pairs as a special case of a table with a single column, then {key, version} gives the corresponding value of that version. I think that's why you say that the version is associated to the key, but if we think more generally about tables, then the version is really about the cell and its value.

I´d rather not call it `KeyVersion` because the key is a coordinate to the value, and the version is not about the coordinate. But, I don't fully understand whether that breaks the internal abstractions (e.g., `TableSegment`), in which case we might not have a choice. Let me know.

client/src/main/java/io/pravega/client/tables/impl/TableSegmentEntry.java
andreipaduroiu: #4647 and assigned to @shrids .

client/src/main/java/io/pravega/client/tables/KeyVersion.java
andreipaduroiu: This is not a multi-column table. It's a 2-column table, one column being key and the other value. We cannot use `version` to execute a lookup (i.e., get version X of key Y). We can only get the latest version of a Key. Versions are only there to enable conditional updates.

This API does nothing to change the meaning of versions. Versions are defined like this in Table Segments and the KeyValue Table API is just passing them through and exposing them. 

If you are looking to make such a semantic change, we'd have to likely redo most of the Table Segment work, which is out of scope for this project.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStoreHelper.java
```diff
@@ -429,69 +445,24 @@ public void invalidateCache(String table, String key) {
...
-     * @param table table
-     * @param fromStringKey function to deserialize key from String.
-     * @param fromBytesValue function to deserialize value from byte array
-     * @param filter filer predicate which takes key and value and returns true or false.
-     * @param limit maximum number of entries to retrieve
-     * @param <K> Type of Key
-     * @param <V> Type of Value
-     * @return CompletableFuture which when completed will have a map of keys and values of size bounded by supplied limit
-     */
-    public <K, V> CompletableFuture<Map<K, V>> getEntriesWithFilter(
```
shrids: +1

client/src/main/java/io/pravega/client/tables/KeyVersion.java
fpj: If we can call it `Version` or add a prefix to version that doesn't make it look like we are versioning the key rather than the value, I'm good. Internally, the version can remain implemented as is, so no work out of scope. Does it work?

andreipaduroiu: Ok. I renamed it to `Version`.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,258 @@
...
+     * and the versions will be in the same order as the entries. Notable exceptions:
+     * <ul>
+     * <li>{@link ConditionalTableUpdateException} If this is a Conditional Update and the condition was not satisfied.
+     * See the {@link KeyValueTable} doc for more details on Conditional Update Responses.
+     * </ul>
+     */
+    CompletableFuture<List<Version>> replaceAll(@NonNull String keyFamily, @NonNull Iterable<TableEntry<KeyT, ValueT>> entries);
+
+    /**
+     * Unconditionally removes a {@link TableKey} from this Table Segment. If the Key does not exist, no action will be
```
tkaitchuck: This links to tableKey but the signature does not take one.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,258 @@
...
+     * @param keyFamily (Optional) The Key Family for the Key to remove.
+     * @param key       The Key to remove.
+     * @return A CompletableFuture that, when completed, will indicate the Key has been removed.
+     */
+    CompletableFuture<Void> remove(@Nullable String keyFamily, @NonNull KeyT key);
+
+    /**
+     * Conditionally Removes a Key from this Table Segment.
+     *
+     * @param keyFamily (Optional) The Key Family for the {@link TableKey} to remove.
```
tkaitchuck: Same here

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,258 @@
...
+ * Conditional Update Responses:
+ * <ul>
+ * <li> Success: the update or removal has been atomically validated and performed; all updates or removals in the
+ * request have been accepted.
+ * <li> Failure: the update or removal has been rejected due to version mismatch; no update or removal has been performed.
+ * <li> {@link NoSuchKeyException}: the update or removal has been conditioned on a specific version (different from
+ * {@link Version#NOT_EXISTS} or {@link Version#NO_VERSION}) but the {@link TableKey} does not exist in the
+ * {@link KeyValueTable}.
+ * <li> {@link BadKeyVersionException}: the update or removal has been conditioned on a specific version (different from
+ * {@link Version#NO_VERSION} but the {@link TableKey} exists in the {@link KeyValueTable} with a different version.
```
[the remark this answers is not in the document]
andreipaduroiu: This doesn't mean "Wrong Version". It means that the user does not care about the server-side version of this update so their update will be unconditional.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,258 @@
...
+ * spreading it across many Table Segments, leading to eventual performance degradation.
+ * </ul>
+ * <p>
+ * Types of Updates:
+ * <ul>
+ * <li> Unconditional Updates will insert and/or overwrite any existing values for the given Key, regardless of whether
+ * that Key previously existed or not, and regardless of what that Key's version is.
+ * <li> Conditional Updates will only overwrite an existing value if the specified version matches that Key's version.
+ * If the key does not exist, the {@link TableKey} or {@link TableEntry} must have been created with
+ * {@link Version#NOT_EXISTS} in order for the update to succeed.
```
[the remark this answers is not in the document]
andreipaduroiu: I updated this entire section.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -0,0 +1,258 @@
...
+ * <li> Unconditional Removals will remove a Key regardless of what that Key's version is. The operation will also
+ * succeed (albeit with no effect) if the Key does not exist.
+ * <li> Conditional Removals will remove a Key only if the specified {@link TableKey#getVersion()} matches that Key's
+ * version. It will also fail (with no effect) if the Key does not exist and Version is not set to
+ * {@link Version#NOT_EXISTS}.
```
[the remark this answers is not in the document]
andreipaduroiu: I Explained it in the reworded section. Yes, it is a valid case, albeit very unusual.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
tkaitchuck: It appears to be used for both. I would like to differentiate the two

PR #4592, opened 2020-03-03
Issue 4590: (SegmentStore) Fixed Read Index concurrent Tier 2 read bug

### LINKED ISSUE #4590: DataCorruptionException in StorageWriter when fetching Read Index contents
**Problem description**
The Storage Writer occasionally trips over a problem where it cannot read data from the Read Index.

Symptoms:
1. Storage Writer has just flushed data to Tier 2 up to offset `B`. 
2. Read Index evicts range `[A, B)` (`A` < `B`)
3. Two concurrent Read Index read requests end up reinserting range `[A, B)` twice
    -  First one completes at time t1 and a new Cache Entry `[A, B)` is created.
4. An Append happens at offset `B` of length `L`. This gets appended to the cache entry at the previous step.
5. The second Read index read request completes with range `[A, B)`.
    - This ends up overwriting the entry created at step 3 and updated at step 4.
    - This is accompanied by a `WARN` in the logs that the second insert overrode the first one
4. The Storage Writer correctly complains that it cannot read offsets `[B, B+L)`.

Alternate situations:
- It is possible to run into this problem even if the concurrent Tier 2 reads do not begin at the same offset, but have overlapping ranges. PR #4596 addresses this as well.

In addition the `WARN` printed out in the logs has cache entries reversed (old is new, new is old). 

Specific examples:
(Note these have been edited to output the correct values for OldLength and NewLength.
```
2020-03-03 10:51:16,364 86007109 [core-21] WARN  i.p.s.s.r.StreamSegmentReadIndex - ReadIndex[10-7]: Insert overrode existing entry (Offset = 7747897, OldLength = 11088, NewLength = 10954).
2020-03-03 10:51:16,605 86007350 [core-25] ERROR i.p.s.server.writer.StorageWriter - StorageWriter[10]: Iteration[112516].CriticalError.
io.pravega.segmentstore.server.DataCorruptionException: Unable to retrieve CacheContents for 'AggregatedAppend: SegmentId = 7, Offsets = [7758851-7758985), SeqNo = 671834'.
```

and another one:
```
2020-03-03 11:15:51,819 87457206 [core-27] WARN  i.p.s.s.r.StreamSegmentReadIndex - ReadIndex[16-296681]: Insert overrode existing entry (Offset = 6153848, OldLength = 1301, NewLength = 680).
2020-03-03 11:15:53,278 87458665 [core-10] ERROR i.p.s.server.writer.StorageWriter - StorageWriter[16]: Iteration[55258].CriticalError.
io.pravega.segmentstore.server.DataCorruptionException: Unable to retrieve CacheContents for 'AggregatedAppend: SegmentId = 296681, Offsets = [6154528-6156006), SeqNo = 534738'.
```

**Problem location**
Segment Store. Either Read Index or Segment Aggregator.

Fix the problem.

andreipaduroiu:
**Change log description**  
Fixed a bug in StreamSegmentReadIndex where it would have been possible for two concurrent Storage read requests to override one another, resulting in potential (in-memory) data loss should there be an append executed in between the two.

**Purpose of the change**  
Fixes #4590.

**What the code does**  
When the Read Index receives concurrent reads for the same range in a Segment that is not cached, it may issue only one Tier 2 read and serve all such reads from it. However sometimes it may happen that each such read will trigger its own Tier 2 read, which will be accompanied by its own callback, when the data is inserted into the Cache. If such a scenario happens, the callback that is invoked later will override the data (and entry) inserted by the callback that is invoked first. Under normal circumstances this is not a big deal.

However, considering the scenario outlined in #4590, it is possible that the cache entry inserted by the first callback may have been modified (appended to) before the second one, and removing it from the cache will result in missing data that is not in Tier 2 nor in memory. This would cause both cache reads to fail and Tier 2 syncs to fail as well.

This PR fixes this problem by checking the size of the cache entry that is about to be overridden. If greater than the one that it is about to be replaced with, it will not perform the operation. Since cache entries can only grow (due to append being the only modification allowed), such a check is sufficient to prevent in-memory data loss.

**How to verify it**  
New unit tests added to verify this specific scenario.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CachePolicy.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CachePolicy.java
@@ -90,7 +90,7 @@ public class CachePolicy {
     public CachePolicy(long maxSize, double targetUtilization, double maxUtilization, Duration maxTime, Duration generationDuration) {
         Preconditions.checkArgument(maxSize > 0, "maxSize must be a positive integer");
         Preconditions.checkArgument(targetUtilization > 0 && targetUtilization <= 1.0,
-                "maxUtilization must be a number in the range (0.0, 1.0].");
+                "targetUtilization must be a number in the range (0.0, 1.0].");
         Preconditions.checkArgument(maxUtilization >= targetUtilization && maxUtilization <= 1.0,
                 "maxUtilization must be a number in the range (0.0, 1.0], at least equal to targetUtilization(%s).", targetUtilization);
         this.maxSize = maxSize;
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
@@ -563,11 +563,11 @@ class StreamSegmentReadIndex implements CacheManager.Client, AutoCloseable {
     private CacheIndexEntry addSingleEntryToCacheAndIndex(BufferView data, long offset, String operationName) {
         int dataAddress = this.cacheStorage.insert(data);
         CacheIndexEntry newEntry;
-        ReadIndexEntry oldEntry;
+        ReadIndexEntry rejectedEntry;
         try {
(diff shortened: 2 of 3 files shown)


RaulGracia:
@sachin-j-joshi please have a look to this PR.

PR #4593, opened 2020-03-03
Issue 4591: Fixed ExtendedS3 exists()

### LINKED ISSUE #4591: ExtendedS3Storage: incorrect implementation of exists
**Problem description**
Currently the ExtendedS3Storage implements exists() by calling listObjects with given path and checking whether list returned is empty or not.
For the path that corresponds with full path to the object this is correct. However this breaks when the path checked is a prefix. In this case list of other object matching that prefix will be returned and the code will incorrectly determine that object exists. 

**Problem location**
ExtendedS3Storage . doExists

Instead of listObjects use getObjectMetadata.

sachin-j-joshi:
**Change log description**  
This change uses getObjectMetadata instead of listObjects to implement exists.
This fixes the current problem where listObject based check behaves incorrectly when prefix instead of full path is provided as parameter.

**Purpose of the change**  
Fixes #4591 
**What the code does**  
This change uses getObjectMetadata instead of listObjects to implement exists.

**How to verify it**  
Unit tests, System Tests and longevity tests should pass and should not show the incorrect behavior anymore.

### DIFF
--- a/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
+++ b/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
@@ -17,7 +17,6 @@ import com.emc.object.s3.bean.AccessControlList;
 import com.emc.object.s3.bean.CanonicalUser;
 import com.emc.object.s3.bean.CopyPartResult;
 import com.emc.object.s3.bean.Grant;
-import com.emc.object.s3.bean.ListObjectsResult;
 import com.emc.object.s3.bean.MultipartPartETag;
 import com.emc.object.s3.bean.Permission;
 import com.emc.object.s3.request.CompleteMultipartUploadRequest;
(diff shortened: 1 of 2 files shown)


bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -263,18 +262,10 @@ private StreamSegmentInformation doGetStreamSegmentInfo(String streamSegmentName
 
     private boolean doExists(String streamSegmentName) {
         try {
-            ListObjectsResult result = client.listObjects(config.getBucket(), config.getPrefix() + streamSegmentName);
-            return !result.getObjects().isEmpty();
+            S3ObjectMetadata result = client.getObjectMetadata(config.getBucket(),
+                    config.getPrefix() + streamSegmentName);
```
medvedevigorek: _streamSegmentName_ is actually a stream segment name but do we actually create objects with such names? I thought   for each segment there will be objects like _streamSegmentName$header_ or _streamSegmentName$attributes_. So if we simple check for presence of an object with _streamSegmentName_ we always get object not found. Am i missing something here?

sachin-j-joshi: The layer above this component (RollingStorage) passes the names like _streamSegmentName$header_ or _streamSegmentName$attributes_ to this component.

sachin-j-joshi:
@andreipaduroiu the system tests and short longevity tests passed. (We don't see this issue. Fix worked.)

PR #4596, opened 2020-03-06
Issue 4590: (SegmentStore) Fixed Read Index concurrent Tier 2 read bug (Part 2)

### LINKED ISSUE #4590: DataCorruptionException in StorageWriter when fetching Read Index contents

andreipaduroiu:
**Change log description**  
Fixed a (second) bug in StreamSegmentReadIndex where it would have been possible for two concurrent Storage read requests to override one another, resulting in potential (in-memory) data loss should there be an append executed in between the two.

**Purpose of the change**  
Fixes #4590.

**What the code does**  
See #4592 for details on the first bug.
#4592 handled the case when the concurrent reads began at  exactly the same Segment offset. This PR covers an additional case when two or more concurrent Tier 2 reads (resulting from cache misses) do not necessarily start at the same offset, but have at least one overlapping byte.

This PR aims to completely eliminate duplicated data in the Read Index due to such situations. Having Read Index entries that overlap can cause inconsistent read behaviors depending on what offset your read originated; this can also cause recently appended data to vanish from the read index (see parent issue for how this can happen).

**How to verify it**  
New unit tests added to verify this specific scenario.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
@@ -617,18 +617,18 @@ class StreamSegmentReadIndex implements CacheManager.Client, AutoCloseable {
                     // Slice the data that we need to insert. We may be able to insert the whole buffer at once.
                     BufferView dataToInsert = overlapLength >= data.getLength() ? data : data.slice(0, (int) overlapLength);
                     CacheIndexEntry newEntry;
-                    int dataAddress = this.cacheStorage.insert(dataToInsert);
+                    int dataAddress = CacheStorage.NO_ADDRESS; // Null address pointer.
                     try {
+                        dataAddress = this.cacheStorage.insert(dataToInsert);
                         newEntry = new CacheIndexEntry(segmentOffset, dataToInsert.getLength(), dataAddress);
                         ReadIndexEntry overriddenEntry = addToIndex(newEntry);
                         assert overriddenEntry == null : "Insert overrode existing entry; " + segmentOffset + ":" + dataToInsert.getLength();
                         lastInsertedEntry = newEntry;
                     } catch (Throwable ex) {
-                        // Clean up the data we inserted if we were unable to add it to the index.
+                        // Clean up the data we might have inserted if we were unable to add it to the index.
                         this.cacheStorage.delete(dataAddress);
                         throw ex;
                     }
-
                 }
 
                 // Slice the remainder of the buffer, or set it to null if we processed everything.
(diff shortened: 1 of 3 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
```diff
@@ -577,41 +586,67 @@ private CacheIndexEntry addSingleEntryToCacheAndIndex(BufferView data, long offs
...
+                    CacheIndexEntry newEntry;
+                    int dataAddress = this.cacheStorage.insert(dataToInsert);
+                    try {
+                        newEntry = new CacheIndexEntry(segmentOffset, dataToInsert.getLength(), dataAddress);
+                        ReadIndexEntry overriddenEntry = addToIndex(newEntry);
+                        assert overriddenEntry == null : "Insert overrode existing entry; " + segmentOffset + ":" + dataToInsert.getLength();
+                        lastInsertedEntry = newEntry;
+                    } catch (Throwable ex) {
+                        // Clean up the data we inserted if we were unable to add it to the index.
+                        this.cacheStorage.delete(dataAddress);
```
tkaitchuck: Is this line reachable? (in practice)

andreipaduroiu: No. Not really. If we end up here then some exception got thrown from `addToIndex` and we must make sure we delete the data from the cache otherwise it will stay there forever. The only exception thrown from there is an `ObjectClosedException`, which indicates the whole index is about to close (or has just been) so we must clean up anyway.

PR #4598, opened 2020-03-09
Issue 4597: Optimization of String splits for Metrics

### LINKED ISSUE #4597: Optimize String Split for Segment Name
**Problem description**
In multiple places of `NameUtils`, we use a reggex input to split a string when a single character would suffice:
```
String[] tokens = originalSegmentName.split("[/]");
```
The problem is that this is being executed in the hot data path and it is adding significant and unnecessary overhead. The `String.split()` method is optimized when we split with a single character:
```
    public String[] split(String regex, int limit) {
        /* fastpath if the regex is a
         (1)one-char String and this character is not one of the
            RegEx's meta characters ".$|()[{^?*+\\", or
         (2)two-char String and the first char is the backslash and
            the second is not the ascii digit or ascii letter.
         */
```
We should make use of the first option for efficiency reasons.

**Problem location**
`NameUtils`

Use single character to split String whenever possible.

RaulGracia:
**Change log description**  
Uses single-character input in String.split() for efficiency reasons.

**Purpose of the change**  
Fixes #4597.

**What the code does**  
This PR attempts to use a variant of the `String.split()` method implementation, which is optimized for single-character split patterns. There are several occurrences in `NameUtils` class that can exploit this approach, but by some reason a reggex expression is used. This PR just replaces these occurrences, as some of them are on the hot write path.

**How to verify it**  
We have executed a micro benchmark to compare both approaches. Improvement in performance of `String.split()` is consistently between 2x and 4x:
```
    public static void main(String[] args) {
        String toSplit = "this/is/a/segment";
        long iterations = 1000000;
        long iniTime = System.nanoTime();
        for (long i = 0; i < iterations; i++) {
            toSplit.split("[/]");
        }
        double timeReggex = System.nanoTime() - iniTime;
        System.out.println("Time Reggex: " + timeReggex/1000000.0 + "ms");

        iniTime = System.nanoTime();
        for (long i = 0; i < iterations; i++) {
            toSplit.split("/");
        }
        double timeNoReggex = System.nanoTime() - iniTime;
        System.out.println("Time No Reggex: " + timeNoReggex/1000000.0 + "ms");
        System.out.println("Improvement: " + timeReggex/timeNoReggex);
    }
```
These are results from multiple executions:
```
Time Reggex: 747.486985ms
Time No Reggex: 208.371695ms
Improvement: 3.587276981165796
```
```
Time Reggex: 775.19952ms
Time No Reggex: 272.466208ms
Improvement: 2.845121696705964
```
```
Time Reggex: 795.027535ms
Time No Reggex: 399.01521ms
Improvement: 1.9924742593145761
```
Even with larger number of iterations (100M), the improvement is still consistent:
```
Time Reggex: 47721.324557ms
Time No Reggex: 17042.942773ms
Improvement: 2.80006365054524
```

### DIFF
--- a/client/src/main/java/io/pravega/client/netty/impl/ClientConnectionImpl.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/ClientConnectionImpl.java
@@ -19,6 +19,7 @@ import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.PromiseCombiner;
 import io.pravega.common.Exceptions;
 import io.pravega.common.Timer;
+import io.pravega.shared.metrics.MetricNotifier;
 import io.pravega.shared.protocol.netty.Append;
 import io.pravega.shared.protocol.netty.AppendBatchSizeTracker;
 import io.pravega.shared.protocol.netty.ConnectionFailedException;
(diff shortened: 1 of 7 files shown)


RaulGracia:
@kevinhan88 metrics are being published, I have seen them in the dashboard.

PR #4599, opened 2020-03-09
Issue 4600: Fix Transaction related logs.

### LINKED ISSUE #4600: Improve logging on client and controller
**Problem description**
Improve the following aspects.
1. Incase of no response from Controller a deadline exceeded exception is thrown. Ensure we log the request details to ensure easier debugging in such a scenario. 
2. A SSS can return an empty response as part of handling `io.pravega.segmentstore.server.host.handler.PravegaRequestProcessor.ReadCancellationException`
3. Ensure the controller logs the txn ids in UUID format instead of lowbits and high bits.

**Problem location**
Logging on client and controller

Improve these issues.

shrids:
**Change log description**  
- Simplify debugging in case of deadline exceed exception on the controller client.
- Reduce logging on the client.
- Improve transaction related logs on the controller.
- Ensure an empty response from SSS should not log a WARN message.

**Purpose of the change**  
Fixes #4600 

**What the code does**  

- Simplify debugging in case of exceptions on the controller client. Ensure the parameters like the transaction ids are logged immediately instead of logging it along with a `RetriesExhaustedException`.
- Reduce logging on the client in case of an empty response from SSS.
- Ensure the txn UUID is logged instead of low and high bits of the transaction.

**How to verify it**  
No changes to functionality. All the existing tests should continue to pass.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
+++ b/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
@@ -9,7 +9,6 @@
  */
 package io.pravega.controller.server.rpc.grpc.v1;
 
-import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 import io.grpc.Status;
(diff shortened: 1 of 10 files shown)


controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -354,56 +356,60 @@ public void createTransaction(CreateTxnRequest request, StreamObserver<Controlle
 
     @Override
     public void commitTransaction(TxnRequest request, StreamObserver<TxnStatus> responseObserver) {
+        final UUID txnId = ModelHelper.encode(Preconditions.checkNotNull(request.getTxnId(), "txnId"));
```
shiveshr: nit: can you move preconditions inside the encode method.

shrids: `ModelHelper.encode(...)` already has this precondition, hence removed this redundant check.

PR #4602, opened 2020-03-09
Issue 4601: (RollingStorage/ExtendedS3) S3 client consumes whole InputStream and causes java.lang.IllegalArgumentException

### LINKED ISSUE #4601: (RollingStorage/ExtendedS3) S3 client consumes whole InputStream and causes java.lang.IllegalArgumentException
**Problem description**

- For Extended S3 write operation we pass in InputStream and offset+length. With rolling storage it is possible to have more data in InputStream than length specified. 

- Assumption is that only bytes upto length would be sent to the server, However the S3 client writes out the entire input stream content. Note that the range parameter is validated on server side and doesn't mean client sends only 'length' bytes. That length basically needs to EXACTLY match content length.

- This results in java.lang.IllegalArgumentException: 

`2020-03-07 16:23:45,938 33445218 [storage-io-152] ERROR i.p.s.server.writer.StorageWriter - StorageWriter[11]: Iteration[40878].Error.
java.lang.IllegalArgumentException: _system/_tables/completedTransactionsBatch-0$attributes.index$offset.0
	at io.pravega.storage.extendeds3.ExtendedS3Storage.throwException(ExtendedS3Storage.java:482)
	at io.pravega.storage.extendeds3.ExtendedS3Storage.execute(ExtendedS3Storage.java:511)
	at io.pravega.storage.extendeds3.ExtendedS3Storage.write(ExtendedS3Storage.java:148)
	at io.pravega.segmentstore.storage.rolling.RollingStorage.write(RollingStorage.java:324)
	at io.pravega.segmentstore.storage.AsyncStorageWrapper.lambda$write$2(AsyncStorageWrapper.java:93)
	at `

**Problem location**
RollingStorage/ExtendedS3

- RollingStorage.write should send in BoundedInputStream to storage adapters to prevent reading more data.
- S3 client mocks should simulate this range behavior to match how real ECS bahaves.

sachin-j-joshi:
**Change log description**  
S3 client consumes whole InputStream and causes java.lang.IllegalArgumentException.
With this change, RollingStorage.write sends in BoundedInputStream to storage adapters to prevent reading more data.
Also, S3 client mocks simulate this InvalidRange behavior to match how real ECS bahaves.

**Purpose of the change**  
Fixes #4601 

**What the code does**  
RollingStorage.write sends in BoundedInputStream to storage adapters to prevent reading more data.
S3 client mocks simulate this range behavior to match how real ECS bahaves.

**How to verify it**  

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
@@ -12,6 +12,7 @@ package io.pravega.segmentstore.storage.rolling;
 import com.google.common.base.Preconditions;
 import io.pravega.common.Exceptions;
 import io.pravega.common.LoggerHelpers;
+import io.pravega.common.io.BoundedInputStream;
 import io.pravega.common.util.ByteArraySegment;
 import io.pravega.common.util.CollectionHelpers;
 import io.pravega.segmentstore.contracts.BadOffsetException;
@@ -29,6 +30,7 @@ import io.pravega.segmentstore.storage.StorageNotPrimaryException;
 import io.pravega.segmentstore.storage.SyncStorage;
 import io.pravega.shared.NameUtils;
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
@@ -301,6 +303,7 @@ public class RollingStorage implements SyncStorage {
     }
 
     @Override
+    @SneakyThrows(IOException.class)
     public void write(SegmentHandle handle, long offset, InputStream data, int length) throws StreamSegmentException {
         val h = getHandle(handle);
         ensureNotDeleted(h);
@@ -321,7 +324,13 @@ public class RollingStorage implements SyncStorage {
             int writeLength = (int) Math.min(length - bytesWritten, h.getRollingPolicy().getMaxLength() - last.getLength());
             assert writeLength > 0 : "non-positive write length";
             long chunkOffset = offset + bytesWritten - last.getStartOffset();
-            this.baseStorage.write(h.getActiveChunkHandle(), chunkOffset, data, writeLength);
+
+            // Use a BoundedInputStream to ensure that the underlying storage does not try to read more (or less) data
+            // than we instructed it to. Invoking BoundedInputStream.close() will throw an IOException if baseStorage.write()
+            // has not read all the bytes it was supposed to.
+            try (BoundedInputStream bis = new BoundedInputStream(data, writeLength)) {
+                this.baseStorage.write(h.getActiveChunkHandle(), chunkOffset, bis, writeLength);
+            }
             last.increaseLength(writeLength);
             bytesWritten += writeLength;
         }
--- a/segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
+++ b/segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
@@ -406,8 +406,7 @@ public abstract class StorageTestBase extends ThreadPooledTestSuite {
             for (int j = 0; j < APPENDS_PER_SEGMENT; j++) {
                 byte[] writeData = String.format(APPEND_FORMAT, segmentName, j).getBytes();
 
-                // Append some garbage at the end to make sure we only write as much as instructed, and not the whole InputStream.
-                val dataStream = new SequenceInputStream(new ByteArrayInputStream(writeData), new ByteArrayInputStream(extraData));
+                val dataStream = new ByteArrayInputStream(writeData);
                 s.write(writeHandle, offset, dataStream, writeData.length, TIMEOUT).join();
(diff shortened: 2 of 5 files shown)


segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
```diff
@@ -406,8 +406,7 @@ private String getSegmentName(int id, String context) {
             for (int j = 0; j < APPENDS_PER_SEGMENT; j++) {
                 byte[] writeData = String.format(APPEND_FORMAT, segmentName, j).getBytes();
 
-                // Append some garbage at the end to make sure we only write as much as instructed, and not the whole InputStream.
-                val dataStream = new SequenceInputStream(new ByteArrayInputStream(writeData), new ByteArrayInputStream(extraData));
+                val dataStream = new ByteArrayInputStream(writeData);
```
andreipaduroiu: Why remove this? This is exactly what we were trying to test for...

sachin-j-joshi: Now that we match mocks to match server behavior the test breaks.

andreipaduroiu: Given that these classes will get rewritten anyway with PDP-34 and that even right now, nobody accesses them directly except via RollingStorage, I think we should be fine with this change.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -321,7 +324,13 @@ public void write(SegmentHandle handle, long offset, InputStream data, int lengt
             int writeLength = (int) Math.min(length - bytesWritten, h.getRollingPolicy().getMaxLength() - last.getLength());
             assert writeLength > 0 : "non-positive write length";
             long chunkOffset = offset + bytesWritten - last.getStartOffset();
-            this.baseStorage.write(h.getActiveChunkHandle(), chunkOffset, data, writeLength);
+
+            // Use a BoundedInputStream to ensure that the underlying storage does not try to read more (or less) data
+            // than we instructed it to. Invoking BoundedInputStream.close() will throw an IOException if baseStorage.write()
+            // has not read all the bytes it was supposed to.
```
kevinhan88: Here we have BoundedInputStream and the Length of Inputstream (or length to be written) - both are used to control the boundary? Not very clear about the relationship between the two - who's taking priority if the two values conflicting?

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/rolling/RollingStorageTestBase.java
```diff
@@ -123,6 +124,37 @@ public void testWriteAfterHeaderMerge() throws Exception {
...
+        val writeHandle = s.create(segmentName, new SegmentRollingPolicy(maxLength), TIMEOUT)
+                .thenCompose(v -> s.openWrite(segmentName)).join();
+
+        val byteInputStream1 = new ByteArrayInputStream(seq1.getBytes());
+        val byteInputStream2 = new ByteArrayInputStream(seq2.getBytes());
+
+        val sequenceInputStream = new SequenceInputStream(byteInputStream1, byteInputStream2);
+
+        // This write should cause 3 rollovers.
+        s.write(writeHandle, 0, sequenceInputStream, totalWriteLength, TIMEOUT).join();
```
kevinhan88: Can the 3 rollovers be reflected in the test code?

sachin-j-joshi: We write 10 bytes, with max rolling length set to 3.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
sachin-j-joshi: Actually the requirement is that we want them to be same. That's why we pass same writeLength on line 331 and 332.

andreipaduroiu: InputStreams do not have lengths. They have `available()` but that doesn't tell you much.

You need to pass the length to the API to tell it how much to write.

You need to (now) use a BoundedInputStream with a specific bound because some of those Tier 2 APIs do not respect that value. We want to make sure they do not under- or over-read.

kevinhan88: Then ideally create another method called "writeWithBoundedInputStream" to encapsulate this logic; or at least add a runtime assert here to call out the logic.
A method calling with overlapping input data is error-prone.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/rolling/RollingStorageTestBase.java
sachin-j-joshi: Added additional checks.

PR #4611, opened 2020-03-10
Issue 4610: (SegmentStore) Read Index data corruption during concurrent segment merge and cache eviction.

### LINKED ISSUE #4610: Read Index corruption possible during concurrent cache eviction and segment merger
**Problem description**
1. We have Segments A, B
2. We receive a request to merge B into A.
3. We execute ReadIndex.BeginMerge B->A
4. After executing the merge in Tier 2 (B ceases to exist), we execute ReadIndex.CompleteMerge B->A (which has the effect of transferring the read index entries from B to A)
5. CacheManager evicts all of B (because it doesn't exist) while the previous step is still executing.

At this point:
- Any append to A will fail because it will try to modify a deallocated Cache Block
- Any read from A for the range where B was merged in will either fail saying that the Cache Blocks are not allocated or will return the wrong data if those blocks have already been reused.

**Problem location**
StreamSegmentReadIndex.

Immediately after transferring the read index entries, and while holding the source segment's cache lock, clear its index (but do not evict anything). This should prevent the Cache Manager from overreaching into it.

andreipaduroiu:
**Change log description**  
Fixed a bug where a concurrent segment merge and cache eviction could corrupt the Read Index.

**Purpose of the change**  
Fixes #4610.

**What the code does**  
In the Read Index, Segments are merged into each other in a two-step process. Step 1 (BeginMerge) is performed when the request is received externally and involves adding a Redirect placeholder in the index for the merged byte range. Step 2 (CompleteMerge) is performed when the merge is performed in Tier 2 and involves atomically removing the Redirect placeholder and copying over the entire Read Index of the source segment.

With the current implementation, Step 2 has a short period of time where, if a Cache Manager eviction executes, it can effectively delete the data from the cache. That is because the source index entries were copied over but they still remained in the source index. If the Cache Manager decided it's time to evict data, those old entries were eligible for eviction since the segment has just been merged. If this does go through, the newly copied entries will now point to invalid Cache addresses, which can have a number of bad side effects:
- A subsequent append will fail because it tries to append to a non-allocated block
- A subsequent read will fail because there's no data to read from the cache.
- The cache blocks could have been reallocated to other entries, so:
    - A subsequent append will fail because the append condition (last block length) doesn't match what's in the cache.
    - A subsequent read will read data belonging to completely unrelated segments and/or events
    - A subsequent cache eviction could deallocate blocks that are used by other entries.    

To resolve this, we need to change the "copy" to a "move". We atomically replace the Redirect Index entry with copies of the source entries and remove the source entries from the source index. This way, they will be out of reach of the Cache Manager. Since the source segment ceased to exist (it was merged), this operation should have no side effects.

**How to verify it**  
Unit tests added to verify scenario.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
@@ -482,7 +482,7 @@ class StreamSegmentReadIndex implements CacheManager.Client, AutoCloseable {
         StreamSegmentReadIndex sourceIndex = redirectEntry.getRedirectReadIndex();
 
         // Get all the entries from the source index and append them here.
-        List<MergedIndexEntry> sourceEntries = sourceIndex.removeAllEntries(redirectEntry.getStreamSegmentOffset());
+        List<MergedIndexEntry> sourceEntries = sourceIndex.removeAllDataEntries(redirectEntry.getStreamSegmentOffset());
 
         synchronized (this.lock) {
             // Remove redirect entry (again, no need to update the Cache Stats, as this is a RedirectIndexEntry).
@@ -1200,7 +1200,7 @@ class StreamSegmentReadIndex implements CacheManager.Client, AutoCloseable {
      * live data in the {@link CacheManager}; as such it is the responsibility of the caller to manage their lifecycle
      * from now on.
      */
-    private List<MergedIndexEntry> removeAllEntries(long offsetAdjustment) {
+    private List<MergedIndexEntry> removeAllDataEntries(long offsetAdjustment) {
         Preconditions.checkState(this.metadata.isDeleted(), "Cannot fetch entries for a Segment that has not been deleted yet.");
         Exceptions.checkArgument(offsetAdjustment >= 0, "offsetAdjustment", "offsetAdjustment must be a non-negative number.");
 
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
@@ -482,7 +482,7 @@ class StreamSegmentReadIndex implements CacheManager.Client, AutoCloseable {
         StreamSegmentReadIndex sourceIndex = redirectEntry.getRedirectReadIndex();
 
         // Get all the entries from the source index and append them here.
-        List<MergedIndexEntry> sourceEntries = sourceIndex.fetchAllEntries(redirectEntry.getStreamSegmentOffset());
+        List<MergedIndexEntry> sourceEntries = sourceIndex.removeAllEntries(redirectEntry.getStreamSegmentOffset());
 
         synchronized (this.lock) {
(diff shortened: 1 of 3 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
```diff
@@ -1192,24 +1192,33 @@ private ReadResultEntryBase createFutureRead(long streamSegmentOffset, int maxLe
     }
 
     /**
-     * Gets a copy of all the ReadIndexEntries in this Index that are not RedirectReadIndices. All returned
-     * entries have their offsets adjusted by the given amount.
+     * Gets a copy of all the ReadIndexEntries in this Index that are not RedirectReadIndices and removes them from this
```
tkaitchuck: It says "gets a copy" but then later says they are removed, so, it's not really a copy.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
```diff
@@ -1192,24 +1192,33 @@ private ReadResultEntryBase createFutureRead(long streamSegmentOffset, int maxLe
...
+     * index. All returned entries have their offsets adjusted by the given amount.
      *
      * @param offsetAdjustment The amount to adjust the offset by.
+     * @return A List of {@link MergedIndexEntry} that represents the contents of this index. All these entries point to
+     * live data in the {@link CacheManager}; as such it is the responsibility of the caller to manage their lifecycle
+     * from now on.
      */
-    private List<MergedIndexEntry> getAllEntries(long offsetAdjustment) {
+    private List<MergedIndexEntry> fetchAllEntries(long offsetAdjustment) {
```
tkaitchuck: This should probably be called "removeAllEntries"

segmentstore/server/src/test/java/io/pravega/segmentstore/server/reading/ContainerReadIndexTests.java
```diff
@@ -91,7 +91,7 @@
             .with(ReadIndexConfig.MEMORY_READ_MIN_LENGTH, 0) // Default: Off (we have a special test for this).
             .with(ReadIndexConfig.STORAGE_READ_ALIGNMENT, 1024)
             .build();
-    private static final Duration TIMEOUT = Duration.ofSeconds(20);
+    private static final Duration TIMEOUT = Duration.ofSeconds(2000000);
```
RaulGracia: Is this change intentional?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StreamSegmentReadIndex.java
andreipaduroiu: Reworded.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/reading/ContainerReadIndexTests.java
andreipaduroiu: No it is not. Fixed it.

PR #4612, opened 2020-03-11
Issue 4603: (KeyValueTables) Client Controller API

### LINKED ISSUE #4603: KeyValueTables: Client Controller API
**Problem description**
1. Move `Controller` interface and impl out of `stream` package.
2. Add KeyValueTable APIs and implement them as no-ops (or throw UnsupportedOpEx).

**Problem location**
Client.

See description.

andreipaduroiu:
**Change log description**  
- Moved `io.pravega.client.stream.impl.Controller` (and related classes) into `io.pravega.client.control.impl`
- Added new stub contracts to `Controller` interface for Key-Value Tables. These methods are not implemented in this PR.

**Purpose of the change**  
Fixes #4603.

**What the code does**  
Refactoring (moved `Controller` into new package), which triggered changes to a large number of files. All these changes should be simply to import a new package.

Added new API stubs.

**How to verify it**  
Build.


client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
```diff
@@ -40,7 +40,7 @@
      *
      * @return the read offset for each segment in the ownedSegments set
      */
-    abstract Map<Segment, Long> getOwnedSegmentsWithOffsets();
+    public abstract Map<Segment, Long> getOwnedSegmentsWithOffsets();
```
tkaitchuck: This was something we very deliberately avoided in the way we layed out the directory structure. This object is returned to users, so even though it is inside of the impl directory there is a temptation to call methods on it. Let's try to work out a way to avoid making this publicly callable. Perhaps by having the map passed to the caller rather than the position object?

client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
```diff
@@ -52,7 +52,7 @@
     /**
      * Gets the set of completely read segments.
      */
-    abstract Set<Segment> getCompletedSegments();
+    public abstract Set<Segment> getCompletedSegments();
```
tkaitchuck: Same here

client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
andreipaduroiu: Looks like this is only invoked from `ModelHelper.getSegmentsFromPositions` which is actually not used anywhere. Same with the one below (invoked in `ModelHelper.toSegmentOffsetMap` which is not used anywhere.

I deleted those methods in `ModelHelper` and reverted the changes in `PositionInternal.java`.

client/src/main/java/io/pravega/client/stream/impl/PositionInternal.java
andreipaduroiu: See reply above.

PR #4615, opened 2020-03-11
Issue 4609 concat using appends for small source

### LINKED ISSUE #4609: (ExtendedS3Storage): For smaller source segments, concat should read and append instead of using multi-part upload
**Problem description**
Currently ExtendedS3Storage implements all concat operations as multi-part uploads. ( target = target+ source ). This is inefficient when source segments are small.

**Problem location**
ExtendedS3Storage

For smaller source segments, concat should **read** complete source segment and **append** it to target instead of using multi-part upload.
For larger source segments continue using current method.

sachin-j-joshi:
**Change log description**  
Currently ExtendedS3Storage implements all concat operations as multi-part uploads. ( target = target+ source ). This is inefficient when source segments are small.
With this change , for smaller source segments, concat should read complete source segment and append it to target instead of using multi-part upload.
For larger source segments continue using current method.

**Purpose of the change**  
Fixes #4609 

**What the code does**  
For smaller source segments, concat should read complete source segment and append it to target instead of using multi-part upload.
For larger source segments continue using current method.

**How to verify it**  
All tests should be green.

### DIFF
--- a/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
+++ b/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
@@ -39,6 +39,8 @@ import io.pravega.segmentstore.contracts.StreamSegmentNotExistsException;
 import io.pravega.segmentstore.contracts.StreamSegmentSealedException;
 import io.pravega.segmentstore.storage.SegmentHandle;
 import io.pravega.segmentstore.storage.SyncStorage;
+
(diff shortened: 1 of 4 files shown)


medvedevigorek:
@sachin-j-joshi shouldn't it be merged to _master_ branch first and then cherry-picked to _r0.7_ ?

sachin-j-joshi:
https://github.com/pravega/pravega/pull/4617 Against master.
I'll create another one for cherry pick

PR #4616, opened 2020-03-11
Issue 3865 unify test utils assert extensions

### LINKED ISSUE #3865: Unify TestUtils.await and AssertExtensions.assertEventuallyEquals
**Problem description**
This is one of the outcome of troubleshooting of https://github.com/pravega/pravega/issues/3785 as we found the code duplication between the two methods.
Both methods executes the supplied function with the specified interval until it times out at the input timeout. 
However, await() needs a function to supply boolean, but assert doesn't return value, so it is not trivial to unify the two.

**Problem location**
TestUtils.await() 
AssertExtensions.assertEventuallyEquals()

Find a way to unify the two in order to reduce the code duplication and improve the usage of the future unified method.

co-jo:
**Change log description**  
* Reuse the  `TestUtils.await()` function to remove duplicate code.

**Purpose of the change**  
Closes #3865 

**What the code does**  
No functional changes.

**How to verify it**  
Run unit & system test. Completed.


test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
```diff
@@ -56,15 +56,13 @@
...
-        while (currentTime < endTime) {
-            if ((expected == null && eval.call() == null) || (expected != null && expected.equals(eval.call()))) {
-                return;
-            }
-            Thread.sleep(checkIntervalMillis);
-            currentTime = System.currentTimeMillis();
-        }
+        TestUtils.await(() -> {
+                try {
+                    return expected.equals(eval.call());
```
andreipaduroiu: You need to handle the case when `expected` is null as well. Check the condition above.

PR #4617, opened 2020-03-11
Issue 4609: ExtendedS3Storage.concat using appends when source segment is small

### LINKED ISSUE #4609: (ExtendedS3Storage): For smaller source segments, concat should read and append instead of using multi-part upload

sachin-j-joshi:
**Change log description**  
Currently ExtendedS3Storage implements all concat operations as multi-part uploads. ( target = target+ source ). This is inefficient when source segments are small.
With this change , for smaller source segments, concat should read complete source segment and append it to target instead of using multi-part upload.
For larger source segments continue using current method.

**Purpose of the change**  
Fixes #4609 

**What the code does**  
For smaller source segments, concat should read complete source segment and append it to target instead of using multi-part upload.
For larger source segments continue using current method.

**How to verify it**  
All tests should be green.

### DIFF
--- a/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
+++ b/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
@@ -39,6 +39,8 @@ import io.pravega.segmentstore.contracts.StreamSegmentNotExistsException;
 import io.pravega.segmentstore.contracts.StreamSegmentSealedException;
 import io.pravega.segmentstore.storage.SegmentHandle;
 import io.pravega.segmentstore.storage.SyncStorage;
+
+import java.io.BufferedInputStream;
 import java.io.InputStream;
 import java.time.Duration;
 import java.util.SortedSet;
@@ -403,39 +405,53 @@ public class ExtendedS3Storage implements SyncStorage {
         Preconditions.checkState(si.isSealed(), "Cannot concat segment '%s' into '%s' because it is not sealed.",
                 sourceSegment, targetHandle.getSegmentName());
 
-        //Copy the first part
-        CopyPartRequest copyRequest = new CopyPartRequest(config.getBucket(),
-                targetPath,
-                config.getBucket(),
-                targetPath,
-                uploadId,
-                1).withSourceRange(Range.fromOffsetLength(0, offset));
-        CopyPartResult copyResult = client.copyPart(copyRequest);
+        if (config.getSmallObjectThreshold() < si.getLength()) {
 
-        partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
+            //Copy the first part
+            CopyPartRequest copyRequest = new CopyPartRequest(config.getBucket(),
+                    targetPath,
+                    config.getBucket(),
+                    targetPath,
+                    uploadId,
+                    1).withSourceRange(Range.fromOffsetLength(0, offset));
+            CopyPartResult copyResult = client.copyPart(copyRequest);
 
-        //Copy the second part
-        S3ObjectMetadata metadataResult = client.getObjectMetadata(config.getBucket(),
-                config.getPrefix() + sourceSegment);
-        long objectSize = metadataResult.getContentLength(); // in bytes
+            partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
 
-        copyRequest = new CopyPartRequest(config.getBucket(),
-                config.getPrefix() + sourceSegment,
-                config.getBucket(),
-                targetPath,
-                uploadId,
-                2).withSourceRange(Range.fromOffsetLength(0, objectSize));
+            //Copy the second part
+            S3ObjectMetadata metadataResult = client.getObjectMetadata(config.getBucket(),
+                    config.getPrefix() + sourceSegment);
+            long objectSize = metadataResult.getContentLength(); // in bytes
 
-        copyResult = client.copyPart(copyRequest);
-        partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
+            copyRequest = new CopyPartRequest(config.getBucket(),
+                    config.getPrefix() + sourceSegment,
+                    config.getBucket(),
+                    targetPath,
+                    uploadId,
+                    2).withSourceRange(Range.fromOffsetLength(0, objectSize));
 
-        //Close the upload
-        client.completeMultipartUpload(new CompleteMultipartUploadRequest(config.getBucket(),
-                targetPath, uploadId).withParts(partEtags));
+            copyResult = client.copyPart(copyRequest);
+            partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
 
-        client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
+            //Close the upload
+            client.completeMultipartUpload(new CompleteMultipartUploadRequest(config.getBucket(),
+                    targetPath, uploadId).withParts(partEtags));
+
+            client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
+
+        } else {
+            try (InputStream reader = client.readObjectStream(config.getBucket(),
+                    config.getPrefix() + sourceSegment, Range.fromOffsetLength(0, si.getLength()))) {
+                client.putObject(this.config.getBucket(),
+                        targetPath,
+                        Range.fromOffsetLength(offset, si.getLength()),
+                        new BufferedInputStream(reader, Math.toIntExact(si.getLength())));
+            } catch (Exception e) {
+                throw Exceptions.sneakyThrow(e);
+            }
+            client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
+        }
         Duration elapsed = timer.getElapsed();
-
         log.debug("Concat target={} source={} offset={} bytesWritten={} latency={}.", targetHandle.getSegmentName(), sourceSegment, offset, si.getLength(), elapsed.toMillis());
 
         ExtendedS3Metrics.CONCAT_LATENCY.reportSuccessEvent(elapsed);
--- a/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
+++ b/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
@@ -388,69 +388,31 @@ public class ExtendedS3Storage implements SyncStorage {
      * completeMultiPartUpload call. Specifically, to concatenate, we are copying the target segment T and the
      * source segment S to T, so essentially we are doing T <- T + S.
      */
-    private Void doConcat(SegmentHandle targetHandle, long offset, String sourceSegment) throws StreamSegmentNotExistsException {
+    private Void doConcat(SegmentHandle targetHandle, long offset, String sourceSegment) throws Exception {
         Preconditions.checkArgument(!targetHandle.isReadOnly(), "target handle must not be read-only.");
         long traceId = LoggerHelpers.traceEnter(log, "concat", targetHandle.getSegmentName(), offset, sourceSegment);
         Timer timer = new Timer();
-        SortedSet<MultipartPartETag> partEtags = new TreeSet<>();
-        String targetPath = config.getPrefix() + targetHandle.getSegmentName();
-        String uploadId = client.initiateMultipartUpload(config.getBucket(), targetPath);
 
+        String targetPath = config.getPrefix() + targetHandle.getSegmentName();
         // check whether the target exists
         if (!doExists(targetHandle.getSegmentName())) {
             throw new StreamSegmentNotExistsException(targetHandle.getSegmentName());
         }
         // check whether the source is sealed
         SegmentProperties si = doGetStreamSegmentInfo(sourceSegment);
+        String sourcePath = config.getPrefix() + sourceSegment;
         Preconditions.checkState(si.isSealed(), "Cannot concat segment '%s' into '%s' because it is not sealed.",
                 sourceSegment, targetHandle.getSegmentName());
 
-        if (config.getSmallObjectThreshold() < si.getLength()) {
-
-            //Copy the first part
-            CopyPartRequest copyRequest = new CopyPartRequest(config.getBucket(),
-                    targetPath,
-                    config.getBucket(),
-                    targetPath,
-                    uploadId,
-                    1).withSourceRange(Range.fromOffsetLength(0, offset));
-            CopyPartResult copyResult = client.copyPart(copyRequest);
-
-            partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
-
-            //Copy the second part
-            S3ObjectMetadata metadataResult = client.getObjectMetadata(config.getBucket(),
-                    config.getPrefix() + sourceSegment);
-            long objectSize = metadataResult.getContentLength(); // in bytes
-
-            copyRequest = new CopyPartRequest(config.getBucket(),
-                    config.getPrefix() + sourceSegment,
-                    config.getBucket(),
-                    targetPath,
-                    uploadId,
-                    2).withSourceRange(Range.fromOffsetLength(0, objectSize));
-
-            copyResult = client.copyPart(copyRequest);
-            partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
-
-            //Close the upload
-            client.completeMultipartUpload(new CompleteMultipartUploadRequest(config.getBucket(),
-                    targetPath, uploadId).withParts(partEtags));
-
-            client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
-
+        if (config.getSmallObjectSizeLimitForConcat() < si.getLength()) {
+            doConcatWithMultipartUpload(targetPath, sourceSegment, offset);
+            ExtendedS3Metrics.LARGE_CONCAT_COUNT.inc();
         } else {
-            try (InputStream reader = client.readObjectStream(config.getBucket(),
-                    config.getPrefix() + sourceSegment, Range.fromOffsetLength(0, si.getLength()))) {
-                client.putObject(this.config.getBucket(),
-                        targetPath,
-                        Range.fromOffsetLength(offset, si.getLength()),
-                        new BufferedInputStream(reader, Math.toIntExact(si.getLength())));
-            } catch (Exception e) {
-                throw Exceptions.sneakyThrow(e);
-            }
-            client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
+            doConcatWithAppend(targetPath, sourcePath, offset, si.getLength());
         }
+        // Now delete the source object.
+        client.deleteObject(config.getBucket(), sourcePath);
+
         Duration elapsed = timer.getElapsed();
         log.debug("Concat target={} source={} offset={} bytesWritten={} latency={}.", targetHandle.getSegmentName(), sourceSegment, offset, si.getLength(), elapsed.toMillis());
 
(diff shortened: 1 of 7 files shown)


bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
```diff
@@ -93,6 +97,7 @@ private ExtendedS3StorageConfig(TypedProperties properties) throws Configuration
         String givenPrefix = Preconditions.checkNotNull(properties.get(PREFIX), "prefix");
         this.prefix = givenPrefix.endsWith(PATH_SEPARATOR) ? givenPrefix : givenPrefix + PATH_SEPARATOR;
         this.useNoneMatch = properties.getBoolean(USENONEMATCH);
+        this.smallObjectThreshold = properties.getInt(SMALL_OBJECT_THRESHOLD);
```
kevinhan88: If the threshold is configurable, then please add it into config.properties as commented line with description.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
```diff
@@ -30,6 +30,7 @@
     public static final Property<String> BUCKET = Property.named("bucket", "");
     public static final Property<String> PREFIX = Property.named("prefix", "/");
     public static final Property<Boolean> USENONEMATCH = Property.named("useNoneMatch", false);
+    public static final Property<Integer> SMALL_OBJECT_THRESHOLD = Property.named("smallObjectThreshold", 1024 * 1024);
```
andreipaduroiu: What does `smallObjectThreshold` mean? Someone who has no idea what this means will not be able to use this properly.

Second you must always include the unit of measure in your property names. Is this bytes, KB, MB, ..., TB?

Please name this something more appropriate, such as `nativeConcatThresholdBytes` or `multiPartUploadConcatThresholdBytes` .

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
```diff
@@ -75,6 +76,9 @@
     @Getter
     private final boolean useNoneMatch;
 
+    @Getter
+    private final int smallObjectThreshold;
```
andreipaduroiu: Add a comment to the field explaining exactly what the behavior is.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
andreipaduroiu: +1. In `config.properties`, please add recommended values and explain what could happen if setting it too small or too large.

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3StorageTest.java
```diff
@@ -227,6 +227,26 @@ public void testExistsWithPrefix() throws Exception {
...
+        val adapterConfig = ExtendedS3StorageConfig.builder()
+                .with(ExtendedS3StorageConfig.CONFIGURI, setup.configUri)
+                .with(ExtendedS3StorageConfig.BUCKET, setup.adapterConfig.getBucket())
+                .with(ExtendedS3StorageConfig.PREFIX, "samplePrefix")
+                .with(ExtendedS3StorageConfig.USENONEMATCH, true)
+                .with(ExtendedS3StorageConfig.SMALL_OBJECT_THRESHOLD, 1)
+                .build();
+        final String context = createSegmentName("Concat");
+        try (Storage s = createStorage(setup.client, adapterConfig, executorService())) {
+            testConcat(context, s);
```
andreipaduroiu: How do you know MPU was used?

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -403,39 +405,53 @@ private Void doConcat(SegmentHandle targetHandle, long offset, String sourceSegm
...
+            try (InputStream reader = client.readObjectStream(config.getBucket(),
+                    config.getPrefix() + sourceSegment, Range.fromOffsetLength(0, si.getLength()))) {
+                client.putObject(this.config.getBucket(),
+                        targetPath,
+                        Range.fromOffsetLength(offset, si.getLength()),
+                        new BufferedInputStream(reader, Math.toIntExact(si.getLength())));
+            } catch (Exception e) {
+                throw Exceptions.sneakyThrow(e);
+            }
+            client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
```
kevinhan88: Only delete original object in case of success, right? move this line into try block.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -403,39 +405,53 @@ private Void doConcat(SegmentHandle targetHandle, long offset, String sourceSegm
...
+
+        } else {
+            try (InputStream reader = client.readObjectStream(config.getBucket(),
+                    config.getPrefix() + sourceSegment, Range.fromOffsetLength(0, si.getLength()))) {
+                client.putObject(this.config.getBucket(),
+                        targetPath,
+                        Range.fromOffsetLength(offset, si.getLength()),
+                        new BufferedInputStream(reader, Math.toIntExact(si.getLength())));
+            } catch (Exception e) {
+                throw Exceptions.sneakyThrow(e);
```
kevinhan88: Please give more information here, particularly this is something new we're trying

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -403,39 +405,53 @@ private Void doConcat(SegmentHandle targetHandle, long offset, String sourceSegm
...
+            partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
 
+            //Close the upload
+            client.completeMultipartUpload(new CompleteMultipartUploadRequest(config.getBucket(),
+                    targetPath, uploadId).withParts(partEtags));
+
+            client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
+
+        } else {
+            try (InputStream reader = client.readObjectStream(config.getBucket(),
```
kevinhan88: Why not use CopyObjectRequest directly?

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
medvedevigorek: Do we use any kind of prefix to highlight which component this config property from? Something like `extendeds3.`

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -403,39 +405,53 @@ private Void doConcat(SegmentHandle targetHandle, long offset, String sourceSegm
...
-
-        //Close the upload
-        client.completeMultipartUpload(new CompleteMultipartUploadRequest(config.getBucket(),
-                targetPath, uploadId).withParts(partEtags));
-
-        client.deleteObject(config.getBucket(), config.getPrefix() + sourceSegment);
-        Duration elapsed = timer.getElapsed();
+        if (config.getSmallObjectThreshold() < si.getLength()) {
+
+            //Copy the first part
```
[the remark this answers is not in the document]
medvedevigorek: Shall the size of a segment to merge be strictly less than the threshold?

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
kevinhan88: yes, there are namespaces for configs. so in config.properties it's going to be something like extendeds3.smallObjectThreshold, which will be only picked up by ExtendedS3Config

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
medvedevigorek: Delete operation may fail too.  Need some error handling here and of course sufficient logging.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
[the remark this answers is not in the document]
sachin-j-joshi: broke those into two methods.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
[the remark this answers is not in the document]
sachin-j-joshi: Actually we don't need to handle exception here. It will bubble up to execute method in this class and then will be handled in throwException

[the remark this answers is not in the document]
sachin-j-joshi: remove catch block, we don't need it.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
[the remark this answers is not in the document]
sachin-j-joshi: check is such that large object is greater than threshold

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
sachin-j-joshi: Good catch @kevinhan88 .Added to config value.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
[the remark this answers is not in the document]
sachin-j-joshi: Fixed.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -108,6 +108,7 @@
     public static final String STORAGE_CREATE_COUNT = PREFIX + "segmentstore.storage.create_count";      // Counter
     public static final String STORAGE_DELETE_COUNT = PREFIX + "segmentstore.storage.delete_count";      // Counter
     public static final String STORAGE_CONCAT_COUNT = PREFIX + "segmentstore.storage.concat_count";      // Counter
+    public static final String STORAGE_LARGE_CONCAT_COUNT = PREFIX + "segmentstore.storage.small_concat_count"; // Counter
```
kevinhan88: It says LARGE concat count but with literal value of "small"

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
kevinhan88: Why the metric says Large but config is called small? Please straight them out.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
kevinhan88: Please add unit into the name as commented.

sachin-j-joshi: renamed

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3StorageTest.java
sachin-j-joshi: added check

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
sachin-j-joshi: actually we don't need to catch exception here.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
sachin-j-joshi: oh no!

config/config.properties
```diff
@@ -448,6 +448,15 @@ extendeds3.configUri=http://localhost:9020?identity=user&secretKey=password
 # Prefix is optional.
 # extendeds3.prefix=
 
+# Size of ECS objects in bytes above which it is no longer considered a small object.
+# This value is used to optimize transactions performance when size of transaction segments is small.
+# For small transaction segments, to implement concat ExtendedS3Storage reads complete source segment and appends it to target
+# instead of using multipart upload.
+# smallObjectSizeLimitForConcat is optional.
+# Valid values: Positive integer.
```
kevinhan88: Please call out what is the default value, which really matters in the case.
Also consider saying something like: "In most cases, the default value should just work. Change this value with cautiousness."

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
sachin-j-joshi: Fixed.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3StorageConfig.java
sachin-j-joshi: value is in bytes so by convention without suffix.

config/config.properties
sachin-j-joshi: Good point. Updated

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
sachin-j-joshi: Yes. actually this could be a nasty bug. If concat succeeds but delete fails then we are leaving target segment in a state where it is bigger than before. But for the caller the operation has failed so it will not update size. If now operation is retried it will be done with older offset which will fail again and again. Solution is to either move delete as a separate step to upper layers or not fail the concat call when delete operation fails.

andreipaduroiu: This is not as bad as you think. If the concat succeeds but the delete fails, the storage writer will reattempt. You will need to throw a `BadOffsetException` because your precondition will not hold. The StorageWriter will them enter a reconciliation state and figure out that the source segment has actually been merged in. 

The only downside is that the old segment will be left behind, occupying space but not being used.

medvedevigorek: Okay, it makes sense then.

sachin-j-joshi: I just realized that in case of delete failures, the concat operation will be retried with same offset and we'll simply overwrite the same byte range with same contents. There is a test case that tests this idempotent. If the delete failure was intermittent then eventually it will succeed. Otherwise its likely that root cause will manifest in other operations too (like object deleted by some zombie etc. in which case read will fail. Same for credential issues. )

medvedevigorek: The other observation is S3Client being used here doesn't declare any exceptions for its methods but actually it does throw [S3Exception](https://github.com/EMCECS/ecs-object-client-java/blob/master/src/main/java/com/emc/object/s3/S3Exception.java) if it receives a response with status code other than 2xx. 

Since S3Exception is defined as a runtime exception we need to ensure the call stack handles it correctly and doesn't lead to a serious disaster as it may not assume to handle a runtime exception at all.

sachin-j-joshi: - Updated to throw BadOffsetException on failure during delete. This will trigger reconciliation.
- In normal course we continue to allow concat to be idempotent.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -386,23 +388,57 @@ private void setPermission(SegmentHandle handle, Permission permission) {
...
+        } else {
+            doConcatWithAppend(targetPath, sourcePath, offset, si.getLength());
+        }
+        // Now delete the source object.
+        client.deleteObject(config.getBucket(), sourcePath);
+
+        Duration elapsed = timer.getElapsed();
+        log.debug("Concat target={} source={} offset={} bytesWritten={} latency={}.", targetHandle.getSegmentName(), sourceSegment, offset, si.getLength(), elapsed.toMillis());
+
+        ExtendedS3Metrics.CONCAT_LATENCY.reportSuccessEvent(elapsed);
```
medvedevigorek: _latency_ and _bytes_ are recorded only in case of success, in case of a failure in `deleteObject()` those will not be recorded. Don't they need to be recorded anyway?

andreipaduroiu: We never record these for failed requests.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -386,23 +388,57 @@ private void setPermission(SegmentHandle handle, Permission permission) {
...
+        ExtendedS3Metrics.CONCAT_BYTES.add(si.getLength());
+        ExtendedS3Metrics.CONCAT_COUNT.inc();
```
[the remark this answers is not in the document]
andreipaduroiu: In other words we should update them around the same time. I prefer after, so that we may only record successful runs. 

If we record it before, then every retry will be counted, which is not what we want.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
kevinhan88: We can start reportFailEvent in case concat (or/and delete) failed. 
We may report success only after both concat and delete succeeded, otherwise report failure.

andreipaduroiu: Yes, we could do that. But this would be the only place in the code we currently do that. We can tackle this as a widespread change in another PR.

kevinhan88: Shall we delete the source segment through ExtendedS3Storage.delete(), instead of calling client.delete() directly? 
Going through that path, we have Tier2 DELETE metrics triggered.

kevinhan88: Yes, just overload the doDelete(SegmentHandle handle) with doDelete(String bucket, String prefix, String segmentName).

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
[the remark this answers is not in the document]
sachin-j-joshi: I think I would agree with @andreipaduroiu , for 0.7 let's record metrics only for successful events. Let's track this as a separate issue to be resolved  in 0.8 along with BadOffsetException change.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
sachin-j-joshi: For 0.7 let's update metrics for only successful requests. 
Let's track the issue as separate issue for 0.8

PR #4640, opened 2020-03-24
Issue 4638: Propagate retryable exception to the client instead of returning Status.Failure on all exceptions

### LINKED ISSUE #4638: Commit and abort Transanction requests should return retryable failure response to the client
**Problem description**
If transaction commit/abort requests fail, for all exceptions controller sends the response with status as failure. 
Instead, for expected retryable failures controller should propagate the error back to the client such that it is retried. 

**Problem location**
ControllerService

Propagate the appropriate error code back to the client instead of sending failure status for retryable expected exceptions.

shiveshr:
**Change log description**  
Controller apis for commit and abort transactions do not propagate error back to the client if the error is retryable. Instead they return the status.Failure which leads client to not perform any retries. 

**Purpose of the change**  
Fixes #4638 

**What the code does**  
In controllerservice.java we throw all retryable exceptions back to grpc layer which translates them to appropriate error codes and fails grpc calls. 

For any other exception, the status code Failure is returned. 

**How to verify it**  
Unit test added.


controller/src/test/java/io/pravega/controller/server/v1/ControllerServiceTest.java
```diff
@@ -65,7 +73,7 @@
     private final String stream2 = "stream2";
     private final ScheduledExecutorService executor = Executors.newScheduledThreadPool(10);
 
-    private final StreamMetadataStore streamStore = StreamStoreFactory.createInMemoryStore(executor);
+    private final StreamMetadataStore streamStore = spy(StreamStoreFactory.createInMemoryStore(executor));
```
shrids: timeout for all the tests should be added.

PR #4641, opened 2020-03-24
Issue 4639: Table store helper to retry only on unknownhost and auth failure for update requests

### LINKED ISSUE #4639: Controller's table client should not retry for updates on Connection Exceptions
**Problem description**
While interacting with segment store for table queries, we retry on store exceptions for all requests. 
This can have undesired effect for conditional updates because they can now fail with write conflict exception. 
Ideally retry for updates should be done for WrongHost but not for other connection failure exceptions. 

**Problem location**
PravegaTableStoreHelper.java

Fix the issue. do not retry on connection failure exceptions for updates.

shiveshr:

**Change log description**  
For table update calls, retry should be done only for WrongHost and auth failed but not for other store connection failure exceptions.

**Purpose of the change**  
Fixes #4639 

**What the code does**  
 While interacting with segment store for table queries, we retry on store exceptions for all requests.
This can have undesired effect for conditional updates because they can now fail with write conflict exception.

**How to verify it**  
Unit test added.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStoreHelper.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStoreHelper.java
@@ -50,6 +50,8 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import static io.pravega.controller.server.WireCommandFailedException.Reason.ConnectionDropped;
+import static io.pravega.controller.server.WireCommandFailedException.Reason.ConnectionFailed;
 import static io.pravega.controller.store.stream.AbstractStreamMetadataStore.DATA_NOT_FOUND_PREDICATE;
 
 /**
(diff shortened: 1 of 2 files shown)


controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStoreHelper.java
```diff
@@ -176,7 +178,7 @@ public void invalidateCache(String table, String key) {
                 new TableEntryImpl<>(new TableKeyImpl<>(key.getBytes(Charsets.UTF_8), KeyVersion.NOT_EXISTS), value));
         Supplier<String> errorMessage = () -> String.format("addNewEntry: key: %s table: %s", key, tableName);
         return withRetries(() -> segmentHelper.updateTableEntries(tableName, entries, authToken.get(), RequestTag.NON_EXISTENT_ID),
-                errorMessage)
+                errorMessage, true)
```
shrids: This change causes a slight change in the behavior of the controller wrt to createScope().

```
This change prevents retries in the case of `ConnectionDropped` or `ConnectionFailed`. In the case of `createScope()` 
workflow, this will prevent retries by the controller. And in this scenario the retries of the client would used.
```

shrids:
pravega-branch-builder/2266/  - system test run is in progress.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStoreHelper.java
shiveshr: Yes, that is correct. But its not necessarily a behaviour change (except the bug fix). Its just fail faster rather than retry. 
Most updates happen from workflows. 
Some happen in the context of client calls.

shrids:
`/job/test-pravega-continuous/2729/` System test run with this change is successful.

PR #4645, opened 2020-03-26
Issue 4273: Revised Throttling Metrics

### LINKED ISSUE #4273: Segment Store Throttling metrics
**Problem description**
1. Reported Throttling values are correct except for interruptible throttling delays (such as the ones induced by the cache). If a throttling delay is interrupted (and as such, cut short), the Throttler will delay less time than it originally reported. This could cause throttling delay metrics to be over-reported .
2. We need a way to report the type of throttling applied (via metrics).

**Problem location**
`Throttler.java`. Most likely in the `throttle`/`throttleOnce` method.

1. Measure throttling delay end-to-end and report that.
2. Report Throttling Type (maybe as a tag?). This should be fetched from the throttler source name.

co-jo:
**Change log description**  
* Added 3 new metrics (and removed one). Instead of a single encompassing `operationProcessDelay` metric, one for each `Throttler` type is created. 
* Adjust for possible throttling interruptions in the `Throttler` and record the source of the throttling into the appropriate metric.
* Add trivial metric tests and also throttling metric tests which test the reported metrics upon interruption.
 
**Purpose of the change**  
Fixes #4273 

**What the code does**  
* Adds new throttling metrics and properly accounts for interruptions during throttling.

**How to verify it**  
Run the provided unit tests.


segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/ThrottlerTests.java
```diff
@@ -145,6 +157,187 @@ public void testInterruptedDurableDataLogDelay() throws Exception {
         testInterruptedDelay(ThrottlerCalculator.ThrottlerName.DurableDataLog);
     }
 
+    /**
```
co-jo: Working on reducing the amount of duplicate code within these tests. 

I tried to avoid importing metrics classes, however it seemed the most appropriate place, as the accurate reporting of the Metrics is entirely determined on their placement in the `Throttler` class.

PR #4646, opened 2020-03-26
Issue 4273: Revised Throtting Metrics.

### LINKED ISSUE #4273: Segment Store Throttling metrics

co-jo:
**Change log description**  
* Added 3 new metrics (and removed one). Instead of a single encompassing `operationProcessDelay` metric, one for each `Throttler` type is created. 
* Adjust for possible throttling interruptions in the `Throttler` and record the source of the throttling into the appropriate metric.
* Add trivial metric tests and also throttling metric tests which test the reported metrics upon interruption.
 
**Purpose of the change**  
Fixes #4273 

**What the code does**  
* Adds new throttling metrics and properly accounts for interruptions during throttling.

**How to verify it**  
Run the provided unit tests.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
```diff
@@ -154,36 +168,57 @@ public OperationProcessor(int containerId) {
...
             this.operationQueueWaitTime = STATS_LOGGER.createStats(MetricsNames.OPERATION_QUEUE_WAIT_TIME, this.containerTag);
-            this.operationProcessorDelay = STATS_LOGGER.createStats(MetricsNames.OPERATION_PROCESSOR_DELAY_MILLIS, this.containerTag);
             this.operationCommitLatency = STATS_LOGGER.createStats(MetricsNames.OPERATION_COMMIT_LATENCY, this.containerTag);
             this.operationLatency = STATS_LOGGER.createStats(MetricsNames.OPERATION_LATENCY, this.containerTag);
             this.memoryCommitLatency = STATS_LOGGER.createStats(MetricsNames.OPERATION_COMMIT_MEMORY_LATENCY, this.containerTag);
             this.memoryCommitCount = STATS_LOGGER.createStats(MetricsNames.OPERATION_COMMIT_MEMORY_COUNT, this.containerTag);
             this.processOperationsLatency = STATS_LOGGER.createStats(MetricsNames.PROCESS_OPERATIONS_LATENCY, this.containerTag);
             this.processOperationsBatchSize = STATS_LOGGER.createStats(MetricsNames.PROCESS_OPERATIONS_BATCH_SIZE, this.containerTag);
+            this.operationProcessorCacheDelay = STATS_LOGGER.createStats(
```
andreipaduroiu: Why do we need three separate metrics here? We can do this using a single metric and change the tag on each publish.

Check out https://github.com/pravega/pravega/blob/master/shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsLogAppender.java

We use a single metric (ok, two of them `MetricsNames.LOG_ERRORS` or `MetricsNames.LOG_WARNINGS`), then in `recordEvent` we use `DYNAMIC_LOGGER` to record a value and attach a tag.

What I think we can do in the throttling metrics is define 3 tags (Cache, DurableDataLog and Batching - or better, just get them out of `ThrottlerCalculator.DelayResult.getThrottlerName().toString()`) and then use a single metric (you can use the one we had before) to report the delay value with the appropriate number.

kevinhan88: After reading the code, seems in the case there are two tags - container (x1) and throttler (x3) - so it's a best practice to pre-create all possible tag values combinations in advance, and use enum/map to locate them.
Please avoid dynamically generating tag values asap. (Tag values should be, in most cases, constant)

kevinhan88: The code could be like this:
this.containerTag = containerTag(containerId);
this.throttler-cache = throttlerTag("cache");
this.throttler-batching = throttlerTag("batching");
...
DYNAMIC_LOGGER.reportGaugeValue(MetricsNames.OPERATION_PROCESSOR_DELAY_MILLIS, millis, containerTag, whateverThrottlerTag);

co-jo: I attempted using the `reportGaugeValue` functionality. But I had issues passing in primitive numeric types (Could not change the value of the gauge after the initial argument reference was lost, i.e the method returned). According to the micrometer documentation this is the expected result.

Using a `Counter` or a `MeteredEvent` also don't seem appropriate for the use case.

co-jo: For example, I wrote two simple tests and stuck them in `ThrottlerTests`:

```
    @Test
    public void a() throws Exception {
        DYNAMIC_LOGGER.reportGaugeValue("gauge-key", 10);
        log.debug("gauge-key: {}", MetricRegistryUtils.getGauge("gauge-key").value());
        DYNAMIC_LOGGER.reportGaugeValue("gauge-key", 15);
        log.debug("gauge-key: {}", MetricRegistryUtils.getGauge("gauge-key").value());
    }
    @Test
    public void b() throws Exception {
        log.debug("gauge-key before: {}", MetricRegistryUtils.getGauge("gauge-key").value());
        DYNAMIC_LOGGER.reportGaugeValue("gauge-key", 20);
        log.debug("gauge-key after: {}", MetricRegistryUtils.getGauge("gauge-key").value());
    }
```

Which produced the output:

```
Stats enabled
gauge-key: 10.0
gauge-key: 15.0
Stats enabled
gauge-key before: 15.0
gauge-key after: 15.0
```

kevinhan88: Please refer to MetricsProviderTest.java for similar test cases, in particular compare how the metrics registry gets initialized.

co-jo: > 
> 
> Please refer to MetricsProviderTest.java for similar test cases, in particular compare how the metrics registry gets initialized.

I'm not sure I understand. I don't see any difference in usages, with the exception of `MetricsProviderTest.java` calling `freezeGaugeValue` before returning.  But that still leaves the question about why the `reportOpenTransacitons` call is able to work properly.

kevinhan88: Unit tests are supposed to be independent, and be run in any sequence.
Seems the test b() carries the result of test a (15), it is a test leaking. The first line of test b() should receive an nullpointerexception as the gauge wasn't created by that time.
Please refer to MetricsProviderTest again, look for @Before which annotates the method to run before EVERY test case.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
```diff
@@ -148,13 +144,16 @@ private void report() {
         private final OpStatsLogger processOperationsLatency;
         private final OpStatsLogger processOperationsBatchSize;
         private final String[] containerTag;
+        private final HashMap<String, String[]> throttlerTags = new HashMap<>();
```
andreipaduroiu, line 147: I'm not a big fan of this type of caching, for a couple of reasons:
1. You need to do a lookup in this map every time you report something.
2. The map's keys are not the same as the ones that are in `ThrottlerCalculator`. If we decide (for whatever reason) to rename, remove or add one in there, this will not work.
3. If (continued from 2 above), there is no entry in this map for a throttler name, then it will return null - how will the reporting work? will it throw a NullPtr? 

I am OK with generating a new array every time we invoke this. Generating a 4-element array is insignificant compared to all the other work that is done in the Operation Processor so there's no need to try to optimize it.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
```diff
@@ -122,6 +122,8 @@ boolean isThrottlingRequired() {
                 if (remaining > 0 && remaining < delay.get().getDurationMillis()) {
                     delay.set(delay.get().withNewDelay(remaining));
                 }
+                int incurredDelay = (int) (existingDelay.remaining.getInitial().toMillis() - existingDelay.remaining.getRemaining().toMillis());
```
andreipaduroiu, line 125: please add a method called `getElapsed` on `TimeoutTimer` that does this calculation. Doing so will remove the need to expose `getInitial` to the outside.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
```diff
@@ -138,13 +140,12 @@ boolean isThrottlingRequired() {
...
 
     private CompletableFuture<Void> throttleOnce(ThrottlerCalculator.DelayResult delay) {
-        this.metrics.processingDelay(delay.getDurationMillis());
         if (delay.isMaximum()
                 || delay.getThrottlerName() == ThrottlerCalculator.ThrottlerName.DurableDataLog) {
             // Increase logging visibility if we throttle at the maximum limit (which means we're likely to fully block
             // processing of operations) or if this is due to us not being able to ingest items quickly enough.
             log.warn("{}: Processing delay = {}.", this.traceObjectId, delay);
-        } else {
+        } else if (delay.getThrottlerName() != null) {
```
andreipaduroiu, line 148: Why would this be null? 

And even if it is, I **still** want to see this logged somewhere - this is important for certain debugging activities.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
```diff
@@ -159,8 +160,17 @@ boolean isThrottlingRequired() {
                             result.delayFuture,
                             ex -> ex instanceof ThrottlingInterruptedException,
                             this::throttle)
-                    .whenComplete((r, e) -> this.currentDelay.set(null));
+                    .whenComplete((r, e) -> {
+                        if (this.currentDelay.get() != null && currentDelay.get().remaining.getRemaining().toMillis() <= 0) {
```
andreipaduroiu, line 164: use `hasRemaining` in `TimeoutTimer`

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/ThrottlerTests.java
```diff
@@ -145,6 +161,191 @@ public void testInterruptedDurableDataLogDelay() throws Exception {
...
+        calculator.setDelayMillis(nextDelay.next());
+        val t1 = t.throttle();
+        Assert.assertFalse("Not expected throttle future to be completed yet.", t1.isDone());
+
+        // For every delay that we want to submit, notify that the cache cleanup has completed, which should cancel the
+        // currently running throttle cycle and request the next throttling value.
+        for (int i = 1; i < suppliedDelays.size(); i++) {
+            // Interrupt the current throttle cycle.
+            t.notifyThrottleSourceChanged();
+            //Assert.assertFalse("Not expected throttle future to be completed yet.", t1.isDone());
```
andreipaduroiu, line 241: Please do not check in commented out code. Either add it back or delete it completely.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/ThrottlerTests.java
```diff
@@ -145,6 +161,191 @@ public void testInterruptedDurableDataLogDelay() throws Exception {
...
+                (int) getThrottlerMetric(ThrottlerCalculator.ThrottlerName.Batching)
+        );
+
+        durableDataLogDelay = max(cacheDelay, batchingDelay) + 1000;
+        // This should force the throttler calculator to pick up the durable data log throttler.
+        durableDataLogCalculator.setDelayMillis(durableDataLogDelay);
+        val t2 = t.throttle();
+        Assert.assertFalse("Not expected throttle future to be completed yet.", t2.isDone());
+        // Arbitrary amount of time goes by.
+        TimeUnit.MILLISECONDS.sleep(throttlerActiveTime);
```
andreipaduroiu, line 326: I really do not like sleeping in tests (or anywhere else for that matter). Can you rewrite this test without the need to sleep?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/ThrottlerTests.java
```diff
@@ -195,6 +396,26 @@ private ThrottlerCalculator wrap(ThrottlerCalculator.Throttler calculatorThrottl
...
+
+    private String[] throttlerTags(ThrottlerCalculator.ThrottlerName name)  {
+        return  new String[] { "container", String.valueOf(this.containerId), "throttler", name.toString() };
+    }
+
+    private double getThrottlerMetric(ThrottlerCalculator.ThrottlerName name) {
+        return MetricRegistryUtils.getGauge(MetricsNames.OPERATION_PROCESSOR_DELAY_MILLIS, throttlerTags(name)).value();
+    }
+
+    private static int min(int a, int b) {
```
andreipaduroiu, line 411: `Math.min`?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/ThrottlerTests.java
```diff
@@ -195,6 +396,26 @@ private ThrottlerCalculator wrap(ThrottlerCalculator.Throttler calculatorThrottl
...
+        return a < b ? a : b;
+    }
+
+    private static int max(int a, int b) {
```
andreipaduroiu, line 415: `Math.max`?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/logs/ThrottlerTests.java
```diff
@@ -195,6 +396,26 @@ private ThrottlerCalculator wrap(ThrottlerCalculator.Throttler calculatorThrottl
         return ThrottlerCalculator.builder().throttler(calculatorThrottler).build();
     }
 
+    private ThrottlerCalculator wrap(ArrayList<ThrottlerCalculator.Throttler> throttlers) {
+        return ThrottlerCalculator.builder().throttlers(throttlers).build();
+    }
+
+    private String[] throttlerTags(ThrottlerCalculator.ThrottlerName name)  {
```
andreipaduroiu, line 403: Can't you get this from `MetricsTags`?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/Throttler.java
co-jo: I did not dig too deeply, but It appears that `throttle` is being called somewhere in the testing code, without having been assigned a `Throttler` and returns a `DelayResult` with a default `null` name. When I called `delay.getThrottlerName()` on this null value, it halts execution. This was the reason for my TravisCI builds failing. Having said that I think this condition is redundant anyways. Will remove.

andreipaduroiu: Well, by looking at the code in this `else` block I don't see how that could happen. `log.debug` is passed `delay`, and it will only invoke `toString` on it if debug-level logging is enabled. I believe it is not (but I am not sure). However if there is a bug like you said, then it must be in the `toString` implementation of `delay`, so if anything it should be fixed there.

co-jo:
Appears there is a problem with the system tests image, but note that the build _does_ pass: https://travis-ci.org/github/co-jo/pravega/builds/673609377.

PR #4649, opened 2020-03-27
Issue 4648: System tests are failing with zookeeper-operator 0.2.6-rc1

### LINKED ISSUE #4648: System tests are failing with zookeeper-operator 0.2.6-rc1
**Problem description**
In system tests, for deploying zookeeper, we are passing `size` field, that got deprecated in `0.2.6` release.

**Problem location**
https://github.com/pravega/pravega/blob/master/test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/ZookeeperK8sService.java

Change the deprecated `size` field to `replicas`

anishakj:
Signed-off-by: anisha.kj <anisha.kj@dell.com>

**Change log description**  
Since `replicas` field is not set, system tests were deploying `3` instances of `zk` instead of `1`. Set the replicas correctly to fix the issue

**Purpose of the change**  
Fixes  #4648

**What the code does**  
Changed the deprecated size field to replicas

**How to verify it**  
System tests should pass

### DIFF
--- a/test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/ZookeeperK8sService.java
+++ b/test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/ZookeeperK8sService.java
@@ -239,7 +239,7 @@ public class ZookeeperK8sService extends AbstractService {
                 .put("kind", CUSTOM_RESOURCE_KIND)
                 .put("metadata", ImmutableMap.of("name", deploymentName))
                 .put("spec", ImmutableMap.builder().put("image",  getImageSpec(DOCKER_REGISTRY + PREFIX + "/" + ZOOKEEPER_IMAGE_NAME, PRAVEGA_ZOOKEEPER_IMAGE_VERSION))
-                                         .put("size", clusterSize)
+                                         .put("replicas", clusterSize)
                                          .build())
                 .build();
     }


test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/ZookeeperK8sService.java
```diff
@@ -239,7 +239,7 @@ private V1Deployment getDeployment() {
                 .put("kind", CUSTOM_RESOURCE_KIND)
                 .put("metadata", ImmutableMap.of("name", deploymentName))
                 .put("spec", ImmutableMap.builder().put("image",  getImageSpec(DOCKER_REGISTRY + PREFIX + "/" + ZOOKEEPER_IMAGE_NAME, PRAVEGA_ZOOKEEPER_IMAGE_VERSION))
-                                         .put("size", clusterSize)
+                                         .put("replicas", clusterSize)
```
shrids: This change seems implies that the newer zookeeper operator is not backward compatible. Is that the case ? if yes could share the zookeeper-operator version above which this change will work.

anishakj: The changes are backward compatible as `replicas` filed is available in previous zookeeper  versions as well.

shrids:
System test run with this change `pravega/job/test-pravega-continuous/2727`

PR #4655, opened 2020-03-30
Issue 4653: Control rate of parallel Segment Container recoveries

### LINKED ISSUE #4653: Control rate of parallel Segment Container recoveries
**Problem description**
At the moment, a Segment Store can process multiple Segment Container recoveries in parallel. However, it may be the case that not all the data to be recovered for each container fits in cache. If this situation arises, the Segment Store throws an exception like this:
```
2020-03-29 07:02:46,010 116663978 [core-28] INFO  i.p.s.s.reading.ContainerReadIndex - ReadIndex[27] Exit RecoveryMode.
2020-03-29 07:02:46,010 116663978 [core-28] ERROR i.p.s.server.logs.DurableLog - DurableLog[27] Recovery FAILED.
io.pravega.segmentstore.storage.cache.CacheFullException: DirectMemoryCache full: Stored = 8570578946, Used = 8587841536, Reserved = 16777216, Allocated = 8589934592, Max = 8589934592.
	at io.pravega.segmentstore.storage.cache.DirectMemoryCache.getNextAvailableBuffer(DirectMemoryCache.java:405)
```` 
We need to better control this issue, otherwise the system may get stuck on this situation for an undetermined amount of time.

**Problem location**
Segment Container recovery

Add a parameter to control the number of Segment Container recoveries to be processes, and setting the default to 1 to be conservative to prevent this issue.

### PR DESCRIPTION
**Pull request by RaulGracia**
**Change log description**  
Enforces sequential start of containers to prevent multiple concurrent container recoveries from overloading the cache.

**Purpose of the change**  
Fixes #4653.

**What the code does**  
Until now, a Segment Store could start Segment Containers in parallel, which also includes the process of recovery in the case that the Segment Container to start is not new. However, if we think about a Segment Store recovering in parallel multiple Segment Containers, it may be the case that the data to process for each container is larger than the size of the cache (not to mention that the cache can still be used by other containers). This PR mainly serializes the execution of Segment Container starts with the main objective of preventing overloading the cache.

To this end, this PR includes the following changes:
-  In `ZKSegmentContainerMonitor`, we use the existing class `OrderedItemProcessor` to serialize the executions of `startContainer` method.
- The behavior of `OrderedItemProcessor` was by default to shut down the processor in case of an exception being thrown by any of the items being processed. While the default behavior remains the same, we added to this class a flag (`closeOnException`) to tolerate exceptions while processing items to continue processing the remaining ones. In our case, we want to start all the containers possible, even in the case that one/few of the containers have problems (i.e., `startContainer` throws an exception), which could be retried.

**How to verify it**  
Added a new test to `OrderedItemProcessor` to verify the behavior of `closeOnException`. The rest of tests for are passing, including system tests (build 2730).

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java
@@ -37,7 +37,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
 import java.util.function.Supplier;
 import lombok.SneakyThrows;
 import lombok.Synchronized;
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java
@@ -102,7 +102,9 @@ public class ZKSegmentContainerMonitor implements AutoCloseable {
         this.hostContainerMapNode = new NodeCache(zkClient, clusterPath);
         this.assigmentTask = new AtomicReference<>();
         this.lastReportTime = new AtomicLong(CURRENT_TIME_MILLIS.get());
-        this.startContainerOrderedProcessor = new OrderedItemProcessor<>(1, this::startContainer, this.executor);
+        // Allow OrderedItemProcessor to throw exceptions, as they may happen while starting containers. But we want to
+        // continue processing subsequent container starts irrespective of such failures.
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ZKSegmentContainerMonitor.java [hunk 1]
```diff
@@ -98,6 +102,9 @@
         this.hostContainerMapNode = new NodeCache(zkClient, clusterPath);
         this.assigmentTask = new AtomicReference<>();
         this.lastReportTime = new AtomicLong(CURRENT_TIME_MILLIS.get());
+        // Allow OrderedItemProcessor to throw exceptions, as they may happen while starting containers. But we want to
+        // continue processing subsequent container starts irrespective of such failures.
+        this.startContainerOrderedProcessor = new OrderedItemProcessor<>(1, this::startContainer, false, this.executor);
```
andreipaduroiu, line 107: I would feel much better if we made this number configurable. Let's try to do that.

RaulGracia: I added the possibility of configuring it.

PR #4658, opened 2020-03-31
Issue 4571: (Key-ValueTables) Client Control Path

### LINKED ISSUE #4571: Key-Value Tables: Implement Client-Control Path
**Problem description**
Implement Control Path (Create, Delete Key Value Tables) on the Client.
Depends on #4603.

See [PDP-39](https://github.com/pravega/pravega/wiki/PDP-39-Key-Value-Tables) for details.

**Problem location**
Client.

Implement.

andreipaduroiu:
**Change log description**  
Implemented Key-Value Tables control path APIs in the Client.

**Purpose of the change**  
Fixes #4571.

**What the code does**  
Added the following:
- `KeyValueTableInfo`: Contains info about a KVT (for now, just Scope and Name)
    - Similar to `StreamInfo`
- `KeyValueTableManager`/`KeyValueTableManagerImpl`: Creates, Deletes, Updates (config) or lists all KVTs.
    - Similar to `StreamManager`.

Refactored the following to support KVTs:
- `MockController`: added support for KVTs, and reused most of the code already in place for Streams.

**How to verify it**  
New unit tests added where needed.


andreipaduroiu:
Looks like this PR's diff code coverage is reported too low. This is because of the following reasons:
- A small number of executable LOCs are touched, which means that even if a few of those are not covered by direct tests (see below), the code cov % will drop significantly.
- In the client, we generally need to use a Controller Client to talk to a real controller. However, in the Client's unit tests, we use a MockController because we cannot instantiate the whole Controller. This generally requires a special constructor in the tested classes.
- Codecov only reports coverage by unit tests in the same package as the target class. It does not report cross-package coverage. For the client, the only way we can test it with the real controller is by adding those tests in the `integration` package (see `KeyValueTableCreateTests.java` in this PR). That exercises the `KeyValueTableManager` factory code and `KeyValueTableManagerImpl` constructor that uses the real controller, so that code is indeed exercised in tests, but not reported.

PR #4668, opened 2020-04-05
Issue 4667: Segment store pods are not coming up with external access enabled

### LINKED ISSUE #4667: Segment store pods are not coming up with external access enabled
**Problem description**
Segment store pods are entering in `crash loop back` state with external access enabled with pravega `0.5.0-rc0` operator

**Problem location**
init_kubernetes.sh

Fix the script

anishakj:
Signed-off-by: anisha.kj <anisha.kj@dell.com>

**Change log description**  
With  pravega-operator `0.5.0`, pod name has changed and it is different from service name. We were using pod name to find the service details 

**Purpose of the change**  
Fixes #4667 

**What the code does**  
Changed the script to use service name instead of service name

**How to verify it**  
Verified external connectivity is working fine.

### DIFF
--- a/docker/pravega/scripts/init_kubernetes.sh
+++ b/docker/pravega/scripts/init_kubernetes.sh
@@ -46,6 +46,8 @@ init_kubernetes() {
         local ns=${POD_NAMESPACE}
         local podname=${POD_NAME}
         # SERVICE_NAME is set by pravega-operator when external access is enabled from version 0.5.0 onwards
+        # SERVICE_NAME and POD_NAME are different in pravega-operator 0.5.0 onwards 
+
         if [ -z "${SERVICE_NAME}" ]; then
           local servicename=${podname}
         else
--- a/docker/pravega/scripts/init_kubernetes.sh
+++ b/docker/pravega/scripts/init_kubernetes.sh
@@ -45,29 +45,34 @@ init_kubernetes() {
 
         local ns=${POD_NAMESPACE}
         local podname=${POD_NAME}
+        if [ -z "${SERVICE_NAME}" ]; then
+          local servicename=${podname}
+        else
+          local servicename=${SERVICE_NAME}"-""${podname: -1}"
+        fi
         export PUBLISHED_ADDRESS=""
         export PUBLISHED_PORT=""
 	
-	export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".metadata.annotations[\"external-dns.alpha.kubernetes.io/hostname\"]" )
+	export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${servicename}" ".metadata.annotations[\"external-dns.alpha.kubernetes.io/hostname\"]" )
 
         if [[ -n ${PUBLISHED_ADDRESS} && "${PUBLISHED_ADDRESS:${#PUBLISHED_ADDRESS}-1}" == "." ]];
         then
           export PUBLISHED_ADDRESS=${PUBLISHED_ADDRESS::-1}
         fi
 
-        service_type=$( k8 "${ns}" "services" "${podname}" ".spec.type" )
+        service_type=$( k8 "${ns}" "services" "${servicename}" ".spec.type" )
         if [ "${service_type}" == "LoadBalancer" ]; then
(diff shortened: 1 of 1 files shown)


docker/pravega/scripts/init_kubernetes.sh
```diff
@@ -45,29 +45,34 @@ init_kubernetes() {
 
         local ns=${POD_NAMESPACE}
         local podname=${POD_NAME}
+        if [ -z "${SERVICE_NAME}" ]; then
```
shrids, line 48: Could we add a comment indicating that the SERVICE_NAME is set by the pravega operator when external access is enabled?.

anishakj: Added comment indicating same

docker/pravega/scripts/init_kubernetes.sh
```diff
@@ -45,29 +45,37 @@ init_kubernetes() {
...
                 	if [ -z "${PUBLISHED_ADDRESS}" ]; then
-                    		export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".status.loadBalancer.ingress[0].hostname" )
+                  	        export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${servicename}" ".status.loadBalancer.ingress[0].hostname" )
                 	fi
 		fi
-                export PUBLISHED_PORT=$( k8 "${ns}" "services" "${podname}" ".spec.ports[].port" )
+                export PUBLISHED_PORT=$( k8 "${ns}" "services" "${servicename}" ".spec.ports[].port" )
             done
         elif [ "${service_type}" == "NodePort" ]; then
             nodename=$( k8 "${ns}" "pods" "${podname}" ".spec.nodeName" )
```
pbelgundi, line 81: Need to change from `podname` to `servicename`  for serviceType=Nodeport as well??

docker/pravega/scripts/init_kubernetes.sh
```diff
@@ -45,29 +45,37 @@ init_kubernetes() {
 
         local ns=${POD_NAMESPACE}
         local podname=${POD_NAME}
+        # SERVICE_NAME is set by pravega-operator when external access is enabled from version 0.5.0 onwards
+        # SERVICE_NAME and POD_NAME are different in pravega-operator 0.5.0 onwards 
+
+        if [ -z "${SERVICE_NAME}" ]; then
+          local servicename=${podname}
```
pbelgundi, line 52: Adding a new service_name variable breaks compatibility of operator with older pravega builds. Could we instead change the value of pod_name in operator 0.5 so it points to service name instead of pod name and hence no changes to pravega startup scripts are required?

docker/pravega/scripts/init_kubernetes.sh
anishakj: Here it is looking at the  pod resource. So it is fine

medvedevigorek:
@anishakj what has changed on Pravega Operator side that requires a change in Pravega itself? Is Pravega Operator change really required or not? 

It will help if you share more details or even link to corresponding Pravega Operator issue as it's not clear what the things used to work with previous versions of Pravega Opearator and suddenly some changes in Pravega are required.

anishakj:
With pravega operator version `5.0` onwards, pod name and service name are different. Due to that `init_kubernetes.sh` script  in pravega fails to start SS containers, as the script uses pod  name to find the service details. Currently i am analyzing whether we can do the changes in operator alone and make it work

> @anishakj what has changed on Pravega Operator side that requires a change in Pravega itself? Is Pravega Operator change really required or not?
> 
> It will help if you share more details or even link to corresponding Pravega Operator issue as it's not clear what the things used to work with previous versions of Pravega Opearator and suddenly some changes in Pravega are required.

pbelgundi:
> used to work with previous versions of Pravega Operator and suddenly some changes in Pravega are required.

The fix for this issue has surfaced this problem: 
https://github.com/pravega/pravega-operator/issues/312
I would say it is incorrect for pravega startup scripts to assume that pod-name will always be the same as Service Name for any given component and the scripts to do need fixing at some point. For now, we'll do what it takes to maintain backward compatibility.

anishakj:
Closing this PR as different approach is taken to fix the above issue, by changing only operator

PR #4672, opened 2020-04-07
Issue 4669: (Bugfix) Fix serialization in PositionImpl.

### LINKED ISSUE #4669: client.stream.Position.toBytes fails to serialize -1
**Problem description**

Calling `client.stream.Position.toBytes` fails with some events returned by ReadEvents. Events consist of short UTF-8 strings such as "123456789". I believe this problem occurs for streams with multiple segments in which 1 segment has no events.

Stack trace is below.
```
Caused by: java.lang.IllegalArgumentException: writeCompactLong can only serialize longs in the interval [0, 2^62), given -1.
	at io.pravega.common.io.serialization.RevisionDataOutputStream.writeCompactLong(RevisionDataOutputStream.java:104)
	at io.pravega.client.stream.impl.PositionImpl$PositionSerializer.lambda$write00$2(PositionImpl.java:138)
	at io.pravega.common.io.serialization.RevisionDataOutputStream.writeMap(RevisionDataOutputStream.java:343)
	at io.pravega.client.stream.impl.PositionImpl$PositionSerializer.write00(PositionImpl.java:137)
	at io.pravega.common.io.serialization.VersionedSerializer$SingleType.serializeContents(VersionedSerializer.java:336)
	at io.pravega.common.io.serialization.VersionedSerializer$SingleType.serialize(VersionedSerializer.java:304)
	at io.pravega.common.io.serialization.VersionedSerializer$WithBuilder.serialize(VersionedSerializer.java:511)
	at io.pravega.common.io.serialization.VersionedSerializer.serialize(VersionedSerializer.java:130)
	at io.pravega.client.stream.impl.PositionImpl.toBytes(PositionImpl.java:174)
	at io.pravega.example.gateway.grpc.PravegaServerImpl.readEvents(PravegaServerImpl.java:204)
	... 9 common frames omitted
```

Below shows the first event in the stream. position.getBytes returns this exception. The description field is from position.toString() which shows a -1 for one of the segments.
```
event: "1234"
 position {
   description: "examples/memory-eater-in/1.#epoch.0=-1, examples/memory-eater-in/0.#epoch.0=12"
 }
 event_pointer {
   bytes: "\000\000\001\000\000\000\000\'\000#examples/memory-eater-in/0.#epoch.0\000\014"
   description: "examples/memory-eater-in/0.#epoch.0:0-12"
 }
 stream_cut {
   text: "H4sIAAAAAAAAADOwSq1IzC3ISS3Wz03NzS+q1E1NLEkt0s3MszLQMQRiAytDIx1dQwCG5bJbKAAAAA=="
   description: "examples/memory-eater-in:0=12, 1=-1"
 }
```

**Problem location**
Pravega Client 0.7.0

The serializer must use signed integers.

shrids:
**Change log description**  
Fix `io.pravega.client.stream.Position#toBytes` to ensure serialization works if the offset is -1L.

**Purpose of the change**  
Fixes #4669

**What the code does**  
The older revision of the client does not support serialization/deserialization with -1L as an offset.
This PR uses `io.pravega.common.io.serialization.RevisionDataOutput#writeCompactSignedLong` instead of `io.pravega.common.io.serialization.RevisionDataOutput#writeCompactLong` tofix the issue.

Both forward and backward compatibility is maintained with this change.

**How to verify it**  

### DIFF
--- a/client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
+++ b/client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
@@ -118,25 +118,34 @@ public class PositionImpl extends PositionInternal {
 
         @Override
         protected byte getWriteVersion() {
-            return 1;
+            return 0;
         }
 
         @Override
         protected void declareVersions() {
             version(0).revision(0, this::write00, this::read00)
-                      .revision(1, this::write01, this::read01);
-            version(1).revision(0, this::write10, this::read10);
+                      .revision(1, this::write01, this::read01)
+                      .revision(2, this::write02, this::read02);
         }
 
         private void read00(RevisionDataInput revisionDataInput, PositionBuilder builder) throws IOException {
-            Map<Segment, Long> map = revisionDataInput.readMap(in -> Segment.fromScopedName(in.readUTF()), RevisionDataInput::readCompactLong);
+            // RevisionDataInput::readCompactSignedLong can read data written using RevisionDataOutput.writeCompactLong
+            Map<Segment, Long> map = revisionDataInput.readMap(in -> Segment.fromScopedName(in.readUTF()), RevisionDataInput::readCompactSignedLong);
             builder.ownedSegments(map);
         }
 
         private void write00(PositionImpl position, RevisionDataOutput revisionDataOutput) throws IOException {
+            // This is used to ensure serialized data can be read by the older client.
+            // Note: the older client does not serialize if the offset has -1L as value.
             Map<Segment, Long> map = position.getOwnedSegmentsWithOffsets();
             revisionDataOutput.writeMap(map, (out, s) -> out.writeUTF(s.getScopedName()),
-                                        (out, offset) -> out.writeCompactLong(offset));
+                                        (out, offset) -> {
+                                            if ( offset < 0) {
+                                                out.writeCompactSignedLong(offset);
+                                            } else {
+                                                out.writeCompactLong(offset);
+                                            }
+                                        });
         }
         
         private void read01(RevisionDataInput revisionDataInput, PositionBuilder builder) throws IOException {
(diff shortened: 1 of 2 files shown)


client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -118,13 +118,14 @@ protected PositionBuilder newBuilder() {
 
         @Override
         protected byte getWriteVersion() {
-            return 0;
+            return 1;
```
andreipaduroiu, line 121: This will break backward compatibility. Old code does not know how to deal with version 1.

Do this using a revision.

shrids: This bug prevents serialization with the older version of the client for -1L.

I used a new version number since the serializing via `RevisionDataOutput#writeCompactSignedLong` (new change) cannot be deserialized via `RevisionedDataOutput#readCompactLong` (old change).

I have also added a test to verify if serialization via version 0 can be deserialized via version 1.

shrids: Done, I have updated the PR to ensure it is compatible both ways.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -147,7 +157,20 @@ private void write01(PositionImpl position, RevisionDataOutput revisionDataOutpu
             Map<Segment, Range> map = position.segmentRanges;
             revisionDataOutput.writeMap(map, (out, s) -> out.writeUTF(s.getScopedName()), PositionSerializer::writeRange);
         }
-        
+
+        private void write02(PositionImpl position, RevisionDataOutput revisionDataOutput) throws IOException {
+            // serialize offset segment again as CompactSignedLong ensure serialization
+            Map<Segment, Long> map = position.getOwnedSegmentsWithOffsets();
+            revisionDataOutput.writeMap(map, (out, s) -> out.writeUTF(s.getScopedName()),
+                    (out, offset) -> out.writeCompactSignedLong(offset));
```
andreipaduroiu, line 165: TIP: you can use a syntactic shortcut like `RevisionDataOutput::writeCompactSignedLong`

shrids:
Thanks @tkaitchuck , I realized that we can fix this issue without increasing the revision number and avoiding an increase in size of the serialized data. I have updated the PR with it.

PR #4674, opened 2020-04-07
Issue 4673: Bump up versions of some of the dependencies

### LINKED ISSUE #4673: Bump versions of dependencies
**Problem description**
We need to upgrade versions of some of our dependencies for v`0.7.1+` and `v0.8.x`, in order to address vulnerabilities founded by running container image scans. 

**Problem location**
`gradle.properties`

ravisharda:
**Change log description**  
Update dependency versions. 

**Purpose of the change**  
Resolves #4673.  

**What the code does**  
Updates versions for a few dependencies. 

* Updates Jackson version from `2.9.9.3` to `2.9.10.3`
* Updates Apache commons Compress version from `1.18` o `1.20` 
* Upgrades Guava to the latest version `28.2-jre`. 

**How to verify it**  
All the tests, including system tests, must pass.


gradle.properties
```diff
@@ -31,10 +31,14 @@ gradleLombokPluginVersion=3.2.0
 gradleMkdocsPluginVersion=1.0.0
 gradleSshPluginVersion=2.9.0
 grpcVersion=1.17.1
-guavaVersion=28.1-jre
-hadoopVersion=2.8.5
+
+# Downgraded the Guava version, to avoid version conflicts with transitiv dependencies of HDFS. Guava versions above
+# this one do not have `com.google.common.util.concurrent.Futures` `addCallback(ListenableFuture, FutureCallback)`
+# method, which is in turn used by HDFS server class `org.apache.hadoop.hdfs.server.datanode.checker.ThrottledAsyncChecker`.
+guavaVersion=25.1-jre
```
andreipaduroiu, line 38: This is not acceptable. We should never downgrade dependencies because we could be reintroducing bugs that were solved by upgrading in the first place. 

Please find another solution to this problem.

ravisharda: Yes, that makes sense. I've revert the change.

PR #4678, opened 2020-04-08
Issue 3994 : List all Segments in Tier 2

### LINKED ISSUE #3994: List all Segments in Tier 2
**Problem description**
In order to implement #3993, we need a way to list all the segments in Tier 2 so that they may be properly reconciled.

**Problem location**
Segment Store/Rolling Storage.

Provide an API of the form `Iterator<SegmentProperties> listSegments()` that will return all segments in Tier 2. The returned `SegmentProperties` objects should have a minimum of: Segment Name, Segment Length, Segment Sealed Status. 

Note that we do not need all files in Tier 2. RollingStorage keeps multiple files per segment in tier 2. We only need the whole segment info (i.e., one item per `$header`).

naveenkothamasu:
**Changelog description**
Added list-segments API to retrieve the segments from Tier 2.

**Purpose of the change**
Fixes #3994.

**What the code does**
Basic implementation of Iterator over the header files in the storage returning SegmentProperties per file.
The iterator logic is tailored to the specific underlying storage depending on how the listing of entries works in that implementation.
For retrieving StreamSegmentInformation for a particular header file, RollingSegmentHandle is used.

**What this doesn't do**
As of now, the API support is only verified for the FileSystemStorage type. For other storage types such as s3 and HDFS, it will be considered later.

**How to verify it**
Unit tests added to verify feature.

### DIFF
--- a/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
+++ b/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
@@ -17,8 +17,10 @@ import com.emc.object.s3.bean.AccessControlList;
 import com.emc.object.s3.bean.CanonicalUser;
 import com.emc.object.s3.bean.CopyPartResult;
 import com.emc.object.s3.bean.Grant;
+import com.emc.object.s3.bean.ListObjectsResult;
 import com.emc.object.s3.bean.MultipartPartETag;
 import com.emc.object.s3.bean.Permission;
+import com.emc.object.s3.bean.S3Object;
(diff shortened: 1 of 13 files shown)


naveenkothamasu:
Travis reported the following tests failed.
> Task :segmentstore:server:test FAILED
But it passes locally.
```
io.pravega.segmentstore.server.writer.StorageWriterTests > testWithStorageNotPrimaryErrors PASSED

BUILD SUCCESSFUL in 3m 41s
18 actionable tasks: 9 executed, 9 up-to-date
````

andreipaduroiu:
Replaced by #4710

PR #4682, opened 2020-04-09
Issue 4642 - Use Java truststore path and password mounted by operator.

### LINKED ISSUE #4642: Pravega to support self-signed certificate of S3 Storage service
**Problem description**
Now Pravega supports ECS as tier2. If an ECS internal installation is only self-signed for TLS, Pravega must trust its self-signed certificate in order to establish TLS connection.

**Problem location**
docker/common.sh
config.properties, standalone.properties

Scenario 1: deployment on K8S
1. Download the ECS self-signed certificate, e.g. ecs-certificate.pem)
2. Add the certificate into K8S secret manually 
  (e.g. kubectl create secret generic **ecs-cert** --from-file ./ecs-certificate.pem)
3. In Pravega.yaml:
tls:
  static:
    caBundle: "**ecs-cert**"
4. Operator mounts caBundle onto "**/etc/secret-volume/ca-bundle/**" in container

5. Segmentstore container, by its docker init scripts, adds everything under "**/etc/secret-volume/ca-bundle/**" into java keystore.

Scenario 2: without container, without K8S (e.g. stand-alone mode)
Download ECS self signed certificate and manually add it into java keystore $JAVA_HOME/jre/lib/security/cacerts

kevinhan88:
**Change log description**  
Instead of hard-coding, now the Java Truststore path and password get mounted from pravega manifest by pravega-operator.
This change is to start using the mounted path and password in the script to add ECS certificate into Java truststore.

**Purpose of the change**  
Fixes #4642 

**What the code does**  
Inside docker/pravega/scripts/common.sh:
Retrieve password from /etc/ssl/certs/java/ecs-certs/JAVA_TRUST_STORE_PASSWORD
Retrieve truststore file path from /etc/ssl/certs/java/ecs-certs/JAVA_TRUST_STORE_PATH
Then scan /etc/ssl/certs/java/ecs-certs to add all the found PEM files into Java Truststore.

**How to verify it**  
With https://... ECS configUri, Pravega segmentstore is able to write/read to/from ECS

### DIFF
--- a/docker/pravega/scripts/common.sh
+++ b/docker/pravega/scripts/common.sh
@@ -37,7 +37,7 @@ add_system_property_ecs_config_uri() {
 # Add ECS certificates into Java truststore
 add_certs_into_truststore() {
     password="changeit"
-    trust="/etc/ssl/certs/java/cacerts"
+    trustStore="/etc/ssl/certs/java/cacerts"
     certificate="ecs-certificate.pem"
 
     if test -f "/etc/secret-volume/ECS_JAVA_TRUST_STORE_PASSWORD"; then
--- a/docker/pravega/scripts/common.sh
+++ b/docker/pravega/scripts/common.sh
@@ -36,12 +36,21 @@ add_system_property_ecs_config_uri() {
 
 # Add ECS certificates into Java truststore
 add_certs_into_truststore() {
-    password=`cat /etc/ssl/certs/java/ecs-certs/JAVA_TRUST_STORE_PASSWORD`
-    trustStore=`cat /etc/ssl/certs/java/ecs-certs/JAVA_TRUST_STORE_PATH`
+    password="changeit"
+    trust="/etc/ssl/certs/java/cacerts"
+    certificate="ecs-certificate.pem"
 
-    CERTS=/etc/ssl/certs/java/ecs-certs/*.pem
-    for cert in $CERTS
-    do
-      yes | keytool -importcert -storepass "${password}" -file "${cert}" -keystore "${trustStore}" || true
-    done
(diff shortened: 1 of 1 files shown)


docker/pravega/scripts/common.sh
```diff
@@ -36,9 +36,12 @@ add_system_property_ecs_config_uri() {
 
 # Add ECS certificates into Java truststore
 add_certs_into_truststore() {
-    CERTS=/etc/ssl/certs/java/ecs-certs/*
+    password=`cat /etc/ssl/certs/java/ecs-certs/JAVA_TRUST_STORE_PASSWORD`
+    trustStore=`cat /etc/ssl/certs/java/ecs-certs/JAVA_TRUST_STORE_PATH`
```
medvedevigorek, line 40: Why do we allow to specify a path to the truststore if we anyway use the default truststore in Segment Store? 

As a user I can provide a different path that would be pointing to some custom keystore inside the container but that keystore wouldn't be loaded by SSS

ravisharda: We can use the Kubernetes secrets mechanism to get a user-specified truststore/password into the pod at deployment time. That's what we do for Controller and Segment store TLS material (see https://github.com/pravega/pravega-operator/blob/master/doc/tls.md)

kevinhan88: Yes, now the ECS TLS materials are to be prepared as k8s secret by admin, and the Pravega-Operator mounts the secret into secret-volume, in the same way we mount other TLS secrets.

The ECS TLS config will be transparent to ECS client side config.

kevinhan88:
Yes @ravisharda now the ECS TLS secret is mounted to the secret volume as well.

kevinhan88:
Per the latest discussion and decision made, this PR is no longer needed

PR #4692, opened 2020-04-15
Issue 4691: Honor delegation token expiration during appends 

### LINKED ISSUE #4691: Writes can continue despite delegation token expiry
**Problem description**
Since delegation tokens are verified only upon `SetupAppend`, writers can continue to write to a segment even if the delegation token has expired, until the connection is reset and `SetupAppend`  interaction occurs again. 

**Problem location**

ravisharda:
**Change log description**  
Prevent writes upon delegation token expiry and ensure that the client re-authenticates with the Controller. 

There were multiple ways of tackling this problem, each one with its own pros and cons as described below. This PR uses approach 5. 

|S.No.| Approach |	Pros  | Cons | 
|:----:|:-----------|:---------------:|:---------------:|
|1. | Perform delegation token expiry check upon each append on the server-side. | After the token expires, no write/append is allowed to go through. | Negative impact on append performance, due to additional cost of expiry checking. | 
|2.| Perform delegation token expiry check for each `AppendBlock` on the server-side | Lower impact on write performance than approach 1 in many practical scenarios. | Still, a significant negative impact on data append performance. |  
|3. | `AppendProcessor` closes the connections at periodic intervals regardless of whether a token has expired, in order to bound the period until which writes can continue despite token expiry. | Lower impact on write performance than the above approaches | Writes can continue after token expiry for a much longer duration than desired/expected.  |
|4.| On client-side, parse token expiry upon `SetupAppend` and store the expiration duration in the writer (SegmentOutputStreamImpl) state. This is so that parsing - an expensive operation - is done only once. Then, perform a "low-cost" token expiration check upon each write. | No additional performance costs on the server-side. | The check depends wholely on client-side validation, plus might impact client application performance. |
|5.| Run a task upon `SetupAppend` that closes the connection upon token expiration. | Token expiry takes effect soon after the event improving security. The impact on performance is minimal. | In some cases, the token expiry can occur slightly slower than expected. | 

**Purpose of the change**  
Resolves #4691 

**What the code does**  
Here are the main changes: 
* Adds a task in `AppendProcessor.setupAppend(..)` for closing the connection upon token expiry. This will ensure that the client will re-authenticate with the Controller and retry the request with a new delegation token. 
* Adds a `ScheduledExecutorService` instance in `PravegaConnectionListener` for running the token expiry handling task described above. 

Here are the ancillary changes:
* Extracts away token expiration parsing logic from a `client` class to a class (`JwtUtils`) in `common` subproject. Without this change, we'd have to use a client class in Segment Store `AppendProcessor`, which'd be odd. In the same class, an additional method was added to identify the "duration to expiry"  for tokens. 
* Relocates a couple of other Security utilities to the same package as the newly created `JwtUtils`. 
* Relocates a few security-related test classes too, to enable the changes made for this PR.  

**How to verify it**  
All unit and integration tests must succeed. In addition, systems tests must also succeed.


ravisharda:
Working on expanding the unit test coverage.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -142,6 +146,30 @@ public void setupAppend(SetupAppend setupAppend) {
...
+                        log.debug("Token has already expired");
+                        throw new TokenExpiredException("Token already expired");
+                    } else {
+                        Futures.delayedTask(() -> {
+                            if (writerStates.containsKey(Pair.of(newSegment, writer))) {
+                                // Closing the connection will result in client authenticating with Controller again
+                                // and retrying.
+                                log.debug("Closing client connection due to token expiry");
+                                try {
+                                    this.connection.close();
```
shrids, line 163: if connection pooling is enabled then all the sessions will be terminated. We should send a `WireCommands.AuthTokenCheckFailed` .

ravisharda: Sure, I've made the change in the latest commits that I just pushed.

build.gradle
```diff
@@ -337,7 +346,7 @@ project ('bindings') {
         compile project(':common')
         compile project(':segmentstore:storage')
         compile project(':shared:metrics')
-        testCompile project(':test:testcommon')
+        testCompile project(':test:testcommon')
```
andreipaduroiu, line 349: Revert this line

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -205,6 +153,7 @@ public boolean populateToken(String token) {
         } else {
             return this.delegationToken.compareAndSet(currentToken, new DelegationToken(token, extractExpirationTime(token)));
         }
+
```
andreipaduroiu, line 156: Revert this

common/src/main/java/io/pravega/common/security/JwtUtils.java
```diff
@@ -0,0 +1,81 @@
...
+
+                // JwtUtils
+
+                String[] expiryTimeFieldParts = matchedString.split(":");
+                if (expiryTimeFieldParts != null && expiryTimeFieldParts.length == 2) {
+                    try {
+                        result = Long.parseLong(expiryTimeFieldParts[1].trim());
+                    } catch (NumberFormatException e) {
+                        // ignore
+                        log.warn("Encountered this exception when parsing JWT body for expiration time: {}", e.getMessage());
```
andreipaduroiu, line 74: "Unable to parse JWT body for expiration time: {}"

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +172,33 @@ public void setupAppend(SetupAppend setupAppend) {
                 });
     }
 
+    @VisibleForTesting
+    void setupTokenExpiryTask(SetupAppend setupAppend, String newSegment, UUID writer) {
+        final Duration durationToExpiry = JwtUtils.durationToExpiry(setupAppend.getDelegationToken());
+
+        if (durationToExpiry != null) { // Can be null if token had no expiry set (for internal communications)
+            if (durationToExpiry.isNegative()) {
+                log.debug("Token has already expired");
```
andreipaduroiu, line 181: Include segment and writer id in the message. Both here and everywhere else you log or throw something.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +172,33 @@ public void setupAppend(SetupAppend setupAppend) {
...
+            } else {
+                Futures.delayedTask(() -> {
+                    if (writerStates.containsKey(Pair.of(newSegment, writer))) {
+                        try {
+                            log.debug("Informing the client about token expiry");
+                            connection.send(new WireCommands.AuthTokenCheckFailed(setupAppend.getRequestId(),
+                                    "Token expired",
+                                    WireCommands.AuthTokenCheckFailed.ErrorCode.TOKEN_EXPIRED));
+                        } catch (RuntimeException e) {
+                            log.warn("Encountered an exception", e);
```
andreipaduroiu, line 192: Better message

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -90,6 +92,10 @@
 
     private FileModificationMonitor tlsCertFileModificationMonitor; // used only if tls reload is enabled
 
+    // Used for running token expiry handling tasks.
+    private final ScheduledExecutorService tokenExpiryHandlerExecutor =
```
andreipaduroiu, line 96: Please do not create random executors. There are already executors created in the Segment Store process; use one of them. Or otherwise try to rework your code to not require an executor.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -280,5 +286,8 @@ public void close() {
         if (tlsCertFileModificationMonitor != null) {
             tlsCertFileModificationMonitor.stopMonitoring();
         }
+        if (tokenExpiryHandlerExecutor != null) {
+            tokenExpiryHandlerExecutor.shutdownNow();
```
andreipaduroiu, line 290: This is the wrong way to shut down an executor. See `ExecutorServiceHelpers.shutdown`. But anyway, see my comment above.

test/testcommon/src/main/java/io/pravega/test/common/JwtTestUtils.java
```diff
@@ -7,20 +7,24 @@
...
+package io.pravega.test.common;
 
 import com.google.gson.Gson;
 
 import java.util.Base64;
 
 public class JwtTestUtils {
 
-    static String createJwtBody(JwtBody jwt) {
+    public static String createJwtBody(JwtBody jwt) {
```
andreipaduroiu, line 18: Javadoc for this class and all its methods.

build.gradle
ravisharda: Done. Thanks for catching it.

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
ravisharda: Done.

common/src/main/java/io/pravega/common/security/JwtUtils.java
ravisharda: Yeah, I had missed that - had thought of changing it later. I have fixed it now.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: Sure, I've made this change.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: Done.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
ravisharda: Fixed.

test/testcommon/src/main/java/io/pravega/test/common/JwtTestUtils.java
ravisharda: Done.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
ravisharda: Ok. I have modified the code to use a preexisting executor that is passed from the `ServiceStarter`.

build.gradle
```diff
@@ -321,6 +322,14 @@ project('segmentstore:storage:impl') {
 }
 
 project ('bindings') {
+    configurations.all {
```
shrids, line 325: Can we move the resolution strategy to the global one in build.gradle, to enforce a common lib across all.
https://github.com/pravega/pravega/blob/729bc23b6370c2b9fb735e83ed496eec66a5991d/build.gradle#L109-L115

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +175,57 @@ public void setupAppend(SetupAppend setupAppend) {
...
+
+        if (expiryDuration == null) {
+            return CompletableFuture.completedFuture(null);
+        } else {
+            return Futures.delayedTask(() -> {
+                if (isSetupAppendCompleted(segment, writerId)) {
+                    try {
+                        log.debug("Informing writer {} that sent request {}, about token expiry for segment {}",
+                                writerId, requestId, segment);
+                        connection.send(new WireCommands.AuthTokenCheckFailed(setupAppend.getRequestId(),
```
shrids, line 194: Can we avoid sending this if the connection is closed? A connection closing should ensure all the pending tasks are cancelled.

ravisharda: The `ServerConnection` interface doesn't seem to have a way to check if the connection is closed. My understanding is that if the connection is closed, sending a message over it will result in an exception being thrown, which we are catching and ignoring here. 

What do you say? Am I missing something else?

build.gradle
ravisharda: Let me explain the rationale for the change a bit. 

We use Gson version 2.8.5 elsewhere, including in `:test:testcommon`. The `:bindings` project depends on both: 
* `:test:testcommon` sub-project, which brings in Gson `2.8.5` as transitive dependency. 
* HDFS and ECS libraries, which depend on/bring Gson `2.5` as transitive dependency. 

Since the two Gson versions are not compatible with each other, we run into version conflicts here. So, I had to force the use of the older version in the `:bindings` sub-project. But, we don't want to use the older version of Gson everywhere, that is why the config change is local to the `:bindings` project. If/when we upgrade HDFS and ECS dependencies, we might be able to remove this change. 

So, I think the suggested change might not be required. What do you say?

shrids: Thanks @ravisharda , yeah we would end up downgrading gson version that is used by test common to avoid this. As discussed, this would change once we upgrade hdfs dependency...

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
shrids: The connection can be closed externally or when due to a connection.close() in by the appendProcessor causing these delayed tasks to be unnecessarily executed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +175,57 @@ public void setupAppend(SetupAppend setupAppend) {
...
+                                String.format("Token sent by writer %s for segment %s in request %s has expired",
+                                        writerId, segment, requestId),
+                                WireCommands.AuthTokenCheckFailed.ErrorCode.TOKEN_EXPIRED));
+                    } catch (RuntimeException e) {
+                        // Log and ignore
+                        log.warn("Unable to inform writer {} that sent request {}, about token expiry for segment {}",
```
shrids, line 200: I think we should have a lower log level here.

build.gradle
ravisharda: yeah, I've added a comment in the build file towards this end.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: Sure, that sounds good. I've reduced the log level.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: I've now introduced an `isClosed()` method to the connection, so that we the task checks for it before trying to send the token expiry notification. This makes it a bit more efficient. However, the core issue remains: whether/when/how do we cancel the task?

common/src/main/java/io/pravega/common/security/JwtUtils.java
```diff
@@ -0,0 +1,100 @@
...
+     *         a) the specified token is blank, or
+     *         b) the specified token is of invalid format, or
+     *         c) expiration time is missing from the token, or
+     *         d) expiration time is not a number.
+     */
+    public static Long extractExpirationTime(String jsonWebToken) {
+        if (jsonWebToken == null || jsonWebToken.trim().equals("")) {
+            return null;
+        }
+        String[] tokenParts = jsonWebToken.split("\\.");
```
andreipaduroiu, line 50: This is an expensive string parse. Can you just look up `exp` and get the number that succeeds it?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +175,59 @@ public void setupAppend(SetupAppend setupAppend) {
...
+                    try {
+                        log.trace("Informing writer {} that sent request {}, about token expiry for segment {}",
+                                writerId, requestId, segment);
+                        if (!connection.isClosed()) {
+                            connection.send(new WireCommands.AuthTokenCheckFailed(setupAppend.getRequestId(),
+                                    String.format("Token sent by writer %s for segment %s in request %s has expired",
+                                            writerId, segment, requestId),
+                                    WireCommands.AuthTokenCheckFailed.ErrorCode.TOKEN_EXPIRED));
+                        }
+                    } catch (RuntimeException e) {
```
andreipaduroiu, line 200: RuntimeException extends Exception. Please catch Exception.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -108,7 +113,9 @@ public static AppendProcessorBuilder defaultBuilder() {
                 .nextRequestProcessor(new FailingRequestProcessor())
                 .statsRecorder(SegmentStatsRecorder.noOp())
                 .connectionTracker(new ConnectionTracker())
-                .replyWithStackTraceOnError(false);
+                .replyWithStackTraceOnError(false)
+                .tokenExpiryHandlerExecutor(ExecutorServiceHelpers.newScheduledThreadPool(2,
```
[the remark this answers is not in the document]
andreipaduroiu: Also:
1. This executor has a weird name.
2. It is never shut down.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -105,7 +110,17 @@
     @VisibleForTesting
     public PravegaConnectionListener(boolean enableTls, int port, StreamSegmentStore streamSegmentStore, TableStore tableStore) {
         this(enableTls, false, "localhost", port, streamSegmentStore, tableStore, SegmentStatsRecorder.noOp(), TableSegmentStatsRecorder.noOp(),
-                new PassingTokenVerifier(), null, null, true);
+                new PassingTokenVerifier(), null, null, true,
+                ExecutorServiceHelpers.newScheduledThreadPool(1, "test-token-expiry-handler"));
```
andreipaduroiu, line 114: Who manages this executor?

build.gradle
```diff
@@ -321,6 +322,17 @@ project('segmentstore:storage:impl') {
...
+        resolutionStrategy {
```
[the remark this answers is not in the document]
ravisharda: Line 110 is in the `allProjects`  block in the Gradle file, and so applies to all sub-projects, as well as the root project. The `configurations.all` one we have here is applicable to this specific sub-project (`bindings`). I think that keeping it here along-side the rest of the sub-project makes this part of the code more cohesive. Besides, I don't think there is a clean way of putting this configuration up in the `allProjects` block and still apply to this specific sub-project.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: I want to be specific here, as the intention is only to catch any runtime exceptions.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
ravisharda: On second thought, I realized that it'd be better to run the token expiry tasks on a separate executor spinning up lower priority threads, just so that these tasks don't end up overwhelming the core executor. So, I have now modified the code again. 

Please take a look again and see what you think.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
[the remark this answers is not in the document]
ravisharda: Yeah, I've removed it altogether in the latest push.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +175,59 @@ public void setupAppend(SetupAppend setupAppend) {
...
+                                writerId, requestId, segment);
+                        if (!connection.isClosed()) {
+                            connection.send(new WireCommands.AuthTokenCheckFailed(setupAppend.getRequestId(),
+                                    String.format("Token sent by writer %s for segment %s in request %s has expired",
+                                            writerId, segment, requestId),
+                                    WireCommands.AuthTokenCheckFailed.ErrorCode.TOKEN_EXPIRED));
+                        }
+                    } catch (RuntimeException e) {
+                        // Log and ignore
+                        log.debug("Unable to inform writer {} that sent request {}, about token expiry for segment {}",
```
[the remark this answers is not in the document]
ravisharda: Yeah, the log level was changed from `warn` to `debug` previously, as the exception would also occur if the task ran after the connection was closed. Now that the main portion of the task is only run if the connection is not already closed, it'd be appropriate to change this back to `warn` level. I've made this change in the latest push.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -199,13 +217,10 @@ public void initChannel(SocketChannel ch) {
...
                          lsh);
 
-                 lsh.setRequestProcessor(new AppendProcessor(store,
-                         lsh,
-                         connectionTracker,
-                         new PravegaRequestProcessor(store, tableStore, lsh, statsRecorder, tableStatsRecorder, tokenVerifier, replyWithStackTraceOnError),
-                         statsRecorder,
-                         tokenVerifier,
-                         replyWithStackTraceOnError));
+                 lsh.setRequestProcessor(new AppendProcessor(store, lsh, connectionTracker,
```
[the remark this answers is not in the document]
ravisharda: Yeah, this is unnecessary. I've reverted it.

common/src/main/java/io/pravega/common/security/JwtUtils.java
ravisharda: It is a pre-existing pice code that was refactored into its own class, and therefore unrelated to the issue at hand. I can raise a separate issue, for this potential performance improvement.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +175,59 @@ public void setupAppend(SetupAppend setupAppend) {
...
+        if (expiryDuration == null) {
+            return CompletableFuture.completedFuture(null);
+        } else {
+            return Futures.delayedTask(() -> {
+                if (isSetupAppendCompleted(segment, writerId)) {
+                    try {
+                        log.trace("Informing writer {} that sent request {}, about token expiry for segment {}",
+                                writerId, requestId, segment);
+                        if (!connection.isClosed()) {
+                            connection.send(new WireCommands.AuthTokenCheckFailed(setupAppend.getRequestId(),
```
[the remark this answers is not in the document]
ravisharda: Yes, that is right. We are depending on the client to meet its contract related to token expiry and reauthenticate. One alternative would have been to close the connection, but that'd be too expensive I suppose.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
ravisharda: Even though the preexisting defaultBuilder() method where this executor was added, was only meant for testing, it was inappropriate to add the executor here. So, I've removed it altogether in the latest push. Please review the revised code.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceHelpers.java
```diff
@@ -66,6 +66,17 @@ public void uncaughtException(Thread t, Throwable e) {
      * @return a thread factory
      */
     public static ThreadFactory getThreadFactory(String groupName) {
+        return getThreadFactory(groupName, null);
```
shrids, line 69: Default priority can be `java.lang.Thread#NORM_PRIORITY`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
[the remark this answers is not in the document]
ravisharda: I can maintain a flag in the `WriteState` such that when the token expires, the task will set it to true. Then, I can check the flag on each append. However, that can introduce additional performance cost for each append. I believe the current solution is adequate and provides a good balance of security and performance.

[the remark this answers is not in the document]
tkaitchuck: This is a question of authentication. You can't rely on the client, that saying "We enforce authorization by asking nicely". The connection needs to be closed.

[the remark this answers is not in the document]
shrids: If connection pooling is enabled then closing the connection would cause all the clients using this connection to see a connection drop. But since it is a question of authentication the other clients would see a connection drop and would re-establish a connection with the SSS.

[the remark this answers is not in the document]
ravisharda: I've modified the code to close the connection, as suggested.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceHelpers.java
ravisharda: Done.

common/src/main/java/io/pravega/common/security/JwtUtils.java
ravisharda: Actually, we do look up the expiry string (for example: `"exp": 1569837434`) in another method `parseExpirationTime()`, after first extracting the JWT body and decoding it here.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceHelpers.java
```diff
@@ -86,8 +100,24 @@ public Thread newThread(Runnable r) {
...
+     * Creates a new ScheduledExecutorService that will use daemon threads with specified priority and names.
+     *
+     * @param size The number of threads in the threadpool
+     * @param poolName The name of the pool (this will be printed in logs)
+     * @param threadPriority The priority to be assigned to the threads
+     * @return A new executor service.
+     */
+    public static ScheduledExecutorService newScheduledThreadPool(int size, String poolName, Integer threadPriority) {
+
+        ThreadFactory threadFactory =
```
shrids, line 116: can we remove this redundant null check? The user would invoke `newScheduledThreadPool(int size, String poolName)` if the priority is default.

ravisharda: Yeah, but since it is a public method, we can't assume that's always true. Besides, having the check makes the intention explicit.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +172,56 @@ public void setupAppend(SetupAppend setupAppend) {
...
+                if (isSetupAppendCompleted(segment, writerId)) {
+                    try {
+                        log.trace("Informing writer {} that sent request {}, about token expiry for segment {}",
+                                writerId, requestId, segment);
+                        if (!connection.isClosed()) {
+                            connection.close();
+                        }
+                    } catch (RuntimeException e) {
+                        // Log and ignore
+                        log.warn("Unable to inform writer {} that sent request {}, about token expiry for segment {}",
```
shrids, line 196: We can improve/remove the warn message here since we are closing the connection and not informing the writer via a Wirecommand.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +172,56 @@ public void setupAppend(SetupAppend setupAppend) {
...
+
+        if (expiryDuration == null) {
+            return CompletableFuture.completedFuture(null);
+        } else {
+            return Futures.delayedTask(() -> {
+                if (isSetupAppendCompleted(segment, writerId)) {
+                    try {
+                        log.trace("Informing writer {} that sent request {}, about token expiry for segment {}",
+                                writerId, requestId, segment);
+                        if (!connection.isClosed()) {
```
shrids, line 191: we do not need this check since we are closing the connection.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +172,56 @@ public void setupAppend(SetupAppend setupAppend) {
...
+                            connection.close();
+                        }
+                    } catch (RuntimeException e) {
```
shrids, line 194: I did not understand the need to catch an exception for `connection.close()`? 
(if it is required, then this pattern should be replicated at all connection.close() invocation instances. )

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +172,56 @@ public void setupAppend(SetupAppend setupAppend) {
...
+        }
+    }
+
+    @VisibleForTesting
+    Duration durationToExpiry(SetupAppend setupAppend) {
+        final Duration duration = JwtUtils.durationToExpiry(setupAppend.getDelegationToken());
+
+        // Note that duration can be null, say if token had no expiry set (for internal communications).
+
+        if (duration != null && duration.isNegative()) {
```
shrids, line 211: Would the `TokenVerifierImpl#verifyToken` verify if the duration is negative? If not, we could move the check there to ensure expired tokens are not allowed.

build.gradle
```diff
@@ -322,6 +323,17 @@ project('segmentstore:storage:impl') {
...
+            // We need to force this version as dependency, since ':test:testcommon' brings in a newer version of
```
tkaitchuck, line 328: This comment makes no sense. If there is a conflict with testcommon why is the `bindings` version being changed? This fix should go into testcommon. Because testcommon is not depended on by any non-test target there should be no conflict.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -105,7 +110,17 @@
...
+                ExecutorServiceHelpers.newScheduledThreadPool(1, "test-token-expiry-handler"));
+    }
+
+    public PravegaConnectionListener(boolean enableTls, boolean enableTlsReload, String host, int port, StreamSegmentStore streamSegmentStore, TableStore tableStore,
+                                     SegmentStatsRecorder statsRecorder, TableSegmentStatsRecorder tableStatsRecorder,
+                                     DelegationTokenVerifier tokenVerifier, String certFile, String keyFile,
+                                     boolean replyWithStackTraceOnError) {
+        this(enableTls, enableTlsReload, host, port, streamSegmentStore, tableStore, statsRecorder, tableStatsRecorder,
+                tokenVerifier, certFile, keyFile, replyWithStackTraceOnError,
+                ExecutorServiceHelpers.newScheduledThreadPool(1, "test-token-expiry-handler"));
```
tkaitchuck, line 123: In this code path, this executor does not look like it will ever be shutdown. That is a thread leak.

test/testcommon/src/main/java/io/pravega/test/common/JwtBody.java
```diff
@@ -20,7 +20,7 @@
 @Builder
 @Getter
 @Setter
-class JwtBody {
+public class JwtBody {
```
tkaitchuck, line 23: This is such a small amount of json, and it doesn't even have nesting. Rather than dealing with dependency managment of Gson can we just use a Format template?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
```diff
@@ -66,13 +66,15 @@
...
-    private static final Duration SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
+    private static final Duration SHUTDOWN_TIMEOUT = Duration.ofSeconds(45);
     private final SegmentStoreMetrics.ThreadPool threadPoolMetrics;
     private final SegmentToContainerMapper segmentToContainerMapper;
     private final ServiceBuilderConfig serviceBuilderConfig;
     @Getter(AccessLevel.PROTECTED)
     private final ScheduledExecutorService coreExecutor;
     private final ScheduledExecutorService storageExecutor;
+    @Getter(AccessLevel.PUBLIC)
+    private final ScheduledExecutorService lowPriorityExecutor;
```
tkaitchuck, line 77: This doesn't appear to be used anywhere.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: IMO, this check is still useful because this task is a background task and the connection might already have closed out-of-band.

build.gradle
ravisharda: The problem really is that the dependencies `HDFS` and `S3` of the `bindings` sub-project bring in an older version of Gson `2.5` (2015) transitively, while the `client` and `testcommon` sub-projects depend upon a newer version `2.8.5` (2018). I don't think it's a good idea to downgrade the Gson dependency used by `client` and `testcommon` subprojects, just because the old versions of `HDFS` and `S3` we use in turn use a rather old version of `Gson`. Also, we already have plans to upgrade HDFS and S3 to newer versions, which should allow us to remove this config. So, IMO this change needs to stay.

What do you say?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: Yes, that check is for finding out if the token has already expired.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: We don't need to catch the exception anymore after we modified the code here to close the connection, so I've removed the code.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: We don't need to catch the exception anymore after we modified the code here to close the connection, so I've removed the code.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
ravisharda: That constructor was being used in unit tests. I've modified it to inject the executor from outside, and have modified the tests.

test/testcommon/src/main/java/io/pravega/test/common/JwtBody.java
ravisharda: This is an existing piece of code and I think we should explore seeing if that'd make sense and do the change separately of this PR.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
ravisharda: We were using it in `ServiceStarter`. Now, we also use it elsewhere as you can see in the latest push.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/ServerConnectionInboundHandler.java
```diff
@@ -84,6 +87,12 @@ public void close() {
             // wait for all messages to be sent before closing the channel.
             ch.eventLoop().execute(() -> ch.writeAndFlush(Unpooled.EMPTY_BUFFER).addListener(ChannelFutureListener.CLOSE));
         }
+        isClosed.compareAndSet(false, true);
```
shrids, line 90: can we ensure close() is invoked only once? i.e 
```
if (isClosed.compareAndSet(false, true) ) {
   Channel ch = ...
}
```

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
shrids: TokenVerifierImpl#verifyToken is already checking for TokenExpired. Hence this check seems redundant.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/ServerConnectionInboundHandler.java
ravisharda: Ok, I've moved the check as suggested.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: I've moved this check inside close itself, and have removed it here.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: I've now removed it.

common/src/main/java/io/pravega/common/security/JwtUtils.java
ravisharda: This PR now doesn't use this method anymore. It continues to be used elsewhere.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/DelegationTokenVerifier.java
```diff
@@ -33,7 +34,8 @@
      * @throws InvalidClaimException  if the {@code token} does not contain the claim representing
      *                                {@code expectedLevel} of access
      * @throws TokenException         if any other failure condition is encountered
+     * @return JsonWebToken           a non-null value if token was parsed and verified successfully, otherwise null.
      */
-    void verifyToken(String resource, String token, AuthHandler.Permissions expectedLevel)
+    JsonWebToken verifyToken(String resource, String token, AuthHandler.Permissions expectedLevel)
```
shrids, line 39: +1

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -328,8 +361,7 @@ private void handleException(UUID writerId, long requestId, String segment, long
             connection.close();
         } else if (u instanceof TokenExpiredException) {
             log.warn(requestId, "Token expired for writer {} on segment {}.", writerId, segment, u);
```
shrids, line 363: Should this be bumped to ERROR log level ?

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
```diff
@@ -107,6 +104,19 @@ public JsonWebToken(@NonNull String subject, @NonNull String audience, @NonNull
...
+                        Date expiry, Map<String, Object> resourcePermissionClaims) {
+        this.subject = subject;
+        this.audience = audience;
+        this.signingKey = signingKey.clone();
+        if (expiry != null) {
+            this.expirationTime = Date.from(expiry.toInstant());
+        } else {
+            this.expirationTime = null;
+        }
+        this.permissionsByResource = resourcePermissionClaims;
```
tkaitchuck, line 117: Should this have a defensive copy?

build.gradle
tkaitchuck: Now that you have changed the code both :client and :test:testcommon can remove the dependency all together.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/AppendProcessorTest.java
```diff
@@ -124,6 +136,101 @@ public void testAppend() {
...
+        // Spy the actual Append Processor, so that we can have some of the methods return stubbed values.
+        AppendProcessor mockProcessor = spy(processor);
+        doReturn(true).when(mockProcessor).isSetupAppendCompleted(streamSegmentName, clientId);
+        // doReturn(Duration.ofMillis(20)).when(mockProcessor);
+
+        JsonWebToken token = new JsonWebToken("subject", "audience", "secret".getBytes(),
+                Date.from(Instant.now().plusMillis(100)), null);
+        SetupAppend setupAppend = new SetupAppend(1, clientId, streamSegmentName, token.toCompactString());
+
+        Thread.sleep(100);
```
andreipaduroiu, line 197: Please do everything you can to avoid sleeping arbitrary amounts of time in unit tests. This just adds unnecessary delays.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
```diff
@@ -66,13 +66,15 @@
 public class ServiceBuilder implements AutoCloseable {
     //region Members
 
-    private static final Duration SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
+    private static final Duration SHUTDOWN_TIMEOUT = Duration.ofSeconds(45);
```
andreipaduroiu, line 69: Why?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
```diff
@@ -122,8 +124,10 @@ private ServiceBuilder(ServiceBuilderConfig serviceBuilderConfig, ServiceConfig
 
         // Setup Thread Pools.
         String instancePrefix = getInstanceIdPrefix(serviceConfig);
-        this.coreExecutor = executorBuilder.apply(serviceConfig.getCoreThreadPoolSize(), instancePrefix + "core");
-        this.storageExecutor = executorBuilder.apply(serviceConfig.getStorageThreadPoolSize(), instancePrefix + "storage-io");
+        this.coreExecutor = executorBuilder.apply(serviceConfig.getCoreThreadPoolSize(), instancePrefix + "core", Thread.NORM_PRIORITY);
+        this.storageExecutor = executorBuilder.apply(serviceConfig.getStorageThreadPoolSize(), instancePrefix + "storage-io", Thread.NORM_PRIORITY);
+        this.lowPriorityExecutor =
```
andreipaduroiu, line 129: What does 2 mean here? Above you use `Thread.NORM_PRIORITY`. Is there another constant you can use?

andreipaduroiu: Also do not hardcode the number of threads. Please add an entry in `ServiceConfig`, default it to 10, so that we may change it at any time if we encounter problems.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupTest.java
```diff
@@ -67,7 +67,7 @@ public void run() {
         }
     }
 
-    @Test(timeout = 20000)
+    @Test//(timeout = 20000)
```
andreipaduroiu, line 70: Undo

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public void shutdown() {
+
+    }
+
+    @Override
+    public List<Runnable> shutdownNow() {
+        return null;
```
andreipaduroiu, line 61: Does this return value match the contract of `Executor.shutdownNow`?

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public boolean isShutdown() {
+        return false;
```
andreipaduroiu, line 66: Should return true if you invoked `shutdown`

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public boolean isTerminated() {
+        return false;
+    }
+
+    @Override
+    public boolean awaitTermination(long l, TimeUnit timeUnit) throws InterruptedException {
+        return false;
```
andreipaduroiu, line 76: Should return true if terminated.

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public <T> Future<T> submit(Callable<T> callable) {
+        return null;
```
andreipaduroiu, line 81: same comment as above.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -167,6 +173,33 @@ public void setupAppend(SetupAppend setupAppend) {
                 });
     }
 
+    @VisibleForTesting
+    CompletableFuture<Void> setupTokenExpiryTask(@NonNull SetupAppend setupAppend, @NonNull JsonWebToken token) {
```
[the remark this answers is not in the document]
ravisharda: A couple of tests use the future to wait for task completion before making the assertion.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
ravisharda: This is to give some allowance as we will also be closing the new `lowPriorityExecutor` in addition to the two preexisting executors `coreExecutor` and `storageExecutor`. I just made a simplistic assumption here that if we give two executors `30` seconds, then we should give three of them `45`.

build.gradle
ravisharda: Yeah, that makes sense. I've removed this configuration section now and the dependency to `testcommon`.

shared/security/src/main/java/io/pravega/shared/security/token/JsonWebToken.java
ravisharda: The class is used internally. I had to make a defensive copy of expiration time and bytes to satisfy Spotbugs.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/AppendProcessorTest.java
ravisharda: Yeah, I agree with you fully. This was supposed to be a stop-gap, until I found a better solution, but I pushed the change by mistake. In any case, the solution turned out to be quite simple, and I have pushed the change in the latest commit.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
ravisharda: `ERROR` would be too harsh here, as the system continues to function normally.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupTest.java
ravisharda: Fixed.

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
ravisharda: Yes. I've modified the code to return an empty list.

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private static final NoOpScheduledExecutor INSTANCE = new NoOpScheduledExecutor();
+
+    public static NoOpScheduledExecutor get() {
+        return INSTANCE;
+    }
+
+    @Override
+    public ScheduledFuture<?> schedule(Runnable runnable, long l, TimeUnit timeUnit) {
+        return null;
```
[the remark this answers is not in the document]
ravisharda: Sure, I've modified the code return a dummy scheduled future.

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
ravisharda: Done.

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
ravisharda: Fixed.

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
ravisharda: Fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
ravisharda: Sure, done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
```diff
@@ -126,8 +126,8 @@ private ServiceBuilder(ServiceBuilderConfig serviceBuilderConfig, ServiceConfig
         String instancePrefix = getInstanceIdPrefix(serviceConfig);
         this.coreExecutor = executorBuilder.apply(serviceConfig.getCoreThreadPoolSize(), instancePrefix + "core", Thread.NORM_PRIORITY);
         this.storageExecutor = executorBuilder.apply(serviceConfig.getStorageThreadPoolSize(), instancePrefix + "storage-io", Thread.NORM_PRIORITY);
-        this.lowPriorityExecutor =
-                executorBuilder.apply(10, instancePrefix + "low-priority-cleanup", 2);
+        this.lowPriorityExecutor = executorBuilder.apply(serviceConfig.getLowPriorityThreadPoolSize(),
+                instancePrefix + "low-priority-cleanup", 2);
```
andreipaduroiu, line 130: Why can't you use `Thread.MIN_PRIORITY`?

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
```diff
@@ -9,110 +9,173 @@
...
     }
 
     @Override
     public void execute(Runnable runnable) {
+    }
+}
+
+@EqualsAndHashCode // Required to keep Spotbugs satisfied (owing to code in `compareTo()`).
+@RequiredArgsConstructor
+class DummyScheduledFuture implements ScheduledFuture<Integer> {
```
andreipaduroiu, line 131: Hmm.. I wonder how come checkstyle didn't catch this. We shouldn't define more than one class/interface per file. 

You should move this as a static nested class inside the `NoOpScheduledExecutor` class.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceBuilder.java
ravisharda: Sure, done.

test/testcommon/src/main/java/io/pravega/test/common/NoOpScheduledExecutor.java
ravisharda: Yeah, that sounds good. Done.

PR #4696, opened 2020-04-15
Issue 4695: Improved batch size calculation algorithm.

### LINKED ISSUE #4695: Improve AppendBatchSizeTracker algorithm
**Problem description**
We have an intelligent batching algorithm that attempts to work without any configuration. It appears to work well in most cases, but does not handle very small events at high rates super effectively.

**Problem location**
AppendBatchSizeTracker

Find a way to tune the algorithm to improve performance for small events without sacrificing it in other cases.

tkaitchuck:

**Change log description**  
Tune the Append batch size tracker's algorithm for calculating batch size.

**Purpose of the change**  
Fixes #4695 

**What the code does**  
Averages two different algorithms. One is based on time with a target of 7ms, and the other is based on the appends outstanding (as it did before). Accuracy is also improved through measuring in nanos rather than millis, and samples are weighted exponentially. The averages are also calculated over a longer period of time to avoid oscillations.

**How to verify it**
We will need several rounds of experimental tests to verify the new parameters.

### DIFF
--- a/client/src/main/java/io/pravega/client/netty/impl/AppendBatchSizeTrackerImpl.java
+++ b/client/src/main/java/io/pravega/client/netty/impl/AppendBatchSizeTrackerImpl.java
@@ -29,17 +29,21 @@ import java.util.function.Supplier;
  */
 class AppendBatchSizeTrackerImpl implements AppendBatchSizeTracker {
     private static final int MAX_BATCH_TIME_MILLIS = 100;
-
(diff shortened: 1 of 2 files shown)


RaulGracia:
@tkaitchuck I have tested this branch and latency is significantly worse compared to the original version of the algorithm. Maybe it will require further debugging to understand why.

PR #4686, opened 2020-04-10
Issue 4676: (PDP-34 )  Part 1 of 4. - ChunkedSegmentStorage and BaseChunkMetadataStore implementation

sachin-j-joshi:
Issue 4676: (PDP-34) Initial Implementation (Part 1 of 4) - Core functionality.
Signed-off-by: Sachin Joshi <sachin.joshi@emc.com>

**Change log description**  

- Initial implementation of PDP 34. 
- The Storage interface is implemented using new design using ChunkedSegmentStorage, ChunkMetadataStore, ChunkStorage. The implementation re-implements RollingStorage functionality.
- For this increment, both chunk metadata store and chunk storage providers are in memory based.
- In addition to new new unit tests, this code is tested using refactored storage tests.

**Purpose of the change**  
Fixes #4676 - Initial implementation of PDP -34

**What the code does**  
_Design :_ 
- This PR implements core parts of following design.  https://github.com/pravega/pravega/wiki/PDP-34:-Simplified-Tier-2

_General Comments:_ 
- The code is added in a manner so that older AsyncStorage+Sync based code continues to work. Reusing and refactoring code wherever possible.
- The ChunkStorage interface is synchronous for simplicity of implementation. We'll change it to Async if required at later date.

_Part 1 of 4 PRs :_ 

1.  **This PR**
2.  ChunkStorage implementation for HDFS, ECS and FileSystem
3.  Table segment based metadata store,  container bootstrap logic. end to end integration tests.
4. Changes for configuration, tests and other incidental changes

_In scope :_ The initial implementation includes following.
- Implementation of Storage interface using new design ( viz. ChunkedSegmentStorage + ChunkMetadataStore + ChunkStorage). The implementation should also re-implement RollingStorage functionality.
- Implementations of No Op and InMemory ChunkStorage for testing.
 

_out of scope :_ 
- This initial implementation does not implement actual provider that supports only immutable writes. (Although it should include design elements to support that.)
- The initial implementation does not support upgrade path.
- Non-critical performance improvements. (Especially on read path. Coming very very soon)

**How to verify it**  
- [ ] New or old unit tests should pass.
- [ ] System tests should pass. 

**Suggested Sequence for review**  

1. Interfaces `ChunkStorage`, `ChunkMetadataStore` 
2. ChunkStorage Implementations - Base class `BaseChunkStorage`.  For testing `NoOpChunkStorage` and `InMemoryChunkStorage`. 
3. The core of the Implementation . `ChunkedSegmentStorage` .
[...]

PR #4699, opened 2020-04-16
Issue 4676: (PDP-34) Part 2 of 4 - Implementation of ChunkStorage for HDFS, FileSystem and ExtendedS3.

### LINKED ISSUE #4676: (PDP-34) Initial Implementation for Simplified Tier-2.
**Problem description**
This issue is for tracking initial implementation of Simplified Tier-2 (PDP-34).

**Problem location**
Segment store tier-2.

_General Comments:_ 

- The code should be added in a manner so that older AsyncStorage+Sync based code continues to work.
- Reuse and refactor code wherever possible.

_In scope :_ The initial implementation should include following.
- Implementation of Storage interface using new design ( viz. ChunkStorageManager + ChunkMetadataStore + ChunkStorageProvider ). The implementation should also re-implement RollingStorage functionality.
- Implementation for ChunkStorageProvider using NFS, HDFS and ExtendedS3
- Implementation of No Op and InMemory ChunkStorageProviders for testing.
- Initial implementation should use TableSegment based implementation for storing storage metadata.
- This implementation should also update container bootstrap logic so that system boots up correctly.
- New configuration options should be provided so that this new experimental feature can be turned on in Pravega deployments.
- The end to end StreamSegmentStore tests should pass.
 
_out of scope :_ 
- This initial implementation need not implement actual provider that supports only immutable writes. (Although it should include design elements to support that.)
- The initial implementation need not support upgrade path.
- Non-critical performance improvements.

**Related Issues**
#4128

**Links**

- The design document is here https://github.com/pravega/pravega/wiki/PDP-34:-Simplified-Tier-2

sachin-j-joshi:
NOTE  This PR depends on https://github.com/pravega/pravega/pull/4686 to be completed first.

Issue 4676: (PDP-34) Initial Implementation (Part 2 of 4) - ChunkStorage for FileSystem, HDFS and ExtendedS3.

**Change log description**  

- Initial implementation of PDP 34. 
- ChunkStorage for FileSystem, HDFS and ExtendedS3. Minor code reuse and refactoring.

**Purpose of the change**  
Fixes #4676 - Initial implementation of PDP -34

**What the code does**  
_Design :_ 
- This PR implements core parts of following design.  https://github.com/pravega/pravega/wiki/PDP-34:-Simplified-Tier-2

_General Comments:_ 
- The code is added in a manner so that older AsyncStorage+Sync based code continues to work. Reusing and refactoring code wherever possible.
- The ChunkStorage interface is synchronous for simplicity of implementation. We'll change it to Async if required at later date.

_Part 1 of 4 PRs :_ 

1.  Core interfaces and core functionality of ChunkedSegmentStorage.
2.  ****This PR** _ `ChunkStorage` implementation for HDFS, ECS and FileSystem_**
3.  Table segment based metadata store,  container bootstrap logic. end to end integration tests.
4. Changes for configuration, tests and other incidental changes

**How to verify it**  
- [ ] New or old unit tests should pass.
- [ ] System tests should pass.


bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,440 @@
...
+
+            //Close the upload
+            client.completeMultipartUpload(new CompleteMultipartUploadRequest(config.getBucket(),
+                    targetPath, uploadId).withParts(partEtags));
+            // Delete all source objects.
+            for (int i = 1; i < chunks.length; i++) {
+                client.deleteObject(config.getBucket(), config.getPrefix() + chunks[i].getName());
+            }
+        } catch (RuntimeException e) {
+            throw e; // make spotbugs happy
```
fpj, line 256: Is spotbugs really complaining about this? We don't do it in other parts of the code and it doesn't complain, why does it complain here, do you know?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    /**
+     * Gets a value indicating whether this Storage implementation supports truncate operation on underlying storage object.
+     *
+     * @return True or false.
+     */
+    @Override
+    public boolean supportsTruncation() {
+        return false;
```
fpj, line 103: Doesn't `FileChannel` have a truncation operation?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,431 @@
...
+    protected long getFileSize(Path path) throws IOException {
+        try {
+            return Files.size(path);
+        } catch (NoSuchFileException e) {
+            throw new FileNotFoundException(path.toString());
+        }
+    }
+
+    /**
+     * Retrieves the ChunkInfo for given name.
```
fpj, line 129: In general, the value I see in javadocs for overridden methods is the additional information about the specifics of the implementation. Otherwise, just copying from parent class or interface doesn't add much. I'm in favor of adding comments to reflect what is relevant for the particular implementation.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /**
+     * Gets a value indicating whether this Storage implementation supports truncate operation on underlying storage object.
+     *
+     * @return True or false.
+     */
+    @Override
+    public boolean supportsTruncation() {
+        return false;
```
fpj, line 108: I also thought that HDFS had a truncate operation, is there a reason for us to say that it doesn't support truncate?

https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/FileSystemShell.html#truncate

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,473 @@
...
+        try {
+            Path fullPath = getFilePath(chunkName);
+            // Create the file, and then immediately close the returned OutputStream, so that HDFS may properly create the file.
+            this.fileSystem.create(fullPath, READWRITE_PERMISSION, false, 0, this.config.getReplication(),
+                    this.config.getBlockSize(), null).close();
+            log.debug("Created '{}'.", fullPath);
+
+            // return handle
+            return ChunkHandle.writeHandle(chunkName);
+        } catch (org.apache.hadoop.fs.FileAlreadyExistsException e) {
```
fpj, line 170: Can we import rather than have the full reference?

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,473 @@
...
+     * @throws IllegalArgumentException  If argument is invalid.
+     * @throws NullPointerException      If the parameter is null.
+     * @throws IndexOutOfBoundsException If the index is out of bounds.
+     */
+    @Override
+    protected int doRead(ChunkHandle handle, long fromOffset, int length, byte[] buffer, int bufferOffset)
+            throws ChunkStorageException, NullPointerException, IndexOutOfBoundsException {
+        ensureInitializedAndNotClosed();
+        try {
+
```
fpj, line 283: Remove new line.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/SegmentRollingPolicy.java
```diff
@@ -16,7 +16,10 @@
  * A generic rolling policy that can be applied to any Storage unit.
  */
 public final class SegmentRollingPolicy {
-    public static final SegmentRollingPolicy NO_ROLLING = new SegmentRollingPolicy(Long.MAX_VALUE);
+    /**
+     * Max rolling length is 2^62 so that we can use CompactLong in serialization everywhere.
```
fpj, line 20: This comment needs to change similarly to the comment in the previous PR, it should be `2^61 - 1` instead of `2^62`. I'm actually wondering why I'm seeing this here, is it because this PR includes changes of Part 1?

fpj: I actually confirmed that this is the case, that the changes of Part 1 are also here. What's the plan exactly? Do you expect git to automatically sort this out when Part 1 is merged?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: We need supportsTruncation to support truncation at end.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: We need supportsTruncation to support truncation at front, not end.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/SegmentRollingPolicy.java
sachin-j-joshi: Re-targeted base branch.

bindings/src/test/java/io/pravega/storage/filesystem/FileSystemChunkStorageMockTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Unit tests for {@link FileSystemChunkStorage} that uses mocks.
+ */
+public class FileSystemChunkStorageMockTest {
```
fpj, line 41: I'm not sure I understand the purpose of this mock, there is only a single test case it is serving, `doReadTest`.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,317 @@
...
+            ChunkInfo information = ChunkInfo.builder()
+                    .name(chunkName)
+                    .length(attrs.size())
+                    .build();
+
+            return information;
+        } catch (IOException e) {
+            throwExeption(chunkName, "doGetInfo", e);
+        }
+        throw new ChunkStorageException(chunkName, "Unreachable code");
```
fpj, line 127: Why is this necessary?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,317 @@
...
+        }
+        if (e instanceof FileAlreadyExistsException) {
+            throw new ChunkAlreadyExistsException(chunkName, message);
+        }
+        throw new ChunkStorageException(chunkName, message, e);
+    }
+
+    @Override
+    protected boolean checkExists(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        return Files.exists(Paths.get(config.getRoot(), chunkName));
```
fpj, line 160: This doesn't really throw `ChunkStorageException`, does it?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,317 @@
...
+                Preconditions.checkState(offset <= getFileSize(targetPath));
+                Preconditions.checkState(length <= getFileSize(sourcePath));
+                try (FileChannel targetChannel = getFileChannel(targetPath, StandardOpenOption.WRITE);
+                     RandomAccessFile sourceFile = new RandomAccessFile(String.valueOf(sourcePath), "r")) {
+                    while (length > 0) {
+                        long bytesTransferred = targetChannel.transferFrom(sourceFile.getChannel(), offset, length);
+                        offset += bytesTransferred;
+                        length -= bytesTransferred;
+                    }
+                    targetChannel.force(false);
```
fpj, line 284: The `false` parameter here indicates that the file metadata change is not flushed. Given that we are changing the size of the file, I'd think that the metadata flush is necessary. What do you think?

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    @Override
+    protected boolean checkExists(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        ensureInitializedAndNotClosed();
+        FileStatus status = null;
+        try {
+            status = fileSystem.getFileStatus(getFilePath(chunkName));
+        } catch (IOException e) {
+            // HDFS could not find the file. Returning false.
```
fpj, line 155: Why is it swallowing the exception? Maybe the file is there and the exception is due to something else, in which case the return value is incorrect.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    protected int doRead(ChunkHandle handle, long fromOffset, int length, byte[] buffer, int bufferOffset)
+            throws ChunkStorageException, NullPointerException, IndexOutOfBoundsException {
+        ensureInitializedAndNotClosed();
+        try {
+            if (fromOffset < 0 || bufferOffset < 0 || length < 0 || buffer.length < bufferOffset + length) {
```
fpj, line 207: Same comment about preconditions and checking `fromOffset` and `bufferOffset`. make sure that we are comprehensively checking everything we need to check.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+
+    @Override
+    public void close() {
+        if (!this.closed.getAndSet(true)) {
+            if (this.fileSystem != null) {
+                try {
+                    this.fileSystem.close();
+                    this.fileSystem = null;
+                } catch (IOException e) {
+                    log.warn("Could not close the HDFS filesystem: {}.", e);
```
fpj, line 109: The `{}` is not needed.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+        }
+        boolean exists = status != null;
+        return exists;
+    }
+
+    @Override
+    protected void doDelete(ChunkHandle handle) throws ChunkStorageException, IllegalArgumentException {
+        ensureInitializedAndNotClosed();
+        try {
+            boolean retValue = this.fileSystem.delete(getFilePath(handle.getChunkName()), true);
```
fpj, line 166: it doesn't look like we need this `boolean retValue` declaration.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+            throw new ChunkNotFoundException(chunkName, message, e);
+        }
+        throw new ChunkStorageException(chunkName, message, e);
+    }
+
+    @Override
+    protected ChunkHandle doOpenRead(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        ensureInitializedAndNotClosed();
+        try {
+            FileStatus status = fileSystem.getFileStatus(getFilePath(chunkName));
```
fpj, line 183: This `status` is not used anywhere.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+            throwException(chunkName, "doOpenRead", e);
+        }
+        return ChunkHandle.readHandle(chunkName);
+    }
+
+    @Override
+    protected ChunkHandle doOpenWrite(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        ensureInitializedAndNotClosed();
+        try {
+            FileStatus status = fileSystem.getFileStatus(getFilePath(chunkName));
```
fpj, line 194: This `status` isn't used anywhere.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+    @SneakyThrows(IOException.class)
+    public void initialize() {
+        Exceptions.checkNotClosed(this.closed.get(), this);
+        Preconditions.checkState(this.fileSystem == null, "HDFSStorage has already been initialized.");
+        Configuration conf = new Configuration();
+        conf.set("fs.default.name", this.config.getHdfsHostURL());
+        conf.set("fs.default.fs", this.config.getHdfsHostURL());
+        conf.set("fs.hdfs.impl", "org.apache.hadoop.hdfs.DistributedFileSystem");
+
+        // FileSystem has a bad habit of caching Clients/Instances based on target URI. We do not like this, since we
```
fpj, line 297: Fix this comment as the last part does not read correctly.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+        // want to own our implementation so that when we close it, we don't interfere with others.
+        conf.set("fs.hdfs.impl.disable.cache", "true");
```
fpj, line 299: Maybe point to documentation about this so that anyone checking this code understands the impact of this choice. it would be particularly important to document the potential impact of setting this to true. A quick search reveals comments going both ways.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,360 @@
...
+    private boolean isReadOnly(FileStatus fs) {
+        return fs.getPermission().getUserAction() == FsAction.READ;
+    }
+
+    private int readInternal(ChunkHandle handle, byte[] buffer, long offset, int bufferOffset, int length) throws IOException {
+        //There is only one file per chunkName.
+        try (FSDataInputStream stream = this.fileSystem.open(getFilePath(handle.getChunkName()))) {
+            stream.readFully(offset, buffer, bufferOffset, length);
+        } catch (EOFException e) {
+            throw new IllegalArgumentException(String.format("Reading at offset (%d) which is beyond the current size of chunk.", offset));
```
fpj, line 354: Could we check this as a precondition?

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+        try {
+            int partNumber = 1;
+
+            SortedSet<MultipartPartETag> partEtags = new TreeSet<>();
+            String targetPath = config.getPrefix() + chunks[0].getName();
+            String uploadId = client.initiateMultipartUpload(config.getBucket(), targetPath);
+
+            // check whether the target exists
+            if (!checkExists(chunks[0].getName())) {
+                throw new FileNotFoundException(chunks[0].getName());
```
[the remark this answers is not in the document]
sachin-j-joshi: This **should** throw ChunkNotFoundException. Looks like missed that one. (I am not sure why compiler did not complain. ).
I'll fix this

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+
+    @Override
+    public int doConcat(ConcatArgument[] chunks) throws ChunkStorageException, UnsupportedOperationException {
+        int totalBytesConcated = 0;
+        try {
+            int partNumber = 1;
+
+            SortedSet<MultipartPartETag> partEtags = new TreeSet<>();
+            String targetPath = config.getPrefix() + chunks[0].getName();
+            String uploadId = client.initiateMultipartUpload(config.getBucket(), targetPath);
```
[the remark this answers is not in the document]
sachin-j-joshi: Based on our experience with ExtendedS3, there is now a config value `minSizeLimitForConcat`. If size of the source object is below this size then concat is not called. Instead concat is performed using appends.

To avoid confusion, we removed concept of separate "concat using native capabilities" vs "concat using append" from `ChunkStorage`. Instead of that`ChunkedSegmentStorage` itself now decides whether to call concat or not and it directly translates "concat using append" into appropriate write() calls on ChunkStorage .

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
andreipaduroiu: It doesn't have to

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemUtils.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.common.collect.ImmutableSet;
+
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.Set;
+
+/**
+ * Utils for File System.
+ */
+public class FileSystemUtils {
+    public static final Set<PosixFilePermission> READ_ONLY_PERMISSION = ImmutableSet.of(
```
andreipaduroiu, line 21: Javadoc for all these fields.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
andreipaduroiu: This code is copied from the existing `HDFSStorage.java`. However the code within the `try` block is also very different. Can we check the contracts for that method to see if it warrants this whole try-catch block in here too?

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,325 @@
...
+        try {
+            setPermission(handle, isReadOnly ? Permission.READ : Permission.FULL_CONTROL);
+            return true;
+        } catch (Exception e) {
+            throw convertException(handle.getChunkName(), "doSetReadOnly", e);
+        }
+    }
+
+    private void setPermission(ChunkHandle handle, Permission permission) {
+        AccessControlList acl = client.getObjectAcl(config.getBucket(), config.getPrefix() + handle.getChunkName());
```
eolivelli, line 208: I see that we are concatenating very often `config.getPrefix() + handle.getChunkName()`:
1) it is better to have a common function to compose the final object name
2) we should validate the chunkName and check for invalid values for S3, probably sanitize it by replacing invalid characters
3) what about adding some know separator between the prefix and the chunkName ? this way it won't be possible to mix data from two separate Pravega clusters that share the same S3 bucket. I know this is a remote possibility because "prefix" should be unique and checkName will be some kind of UUID, but adding a check does not cost very much and will make this safer

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: According to the fsync documentation, if file metadata is changed as a result of file data change then metadata is also forced sync. (any metadata change that affects ability to read data correctly). It will not however sync if the only change is related to non data attributes.
In this case, this is effectively same as force(true).  I'll make it true to avoid confusion.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: Removed, this check was unnecessary. The preconditions are checked already.

bindings/src/test/java/io/pravega/storage/hdfs/HDFSSimpleStorageTest.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    public static class HDFSChunkStorageTests extends ChunkStorageTests {
+        @Rule
+        public Timeout globalTimeout = Timeout.seconds(TIMEOUT.getSeconds());
+        private TestContext testContext = new TestContext();
+
+        @Before
+        public void before() throws Exception {
+            testContext.setUp();
+            super.before();
```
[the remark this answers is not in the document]
sachin-j-joshi: Test context should be initialized before base class methods are called (which call overridden methods in this class).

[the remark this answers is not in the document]
sachin-j-joshi: Fixed after().

bindings/src/test/java/io/pravega/storage/filesystem/FileSystemChunkStorageMockTest.java
sachin-j-joshi: There are some corner cases in read logic that can only be tested using Mockito mock.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+                            partNumber++).withSourceRange(Range.fromOffsetLength(0, chunks[i].getLength()));
+
+                    CopyPartResult copyResult = client.copyPart(copyRequest);
+                    partEtags.add(new MultipartPartETag(copyResult.getPartNumber(), copyResult.getETag()));
+                    totalBytesConcated += chunks[i].getLength();
+                }
+            }
+
+            //Close the upload
+            client.completeMultipartUpload(new CompleteMultipartUploadRequest(config.getBucket(),
```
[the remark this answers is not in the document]
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+                    targetPath, uploadId).withParts(partEtags));
+            // Delete all source objects.
+            for (int i = 1; i < chunks.length; i++) {
+                client.deleteObject(config.getBucket(), config.getPrefix() + chunks[i].getName());
+            }
+        } catch (RuntimeException e) {
+            // Make spotbugs happy. Wants us to catch RuntimeException in a separate catch block.
```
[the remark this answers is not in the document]
sachin-j-joshi: No idea! Couldn't get rid of it.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+        return null;
+    }
+
+    @Override
+    protected boolean checkExists(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        try {
+            S3ObjectMetadata result = client.getObjectMetadata(config.getBucket(),
+                    config.getPrefix() + chunkName);
+            return true;
+        } catch (S3Exception e) {
```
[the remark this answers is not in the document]
sachin-j-joshi: Yes that is correct it needs to return boolean.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,317 @@
...
+    //endregion
+
+    //region
+
+    @VisibleForTesting
+    protected FileChannel getFileChannel(Path path, StandardOpenOption openOption) throws IOException {
+        try {
+            return FileChannel.open(path, openOption);
+        } catch (NoSuchFileException e) {
+            throw new FileNotFoundException(path.toString());
```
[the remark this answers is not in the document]
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemUtils.java
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,440 @@
...
+        return true;
+    }
+
+    /**
+     * Gets a value indicating whether this Storage implementation supports append operation on underlying storage object.
+     *
+     * @return True or false.
+     */
+    @Override
+    public boolean supportsAppend() {
```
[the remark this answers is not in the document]
sachin-j-joshi: It should append only at the end.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
sachin-j-joshi: I fixed the first issue. 
Not sure about 2nd and 3rd. These names are generated by us and should not contain invalid chars.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3SegmentHandle.java
```diff
@@ -17,7 +17,6 @@
     private final String segmentName;
     private final boolean isReadOnly;
 
-
```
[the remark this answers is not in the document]
sachin-j-joshi: done.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -107,7 +107,6 @@ public ExtendedS3Storage(S3Client client, ExtendedS3StorageConfig config) {
 
     //endregion
 
-
```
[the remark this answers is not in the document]
sachin-j-joshi: done

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+        if (!checkExists(chunkName)) {
+            throw new ChunkNotFoundException(chunkName, "Chunk not found");
+        }
+        return ChunkHandle.writeHandle(chunkName);
+    }
+
+    @Override
+    protected int doRead(ChunkHandle handle, long fromOffset, int length, byte[] buffer, int bufferOffset) throws ChunkStorageException, NullPointerException, IndexOutOfBoundsException {
+        try {
+            if (fromOffset < 0 || bufferOffset < 0 || length < 0) {
```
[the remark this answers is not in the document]
sachin-j-joshi: This check is not required as it is already checked by the caller method in base class.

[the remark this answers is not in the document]
sachin-j-joshi: removed

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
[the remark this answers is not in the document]
sachin-j-joshi: fixed. Was leftover from refactor.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: fixed.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: fixed.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: Precondition is checked elsewhere. But still need to handle runtime exception.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+                if (reader == null) {
+                    throw new ChunkNotFoundException(handle.getChunkName(), "Chunk not found");
+                }
+
+                int bytesRead = StreamHelpers.readAll(reader, buffer, bufferOffset, length);
+
+                return bytesRead;
+            }
+        } catch (Exception e) {
+            throwException(handle.getChunkName(), "doRead", e);
```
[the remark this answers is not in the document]
sachin-j-joshi: Exactly - that is what I did.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,317 @@
...
+            PosixFileAttributes attrs = Files.readAttributes(Paths.get(config.getRoot(), chunkName),
+                    PosixFileAttributes.class);
+            ChunkInfo information = ChunkInfo.builder()
+                    .name(chunkName)
+                    .length(attrs.size())
+                    .build();
+
+            return information;
+        } catch (IOException e) {
+            throwExeption(chunkName, "doGetInfo", e);
```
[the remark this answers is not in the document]
sachin-j-joshi: fixed.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: Fixed.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
sachin-j-joshi: I'm not sure why

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: removed redundant comments.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+                    Range.fromOffsetLength(offset, length), data);
+            return length;
+        } catch (Exception e) {
+            throwException(handle.getChunkName(), "doWrite", e);
+        }
+        return 0;
+    }
+
+    @Override
+    public int doConcat(ConcatArgument[] chunks) throws ChunkStorageException, UnsupportedOperationException {
```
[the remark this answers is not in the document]
sachin-j-joshi: fixed

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,328 @@
...
+        } catch (Exception e) {
+            throwException(chunkName, "doGetInfo", e);
+        }
+        return null;
+    }
+
+    @Override
+    protected ChunkHandle doCreate(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        try {
+            if (!client.listObjects(config.getBucket(), config.getPrefix() + chunkName).getObjects().isEmpty()) {
```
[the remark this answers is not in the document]
sachin-j-joshi: We must throw chunk already exists exception.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: fixed

bindings/src/test/java/io/pravega/storage/filesystem/FileSystemChunkStorageMockTest.java
fpj: Ok, this name confused me. if this is about testing the read logic in a given way, then name it accordingly. I thought you're defining a mock to be used across test cases, while it sounds like there are aspects of the read logic that require mocking, so the test is about read, not mocking.

sachin-j-joshi: This is not only about the read logic, but testing FileSystemChunkStorage functionality by mocking underlying file system apis.  In this context the name correctly reflects the intent.
Mocks are useful to create hard to reproduce scenario. In future I do intent to add more tests around handling exception.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,305 @@
...
+            int totalBytesRead = 0;
+            long readOffset = fromOffset;
+            do {
+                ByteBuffer readBuffer = ByteBuffer.wrap(buffer, bufferOffset, length);
+                int bytesRead = channel.read(readBuffer, readOffset);
+                bufferOffset += bytesRead;
+                totalBytesRead += bytesRead;
+                length -= bytesRead;
+                readOffset += bytesRead;
+            } while (length != 0);
```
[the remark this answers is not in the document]
andreipaduroiu: +1.
I'm always advocating for a broader condition in these cases for the same reason as Raul mentioned.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: updated.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,339 @@
...
+            if (errorCode.equals("NoSuchKey")) {
+                retValue =  new ChunkNotFoundException(chunkName, message, e);
+            }
+
+            if (errorCode.equals("PreconditionFailed")) {
+                retValue =  new ChunkAlreadyExistsException(chunkName, message, e);
+            }
+
+            if (errorCode.equals("InvalidRange")
+                    || errorCode.equals("InvalidArgument")
```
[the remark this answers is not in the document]
andreipaduroiu: IMO, This is the best place to keep them. They are ExtendedS3 specific and nobody else should need to know about them.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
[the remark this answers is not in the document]
andreipaduroiu: I don't understand how this is fixed. How is the garbage cleaned up? Is there garbage? If it's already handled, let's write it in a comment.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,339 @@
...
+            SortedSet<MultipartPartETag> partEtags = new TreeSet<>();
+            uploadId = client.initiateMultipartUpload(config.getBucket(), targetPath);
+
+            // check whether the target exists
+            if (!checkExists(chunks[0].getName())) {
+                throw new ChunkNotFoundException(chunks[0].getName(), "Target segment does not exist");
+            }
+
+            //Copy the parts
+            for (int i = 0; i < chunks.length; i++) {
```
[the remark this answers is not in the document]
sachin-j-joshi: Extended S3 guarantees that MPU operation is automatic.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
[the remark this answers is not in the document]
andreipaduroiu: So what is the Spotbugs actual message? What method in the `try` block above has such a weird declaration that spotbugs isn't happy?

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
andreipaduroiu: I think one way to disambiguate this is to name this operation `HeadTruncation`. We may want to rename these methods to `supportsHeadTruncation`. We can do it in a future issue.

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
```diff
@@ -0,0 +1,62 @@
...
+                .map(object -> new ObjectKey(object.getKey()))
+                .collect(Collectors.toList());
+
+        if (!keys.isEmpty()) {
+            client.deleteObjects(new DeleteObjectsRequest(bucketName).withKeys(keys));
+        }
+    }
+
+    public void close() throws Exception {
+        if (client != null) {
```
andreipaduroiu, line 57: Why would client be null? you set it in the constructor

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
```diff
@@ -0,0 +1,62 @@
...
+    public ExtendedS3TestContext() throws Exception {
+        String bucketName = BUCKET_NAME_PREFIX + UUID.randomUUID().toString();
+        this.adapterConfig = ExtendedS3StorageConfig.builder()
+                .with(ExtendedS3StorageConfig.CONFIGURI, configUri)
+                .with(ExtendedS3StorageConfig.BUCKET, bucketName)
+                .with(ExtendedS3StorageConfig.PREFIX, "samplePrefix")
+                .build();
+        s3Config = new ConfigUri<>(S3Config.class).parseUri(configUri);
+        s3Proxy = new S3ProxyImpl(configUri, s3Config);
+        s3Proxy.start();
```
andreipaduroiu, line 44: There is a chance that something in the constructor fails. We want to put a try-catch block here to shut down the s3proxy if we fail before we exit the constructor.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,339 @@
...
+
+        if (retValue == null) {
+            retValue = new ChunkStorageException(chunkName, message, e);
+        }
+
+        return retValue;
+    }
+
+    private String getObjectPath(String objectName) {
+        return config.getPrefix() + objectName;
```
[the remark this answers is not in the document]
sachin-j-joshi: Not really. concatenation is not in loop or anything. This concat is unavoidable we need to add object name to prefix and then pass it to S3 client as parameter.
I am sure this is the fastest way to concat two strings. String.format or other things like StringBuffer etc would have additional overhead that is not necessary.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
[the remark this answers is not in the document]
sachin-j-joshi: agree.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,339 @@
...
+    private ChunkStorageException convertException(String chunkName, String message, Exception e)  {
+        ChunkStorageException retValue = null;
+        if (e instanceof ChunkStorageException) {
+            return (ChunkStorageException) e;
+        }
+        if (e instanceof S3Exception) {
+            S3Exception s3Exception = (S3Exception) e;
+            String errorCode = Strings.nullToEmpty(s3Exception.getErrorCode());
+
+            if (errorCode.equals("NoSuchKey")) {
```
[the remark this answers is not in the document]
sachin-j-joshi: see below

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
sachin-j-joshi: The close is called in  @After in test classes.

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
sachin-j-joshi: in case close is called on half initialized object.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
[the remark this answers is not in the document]
sachin-j-joshi: **Summary**
Warning Type 	Number
Dodgy code Warnings 	1
Total 	1
Warnings

Click on a warning row to see full context information.
Dodgy code Warnings
Code 	Warning
REC 	Exception is caught when Exception is not thrown in io.pravega.storage.extendeds3.ExtendedS3ChunkStorage.doConcat(ConcatArgument[])
	
**Details**
REC_CATCH_EXCEPTION: Exception is caught when Exception is not thrown

This method uses a try-catch block that catches Exception objects, but Exception is not thrown within the try block, and RuntimeException is not explicitly caught. It is a common bug pattern to say try { ... } catch (Exception e) { something } as a shorthand for catching a number of types of exception each of whose catch blocks is identical, but this construct also accidentally catches RuntimeException as well, masking potential bugs.

A better approach is to either explicitly catch the specific exceptions that are thrown, or to explicitly catch RuntimeException exception, rethrow it, and then catch all non-Runtime Exceptions, as shown below:

```
try {
    ...
} catch (RuntimeException e) {
    throw e;
} catch (Exception e) {
    ... deal with all non-runtime exceptions ...
}
```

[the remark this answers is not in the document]
sachin-j-joshi: S3Exception is derived from RuntimeException

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    @Override
+    protected boolean doSetReadOnly(ChunkHandle handle, boolean isReadOnly) throws ChunkStorageException {
+        try {
+            setPermission(handle, isReadOnly ? Permission.READ : Permission.FULL_CONTROL);
+            return true;
+        } catch (Exception e) {
+            throw convertException(handle.getChunkName(), "doSetReadOnly", e);
+        }
```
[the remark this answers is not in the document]
sachin-j-joshi: fixed

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
andreipaduroiu: You will have nothing to call `close` on since this is the constructor. If the constructor throws, there will be no object to call `close` on, but the side effects will remain. It is your responsibility to clean up any such side effects in the constructor should it fail.

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
andreipaduroiu: What is a half initialized object? Can you show me an example where only half of a constructor executes and then you get an object as a result?

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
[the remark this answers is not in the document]
sachin-j-joshi: Moved the deletion and garbage collection to upper layers (ChunkedSegmentStorage).

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
sachin-j-joshi: line 56 above.

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
sachin-j-joshi: fixed

bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java
sachin-j-joshi: You can not write following code 
   ``` java
try {
   } catch (Exception e) {
            if (client != null) {
                client.destroy();
            }
            if (s3Proxy != null) {
                s3Proxy.stop();
            }
            throw e;
        }
```

> > Task :bindings:compileTestJava
> /home/sachin/workspace/active5/pravega/bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java:56: error: variable client might not have been initialized
>             if (client != null) {
>                 ^
> /home/sachin/workspace/active5/pravega/bindings/src/test/java/io/pravega/storage/extendeds3/ExtendedS3TestContext.java:59: error: variable s3Proxy might not have been initialized
>             if (s3Proxy != null) {
>                 ^
> 2 errors
> 
> > Task :bindings:compileTestJava FAILED

but following works so keeping it that way.

``` java
} catch (Exception e) {
            close();
            throw e;
        }


  public void close() throws Exception {
        if (client != null) {
            client.destroy();
        }
        if (s3Proxy != null) {
            s3Proxy.stop();
        }
    }
}
```

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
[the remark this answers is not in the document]
sachin-j-joshi: fixed.

RaulGracia:
@sachin-j-joshi the codecov check is very close to pass. I recommend to look to the codecov report, as mainly the uncovered regions correspond to untested `if` clauses within methods and `catch` blocks. I would suggests to have a look to them and find a way to exercise at least the most important ones, as they embody corner cases that might appear in the wild and we need to be sure about their behavior.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,294 @@
...
+
+    @VisibleForTesting
+    protected long getFileSize(Path path) throws IOException {
+        return Files.size(path);
+    }
+
+    @Override
+    protected ChunkInfo doGetInfo(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        try {
+            PosixFileAttributes attrs = Files.readAttributes(Paths.get(config.getRoot(), chunkName),
```
eolivelli, line 107: as for S3 storage, we should have a common method to perform `Paths.get(config.getRoot(), chunkName)`

it can also be a security issue to let **chunkName** be anystring. 
we should have some validation that **at least** blocks:
- null chars
- relative path sequences ("..", "/","\",".")
- anche chunkname cannot be empty

we cannot assume that the caller is a good guy, even if this is our own code

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,294 @@
...
+                    PosixFileAttributes.class);
+            ChunkInfo information = ChunkInfo.builder()
+                    .name(chunkName)
+                    .length(attrs.size())
```
eolivelli, line 111: isn't Files.size(path)  enough ?
why are we passing thru PosixFileAttributes ?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,294 @@
...
+            throw new ChunkNotFoundException(chunkName, "FileSystemChunkStorage::doOpenRead");
+        }
+
+        return ChunkHandle.readHandle(chunkName);
+    }
+
+    @Override
+    protected ChunkHandle doOpenWrite(String chunkName) throws ChunkStorageException, IllegalArgumentException {
+        Path path = Paths.get(config.getRoot(), chunkName);
+        if (!Files.exists(path)) {
```
eolivelli, line 176: please also check that this is a RegularFile, not a directory or a link (for instance)

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
eolivelli: you also have this getFileSize
https://github.com/pravega/pravega/pull/4699/files#diff-17a1fa6c3b4ce4b99b95fde90189ccc5R190

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,294 @@
...
+
+            for (int i = 1; i < chunks.length; i++) {
+                val source = chunks[i];
+                Preconditions.checkArgument(!chunks[0].getName().equals(source.getName()), "target and source can not be same.");
+                Path sourcePath = Paths.get(config.getRoot(), source.getName());
+                long length = chunks[i].getLength();
+                Preconditions.checkState(offset <= getFileSize(targetPath));
+                Preconditions.checkState(length <= getFileSize(sourcePath));
+                try (FileChannel targetChannel = getFileChannel(targetPath, StandardOpenOption.WRITE);
+                     RandomAccessFile sourceFile = new RandomAccessFile(String.valueOf(sourcePath), "r")) {
```
eolivelli, line 259: this is bad from a security perspective
`String.valueOf(sourcePath)`

you can use sourcePath.toFile()

eolivelli: why are you using RandomAccessFile ?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,294 @@
...
+            long offset = chunks[0].getLength();
+
+            for (int i = 1; i < chunks.length; i++) {
+                val source = chunks[i];
+                Preconditions.checkArgument(!chunks[0].getName().equals(source.getName()), "target and source can not be same.");
+                Path sourcePath = Paths.get(config.getRoot(), source.getName());
+                long length = chunks[i].getLength();
+                Preconditions.checkState(offset <= getFileSize(targetPath));
+                Preconditions.checkState(length <= getFileSize(sourcePath));
+                try (FileChannel targetChannel = getFileChannel(targetPath, StandardOpenOption.WRITE);
```
eolivelli, line 258: why are you opening the same file more times ?
it is better to open the file for write only once.

I suggest you also to:
- create a temporary file
- fill it
- rename new the file with ATOMIC_MOVE flag to the targetPath
This way we are protected from partial writes

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
```diff
@@ -0,0 +1,294 @@
...
+            }
+            return totalBytesConcated;
+        } catch (IOException e) {
+            throw convertExeption(chunks[0].getName(), "doConcat", e);
+        }
+    }
+
+    @Override
+    protected boolean doTruncate(ChunkHandle handle, long offset) throws UnsupportedOperationException {
+        throw new UnsupportedOperationException();
```
eolivelli, line 279: Maybe this `throw new UnsupportedOperationException() ` should be the default implementation in the base class

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,350 @@
...
+        return length;
+    }
+
+    @Override
+    public int doConcat(ConcatArgument[] chunks) throws ChunkStorageException {
+        ensureInitializedAndNotClosed();
+        int length = 0;
+        try {
+            val sources = new Path[chunks.length - 1];
+            this.fileSystem.truncate(getFilePath(chunks[0].getName()), chunks[0].getLength());
```
eolivelli, line 244: what happens in case of failure during the execution of this method ?
we are losing the contents of the file IMHO

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,350 @@
...
+    //region Storage Implementation
+
+    @SneakyThrows(IOException.class)
+    public void initialize() {
+        Exceptions.checkNotClosed(this.closed.get(), this);
+        Preconditions.checkState(this.fileSystem == null, "HDFSStorage has already been initialized.");
+        Configuration conf = new Configuration();
+        conf.set("fs.default.name", this.config.getHdfsHostURL());
+        conf.set("fs.default.fs", this.config.getHdfsHostURL());
+        conf.set("fs.hdfs.impl", "org.apache.hadoop.hdfs.DistributedFileSystem");
```
eolivelli, line 284: what about org.apache.hadoop.hdfs.DistributedFileSystem.class.getName() ? 
this way we will have a compile time guarantee that the name is good

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
```diff
@@ -0,0 +1,350 @@
...
+
+        // We do not want FileSystem to cache clients/instances based on target URI.
+        // This allows us to close instances without affecting other clients/instances. This should not affect performance.
+        conf.set("fs.hdfs.impl.disable.cache", "true");
+        if (!this.config.isReplaceDataNodesOnFailure()) {
+            // Default is DEFAULT, so we only set this if we want it disabled.
+            conf.set("dfs.client.block.write.replace-datanode-on-failure.policy", "NEVER");
+        }
+
+        this.fileSystem = openFileSystem(conf);
```
eolivelli, line 294: do we have a way to pass additional configuration parameters to this client ?
I mean, probably users would like to add additional configurations, in segment store configuration file
`hdfs.customconfig.xxxx=yyyyy `

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -847,6 +857,11 @@ private void defrag(MetadataTransaction txn, SegmentMetadata segmentMetadata, St
                     concatUsingAppend(concatArgs);
                 }
 
+                // Delete chunks.
+                for (int i = 1; i < chunksToConcat.size(); i++) {
+                    chunksToDelete.add(chunksToConcat.get(i).getName());
```
eolivelli, line 862: what happens in case of failure ? where is recovery handled ?

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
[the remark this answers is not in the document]
andreipaduroiu: Ok. I see now.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
sachin-j-joshi: - The list of chunks to delete is returns to the caller in this case ChunkedSegmentStorage::concat
- ChunkedSegmentStorage::concat calls ChunkedSegmentStorage::collectGarbage which then immediately deletes the garbage chunks. This is called only after the whole concat operation and the metadata commit succeeds.
- What happens when delete fails ? We add it a list of "to delete list" of chunks which will be deleted by a background task.
- There is a pending issue to create background garbage collection task https://github.com/pravega/pravega/issues/4903

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: There is a very specific segment store fail over situation when actual chunk has more data than its length recorded in metadata. This can happen when zombie instance keeps writing to chunk even after it is no longer the owner of the segment. The new segment store instance will effectively "seal" that chunk by recording the most recent length and adding a fresh new chunk. In short data after recorded length is garbage that needs to be ignored anyway.
See https://github.com/pravega/pravega/wiki/PDP-34:-Simplified-Tier-2#segment-store-failover

HDFS concat does not have a concept of concatenating a partial  file. So removing unneeded end part.

Segment concat is called in only two cases 
- for cases where transaction segment is merged into the main segment. 
- de-fragmenting an existing segment by merging smaller chunks into bigger chunks.
In both the cases above there is no possibility of having garbage data/invalid data at the front of the chunks.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: Ok. Makes sense.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: Good catch.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: Good catch that I was opening same file multiple times. Fixing it.

the target chunk could be multi GB file.

As for why partial writes are still okay...
- chunk metadata is not updated unless the whole operation succeeds. So even if resultant chunk has more data we never consider it valid part of chunk.
- The source chunks are not deleted.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: Good check to add. Adding

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
andreipaduroiu: The purpose of this PR was not to tweak the HDFS client. We can take that challenge in a separate issue. We are setting all the configs that we believe we need, and define the "pass-through" values in our own config. Otherwise we'd have an unmanageable zoo of configurations, most of which will be beyond our control.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
andreipaduroiu: This sounds sensible. 

@sachin-j-joshi since this this not new code (it's borrowed from the other HDFSStorage.java), may I suggest collecting all these HDFS improvement ideas (there's at least one other below) in a new GitHub issue and we can address them later? 

This particular change is unrelated to PDP34 so I don't want to tackle it here.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
eolivelli: @andreipaduroiu agreed.
@sachin-j-joshi probably there is not need to create a ticket if no user asks for it.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: This was from the time when I had isReadOnly as property of Chunk .
Fixed

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: Fixed

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: Fixed for loop

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: done

bindings/src/main/java/io/pravega/storage/hdfs/HDFSChunkStorage.java
sachin-j-joshi: Fixed
conf.set("fs.hdfs.impl", org.apache.hadoop.hdfs.DistributedFileSystem.class.getName());

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java
sachin-j-joshi: fixed the empty name part.
Now also checking that path is not directory, that should handle second case.

sachin-j-joshi:
Added more tests ...(wrapping all calls to File system in FileSystemWrapper  and using mocks to add few last tests)

image

andreipaduroiu:
@eolivelli In the interest of making progress on this feature, I merged the PR into the feature branch. I understand you may still have some open questions. Let's follow up on them and see if we can tackle in a different PR on this feature branch (if related to this work) or in a separate GitHub issue (if they were "inherited" from old code).

PR #4703, opened 2020-04-16
Issue 4702: (SegmentStore) Fixed a deadlock condition in the Read Index.

### LINKED ISSUE #4702: Read Index Deadlock with concurrent Storage Reads that hit a full cache
**Problem description**
Steps:
1. Segment 1: Storage Read Complete -> Ack -> Insert in Index -> Acquire (ReadIndex1.Lock[Thread1]) -> Insert in Cache [Request1]
 2. Segment 2: Storage Read Complete -> Ack -> Insert in Index -> Acquire (ReadIndex2.Lock[Thread2]) -> Insert in Cache [Request2]
 3. Cache is full. Deadlock occurs if:
 3.1. [Request1] invokes Cache Eviction, which wants to acquire ReadIndex2.Lock, but it is owned by Thread2.
 3.2. [Request2] invokes Cache Eviction, which wants to acquire ReadIndex1.Lock, but it is owned by Thread1.

**Problem location**
Read Index/CacheManager/Cache

Break the deadlock by choosing a victim and have it back off and retry.

andreipaduroiu:
Signed-off-by: Andrei Paduroiu <andrei.paduroiu@emc.com>

**Change log description**  
- Changed the `CacheManager` to reject additional concurrent cache cleanup requests if one is ongoing (as opposed from using a mutex to serialize them).
- Changed the `DirectMemoryCache` to use an exp backoff retry if it cannot clean up the cache after it detected the cache was full.

**Purpose of the change**  
Fixes #4702.

**What the code does**  
See #4702 for a description of a scenario that leads to this. 
The problem arises if 2 concurrent Storage reads complete at about the same time, and both want to insert into a full cache at the same time. Each callback would execute on its own thread, acquire the Read Index lock associated with their segment, then acquire the Cache Manager eviction lock (whomever does it first), and the Cache Manager would attempt to clean up the index by accessing the other Read Indices (leading to a deadlock).

This has been fixed in 2 steps:
1. Removed the Cache Manager lock altogether. If another request comes in while a cache eviction is in progress, the second request will be rejected. 
2. Changed the Streaming Cache (`DirectMemoryCache`) to use an exp backoff retry in case it is unable to trigger a cache eviction
    - Previously it would retry immediately and give up as soon as the CM said nothing could be done.
    - Now it will try up to 5 times, waiting 50-250ms in between requests. 
    - If after 5 attempts it is not able to do anything, then it will throw up a `CacheFullException`
        - This is the same as before, except that now it adds the extra delays in between checks.
    - On the append path, a `CacheFullException` will shut down the container (this has not changed)
    - On the Storage Read Path, a `CacheFullException` has no effect on the request. The original request has already been ack-ed/sent to the client before the attempt to write to the cache. If we cannot write to the cache due to cache full, we will log a warning now so we can trace this in the logs.

**How to verify it**  
New, specific unit test added that verifies this exact situation. The unit test would not complete if there were a deadlock.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
@@ -26,6 +26,8 @@ import java.util.Iterator;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 import lombok.Getter;
@@ -47,17 +49,15 @@ import lombok.extern.slf4j.Slf4j;
 @ThreadSafe
 public class CacheManager extends AbstractScheduledService implements AutoCloseable {
     //region Members
-
+    private static final int CACHE_FULL_RETRY_BASE_MILLIS = 50;
     private static final String TRACE_OBJECT_ID = "CacheManager";
     @GuardedBy("lock")
     private final Collection<Client> clients;
     private final ScheduledExecutorService executorService;
-    @GuardedBy("lock")
-    private int currentGeneration;
-    @GuardedBy("lock")
-    private int oldestGeneration;
-    @GuardedBy("lock")
-    private CacheState lastCacheState;
+    private final AtomicInteger currentGeneration;
+    private final AtomicInteger oldestGeneration;
+    private final AtomicReference<CacheState> lastCacheState;
+    private final AtomicBoolean running;
     private final CachePolicy policy;
     private final AtomicBoolean closed;
     private final SegmentStoreMetrics.CacheManager metrics;
@@ -85,7 +85,7 @@ public class CacheManager extends AbstractScheduledService implements AutoClosea
      * Creates a new instance of the CacheManager class.
      *
      * @param policy          The policy to use with this CacheManager.
-     * @param cacheStorage       The CacheStorage to maintain.
+     * @param cacheStorage    The CacheStorage to maintain.
      * @param executorService An executorService to use for scheduled tasks.
      */
     @VisibleForTesting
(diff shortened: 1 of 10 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
```diff
@@ -399,15 +403,23 @@ private DirectMemoryBuffer getNextAvailableBuffer() {
...
         throw new CacheFullException(String.format("%s full: %s.", DirectMemoryCache.class.getSimpleName(), getState()));
     }
 
-    private boolean tryCleanup() {
+    private void tryCleanup(int attempts) {
         val c = this.tryCleanup.get();
-        return c != null && c.get();
+        if (c != null && !c.get()) {
+            // Unable to clean up the cache. Wait a bit, then try again.
+            int sleepMillis = attempts * this.retryDelayBaseMillis.get();
```
tkaitchuck, line 418: RetryPolicy in common is built to do this.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
```diff
@@ -47,17 +49,15 @@
...
     @GuardedBy("lock")
     private final Collection<Client> clients;
     private final ScheduledExecutorService executorService;
-    @GuardedBy("lock")
-    private int currentGeneration;
-    @GuardedBy("lock")
-    private int oldestGeneration;
-    @GuardedBy("lock")
-    private CacheState lastCacheState;
+    private final AtomicInteger currentGeneration;
```
tkaitchuck, line 57: As usual when changing from course locking to fine locking make double sure there aren't code paths that implicitly assume that another thread isn't going to change something in the mean time. For example here one might make the assumption that currentGeneration would not change in the middle of a method, or that current and oldest generation are different or that older is newer that current etc.

RaulGracia:
@tkaitchuck @andreipaduroiu this PR has been validated and works, we will need to cherry pick it to 0.7, so please let's see if we can merge it soon.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/DirectMemoryCache.java
andreipaduroiu: That was my first choice, however `Retry` only retries on exceptions. Here I retry based on the response from a method.

RaulGracia:
@andreipaduroiu it looks like a new experiment has shown some problem with this branch, it will be good to check if the issue is related to these changes or not before merging (my previous commend was based on incomplete testing, apparently).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/CacheManager.java
andreipaduroiu: Agreed. In this class, the only thing changing the internal state is `applyCachePolicyInternal`, which is guarded by the lock. I decided to make all these fields `AtomicXYZ` because I didn't want to acquire some lock just to query them. To be clear, `applyCachePolicyInternal` modifies these values, but they are queried in other places as well, which should be OK for thread safety purposes.

The only exception to this rule is `this.clients`, which keeps track of all the Cache Manager's Clients, and is a collection that is modified when new segments are activated or evicted from memory, hence out of the control of the Cache Manager. However, no external APIs are invoked while using the `clients` lock, so no chance for deadlock there.

PR #4710, opened 2020-04-20
Issue 3994 : List all Segments in Tier 2

### LINKED ISSUE #3994: List all Segments in Tier 2

ManishKumarKeshri:
**Change log description**  
Added list-segments API to retrieve the segments from Tier 2.

**Purpose of the change**  
Fixes #3994 

**What the code does**  
Basic implementation of Iterator over the header files in the storage returning SegmentProperties per file.
The iterator logic is tailored to the specific underlying storage depending on how the listing of entries works in that implementation.
For retrieving StreamSegmentInformation for a particular header file, RollingSegmentHandle is used.

**What this doesn't do**
As of now, the API support is only verified for the FileSystemStorage type. For other storage types such as s3 and HDFS, it will be considered later.

**How to verify it**  
Unit tests added to verify feature.


ManishKumarKeshri:
Build passes now. Please leave your review @andreipaduroiu @sachin-j-joshi

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/rolling/RollingStorageSegmentIteratorMockTests.java
```diff
@@ -0,0 +1,136 @@
...
+     * @throws StreamingException openHandle method mocked to throw this exception.
+     */
+    @Test
+    public void testNext() throws StreamingException {
+        RollingStorageSegmentIteratorMockTests.TestRollingStorageSegmentIterator testRollingStorageSegmentIterator = new
+                RollingStorageSegmentIteratorMockTests.TestRollingStorageSegmentIterator(null, null,
+                null);
+        val baseStorage = new InMemoryStorage();
+        testRollingStorageSegmentIterator.instance = Mockito.spy(new RollingStorage(baseStorage, DEFAULT_ROLLING_POLICY));
+        testRollingStorageSegmentIterator.current = new SegmentProperties() {
```
sachin-j-joshi, line 52: I suggest returning StreamSegmentInformation.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -187,6 +189,37 @@ public boolean supportsTruncation() {
         return false;
     }
 
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(value = "OS_OPEN_STREAM", justification = "Rare operation. The leaked object is collected by GC. In case of a iteraror in a for loop this would be fast.")
+    @Override
+    public Iterator<SegmentProperties> listSegments() {
+        if (log.isDebugEnabled()) {
```
sachin-j-joshi, line 195: I don't think we need to check isDebugEnabled

ManishKumarKeshri: Done.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/rolling/RollingStorageSegmentIteratorMockTests.java
ManishKumarKeshri: Thanks for suggesting. Used StreamSegmentInformation.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -181,6 +185,71 @@ public boolean supportsTruncation() {
...
+    @Override
+    public Iterator<SegmentProperties> listSegments() {
+        return new ExtendedS3SegmentIterator(s3object -> true);
+    }
+
+    /**
+     * Iterator for segments in ExtendedS3Storage.
+     */
+    public class ExtendedS3SegmentIterator implements Iterator<SegmentProperties> {
+        ListObjectsResult results;
```
andreipaduroiu, line 197: Make all of these private.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -181,6 +185,71 @@ public boolean supportsTruncation() {
...
+            }
+            results = null;
+            innerIterator = null;
+            current = null;
+            return false;
+        }
+
+        @Override
+        public SegmentProperties next() {
+            if (null != current) {
```
andreipaduroiu, line 239: This doesn't work as it should. In an `Iterator`, when you invoke `next`, you get the "next" result and on a subsequent invocation, you should get the following item.

Your implementation does not do this. Actually it seems to to advance the "next" every time you invoke `hasNext` ... so if you do: 
```
i.hasNext();
i.hasNext();
```

then your iterator has advanced two positions. `hasNext` should be a side-effect-free operation.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
andreipaduroiu: Also some of them can be made `final`

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -187,6 +189,34 @@ public boolean supportsTruncation() {
...
+    @Override
+    public Iterator<SegmentProperties> listSegments() {
+        try {
+            return Files.find(Paths.get(config.getRoot()),
+                    Integer.MAX_VALUE,
+                    (filePath, fileAttr) -> fileAttr.isRegularFile())
+                    .map(path -> (SegmentProperties) getStreamSegmentInformation(config.getRoot(), path))
+                    .iterator();
+        } catch (IOException e) {
+            log.error("Exception occurred while listing the segments.", e);
```
andreipaduroiu, line 202: You should throw this exception. If you just return an empty iterator then the caller will never know something went amiss.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -187,6 +189,34 @@ public boolean supportsTruncation() {
         return false;
     }
 
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(value = "OS_OPEN_STREAM", justification = "Rare operation. The leaked object is collected by GC. In case of a iteraror in a for loop this would be fast.")
```
andreipaduroiu, line 192: What is this?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -187,6 +189,34 @@ public boolean supportsTruncation() {
...
+                    (filePath, fileAttr) -> fileAttr.isRegularFile())
+                    .map(path -> (SegmentProperties) getStreamSegmentInformation(config.getRoot(), path))
+                    .iterator();
+        } catch (IOException e) {
+            log.error("Exception occurred while listing the segments.", e);
+        }
+        return Collections.emptyIterator();
+    }
+
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings
```
andreipaduroiu, line 207: and here

bindings/src/main/java/io/pravega/storage/hdfs/HDFSStorage.java
```diff
@@ -487,6 +491,78 @@ public SegmentHandle create(String streamSegmentName) throws StreamSegmentExcept
...
+                        }
+                        try {
+                            Long.parseLong(fileName.substring(index + 1));
+                        } catch (NumberFormatException nfe) {
+                            return false;
+                        }
+                        return true;
+                    });
+        } catch (Exception e) {
+            log.error("Exception occurred while listing the segments.", e);
```
andreipaduroiu, line 513: throw it

bindings/src/main/java/io/pravega/storage/hdfs/HDFSStorage.java
```diff
@@ -487,6 +491,78 @@ public SegmentHandle create(String streamSegmentName) throws StreamSegmentExcept
...
+        }
+        return Collections.emptyIterator();
+    }
+
+    /**
+     * Iterator for segments in HDFS Storage.
+     */
+    public static class HDFSSegmentIterator implements Iterator<SegmentProperties> {
+        RemoteIterator<FileStatus> results;
```
andreipaduroiu, line 522: Private and final

bindings/src/main/java/io/pravega/storage/hdfs/HDFSStorage.java
```diff
@@ -487,6 +491,78 @@ public SegmentHandle create(String streamSegmentName) throws StreamSegmentExcept
...
+        java.util.function.Predicate<FileStatus> patternMatchPredicate;
+
+        HDFSSegmentIterator(RemoteIterator<FileStatus> results, java.util.function.Predicate<FileStatus> patternMatchPredicate) {
+            this.results = results;
+            this.patternMatchPredicate = patternMatchPredicate;
+        }
+
+        @Override
+        public boolean hasNext() {
+            try {
```
andreipaduroiu, line 533: same comments as in ExtendedS3Storage.

bindings/src/test/java/io/pravega/storage/hdfs/HDFSSegmentIteratorMockTests.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.function.Predicate;
+
+import org.mockito.Mockito;
+
+public class HDFSSegmentIteratorMockTests {
+    static final Duration TIMEOUT = Duration.ofSeconds(30);
+
+    @Rule
+    public Timeout globalTimeout = Timeout.seconds(TIMEOUT.getSeconds());
+    private RemoteIterator<FileStatus> results;
```
andreipaduroiu, line 33: Why is this a shared member?

bindings/src/test/java/io/pravega/storage/hdfs/HDFSSegmentIteratorMockTests.java
```diff
@@ -0,0 +1,137 @@
...
+    private java.util.function.Predicate<FileStatus> patternMatchPredicate;
```
andreipaduroiu, line 34: this can be made static-final; then you won't need the `setup` method.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/AsyncStorageWrapper.java
```diff
@@ -138,6 +139,11 @@ public boolean supportsTruncation() {
         return supplyAsync(() -> this.syncStorage.exists(streamSegmentName), streamSegmentName);
     }
 
+    @Override
+    public CompletableFuture<Iterator<SegmentProperties>> listSegments() {
```
andreipaduroiu, line 143: So ... why is this returning a `CompletableFuture` if you are always wrapping a sync call?

Just return `Iterator<SegmentProperties` from the method.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/mocks/InMemoryStorage.java
```diff
@@ -195,6 +196,11 @@ public boolean supportsTruncation() {
         return false;
     }
 
+    @Override
+    public Iterator<SegmentProperties> listSegments() {
+        return this.streamSegments.values().stream().map( s -> s.getInfo()).iterator();
```
andreipaduroiu, line 201: Did you notice that `streamSegments` is guarded by a lock? 

This tells you two things:
1. Wrap this call in a `synchronized` block
2. Make a copy of the `values()` collection in that block and iterate over the copy. That will prevent your iterator from failing if some other thread modifies the collection while you iterate through it.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -516,6 +518,59 @@ public boolean supportsTruncation() {
...
+    public Iterator<SegmentProperties> listSegments() {
+        return new RollingStorageSegmentIterator(this, this.baseStorage.listSegments(),
+                props -> NameUtils.isHeaderSegment(props.getName()));
+    }
+
+    /**
+     * Iterator for segments in Rolling storage.
+     */
+    public static class RollingStorageSegmentIterator implements Iterator<SegmentProperties> {
+        RollingStorage instance;
```
andreipaduroiu, line 531: Private final for all of these

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -516,6 +518,59 @@ public boolean supportsTruncation() {
...
+            this.patternMatchPredicate = patternMatchPredicate;
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (results != null) {
+                while (results.hasNext()) {
+                    current = results.next();
+                    if (patternMatchPredicate.test(current)) {
+                        return true;
```
andreipaduroiu, line 548: Same comment as in the other iterators.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -516,6 +518,59 @@ public boolean supportsTruncation() {
...
+            if (null != this.current) {
+                try {
+                    String segmentName = NameUtils.getSegmentNameFromHeader(current.getName());
+                    val handle = instance.openHandle(segmentName, true);
+                    return StreamSegmentInformation.builder()
+                            .name(segmentName)
+                            .length(handle.length())
+                            .sealed(handle.isSealed()).build();
+                } catch (StreamSegmentException e) {
+                    log.error("Exception occurred while fetching the next segment.", e);
```
andreipaduroiu, line 567: throw

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -666,7 +721,7 @@ private boolean shouldConcatNatively(RollingSegmentHandle source, RollingSegment
                 && lastTarget.getLength() + lastSource.getLength() <= target.getRollingPolicy().getMaxLength();
     }
 
-    private RollingSegmentHandle openHandle(String segmentName, boolean readOnly) throws StreamSegmentException {
+    public RollingSegmentHandle openHandle(String segmentName, boolean readOnly) throws StreamSegmentException {
```
andreipaduroiu, line 724: Why public?

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
```diff
@@ -40,17 +47,18 @@
...
  */
 public abstract class StorageTestBase extends ThreadPooledTestSuite {
     //region General Test arguments
-
     protected static final Duration TIMEOUT = Duration.ofSeconds(30);
     protected static final long DEFAULT_EPOCH = 1;
     protected static final int APPENDS_PER_SEGMENT = 10;
     protected static final String APPEND_FORMAT = "Segment_%s_Append_%d";
     private static final int SEGMENT_COUNT = 4;
+    private static final Random RANDOM = new Random();
```
andreipaduroiu, line 61: In your tests, always initialize your RNG with a constant. That way, if you get a failure, it will be easy to reproduce since you'll always have the same input.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/rolling/RollingStorageSegmentIteratorMockTests.java
```diff
@@ -0,0 +1,101 @@
...
+                .name("x$header")
+                .build();
+        Mockito.doThrow(mock(StreamSegmentException.class)).when(testRollingStorageSegmentIterator.instance).openHandle(anyString(), anyBoolean());
+        boolean caughtException = false;
+        try {
+            testRollingStorageSegmentIterator.next();
+        } catch (NoSuchElementException e) {
+            caughtException = true;
+        }
+        Assert.assertTrue(caughtException);
```
andreipaduroiu, line 60: You can replace this try-catch-assert with  `AssertExtensions.assertThrows`.

Please fix this anywhere else you used this pattern.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -194,24 +194,36 @@ public boolean supportsTruncation() {
      * Iterator for segments in ExtendedS3Storage.
      */
     public class ExtendedS3SegmentIterator implements Iterator<SegmentProperties> {
-        ListObjectsResult results;
-        S3Object current;
-        Iterator<S3Object> innerIterator;
-        java.util.function.Predicate<S3Object> patternMatchPredicate;
+        private final java.util.function.Predicate<S3Object> patternMatchPredicate;
+        private ListObjectsResult results;
```
andreipaduroiu, line 198: I think this can be final too.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -194,24 +194,36 @@ public boolean supportsTruncation() {
...
 
+        /**
+         * Method to check the presence of next element in the iterator.
+         * It also sets the position of the current element for Next method, but repetitive call to this method before Next
+         * will not advance the current element.
+         * @return true if the next element is there, else false.
+         */
         @Override
         public boolean hasNext() {
+            if (isAvailable) { // If the method was already called before next method call, return true if last call returned true.
```
andreipaduroiu, line 218: I've been thinking. ...

This could be achieved by doign something like 
(constructor)
```
innerIterator = client.listObjects(config.getBucket(), config.getPrefix()).stream()
                      .flatMap(r -> r.getObjects().stream())
                      .filter(patternMatchIPredicate)
                      .map(this::toSegmentProperties)
                      .iterator()
```

`hasNext`: `return innerIterator.hasNext()`
`next`: `return innerIterator.next()`
`toSegmentProperties`: put the code you have in `next` here

Does this work? It may require a couple of adjustments but it should eliminate the need for all this extra code.
Please fix the other iterators too.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -194,24 +194,36 @@ public boolean supportsTruncation() {
      * Iterator for segments in ExtendedS3Storage.
      */
     public class ExtendedS3SegmentIterator implements Iterator<SegmentProperties> {
```
andreipaduroiu, line 196: 1. This class does not need to be public. 
2. Can it be static?
3. To better organize the code, please move the class at the bottom of this file. Do not mix methods and inner classes together.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
andreipaduroiu: Ideall the `toSegmentProperties` would share code with `doGetStreamSegmentInfo`. Both places extract info from S3ObjectMetadata into our own types.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSStorage.java
```diff
@@ -511,30 +510,42 @@ public SegmentHandle create(String streamSegmentName) throws StreamSegmentExcept
...
             log.error("Exception occurred while listing the segments.", e);
+            throw e;
         }
-        return Collections.emptyIterator();
     }
 
     /**
      * Iterator for segments in HDFS Storage.
      */
     public static class HDFSSegmentIterator implements Iterator<SegmentProperties> {
```
andreipaduroiu, line 520: Same comments here:
1. Make private.
2. Move to the bottom of the file
3. I don't think you can use `stream()` here since you are already working with an iterator. Use the `Iterators` class. There's a `filter` method that you can use.

bindings/src/main/java/io/pravega/storage/hdfs/HDFSStorage.java
```diff
@@ -654,5 +667,77 @@ private int readInternal(SegmentHandle handle, byte[] buffer, long offset, int b
...
+        /**
+         * Method to return the next element in the iterator.
+         * @return A newly created StreamSegmentInformation class.
+         * @throws NoSuchElementException in case of an unexpected failure.
+         */
+        @Override
+        public SegmentProperties next() throws NoSuchElementException {
+            try {
+                return results.next();
+            } catch (NoSuchElementException e) {
```
andreipaduroiu, line 737: There is no need for this try-catch as the exception will automatically bubble up.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/mocks/InMemoryStorage.java
```diff
@@ -195,6 +197,14 @@ public boolean supportsTruncation() {
         return false;
     }
 
+    @Override
+    public Iterator<SegmentProperties> listSegments() {
+        synchronized (this) {
+            Collection<StreamSegmentData> copyValues = this.streamSegments.values();
```
andreipaduroiu, line 203: This is not making a copy. You need to do `new ArrayList<>(this.streamSegments.values())`  inside the `synchronized` block, and then move the following line out of that block.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -834,5 +845,57 @@ private void ensureOffset(RollingSegmentHandle handle, long offset) throws Strea
         }
     }
 
+    /**
+     * Iterator for segments in Rolling storage.
+     */
+    private static class RollingStorageSegmentIterator implements Iterator<SegmentProperties> {
+        protected RollingStorage instance;
```
andreipaduroiu, line 852: final

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -834,5 +845,57 @@ private void ensureOffset(RollingSegmentHandle handle, long offset) throws Strea
...
+        /**
+         * Method to return the next element in the iterator.
+         * @return A newly created StreamSegmentInformation class.
+         * @throws NoSuchElementException in case of an unexpected failure.
+         */
+        @Override
+        public SegmentProperties next() throws NoSuchElementException {
+            try {
+                return results.next();
+            } catch (NoSuchElementException e) {
```
andreipaduroiu, line 895: same here

eolivelli:
Is there any particular reason for using `java.util.Iterator` ?
that interface supports "removal" of elements, but here we have only a readonly view.

PR #4713, opened 2020-04-21
Issue 4712: Make configuration keys consistent and uniform

### LINKED ISSUE #4712: Inconsistently and confusingly named configuration keys
**Problem description**
Some of Pravega's configuration keys have odd and even misleading names. For example,
```properties
autoScale.tlsCertFile -- Actually represents the truststore for autoscaler-to-controller communications
controller.auth.userPasswordFile -- actually represents the default password auth handler database file
```
The name-space schemes of the configuration items are also quite inconsistent. For example,

```
controller.auth.tlsEnabled  -- Enable TLS for the controller
pravegaservice.enableTls    -- Enable TLS for the Segment Store 
```

**Problem location**

* Controller config
* Segment store config
* Client config

Use a consistent naming scheme. Use precise names to represent items that are confusing.

ravisharda:
**Change log description**  
Make configuration keys consistent. 

The naming scheme used in the PR is described [here](https://github.com/pravega/pravega/wiki/Configuration-Naming-Guidelines). 

**Purpose of the change**  
Resolves #4712 

**What the code does**  
* Modifies configuration keys across the board. 
* Supports backward compatibility, so that old keys/names can be migrated to new ones asynchronously. Old names continue to work. Also, old and new names can coexist. 

Pending items (to be done during the course of the review): 
* Update preexisting documentation. 
* Notify dependent projects. 
* Document the change. 

**How to verify it**  
All unit, integration, and system tests must pass.


ravisharda:
> If you do this, you need to do it for all the Segment Store components. Check out all classes that have a `Config` suffix.
> 
> In addition, this will break any existing installation, script or deployment that references these settings.
> 
> Have you tried running the standalone, the [self tester ](https://github.com/pravega/pravega/wiki/Local-Stress-Testing)or system tests with this change?

This is an early draft right now, and other changes, as well as test runs, are pending. I wanted to obtain early feedback about the new naming scheme being used here (as I mentioned in Slack).

ravisharda:
Removed the `tags/breaksAPI` label, as there is support for backward compatibility now.

config/config.properties
```diff
@@ -29,82 +29,82 @@
...
-#storageextra.storageNoOpMode=false
+#storageextra.noOpMode.enable=false
 
 # When Storage NO-OP Mode is true, storageWriteNoOpModeLatencyMillis is applied to write operation.
 # It is used to compensate the supposed latency if the storage is in operation, in order to make the test as real as possible.
 # This latency has no effect if storage is not in No-Op mode.
 # Valid values: non-negative integer
 # Default vale: 20 (milliseconds)
-#storageextra.storageWriteNoOpLatencyMillis=20
+#storageextra.write.noOp.latency.milliseconds=20
```
fpj, line 107: `write` here seems to refer to an IO write, should it be `writer` instead? Also, is it related to the previous configuration option above such that we need to add `.noOp.` there too?

ravisharda: @andreipaduroiu could you please clarify the purpose of this configuration item? Could you also suggest a better name for this?

andreipaduroiu: @kevinhan88 wrote this thing. I would drop the `write` and just emphasize `noOp.

common/src/main/java/io/pravega/common/util/Property.java
```diff
@@ -45,7 +47,11 @@
      * @return A new instance of the Property class with the given default value.
      */
     public static <T> Property<T> named(String name, T defaultValue) {
-        return new Property<>(name, defaultValue);
+        return new Property<>(name, defaultValue, null);
+    }
+
+    public static <T> Property<T> named(String name, T defaultValue, String legacyName) {
```
andreipaduroiu, line 53: Javadoc

common/src/main/java/io/pravega/common/util/Property.java
```diff
@@ -55,6 +61,10 @@ boolean hasDefaultValue() {
         return this.defaultValue != null;
     }
 
+    boolean hasLegacyName() {
```
andreipaduroiu, line 64: Javadoc

config/config.properties
```diff
@@ -29,82 +29,82 @@
...
 # Valid values: Positive integer.
 # Recommended setting: 2 * Number of containers per node, minimum 20.
-#pravegaservice.threadPoolSize=30
+#pravegaservice.threadPool.size=30
 
 # Maximum number of threads in the Thread Pool used for Tier2 tasks (reading, writing, create, delete, etc.).
 # Valid values: Positive integer.
 # Recommended setting: reasonably large number which does not cause thrashing, minimum 20, recommended value 200.
-#pravegaservice.storageThreadPoolSize=200
+#pravegaservice.storageThreadPool.size=200
```
andreipaduroiu, line 51: `pravegaservice.threadPool.storage.size`

config/config.properties
```diff
@@ -29,82 +29,82 @@
...
-pravegaservice.containerCount=4
+pravegaservice.container.count=4
 
 # Maximum number of threads in the Core SegmentStore Thread Pool. This pool is used for all SegmentStore-related
 # activities, except Netty-related tasks and Tier2 Storage activities. Examples include: handling inbound requests,
 # processing reads, background maintenance operations and background operation processing.
 # Valid values: Positive integer.
 # Recommended setting: 2 * Number of containers per node, minimum 20.
-#pravegaservice.threadPoolSize=30
+#pravegaservice.threadPool.size=30
```
andreipaduroiu, line 46: `pravegaservice.threadPool.core.size`

config/config.properties
```diff
@@ -29,82 +29,82 @@
...
-#pravegaservice.zkTrustStorePasswordPath=
+#pravegaservice.zk.connect.security.tls.trustStore.pwd.location=
 
 # DataLog implementation for Tier 1 storage.
 # Valid values: BOOKKEEPER, INMEMORY.
 # Default value: BOOKKEEPER
-pravegaservice.dataLogImplementation=BOOKKEEPER
+pravegaservice.dataLog.impl.name=BOOKKEEPER
 
 # Storage implementation for Tier 2 storage.
```
andreipaduroiu, line 89: Since you are editing this file, can you please replace `Tier 2` with `Long Term Storage` here and `Tier 1` with `DurableDataLog`. We have recently consolidated our naming to follow what is in the code.

config/config.properties
```diff
@@ -29,82 +29,82 @@
...
 # Valid values: HDFS, FILESYSTEM, EXTENDEDS3, INMEMORY.
 # Default value: HDFS
-# pravegaservice.storageImplementation=HDFS
+# pravegaservice.storage.impl.name=HDFS
 
 # Storage NO-OP Mode: in No-Op mode, user stream segment writing is no-oped; user stream segment reading is not supported.
 # This mode is used to avoid storage interference in testing while still keep the system functioning as usual.
 # NOTE: pravegaservice.storageImplementation is still used to store metadata and system segments, which are required for the functioning of the Pravega Cluster.
```
andreipaduroiu, line 96: Did you fix this reference? There may be others, so please check this file for references in the comments.

config/config.properties
andreipaduroiu: Actually, please be consistent with the above property. You use `noOpMode` and here you use `noOp`.

config/config.properties
```diff
@@ -542,12 +542,12 @@ extendeds3.configUri=http://localhost:9020?identity=user&secretKey=password
 # Recommended values: 16KB. Various load tests showed that this maximum page size produces the smallest B+Tree index
 # segment when loaded with either sequential values (very rare) or arbitrary values in arbitrary order (most common); this
 # value also produces the best balance between read/write performance and index fragmentation.
-#attributeindex.maxIndexPageSizeBytes=32767
+#attributeindex.index.page.size.bytes.max=32767
```
andreipaduroiu, line 545: `index.page` should be one word.

docker/compose/docker-compose.yml
```diff
@@ -7,10 +7,10 @@
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
-version: '2'
+version: '3'
 services:
   zookeeper:
-    image: zookeeper:3.5.4-beta
```
andreipaduroiu, line 13: Please do not sneak in changes like this. If you want to upgrade dependencies, do it in another PR.

docker/compose/docker-compose.yml
```diff
@@ -72,8 +72,7 @@ services:
       ZK_URL: zookeeper:2181
       REST_SERVER_PORT: 10080
       JAVA_OPTS: |
-        -Dcontroller.service.port=9090
-        -Dconfig.controller.metricenableCSVReporter=false
```
andreipaduroiu, line 76: Why did you remove this?

docker/compose/docker-compose.yml
```diff
@@ -96,11 +95,10 @@ services:
       ZK_URL: zookeeper:2181
       CONTROLLER_URL: tcp://${HOST_IP}:9090
       JAVA_OPTS: |
-        -Dmetrics.enableCSVReporter=false
```
andreipaduroiu, line 99: and this?

derekm:
Since this breaks backwards compatibility of existing deployments that might seek to upgrade, can it include an AWK, sed or Perl script that performs all the necessary substitutions on users' custom config files?

docker/compose/docker-compose.yml
ravisharda: Here's the reason: 
* I didn't find `config.controller.metricenableCSVReporter` config param anywhere. So, I assume this is not used. 
* `controller.service.port` is replaced by `controller.service.rpc.listener.port`.

docker/compose/docker-compose.yml
ravisharda: I don't see `metrics.enableCSVReporter` in `MetricsConfig`. So, I am assuming this is unused.

ravisharda:
> Since this breaks backwards compatibility of existing deployments that might seek to upgrade, can it include an AWK, sed or Perl script that performs all the necessary substitutions on users' custom config files?

The intent is not to break backward compatibility. It supports backward compatibility, so that old configuration keys/names can be migrated to new ones asynchronously. Old names continue to work. Also, old and new names can coexist. 

That being said, it'd still be a good idea to provide a script that migrates old config names to new ones. Maybe as a separate PR.

ravisharda:
**Important note for reviewers:** Consecutive system test runs for this PR have failed. So, it is not ready to get merged even if the PR is approved, until we identify the problem and fix it. 

Standalone and Compose-based distributed mode cluster ran successfully, but apparently it fails on Kubernetes. 

I believe we can continue to go through the review process (except merge). I'll debug/fix any issue that might be causing the system test issue and let you know.

docker/compose/docker-compose.yml
ravisharda: The Zookeeper operator for K8s upgraded Zookeeper to `3.5.5` in Oct 2019: https://github.com/pravega/zookeeper-operator/pull/83. I noticed this anomaly where these Compose files still refers to an older/`beta` version of Zookeeper, and so I made this change. If you guys prefer, I can raise another PR.

config/config.properties
ravisharda: Done.

config/config.properties
ravisharda: Done.

common/src/main/java/io/pravega/common/util/Property.java
ravisharda: Done.

common/src/main/java/io/pravega/common/util/Property.java
ravisharda: Added.

ravisharda:
> Standalone and Compose-based distributed mode cluster ran successfully, but apparently it fails on Kubernetes.
> 

With the latest changes, it works on Kubernetes as well. A system test run (branch builder Jenkins build no. 2443), which runs system tests on Kubernetes, was also successful.

config/config.properties
ravisharda: Fixed. Also, I left the `write` in the name as it is apparently referring to `write` latency.

config/config.properties
ravisharda: I made the changes in the comments everywhere in this file, as suggested.

config/config.properties
ravisharda: Fixed.

ravisharda:
@fpj , @andreipaduroiu , @RaulGracia and @sachin-j-joshi could you please action on this?

derekm:
> > Since this breaks backwards compatibility of existing deployments that might seek to upgrade, can it include an AWK, sed or Perl script that performs all the necessary substitutions on users' custom config files?
> 
> The intent is not to break backward compatibility. It supports backward compatibility, so that old configuration keys/names can be migrated to new ones asynchronously (and manually). Old names continue to work. Also, old and new names can coexist within the same configuration file.
> 
> That being said, it'd still be a good idea to provide a script that migrates old config names to new ones, for people that may want to modify their existing config files to use new names instead of the old ones. Maybe as a separate PR later, as:
> 
> * It is not strictly necessary for this PR.
> * We also need to make config file names consistent across controller and segment store, which we plan to do separately in a different PR. That'll require rework on the scripts if we do the scripts now.
> 
> @derekm what do you say?

I must have misread the PR when I made my comment.

I wouldn't think this PR needs to be blocked by a migration script if old keys are honored. When I made my comment, I was under the assumption that there was no fall back to the old config key names, but at second glance, sure enough, they old keys are

Sorry for the confusion! I agree with you, migration script is a nice to have, but with old keys being honored, it isn't mandatory.

ravisharda:
I've created issue #4834 for removing deprecated/old configuration keys in the future.

config/config.properties
```diff
@@ -20,162 +20,162 @@
...
 
 # Storage NO-OP Mode: in No-Op mode, user stream segment writing is no-oped; user stream segment reading is not supported.
 # This mode is used to avoid storage interference in testing while still keep the system functioning as usual.
-# NOTE: pravegaservice.storageImplementation is still used to store metadata and system segments, which are required for the functioning of the Pravega Cluster.
+# NOTE: pravegaservice.storage.impl.name is still used to store metadata and system segments, which are required for the functioning of the Pravega Cluster.
 # WARNING: Do NOT set NO-OP Mode to true in production environment! Data loss will happen!
 # Valid values: true, false
 # Default value: false
-#storageextra.storageNoOpMode=false
+#storageextra.noOp.mode.enable=false
```
RaulGracia, line 100: Since this was introduced, I think that this could me more understandable as `pravegaservice.storage.noOp.mode.enable`.

config/config.properties
```diff
@@ -20,162 +20,162 @@
...
 
 # When Storage NO-OP Mode is true, storageWriteNoOpModeLatencyMillis is applied to write operation.
 # It is used to compensate the supposed latency if the storage is in operation, in order to make the test as real as possible.
 # This latency has no effect if storage is not in No-Op mode.
 # Valid values: non-negative integer
 # Default vale: 20 (milliseconds)
-#storageextra.storageWriteNoOpLatencyMillis=20
+#noOp.write.latency.milliseconds=20
```
RaulGracia, line 107: Why this does not have the same structure as the previous property? It refers to the same, so I think the structure should be the same: `storageextra.noOp.write.latency.milliseconds` (unless you replace `storageextra` -> `pravegaservice.storage`, which I think it is easier to understand)

config/config.properties
```diff
@@ -469,13 +469,13 @@ extendeds3.configUri=http://localhost:9020?identity=user&secretKey=password
 
 ##region DurableLog Settings
 
-# A Metadata Checkpoint is a special internal write inside the DurableLog (persisted to Tier1 DurableDataLog) that takes
+# A Metadata Checkpoint is a special internal write inside the DurableLog (persisted to Durable Data Log) that takes
```
RaulGracia, line 472: I'm aware that is better to replace Tier 2 -> Long Term Storage for clarity. But I'm not sure whether making Tier 1 to disappear as well is making things clearer or not (is an open question).

controller/src/test/resources/controller.config.properties
```diff
@@ -12,38 +12,38 @@
 # The commented out settings are not different than the hardcoded defaults in Config.java.
 
 
-controller.containerCount=4
```
RaulGracia, line 15: Now that you are addressing the configuration of Pravega, I ask the following: could we get rid of this file completely? I mean, I have always wondered why are there 2 separate files for configuration, one for the Segment Store and another for the Controller. Would it be better to have a single file with different "sections"? Maybe this is something to discuss with @shiveshr and @andreipaduroiu, but from an usability perspective, I don't know if doubling the number of configuration files is going to make things better for users.

controller/src/conf/controller.config.properties
```diff
@@ -18,49 +18,52 @@
 #                java -Dcontroller.service.port=5689 ...
 #          which will set the value for that setting to 5689, regardless of what this file contains.
 
+controller.cluster.name=${CLUSTER_NAME}
+controller.container.count=${CONTAINER_COUNT}
+controller.hostMonitor.enable=${HOST_MONITOR_ENABLE}
+controller.rebalance.interval.seconds.min=${CLUSTER_MIN_REBALANCE_INTERVAL}
 
-controller.containerCount=${CONTAINER_COUNT}
```
RaulGracia, line 22: Same here as in the previous comment related to split controller and Segment Store config in different files.

config/config.properties
andreipaduroiu: We've agreed to eliminate the Tier 1 term as well. DurableLog/DurableDataLog is the term to use to be consistent with the code.

config/config.properties
andreipaduroiu: This smells like a bug. This value won't be picked up by the code.

config/config.properties
ravisharda: This config property is defined in `StorageExtraConfig.java` that has a component code `storageextra`, which is resulting in the `storageextra` prefix. Refactoring it to bring it under `pravegservice` (`ServiceConfig.java`) would require kind of changes that are out-of-scope of this PR. We only intend to change configuration property keys/names here, and that too in a backward compatible manner.

controller/src/test/resources/controller.config.properties
ravisharda: At the outset, I think its a good idea to have a single config file. I've created a separate issue #4843 for this item, as I don't think its in the scope of this PR.

controller/src/conf/controller.config.properties
ravisharda: https://github.com/pravega/pravega/pull/4713#discussion_r434296932

config/config.properties
ravisharda: Yes, that's indeed a bug. I've fixed it now. Thanks for spotting it in [this](https://github.com/pravega/pravega/pull/4713/commits/5481973b7a13aae092d96fa652a9032df33e7074) commit.

config/config.properties
ravisharda: So, I'm leaving it as is. @RaulGracia / @andreipaduroiu hope that is OK.

ravisharda:
> Please fix OutOfProcessAdapter.java. It's setting some values directly.
> Also search for any other places in the code that might be hardcoding these values as strings.

@andreipaduroiu I didn't set them in this PR, and believe replacing those hardcoded values is out-of-scope of this PR. Maybe we should tackle that in a separate PR.

RaulGracia:
@ravisharda please execute system tests on this branch before merging this one ok?

andreipaduroiu:
That is not "out of scope" for this PR. If you make such a change across the board, you should be fixing all places within your reach that still use the old values.

ravisharda:
> That is not "out of scope" for this PR. If you make such a change across the board, you should be fixing all places within your reach that still use the old values.

Ah, I later realized what change you were referring to. I had confused it with a numeric non-configuration variable used in that file. I've now modified the adapter and some other files. Also, I had already modified configuration keys elsewhere (including shell scripts).

ravisharda:
I've triggered a system test run using the branch builder, with both security enabled and disabled. The job id is: 2509.

ravisharda:
FYI, the system test run with the latest code passed. Branch builder job id 2509. Test Pravega Continuous job id 2899.

PR #4719, opened 2020-04-22
Issue-4642: load TLS certificates from /etc/secret-volume/ca-bundle

### LINKED ISSUE #4642: Pravega to support self-signed certificate of S3 Storage service

kevinhan88:

**Change log description**  
Now Pravega supports ECS as tier2. If an ECS internal installation is only self-signed for TLS, Pravega must trust its self-signed certificate or the signing CA in order to establish TLS connection.
Per the discussion and agreement, ECS certificate will be mounted to /etc/secret-volume/ca-bundle, hence Pravega container script needs to find the certificates from this location and then add them into truststore.

**Purpose of the change**  
Closes #4642 

**What the code does**  
The container script loads certificates found under /etc/secret-volume/ca-bundle into the default Java Truststore in order to establish TLS connection with ECS

**How to verify it**  
ECS certificate should show up at /etc/secret-volume/ca-bundle, and ECS TLS works.

### DIFF
--- a/docker/pravega/scripts/common.sh
+++ b/docker/pravega/scripts/common.sh
@@ -41,5 +41,4 @@ add_certs_into_truststore() {
     do
       yes | keytool -importcert -storepass changeit -file "${cert}" -alias "${cert}" -keystore /etc/ssl/certs/java/cacerts || true
     done
-    rm -rf /etc/secret-volume/ca-bundle/* || true
 }
--- a/docker/pravega/scripts/common.sh
+++ b/docker/pravega/scripts/common.sh
@@ -39,7 +39,7 @@ add_certs_into_truststore() {
     CERTS=/etc/secret-volume/ca-bundle/*
     for cert in $CERTS
     do
-      yes | keytool -importcert -storepass changeit -file "${cert}" -keystore /etc/ssl/certs/java/cacerts || true
+      yes | keytool -importcert -storepass changeit -file "${cert}" -alias "${cert}" -keystore /etc/ssl/certs/java/cacerts || true
     done
     rm -rf /etc/secret-volume/ca-bundle/* || true
 }
--- a/docker/pravega/scripts/common.sh
+++ b/docker/pravega/scripts/common.sh
@@ -41,4 +41,5 @@ add_certs_into_truststore() {
     do
       yes | keytool -importcert -storepass changeit -file "${cert}" -keystore /etc/ssl/certs/java/cacerts || true
     done
+    rm -rf /etc/secret-volume/ca-bundle/* || true
 }
--- a/docker/pravega/scripts/common.sh
+++ b/docker/pravega/scripts/common.sh
@@ -36,7 +36,7 @@ add_system_property_ecs_config_uri() {
 
 # Add ECS certificates into Java truststore
 add_certs_into_truststore() {
-    CERTS=/etc/ssl/certs/java/ecs-certs/*
+    CERTS=/etc/secret-volume/ca-bundle/*
     for cert in $CERTS
     do
       yes | keytool -importcert -storepass changeit -file "${cert}" -keystore /etc/ssl/certs/java/cacerts || true
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentStoreMetrics.java
@@ -21,7 +21,10 @@ import io.pravega.shared.metrics.OpStatsLogger;
 import io.pravega.shared.metrics.StatsLogger;
 import java.time.Duration;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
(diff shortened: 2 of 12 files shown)


docker/pravega/scripts/common.sh
```diff
@@ -36,7 +36,7 @@ add_system_property_ecs_config_uri() {
 
 # Add ECS certificates into Java truststore
 add_certs_into_truststore() {
-    CERTS=/etc/ssl/certs/java/ecs-certs/*
+    CERTS=/etc/secret-volume/ca-bundle/*
```
ravisharda, line 39: Curious, what is this `ca-bundle` directory under `/etc/secret-volume`? It's not clear to me from the PR description. Is it the certificate of the CA that was used to sign the ECS server certificates?

Could we use a more specific name? Not sure what the ca-bundle represents exactly.

ravisharda: Ok, I found the context in the corresponding issue: #4642. I'd suggest you add a bit of context to the PR description too, for future reference.

docker/pravega/scripts/common.sh
```diff
@@ -36,7 +36,7 @@ add_system_property_ecs_config_uri() {
...
     for cert in $CERTS
     do
       yes | keytool -importcert -storepass changeit -file "${cert}" -keystore /etc/ssl/certs/java/cacerts || true
```
ravisharda, line 42: Should we delete the certificate after successful import into the keystone?

kevinhan88: added line to delete certs afterwards.

docker/pravega/scripts/common.sh
kevinhan88: PR description updated.

docker/pravega/scripts/common.sh
```diff
@@ -36,9 +36,10 @@ add_system_property_ecs_config_uri() {
...
     done
+    rm -rf /etc/secret-volume/ca-bundle/* || true
```
maddisondavid, line 44: This is unrequired and potentially dangerous.  If Pravega is running bare metal then this will delete any certificates that were placed in `/etc/secret-volume/ca-bundle`.  If this is running in Kubernetes then it would be a secret and you can't delete the certificates from the secret.

Is there a reason this was added?

docker/pravega/scripts/common.sh
```diff
@@ -36,9 +36,10 @@ add_system_property_ecs_config_uri() {
 
 # Add ECS certificates into Java truststore
 add_certs_into_truststore() {
-    CERTS=/etc/ssl/certs/java/ecs-certs/*
+    CERTS=/etc/secret-volume/ca-bundle/*
     for cert in $CERTS
     do
       yes | keytool -importcert -storepass changeit -file "${cert}" -keystore /etc/ssl/certs/java/cacerts || true
```
maddisondavid, line 42: Does this work with multiple certificates?  I was under the impression from other tests I've done with `keytool` that you need to specify an alias otherwise the certificate gets added as `<mykey>` which isn't a trusted entry.

docker/pravega/scripts/common.sh
kevinhan88: This is per our security specialist Ravi's comment.
Functional-wise, it's tested okay to delete certificates after adding them into truststore. Those certificates have been mounted into the pod already so it wont affect k8s secrets.

docker/pravega/scripts/common.sh
kevinhan88: Good point. Added file name as alias to avoid the potential conflict.
(OpenJDK's default truststore also uses filename as alias)

docker/pravega/scripts/common.sh
ravisharda: Actually, @maddisondavid is right. You shouldn't delete a file-mounted K8s secret using an OS command. If you manage to do it, it'll probably automatically be recreated (I am not 100% sure). Maybe we should leave it as is as @maddisondavid suggests.

ravisharda:
LGTM. @kevinhan88 you might want to address @maddisondavid 's comment about avoiding K8s  secrets using OS command, before this PR is approved by the rest of the reviewers and merged.

docker/pravega/scripts/common.sh
kevinhan88: fixed

PR #4744, opened 2020-04-28
Issue 4743: (SegmentStore) Infinite loop in StorageWriter if deleting a segment while flushing.

### LINKED ISSUE #4743: StorageWriter can enter infinite loop if segment is deleted while flushing.
**Problem description**
If a segment is deleted while flushing, the StorageWriter can enter in an infinite loop.

**Problem location**
SegmentAggregator.

Abort the flush if the segment has been deleted.

andreipaduroiu:
**Change log description**  
Fixed a bug in SegmentAggregator where it was possible to enter an infinite loop if deleting a segment while flushing.

**Purpose of the change**  
Fixes #4743.

**What the code does**  
Exiting the flush loop if a segment deletion is detected.

**How to verify it**  
New unit test added. The unit test would have never finished (timed out) if the fix hadn't been there.


RaulGracia:
We have verified this fix with high-load performance benchmarks involving Stream deletions and system test are passing.

PR #4745, opened 2020-04-28
Issue 4680: Enable truncation of Byte Stream.

### LINKED ISSUE #4680: On demand truncation of a byte stream.
**Problem description**
It is a valid case that an application needs to actively reduce the storage usage for old data in a stream that is no longer used for sure. Basically we can rely on the retention policy to do so, but the truncation triggered by retention policy is out of control from the application who consumes the stream. For example, if I need to save some data that I'd like to consume only once, after I consumed the data, it can be removed from the storage. Meanwhile I don't want to define a scope either by size or event count for retention purpose because I cannot affort to lose any data that I haven't consumed yet. 
The truncation APIs now serves as the function to explictly truncate a stream from a point described as a StreamCut. Now days we have several ways to truncate a stream on demand. 

1. Through StreamManager::truncateStream which only accepts a StreamCut.
2. Through RevisionedStreamClient::truncateToRevision if we versionize the events we read.
3. Implict truncation caused by StateSynchronizer::compact() action

Unfortunately, none of the above truncation methods fits the "ByteStream", whose design goal is to read/write a stream without any framing or demarcation of events, and it is consumed by ByteStreamReader rather than ReaderGroups since we read byte stream from one offset to another as "bytes" rather than "events". In order to truncate a byte stream we currently has to do some tricky by following:
1. Create a ReaderGroup and points the reader group to the byte stream we'd like to truncate
2. Call ReaderGroup::getStreamCuts() which is supposed to contain only one StreamCut
3. Call StreamCut::getPositions() to list position map which is supposed to contain only one Position
4. Extract the Segment object from the Position object we get from 3.
5. Construct a StreamCutImpl object by our own, put the Segment object we get from 4., along with Offset where we'd like to do the truncation.
6. Pass the StreamCutImpl object to StreamManager::truncateStream() to do the real truncation job.

This is not a convenient way we think it should be, and it exposes some inside implementations which disobeys the rule that we shall follow to communicate with Pravega using Pravega Client.

**Problem location**
Pravega Client (maybe the controller need to support such a call as well)?

Pravega client add a truncation function based on "Offset", dedicated for "Byte Streams" which is one segment only and do not have framing/demarcation of events.

shrids:
**Change log description**  
- Add new truncation api to ByteStreamWriter 

**Purpose of the change**  
Fixes #4680 

**What the code does**  
Add a truncation API to ByteStreamWriter 
```
public void truncateDataBefore(long offset);
````

**How to verify it**  
All the existing and newly added unit and integration tests should continue to pass.


shrids:
I have created issue https://github.com/pravega/pravega/issues/4777 for implementing a system test.

PR #4749, opened 2020-04-28
Issue 4737: Controller's Stream Workflows can throw StartException if the input is not found, leading to wasteful compute cycles 

### LINKED ISSUE #4737: Controller's Stream Workflows can throw StartException if the input is not found, leading to wasteful compute cycles
**Problem description**
Controller event processing tasks throw StartException if they do not find the corresponding input for the task in metadata table. The StartException is retryable and results in the whole processing for the event being retried. 
This is wasteful because input is ALWAYS written before writing the event. So absence of input implies completion of processing. 
PS: auto scale is the only exception here.  

**Problem location**
Stream Tasks in event processor.

Do not throw StartException if input is not found.

shiveshr:

**Change log description**  
Remove StartException as it is unnecessary as we always update metadata and then write the event. 
So not finding metadata should not throw StartException which leads to wasted compute retries. 

**Purpose of the change**  
Fixes #4737 

**What the code does**  
Controller event processing tasks throw StartException if they do not find the corresponding input for the task in metadata table. This leads to workflows being retried for sometime before failing. 
This is unnecessary because we always write metadata and then post the event. So if event is picked for processing and metadata does not match the expected state, it is because the processing is already complete. So we do not need to unnecessarily retry in such situations. 

Changes:
1. Removed StartException class. Removed Postpone logic based on StartException from stream request handler. 
2. Removed its usage in UpdateStreamTask: if StreamConfigurationRecord.isUpdating = false, complete the task treating it as idempotent case
3. Removed its usage in TruncateStreamTask: if StreamTrunctionRecord.isUpdating = false, complete the task treating it as idempotent case
4. Removed its usage in ScaleOperationTask: if Epoch Transition is empty for manual scale, scale is already complete and this is idempotent case
5. Removed its usage in SealStreamTask: if stream.state != SEALING/SEALED, throw illegal state exception. 

**How to verify it**  
Unit tests updated


shrids:
System test run is successful `/job/test-pravega-continuous-1/685/`

PR #4756, opened 2020-04-29
Issue 4736: Run scale workflow only if input matches the existing Epoch transition record

### LINKED ISSUE #4736: Scale Operation Task should log processing logs with respect to the input it gets Epoch Transition Record
**Problem description**
ScaleOperationTask uses the input in EpochTransitionRecord to process an event but logs the value corresponding to the scale input (which is there for purposes of auto scaling requests where we call `submitScale` first before processing it. 

So logging should reflect the actual processing that was performed against the event. 

**Problem location**
ScaleOperationTask.java

Fix the logging in ScaleOperationTask

shiveshr:

**Change log description**  
Ensure that scale workflow for an event is processed for that event and if the stream metadata (epoch transition record) is different from the scale event then exit the workflow. 

**Purpose of the change**  
Fixes #4736

**What the code does**  
ScaleOperationTask runs the scale workflow by checking if the epoch transition record exists or not. 
If it exists, it should compare the scale input and only process the event if the input is same as epoch transition record. 
For manual scale, since the epoch transition record is written before the event is posted, so if the epoch transition record is different from the event, then the event has already been processed and it is idempotent case so complete the workflow. 
For auto scale case, if the epoch transition record exists, it means a different scale is being processed, so throw conflict exception so that this event is postponed and retried later. 

**How to verify it**  
Unit test added.


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/ScaleOperationTask.java
```diff
@@ -126,7 +127,19 @@ public ScaleOperationTask(final StreamMetadataTasks streamMetadataTasks,
...
+                            // ensure that we process the event only if the input matches the epoch transition record.
+                            // if its manual scale and the event doesnt match, it means the scale has already completed. 
+                            // for auto scale, it means another scale is on going and just throw scale conflict exception 
+                            // which will result in this event being postponed. 
+                            if (isManualScale) {
+                                log.info("Scale already completed.");
+                                return CompletableFuture.completedFuture(null);
+                            } else {
+                                log.info("Scale conflict.");
+                                throw new EpochTransitionOperationExceptions.ConflictException();
```
shrids, line 140: Throwing a ConfilictException cause the writeBackPredicate of `processScaleOpRequest` to be true causing the event to written back, which effectively postpones the event, right ?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/ScaleOperationTask.java
```diff
@@ -126,7 +127,19 @@ public ScaleOperationTask(final StreamMetadataTasks streamMetadataTasks,
...
                                         new ArrayList<>(scaleInput.getNewRanges()), scaleInput.getScaleTime(), record, context, executor));
                             }
-                        } 
+                        } else if (!RecordHelper.verifyRecordMatchesInput(scaleInput.getSegmentsToSeal(), scaleInput.getNewRanges(), isManualScale, record.getObject())) {
+                            // ensure that we process the event only if the input matches the epoch transition record.
+                            // if its manual scale and the event doesnt match, it means the scale has already completed. 
+                            // for auto scale, it means another scale is on going and just throw scale conflict exception 
+                            // which will result in this event being postponed. 
+                            if (isManualScale) {
+                                log.info("Scale already completed.");
```
shrids, line 136: The log can have additional data about the scale operation.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/ScaleOperationTask.java
```diff
@@ -126,7 +127,19 @@ public ScaleOperationTask(final StreamMetadataTasks streamMetadataTasks,
...
+                                return CompletableFuture.completedFuture(null);
+                            } else {
+                                log.info("Scale conflict.");
```
shrids, line 139: Scale conflict already causes a WARN message to be thrown, Can we add more details to this log line.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/ScaleOperationTask.java
shiveshr: Yes. And that is the desired behaviour.

PR #4763, opened 2020-04-30
Issue 4656: (KeyValue Tables) Sorted Table Segments

### LINKED ISSUE #4656: Key-Value Tables: Segment Iterator Filters
**Problem description**
Implement Prefix filters in Table Segment Iterators (both key and entry iterators).
Depends on: #4333.

**Problem location**
Segment Store.

TBD

andreipaduroiu:
**Change log description**  
- Added a new type of Table Segment that can iterate through Keys/Entries in a bitwise lexicographical order.
    - Existing Table Segment is untouched and provides the same functionality.

**Purpose of the change**  
Fixes #4656.  

**What the code does**  
TableSegments (TS) are a hash-table based structure that keep their indices in Segment Attributes. However they provide no order between keys, making range queries impossible. 

Sorted Table Segments (STS) are an **extension** to Table Segments in the following way:
- STS can TS can coexist side-by-side. TSs have not been touched or affected in any way by the existences of STSs.
- STS is built on top of TS using the following scheme:
    - All Keys are additionally stored in a _B+Tree Set_ (BTS - see below) that provides order between keys. 
    - The BTS stores its data (nodes) alongside the original TS data, but with its keys being differentiated from external keys. BTS keys are prefixed with a byte (`I`) denoting they're **I**nternal, while external keys are prefixed with a byte (`E`) denoting they're **E**xternal. No external query can access internal keys, so they cannot leak out.
    - STS delegates all operations to the underlying TS implementation exception Key/Entry iterators. Iterators are delegated to the BTS iterator, which provides ordering and range queries. 
        - This is the only difference between TSs and STSs. They both function as a Hash Table, but STSs allow for iterating keys in order.
    - For TS, the `WriterTableProcessor` is the one that updates the index, so it knows which keys are inserted or removed. For STSs, this component is the one that updates the BTS, and it only does so when it detects a new key or that a key has been removed.
    - The `ContainerKeyIndex` component keeps a small tail cache of keys in sorted order (see below) that aids iterators for the tail (unidexed) section of the segment.

**B+Tree Set** (BTS)
- This is a an ordinary B+Tree but with no values. Only Keys are stored. 
- Wholly contained in `io.pravega.common.util.btree.sets`
    - Not to be confused with `io.pravega.common.util.btree.BTreeIndex` which is a fixed-length KV-pair on append-only storage index.
- The only operations implemented are: Update (Insert or Remove) and iterate.
- This follows the typical B+Tree implementation with node splits, but it does not do mergers (yet). It only removes a node when it becomes empty (which is how the tree shrinks). 
- It supports variable-length keys with a maximum node size of 1MB.
- Since it does not have to store data directly in Durable Log (formerly known as Tier 1) or Long-Term Storage (formerly known as Tier 2), it does not have to worry about how to store or compact data; the underlying TS takes care of that.

Integrating **B+TreeSet** with **Table Segments** to form **Sorted Table Segments**:
- The `SortedKeyIndex` component has been created, and it only is present for STSs. 
    - This keeps an instance of the BTS, and updates it from the `WriterTableProcessor` (when it performs final indexing) and queries it during iteration. 
    - It also keeps an in-memory `TreeSet` for the tail cache, which it cleans very aggressively as the `WriterTableProcessor` makes indexing progress; it uses this to complete iterator queries.
- The `SortedKeyIndex` instance is shared between the `WriterTableProcessor` and the "front-end" `ContainerKeyIndex`; it provides a unified view of the STS keys.

Resource and performance considerations:
- The presence of STSs in the Segment store will require more resources and more CPU utilization in order to maintain the B+TreeSet and sorted tail index. 
    - No impact is anticipated for cases where STSs aren't used - regular Table Segments are not affected by the new code.
- The BTS is only updated for new and removed keys. Updates to existing keys do not touch the BTS index, so this should only take up processing resources for STSs where we have a high degree of key churn.
- The sorted tail cache now stores the tail keys in Java heap arrays. This may increase the heap requirements, however the throttling mechanisms inside the segment store will slow down the intake rate of new update requests if LTS and `WriterTableProcessor` falls too much behind, so this should be mostly contained.

**How to verify it**  
Extensive new unit tests added that verify new functionality.
Existing unit tests validate that no existing functionality was affected.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentSortedKeyIndex.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Includes the given {@link TableKeyBatch} which contains Keys that have recently been updated or removed, but not
+     * yet indexed. These will be stored in a memory data structure until {@link #updateSegmentIndexOffset} will be invoked
+     * with an offset that exceeds their offset.
+     *
+     * @param batch              The {@link TableKeyBatch} to include.
+     * @param batchSegmentOffset The offset of the {@link TableKeyBatch}.
+     */
+    void includeTailUpdate(TableKeyBatch batch, long batchSegmentOffset);
```
fpj, line 46: The next three calls are to maintain a tail cache, if I understand it right. By reading these, it is not clear to me the expected flow of these calls for a user of this interface, and it is also not not clear what the consistency guarantee is.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableService.java
```diff
@@ -48,9 +48,9 @@ public TableService(SegmentContainerRegistry segmentContainerRegistry, SegmentTo
     //region TableStore Implementation
 
     @Override
-    public CompletableFuture<Void> createSegment(String segmentName, Duration timeout) {
+    public CompletableFuture<Void> createSegment(String segmentName, boolean sorted, Duration timeout) {
```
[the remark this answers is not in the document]
andreipaduroiu: We could very well do that, but sorting doesn't come for free. There will be additional space requirement (keep all keys in sorted order) and some computational overhead to maintain that sorted index. Not all Table Segments may want that. If all they want is a Hash-Table-like structure with no order between keys, then regular Table Segments will do the job without the overhead of sorted index.

Additionally, there is no upgrade path for Hash Table Segments to Sorted Table Segments. IF we converted all HTS into STS, then we'd have to auto-upgrade existing Segments. That would be another challenge by its own.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentSortedKeyIndex.java
```diff
@@ -0,0 +1,141 @@
...
+interface SegmentSortedKeyIndex {
+    /**
+     * Include and persist updates that have been included in a Table Segment Index.
+     *
+     * @param bucketUpdates A Collection of {@link BucketUpdate} instances that reflect what keys have been added and/or
+     *                      removed.
+     * @param timeout       Timeout for the operation.
+     * @return A CompletableFuture that, when completed, will indicate that the operation has completed.
+     */
+    CompletableFuture<Void> persistUpdate(Collection<BucketUpdate> bucketUpdates, Duration timeout);
```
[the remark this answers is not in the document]
andreipaduroiu: The Sorted Key Index (SKI) is updated just like the regular Container Key Index (CKI). There is a tail section (you found it below), which gets updated every time we get a user update. The non-tail section (indexed section) gets updated by the WriterTableProcessor when it indexes the next entries (to be indexed) from the tail. At this point, the tail is drained of those Keys, which are then only available in the indexed portion.

We update the SKI (via tail portion below) before we ack to the user. As such, the Key will be visible to the user both via the regular (get) API and in the Key/Entry iterator; there will be no inconsistency. The `persistUpdate` call is done in the background and is completely transparent to upstream code, and hence to the user.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentSortedKeyIndex.java
andreipaduroiu: This mirrors `ContainerKeyIndex` and `ContainerKeyCache`. 
- `includeTailUpdate` is called every time we insert, update or remove a key (from the API)
- `includeTailCache` is called (optionally) after a container failover, if someone wants to access the Segment **prior to** it being fully indexed by the `WriterTableProcessor` (via `persistUpdate`. 
    - Why is this important? Because the in-memory part is volatile; it is gone after a crash. We accumulate data into it via `includeTailUpdate`, but that information is gone after a crash. There is special code in the `ContainerKeyIndex` (pre-existing; not part of this PR) that auto-recovers a Table Segment if it is accessed immediately after a recovery but before it is indexed. (If you remember, we had a few problems last year when the Controller was complaining that Table Segment data was unavailable shortly after a recovery - this pre-recovery was introduced to handle that kind of problems).
- `updateSegmentIndexOffset` is called by the `WriterTableProcessor` (indirectly, via `ContainerKeyIndex` to tell the SKI that everything prior to a certain offset has been indexed and it can remove it from the "tail cache" (i.e., if anyone needs that info, they can get it from the actual index, and no need to keep a heap data structure filled with redundant info).


Does this explanation make sense? These are all pre-existing concepts, and have just been applied to the SKI. The consumers of this interface are the `ContainerKeyIndex`, `WriterTableProcessor` which work in tandem to offer a unified view of the TableSegment to the upstream code.

fpj: What actually got me to ask about it is whether this distinction between cached vs. indexed  belongs in the interface or whether it is an implementation detail leaking to the interface. But, if this is symmetric to an existing concept, then there is no point in arguing about it for this PR. At best, we should think about it separately.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentSortedKeyIndex.java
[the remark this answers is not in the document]
fpj: Ok, makes sense.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableService.java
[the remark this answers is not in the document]
fpj: I guess we don't have a concrete answer to this, but I'm wondering about the recommendation we would give about the use of STS, like what's a reasonable number of sorted tables or what is a reasonable amount of traffic to guide users. 

The migration is indeed a concern.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -269,16 +311,49 @@ public void close() {
...
+        return item == null ? null : translateItem.apply(translator, item);
+    }
+
     @Override
     public CompletableFuture<AsyncIterator<IteratorItem<TableKey>>> keyIterator(String segmentName, IteratorArgs args) {
-        logRequest("keyIterator", segmentName);
-        return newIterator(segmentName, args, TableBucketReader::key);
+        return this.segmentContainer.forSegment(segmentName, args.getFetchTimeout())
+                .thenComposeAsync(segment -> {
+                    if (ContainerSortedKeyIndex.isSortedTableSegment(segment.getInfo())) {
```
fpj, line 334: If ai understand the flow right, `ContainerTableExtensionImpl` implements `ContainerTableExtension`, which implements `TableStore`. `keyIterator` and `entryIterator` are calls in the `TableStore` interface, so that's how code using table segments use the sorted table segments.

`SegmentSortedKeyIndex` is made available via `ContainerKeyIndex`, which is used in this class. This is how the `TableStore` interface connects with the `SegmentSortedKeyIndex`, is it right?

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
```diff
@@ -0,0 +1,414 @@
...
+    //region Queries
+
+    /**
+     * Returns an {@link AsyncIterator} that will iterate through all the items in this {@link BTreeSet} within the
+     * specified bounds. All iterated items will be returned in lexicographic order (smallest to largest).
+     * See {@link ByteArrayComparator} for ordering details.
+     *
+     * @param firstItem          An {@link ArrayView} indicating the first Item to iterate from. If null, the iteration
+     *                           will begin with the first item in the index.
+     * @param firstItemInclusive If true, firstIem will be included in the iteration (provided it exists), otherwise it
```
fpj, line 296: I'm curious, why you think it is a good idea to have booleans for inclusion rather than pick one option and go with it?

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSetPage.java
```diff
@@ -0,0 +1,879 @@
...
+ * Represents a Page (Node) within a {@link BTreeSet}. Pages can be of type {@link IndexPage} or {@link LeafPage}.
+ */
+@NotThreadSafe
+abstract class BTreeSetPage {
+    //region Serialization format
+
+    /**
+     * Format Version related fields. The version itself is the first byte of the serialization. When we will have to
+     * support multiple versions, we will need to read this byte and choose the appropriate deserialization approach.
+     * We cannot use VersionedSerializer in here - doing so would prevent us from efficiently querying and modifying the
```
fpj, line 42: It is not great that we have to create an exception for the serialization versioning. I understand that the concern is the overhead of serializing, but could you elaborate on why you believe this is concern that justifies creating an exception for not using `VersionedSerializer`? Is there a way of using `VersionedSerializer` that could solve the issue, e.g., changing a version such that it is not backwards compatible when the overhead is high?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableService.java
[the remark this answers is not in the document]
andreipaduroiu: For now, I think we should keep all metadata (Segment Store and Controller) in hash tables - there is no need for sorted keys there. 

For KVTs, we will use STSs because we want to use iterators. In the (near) future, we can add an option at the creation of KVTs that says "I don't need iterators" which should result in the use of Hash Table Segments as backing store (the API is identical between STS and HTS). 

* I say "near future" because KVTs are going out as **Beta**, which gives us certain freedoms in altering the API without need for backwards compatibility while having this tag.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentSortedKeyIndex.java
andreipaduroiu: While this is an interface, it's an internal one, and the whole point of it was to be able to mock it during unit tests. This is buried so deep within the stack that it will be hard for anyone to surface this if they want to. 

Yes, it exposes details about the implementation, but this **is** the implementation.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
```diff
@@ -0,0 +1,414 @@
...
+/**
+ * A B+Tree-backed Set. Stores all items in a B+Tree Structure using a {@link ByteArrayComparator} for ordering them.
+ *
+ * NOTE: This component is in {@link Beta}. There are no guarantees about data or API compatibility with future versions.
+ * Any component that is directly dependent on this one should either be in {@link Beta} as well.
+ */
+@NotThreadSafe
+@Beta
+@Slf4j
+public class BTreeSet {
```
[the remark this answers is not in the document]
andreipaduroiu: If I haven't already, we should file a follow-up issue to unify as much as we can between the existing BTreeIndex and this BTreeSet.

I tried to do that during development of this class, however they are so different that I preferred to keep them apart. 

Why are they different?
- BTreeIndex is a Key-Value index with all keys of the same length and all values of the same length. BTree set has no values and allows variable-length keys.
    - This allows BTreeIndex Nodes/Pages to more efficiently pack the data together (which is necessary (see next bullet)) and execute binary searches for lookups. For BTreeSet I have to keep a header with pointers to data so I can "jump" to an entry.
- BTreeIndex is meant to store data on an append-only medium, while BTreeSet couldn't care less (literally). BtreeIndex is stored directly on Tier2/LTS, while BTreeSet uses the Hash Table Segment itself to store nodes (each node is a Key-Value entry in the Table Segment it keeps keys for). As such, BTreeSet does a lot less work in managing data integrity and compaction (it does none of that!) since it delegates all of it to the Table Segment underneath it.
    - Due to the fact that BTreeIndex stores on Append only media, it has to deal with write amplification so the amount of data that we need to write must be kept to a minimum, hence the branching factor of the node. BTreeIndex has a max node/page size of 32KB, which gives it a branching factor of about 1000 (more or less), thus reducing the depth of the tree and the size of the writes on updates.
    - BTreeSet has no such limitation so its max page/node size is whatever the TableSegment allows as values (1MB as of today). Given that it has variable-length keys, the branching factor will vary, however there is a lower bound on it due to the fact that keys themselves have an upper bound on their length.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
andreipaduroiu: First paragraph is 100% spot on.

Second paragraph:
- `SortedKeyIndex` is indeed exposed via `ContainerKeyIndex`, but it is only used in `ContainerTableExtensionImpl`, which, depending on whether the Segment is Sorted or not, uses it or uses the default iterator. 
- The connection is done in the `ContainerTableExtensionImpl` and again (in the back-end) in `WriterTableProcessor` that invokes its `persistUpdate` API to insert or remove keys (only when keys are inserted or removed, but not on update).

* Clarification in case I didn't explain it properly: the BTreeSet behind the SKI is only updated when a Key is inserted or removed. It is not touched when a Key is updated. Hence the frequency of writes to that data structure is a function of inserts or deletes, and not in-place updates.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
andreipaduroiu: Because sometimes I need to iterate starting with an element and sometimes I need to iterate immediately after an element (i.e., I know the last element I processed, so give me the next one).

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSetPage.java
andreipaduroiu: The problem with VersionedSerializer is that it forces me to read the whole input (buffer) and store it in another buffer. That is, it reads the data before it gives it to me. 

In this case, since I'm always dealing with huge chunks of data, reading it (and copying it, since that's what VersionedSerializer does) will add extra CPU and GC overhead that will slow down the whole process.

This reason alone was enough for me to choose not to use the convenience of `VersionedSerializer` and use a version flag instead.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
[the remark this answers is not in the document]
fpj: Ok, I searched and couldn't find an issue, so I created #4840, and assigned it to you for now, feel free to reassign as needed. You can resolve this thread once you see this message, I want to make sure you see the created issue.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/tables/IteratorArgs.java
```diff
@@ -24,9 +24,11 @@
 @Builder
 public class IteratorArgs {
     /**
+     * EXPERIMENTAL!
```
fpj, line 27: To clarify, is it only the `prefixFilter` that is experimental? Basically, `getpPrefixFilter` returns an `ArrayView` that is not guaranteed to be correct? Do you know whether this is displayed correctly with javadocs (because of the @Data lombok annotation)?

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/tables/TableStore.java
```diff
@@ -41,8 +39,19 @@
  * will be atomically checked-and-applied.
  * * Unconditional Updates (insert, update, remove) will take effect regardless of what the current Key version exists in
  * the Table Segment.
+ *
+ * Sorted vs Non-Sorted Table Segments:
+ * * All Table Segments are a Hash-Table-like data structure, where Keys are mapped to Values.
+ * * Non-Sorted Table Segments provide no ordering guarantees for {@link #keyIterator} or {@link #entryIterator}.
+ * * Sorted Table Segments store additional information about the Keys and will return results for {@link #keyIterator}
+ * or {@link #entryIterator} in lexicographic bitwise order. All other contracts are identical to the Non-Sorted variant.
+ * * Sorted Table Segments will require additional storage space to store the ordered Keys and may require additional
```
fpj, line 48: Do you want to add that this additional space is proportional to the number and length of keys, but not the size of the value?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerKeyIndex.java
```diff
@@ -538,6 +546,25 @@ void notifyIndexOffsetChanged(long segmentId, long indexOffset) {
...
+     * Gets the {@link SegmentSortedKeyIndex} associated with the given Segment, as provided by
+     * {@link ContainerSortedKeyIndex#getSortedKeyIndex}. This can be used to safely iterate through Keys of a Sorted
+     * Table Segment while including both fully indexed and tail updates.
+     *
+     * Note: this will return the same result as {@link ContainerSortedKeyIndex#getSortedKeyIndex}, however this method
+     * will wait on any Segment-specific recovery (and trigger it) to complete before executing, which should enable a
+     * safe iteration for recently recovered Table Segments.
+     *
+     * @param segment A {@link DirectSegmentAccess} representing the Segment for which to get the {@link SegmentSortedKeyIndex}.
+     * @return A CompletableFuture that, when completed, will contain the desired result. This Future will wait on any
```
fpj, line 559: > This Future will wait on any Segment-specific recovery to complete before executing.

This is a good information to have, but it made me wonder if that's not true of any segment-related operation. Is there anything special about this call to get the sorted key index with respect to recovery?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerSortedKeyIndex.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Notifies that the indexed offset for a particular Segment Id has been changed.
+     *
+     * @param segmentId   The Segment Id whose indexed offset has changed.
+     * @param indexOffset The new indexed offset. If -1, and if the given Segment is currently registered, it will be
+     *                    de-registered (since -1 indicates it has been evicted).
+     */
+    void notifyIndexOffsetChanged(long segmentId, long indexOffset) {
```
fpj, line 73: I'm not sure what this notification is about, could you elaborate?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/KeyTranslator.java
```diff
@@ -0,0 +1,207 @@
...
+import io.pravega.common.util.ArrayView;
+import io.pravega.common.util.ByteArraySegment;
+import io.pravega.segmentstore.contracts.tables.TableEntry;
+import io.pravega.segmentstore.contracts.tables.TableKey;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Translates Table Segment Keys from an external form into an internal one and back.
+ */
+abstract class KeyTranslator {
```
fpj, line 22: What's this `KeyTranslator` about?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/ContainerSortedKeyIndexTests.java
```diff
@@ -0,0 +1,392 @@
...
+import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import lombok.RequiredArgsConstructor;
+import lombok.val;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Unit tess for {@link ContainerSortedKeyIndex} and {@link SegmentSortedKeyIndexImpl}.
```
fpj, line 43: Typo.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
andreipaduroiu: > if the caller has the first item and wants it inclusive, then it can use the first item it already has.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
```diff
@@ -0,0 +1,414 @@
...
+        this.traceLogId = traceLogId == null ? "" : traceLogId;
+    }
+
+    //endregion
+
+    //region Updates
+
+    /**
+     * Atomically inserts the items in 'toInsert' into the {@link BTreeSet} and removes the items in 'toRemove'
+     * from the {@link BTreeSet}. No duplicates are allowed; the same item cannot exist multiple times in either 'toInsert'
```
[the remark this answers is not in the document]
andreipaduroiu: An `IllegalArgumentException` is thrown.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/tables/IteratorArgs.java
andreipaduroiu: We do not generate or publish Javadocs for internal services since everything is behind the wire protocol, so this is just for someone looking at the code. 

My IDE correctly displays Javadoc when I hover over a method, so I am not worried about it here.

And yes, it's only the `prefixFilter` that is experimental, since the whole Sorted Table Segment is experimental too.

> getpPrefixFilter returns an ArrayView that is not guaranteed to be correct
No. Making use of this feature is experimental. There is nothing here that indicates something isn't going to be correct.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/tables/TableStore.java
andreipaduroiu: Good point. Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerKeyIndex.java
andreipaduroiu: It's true of any segment-related operation that have this phrase attached to them. Not all operations need to wait on that. Conditional updates must wait, but unconditional updates do not. Retrievals and iterators must wait.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerSortedKeyIndex.java
andreipaduroiu: It notifies that "IndexedOffset" has changed for this segment. If you look at the method with same name in `SegmentSortedKeyIndex`, you'll see that this is used to clear the tail cache.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/KeyTranslator.java
andreipaduroiu: I need to separate user keys from the Sorted Key Index keys (since I store the Sorted Key Index alongside the rest of the keys). I do not want internal and external keys to clash, or for a user to (accidentally) modify a SKI key-value or viceversa.

For Sorted Table Segments only, this translator adds a 1-byte prefix to each key, as follows:
- User-keys: byte equivalent of `E` (for External)
- SKI keys: byte equivalent of `I` (for Internal).

For Hash Table Segments, the no-op translator is used which doesn't do anything (so no effect on existing keys). 

Currently this prepend is inefficient due to a mandatory array copy. However with #4837 this will be a simple, zero-copy composition which has no overhead.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSetPage.java
fpj: I'm concerned that at this point I don't understand when it is acceptable to not comply and version directly instead of using the versioned serializer. One of the main goals of versioned serializer was to make sure that we are versioning persistent and external data structures across Pravega in a consistent manner. 

It might be acceptable to do it due to the performance concern, but I'm wondering what the criteria is for exceptions for other PRs.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
fpj: > The caller may have the first item, but the first item may not be part of the set. If we always make it exclusive, the caller may erroneously include that first item in the result even if it is not part of the set.

It is correct that making it always exclusive requires a different code path to add the first or last when you want it inclusive, and that makes it more prone to errors. The developer can always set the flag incorrectly, though.

> This is a low-level API.

This is part of common, which is principle should be usable even outside Pravega. I'm curious, is this common in the implementation of iterators?  Do you know of other examples outside Pravega that implement iterators like this?

> Another reason would be to make it consistent

We should make it consistent, at least for this PR as we are not really changing `BTreeIndex`.

andreipaduroiu: Some examples I could quickly came up with:
- Java `IntStream.range(start, end)` and `IntStream.rangeClosed(start, end)`
- Java `NavigableSet`: [headSet](https://docs.oracle.com/javase/7/docs/api/java/util/NavigableSet.html#headSet(E,%20boolean) and [subSet](https://docs.oracle.com/javase/7/docs/api/java/util/NavigableSet.html#subSet(E,%20boolean,%20E,%20boolean))
    - Similarly for `NavigableMap`

NavigableSet and NavigableMap are what I based the APIs for BTreeSet and BTreeIndex on (more or less).

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSet.java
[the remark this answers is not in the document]
andreipaduroiu: Added a `@throws` tag.

common/src/main/java/io/pravega/common/util/btree/sets/BTreeSetPage.java
andreipaduroiu: (copy-pasting offline conversation conclusion)
The way to perform an upgrade with this format is as follows:
- We can create new serialization versions by adding new fields at the end of the current serialization.
- There isn't anything to do at this time in terms of accepting a larger input for BTreeSetPage. The format encodes the length of the BTreeSetPage and the parsing logic validates that the given buffer isn't smaller.
    - The page format is: `{Header}{ItemOffsets}{ItemData}{PageId}`. 
    - The length of an item is calculated by getting the difference between that item's offset and the next item's offset.
    - Then line 152 slices the contents buffer to the size encoded in the header.
- So in case of a backwards compatible upgrade (from Version 1), it would work like this:
    - New code defines Version 2, but does not use it yet (it just understands it)
    - New code will serialize using Version 1, but encode its additional data after the end of the buffer. It will make sure that the size encoded in the header ends where the old code expects it to.
    - Old code will read this size and slice the input, essentially dropping new data.
    - New code will be programmed to read past this (if anything) and process any additional fields.
    - Eventually, in a subsequent release, we will change the code to serialize to Version 2, thus completing the upgrade.

Not as smooth as VersionedSerializer (which was doing all this legwork for us), but this is the tradeoff I'm willing to accept for better perf.

PR #4765, opened 2020-05-01
Issue 4764: Optimized AppendDecoder to make fewer buffer copies

### LINKED ISSUE #4764: Optimize AppendDecoder
**Problem description**
AppendDecoder and associated WireCommands make too many buffer copies before generating the final append.

**Problem location**
AppendDecoder, WireCommands.

Reduce the number of copies. Ideally the generated Append should have a ByteBuf that references the Netty buffers directly without having to do intermediate copies.

andreipaduroiu:
**Change log description**  
- Optimized AppendDecoder to make fewer buffer copies from Netty to the Segment Store.
- Optimized the ingestion path in the Segment Store to choose the most efficient method of copying data based on the source and target buffer types.

**Purpose of the change**  
Fixes #4764.

**What the code does**  
`AppendDecoder`, in conjunction with `WireCommands.AppendBlock/AppendBlockEnd/ConditionalAppend/Event/PartialEvent`, is responsible for translating between Wire Commands and an `Append` that can be processed by the Segment Store it's sitting on top of.

However all these classes make numerous buffer allocations and copies for data to get from Netty's network bufs to the Segment Store:
- Netty -> AppendBlockEnd -> Append
- Netty -> AppendBlock/AppendBlockEnd -> Append
- Netty -> ConditionalAppend -> Event -> Append

This PR changes these classes to not execute those copies anymore but rely on ByteBuf slicing and retaining for this:
- The Wire Commands listed above use `EnhancedByteBufInputStream` to fetch a `ByteBuf` slice out of Netty's network bufs (direct memory buffers)
- These are `retain`-ed and sent out to `AppendDecoder`
- `AppendDecoder` stitches them into one or more `Append` instances which exposes those bufs as a composite `ByteBuf`. This `Append is then passed on to the underlying Segment Store via `AppendProcessor`.

AppendProcessor and Segment Store already manage the lifecycle of these buffers by invoking `retain`/`release` in `StreamSegmentAppendOperation` and `release` in `AppendProcessor` after the append is complete. 

Additionally, inside the ingestion pipeline in the Segment Store, the following modifications have been made:
- `DataFrameOutputStream` implements the new `BufferViewSink`. This allows using direct buffer copies (from Netty via `StreamSegmentAppendOperation` to the `DataFrame` that ends up getting sent to BK). Previously we were using the `OutputStream.write(...)` methods which were less efficient.

**How to verify it**  
Unit tests, integration tests and system tests must pass.
Unit tests for these classes have been updated to check for memory leaks and under-referenced buffers.

### DIFF
--- a/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
+++ b/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
@@ -203,7 +203,6 @@ public class AppendDecoder extends MessageToMessageDecoder<WireCommand> {
                 // Work around a bug in netty:
                 // See https://github.com/netty/netty/issues/5597
                 if (appendDataBuf.readableBytes() == 0) {
-                    assert currentBlock.getData().readableBytes() == 0;
                     currentBlock.release();
                     appendDataBuf = wrappedBuffer(((WireCommands.PartialEvent) cmd).getData(), blockEnd.getData());
                 } else {
--- a/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
+++ b/shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
@@ -33,6 +33,9 @@ import static io.netty.buffer.Unpooled.wrappedBuffer;
  * which must be an AppendBlockEnd.
  * The AppendBlockEnd command should have all of the information need to construct a single
  * Append object with all of the Events in the block.
+ * <p>
+ * The {@link #decode} method invokes {@link #processCommand(WireCommand)} to generate {@link Request} instances. Please
+ * refer to {@link #processCommand(WireCommand)} Javadoc for important information about returned objects.
  *
  * @see CommandEncoder For details about handling of PartialEvents
  */
@@ -62,8 +65,24 @@ public class AppendDecoder extends MessageToMessageDecoder<WireCommand> {
         }
     }
 
+    /**
+     * Processes inbound {@link WireCommand} instances and converts them to {@link Request} instances.
+     *
+     * @param command Inbound {@link WireCommand}. Supported types are: {@link WireCommandType#PADDING},
+     *                {@link WireCommandType#SETUP_APPEND}, {@link WireCommandType#CONDITIONAL_APPEND},
+     *                {@link WireCommandType#APPEND_BLOCK} and {@link WireCommandType#APPEND_BLOCK_END}.
+     * @return One of the following:
+     * - null if command type is {@link WireCommandType#PADDING} or {@link WireCommandType#APPEND_BLOCK}.
+     * - command if command type is A {@link WireCommandType#SETUP_APPEND}.
+     * - An {@link Append} instance for any other types. {@link Append#getData()} will be a {@link ByteBuf} that wraps
+     * one or more {@link ByteBuf}s that are part of the supplied command instance(s) and will have a {@link ByteBuf#refCnt()}
+     * at least equal to 1. It is very important that {@link Append#getData()}{@link ByteBuf#release()} is invoked on this
+     * instance when it is no longer needed so that any direct memory used by it can be properly freed.
+     * @throws InvalidMessageException If commands are received out of order or otherwise form an illegal Append.
+     * @throws IOException             If unable to read from the command's data buffer.
+     */
     @VisibleForTesting
-    public Request processCommand(WireCommand command) throws Exception {
+    Request processCommand(WireCommand command) throws InvalidMessageException, IOException {
         try {
             return processCommandInternal(command);
         } catch (Throwable ex) {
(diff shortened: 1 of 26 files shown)


andreipaduroiu:
System tests have passed successfully.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
```diff
@@ -163,22 +205,15 @@ private ByteBuf getAppendDataBuf(WireCommands.AppendBlockEnd blockEnd, int sizeO
...
-                    appendDataBuf.release();
+                    currentBlock.release();
                     appendDataBuf = wrappedBuffer(((WireCommands.PartialEvent) cmd).getData(), blockEnd.getData());
                 } else {
                     appendDataBuf = wrappedBuffer(appendDataBuf, ((WireCommands.PartialEvent) cmd).getData(), blockEnd.getData());
                 }
             }
         }
 
-        // Make a copy of the ByteBuf as the readable bytes of the result may be significantly less than the total allocated
```
tkaitchuck, line 174: See this comment for why this copy was added. We need to test to confirm this will not cause problemms.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -582,26 +591,32 @@ public void writeFields(DataOutput out) throws IOException {
...
             long lastEventNumber = in.readLong();
             long requestId = in.available() >= Long.BYTES ? in.readLong() : -1L;
-            return new AppendBlockEnd(writerId, sizeOfHeaderlessAppends, wrappedBuffer(data), numEvents, lastEventNumber, requestId);
+            return new AppendBlockEnd(writerId, sizeOfHeaderlessAppends, data.retain(), numEvents, lastEventNumber, requestId)
+                    .requireRelease();
+        }
+
+        @Override
+        void releaseInternal() {
+            this.data.release();
```
tkaitchuck, line 613: I don't see a corresponding `retain`

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/AppendEncodeDecodeTest.java
```diff
@@ -219,23 +219,37 @@ public void execute(Runnable runnable) {
...
-
     @Mock
     private ChannelHandlerContext ctx;
     @Mock
     private Channel ch;
+    private ByteBuf fakeNetwork;
+    private BufferReleaseVerifier releaseVerifier;
+
+    protected int getThreadPoolSize() {
+        return 1;
```
tkaitchuck, line 230: Does this number have to be 1 for the test to pass? If so there should be a comment explaining it.

andreipaduroiu: If it's 0, it uses the InlineExecutor. If it's 1, it uses an executor with 1 thread. This is exactly as it was before, if you look at the modifications in this class, except that now I am using the services from the inherited class as opposed from doing it again.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
andreipaduroiu: 6 lines above.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
andreipaduroiu: @RaulGracia is testing this with unthrottled loads. If we don't get any memory problems we should be OK. 

Since we've added that extra copy we have made numerous improvements in the Segment Stores' throttler and it is now a lot more aggressive in keeping incoming data flow under control. My expectation is that this should no longer be an issue and I want to see if it actually serves any purpose.

In addition, there is no longer a heap issue, as none of the source buffers are stored in the heap (the changes I made in WireCommands eliminated the heap buffer allocations). This will now point to sliced direct mem buffers from Netty.

RaulGracia:
@andreipaduroiu please, could address one checkstyle issue that is preventing the build from passing?
```> Task :test:integration:checkstyleMain FAILED
[ant:checkstyle] [ERROR] /home/travis/build/pravega/pravega/test/integration/src/main/java/io/pravega/test/integration/selftest/adapters/TruncateableArray.java:154: Overload methods should not be split. Previous overloaded method located at line '144'. [OverloadMethodsDeclarationOrder]
```
Apart from that, I can confirm that this change provides significant latency improvements, specially for readers in moderate/high throughput scenarios. Also, I system tests are passing.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/AppendDecoder.java
andreipaduroiu: We executed several perf tests and had no issues.

PR #4768, opened 2020-05-04
Issue 4774: Optimize reader computations

### LINKED ISSUE #4774: Optimize reader computations
**Problem description**
In our analysis, we observe that there are a couple of methods that are being called extensively while reading events: `ReaderGroupStateManager.doesReaderOwnTooManySegments` and `EventStreamReaderImpl.getPosition`. These methods seem to be executed on a per-event basis and the compute cycles related to their execution make readers to achieve lower throughput, which is visible for small events.

**Problem location**
Reader

Minimize the calls to these methods to save up cpu cycles, while keeping the behavior of the client intact.

RaulGracia:
**Change log description**  
This PR saves up compute cycles for a reader by minimizing and optimizing calls to several costly functions related to manage its internal state.

**Purpose of the change**  
Fixes #4774.

**What the code does**  
This PR attempts to minimize the number of calls to the following methods:
- `doesReaderOwnTooManySegments()`: This method calculates whether a reader is owning too many segments, so it can release them and let other readers to acquire them. However, this method involves a non-negligible computation and is being evaluated on a per-event read basis (`readNextEventInternal()` -> `updateGroupStateIfNeeded()` -> `acquireNewSegmentsIfNeeded()` -> `shouldAcquireSegment()`). This PR drastically reduces the number of calls to this method by executing it only when the appropriate timers are consumed.

- We now also reuse the same segment `ranges` map in `EventStreamReaderImpl` for multiple `PositionImpl` objects, until there is a change in segments. When this happens, we just create a new `ranges` map (i.e., this map has a "copy-on-write" behavior).

- `getCheckpointForReader()`: In method `shouldAcquireSegment()`, readers also execute `getCheckpointForReader()`. The number of calls to this method has been reduced by executing it only when the appropriate timers are consumed. Besides, the actual implementation of this method has been optimized (e.g., keeping a sorted list of checkpoints for reader to directly get the min checkpoint).

- `getPosition()`: The other important performance bottleneck is the execution of `getPosition()` per event read. Before, this method was called twice per event read (in `readNextEventInternal()` as part of the return object of the function and in `updateGroupStateIfNeeded()`). This PR reduces drastically the number of calls to this method as follows:
  - `EventStreamReaderImpl` has now two new fields called `ownedSegments` and `segmentOffsetUpdates` that will be used to build `PositionImpl` objects.
  - We cannot avoid "refreshing" `ownedSegments` and `segmentOffsetUpdates` every time that there is a change in the segments owned for a reader (e.g., truncation, checkpoint, end of segment,...). This has a similar cost to the former `getPosition()` method, but it happens rarely compared to the number of event read.
  - Whenever there is no change in the segments owned for a reader (general case), is when the optimization comes into play. In this case, `EventStreamReaderImpl` keeps `ownedSegments` in the same state and for every new event read it appends a new element in `segmentOffsetUpdates`, indicating the new offset for the event read on a segment. As you may infer, this has a constant cost (i.e., one list append) compared to the previous call to `getPosition()` which required to generate the new `ownedSegments` map per event read (the cost of creating this map is in turn related to the number of owned segments). While there is no change in the segments, `EventStreamReaderImpl` can buffer up to `MAX_BUFFERED_SEGMENT_OFFSET_UPDATES` of segment updates.
  - The `PositionImpl` class contains two new fields: i) one to refer to `segmentOffsetUpdates` list and ii) another (called `version`) to determine the updates that should be applied for this particular `PositionImpl` object. That is, while N `PositionImpl` objects with refer to the same list of `segmentOffsetUpdates`, they will build their internal state by applying a subset of these updates to represent the actual point at which the event at hand was read. Moreover, the application of such updates of offsets to the internal state of `PositionImpl` will be executed lazily, only if this `PositionImpl` is used.

In terms of performance, this PR increases between 3x (1 segment) to 4x (16 segments) the throughput of a reader for small events (e.g., 100B).

**How to verify it**  
Added new tests to verify the behavior of `PositionImpl` and `CopyOnWriteMapUtils` classes. Rest of tests should be passing as before. System tests are passing (2/2).


client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -154,8 +156,8 @@
         } 
         lastRead = Sequence.create(segment.getSegmentId(), offset);
         int length = buffer.remaining() + WireCommands.TYPE_PLUS_LENGTH_SIZE;
-        return new EventReadImpl<>(deserializer.deserialize(buffer), getPosition(),
-                                   new EventPointerImpl(segment, offset, length), null);
+        lastPosition = getPosition();
```
tkaitchuck, line 159: Using a member variable is error prone. (In fact there is a case this logic does not cover: if the segments change as a result of calling updateGroupStateIfNeeded, the next time through the above while loop it will pass in stale data resulting in an invalid state.)

Instead it may be better to just use a local variables by computing the position on 128 an passing it into updateGroupStateIfNeeded, then modifying the offset for the segment read from before returning it.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -339,15 +348,15 @@ private boolean shouldAcquireSegment() throws ReaderNotInReaderGroupException {
...
-            if (state.getCheckpointForReader(readerId) != null) {
-                return false;
-            }
             if (state.getNumberOfUnassignedSegments() == 0) {
                 if (doesReaderOwnTooManySegments(state)) {
                     acquireTimer.reset(calculateAcquireTime(readerId, state));
                 }
                 return false;
             }
+            if (state.getCheckpointForReader(readerId) != null) {
```
tkaitchuck, line 357: I optimized this method, I don't think this reordering is needed anymore.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -213,13 +216,19 @@ Segment findSegmentToReleaseIfRequired() {
      * the reader with the least assigned to it.
      */
     private boolean doesReaderOwnTooManySegments(ReaderGroupState state) {
-        Map<String, Double> sizesOfAssignemnts = state.getRelativeSizes();
+        // The previous calculation for this method is cached during TIME_UNIT to prevent excessive per-event computations.
```
tkaitchuck, line 219: I am confused by this, because this method should only be invoked once per checkpoint interval. If that's set to say 20sec then this new time logic is totally redundant. Am I missing something?

RaulGracia: You are right, this method is being invoked much more often than it should, as I understand its goal. This method is invoked in two different places, and the one that takes a lot of CPU cycles comes from `shouldAcquireSegment()`, which as you can see, it can be potentially invoked on every read event only if `acquireTimer.hasRemaining()` is `false` most of the time, which seems to be what is going on according to the flamegraphs. The `acquireTimer` should be generally in the order of seconds, right? So maybe I can spend some more time to see if this timer is incorrectly set at some point.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
RaulGracia: It does if `acquireTimer.hasRemaining()` is `false` quite often, which seems to be the case. By doing this, we are preventing the execution of `getCheckpointForReader()` thanks to the modification in the implementation of `doesReaderOwnTooManySegments`. If it finally turns out that there is a problem in the way `acquireTimer` is being set, then we can fix it and revert this change.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
RaulGracia: Maybe I'm mistaken, but your suggestion implies that we should be able to "update" the offsets of a `PositionImpl` object in some way, so we don't have to instantiate it twice anyway, right? I don't know if this is possible at the moment, as this object seems to be read-only and uses unmodifiable collections in its getters. 

But your comment made me think on a more aggressive optimization: let's assume that we still keep a member variable for the "last position" object. As you mention, we fully recompute the last position object whenever there is a change in the segments, so the state is valid upon segment changes. If there are no segment changes, which is the majority of cases, we can modify the offset of the segment read on a particular iteration (which would imply to be able to be able to safely update the Position object) reusing the "last position" object. If we do that, we would be able to avoid 99% of the `getPosition()` execution calls, which is the main performance bottleneck we have now in the reader side. Will try to propose this change so we can discuss about it.

RaulGracia:
@tkaitchuck looking at when the `acquireTimer` is being updated, I have added some debug logs and found the following when starting a Pravega Benchmark using this branch on the client:
```
(started reading)
...
initializeReader 1 - SETTING ACQUIRE TIMER TO: 0
shouldAcquireSegment 2 - SETTING ACQUIRE TIMER TO: 30
acquireSegment 1 - SETTING ACQUIRE TIMER TO: 1
...
(continue reading for the rest of the experiment)
```
That is, just after the start of the experiment, the reader sets `acquireSegment` to 1 second, and that timer is not updated anymore. This means that for most of the experiment, the condition on `acquireTimer.hasRemaining()` inside `shouldAcquireSegment` is `false` and the next condition that is evaluated on that method is always `true` (`state.getNumberOfUnassignedSegments() == 0`). This leads to the continuous execution of `doesReaderOwnTooManySegments()`. 

The bottom line is that `acquireTimer` seems not to be updated as often as it should, which has an impact in terms of performance. We have two options: if the behavior in which the client resets `acquireTimer` is the expected, we can keep the current timer `periodicSegmentOffloadTimer` to prevent `doesReaderOwnTooManySegments()` from being executed too often. On the other hand, if this is not the expected behavior, then we can just discuss where else `acquireTimer` should be reset.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -97,12 +97,12 @@ void beginNewCheckpoint(String checkpointId, Set<String> currentReaders, Map<Seg
...
     String getCheckpointForReader(String readerName) {
-        OptionalInt min = getCheckpointsForReader(readerName).stream().mapToInt(checkpoints::indexOf).min();
-        if (min.isPresent()) {
-            return checkpoints.get(min.getAsInt());
-        } else {
+        List<String> checkpointsForReader = getCheckpointsForReader(readerName);
+        if (checkpointsForReader.isEmpty()) {
             return null;
         }
+        OptionalInt min = checkpointsForReader.stream().mapToInt(checkpoints::indexOf).min();
```
tkaitchuck, line 104: Replace this with a standard for loop. It will be faster.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -73,16 +74,19 @@
     private boolean closed;
     @GuardedBy("readers")
     private final List<EventSegmentReader> readers = new ArrayList<>();
+    // Ranges map is heavily used to build Position objects that are returned to the client. While there is no change
+    // in segment distribution, we reuse the same map instance for performance reasons. But this map should have a
+    // copy-on-write behavior to do not impact all the Position objects referencing the previous version of it.
     @GuardedBy("readers")
-    private final Map<Segment, Range> ranges = new HashMap<>();
+    private final AtomicReference<Map<Segment, Range>> ranges = new AtomicReference<>(new HashMap<>());
```
tkaitchuck, line 81: This does not need to be an atomic reference, it is guarded by the readers lock, so it can be re-assigned.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -172,11 +179,12 @@ private void blockFor(long timeoutMs) {
     }
 
     private PositionInternal getPosition() {
-        Map<Segment, Long> ownedSegments = new HashMap<>(sealedSegments);
+        Map<Segment, Long> ownedSegments = new HashMap<>(sealedSegments.size() + readers.size());
```
tkaitchuck, line 182: Let's make ownedSegments a List<Entry<Segment, Long>> and adapt positionImpl to store this. This will be much cheaper to construct.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
tkaitchuck: I think we can avoid this. See my comments below.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -196,8 +204,7 @@ private PositionInternal getPosition() {
      * have been persisted.
      */
     @GuardedBy("readers")
-    private String updateGroupStateIfNeeded() throws ReaderNotInReaderGroupException {
-        PositionInternal position = (lastPosition == null) ? getPosition() : lastPosition;
+    private String updateGroupStateIfNeeded(PositionInternal position) throws ReaderNotInReaderGroupException {
```
tkaitchuck, line 207: Rather than passing this in, we can probably avoid creating it all together. `position` is used in four places:
1. on line 209 where the reader is checkpointing. (This is rare and we can call getPosition() inside of this if without concern)
2. on line 218. This is the same situation as the above. It only occurs on checkpoint.
3. when calling `acquireSegmentsIfNeeded`. However this method doesn't actually use the `position` passed in unless a timer has passed. So perhaps having this method call `shouldAquireSegment` first would avoid the need for position most of the time.
4. `updateLagIfNeeded` here the situation is the same as the above, it's based on a timer. So we could easily have a block that does:
```
if (shouldAquireSegment() || shouldUpdateLag()) {
    PositionImpl position = getPosition();
    // The code currently in lines 231-235 here.
}
```

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -377,14 +386,36 @@ public Type fetchEvent(EventPointer pointer) throws NoSuchEventException {
...
+
+    private void copyOnRemoveSegmentRange(Segment segment) {
+        Map<Segment, Range> newRanges = copyRanges();
+        newRanges.remove(segment);
+        ranges.set(newRanges);
+    }
+
+    private Map<Segment, Range> copyRanges() {
+        Map<Segment, Range> newRanges = new HashMap<>(2 * ranges.get().size());
+        newRanges.putAll(ranges.get());
```
tkaitchuck, line 413: HashMap already provides a constructor that does these two lines.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -405,7 +403,9 @@ private int calculateNumSegmentsToAcquire(ReaderGroupState state) {
 
     @VisibleForTesting
     static Duration calculateAcquireTime(String readerId, ReaderGroupState state) {
-        return TIME_UNIT.multipliedBy(state.getNumberOfReaders() - state.getRanking(readerId));
+        int multiplier = state.getNumberOfReaders() - state.getRanking(readerId);
+        Preconditions.checkArgument(multiplier >= TIME_UNIT.getSeconds(), "Invalid acquire timer multiplier");
```
tkaitchuck, line 407: Shouldn't that be `Preconditions.checkArgument(multiplier >= 1, ...`

RaulGracia:
@tkaitchuck I'm addressing the comments you made, but meanwhile, I have another question: We can see in the flamegraphs that almost 30% of the execution samples within the `getPosition()` method are due to `EventSegmentReaderImpl:getOffset()`, which is in turn calling to `SegmentInputStreamImpl:getOffset()`. I observe that both methods are `@Synchronized`, so we are inducing a double lock layer here (in both cases, the same lock is shared across multiple methods). Is it necessary to synchronize both getters in that way? cc/@shrids

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
RaulGracia: Fixed.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
RaulGracia: Done.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
RaulGracia: I did that change but didn't show much improvement. Instead, I opted for another kind of optimization in which `ownedSegments` is re-used and we batch the changes of offsets for every read. A set of Position objects that reference the same `ownedSegments` object, will also have a reference to a list of ` List<Entry<Segment, Long>> segmentOffsetUpdates`. Each Position object will then lazily apply the updates to segment offsets up to the point the event was read. This has proved to make readers much faster reading small events, specially when multiple segments are involved.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
RaulGracia: Thanks, I have applied this change (with some additions to that condition).

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -377,14 +386,36 @@ public Type fetchEvent(EventPointer pointer) throws NoSuchEventException {
             return ImmutableList.copyOf(readers);
         }
     }
-    
+
+    // Ranges management region
```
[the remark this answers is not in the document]
RaulGracia: Created a `CopyOnWriteMapUtils` class to contain all this functionality.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
RaulGracia: In this case is the same, but I changed it to what you suggest.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -343,9 +343,7 @@ private boolean shouldAcquireSegment() throws ReaderNotInReaderGroupException {
                 return false;
             }
             if (state.getNumberOfUnassignedSegments() == 0) {
```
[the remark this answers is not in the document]
RaulGracia: Done.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -240,7 +277,7 @@ private void releaseSegmentsIfNeeded(PositionInternal position) throws ReaderNot
             if (reader != null) {
                 if (groupState.releaseSegment(segment, reader.getOffset(), getLag(), position)) {
                     readers.remove(reader);
-                    ranges.remove(reader.getSegmentId());
+                    ranges = CopyOnWriteMapUtils.remove(ranges, reader.getSegmentId());
```
shrids, line 280: Given that `ranges` is guarded by readers   `@GuardedBy("readers")` why do we need to create a copy? isn't it redundant ?

RaulGracia: The reason is that the `ranges` map is referenced by many `PositionImpl` objects (i.e, every time that `readNextEvent()` returns an event and there is no change in segments, the same `ranges` map is reused). This means that we cannot modify this map if there is a change in segments, as it would impact to all the previous `PositionImpl` objects and they would be in an inconsistent state. So, every time there is a change in the segments, we need the variable `ranges` to point to a new map object and apply the change in segments to the new map.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -114,14 +109,14 @@ private void recomputeCheckpointIndex() {
         for (Entry<String, List<String>> entry : uncheckpointedHosts.entrySet()) {
             String checkpointId = entry.getKey();
             for (String host : entry.getValue()) {
-                List<String> checkpointsForHost = checkpointIndex.get(host);
-                if (checkpointsForHost == null) {
-                    checkpointsForHost = new ArrayList<>();
-                    checkpointIndex.put(host, checkpointsForHost);
-                }
+                List<String> checkpointsForHost = checkpointIndex.computeIfAbsent(host, k -> new ArrayList<>());
```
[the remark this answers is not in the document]
RaulGracia: The code that this line is replacing was basically putting a new ArrayList object as an initialization value for a non-existing key. Just wanted to express the same logic in a more concise way.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -154,8 +168,18 @@
         } 
         lastRead = Sequence.create(segment.getSegmentId(), offset);
         int length = buffer.remaining() + WireCommands.TYPE_PLUS_LENGTH_SIZE;
-        return new EventReadImpl<>(deserializer.deserialize(buffer), getPosition(),
-                                   new EventPointerImpl(segment, offset, length), null);
+        addSegmentOffsetUpdateIfNeeded(segment, offset + length);
+        return new EventReadImpl<>(deserializer.deserialize(buffer),
+                PositionImpl.builder().ownedSegments(ownedSegments).segmentRanges(ranges).updatesToSegmentOffsets(segmentOffsetUpdates).build(),
```
shrids, line 173: the maps need to be wrapped with `java.util.Collections#unmodifiableMap` and the updates should be wrapped with `Collections.unmodifiableList()`.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -66,26 +75,29 @@ public PositionImpl(Map<SegmentWithRange, Long> segments) {
...
     public Map<Segment, Long> getOwnedSegmentsWithOffsets() {
+        applySegmentOffsetUpdatesIfNeeded();
         return Collections.unmodifiableMap(ownedSegments);
     }
     
     @Override
     Map<SegmentWithRange, Long> getOwnedSegmentRangesWithOffsets() {
+        applySegmentOffsetUpdatesIfNeeded();
         HashMap<SegmentWithRange, Long> result = new HashMap<>();
         for (Entry<Segment, Long> entry : ownedSegments.entrySet()) {
```
shrids, line 92: ownedSegments can be modified by the different thread while iterating here.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
RaulGracia: Thanks, I have done this change.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
RaulGracia: Correct, good catch. Now, when `applySegmentOffsetUpdatesIfNeeded()` is executed, the `PositionImpl` object copies the `ownedSegments` map and only then sets the right offsets representing the point at which the event was read.

RaulGracia:
@shrids @tkaitchuck let me know if you have any additional comment on this PR.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -196,4 +230,20 @@ public static Position fromBytes(ByteBuffer buff) {
...
+        // No updates, so nothing to do.
+        if (this.updatesToSegmentOffsets == null) {
+            return;
+        }
+        // We are going to modify ownedSegments to match the segment offsets at the time this event was read, so copying
+        // the map of ownedSegments to do not impact to the map referenced by other PositionImpl objects.
+        this.ownedSegments.set(new HashMap<>(ownedSegments.get()));
+        // Apply all the Segment offset updates up to the point in which this event was read.
+        for (int i = 0; i < this.version; i++) {
+            this.ownedSegments.get().put(this.updatesToSegmentOffsets.get(i).getKey(), this.updatesToSegmentOffsets.get(i).getValue());
```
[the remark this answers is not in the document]
RaulGracia: Thanks for the comment @shrids. I assume that your concern is about the possibility of getting a `ConcurrentModificationException` while looping over a list that could be subject to concurrent appends (totally valid concern). To test whether this is thread safe or not, I have tested the following:
1) Thread is continuously adding elements on a list.
2) Another thread is continuously iterating the same list.

I have tried 2 variants for thread 2:
- Looping over the entire list:
```
for (Integer element: elements) { //do X on element }
```
- Emulating what we do in this PR (take a point in the list before the end and iterate up to that point):
```
for (int j = 0; j < elements.size() / 2; j++) {//do X on element}
```
For case 1, I consistently get `ConcurrentModificationException` on every execution. However, after many executions of this experiments, I have not seen any occurrence of this problem for case 2. I accept that this may not be a usual pattern, but is the most efficient approach I could come up so far and it seems to be thread safe.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
RaulGracia: @shrids I just moved the immutability of these collections to the PositionImpl builder itself, so it takes effect to all the usages of such builder.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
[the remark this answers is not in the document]
shrids: With the current implementation of ArrayList i do not see the possibility of an errror, given that we guarantee only one thread is performing the add. The `resize` impl of the ArrayList would not cause an error here too. It is just that the pattern is unconventional and I have always avoided it.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -22,23 +22,28 @@
...
-@EqualsAndHashCode(callSuper = false)
 public class PositionImpl extends PositionInternal {
 
     private static final PositionSerializer SERIALIZER = new PositionSerializer();
-    private final Map<Segment, Long> ownedSegments;
+    private Map<Segment, Long> ownedSegments;
     private final Map<Segment, Range> segmentRanges;
 
+    // If this field is set, it means that we will need to apply the updates on the ownedSegments.
+    private transient List<Entry<Segment, Long>> updatesToSegmentOffsets;
```
tkaitchuck, line 43: Given the constants you selected this is going to add ~30MB of memory overhead per reader in a user's application. I don't consider this tradeoff worthwhile. See above for how we can avoid this.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -55,37 +62,42 @@ public PositionImpl(Map<SegmentWithRange, Long> segments) {
     }
     
     @Builder(builderClassName = "PositionBuilder")
-    PositionImpl(Map<Segment, Long> ownedSegments, Map<Segment, Range> segmentRanges) {
-        this.ownedSegments = ownedSegments;
+    PositionImpl(Map<Segment, Long> ownedSegments, Map<Segment, Range> segmentRanges, List<Entry<Segment, Long>> updatesToSegmentOffsets) {
+        this.ownedSegments = Collections.unmodifiableMap(ownedSegments);
+        this.updatesToSegmentOffsets = Collections.unmodifiableList((updatesToSegmentOffsets == null) ? Collections.emptyList() : updatesToSegmentOffsets);
```
tkaitchuck, line 67: This computation is wrong. The updatesToSegmentOffsets list is passed in without copying it. This means that if a caller holds onto one position object, and then calls readNext the first position object will be updated and have the wrong value. Please add a test for this.

common/src/main/java/io/pravega/common/util/CopyOnWriteMapUtils.java
```diff
@@ -0,0 +1,54 @@
...
+package io.pravega.common.util;
+
+import javax.annotation.concurrent.NotThreadSafe;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utils class that contains copy-on-write methods for maps.
+ */
+@NotThreadSafe
```
tkaitchuck, line 19: How?! This is all static methods?

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -55,37 +62,42 @@ public PositionImpl(Map<SegmentWithRange, Long> segments) {
     }
     
     @Builder(builderClassName = "PositionBuilder")
-    PositionImpl(Map<Segment, Long> ownedSegments, Map<Segment, Range> segmentRanges) {
-        this.ownedSegments = ownedSegments;
+    PositionImpl(Map<Segment, Long> ownedSegments, Map<Segment, Range> segmentRanges, List<Entry<Segment, Long>> updatesToSegmentOffsets) {
```
tkaitchuck, line 65: The semantics of which of these collections are copied or not should be spelled out in a javadoc. It would avoid the bug below.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -60,6 +62,10 @@
 
     // Base waiting time for a reader on an idle segment waiting for new data to be read.
     private static final long BASE_READER_WAITING_TIME_MS = ReaderGroupStateManager.TIME_UNIT.toMillis();
+    // As an optimization to do not generate Position objects on every event read, we define a base map of segments and
+    // then a batch of updates to the offsets of these segments, one per event read. Internally, the Position object can
+    // derive the right offsets at which the event was read by lazily replying such updates up to the point it was read.
+    private static final long MAX_BUFFERED_SEGMENT_OFFSET_UPDATES = 1000;
```
tkaitchuck, line 68: Because a copy is needed anyway (see comments below) this is actually going to make things slower. It would be better to simply in the `getPosition()` method iterate over the `readers` and construct the `List<Entry<Segment, Long>>` with only one entry per-segment. This is faster, avoids both of the need collections here, and the need to try to cache the position object.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
RaulGracia: I have added an integration test to verify that the concurrent access and appends to `updatesToSegmentOffsets`. It shows that the optimization is safe and the offsets are correctly calculated.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
RaulGracia: You are right in the case that the copy of the list is necessary, but I think that the current approach of referencing it may be feasible and is the one that provides the actual read throughput speedup.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -114,14 +109,14 @@ private void recomputeCheckpointIndex() {
...
                 checkpointsForHost.add(checkpointId);
             }
         }
+        // Sort checkpoints per reader so we have fast access based on time ordering.
+        for (List<String> checkpointsForHost: checkpointIndex.values()) {
+            checkpointsForHost.sort(String::compareTo);
```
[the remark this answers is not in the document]
RaulGracia: Done.

common/src/main/java/io/pravega/common/util/CopyOnWriteMapUtils.java
RaulGracia: Removed.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
RaulGracia: Yes, I definitely can add javadoc explaining the main purpose of this builder and the way it is currently used. But I prefer to defer this task until we agree the final way to do this (whether passing `updatesToSegmentOffsets` is valid or not).

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -196,4 +229,19 @@ public static Position fromBytes(ByteBuffer buff) {
         return SERIALIZER.deserialize(new ByteArraySegment(buff));
     }
 
+    private synchronized void applySegmentOffsetUpdatesIfNeeded() {
```
tkaitchuck, line 232: This lock is not guarding `updatesToSegmentOffsets` because it is shared by many objects.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
```diff
@@ -196,4 +229,19 @@ public static Position fromBytes(ByteBuffer buff) {
...
+        // No updates, so nothing to do.
+        if (this.updatesToSegmentOffsets == null || this.updatesToSegmentOffsets.isEmpty()) {
+            return;
+        }
+        // We are going to modify ownedSegments to match the segment offsets at the time this event was read, so copying
+        // the map of ownedSegments to do not impact to the map referenced by other PositionImpl objects.
+        this.ownedSegments = new HashMap<>(ownedSegments);
+        // Apply all the Segment offset updates up to the point in which this event was read.
+        for (int i = 0; i < this.version; i++) {
+            this.ownedSegments.put(this.updatesToSegmentOffsets.get(i).getKey(), this.updatesToSegmentOffsets.get(i).getValue());
```
tkaitchuck, line 242: @shrids 's intuition was correct. This call to `updatesToSegmentOffsets.get(i)` is not thread safe. The list may not be synchronized in any way between the thread calling readNextEvent and the thread calling this method. As such this could endup reading garbage data.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
tkaitchuck: No. That map will be ~30mb and there is one per EventStreamReader.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
tkaitchuck: Perhaps I am missing something, but I don't see any actual concurrency in the test.

RaulGracia:
I have done a micro benchmark of the CPU cost of different approaches to manage the internal collections in `EventStreamReaderImpl` and how these are passed to individual `PositionImpl` objects:
1) Copy `ownedSegments` map for each `PositionImpl` object: This is the existing behavior. Therefore, each event read involves a _copy of a map_ with a cost `O(num_segments)`.
2) Make `ownedSegments` a list-like data structure: Copying a list is assumed to be faster than copying a map, so we assume that we could turn `ownedSegments` into a list-like data structure with indexing capabilities. This means that each `PositionImpl` would have copy of that list. Alternatively, we can also assume a `updatesToSegmentOffsets` with the size of the number of segments and copy it with the right offsets to each `PositionImpl` object. In any case, each event read involves a _copy of a list_ with a cost `O(num_segments)`.
3) Shared `updatesToSegmentOffsets` with a given batch size: This is the approach used in this PR. In this case, the reader accumulates updates to segment offsets in `updatesToSegmentOffsets` up to a given size (e.g., `MAX_BUFFERED_SEGMENT_OFFSET_UPDATES`). Up to `MAX_BUFFERED_SEGMENT_OFFSET_UPDATES` `PositionImpl` objects will point to that list and, only if needed, read from it to lazily build their internal offsets to the right values. This has a constant cost in the general case (i.e., append to a list).

Comparing these approaches in a micro benchmark, we get the following:
```
Iterations: 10M
1 segment:
Mean time method 1: 933.4
Mean time method 2: 295.1
Mean time method 3: 319.1

32 segments:
Mean time method 1: 1393.1
Mean time method 2: 610.6
Mean time method 3: 375.2
```
As expected, for the 1 segment case, the current approach is significantly worse, while optimizations 2 and 3 are virtually the same. This supports the hidden costs of copying maps we currently have in our reader. However, as we consider more segments, we observe that the cost of copying larger arrays significantly impacts on method 2 as well, whereas method 3 remains with similar performance. This is also expected, as the computation cost of method 3 in the general case of event reads is not related to the number of segments.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
RaulGracia: You are right. In the case of using an `ArrayList` in that way, an internal resize can be triggered by one thread adding new elements, which might lead another thread to read garbage when executing this method in `PositionImpl` that reads that list. A solution to this is to use an "structurally immutable" list, wrapped by a `List` for convenience. This is currently achieved by using `ArrayList.asList()` for initializing `updatesToSegmentOffsets`. Any attempt to `add` or `delete` elements from the list will result in `UnsupportedOperationException`, while `set` and `get` methods are allowed (similar to an array).

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
RaulGracia: I have added a performance comparison comparing your suggestion with this PR. For 1 segment, your approach is at par with this PR in terms of speed. But as more segments are handled by the reader, the performance decreases, given that the list to copy is larger.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
RaulGracia: I have added javadoc to clarify the goal of the builder method.

client/src/main/java/io/pravega/client/stream/impl/PositionImpl.java
RaulGracia: Now the new test (`integration/ReadTest.java`) involves different threads for writing events, reading events and checking positions objects, so all this activity should happen concurrently. If there is any issue in `PositionImpl` when reading the `segmentOffsetUpdates` list is also being used by `EventStreamReaderImpl`, the test should fail.

RaulGracia:
With current version of the PR I got 2/2 system tests passing.

tkaitchuck:
> Comparing these approaches in a micro benchmark, we get the following:
> 
> ```
> Iterations: 10M
> 1 segment:
> Mean time method 1: 933.4
> Mean time method 2: 295.1
> Mean time method 3: 319.1
> 
> 32 segments:
> Mean time method 1: 1393.1
> Mean time method 2: 610.6
> Mean time method 3: 375.2
> ```

Do these number include the cost of serializing the position object? I would expect that in this case 3 would have some additional overhead due to the need to compute the actual offset for each segment.

RaulGracia:
@tkaitchuck no, none of the microbenchmarks posted involve serialization costs, just the costs of internal management of collections. For option 3, serializing (or invoking any other method on `PositionImpl`), involves executing only once `applySegmentOffsetUpdatesIfNeeded()`, which you are right that requires to recompute the actual offsets of segments. But, if an application is going to serialize `PositionImpl` or do any other computations on it, then the compute cost itself would be higher in general, and the reader may not be able to read at peak throughput anyway. This PR is targeted to reduce computations within the reader read path (i.e., inside `readNextEvent()` method), so peak throughput can significantly increase. Note that with this PR, serializing or using `PositionImpl` in any other way can be done in parallel by other threads without impacting the performance of `readNextEvent()` method (for the other methods, the more segments a reader owns, the highest the computation cost for `readNextEvent()` method is).

common/src/main/java/io/pravega/common/util/CopyOnWriteMapUtils.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package io.pravega.common.util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utils class that contains copy-on-write methods for maps.
+ */
+public final class CopyOnWriteMapUtils {
```
tkaitchuck, line 18: Rather than externalizing the logic, I would rather have it internalized to a single class. For example can we Make a `CopyOnWriteHashMap` which internally does the copies? Because this would allow callers to not need to understand the details of the implementation.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -60,6 +63,10 @@
 
     // Base waiting time for a reader on an idle segment waiting for new data to be read.
     private static final long BASE_READER_WAITING_TIME_MS = ReaderGroupStateManager.TIME_UNIT.toMillis();
+    // As an optimization to avoid creating a new ownedSegments map per event read, we define a base map of segments and
```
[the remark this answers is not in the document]
RaulGracia: As I have reported, we have performed a benchmark in which we have serialized each and every `PositionImpl` event. While this is not a realistic case, it was just for benchmarking purposes. In this case, the current state of the PR is 1.5x-2x better than `master`. While this is not the goal of this optimization, we just verified that this scenario is also improved.

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -74,12 +81,18 @@
...
     @GuardedBy("readers")
-    private final Map<Segment, Range> ranges = new HashMap<>();
+    private Map<Segment, Range> ranges = new HashMap<>();
     @GuardedBy("readers")
     private final Map<Segment, Long> sealedSegments = new HashMap<>();
     @GuardedBy("readers")
     private Sequence lastRead;
     @GuardedBy("readers")
+    private Map<Segment, Long> ownedSegments = new HashMap<>();
+    @GuardedBy("readers")
```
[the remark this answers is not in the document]
RaulGracia: One of the approaches taken by this PR is to use as input for building `PositionImpl` objects references to the collections they need to build their state instead of copies (specially map copies are a main performance bottleneck and a source of GC activity). While we do not make unsafe accesses or manipulations from `PositionImpl` objects to these shared collections, multiple `PositionImpl` object may point to `ownedSegments`, `ranges` and `segmentOffsetUpdates`. Shall I remove `@GuardedBy("readers")` from these three collections? 
(Note that start adding locks here would impact on performance, so I prefer not to use additional synchronization. In my view, the current code has been reviewed and intensively benchmarked to be sure that is thread safe).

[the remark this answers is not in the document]
RaulGracia: I have removed the annotation from the collections that are passed as references to the `PositionImpl` builder.

common/src/main/java/io/pravega/common/util/CopyOnWriteMapUtils.java
RaulGracia: Added `CopyOnWriteHashMap` according to your comment.

RaulGracia:
@tkaitchuck I have addressed your last comments. System tests are passing with the last commit and we have verified that performance is good.

PR #4776, opened 2020-05-08
Issue 4775: Add snapshot stage for all release branches

### LINKED ISSUE #4775: Add snapshot stage for all release branches
**Problem description**

**Problem location**
`.travis.yml`

crazyzhou:

**Change log description**  
Add the release branches into the travis snapshot stage.

**Purpose of the change**  
Fixes #4775 

**What the code does**  
Add the release branches into the travis snapshot stage with regex `/^r[0-9]\.[0-9]/`.

**How to verify it**  
Flink connector has similar changes that is working fine.


.travis.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       script: ./gradlew javadocs
 
     - stage: snapshot
-      if: branch = master AND NOT (type = pull_request)
+      if: (branch = master OR branch =~ /^r[0-9]\.[0-9]/) AND NOT (type = pull_request)
```
fpj, line 37: I'm not sure this is going to work. If we merge it to `master`, then builds of the release branches are not going to have this change. Instead, for each for which you want snapshots from, we need to change `.travis.yml`.

crazyzhou: Yes, this change targetting master branch is really for the future releases, i.e. when we have cut branch for r0.8 and we can have `0.8.1` snapshot versions. I'm not sure if it should be cherry-picked to `r0.7` branch because it is now code freeze for release candidate.

andreipaduroiu: Nothing will get cherrypicked in 0.7.1 at this point unless it's a **critical** bug. RC0 went out on Wednesday and it's about to be promoted to a full release in the next couple of days.

crazyzhou: That's fine. If this goes into master, the future release branches will be cut with this change, then this will still benefit the releases after 0.8 in the future.

fpj: Is this regex going to fail if we ever have multiple digits per version? E.g., `r0.10` or `r.10.0`. I think it will match the former, but not the latter. Not that we are any close to such an example... :-)

crazyzhou: Yes, that's right. I only considered single digit, I will make the change to `/^r[0-9]*\.[0-9]*/`.

.travis.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       script: ./gradlew javadocs
 
     - stage: snapshot
-      if: branch = master AND NOT (type = pull_request)
+      if: (branch = master OR branch =~ /^r[0-9]*\.[0-9]*/) AND NOT (type = pull_request)
```
fpj, line 37: It should be `+ `, not `*` as we want at least one digit.

crazyzhou: Fixed as suggestions

andreipaduroiu:
@crazyzhou What is the status of this PR? Do you still wish to pursue it?

@fpj Have your review comments been addressed?

crazyzhou:
@andreipaduroiu @fpj  I still want to pursue this to be merged before 0.8 branch cut.

andreipaduroiu:
@fpj Can you please take a look at this PR? It's waiting on you.

fpj (approved):

crazyzhou:
@fpj I don't have the committer permission, will you or someone else in the team help to click the merge button?

PR #4787, opened 2020-05-13
Issue 4778: Using Unpooled.wrappedUnmodifableBuffer where possible.

### LINKED ISSUE #4778: Reduce buffer copies when creating DataFrames
**Problem description**
DataFrames allocate their own buffers, which forces the upstream code to copy data into that buffer.

**Problem location**
DataFrame.

Reduce buffer copies when creating DataFrames. This can be done by redesigning the internals of the DataFrame to be an ordered list of components, where a component is either a byte array (for random data) or a reference to a BufferView (external data).

andreipaduroiu:
**Change log description**  
- Using `Unpooled.wrappedUnmodifiableBuffer` instead of `Unpooled.wrappedBuffer(...)` in some places.

**Purpose of the change**  
Fixes #4778.

**What the code does**  
- `wrappedUnmodifiableBuffer` takes in an array of `ByteBufs` and uses it as is; this is different from `wrappedBuffer` which creates a `CompositeByteBuf` that can be modified, and as such takes more time to construct.

**How to verify it**  
All tests must pass.


RaulGracia:
@andreipaduroiu it looks like there is some issue now with codecov that prevents me to merge this (it wasn't before).

ravisharda:
> @andreipaduroiu it looks like there is some issue now with codecov that prevents me to merge this (it wasn't before).

Strange that the code coverage diff reflects files changed in a different PR #4692 . The same issue is also failing code coverage for other PRs like 4799. 

Does anyone know why this might be happening?

fpj:
Codecov seems to be checking the coverage difference for the commits in the pull request, which includes the merges. I checked the files added here individually, and they seem to be adding to the coverage, but perhaps the additions to the coverage from the changes here are no compensating for the negative changes to coverage?

if anyone has a different explanation, then I'd love to hear. To test, I have changed the requirement from patch to project to see if it changes and retriggered the build.

fpj:
I'm not sure why codecov is not producing a codecov/project report. In any case, I have rebuilt another PR and it worked fine, so it might be something with this PR. I have changed back the requirement for codecov/patch to pass until we understand better what needs to happen.

PR #4800, opened 2020-05-19
Issue 4783: Timebound check for completion of workflow to avoid log flooding

### LINKED ISSUE #4783: Timebound checking for completion of workflow, and throw timeout exception to caller if limit is exceeded
**Problem description**
Currently, when client requests for a workflow from controller, if the request is accepted, its processing is scheduled asynchronously, and controller keeps checking for completion of workflow indefinitely. Timebound this check and throw time exceeded exception to the caller and let them retry their request. 

**Problem location**
StreamMetadataTasks.java

Add a timeout in `checkDone` method.

shiveshr:
**Change log description**  
Add timeout to workflow completion check and throw timeout exception if the workflow is not finished within that time. 

**Purpose of the change**  
Fixes #4783 

**What the code does**  
StreamMetadataTasks.java `checkDone` method was checked for in a loop indefinitely. Added timeout which will throw the timeout exception to the caller. 
ControllerServiceImpl.java will throw DEADLINE_EXCEEDED response to the client upon receiving timeout exception. 
RetryHelper.java adds a new helper method that terminates the loop if timeout elapses. 

**How to verify it**  
Unit tests added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
@@ -1078,7 +1078,7 @@ public class StreamMetadataTasks extends TaskBase {
     }
 
     @VisibleForTesting
-    void setCompletionTimeoutMillis(long timeoutMillis) {
+    public void setCompletionTimeoutMillis(long timeoutMillis) {
         completionTimeoutMillis.set(timeoutMillis);
     }
 }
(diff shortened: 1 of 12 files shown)


controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -104,6 +106,7 @@
 
     private static final TagLogger log = new TagLogger(LoggerFactory.getLogger(StreamMetadataTasks.class));
     private static final long RETENTION_FREQUENCY_IN_MINUTES = Duration.ofMinutes(Config.MINIMUM_RETENTION_FREQUENCY_IN_MINUTES).toMillis();
+    private static final long COMPLETION_TIMEOUT_MILLIS = Duration.ofMinutes(2).toMillis();
```
shrids, line 109: Should this be configurable?

shiveshr: There is no point making this a controller level configuration. I have ensured that for testing we are able to set a much smaller value through a setter.

shrids:
This can be merged once `pravega-branch-builder/2450/` is complete.

shrids:
`test-pravega-continuous/2874/` is successful with this change set.

PR #4806, opened 2020-05-21
Issue 4805: Micro benchmark for client append path

### LINKED ISSUE #4805: Benchmark for Client side append path using mock server
**Problem description**
It is nice to have a performance benchmark for the client append path using a mock server that replies ok immediately to any incoming requests. The latency measured in this benchmark would include client code latency and kernel latency and it does not include network and server side latency. 

The result can be used to compare with the Native client performance. 

**Problem location**
Client code

Tristan1900:
**Change log description**  
Print out total latency in the self testing tool for better performance comparision.

**Purpose of the change**  
Closes #4805 

**What the code does**  

**How to verify it**


RaulGracia:
@Tristan1900 is this different from the existing SelfTester tool? https://github.com/pravega/pravega/wiki/Local-Stress-Testing

Tristan1900:
Thanks @RaulGracia . Yes, according to the doc you shared, I think SelfTester tool is mainly to test segmentstore performance given various situations. However, this benchmark is to test only client performance with a mocking segmentstore that replies ok to every request without checking anything. The goal of this benchmark to compare with the Rust client that uses the same setup.

RaulGracia:
Ok, but note that there is an option in this tool that says:
```
InProcessMock: Executes a Streaming Append test using the Client but with a no-op Segment Store (zero-latency store).
```
Isn't that what you want? A mock Segment Store to stress the client? Maybe @andreipaduroiu can clarify this, but it looks that could fit your purposes. And if it does, maybe improving this existing tool could be better than adding something new, right?

Tristan1900:
@RaulGracia Thanks for pointing out! I didn't see that option. I will check with Andrei to see if it fits the purpose.

Tristan1900:
@andreipaduroiu Thanks Andrei. Right now the Native client doesn't have the read path fully completed, maybe I can improve the `InProcessMockClientAdapter` after I finish the Native client Impl since by then I will have a better understanding of the read path. I just spent a couple of hours looking at the `read` in the segmentstore and it doesn't look trivial to me.

Issue #4813, opened 2020-05-27
Rate computation in SegmentAggregates should account for silent period


shiveshr:
**Problem description**
We compute rate as weighted mean with a decay. This applies a weight to new instant rate to basically give it lower weight and decays by same factor.

So if there are spikes, they do not make the rate jump up. But this also has counter productive effect of when there is no traffic for a long, the new super low rate also decays the existing rate more slowly.

The problem really happens because we rate computation only whenever we receive an new bath of event and not periodically even for silent segments.

A rate is computed as following:
```
timeDelta = Current clock time - last clock time
instant rate = number of events / timeDelta
rate = rate + (alpha * (instantRate - rate))
```
Alpha for 2 minute rate is about 0.05.

So whats happening is, within first few seconds about 26 events are received which lets the rate computation to be around 4 events per second.

Now for 10 minutes there is no traffic.

When the traffic is received after 10 minutes, the previous rate is decayed slowly and new rate does not affect it too much as the new rate is `26/60000`.

So the math is `4 * 0.95 + 0.05 * 26/60000`. 

So the fix is to decay the rate even for time when there is no traffic.

This can be done by decaying the rate multiple times for an equivalent of 10 minutes in this example with instant rate being 0.

**Problem location**
SegmentAggregates

**Suggestions for an improvement**
Account for silent periods in instant rate computation.

shrids:
closing this since the PR has been merged.

PR #4817, opened 2020-05-27
Issue 4813: Rate computation in SegmentAggregates should account for silent period 

shiveshr:

**Change log description**  
Calls the delay function for each silent period between two update notifications on segmentaggregates.  

**Purpose of the change**  
#4813 

**What the code does**  
The weighted average rate computation takes the instant rate as events / time between two subsequent update calls. If there is a long delay between the two calls, then it calls the decay function with a small instant rate which effectively decays the rate by a very small factor without accounting for additional decays for silent periods.
So what we do now instead is to take the instant rate as the rate in the latest tick. And all ticks between current update call and the last update call are treated as silent periods and rate is decayed for each of those ticks before applying the current instant rate. 

**How to verify it**  
Unit test added.


shrids:
`test-pravega-continuous/2875/` system test run.

PR #4821, opened 2020-05-28
Issue 4820: Hello Reply doesn't flush to the Channel immediately

### LINKED ISSUE #4820: Hello Reply doesn't flush to the Channel immediately
**Problem description**
When a client creates a connection to the server, it will send Hello to the server, and the server sends a Hello back either, however, the CommandEncoder doesn't flush the Hello immediately, which would make the native client block at verifying connection.
**Problem location**
CoomandEncoder.Java

Add a FlushRequire() for the Hello

zwx14700:
**Change log description**  
Flush the hello when it is encoded.

**Purpose of the change**  
 Fixes #4820 

**What the code does**  
Flush the hello when it is encoded.

**How to verify it**  
All tests should pass.


RaulGracia:
@tkaitchuck @shrids @zwx14700 I have a question about this: would the flushing of "Hello" messages have any impact on the size of batches? Just wondering that this could be detrimental to the client's performance by any chance.

PR #4829, opened 2020-05-28
Issue 4791: Add a new controller API to fetch segments for a given epoch.

### LINKED ISSUE #4791: Enable a Controller API to fetch all segments in a given transaction.
**Problem description**
Create a new Controller API to fetch all segments for a given transaction id or epoch.

**Problem location**
Controller

Add a new GRPC API to enable this on the controller.

shrids:
**Change log description**  
- Add a new controller API to fetch segments for a given epoch.
- Ensure `io.pravega.client.stream.impl.TransactionalEventStreamWriterImpl#getTxn` API uses this API.

**Purpose of the change**  
Fixes #4791 

**What the code does**  
The code enables a new GRPC controller API to fetch segments for a given epoch. It also ensures the `TransactionalEventStreamWriter` uses this API to fetch segments for a given Txn to fetch the segments. 

**How to verify it**  
Note: Junit tests are being added here.


shrids:
The Transaction event stream writer API `io.pravega.client.stream.TransactionalEventStreamWriter#getTxn` earlier used `controller.getCurrentSegments()` API which is incorrect. This PR ensures the correct segments are picked for a given transaction id.

client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java
```diff
@@ -198,13 +199,19 @@ private void throwIfClosed() throws TxnFailedException {
...
     public Transaction<Type> getTxn(UUID txId) {
-        StreamSegments segments = getAndHandleExceptions(
-                controller.getCurrentSegments(stream.getScope(), stream.getStreamName()), RuntimeException::new);
+        // check if the transaction is open.
         Status status = getAndHandleExceptions(controller.checkTransactionStatus(stream, txId), RuntimeException::new);
         if (status != Status.OPEN) {
             return new TransactionImpl<>(writerId, txId, controller, stream);
         }
-        
+
```
shiveshr, line 207: question: why do we need to get the segments for a txn again? when you create a txn, controller returns the txn segments, right?

so when is this getTxn api useful?

shrids: `TransactionalEventStreamWriter` enables users to created transactions. The user can use 

1. `TransactionalEventStreamWriter#beginTxn()` to create a new transaction
2. `TransactionalEventStreamWriter#getTxn(UUID txnId)` to fetch an already created transaction given its id. This is useful when the user knows a txnId and wants to perform operations on that transaction.

Both the above APIs return a `io.pravega.client.stream.Transaction` using which the user can 
a. Write events to a transaction, `Transaction#writeEvent`
b. Commit a transaction, `Transaction#commit() `
c. Abort a transaction, `Transaction#abort()`
d. Check the status of the transaction `Transaction#checkStatus()`.

The new controller API ensures the correct segments are fetched when the user uses  [2.] getTxn(txnId) API.

Issue #4819, opened 2020-05-27
Leak reported by Netty apparently on SecureConnectionFactoryImplTest 


eolivelli:
**Problem description**
I saw this failure while running the full tests on current git sha 37614da10725e0172ff0d3b8bc2e9ea81795258a

`./gradlew clean build`

The error is not reproducible but the stacktrace gives an hint about a possible OpenSslServerContext created during the setup of SecureConnectionFactoryImplTest test but never released

```
io.pravega.client.netty.impl.SecureConnectionFactoryImplTest > getActiveChannelTestWithConnectionPooling FAILED
    java.lang.AssertionError: RESOURCE LEAK: LEAK: {}.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.{}
        at org.junit.Assert.fail(Assert.java:88)
        at io.pravega.test.common.LeakDetectorTestSuite$ResourceLeakLoggerFactory$ResourceLeakAssertionLogger.error(LeakDetectorTestSuite.java:241)
        at io.pravega.test.common.LeakDetectorTestSuite$ResourceLeakLoggerFactory$ResourceLeakAssertionLogger.error(LeakDetectorTestSuite.java:235)
        at io.netty.util.ResourceLeakDetector.reportTracedLeak(ResourceLeakDetector.java:311)
        at io.netty.util.ResourceLeakDetector.reportLeak(ResourceLeakDetector.java:300)
        at io.netty.util.ResourceLeakDetector.track0(ResourceLeakDetector.java:257)
        at io.netty.util.ResourceLeakDetector.track(ResourceLeakDetector.java:245)
        at io.netty.buffer.AbstractByteBufAllocator.toLeakAwareBuffer(AbstractByteBufAllocator.java:41)
        at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:331)
        at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:185)
        at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:176)
        at io.netty.buffer.AbstractByteBufAllocator.buffer(AbstractByteBufAllocator.java:113)
        at io.netty.handler.codec.base64.Base64.encode(Base64.java:123)
        at io.netty.handler.ssl.SslUtils.toBase64(SslUtils.java:347)
[...]

PR #4831, opened 2020-05-29
Issue 4819: SecureConnectionFactoryImplTest - release netty resources

eolivelli:
**Change log description**  
Ensure that SecureConnectionFactoryImplTest releases Netty resources

**Purpose of the change**  
Relates to #4819

**What the code does**  
Release all Netty resources

**How to verify it**  
The test suite should pass


eolivelli:
@ravisharda can you please take a look ?

client/src/test/java/io/pravega/client/netty/impl/ConnectionFactoryImplTest.java
```diff
@@ -106,8 +107,12 @@ public void initChannel(SocketChannel ch) throws Exception {
 
     @After
     public void tearDown() throws Exception {
-        serverChannel.close();
-        serverChannel.closeFuture();
+        serverChannel.close().awaitUninterruptibly();
+        bossGroup.shutdownGracefully().await();
+        workerGroup.shutdownGracefully().await();
+        if (sslCtx != null) {
+            ReferenceCountUtil.safeRelease(sslCtx);
```
ravisharda, line 114: Do we really need to release the SSL Context explicitly here? Could you please describe the reason? 

For what it's worth, the TLS example from Netty also doesn't do it: 
https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/securechat/SecureChatServer.java

eolivelli: That **sslCtx** is an [ReferenceCountedOpenSslContext](https://netty.io/4.1/api/io/netty/handler/ssl/ReferenceCountedOpenSslContext.html)
and it needs an explicit release.
The example you are linking is very trivial and maybe it is not using OpenSSL

PR #4838, opened 2020-06-01
Issue 4808: (SegmentStore) Using BufferViews for Table Segment APIs

### LINKED ISSUE #4808: Optimize Table Segment API Calls
**Problem description**
Apply some of the recent improvements on the Stream Segment Append/Read path to the Table Segment API path:
- Reduce buffer copies both in and out.
- Reduce the number of list copies.

**Problem location**
PravegaRequestProcessor, WireCommands, Segment Store (Table Segment Extension).

See description above.

andreipaduroiu:
**Change log description**  
- Changed `TableStore` APIs to accept `BufferView` instead of `ArrayView`. This helps reduce the number of times we copy key/values both inbound (from 3 to 1) and outbound (from 2 to 1).
- Unified all implementations of `BufferView` to have a common `equals` and `hashCode`. 
    - Removed `HashArray` - it was made redundant by the change above.
- Extended `BufferView.Reader` with new method to help reduce number of copies.

**Purpose of the change**  
Fixes #4808.

**What the code does**  
TBD

**How to verify it**  
All existing tests must pass. New tests added where needed.


common/src/main/java/io/pravega/common/util/ByteArraySegment.java
```diff
@@ -169,6 +171,45 @@ public void copyTo(OutputStream stream) throws IOException {
...
+    }
+
+    private boolean equals(ArrayView other) {
+        if (this.length != other.getLength()) {
+            return false;
+        }
+
+        byte[] otherArray = other.array();
+        int otherOffset = other.arrayOffset();
+        for (int i = 0; i < this.length; i++) {
```
eolivelli, line 199: java.util.Arrays has HotSpot  intrinsified methods for array comparisons.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/ServerConnectionInboundHandler.java
```diff
@@ -53,7 +53,15 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
         if (requestProcessor == null) {
             throw new IllegalStateException("No command processor set for connection");
         }
-        cmd.process(requestProcessor);
+        try {
+            cmd.process(requestProcessor);
+        } catch (Throwable ex) {
+            // Release buffers in case of an unhandled exception.
+            if (cmd instanceof WireCommands.ReleasableCommand) {
```
eolivelli, line 60: What about adding an Utility method
Wire command.ReleasableCommand#safeRelease(cmd) ?

This way we can save from writing **instanceof** that most of the times looks like a code smell

andreipaduroiu:
GitHub and Travis managed to confuse themselves with this PR. I am closing in favor of #4841. @eolivelli I will answer your comments here and take any feedback in that PR.

common/src/main/java/io/pravega/common/util/ByteArraySegment.java
andreipaduroiu: They do, but they only compare whole arrays. I have array slices, so those methods aren't very helpful. Looking at them, they do exactly what I'm doing (byte-by-byte comparison) so not much of an improvement there.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/ServerConnectionInboundHandler.java
andreipaduroiu: I thought about that, but we are actually aiming to eliminate Netty from our Wire Protocol altogether due to poor performance in some cases, so we might not need to release anything afterall. That's also a reason why we're consolidating our buffers under `BufferView` and not exposing `ByteBuf` in our internal/external APIs.

PR #4844, opened 2020-06-04
Issue 4845: StreamInfo should indicate if the Stream is sealed.

### LINKED ISSUE #4845: StreamInfo should indicate if the Stream is sealed
**Problem description**
StreamInfo obtained via `io.pravega.client.admin.impl.StreamManagerImpl#getStreamInfo(java.lang.String, java.lang.String)`
 should indicate if the Stream is sealed. 

**Problem location**
StreamManagerImpl

fix the issue.

shrids:
**Change log description**  
StreamInfo obtained via `io.pravega.client.admin.impl.StreamManagerImpl#getStreamInfo(java.lang.String, java.lang.String)`
should indicate if the Stream is sealed.

**Purpose of the change**  
Fixes #4845 

**What the code does**  
`io.pravega.client.admin.impl.StreamManagerImpl#getStreamInfo(java.lang.String, java.lang.String)`
now indicates if the Stream is sealed.
`Controller.getCurrentSegments()` returns an empty list if the Stream is sealed. The code uses this behavior to check if the Stream is sealed.

**How to verify it**  


client/src/main/java/io/pravega/client/admin/StreamInfo.java
```diff
@@ -45,4 +46,9 @@
      * {@link StreamCut} representing the current HEAD of the stream.
      */
     private final StreamCut headStreamCut;
+
+    /**
+     * Represents the stream state. It is set to True if the stream is Sealed.
```
andreipaduroiu, line 51: I wouldn't say "stream state". Maybe at this point sealed is the only way to describe a Stream's state, but maybe in the future we want to add some other info. I would keep this simple. For example: "Indicates whether the Stream is sealed (true) or not (false). If a stream is sealed, the no further Events can be written to it." ( <-- feel free to include any other operations that will be denied)

shrids: thanks, improved the comment.

client/src/test/java/io/pravega/client/admin/impl/StreamManagerImplTest.java
```diff
@@ -127,9 +132,75 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
...
+        // Create a StreamManager
+        @Cleanup
+        final StreamManager streamManager = new StreamManagerImpl(mockController, connectionFactory);
+
+        // Create a scope and stream and seal it.
+        streamManager.createScope(defaultScope);
+        streamManager.createStream(defaultScope, streamName, StreamConfiguration.builder()
+                                                                                .scalingPolicy(ScalingPolicy.fixed(3))
+                                                                                .build());
+        streamManager.sealStream(defaultScope, streamName);
```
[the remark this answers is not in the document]
shrids: I have added an assert to verify that `info.isSealed()` returns false in the test `io.pravega.client.admin.impl.StreamManagerImplTest#testStreamInfo`.
Based on this comment I have improved the integration test case ( `io.pravega.test.integration.StreamRecreationTest#testStreamRecreation`) to validate `info.isSealed()` before and after sealing the Stream

PR #4850, opened 2020-06-06
Issue 4849: Update NOTICE

### LINKED ISSUE #4849: Update NOTICE
**Problem description**
We need to update the year in the NOTICE file.

**Problem location**
NOTICE file.

Change the range to `2017-2020`.

fpj:

**Change log description**  
* Updates NOTICE file with the year range `2017-2020`.

**Purpose of the change**  
Fixes #4849 

**What the code does**  
No code change, only a change to the NOTICE file to update the range of years the copyright applies to.

**How to verify it**  
It should build as usual.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,4 +1,4 @@
-Copyright (c) 2017-2019 Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright (c) 2017-2020 Dell Inc., or its subsidiaries. All Rights Reserved.
 
 This software contains source code from Apache BookKeeper, distributed under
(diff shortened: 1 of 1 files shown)


NOTICE
```diff
@@ -1,6 +1,6 @@
-Copyright (c) 2017-2019 Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright (c) 2017-2020 Dell Inc., or its subsidiaries. All Rights Reserved.
```
derekm, line 1: Other Dell license headers omit the years. Should we do the same here?

PR #4855, opened 2020-06-09
Issue 4854: Fix erroneous configuration property names and use new names in system tests

### LINKED ISSUE #4854: Bugs in names of updated configuration properties
**Problem description**

Upon [manually listing](https://github.com/pravega/pravega/wiki/Configuration-Name-Changes) all the new and old configuration names side-by-side, a couple of bugs were found: 
1. `singlenode.security.tls.keyStore.pwd.location` should have used `singlenode.keyStoreJKSPasswordFile` as legacy name. It instead uses it as the default value. 

https://github.com/pravega/pravega/blob/efa9e08b205c6c45631b5bd3b492514eeae36421/standalone/src/main/java/io/pravega/local/SingleNodeConfig.java#L39

2. AutoScaler config uses `threadPoolSize` as new name and `threadPool.size` as old name. It should have been the other way around. 

https://github.com/pravega/pravega/blob/efa9e08b205c6c45631b5bd3b492514eeae36421/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScalerConfig.java#L35

**Problem location**

ravisharda:
**Change log description**  
Fix erroneous configuration property names and use new names in system tests. 

**Purpose of the change**  
Fixes #4854 . 

**What the code does**  
* Fixes bugs in configuration properties setup. 
* Replaces configuration property names for system tests to new ones.

**How to verify it**  
All unit and integration tests must pass. Additionally, system tests must pass.


shrids:
https://github.com/pravega/pravega/blob/efa9e08b205c6c45631b5bd3b492514eeae36421/docker/pravega/scripts/init_controller.sh#L15
This config entry is not present in https://github.com/pravega/pravega/wiki/Configuration-Name-Changes and is not being used.

ravisharda:
> https://github.com/pravega/pravega/blob/efa9e08b205c6c45631b5bd3b492514eeae36421/docker/pravega/scripts/init_controller.sh#L15
> 
> 
> This config entry is not present in https://github.com/pravega/pravega/wiki/Configuration-Name-Changes and is not being used.

Yes, this property isn't managed by the Config files. But it is set directly in the shell script. I am not aware of why and how it is used.

ravisharda:
@shrids 
> Can we also modify the properties used by system tests....

That's a good catch. I've pushed the change to update them.

ravisharda:
I've triggered a systems test run on branch builder. The job id is 2519.

ravisharda:
System tests passed with the latest changes - test-pravega-continuous job 2918.

shiveshr:
> add_system_property "controller.server.store.host.type" "Zookeeper" 
This is telling the system to use zookeeper for cluster management (host store type). i think this is already the default value too. and anyway its ok if its supplied explicitly too.

PR #4863, opened 2020-06-10
Issue 4830: Lazily instantiate Auto Scale Processor's writer 

### LINKED ISSUE #4830: Lazily instantiate writer in AutoScaleProcessor
**Problem description**
During bootstrap, if containers are still coming up and we attempt to create auto scale processor's event stream writer, it fails and creates noisy logs about the failure which eventually goes away. The error continues until all internal streams are created, so that leads to log pollution on controller and segment store. 

Its perhaps better to lazily instantiate the writer when the first scale event needs to be sent. That will avoid all noisy error logs at the time of segment store instantiation and equivalent logs in controller too. 

**Problem location**
AutoScaleProcessor.java

Create the writer lazily

shiveshr:
**Change log description**  
Lazily instantiate the writer in auto scale processor when the scale signal needs to be sent. 

**Purpose of the change**  
Fixes #4830

**What the code does**  
In AutoScaleProcessor, we have a writer that is created when the Autoscaleprocessor is instantiated. 
This writer attempts to contact controller and get segments for the internal stream. 
During bootstrap of a cluster as controller is still coming up and attempting to create the internal stream, any requests from this writer can fail (and often does) with warn or error level client logs describing the failure in segment store process. Equivalent warn level logs are present in controller as it is unable to process the request. 
This leads to unnecessary log pollution and people often focus on this log while facing difficulty during bootstrap. 

By starting the writer lazily, at the time when first scale request needs to be sent out, ensures that we avoid this unnecessary error log in multiple places. 

Code changes:
1. the bootstrap method is replaced by `getWriter` method which idempotently sets writer future. 
2. the isInitialized method is no longer required as the request is asynhronously chained to the `getWriter` future. 
3. Added a protected `getTimeMillis` method to effectively make it testable with a clock that can be progressed without having to rely on system clock. 

**How to verify it**  
Unit test added. 
All integration and system tests should pass.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
```diff
@@ -116,48 +122,51 @@
...
+    @Synchronized
     public void close() {
-        val w = this.writer.get();
-        if (w != null) {
-            w.close();
-            this.writer.set(null);
+        if (writer != null) {
+            writer.cancel(true);
+
+            if (!writer.isCancelled() && !writer.isCompletedExceptionally()) {
```
shrids, line 139: we can use `io.pravega.common.concurrent.Futures#isSuccessful` here.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
```diff
@@ -61,10 +66,12 @@
 
     private final EventStreamClientFactory clientFactory;
     private final Cache<String, Pair<Long, Long>> cache;
-    private final AtomicReference<EventStreamWriter<AutoScaleEvent>> writer;
+    @GuardedBy("$lock")
+    private CompletableFuture<EventStreamWriter<AutoScaleEvent>> writer;
```
andreipaduroiu, line 70: Non-final fields make it look like something's not right. Let's see if we can make this better. Would this work?
1. Make this final and remove the "GuardedBy"
2. In the `getWriter` method, do exactly what you're doing, but when the retry loop is done, set the result on this future, whether it be the writer or an exception. You can use `Futures.completeAfter` (I may forget the actual name). 

This way we don't need a new lock and we can make this field final. `CompletableFuture` internally is thread safe so it takes care of any concurrency issues for us.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
```diff
@@ -116,48 +123,51 @@
...
     }
 
     @Override
+    @Synchronized
     public void close() {
-        val w = this.writer.get();
-        if (w != null) {
-            w.close();
-            this.writer.set(null);
+        if (writer != null) {
```
andreipaduroiu, line 137: Writer should always be non-null.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
```diff
@@ -116,48 +123,51 @@
...
-        executor.schedule(
-                () -> Retry.indefinitelyWithExpBackoff(100, 10, 10000, this::handleBootstrapException)
-                        .runInExecutor(() -> bootstrapOnce(clientFactory), executor),
-                10, TimeUnit.SECONDS);
+    @Synchronized
+    private CompletableFuture<EventStreamWriter<AutoScaleEvent>> getWriter() {
+        if (writer == null) {
+            AtomicReference<EventStreamWriter<AutoScaleEvent>> w = new AtomicReference<>();
+            EventWriterConfig writerConfig = EventWriterConfig.builder().build();
+            writer = Retry.indefinitelyWithExpBackoff(100, 10, 10000, this::handleBootstrapException)
```
andreipaduroiu, line 157: This indefinite retry won't work well in case we fail to bootstrap and we decide to shut down. It will likely leave this running forever. You should include a condition that can cancel it. My recommendation would be to check `this.writer.isDone` (it will be true when either a writer has been set or when it got completed with an exception).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
```diff
@@ -192,113 +202,99 @@ static boolean hasTlsEnabled(@NonNull final URI controllerURI) {
...
-        } else {
-            writer.writeEvent(event.getKey(), event)
+        getWriter().thenCompose(w -> w.writeEvent(event.getKey(), event)
                     .whenComplete((r, e) -> {
                         if (e != null) {
                             log.error(event.getRequestId(), "Unable to post Scale Event to RequestStream '{}'.",
                                     this.configuration.getInternalRequestStream(), e);
                         } else {
-                            log.debug(event.getRequestId(), "Scale Event posted successfully: {}.", event);
+                            log.info(event.getRequestId(), "Scale Event posted successfully: {}.", event);
```
andreipaduroiu, line 258: Please revert this. This will result in too much garbage in the segment store logs.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
shiveshr: We can avoid non final writer. But the objective is to make the bootstrap of writer be triggered exactly once which will eventually complete the future. 
In existing code, the bootstrap was being invoked from the constructor.
But now we want to do it lazily, upon first request. But there could be multiple concurrent requests. 
So we need some syncrhonization to make sure exactly one of them end up attempting to create a writer. 
So syncrhonization is really for that. And instead of using a non-final boolean which is synchronized (or atomic boolean) i had chosen to do with writerFuture = null. 

I have an alternate idea though -- which would still require using atomic boolean.. let me make and push that change and see if that is better.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
```diff
@@ -127,35 +131,44 @@
...
-                10, TimeUnit.SECONDS);
+        AtomicReference<EventStreamWriter<AutoScaleEvent>> w = new AtomicReference<>();
+
+        Futures.completeAfter(() -> Retry.indefinitelyWithExpBackoff(100, 10, 10000, this::handleBootstrapException)
+                                         .runInExecutor(() -> bootstrapOnce(clientFactory, w), 
+                                                 executor).thenApply(v -> w.get()), writer);
     }
 
     private void handleBootstrapException(Throwable e) {
         log.warn("Unable to create writer for requeststream: {}.", LoggerHelpers.exceptionSummary(log, e));
```
shrids, line 158: If `startInitWriter` is `false` then we would be logging this line repeatedly. 
This would clutter the logs with WARN messages. (including standalone) Can we avoid it?

PR #4867, opened 2020-06-11
Issue 4677: Offset/Version Table Entry Iteration

### LINKED ISSUE #4677: Ability to read new entries in a TableSegment
**Problem description**
Currently table segment allows for updating entries, and iterating over all entries in the segment. However it would be useful to iterate over all the TableEntries after a given offset. 

This could be used to implement a KV version of StateSynchronizer. Internally it would need to implement 'fetchUpdates' to get all the changes to the table since the last time it called. 

Depends on :
- #4792.
- #4793 

**Problem location**
TableSegment

We can add this to TableSegment directly. It does not need to be on TableStream.

co-jo:
**Change log description**  
* Implement the iteration based APIs for position based `TableSegment` reads.
* This adds a new iterator which starts at a given offset/position in a `TableSegment`, and deserializes these reads into TableEntries (as apposed to the `TableBucket` approach that currently exists). 

**Purpose of the change**  
Closes #4677.

**What the code does**  

In addition to the change log description, the new `entryIterator` has the following semantics:

* If the start position provided in the `ReadTableEntriesDelta` wire command proceeds the `COMPACTION_OFFSET`, we begin starting at the `COMPACTION_OFFSET`. 
* At the time the `newIterator` is constructed, the current segment length is read, which will act as our upper bound. Meaning that even if other entries are appended, we will still consider this point as our segment end.
* The `TableEntryIterator` also uses the same `MAX_READ_SIZE` from the `PravegaRequestProcessor`.

Three new classes (related to iteration) are introduced: `TableEntryIterator`, `EntryIteratorItemImpl` and `EntryIteratorState`. When designing these classes, one of the central goals was to closely follow the general design of the previous `key/entryIterator` that already exist. I believe this may introduce some inefficiencies with you will likely want to address:

**Areas of concern**

`PravegaRequestProcessor.readTableEntriesDelta`: Conforming to the existing iterator structure, I introduce a potentially redundant packing/unpacking of our `EntryIteratorState` object for every `TableEntry` in the iteration.

`EntryIteratorItemImpl`: Unlike a `TableBucket` which may map to mutliple `TableEntry`s, each iteration will always return just one `TableEntry`. Again, using these existing structures, the `entries` field in our `EntryIteratorItem` is always a `singletonList`.

**Testing**

Besides the standard test classes for each of these new iteration classes that were introduced, you will notice two new classes: `TableContext` and `ContainerMock`. These classes were parsed out of `ContainerTableExtensionImplTests` to promote reusability.

The `TableEntryIterator` very much requires some `TableSegment` to add to and iterate over to ensure correctness.

**In Progress**

You will notice that I do not have any tests directly testing the new methods in the `PravegaRequestProcessor`. Following troubles with setting up the mocks, it is still in progress.

**How to verify it**  

Run the unit tests. More involved testing is forth coming.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -466,6 +504,29 @@ public String toString() {
         }
     }
 
+    // endregion
+
+    // region EntryIteratorItemImpl
+
+    @RequiredArgsConstructor
+    private class EntryIteratorItemImpl<T> implements IteratorItem<T> {
```
andreipaduroiu, line 512: Is there really no way to unify this with `IteratorItemImpl`?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/EntryIteratorState.java
```diff
@@ -0,0 +1,137 @@
...
+        this.shouldClear = shouldClear;
+        this.deletionRecord = deletionRecord;
+    }
+
+    boolean isValid(long position) {
+        return position >= 0;
+    }
+
+    @Override
+    public String toString() {
```
andreipaduroiu, line 64: Tip: If all you need `toString` for is for debugging (watching), consider using Lombok's `@ToString` annotation which will auto-generate this for you (you can even pick the fields you want to include).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/EntryIteratorState.java
```diff
@@ -0,0 +1,137 @@
...
+     * Serializes this IteratorState instance into an {@link ArrayView}.
+     *
+     * @return The {@link ArrayView} that was used for serialization.
+     */
+    @SneakyThrows(IOException.class)
+    public ArrayView serialize() {
+        return SERIALIZER.serialize(this);
+    }
+
+    private static class IteratorStateBuilder implements ObjectBuilder<EntryIteratorState> {
```
andreipaduroiu, line 92: You don't need to specify these fields in here. The `@Builder` annotation will take care of this for you. The only reason we need to declare this class is because we want it to implement `ObjectBuilder<>`.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/EntryIteratorState.java
```diff
@@ -0,0 +1,137 @@
...
+            return 0;
+        }
+
+        @Override
+        protected void declareVersions() {
+            version(0).revision(0, this::write00, this::read00);
+        }
+
+        private void read00(RevisionDataInput revisionDataInput, IteratorStateBuilder builder) throws IOException {
+            builder.position = revisionDataInput.readCompactLong();
```
andreipaduroiu, line 121: Just a heads up that CompactLong can only serialize up to 2^62 (and no negatives). This sounds OK for now, since we do not expect negative offsets and 2^62 is still a huge number.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryIterator.java
```diff
@@ -0,0 +1,328 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Iterates through {@link TableBucket}s in a Segment.
+ * @param <T> Type of the final, converted result.
+ */
+@Slf4j
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@ThreadSafe
+class TableEntryIterator<T> implements AsyncIterator<T> {
```
andreipaduroiu, line 53: `TableEntryDeltaIterator`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryIterator.java
```diff
@@ -0,0 +1,328 @@
...
+                    if (entry == null) {
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        return this.resultConverter.apply(entry);
+                    }
+                });
+    }
+
+    public boolean endOfSegment() {
+        return this.currentBatchOffset >= (startOffset + maxLength);
```
andreipaduroiu, line 94: Nit: either use `this.` for all private fields or for none. Try to be consistent.

andreipaduroiu: `this.currentBatchOffset` is Guarded By `this` but you don't use a synchronized block here.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryIterator.java
```diff
@@ -0,0 +1,328 @@
...
+        if (entry != null) {
+            return CompletableFuture.completedFuture(entry);
+        }
+
+        return fetchNextTableEntriesBatch().thenApply(val -> getNextEntryFromBatch());
+    }
+
+    private synchronized Map.Entry<EntryIteratorState, TableEntry> getNextEntryFromBatch() {
+        if (this.currentEntry != null) {
+            Map.Entry<EntryIteratorState, TableEntry> next = this.currentEntry.next();
```
andreipaduroiu, line 108: Tip: use `val` to avoid having to write long type names (i.e `val next = this.currentEntry.next()`)

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryIterator.java
```diff
@@ -0,0 +1,328 @@
...
+        }
+        ReadResult result = this.segment.read(startOffset, length, timer.getRemaining());
+        return AsyncReadResultProcessor.processAll(result, this.executor, timer.getRemaining())
+                .thenApply(data -> parseEntries(data, startOffset, length));
+    }
+
+    @SneakyThrows(IOException.class)
+    private List<Map.Entry<EntryIteratorState, TableEntry>> parseEntries(BufferView data, long startOffset, int readLength) {
+
+        long maxOffset = startOffset + readLength;
```
andreipaduroiu, line 144: Tip: if you declare a variable that you do not expect (or want) to change throughout your method, add the `final` modifier or declare it with `val` (which makes it final). This way the compiler will catch any attempt you make to modify it, which can, in turn, prevent bugs from developing.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryIterator.java
```diff
@@ -0,0 +1,328 @@
...
+    static <T> Builder<T> builder() {
+        return new Builder<>();
+    }
+
+    /**
+     * Builder for the {@link TableIterator} class.
+     */
+    static class Builder<T> {
+
+        private int maxLength = MAX_READ_SIZE;
```
andreipaduroiu, line 206: You don't need to declare this class. Adding `@Builder` will do it for you.

The reason I did it "manually" in `TableIterator` was because `build()` had to do a significant amount of custom work, so the `@Builder` annotation wouldn't have helped.

In your case, however, this is plain boilerplate code so that annotation will save you from having to do that.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/TableContext.java
```diff
@@ -0,0 +1,101 @@
...
+import io.pravega.segmentstore.server.SegmentContainer;
+import io.pravega.segmentstore.server.UpdateableSegmentMetadata;
+import io.pravega.segmentstore.server.containers.StreamSegmentMetadata;
+import io.pravega.segmentstore.storage.cache.CacheStorage;
+import io.pravega.segmentstore.storage.cache.DirectMemoryCache;
+import java.util.Random;
+import java.util.concurrent.ScheduledExecutorService;
+import lombok.val;
+
+public class TableContext implements AutoCloseable {
```
andreipaduroiu, line 23: Please add a quick Javadoc on this class explaining its purpose. I know it's in the test package, but since you took it out of the class it was originally in, we need to explain what it should be used for.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/ContainerMock.java
```diff
@@ -0,0 +1,220 @@
...
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+import org.junit.Assert;
+
+public class ContainerMock implements SegmentContainer {
```
andreipaduroiu, line 40: Is there a reason why this mock class got taken out of the `TableContext` class?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2182,6 +2190,79 @@ public boolean isFailure() {
...
+    public static final class TableEntriesDeltaRead implements Reply, WireCommand {
+        final WireCommandType type = WireCommandType.TABLE_ENTRIES_DELTA_READ;
+        final long requestId;
+        final String segment;
+        final TableEntries tableEntries;
+        final boolean shouldClear;
+        final boolean reachedEnd;
+        final long lastPosition;
+
+        // TODO: Will be implemented as apart of: https://github.com/pravega/pravega/issues/4677
```
andreipaduroiu, line 2238: I think you can remove this TODO now.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessorTest.java
```diff
@@ -985,6 +986,115 @@ public void testGetTableEntries() throws Exception {
...
+        order.verify(connection).send(new WireCommands.SegmentCreated(1, tableSegmentName));
+        verify(recorderMock).createTableSegment(eq(tableSegmentName), any());
+        processor.updateTableEntries(new WireCommands.UpdateTableEntries(2, tableSegmentName, "",
+                getTableEntries(asList(e1, e2, e3)), WireCommands.NULL_TABLE_SEGMENT_OFFSET));
+        verify(recorderMock).updateEntries(eq(tableSegmentName), eq(3), eq(false), any());
+
+        // 1. Now read the table entries where suggestedEntryCount is equal to number of entries in the Table Store.
+        processor.readTableEntriesDelta(new WireCommands.ReadTableEntriesDelta(3, tableSegmentName, "", 0, 3));
+        // Does not work without some delay.
+        TimeUnit.SECONDS.sleep(1);
```
co-jo, line 1024: Note.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -315,9 +322,40 @@ protected int getMaxCompactionSize() {
         return segment.append(new ByteArraySegment(s), null, timeout);
     }
 
-    private <T> CompletableFuture<AsyncIterator<IteratorItem<T>>> newIterator(@NonNull String segmentName, byte[] serializedState,
-                                                                              @NonNull Duration fetchTimeout,
-                                                                              @NonNull GetBucketReader<T> createBucketReader) {
+    @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
co-jo: The cast in the lambda expression on line 342.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/ContainerMock.java
co-jo: It doesn't add much value with respect to this PR, however I thought it could sufficiently stand alone and might come in handy for some other work down the line. Will revert it.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessorTest.java
andreipaduroiu: This is a classic example of some async operation that hasn't completed yet. Most likely even your 1s delay may not help in some cases if wherever you are running this is too slow. 

What happens here is that somewhere inside the execution of `readTableEntriesDelta` we execute some async operation without waiting for it to finish. However since that's how `PravegaRequestProcessor` works internally, there is little we can do there to help us in this case. To solve this problem we've created this `InlineExecutor` which is a single-thread executor that executes every task "inline" (i.e., awaits its completion before moving on). This is why all the other tests in this class work consistently well.

This means that somewhere in your code (test code or not), you are not using this executor or are doing some other "funny business" (pretty vague term, but it's mostly a "catch-all" scenario). Let's focus on the first case instead (not using `InlineExecutor`).

If we look at the `CompletableFuture` doc (and implementation) we learn the following things:
1. Any continuation callback method (`thenXYZ`) that does not have the `Async` suffix will be executed on the same thread where the previous task completed (and implicitly on the same thread pool).
2. Any continuation callback method that has the `Async` method will execute as follows :
    1. If you pass an Executor as the last argument to that method, the callback code will be transferred to that Executor.
    2. If you do not pass an Executor, then the `ForkJoinPool.commonPool()` will be used instead.

In case 1.2. above, if you have such a callback, your execution will jump out of InlineExecutor to a regular thread pool, thus the runtime will not await its execution to complete so this test will likely not work.

I did a quick check on the code path that this is executing (and compared it to the regular iterator) and found that you do not pass `this.executor` to this method: https://github.com/pravega/pravega/pull/4867/files#diff-3dfd10edcdb2751b3655ff97953c1612R330

There may be others downstream, but I'm willing to bet this is the cause of your issues.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -815,6 +821,62 @@ public void readTableEntries(WireCommands.ReadTableEntries readTableEntries) {
...
+                                } else {
+                                    result.add(entry.getKey().getKey(), new AbstractMap.SimpleImmutableEntry<>(k, v), k.size() + v.size());
+                                }
+                            }
+                            result.setState(state);
+                            // Update total read data.
+                            return true;
+                        }))
+                .thenAccept(v -> {
+                    connection.send(new WireCommands.TableEntriesDeltaRead(
```
andreipaduroiu, line 868: Can we add a `log.debug` here (just like the other iterator methods) to indicate that we are sending something back. Feel free to follow the message pattern used for the other iterators.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -815,6 +821,62 @@ public void readTableEntries(WireCommands.ReadTableEntries readTableEntries) {
...
+                            }
+                            TableEntry entry = e.getEntries().iterator().next();
+                            DeltaIteratorState state = DeltaIteratorState.deserialize(e.getState().array());
+                            // Store all TableEntries.
+                            val k = new WireCommands.TableKey(toByteBuf(entry.getKey().getKey()), entry.getKey().getVersion());
+                            val v = new WireCommands.TableValue(toByteBuf(entry.getValue()));
+                            if (state.isDeletionRecord()) {
+                                result.remove(entry.getKey().getKey(), k.size() + v.size());
+                            } else {
+                                Map.Entry<WireCommands.TableKey, WireCommands.TableValue> old = result.getItem(entry.getKey().getKey());
```
andreipaduroiu, line 855: This complex calculation could be simplified in the following way:
1. Just invoke `.add` on the `DeltaIteratorState`.
2. Inside `DeltaIteratorState.add`, when you invoke `put` on the map, that method will actually return the existing (replaced) value, or null if no such value existed. If you get such a value, then subtract its size from the size tracker.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -292,6 +294,12 @@ public void close() {
         return newIterator(segmentName, serializedState, fetchTimeout, TableBucketReader::entry);
     }
 
+    @Override
+    public CompletableFuture<AsyncIterator<IteratorItem<TableEntry>>> entryDeltaIterator(String segmentName, long fromPosition, Duration fetchTimeout) {
+        logRequest("entryIterator", segmentName);
```
andreipaduroiu, line 299: entryDeltaIterator

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/DeltaIteratorState.java
```diff
@@ -0,0 +1,123 @@
...
+import io.pravega.segmentstore.contracts.tables.IteratorState;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.SneakyThrows;
+import lombok.ToString;
+
+import java.io.IOException;
+
+/**
+ * Represents the state of a resumable iterator.
```
andreipaduroiu, line 27: "resumable delta iterator"


Let's differentiate between `IteratorState` and `DeltaIteratorState`. The Javadoc should explain what is different about this one.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/DeltaIteratorState.java
```diff
@@ -0,0 +1,123 @@
...
+    private final long fromPosition;
+    @Getter
+    private final boolean reachedEnd;
+    @Getter
+    private final boolean shouldClear;
+    @Getter
+    private final boolean deletionRecord;
+
+    /**
+     * Creates a new instance of the EntryIteratorState class.
```
andreipaduroiu, line 44: DeltaIteratorState

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/IteratorStateImpl.java
```diff
@@ -25,7 +26,7 @@
 /**
  * Represents the state of a resumable iterator.
  */
-class IteratorState {
+class IteratorStateImpl implements  IteratorState {
```
andreipaduroiu, line 29: extra space after `implements`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryDeltaIterator.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        return null;
+    }
+
+    private CompletableFuture<Void> fetchNextTableEntriesBatch() {
+        return toEntries(currentBatchOffset)
+                .thenAccept(entries -> {
+                    if (!entries.isEmpty()) {
+                        this.currentEntry = entries.iterator();
```
andreipaduroiu, line 113: You are assigning `currentEntry` outside of declared guards. This is not thread safe.

Doesn't Intellij highlight this as a warning? It should be enabled by default.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryDeltaIterator.java
```diff
@@ -0,0 +1,188 @@
...
+                boolean reachedEnd = currentOffset + entry.getHeader().getTotalLength() >= this.maxLength + startOffset;
+                // We must preserve deletions to accurately construct a delta.
+                BufferView value = entry.getValue() == null ? BufferView.empty() : entry.getValue();
+                currentOffset += entry.getHeader().getTotalLength();
+                entries.add(new AbstractMap.SimpleEntry<>(
+                        new DeltaIteratorState(currentOffset, reachedEnd, this.shouldClear, entry.getHeader().isDeletion()),
+                        TableEntry.versioned(entry.getKey(), value, entry.getVersion())));
+            }
+
+        } catch (BufferView.Reader.OutOfBoundsException ex) {
```
andreipaduroiu, line 152: I believe I know why you are silently ignoring this exception, but please add a comment explaining why.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableService.java
```diff
@@ -109,6 +109,13 @@ public TableService(SegmentContainerRegistry segmentContainerRegistry, SegmentTo
                 "get", segmentName, serializedState != null, fetchTimeout);
     }
 
+    @Override
+    public CompletableFuture<AsyncIterator<IteratorItem<TableEntry>>> entryDeltaIterator(String segmentName, long fromPosition, Duration fetchTimeout) {
+        return invokeExtension(segmentName,
+                e -> e.entryDeltaIterator(segmentName, fromPosition, fetchTimeout),
+                "get", segmentName, fromPosition, fetchTimeout);
```
andreipaduroiu, line 116: get -> entryDeltaIterator

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2238,31 +2238,31 @@ public boolean isFailure() {
...
         public static WireCommand readFrom(ByteBufInputStream in, int length) throws IOException {
             long requestId = in.readLong();
             String segment = in.readUTF();
             String delegationToken = in.readUTF();
-            long fromVersion = in.readLong();
+            long fromPosition = in.readLong();
             int suggestedEntryCount = in.readInt();
 
-            return new ReadTableEntriesDelta(requestId, segment, delegationToken, fromVersion, suggestedEntryCount);
+            return new ReadTableEntriesDelta(requestId, segment, delegationToken, fromPosition, suggestedEntryCount);
```
andreipaduroiu, line 2265: you need to add `.requireRelease()` before returning this object. See the changes that were made to `TableEntriesRead` ,for example.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
co-jo: Do you mean `DeltaIteratorResult`? From my understanding it is not guaranteed that entries are read monotonically, i.e we could receive some value later on in the iteration, with an earlier version. Assuming that is true, we'd have to check these conditions prior to the `put`.  Certainly I can place that logic within the `add` call, if losing the generic functionality is acceptable.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryDeltaIterator.java
co-jo: Nope, no warnings. Will look to enable it.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
co-jo: Which resources need releasing? I extended `TableEntriesReadDelta` to release its `entries` but it didn't seem like `ReadTableEntriesDelta` required it.

andreipaduroiu: My bad. This comment relates to line 2303. See below.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2299,14 +2298,14 @@ public static WireCommand readFrom(EnhancedByteBufInputStream in, int length) th
             TableEntries entries = TableEntries.readFrom(in, in.available());
             boolean shouldClear = in.readBoolean();
             boolean reachedEnd = in.readBoolean();
-            long lastVersion = in.readLong();
+            long lastPosition = in.readLong();
 
-            return new TableEntriesDeltaRead(requestId, segment, entries, shouldClear, reachedEnd, lastVersion);
+            return new TableEntriesDeltaRead(requestId, segment, entries, shouldClear, reachedEnd, lastPosition);
```
andreipaduroiu, line 2303: This is where you need to add it. See other cases in this class for examples how to do it.

If you want to know what, then follow through the `.release` call at line 2308 and see which `ByteBuf`s get released. Without invoking that (and requiring it), those buffers will stick around for a long time and lead to memory leaks).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
andreipaduroiu: I am talking about the HashMap internal to `DeltaIteratorResult`. That method is synchronized so you don't have to worry about concurrency. Actually you don't need to worry about concurrency at all in this case; we only need synchronized because we pass this between threads (in sequence).

If you use the HashMap.put while synchronized you are guaranteed to get the previous value assigned to that key. It will make your code simpler and (slightly) faster.

andreipaduroiu: Ok, nevermind. I completely missed the second part of the `if` condition, where you are comparing the versions. Then what you are doing is good.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryDeltaIterator.java
```diff
@@ -0,0 +1,193 @@
...
+    private Iterator<Map.Entry<DeltaIteratorState, TableEntry>> currentEntry;
+    @GuardedBy("this")
+    private long currentBatchOffset;
+
+    //endregion
+
+    //region AsyncIterator Implementation
+
+    @Override
+    public CompletableFuture<T> getNext() {
```
tkaitchuck, line 75: It says this is thread safe, but I am not clear on the semantics of this method being called in parallel. 
In particular `getNextEntry` could have two calls to `getNextEntryFromBatch` in parallel both get null and hence end up calling `fetchNextTableEntriesBatch` twice. When if those calls had been ordered it would have only resulted to one call to `fetchNextTableEntriesBatch`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -1001,5 +1067,45 @@ synchronized BufferView getContinuationToken() {
...
+    private static class DeltaIteratorResult<K, V, S> {
+        @Getter
+        @Setter
+        @GuardedBy("this")
+        private S state;
+        @GuardedBy("this")
+        private final Map<K, V> items = new HashMap<>();
+        @Getter
+        @GuardedBy("this")
+        private int sizeBytes;
```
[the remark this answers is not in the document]
co-jo: I'm not sure I understand the question. Its being used to bound the aggregate size of entries that can be returned.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableEntryDeltaIterator.java
co-jo: I see, good catch. It would seem the only conflict that can occur is that multiple calls to `getNextEntry` can return `null`, but otherwise concurrent calls won't point to the same non-null `currentEntry`. So I see two solutions: call `if (this.currentEntry != null) return CompletableFuture.completedFuture(null);` from within `fetchNextTableEntriesBatch` or make `getNextEntry` synchronized. I'm not sure the former provides any substantial improvement.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -457,13 +463,48 @@ protected int getMaxCompactionSize() {
                 });
     }
 
+    @SuppressWarnings("unchecked")
+    public <T> CompletableFuture<AsyncIterator<IteratorItem<T>>> newDeltaIterator(@NonNull String segmentName, long fromPosition, @NonNull Duration fetchTimeout) {
+        return this.segmentContainer
+                .forSegment(segmentName, fetchTimeout)
+                .thenComposeAsync(segment -> {
+                    SegmentProperties properties = segment.getInfo();
+                    long compactionOffset = properties.getAttributes().getOrDefault(TableAttributes.COMPACTION_OFFSET, 0L);
```
andreipaduroiu, line 472: It just occurred to me that this will not work on sorted table segments. It does not have to. 

Let's add a check here (it's an attribute stored in the SegmentProperties) whether this is a sorted table segment. If it is, we should be throwing/returning an UnsupportedOperationException explaining why.

Issue #2423, opened 2018-04-03
Configure default repoUri for mkdocs


vijikarthi:
**Problem description**
Flink connector project is making use of Pravega dependency as a git sub-module snapped at a specific commit version. The introduction of `mkdocs` plugin in Pravega is causing build issue in `flink-connector` project since the plugin is not able to resolve the project repo URI 

https://github.com/pravega/pravega/blob/master/build.gradle#L27

**Problem location**

The plugin appears to break when applied to a project that is checked out as a git sub-module (as is the case for Pravega in the connector repository).   This is due to a slight variation in how the `.git` directory works.

**Suggestions for an improvement**
The `mkdocs` plugin breaks when attempting to discover the origin uri of the source repository for which documentation is being generated.  This appears to be for informational purposes only.  By default the plugin inspects the git remotes to find `origin`, i.e. `https://github.com/pravega/pravega.git`.

A simple workaround is to explicitly configure the repo uri so that the faulty discovery logic isn't exercised.   Update `mkdocs.gradle` ([ref](https://github.com/pravega/pravega/blob/master/gradle/mkdocs.gradle#L17)) to include a  `repoUri` block that points to the Pravega source repository.  e.g. 
```
mkdocs {
    sourcesDir = "$rootDir/documentation/src"
    buildDir = "$rootDir/documentation/generated"
    updateSiteUrl = false
    publish {
            repoUri = "https://github.com/pravega/pravega.git"
    }
}
```

The underlying plugin code may be seen [here ](https://github.com/xvik/gradle-mkdocs-plugin/blob/master/src/main/groovy/ru/vyarus/gradle/plugin/mkdocs/MkdocsPlugin.groovy#L61) and [here](https://github.com/xvik/gradle-mkdocs-plugin/blob/master/src/main/groovy/ru/vyarus/gradle/plugin/mkdocs/MkdocsPlugin.groovy#L153).

Prabha-Veerubhotla:
https://github.com/pravega/pravega/pull/2424

vijikarthi:
My understanding is that `repo_url` is used by the plugin to provide a link to the project repository from the documentation site (pages).

PR #2169, opened 2017-12-11
Issue 1127: Adding gradle plugin to convert docs from markdown to html format

Prabha-Veerubhotla:
Signed-off-by: PrabhaVeerubhotla <vijayalakshmi.veerubhotla@emc.com>

**Change log description**
* Adds two plugins: `gradle-ssh-plugin`, `gradle-mkdocs-plugin`. 
* This adds gradle tasks to generate docs in html from with specified theme and to upload them to  `pravega.io` server.

 **Purpose of the change**
We have documents under docs/ folder in code repository on GitHub.  Every time there is a change on source docs, we need to manually build them using mkdocs to generate html versions and upload to http://pravega,io/docs

**What the code does**
The `gradle-mkdocs-plugin` convert docs from markdown(`md`) format to `html` format. This eliminates the explicit format conversion using tool `mkdocs`. The input(.md)  is taken from `/documentation/src/docs`  folder.  `documentatiom/src/mkdocs.yml` is used as config file. Check the output html docs in the folder `documentation/generated/<pravegaVersion>`.
The `Gradle SSH Plugin` is a Gradle plugin which provides SSH facilities such as command execution or file transfer for continuous delivery.  The docs from `documentation/generated/<version>/` are transferred to `var/www/pravega.io/docs`. 

Fixes #1127 

**How to verify it** 
Prerequisites: 
1.  Upgrade to gradle wrapper version 4.4.1  using `./gradlew wrapper --gradle-version=4.4.1 --
     distribution-type=bin`
2.  Python >= 2.7 and pip (preferrably latest version)
3.  Copy the pravegaio.pem(private key to connect to pravega.io server) to pravega root directory
4.  Run  `./gradlew clean :deploy -Pusername= <> -Phostname=<> `. (Provide the username and hostname to connect to `pravega.io` server)

PR #4871, opened 2020-06-15
Issue 4868: Enable mkdocs plugin in build scripts 

### LINKED ISSUE #4868: Enable mkdocs plugin in gradle build scripts
**Problem description**
Currently in order to generate a documentation for pravega.io website one needs to install all necessary dependencies in order to use `mkdocs` tool. This a bit odd and there should be a better way to handle this.  

**Problem location**
Build scripts 

Enable `mkdocs` plugin in gradle so it will take care of pulling all the dependencies

medvedevigorek:
**Change log description**  
Enables `mkdocs` gradle plugin to make it easier to generated the documentation as well as modify it. In addition also fixes most of the warnings by adding missing pages or fixing the broken links. 

**Purpose of the change**  
 Closes #4868 

**What the code does**  

1. Enables `mkdocs` plugin in `gradle.properties` and also bumps its version to 1.1 
2. Replaces deprecated `pages` configuration with `nav` 
3. Fixes broken link to `README` file in `Getting Started` page
4. Adds `Securing Pravega in Standalone Mode` under `Pravega Security` as the page was in there but never made available
5. Updates a link to Client javadoc location as it gets corrected every time when a new version of documentation is published.  
6. Removes configuration related to `publish` task as auto-publishing is not yet fully supported

**How to verify it**  
Manually verified that `mkdocs` tasks are now available as gradle tasks and that both `mkdocsBuild` as well as `mkdocsServe` work in both Windows and Linux environments.


shrids:
The changes look good on inspection; Couple of observations
- The latest plugin version is 2.0.1 (https://github.com/xvik/gradle-mkdocs-plugin/tags) We are not using it since it does not work with gradle 4.x
- I tried `gradle mkdocsBuild` with this changeset and it failed on my Linux box. (perhaps it is a setup issue on my laptop)
```
[python] /usr/bin/python3.6 -m mkdocs build -c -d /home/osboxes/code/pravega/pravega/documentation/generated/0.8.0-2582.f550e4a-SNAPSHOT
         /usr/bin/python3.6: No module named mkdocs
FAILURE: Build failed with an exception.
```

medvedevigorek:
That's a good catch @shrids! While it works on both of my machines (windows and linux) I gave it another try on a fresh linux box (deployed an ubuntu container and installed all missing dependencies including python+pip) and observed the same issue. 

What actually happens is `gradle-mkdocs-plugin` depends on `ru.vyarus:gradle-use-python-plugin` plugin and the latter works in getting `virtualenv` prepared but fails to execute other tasks in the created virtualenv. The actual fix was in both of them and the latest version of both should work fine however since we are locked with gradle 4.x there is no way we can upgrade them. at least not now. 

It used to work for me before because I installed all necessary dependencies in the system pythong env hence everything works when plugin invoked system python. 

I've been trying various options on getting `mkdocsBuild` tasks executed in the created virtualenv however so far no luck. Currently I am getting more and more towards idea to let the plugin to install all the dependencies in the user python installation and keep the idea of using the dedicated `virtualenv` till better times. 

Will update PRs in coming day once test it be a bit better.

medvedevigorek:
@shrids I've finally made the changes, there will be no virtualenv used at all, instead all the dependencies are installed in the user python directory.

gradle.properties
```diff
@@ -66,3 +66,6 @@ bookkeeperBaseTag=pravega/bookkeeper
 signing.keyId=
 # This will be defaulted to ~/.gnupg/secring.gpg if not provided as a command line property
 signing.secretKeyRingFile=
+
+enableMkdocs=true
```
shrids, line 70: Is there a reason why it was disabled in the older build?

medvedevigorek: It was enabled by default initially (see PR #2169) but then your question made me checking the history of some other files and it seems there was an issue with building Flink Connector (see Issue #2423).  @crazyzhou  is helping to check whether the issue is still present if `mkdocs`  plugin is enabled by default or not.

andreipaduroiu (dismissed):
At the request of @medvedevigorek , blocking this until we resolve downstream issues with the connectors.

derekm (approved):
I think this is read to merge. This PR is needed to collaborate on docs tech debt items.

RaulGracia (approved):

shrids (approved):

RaulGracia:
Checked with @crazyzhou that Flink connector removed the Pravega submodule in PR https://github.com/pravega/flink-connectors/pull/470.

Issue #3538, opened 2019-03-28
StreamSegmentContainerTests fails with unexpected exception


tkaitchuck:
**Problem description**
```
io.pravega.segmentstore.server.containers.StreamSegmentContainerTests > testWriteFenceOut FAILED
    java.lang.AssertionError: Original container did not reject an append operation after being fenced out. Exception thrown was of unexpected type: io.pravega.common.ObjectClosedException: Object 'BlockingDrainingQueue' has been closed and cannot be accessed anymore.
        at org.junit.Assert.fail(Assert.java:88)
        at io.pravega.test.common.AssertExtensions.assertSuppliedFutureThrows(AssertExtensions.java:118)
        at io.pravega.segmentstore.server.containers.StreamSegmentContainerTests.testWriteFenceOut(StreamSegmentContainerTests.java:1444)
```

**Problem location**
`StreamSegmentContainerTests.testWriteFenceOut`

**Suggestions for an improvement**
Add `ObjectClosedException` object closed exception to list of acceptable exceptions for the method to throw.

derekm:
@tkaitchuck @andreipaduroiu -- I saw a timeout failure on this test in a travis build just now:
```
io.pravega.segmentstore.server.containers.StreamSegmentContainerTests > testWriteFenceOut FAILED
    org.junit.runners.model.TestTimedOutException: test timed out after 100000 milliseconds
        at sun.misc.Unsafe.park(Native Method)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
        at java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
        at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
        at com.google.common.util.concurrent.Monitor.enterIf(Monitor.java:679)
        at com.google.common.util.concurrent.AbstractService.stopAsync(AbstractService.java:264)
        at io.pravega.common.concurrent.Services.stopAsync(Services.java:61)
[...]

PR #4872, opened 2020-06-15
Issue 4901: Fix not null check on target param in StreamHelpers#readAll

### LINKED ISSUE #4901: Fix & test parameter preconditions in StreamHelpers#readAll
**Problem description**
While reading through code, I eyeballed that StreamHelpers#readAll function argument preconditions were erroneously asserting not-null on the same parameter twice.

**Problem location**
https://github.com/pravega/pravega/blob/fec616b2d6b0fbf834cc7f59e2d898cdc5ac398f/common/src/main/java/io/pravega/common/io/StreamHelpers.java#L38

I've fixed the typo and implemented negative tests that exercise the preconditions. In doing so, I found a few small wrinkles in exception text which I've also fixed:
* maxLength precondition listed its name twice, and
* AssertExtensions#assertThrows with tester predicate misidentified failure reason.

derekm:
**Change log description**  
Not null precondition checks in `StreamHelpers#readAll` were testing the `stream` param twice.

**Purpose of the change**  
To fix a bug where the NPE for null `target` param would be thrown 1 line late with the wrong error message.
Fixes #4901

**What the code does**  
Makes second not null precondition check match its error message. Makes null `target` throw NPE with a better error message.

**How to verify it**  
Negative tests in `StreamHelpers` unit tests were added to confirm the correct exception is thrown when `target` param is left null. Negative tests were implemented for the other preconditions, which resulting in improving exception text for `maxLength` parameter and for `assertThrows` failures.

### DIFF
--- a/common/src/main/java/io/pravega/common/io/StreamHelpers.java
+++ b/common/src/main/java/io/pravega/common/io/StreamHelpers.java
@@ -37,7 +37,7 @@ public final class StreamHelpers {
         Preconditions.checkNotNull(stream, "stream");
         Preconditions.checkNotNull(target, "target");
         Preconditions.checkElementIndex(startOffset, target.length, "startOffset");
-        Exceptions.checkArgument(maxLength >= 0, "maxLength", "maxLength must be a non-negative number.");
+        Exceptions.checkArgument(maxLength >= 0, "maxLength", "must be a non-negative number.");
 
         int totalBytesRead = 0;
         while (totalBytesRead < maxLength) {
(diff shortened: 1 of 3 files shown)


derekm:
@andreipaduroiu -- I've updated this PR in an effort to comply with contributor guidelines, and I've added tests to verify the fix, which resulted in some additional minor cleanups.

This PR might seem tiny & pedantic, but it was good exercise in becoming more familiar with tests in Pravega and with contributor guidelines.

derekm:
Travis build failed with an intermittent failure on a unit test: https://github.com/pravega/pravega/issues/3538#issuecomment-649751781

I am unable to kick off a new travis build.

PR #4873, opened 2020-06-16
Issue #4924: Remove netty from the client

tkaitchuck:
**Change log description**  
Removes netty connections from the client

**Purpose of the change**  
The goal if for this to improve performance. 
Fixes: https://github.com/pravega/pravega/issues/4924
This does not yet provide a full solution to #4923 as there is still a dependency on ByteBuf.

**What the code does**  
This removes netty from the client. This should be fully compatible with existing logic.

**How to verify it**  
Manual verification required. System/integration/longevity tests should all be run. Especially with TLS enabled.


ravisharda:
@tkaitchuck Pravega standalone mode cluster is not bootstrapping properly with this change. So, apparently the changes are not working. 

I wanted to verify Standalone mode first, before verifying the distributed mode cluster.

RaulGracia:
@ravisharda you are right, Pavel has also tried this branch in a distributed mode and it didn't work neither. I also see that the branch does not contain the changes from `master` since end of April, so it would be great to do a rebase to compare the performance improvement of these changes considering everything that has been done in the meantime.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -27,19 +27,24 @@
  * synchronous writers. Otherwise the batch size is set to the amount of data that will be written in the next
  * {@link #MAX_BATCH_TIME_MILLIS} or half the server round trip time (whichever is less)
  */
-class AppendBatchSizeTrackerImpl implements AppendBatchSizeTracker {
-    private static final int MAX_BATCH_TIME_MILLIS = 100;
-
+public class AppendBatchSizeTrackerImpl implements AppendBatchSizeTracker {
```
RaulGracia, line 30: Please, revert this change as it may be inducing artifacts on the client latency and makes it difficult to compare with the `master` version. It would be great to have the batching-related changes on a separate PR.

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
```diff
@@ -0,0 +1,59 @@
...
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.pravega.common.MathHelpers;
+import io.pravega.shared.protocol.netty.WireCommands;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+class IoBuffer {
```
RaulGracia, line 20: Should this class be in package `common` along with all the other Byte-related classes?

client/src/main/java/io/pravega/client/segment/impl/SegmentInputStreamImpl.java
```diff
@@ -178,7 +178,9 @@ private void issueRequestIfNeeded() {
         //compute read length based on current offset up to which the events are read.
         int updatedReadLength = computeReadLength(offset + buffer.dataAvailable());
         if (!receivedEndOfSegment && !receivedTruncated && updatedReadLength > 0 && outstandingRequest == null) {
-            log.trace("Issuing read request for segment {} of {} bytes", getSegmentId(), updatedReadLength);
+            if (log.isTraceEnabled()) {
```
RaulGracia, line 181: Is this faster compared to what it was?

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
tkaitchuck: I don't think the changes here are function in nature. Please feel free to verify this.

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
```diff
@@ -0,0 +1,59 @@
...
+    
+    public ByteBuf getBuffOfSize(InputStream in, int size) throws IOException {
+        if (size > maxBufferSize) {
+            throw new IllegalArgumentException("Requested buffer size " + size + " is larger than maximum allowed"
+                    + maxBufferSize);
+        }
+        if (buffer == null) {
+            int bufferSize = MathHelpers.minMax(in.available(), size, maxBufferSize);
+            byte[] newBuffer = new byte[bufferSize];
+            int read = in.read(newBuffer);
```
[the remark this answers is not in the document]
tkaitchuck: Yes. It can. I believe this is handled. Notice line 46.

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
```diff
@@ -0,0 +1,59 @@
...
+            buffer = ByteBuffer.wrap(newBuffer, 0, read); 
+        } 
+
+        if (buffer.remaining() >= size) {
+            return sliceOut(size);
+        } else {
+            int firstSize = buffer.remaining();
+            ByteBuf first = sliceOut(firstSize);
+            assert buffer == null; //Should have been fully sliced out
+            byte[] remaining = new byte[size - firstSize];
```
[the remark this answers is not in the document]
tkaitchuck: Size is bounded by the check on line 35. firstSize is always a positive number, so remaining is always smaller than `maxBufferSize`

RaulGracia:
@tkaitchuck there is a problem compiling the tests that prevents the build from succeeding.

client/src/main/java/io/pravega/client/segment/impl/SegmentInputStreamImpl.java
tkaitchuck: It showed up in the flamegraph before, so yes.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,306 @@
...
+        private final ScheduledExecutorService thread;
+        private final AppendBatchSizeTracker batchSizeTracker;
+        private final AtomicBoolean stop = new AtomicBoolean(false);
+
+        public ConnectionReader(String name, InputStream in, ReplyProcessor callback,
+                                AppendBatchSizeTracker batchSizeTracker) {
+            this.name = name;
+            this.in = in;
+            this.callback = callback;
+            this.thread = ExecutorServiceHelpers.newScheduledThreadPool(1, "Reading from " + name);
```
andreipaduroiu, line 98: We should find a better thread pool name.

andreipaduroiu: And name it "executor" or "threadPool" since it's not really a thread. Or If you really want a single thread, then just create a thread, not a pool.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,306 @@
...
+            }
+
+            ByteBuf payload = buffer.getBuffOfSize(in, length);
+            
+            return type.readFrom(new EnhancedByteBufInputStream(payload), length);
+        }
+        
+        public void stop() {
+            stop.set(true);
+            thread.shutdown();
```
andreipaduroiu, line 157: `shutdownNow` or `ExecutorServiceHelpers.shutdown`; otherwise there's no guarantee that its threads are properly interrupted.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,306 @@
...
+                InputStream inputStream = socket.getInputStream();
+                AppendBatchSizeTrackerImpl batchSizeTracker = new AppendBatchSizeTrackerImpl();
+                ConnectionReader reader = new ConnectionReader(location.toString(), inputStream, callback, batchSizeTracker);
+                reader.start();
+                CommandEncoder encoder = new CommandEncoder(l -> batchSizeTracker, null, socket.getOutputStream());
+                return new TcpClientConnection(socket, encoder, reader, location, onClose, executor);
+            } catch (IOException e) {
+                try {
+                    onClose.run();
+                    socket.close();
```
andreipaduroiu, line 186: Should we ensure that the socket is closed before invoking onClose (as it may itself throw an exception)?

andreipaduroiu: I would close the socket on any exception that comes out of this code, not just IOException.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,306 @@
...
+            result = s;
+
+        } else {
+            result = new Socket();
+        }
+        result.setSendBufferSize(TCP_BUFFER_SIZE);
+        result.setReceiveBufferSize(TCP_BUFFER_SIZE);
+        result.setTcpNoDelay(true);
+        result.connect(new InetSocketAddress(location.getEndpoint(), location.getPort()));
+        return result;
```
andreipaduroiu, line 240: You should put a try-catch around all this code and ensure the socket is closed if it has been created and some exception happens after that and until you return.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,306 @@
...
+
+    @Override
+    public void send(WireCommand cmd) throws ConnectionFailedException {
+        if (closed.get()) {
+            throw new ConnectionFailedException("Connection is closed");
+        }
+        try {
+            encoder.write(cmd);
+        } catch (IOException e) {
+            log.warn("Error writing to connection");
```
andreipaduroiu, line 251: We should at least log the message of this exception (i.e. `e.toString()`) so that we know a bit more about why it failed.

Below too.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,306 @@
...
+    @Override
+    public void close() {
+        if (closed.compareAndSet(false, true)) {
+            reader.stop();
+            timeoutFuture.cancel(false);
+            try {
+                if (onClose != null) {
+                    onClose.run();
+                }
+                socket.close();
```
andreipaduroiu, line 280: Close before invoking onClose?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/Append.java
```diff
@@ -18,8 +18,8 @@
 @Data
 @RequiredArgsConstructor
 public class Append implements Request, Comparable<Append> {
-    final String segment;
-    final UUID writerId;
+    public final String segment;
```
andreipaduroiu, line 21: Why did you make these public when there are getters for them?

client/src/test/java/io/pravega/client/connection/impl/ClientConnectionTest.java
```diff
@@ -0,0 +1,148 @@
...
+        wireCommand = messages.take();
+        assertEquals(WireCommandType.SETUP_APPEND, wireCommand.getType());
+        assertNull(messages.poll());
+        clientConnection.send(new Append("segment", new UUID(1, 2), 1, new Event(Unpooled.EMPTY_BUFFER), 2));
+        wireCommand = messages.take();
+        assertEquals(WireCommandType.APPEND_BLOCK, wireCommand.getType());
+        assertFalse(processor.falure.get());
+    }
+
+    //    @Test
```
andreipaduroiu, line 80: Why is this commented out?

client/src/test/java/io/pravega/client/connection/impl/ConnectionFactoryImplTest.java
```diff
@@ -0,0 +1,234 @@
...
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.connection.impl;
+//
```
andreipaduroiu, line 11: Should we delete this file?

client/src/test/java/io/pravega/client/connection/impl/ConnectionPoolImplTest.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.connection.impl;
+//
+//import io.netty.handler.ssl.SslContext;
```
andreipaduroiu, line 12: Should we uncomment this?

client/src/test/java/io/pravega/client/connection/impl/ConnectionPoolingTest.java
```diff
@@ -0,0 +1,306 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.connection.impl;
+//
+//import io.netty.bootstrap.ServerBootstrap;
```
andreipaduroiu, line 12: Uncomment

client/src/test/java/io/pravega/client/connection/impl/FlowHandlerTest.java
```diff
@@ -0,0 +1,436 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.connection.impl;
+
+//
```
andreipaduroiu, line 12: Uncomment

client/src/test/java/io/pravega/client/connection/impl/RawClientTest.java
```diff
@@ -0,0 +1,142 @@
...
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.connection.impl;
+
+//import io.netty.buffer.Unpooled;
```
andreipaduroiu, line 11: uncomment

client/src/test/java/io/pravega/client/connection/impl/SecureConnectionFactoryImplTest.java
```diff
@@ -7,16 +7,16 @@
...
-public class SecureConnectionFactoryImplTest extends ConnectionFactoryImplTest {
-
-    @Override
-    @Before
-    public void setUp() throws Exception {
-        this.ssl = true;
-        super.setUp();
-    }
-}
+//import org.junit.Before;
```
andreipaduroiu, line 12: Uncomment

client/src/test/java/io/pravega/client/security/auth/JwtTestUtils.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.security.auth;
+
+import com.google.gson.Gson;
+import io.pravega.test.common.JwtBody;
+import java.util.Base64;
+
+public class JwtTestUtils {
```
andreipaduroiu, line 16: There is a class by this name in `io.pravega.common`. May I suggest merging these methods in that one?

common/src/main/java/io/pravega/common/util/TypedProperties.java
```diff
@@ -95,6 +95,18 @@ public long getLong(Property<Long> property) throws ConfigurationException {
...
 
+    /**
+     * Gets the value of a Double property.
+     *
+     * @param property The Property to get.
+     * @return The property value or default value, if no such is defined in the base Properties.
+     * @throws ConfigurationException When the given property name does not exist within the current component and the property
+     *                                does not have a default value set, or when the property cannot be parsed as a Double.
+     */
+    public double getDouble(Property<Double> property) throws ConfigurationException {
```
andreipaduroiu, line 106: Need unit tests for this one.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -27,19 +27,26 @@
...
+    
     private final Supplier<Long> clock;
     private final AtomicLong lastAppendNumber;
     private final AtomicLong lastAppendTime;
     private final AtomicLong lastAckNumber;
-    private final ExponentialMovingAverage eventSize = new ExponentialMovingAverage(1024, 0.1, true);
-    private final ExponentialMovingAverage millisBetweenAppends = new ExponentialMovingAverage(10, 0.1, false);
-    private final ExponentialMovingAverage appendsOutstanding = new ExponentialMovingAverage(2, 0.05, false);
+    private final ExponentialMovingAverage eventSize = new ExponentialMovingAverage(1024, 0.01, true);
+    private final ExponentialMovingAverage nanosBetweenAppends = new ExponentialMovingAverage(10 * NANOS_PER_MILLI, 0.001, true);
```
eolivelli, line 45: This switch from _millis_ to _nanos_ seems unrelated to the removal of Netty.
It is required for this patch to work ?

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -73,15 +79,28 @@ public int getAppendBlockSize() {
...
         return MAX_BATCH_TIME_MILLIS;
     }
+    
+    private static int initializeConstant(String name, int defaultValue) {
+        String val = System.getenv(name);
+        if (val != null) {
+            try {
+                return Integer.parseInt(val);
+            } catch (NumberFormatException e) {
+                //Use default
```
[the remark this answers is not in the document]
eolivelli: IMHO I would throw an exception, otherwise the user has no way to check that the configured value is in use.
I think that not everyone checks the log, you usually check logs only when you have a problem.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -73,15 +79,28 @@ public int getAppendBlockSize() {
...
+        int size = (int) (appendsInBatch * eventSize.getCurrentValue()) + BASE_SIZE;
+        return MathHelpers.minMax(size, 0, MAX_BATCH_SIZE);
     }
 
     @Override
     public int getBatchTimeout() {
         return MAX_BATCH_TIME_MILLIS;
     }
+    
+    private static int initializeConstant(String name, int defaultValue) {
```
eolivelli, line 95: is it better to move this function to a common utility class ?

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -73,15 +79,28 @@ public int getAppendBlockSize() {
...
+        String val = System.getenv(name);
```
eolivelli, line 96: is is a common practice in Pravega to use System env for the client ?

Here we are in the client, we don't know much about the running environment of the client application.

Additionally using ENV may be seen as a security issue, as the application is going to use parameters configured outside the configuration files of the application itself, without even noticing it.

We should at least log a line telling that we picked up a value from system ENV

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
```diff
@@ -0,0 +1,207 @@
...
+        Preconditions.checkNotNull(rp, "ReplyProcessor");
+        Exceptions.checkNotClosed(closed.get(), this);
+
+        // create a new connection.
+        CompletableFuture<FlowHandler> handler = establishConnection(location);
+        Connection connection = new Connection(location, handler);
+        return connection.getFlowHandler().thenApply(h -> h.createConnectionWithFlowDisabled(rp));
+    }
+
+    private boolean isUnused(Connection connection) {
```
eolivelli, line 147: nit: static ?

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
```diff
@@ -0,0 +1,207 @@
...
+     * Used only for testing.
+     */
+    @VisibleForTesting
+    @Synchronized
+    public void pruneUnusedConnections() {
+        for (List<Connection> connections : connectionMap.values()) {
+            for (Iterator<Connection> iterator = connections.iterator(); iterator.hasNext();) {
+                Connection connection = iterator.next();
+                if (isUnused(connection)) {
+                    connection.getFlowHandler().join().close();
```
eolivelli, line 161: should we log a line ?
this .join() may take time to complete, if there is something stuck.

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+class IoBuffer {
+    private int maxBufferSize = WireCommands.MAX_WIRECOMMAND_SIZE;
+    private ByteBuffer buffer = null;
+    
+    private ByteBuf sliceOut(int size) {
+        ByteBuf result = Unpooled.wrappedBuffer(buffer.array(), buffer.arrayOffset() + buffer.position(), size);
```
eolivelli, line 26: Are we assuming that this is only a Heap ByteBuffer ?
`buffer.array()` won't work with a Direct Buffer

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
eolivelli: shall we wait for the threadPool to terminate execution ?

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,306 @@
...
+            SSLContext tlsContext = SSLContext.getInstance("TLS");
+            tlsContext.init(null,
+                    trustMgrFactory != null ? trustMgrFactory.getTrustManagers() : null,
+                    null);
+
+            SSLSocket s = (SSLSocket) tlsContext.getSocketFactory().createSocket();
+            // SSLSocket does not perform hostname verification by default. So, we must explicitly enable it.
+            if (clientConfig.isValidateHostName()) {
+                SSLParameters tlsParams = new SSLParameters();
+                tlsParams.setEndpointIdentificationAlgorithm("HTTPS");
```
eolivelli, line 228: HTTPS ?

common/src/main/java/io/pravega/common/util/CertificateUtils.java
```diff
@@ -0,0 +1,111 @@
...
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Collection;
+
+public class CertificateUtils {
```
eolivelli, line 28: Are we using lombok @UtilityClass  ?
https://projectlombok.org/features/experimental/UtilityClass

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
andreipaduroiu: These are "hardcoded" constants which we do not want users to set (we will not document them). The point of making them ENV vars is because we want to try out various combinations in our perf tests without having to do a full rebuild.

eolivelli: @andreipaduroiu 
What about adding a system property to enable this behaviour ?
`-Dpravega.allowreadconstantsfromenv=true`

This way we will protect clients from running into (security) issues.
System properties are not read automatically from the ENV or by the JVM, so you have to explicitly opt-in to this way of configuring these parameters.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
ravisharda: Yeah, it does look strange as it is connecting to TCP service exposed by Segment Store, and not to an HTTP (application-layer) server. 

However, note that SSLSocket does not perform TLS hostname verification by default and this is a common way of turning it on regardless of the application protocol. We were using the same mechanism elsewhere too: 

https://github.com/pravega/pravega/blob/c7ac009970787df2633163644691aa3d72fc1a4e/client/src/main/java/io/pravega/client/netty/impl/ConnectionPoolImpl.java#L249

So, this code is alright.

eolivelli: Thanks for the explanation.
What about adding a little comment?
This way who will read this code won't be worried

ravisharda: I've added the comment in the latest commit I just pushed.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
shrids: a system property approached is used here. ref: https://github.com/pravega/pravega/pull/4873/files#diff-642bfb8d9badad7de4a71c8d06a9a9baR67 io.pravega.client.connection.impl.SocketConnectionFactoryImpl#getThreadPoolSize

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: Connect is the very last line of this block, if it fails anywhere prior to that there is nothing to close, if it fails on that it never opened.

bindings/src/test/java/io/pravega/storage/extendeds3/S3ProxyImpl.java
```diff
@@ -177,7 +177,7 @@ public void deleteObject(String bucketName, String key) {
 
     @Override
     public DeleteObjectsResult deleteObjects(DeleteObjectsRequest request) {
-        request.getDeleteObjects().getKeys().forEach(key -> aclMap.remove(key));
+        request.getDeleteObjects().getKeys().forEach(key -> aclMap.remove(key.getKey()));
```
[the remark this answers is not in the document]
tkaitchuck: It is.

client/src/main/java/io/pravega/client/admin/impl/ReaderGroupManagerImpl.java
```diff
@@ -57,22 +59,22 @@
...
-    private final ConnectionFactory connectionFactory;
+    private final ConnectionPool connectionPool;
 
     public ReaderGroupManagerImpl(String scope, ClientConfig config, ConnectionFactory connectionFactory) {
         this.scope = scope;
         this.controller = new ControllerImpl(ControllerImplConfig.builder().clientConfig(config).build(),
                 connectionFactory.getInternalExecutor());
 
-        this.connectionFactory = connectionFactory;
+        this.connectionPool = new ConnectionPoolImpl(config, connectionFactory);
```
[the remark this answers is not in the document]
tkaitchuck: Fixed

client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
```diff
@@ -164,8 +164,8 @@ public void close() {
         if (this.executor != null) {
             ExecutorServiceHelpers.shutdown(this.executor);
         }
-        if (this.connectionFactory != null) {
-            this.connectionFactory.close();
+        if (this.connectionPool != null) {
```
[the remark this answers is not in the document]
tkaitchuck: It is being set, but it's not actually null checked in the constructor, so technically it could be null. (In practice it only ever is in one test)

client/src/main/java/io/pravega/client/byteStream/impl/ByteStreamClientImpl.java
```diff
@@ -89,6 +77,6 @@ public ByteStreamWriter createByteStreamWriter(String streamName) {
     @Override
     public void close() {
         controller.close();
-        connectionFactory.close();
+        connectionPool.close();
```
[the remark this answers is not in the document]
tkaitchuck: Looking at the constructor and how it it called, for this alone with all of the arguments have their ownership transferred to this class. So it is consistent with both the prior behavior and the behavior of the other fields. I'll add a javadoc.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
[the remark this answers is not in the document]
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -0,0 +1,419 @@
...
+    private final ByteBuf buffer = Unpooled.buffer(1024 * 1024);
+
+    @RequiredArgsConstructor
+    private final class Session {
+        private static final int MAX_EVENTS = 500;
+        private static final int MAX_DATA_SIZE = 1024 * 1024; // 1MB
+        private final UUID id;
+        private final long requestId;
+        private final List<ByteBuf> pendingList = new ArrayList<>();
+        private int  pendingBytes = 0;
```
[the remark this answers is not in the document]
tkaitchuck: DOne

client/src/main/java/io/pravega/client/connection/impl/ConnectionFactory.java
```diff
@@ -30,23 +30,9 @@
...
-    /**
-     * Get the internal executor which is used by the client.
-     * @return A ScheduledExecutorService.
-     */
-    ScheduledExecutorService getInternalExecutor();
-
     @Override
     void close();
 
+    ScheduledExecutorService getInternalExecutor();
```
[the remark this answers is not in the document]
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+        }
+    }
+    
+    private final ClientConfig clientConfig;
+    private final MetricNotifier metricNotifier;
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+    @GuardedBy("$lock")
+    private final Map<PravegaNodeUri, List<Connection>> connectionMap = new HashMap<>();
+    private ConnectionFactory connectionFactory;
```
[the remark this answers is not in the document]
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
```diff
@@ -0,0 +1,67 @@
...
+import io.netty.buffer.Unpooled;
+import io.pravega.common.MathHelpers;
+import io.pravega.shared.protocol.netty.WireCommands;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+class IoBuffer {
+    private int maxBufferSize = WireCommands.MAX_WIRECOMMAND_SIZE;
```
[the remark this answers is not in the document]
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: Done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/Append.java
tkaitchuck: Fixed. (Automated tool did this)

client/src/test/java/io/pravega/client/connection/impl/ClientConnectionTest.java
tkaitchuck: Fixed

client/src/test/java/io/pravega/client/connection/impl/ConnectionFactoryImplTest.java
tkaitchuck: Fixed

client/src/test/java/io/pravega/client/connection/impl/ConnectionPoolingTest.java
tkaitchuck: Done

client/src/test/java/io/pravega/client/connection/impl/ConnectionPoolImplTest.java
tkaitchuck: DOne

client/src/test/java/io/pravega/client/connection/impl/RawClientTest.java
tkaitchuck: Done

client/src/test/java/io/pravega/client/connection/impl/SecureConnectionFactoryImplTest.java
tkaitchuck: Done

client/src/test/java/io/pravega/client/security/auth/JwtTestUtils.java
tkaitchuck: Deleted this one

common/src/main/java/io/pravega/common/util/TypedProperties.java
tkaitchuck: Added

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
tkaitchuck: Nanos vs millis is not requred, but it did give bad results which interfered with testing. (It turns out it's fairly common for the difference in time in millis to be "0".)

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
tkaitchuck: done

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
tkaitchuck: If the `isUnused` function first checks `Futures.isSuccessful` so if it is considered unused the `join` should be instant.

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
tkaitchuck: The buffer is allocated locally in this class by wrapping an array. (And all accesses to it are private) So there is no way for it to be a heap byteBuffer.

common/src/main/java/io/pravega/common/util/CertificateUtils.java
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -73,15 +79,28 @@ public int getAppendBlockSize() {
...
         }
-        double appendsInMaxBatch = Math.max(1.0, MAX_BATCH_TIME_MILLIS / millisBetweenAppends.getCurrentValue());
-        double targetAppendsOutstanding = MathHelpers.minMax(appendsOutstanding.getCurrentValue() * 0.5, 1.0,
-                                                             appendsInMaxBatch);
-        return (int) MathHelpers.minMax((long) (targetAppendsOutstanding * eventSize.getCurrentValue()), 0,
-                                        MAX_BATCH_SIZE);
+        double nanosPerAppend = nanosBetweenAppends.getCurrentValue();
+        double appendsInMaxBatchTime = (MAX_BATCH_TIME_MILLIS * NANOS_PER_MILLI) / nanosPerAppend;
+        double appendsInTime = Math.max(1.0, BASE_TIME_NANOS / nanosPerAppend);
+        double appendsInBatch = MathHelpers.minMax(appendsOutstanding.getCurrentValue() * OUTSTANDING_FRACTION, appendsInTime, appendsInMaxBatchTime);
```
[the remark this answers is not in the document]
tkaitchuck: Fixed

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
```diff
@@ -0,0 +1,67 @@
...
+    private ByteBuf sliceOut(int size) {
+        ByteBuf result = Unpooled.wrappedBuffer(buffer.array(), buffer.arrayOffset() + buffer.position(), size);
+        buffer.position(buffer.position() + size);
+        if (!buffer.hasRemaining()) {
+            buffer = null;
+        }
+        return result;
+    }
+    
+    public ByteBuf getBuffOfSize(InputStream in, int size) throws IOException {
```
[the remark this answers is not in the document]
tkaitchuck: Added a comment.

> ByteBuf returned can have a reference to two byte-arrays (at max) where the size of each byte-array can be at max 16MB

Each can be at most 8MB. In practice however one is going to be small and the other big. Because they are allocated according to what is available (which is bounded by `TcpClientConnection.TCP_BUFFER_SIZE = 256kb` ) and then what is requested. Because the request is bounded by 8mb this means that the 'tail' of the first buffer plus the whole of the second buffer is <= 8mb. 

> Both the byte arrays will be freed only when the ByteBuf.release is invoked. And in case of small "size" params all the byteBufs should invoke release before the byte arrays are freed.

It's actually different depending on how much data is read. If the amount read out is smaller than the size of the buffer a view of it is returned and the array is still held by this class. This means that even once release is called the buffer will not be freed. (Which is good because it still has unread data)

In the event that the read is larger than the buffer the whole of the buffer is passed into the returned buffer and the IoBuffer class does not hold onto it. Similarly the remaining bytes are passed in the returned buffer and not held elsewhere. In this case because the only reference to the data is inside the ByteBuf, it actually does not matter if `free` is called or not, because once the ByteBuf is dropped there will be no references to the data anywhere and it can be garbage collected normally.

So while ByteBuf is designed to let you pool and re-use memory by using reference counting, this code is not actually taking advantage of that feature and is just relying on normal garbage collection. If that were to change in the future, the code would still be correct because the code that receives the ByteBufs still calls release (according to our unit tests) so it would be OK to add such functionality.

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        /**
+         * Queue the Append data to Session' list.
+         *
+         * @param data  data bytes.
+         * @param out   Network channel buffer.
+         */
+        private void write(ByteBuf data) {
+            pendingWrites.putIfAbsent(id, this);
```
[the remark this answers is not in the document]
tkaitchuck: I don't want to special case that too much. It is not outlawed by the wireprotocol to append zero bytes. (I don't have a reason we should, but we did and the wire command encoding layer is 'smart' and optimizes out the write it might be confusing)

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: Removed the pool

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: Changed

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
tkaitchuck: Done

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
tkaitchuck: @RaulGracia Would you prefer prefer these to: a) stay as-is. b) change into system properties
or are we done with testing these and we can go back to hardcoding them?

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -27,19 +30,31 @@
  * synchronous writers. Otherwise the batch size is set to the amount of data that will be written in the next
  * {@link #MAX_BATCH_TIME_MILLIS} or half the server round trip time (whichever is less)
  */
-class AppendBatchSizeTrackerImpl implements AppendBatchSizeTracker {
-    private static final int MAX_BATCH_TIME_MILLIS = 100;
-
+public class AppendBatchSizeTrackerImpl implements AppendBatchSizeTracker {
```
RaulGracia, line 33: I have done a new iteration on this branch. The results are much better than before. The only point that needs to be addressed is for small events and 1 segment. In that scenario, for throughput rates between 5K and 300K events per second, the write latency seems few milliseconds higher compared to the PoC. Note that for throughput rates lower and higher than this interval, this branch has a performance equal or better compared to the PoC, which may indicate that this is related to the changes in client batching and not related to any other performance bottleneck.

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -0,0 +1,419 @@
...
+         * @param data              Remaining data
+         */
+        private void flushToBuffer(int sizeOfWholeEvents, ByteBuf data) {
+            writeMessage(new AppendBlockEnd(id, sizeOfWholeEvents, data, eventCount, lastEventNumber, requestId), buffer);
+            eventCount = 0;
+        }
+    }
+    
+    /**
+     * Write/flush Multi session's buffered data to network channel buffer by iterating over all the pending sessions .
```
RaulGracia, line 159: Nit: `Multi` -> `multi`. Extra space at the end.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,320 @@
...
+                } catch (Exception e) {
+                    log.warn("Error processing data from from server " + name, e);
+                    stop();
+                }
+            }
+        }
+
+        @VisibleForTesting
+        static WireCommand readCommand(InputStream in, IoBuffer buffer) throws IOException {
+            ByteBuf header = buffer.getBuffOfSize(in, 8);
```
eolivelli, line 141: here we can use ByteBuf and recycle it.
Using the ByteBufAllocator is the simpler way.
it will make code simpler, without the need of the IoBuffer class

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,320 @@
...
+            if (type == null) {
+                throw new InvalidMessageException("Unknown wire command: " + t);
+            }
+
+            int length = header.getInt(4);
+            if (length < 0 || length > WireCommands.MAX_WIRECOMMAND_SIZE) {
+                throw new InvalidMessageException("Event of invalid length: " + length);
+            }
+
+            ByteBuf payload = buffer.getBuffOfSize(in, length);
```
eolivelli, line 154: the same here.
we can ask Netty to take the buffer from its pool (as we already know the length we also won't fall into buffer reallocations).
Netty will be able to manage direct memory for us.
and EnhancedByteBufInputStream will call release on "close()" so Netty will be able to recycle the buffer/memory.

It would be nice to make EnhancedByteBufInputStream recyclable in order to save another object allocation at every command unmashalling.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
eolivelli: It looks like that if `onClose` throws a RuntimeException we are still not calling `close`

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: My benchmarks showed that doing that was slower.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: I don't want to add further dependency on netty functionality here given that I am trying to remove it all together: https://github.com/pravega/pravega/issues/4923

It may make sense to add a dependency on the same buffer manager we use on the server side which we benchmarked as being faster than netty's version. That change however is large enough in it's own right that I don't want to lump it into this PR which is already quite large. I agree the above is not perfect it shows large perf improvements over the current master.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
eolivelli: Interesting.
Maybe the problem for netty is that we passing thru a generic InputStream and probably we can't leverage zero copy features.
If we had access to the SocketChannel maybe we could skip the copies to the heap.

Btw it is something we can do in some follow up work.

I am fine with this approach currently

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
eolivelli: Agreed

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -0,0 +1,419 @@
...
+            if (session.isFree()) {
+                session.record(append);
+                startAppend(blockSizeSupplier, append);
+                continueAppend(data);
+                if (bytesLeftInBlock == 0) {
+                    completeAppend(null);
+                    flushBuffer();
+                }
+            } else {
+                session.record(append);
```
tkaitchuck, line 212: I'm not sure this line is actually reachable. 
See discussion here: https://github.com/pravega/pravega/pull/4121#discussion_r330332792
from when it was added.

client/src/main/java/io/pravega/client/connection/impl/IoBuffer.java
```diff
@@ -0,0 +1,59 @@
...
+
+        if (buffer.remaining() >= size) {
+            return sliceOut(size);
+        } else {
+            int firstSize = buffer.remaining();
+            ByteBuf first = sliceOut(firstSize);
+            assert buffer == null; //Should have been fully sliced out
+            byte[] remaining = new byte[size - firstSize];
+            for (int offset = 0; offset < remaining.length;) {
+                offset += in.read(remaining, offset, remaining.length - offset);
```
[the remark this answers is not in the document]
tkaitchuck: Unfortunately timeouts are not available on InputStream. I have filed: https://github.com/pravega/pravega/issues/4943
Once that is fixed we should be able to add one easily.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: @eolivelli I think you need to refresh. close is now called first.

client/src/main/java/io/pravega/client/admin/impl/ReaderGroupManagerImpl.java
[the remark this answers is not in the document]
andreipaduroiu: I don't see any changes in `close`. Am I missing something?

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
```diff
@@ -157,96 +160,18 @@ public AppendBatchSizeTracker getAppendBatchSizeTracker(final long requestID) {
...
-        super.channelUnregistered(ctx);
-    }
-
-    @Override
-    public void channelRead(ChannelHandlerContext ctx, Object msg) {
-        Reply cmd = (Reply) msg;
-        log.debug(connectionName + " processing reply {} with flow {}", cmd, Flow.from(cmd.getRequestId()));
-        recentMessage.set(true);
+    public void process(Reply cmd) {
+        log.debug("{} processing reply {} with flow {}", location, cmd, Flow.from(cmd.getRequestId()));
```
andreipaduroiu, line 173: What does `Flow.from` do? Most of the time we won't have `debug` logs enabled, but this will be evaluated each time. Is this something we could optimize by checking `log.isDebugEnabled()` beforehand?

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
andreipaduroiu: I still see 
```
                    onClose.run();
                    socket.close();
```

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
tkaitchuck: It is not expensive, but I added the check.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: My mistake. I was looking in the wrong place.
fixed.

client/src/main/java/io/pravega/client/admin/impl/ReaderGroupManagerImpl.java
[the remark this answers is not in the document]
tkaitchuck: I changed it so that the clientFactory owns the pool and hence is responsible for it. 
I have made that more explicit by dropping the member variable.

RaulGracia:
@tkaitchuck some comments on the state of this PR:
- Can you please fix the DCO check? It looks like it is not happy.
- The build is not passing, and there are failures like the last one that could be related to changed in this PR:
```
io.pravega.client.connection.impl.ClientConnectionTest > testSendAsync FAILED
    java.util.concurrent.TimeoutException: Expected value: 2 observed: 5
        at io.pravega.test.common.AssertExtensions.assertEventuallyEquals(AssertExtensions.java:70)
        at io.pravega.test.common.AssertExtensions.assertEventuallyEquals(AssertExtensions.java:47)
        at io.pravega.client.connection.impl.ClientConnectionTest.testSendAsync(ClientConnectionTest.java:175)
```

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,320 @@
...
+        @Override
+        public void run() {
+            token.set(encoder.batchTimeout(token.get()));
+        }    
+    }
+
+    public static CompletableFuture<TcpClientConnection> connect(PravegaNodeUri location, ClientConfig clientConfig, ReplyProcessor callback,
+                                              ScheduledExecutorService executor, Runnable onClose) {
+        return CompletableFuture.supplyAsync(() -> {
+            Socket socket = createClientSocket(location, clientConfig);
```
shrids, line 183: This can throw an exception during socket.connect(...) causing the thread to silently fail.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,320 @@
...
+
+    @Override
+    public void send(Append append) throws ConnectionFailedException {
+        if (closed.get()) {
+            throw new ConnectionFailedException("Connection is closed");
+        }
+        try {
+            encoder.write(append);
+        } catch (IOException e) {
+            log.warn("Error writing to connection", e.toString());
```
shrids, line 279: nit: { } in the log .warn?

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,320 @@
...
+
+    @Override
+    public void send(WireCommand cmd) throws ConnectionFailedException {
+        if (closed.get()) {
+            throw new ConnectionFailedException("Connection is closed");
+        }
+        try {
+            encoder.write(cmd);
+        } catch (IOException e) {
+            log.warn("Error writing to connection", e.toString());
```
shrids, line 265: nit: { } in the log statement ?

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -0,0 +1,320 @@
...
+
+    @Override
+    public void sendAsync(List<Append> appends, CompletedCallback callback) {
+        try {
+            for (Append append : appends) {
+                encoder.write(append);
+            }
+            callback.complete(null);
+        } catch (IOException e) {
+            log.warn("Error writing to connection", e.toString());
```
shrids, line 309: nit: {} log .

tkaitchuck:
@RaulGracia There is no way to fix the DCO check now except to set it to pass. 
I tried manually re-doing all 186 commits since the one with the missing signoff here: https://github.com/pravega/pravega/pull/4945
but as you can see it now makes DCO look at master because of the merges which means it insists on re-writing master's history to resolve it, and conflicts on already resolved conflicts with master. So we will need to introduce code changes in a force push, which I am unwilling to do.
So I can only set it to pass to override.

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
```diff
@@ -0,0 +1,207 @@
...
+        return FlowHandler.openConnection(location, clientConfig, metricNotifier, connectionFactory);
+    }
+
+    @Override
+    public void close() {
+        log.info("Shutting down connection pool");
+        if (closed.compareAndSet(false, true)) {
+            metricNotifier.close();
+            connectionFactory.close();
+            for (List<Connection> connections : connectionMap.values()) {
```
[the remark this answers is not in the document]
tkaitchuck: I am not sure I understand the question. 
getClientConnection() and close() are both synchronized so they can't be called in parallel, and getClientConnection() checks if the closed flag has been set so if it is called after close it will just throw.

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
tkaitchuck: If socket.connect throws the Runnable supplied to `supplyAsync` will fail, which will cause the future returned from connect to fail with the exception. It would only be silent if the caller does not join on the future returned from `connect()`. (Which is unlikely as they need the connection)

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
```diff
@@ -98,20 +102,19 @@ public ClientConnection createFlow(final Flow flow, final ReplyProcessor rp) {
...
-    public void closeFlow(ClientConnection clientConnection) {
-        final ClientConnectionImpl clientConnectionImpl = (ClientConnectionImpl) clientConnection;
-        int flow = clientConnectionImpl.getFlowId();
-        log.info("Closing Flow {} for endpoint {}", flow, clientConnectionImpl.getConnectionName());
+    void closeFlow(FlowClientConnection clientConnection) {
+        int flow = clientConnection.getFlowId();
+        log.info("Closing Flow {} for endpoint {}", flow, clientConnection.getConnectionName());
         flowIdReplyProcessorMap.remove(flow);
         flowIDBatchSizeTrackerMap.remove(flow);
         if (flow == FLOW_DISABLED) {
```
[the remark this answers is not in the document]
tkaitchuck: DOne

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
```diff
@@ -40,37 +36,45 @@
...
-    }
-
-    public FlowHandler(String connectionName, MetricNotifier updateMetric) {
-        this.connectionName = connectionName;
+    private FlowHandler(PravegaNodeUri location, MetricNotifier updateMetric) {
+        this.location = location;
         this.metricNotifier = updateMetric;
     }
 
+    static CompletableFuture<FlowHandler> openConnection(PravegaNodeUri location, ClientConfig clientConfig, MetricNotifier updateMetric, ConnectionFactory connectionFactory) {
```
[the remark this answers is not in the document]
tkaitchuck: Removed

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
```diff
@@ -299,58 +231,67 @@ private void invokeProcessingFailureForAllFlows(Throwable cause) {
...
-            public void operationComplete(ChannelFuture future) throws Exception {
-                recentMessage.set(true);
-                if (!future.isSuccess()) {
-                    log.warn("Keepalive failed for connection {}", connectionName);
-                    close();
-                }
-            }
-        };
-        
+        private AtomicInteger concurrentlyRunning = new AtomicInteger(0);
```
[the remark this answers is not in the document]
tkaitchuck: done

client/src/main/java/io/pravega/client/connection/impl/FlowClientConnection.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        channel.send(append);
+    }
+
+    @Override
+    public void sendAsync(List<Append> appends, CompletedCallback callback) {
+        if (closed.get()) {
+            callback.complete(new ConnectionFailedException("Connection is closed"));
+        }
+        channel.sendAsync(appends, callback);
```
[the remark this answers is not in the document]
tkaitchuck: done

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
```diff
@@ -299,58 +231,67 @@ private void invokeProcessingFailureForAllFlows(Throwable cause) {
...
-                if (openFlowCount != 0) {
-                    log.debug("{} flows are not closed", openFlowCount);
-                    // ensure all the ReplyProcessors are informed immediately about the channel being closed.
-                    invokeProcessingFailureForAllFlows(new ConnectionClosedException());
-                }
-                final int appendTrackerCount = flowIDBatchSizeTrackerMap.size();
-                if (appendTrackerCount != 0) {
-                    log.warn("{} AppendBatchSizeTrackers are not closed", appendTrackerCount);
+            if (keepAliveFuture != null) {
+                keepAliveFuture.cancel(false);
```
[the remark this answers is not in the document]
tkaitchuck: It is more noisy, and not needed because closing a stream will unblock any read/write calls blocked on it.

common/src/main/java/io/pravega/common/util/CertificateUtils.java
```diff
@@ -0,0 +1,110 @@
...
+    @VisibleForTesting
+    static KeyStore createTrustStore(X509Certificate[] certificateChain)
+            throws KeyStoreException, IOException, NoSuchAlgorithmException, CertificateException {
+
+        final KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
+        trustStore.load(null, null);
+
+        int i = 0;
+        for (X509Certificate cert: certificateChain) {
+            String alias = i++ + "";
```
shrids, line 104: Redundant + "" , perhaps we can use String.valueOf()

ravisharda: Sure, have modified the code and pushed the change.

shrids: Thanks @ravisharda This pattern is also present @ `io.pravega.common.util.ConfigurationOptionsExtractor#extractInt`

ravisharda: Sure, thanks for spotting it. Fixed that one too.

RaulGracia:
@tkaitchuck Executed a couple of branch builders against this branch, but both failed on the same test related to this PR: `io.pravega.client.connection.impl.ClientConnectionTest.testAppend`
[branch-builder-failure.txt](https://github.com/pravega/pravega/files/4924058/branch-builder-failure.txt)

tkaitchuck:
@RaulGracia That test error should be fixed.

RaulGracia:
@tkaitchuck thanks, will trigger again tests against this branch. I also see that the DCO check is passing now, which is great. As the codecov check is really close to the target, I suggest to adding few simple tests based on the codecov report to increase coverage and make it pass too:
- `BatchClientFactory:withScope()`: This should be simple to test and will increase coverage of that class.
- `CommandEncoder:conditionalFlush()`: Whatever is inside this method is not being used, as the condition is never met. I actually have found something similar when logging the client in the past. Is that expected?
- `CommandEncoder:validateAppend()`: This method is basically untested. Maybe we should add a test case with invalid appends to check that the invalid append responses are the expected ones.
- `TcpClientConnection`: This class is very important and in general lacks from coverage in many `catch` blocks. While I understand that this may be difficult to test some times, I wonder if we could exercise some of the `catch` blocks given that exceptions will appear in the wild and we need to know that the handling is correct.
- `SocketConnectionFactoryImpl`: It seems that no test exercises this constructor `SocketConnectionFactoryImpl(ClientConfig clientConfig, ScheduledExecutorService executor)`, it should be a simple thing to cover.
- `StreamManagerImpl`: Similarly to the previous one, there is no test exercising the constructor `StreamManagerImpl(ClientConfig clientConfig)` and should be a simple thing to cover.
- `ReaderGroupManager:withScope`: This method seems not tested.
- `CertificateUtils`: This class contains 2 methods that are not tested that they may require to be (`extractCerts` and `createTrustStore`).
- `ReaderGroupManagerImpl`: This class seems to report a pretty low coverage.
- `SynchronizerClientFactory:withScope`: This method is not being tested.
- `FlowClientConnection:sendAsync`: This method is not being tested.
- `EnvVars:readIntegerFromEnvVar`: This new utility class contains an untested method.
- `ByteStreamClientFactory:withScope`: This method seems not covered neither by any test.

While some of these classes are only tangentially affected by this PR, improving some of the above points will make the codecov check to pass for sure.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -50,8 +65,7 @@ public void recordAppend(long eventNumber, int size) {
         long now = Math.max(lastAppendTime.get(), clock.get());
         long last = lastAppendTime.getAndSet(now);
         lastAppendNumber.set(eventNumber);
-        millisBetweenAppends.addNewSample(now - last);
-        appendsOutstanding.addNewSample(eventNumber - lastAckNumber.get());
+        nanosBetweenAppends.addNewSample(now - last);
```
RaulGracia, line 68: Why are we dropping `appendsOutstanding.addNewSample(eventNumber - lastAckNumber.get());` from this method? Why this was before?

tkaitchuck:
@RaulGracia Several of these methods you mention do not have coverage exist solely for tests and are not called in production code. They are called from tests, but because the tests are in different packages they aren't counted towards coverage.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
tkaitchuck: It performs better (in terms of cpu use, this showed up on a flamegraph) and provides a more consistent value if 'appendsOutstanding' is defined in a uniform way as opposed to being an mixture of two different ways of measuring. 

Before this method was called from the netty thread after the handoff had occurred. Now there is no netty thread so this is being called in-line in the write path. So the meaning of the 'outstanding' is subtly different. So I decided to simplify things and remove it from the hot path, by only sampling the value after an ack. 

This has the effect that rather than attempting to average a sawtooth distribution we are just sampling the low points. Hence this will have the effect of lowering the equivalent batch size to below where it was before. However because we are including more data in the definition of being 'outstanding' by virtue of having eliminated the netty thread, this may end up being a wash.

RaulGracia:
@tkaitchuck there is a checkstyle issue that prevent the build from passing:
```
> Task :client:checkstyleTest
[ant:checkstyle] [ERROR] /home/travis/build/pravega/pravega/client/src/test/java/io/pravega/client/connection/impl/ClientConnectionTest.java:172:23: '=' is not followed by whitespace. [WhitespaceAround]
[ant:checkstyle] [ERROR] /home/travis/build/pravega/pravega/client/src/test/java/io/pravega/client/connection/impl/ClientConnectionTest.java:172:23: '=' is not preceded with whitespace. [WhitespaceAround]
```

ravisharda:
> @tkaitchuck there is a checkstyle issue that prevent the build from passing:

Fixed.

RaulGracia:
@tkaitchuck Two tests related to this PR seem failing:
```
io.pravega.client.connection.impl.ClientConnectionTest > testFailedSendAppend FAILED
    java.lang.AssertionError: No exception thrown where: io.pravega.shared.protocol.netty.ConnectionFailedException was expected
        at io.pravega.test.common.AssertExtensions.assertThrows(AssertExtensions.java:100)
        at io.pravega.client.connection.impl.ClientConnectionTest.testFailedSendAppend(ClientConnectionTest.java:152)

io.pravega.client.connection.impl.ClientConnectionTest > testFailedSendAsync FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertNotNull(Assert.java:712)
        at org.junit.Assert.assertNotNull(Assert.java:722)
        at io.pravega.client.connection.impl.ClientConnectionTest.testFailedSendAsync(ClientConnectionTest.java:199)
```

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -73,11 +87,12 @@ public int getAppendBlockSize() {
...
         }
-        double appendsInMaxBatch = Math.max(1.0, MAX_BATCH_TIME_MILLIS / millisBetweenAppends.getCurrentValue());
-        double targetAppendsOutstanding = MathHelpers.minMax(appendsOutstanding.getCurrentValue() * 0.5, 1.0,
-                                                             appendsInMaxBatch);
-        return (int) MathHelpers.minMax((long) (targetAppendsOutstanding * eventSize.getCurrentValue()), 0,
-                                        MAX_BATCH_SIZE);
+        double nanosPerAppend = nanosBetweenAppends.getCurrentValue();
+        double appendsInMaxBatchTime = Math.max(1.0, (MAX_BATCH_TIME_MILLIS * NANOS_PER_MILLI) / nanosPerAppend);
+        double appendsInTime = Math.max(1.0, BASE_TIME_NANOS / nanosPerAppend);
+        double appendsInBatch = MathHelpers.minMax(appendsOutstanding.getCurrentValue() * OUTSTANDING_FRACTION + appendsInTime, 1.0, appendsInMaxBatchTime);
```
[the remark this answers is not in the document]
tkaitchuck: I have set PRAVEGA_BATCH_BASE_TIME_NANOS to zero.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -27,19 +30,31 @@
...
     private final Supplier<Long> clock;
     private final AtomicLong lastAppendNumber;
     private final AtomicLong lastAppendTime;
     private final AtomicLong lastAckNumber;
-    private final ExponentialMovingAverage eventSize = new ExponentialMovingAverage(1024, 0.1, true);
-    private final ExponentialMovingAverage millisBetweenAppends = new ExponentialMovingAverage(10, 0.1, false);
-    private final ExponentialMovingAverage appendsOutstanding = new ExponentialMovingAverage(2, 0.05, false);
+    private final ExponentialMovingAverage eventSize = new ExponentialMovingAverage(1024, 0.01, true);
+    private final ExponentialMovingAverage nanosBetweenAppends = new ExponentialMovingAverage(10 * NANOS_PER_MILLI, 0.001, false);
+    private final ExponentialMovingAverage appendsOutstanding = new ExponentialMovingAverage(20, 0.001, false);
```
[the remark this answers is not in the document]
tkaitchuck: I selected the new values because my local testing showed they greatly improved the max latency. 
Pavel's tests confirmed this as they showed much lower tail variance in batch sizes, and with them reverted it doubled the p99 and tripled the p999. 
I think they are better as they are.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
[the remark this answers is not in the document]
RaulGracia: The calculation is not exactly the same, right? This is adding by default 1 more append in each batch compared to removing `+ appendsInTime`. If this has any consequence, can be checked after we merge this.

andreipaduroiu:
@tkaitchuck Let's resolve the merge conflicts, build and DCO so we can get this PR merged in .

RaulGracia:
@andreipaduroiu this PR is not passing system tests. Wondering if we need to block it until they are green.

RaulGracia:
@andreipaduroiu @tkaitchuck @shrids we have had 2 green system test builds with this branch. This may unblock this PR and make it ready to get merged. If there is any unseen, sporadic failure in system tests, we can investigate it separately.

andreipaduroiu:
There are still a number of issues with this PR:
- DCO
- Conflicts
- Build
- Code con (82%)

fpj:
The codecov report says that this is below the target. I had a look at the report, and couldn't see anything that is hard to fix test-wise. codecov is not always right and often points to some naive stuff (maybe we need to tune it too), but given that overall it helps us keep code coverage under control, I'm inclined to make it happy every time we can. There is a great amount of work behind this PR, and these changes are important for Pravega, so I'm eager to see this merged in the best shape possible. Can we simply fix the issues it is pointing to, even if minor, so that we don't have to rely on an exception?

RaulGracia:
@tkaitchuck you last commit adding a new test increased coverage 0.8%, now the PR less than 1% from passing the coverage check.

PR #4876, opened 2020-06-17
Issue 4793: Tablesegment Update APIs.

### LINKED ISSUE #4793: Implement Length-Conditional Table Updates and Removals
**Problem description**
Add the ability to specify an update condition on Table Segment Updates and Removals; this condition should verify the length of the segment to match a specific one, given by the client.

**Problem location**
Segment Store/ContainerTableExtensionImpl.java

After #4792 is done, change `ContainerTableExtensionImpl` to either do a regular append (if no such offset was specified) or an offset-conditional append (if such an offset was specified).

co-jo:
**Change log description**  
* Extends the logic of `removeTableKeys` and `updateTableEntries` to allow for conditional (conditioned on a matching TableSegment offset) appends.

**Purpose of the change**  
* Closes #4793 
* Alongside PR: #4867, this should conclude #4677.

**What the code does**  
* Adds table segment offset checks in `PravegaRequestProcessor` to both it's `removeTableKeys` and `updateTableEntries` methods. These checks allow these operations to conditonally complete based on whether the provided `tableSegmentOffset` matches the expected offset.
* It first parses the `tableSegmentOffset` from the request(s), compares its value and controls program flow based on this comparison.

**How to verify it**  

* Run the `PravegaRequestProcessorTests`:
    * This PR only makes changes at the `PravegaRequestProcessor` level. Therefore it is sufficient to only test whether the new code paths are executed given the appropriate conditions.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -648,12 +651,19 @@ public void updateTableEntries(final WireCommands.UpdateTableEntries updateTable
...
-        tableStore.put(segment, entries, TIMEOUT)
-                .thenAccept(versions -> {
-                    connection.send(new WireCommands.TableEntriesUpdated(updateTableEntries.getRequestId(), versions));
-                    this.tableStatsRecorder.updateEntries(updateTableEntries.getSegment(), entries.size(), conditional.get(), timer.getElapsed());
-                })
-                .exceptionally(e -> handleException(updateTableEntries.getRequestId(), segment, operation, e));
+        long tableSegmentOffset = updateTableEntries.getTableSegmentOffset();
+        segmentInfo.thenCompose(properties -> {
+            if (tableSegmentOffset == NULL_TABLE_SEGMENT_OFFSET || tableSegmentOffset == properties.getLength()) {
+                return tableStore.put(segment, entries, TIMEOUT);
```
tkaitchuck, line 657: This offset check needs to be preformed atomically with respect to this put.
As is, this code will only check the length at the time the request arrived, but it could change by the time `put` is processed. So the value needs to be passed into tablestore to make the update atomic.

andreipaduroiu: @co-jo You can push this check down all the way into `ContainerTableExtensionImpl` in the `commit` method. That method invokes `DirectSegmentAccess.append` which should have an overload that takes an offset (if it doesn't then we should add one (the plumbing underneath is already done)). That will guarantee this to be an atomic length-conditional append.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1662,7 +1662,7 @@ public static WireCommand readFrom(ByteBufInputStream in, int length) throws IOE
             String segment = in.readUTF();
             String delegationToken = in.readUTF();
             TableEntries entries = (TableEntries) TableEntries.readFrom(in, in.available());
-            long tableSegmentOffset = (in.available() > 0 ) ? in.readLong() : NULL_TABLE_SEGMENT_OFFSET;
+            long tableSegmentOffset = (in.available() >= 0 ) ? in.readLong() : NULL_TABLE_SEGMENT_OFFSET;
```
co-jo, line 1665: Note: I'm assuming that an offset of 0 is valid.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/tables/TableStore.java
```diff
@@ -144,6 +145,32 @@ default int maximumValueLength() {
      */
     CompletableFuture<List<Long>> put(String segmentName, List<TableEntry> entries, Duration timeout);
 
+    /**
+     * Inserts new or updates existing Table Entries into the given Table Segment.
```
andreipaduroiu, line 149: add " conditioned on the Segment's Length matching an expected value".

do this for `remove` too.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/tables/TableStore.java
```diff
@@ -144,6 +145,32 @@ default int maximumValueLength() {
...
+     * Key provided. The versions will be in the same order as the TableEntry instances provided. If the operation failed,
+     * the future will be failed with the causing exception. Notable exceptions:
+     * <ul>
+     * <li>{@link StreamSegmentNotExistsException} If the Table Segment does not exist.</li>
+     * <li>{@link TableKeyTooLongException} If {@link TableEntry#key} exceeds {@link #maximumKeyLength()}.</li>
+     * <li>{@link TableValueTooLongException} If {@link TableEntry#value} exceeds {@link #maximumValueLength()}.</li>
+     * <li>{@link ConditionalTableUpdateException} If {@link TableEntry#key} {@link TableKey#hasVersion() hasVersion() } is true and
+     * {@link TableEntry#key} {@link TableKey#version} does not match the Table Entry's Key current Table Version. </li>
+     * <li>{@link BadSegmentTypeException} If segmentName refers to a non-Table Segment. </li>
+     * <li>{@link BadOffsetException} If there is a mismatch between the provided {@param tableSegmentOffset} and the actual offset.
```
andreipaduroiu, line 169: "the actual offset" -> "the actual segment length"

change it below too.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/DirectSegmentAccess.java
```diff
@@ -51,14 +51,34 @@
...
+    /**
+     * Appends a range of bytes at the end of the Segment and atomically updates the given attributes. The byte range
+     * will be appended as a contiguous block, however there is no guarantee of ordering between different calls to this
+     * method.
+     * @see io.pravega.segmentstore.contracts.StreamSegmentStore#append(String, BufferView, Collection, Duration)
+     *
+     * @param data             The data to add.
+     * @param attributeUpdates A Collection of Attribute-Values to set or update. May be null (which indicates no updates).
+     *                         See Notes about AttributeUpdates in the interface Javadoc.
+     * @param offset           The expected offset that the append will occur at.
```
andreipaduroiu, line 63: "The expected length of the segment. If the current length of the Segment does not equal this value, the operation will fail with a {@link BadOffsetException}."

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -882,6 +882,15 @@ private void shutdownWhenStopped(Service component, String componentName) {
                     .thenApply(v -> operation.getStreamSegmentOffset());
         }
 
+        @Override
+        public CompletableFuture<Long> append(BufferView data, Collection<AttributeUpdate> attributeUpdates, long tableSegmentOffset, Duration timeout) {
```
andreipaduroiu, line 886: tableSegmentOffset -> offset

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -53,6 +54,7 @@
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 
+import static io.pravega.shared.protocol.netty.WireCommands.NULL_TABLE_SEGMENT_OFFSET;
```
andreipaduroiu, line 57: No. Do not import this value. This code should not depend on `io.pravega.shared.protocol`.

Define an internal constant `NO_OFFSET`, set it to `-1`, and use it just in this class. Then in `commit`, invoke the appropriate API of `DirectSegmentAccess` based on whether you have this value or not. (actually you already make this check, so just change it to the internal value)

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -320,9 +332,15 @@ protected int getMaxCompactionSize() {
     }
 
     private <T> CompletableFuture<Long> commit(Collection<T> toCommit, Function<Collection<T>, BufferView> serializer,
-                                               DirectSegmentAccess segment, Duration timeout) {
-        BufferView s = serializer.apply(toCommit);
-        return segment.append(s, null, timeout);
+                                               DirectSegmentAccess segment, long tableSegmentOffset, Duration timeout) {
+        if (tableSegmentOffset == NULL_TABLE_SEGMENT_OFFSET || tableSegmentOffset == segment.getInfo().getLength()) {
```
andreipaduroiu, line 336: Remove the second part of this check. You cannot rely on the length of the segment at this point as there may be multiple operations in flight that will affect it before your operation is processed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -320,9 +332,15 @@ protected int getMaxCompactionSize() {
...
+            BufferView s = serializer.apply(toCommit);
+            return segment.append(s, null, timeout);
+        } else {
+            CompletableFuture<Long> future = new CompletableFuture<>();
```
andreipaduroiu, line 340: Invoke the appropriate `DirectSegmentAccess.append` method here instead of throwing this.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
andreipaduroiu: An offset of 0 is valid, but why are you making this change? This checks for how many bytes are available for reading from the buffer stream.

co-jo: Touche.. Had a little bit of a brain fart. Will revert.

andreipaduroiu:
You'll need to add a bit more code coverage. GitHub will prevent me from merging if codecov isn't happy.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerTableExtensionImpl.java
```diff
@@ -333,14 +336,8 @@ protected int getMaxCompactionSize() {
...
                                                DirectSegmentAccess segment, long tableSegmentOffset, Duration timeout) {
-        if (tableSegmentOffset == NULL_TABLE_SEGMENT_OFFSET || tableSegmentOffset == segment.getInfo().getLength()) {
             BufferView s = serializer.apply(toCommit);
-            return segment.append(s, null, timeout);
-        } else {
-            CompletableFuture<Long> future = new CompletableFuture<>();
-            future.completeExceptionally(new BadOffsetException(segment.getInfo().getName(), segment.getInfo().getLength(), tableSegmentOffset));
-            return future;
-        }
+            return segment.append(s, null, tableSegmentOffset, timeout);
```
andreipaduroiu, line 340: I was only asking you to do the conditional commit for the second case. Please bring back the `if-else` block and only pass the explicit offset in the `else` case. Do not pass `tableSegmentOffset` in the `if` case as it's just a coincidence that `-1` is the "no-offset" append that is now in the Segment Store. By using the other `append` (non-conditional), you shield yourself from any future change in the segment store that may affect how that works.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/DirectSegmentAccess.java
```diff
@@ -51,14 +51,35 @@
...
     CompletableFuture<Long> append(BufferView data, Collection<AttributeUpdate> attributeUpdates, Duration timeout);
 
+    /**
+     * Appends a range of bytes at the end of the Segment and atomically updates the given attributes. The byte range
+     * will be appended as a contiguous block, however there is no guarantee of ordering between different calls to this
+     * method.
+     * @see io.pravega.segmentstore.contracts.StreamSegmentStore#append(String, BufferView, Collection, Duration)
+     *
+     * @param data             The data to add.
+     * @param attributeUpdates A Collection of Attribute-Values to set or update. May be null (which indicates no updates).
```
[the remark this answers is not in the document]
co-jo: `StreamSegmentContainer` handles and empty list and `null` equally. May need @andreipaduroiu's thoughts on this.

PR #4882, opened 2020-06-19
Issue 4875: Modify name resolution strategy of configuration properties for backward compatibility

### LINKED ISSUE #4875: Segment Store doesn't expose metrics with legacy property names.
**Problem description**
I've used Pravega docker image built from  https://github.com/pravega/pravega/commit/3a0164c0bf584d637f73aeb9991c82cb7fc348f6 while I was deploying Pravega on Kubernetes cluster. This manifest A / B I've used could be found below.
Using **Manifest A** (legacy property names), I was able to get only Controller's metrics in my Influx DB while Segment Store haven't sent them. With **Manifest B** (new property names), I was able to get metrics for both.
Note that I haven't observed any metrics-related errors in Segment Store's logs.
According to [MetricsConfig class](https://github.com/pravega/pravega/blob/master/shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java#L23), I assume Pravega should support backward compatibility for the legacy property names. 

**Manifests**
`<InfluxDB target DB name>` - name of DB inside my Influx DB instance;
`<INFLUX DB URI>` - URI of my InfluxDB;

- **1. Manifest A: legacy metrics' property names, Segment Store doesn't expose metrics**
```
apiVersion: "pravega.pravega.io/v1beta1"
kind: "PravegaCluster"
metadata:
  name: "pravega-cluster"
spec:
  version: "0.8.0-2568.efa9e08b2"
  zookeeperUri: zookeeper-client:2181
  bookkeeperUri: "bookkeeper-bookie-0.bookkeeper-bookie-headless.default.svc.cluster.local:3181"

  pravega:
    controllerReplicas: 1
    controllerResources:
      requests:
        memory: "3Gi"
        cpu: "500m"
      limits:
        memory: "5Gi"
        cpu: "3000m"

    segmentStoreReplicas: 1
    segmentStoreResources:
      requests:
        memory: "4Gi"
        cpu: "1000m"
      limits:
        memory: "9Gi"
        cpu: "7000m"

    cacheVolumeClaimTemplate:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "standard"
      resources:
        requests:
          storage: 1Gi

    image:
      repository: "andreykoltsov/pravega"
      pullPolicy: Always

    options:
      metrics.enableStatistics: "true"
      metrics.outputFrequencySeconds: "10"
      metrics.enableInfluxDBReporter: "true"
      metrics.influxDBURI: "<INFLUX DB URI>"
      metrics.influxDBName: "<InfluxDB target DB name>"
      metrics.influxDBRetention: "two_hour"

      controller.metrics.enableStatistics: "true"
      controller.metrics.outputFrequencySeconds: "10"
      controller.metrics.enableInfluxDBReporter: "true"
      controller.metrics.influxDBURI: "<INFLUX DB URI>"
      controller.metrics.influxDBName: "<InfluxDB target DB name>"
      controller.metrics.influxDBRetention: "two_hour"

      controller.containerCount: "32"
      controller.retention.bucketCount: "10"
      controller.service.asyncTaskPoolSize: "20"
      controller.retention.threadCount: "4"
      writer.flushThresholdBytes: "1048576"
      writer.maxFlushSizeBytes: "1048576"
      pravegaservice.containerCount: "32"
```

- *2. Manifest B*: updated property names, Segment Store exposes metrics just fine. Changed section has been marked with ` Start / End of changes`. Basically, I've changed all of the properties that're under `metrics.*` field:

```
apiVersion: "pravega.pravega.io/v1beta1"
kind: "PravegaCluster"
metadata:
  name: "pravega-cluster"
spec:
  version: "0.8.0-2568.efa9e08b2"
  zookeeperUri: zookeeper-client:2181
  bookkeeperUri: "bookkeeper-bookie-0.bookkeeper-bookie-headless.default.svc.cluster.local:3181"

  pravega:
    controllerReplicas: 1
    controllerResources:
      requests:
        memory: "3Gi"
        cpu: "500m"
      limits:
        memory: "5Gi"
        cpu: "3000m"

    segmentStoreReplicas: 1
    segmentStoreResources:
      requests:
        memory: "4Gi"
        cpu: "1000m"
      limits:
        memory: "9Gi"
        cpu: "7000m"

    cacheVolumeClaimTemplate:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "standard"
      resources:
        requests:
          storage: 1Gi

    image:
      repository: "andreykoltsov/pravega"
      pullPolicy: Always

    options:

      # -- Start of changes
      metrics.statistics.enable: "true"
      metrics.influxDB.reporter.enable: "true"
      metrics.influxDB.connect.uri: "<INFLUX DB URI>"
      metrics.influxDB.connect.db.name: "<InfluxDB target DB name>"
      metrics.influxDB.retention: "two_hour"
      metrics.output.frequency.seconds: "10"
      controller.metrics.statistics.enable: "true"
      controller.metrics.influxDB.reporter.enable: "true"
      controller.metrics.influxDB.connect.uri: <INFLUX DB URI>
      controller.metrics.influxDB.connect.db.name: "<InfluxDB target DB name>"
      controller.metrics.influxDB.retention: "two_hour"
      controller.metrics.output.frequency.seconds: "10"
      # -- End of changes

      controller.containerCount: "32"
      controller.retention.bucketCount: "10"
      controller.service.asyncTaskPoolSize: "20"
      controller.retention.threadCount: "4"
      writer.flushThresholdBytes: "1048576"
      writer.maxFlushSizeBytes: "1048576"
      pravegaservice.containerCount: "32"
```

**Problem location**
I assume `Segment Store Service`

None

ravisharda:
**Change log description**  
Modify the name resolution strategy for configuration properties to use legacy names first, for backward compatibility.

**Purpose of the change**  
Resolves #4875. 

**What the code does**  

Previously, the properties were resolved in a way that if a property with a new name was present, it'd override any property specified using the old name. For example, if you specified both `metrics.enableStatistics=true` (old name) and `metrics.statistics.enable=false` (new name), segment store metrics would be `enabled`. These properties may be specified either via properties file (`config.properties`) or Java system properties (say, via the deployment manifest file). 

However, this behavior doesn't work in some cases. For example, Segment Store loads all the properties from `config.properties` as well as the system properties. In the `config.properties` a preconfigured property `metrics.statistics.enable=false` is already set.  So, if you specified the above property as `metrics.enableStatistics=true` in the deployment manifest, the default from `config.properties` file would still apply. That is why segment store metrics were not showing up, as described in issue #4875.  

https://github.com/pravega/pravega/blob/3a0164c0bf584d637f73aeb9991c82cb7fc348f6/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java#L256-L260

Note that this problem was limited to configuration properties that had default values specified in the configuration properties file. 

To ensure that such properties can be overridden properly using old property names, this PR changes the resolution order. So, if you specify `metrics.enableStatistics=true` in the deployment manifest, segment store metrics would be enabled. If you specified it using the new name `metrics.statistics.enable=true`, then too metrics will be enabled. 

**How to verify it**  
All tests, including system tests, must pass.

### DIFF
--- a/common/src/main/java/io/pravega/common/util/TypedProperties.java
+++ b/common/src/main/java/io/pravega/common/util/TypedProperties.java
@@ -125,27 +125,35 @@ public class TypedProperties {
     }
 
     private <T> T tryGet(Property<T> property, Function<String, T> converter) {
-        String fullKeyName = this.keyPrefix + property.getName();
+        String propNewName = this.keyPrefix + property.getName();
+        String propOldName = this.keyPrefix + property.getLegacyName();
+        String propValue = null;
 
-        // Get value from config.
-        String value = this.properties.getProperty(fullKeyName, null);
-
-        if (value == null && property.hasLegacyName()) {
-            value = this.properties.getProperty(this.keyPrefix + property.getLegacyName(), null);
+        if (property.hasLegacyName()) {
+            // Value of property with old name
+            propValue = this.properties.getProperty(propOldName, null);
         }
-        if (value == null) {
-            // 2. Nothing in the configuration for this Property.
+
+        if (propValue == null) {
+            // Value of property with new name
+            propValue = this.properties.getProperty(propNewName, null);
+        }
+
+        // A property with neither old not new name was defined, so the property value is still null
+        if (propValue == null) {
             if (property.hasDefaultValue()) {
                 return property.getDefaultValue();
             } else {
-                throw new MissingPropertyException(fullKeyName);
+                throw new MissingPropertyException(
+                        String.format("Missing property with name [%s] (new name) / [%s] (old name}",
+                        propNewName, propOldName));
             }
         }
 
         try {
-            return converter.apply(value.trim());
+            return converter.apply(propValue.trim());
         } catch (IllegalArgumentException ex) {
-            throw new InvalidPropertyValueException(fullKeyName, value, ex);
+            throw new InvalidPropertyValueException(propNewName, propValue, ex);
         }
     }
 
--- a/common/src/main/java/io/pravega/common/util/TypedProperties.java
+++ b/common/src/main/java/io/pravega/common/util/TypedProperties.java
@@ -11,6 +11,7 @@ package io.pravega.common.util;
 
 import com.google.common.base.Preconditions;
 import io.pravega.common.Exceptions;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.Properties;
 import java.util.function.Function;
@@ -29,6 +30,7 @@ import java.util.function.Function;
  * </ul>
  * Indicate that namespace "foo" has Key-Values (key1=value1, key2=value2), and namespace "bar" has (key1=value3, key3=value4).
  */
+@Slf4j
 public class TypedProperties {
     //region Members
     private static final String SEPARATOR = ".";
@@ -132,6 +134,10 @@ public class TypedProperties {
         if (property.hasLegacyName()) {
             // Value of property with old name
             propValue = this.properties.getProperty(propOldName, null);
+            if (propValue != null) {
+                log.warn("Deprecated property name '{}' used. Please use '{}' instead. Support for the old " +
+                        "property name will be removed in a future version of Pravega.", propOldName, propNewName);
+            }
         }
 
         if (propValue == null) {
@@ -139,7 +145,7 @@ public class TypedProperties {
             propValue = this.properties.getProperty(propNewName, null);
         }
 
-        // A property with neither old not new name was defined, so the property value is still null
+        // This property wasn't specified using either new or old name, so the property value is still null
         if (propValue == null) {
             if (property.hasDefaultValue()) {
                 return property.getDefaultValue();
--- a/common/src/test/java/io/pravega/common/util/TypedPropertiesTests.java
+++ b/common/src/test/java/io/pravega/common/util/TypedPropertiesTests.java
@@ -98,15 +98,17 @@ public class TypedPropertiesTests {
         assertEquals("configValue", config);
     }
 
+
+
     @Test
-    public void testNewPropertyTakesPrecedenceOverLegacy() {
+    public void testLegacyPropertyTakesPrecedenceOverNewOnesWhenSpecified() {
         Properties props1 = new Properties();
-        props1.setProperty("ns1.newPropertyKey", "configValue");
-        props1.setProperty("ns1.legacyPropertyKey", "configValue");
+        props1.setProperty("ns1.newPropertyKey", "new");
+        props1.setProperty("ns1.legacyPropertyKey", "old");
         TypedProperties typedProps = new TypedProperties(props1, "ns1");
         String config = typedProps.get(Property.named(
                 "newPropertyKey", "default", "legacyPropertyKey"));
-        assertEquals("configValue", config);
+        assertEquals("old", config);
     }
 
     private <T> void testData(Properties props, ExtractorFunction<T> methodToTest, Predicate<String> valueValidator) throws Exception {


common/src/main/java/io/pravega/common/util/TypedProperties.java
```diff
@@ -125,27 +125,35 @@ public boolean getBoolean(Property<Boolean> property) throws ConfigurationExcept
...
-        String fullKeyName = this.keyPrefix + property.getName();
+        String propNewName = this.keyPrefix + property.getName();
+        String propOldName = this.keyPrefix + property.getLegacyName();
+        String propValue = null;
 
-        // Get value from config.
-        String value = this.properties.getProperty(fullKeyName, null);
+        if (property.hasLegacyName()) {
+            // Value of property with old name
+            propValue = this.properties.getProperty(propOldName, null);
```
andreipaduroiu, line 134: Let's add a `log.warn` to this code and notify users that they are using a deprecated property name. 

maybe something like 
`Deprecated property name '{}' used. Please use '{}' instead. Support for the old property name will be removed in a future version of Pravega.`

common/src/main/java/io/pravega/common/util/TypedProperties.java
```diff
@@ -125,27 +125,35 @@ public boolean getBoolean(Property<Boolean> property) throws ConfigurationExcept
...
+        }
 
-        if (value == null && property.hasLegacyName()) {
-            value = this.properties.getProperty(this.keyPrefix + property.getLegacyName(), null);
+        if (propValue == null) {
+            // Value of property with new name
+            propValue = this.properties.getProperty(propNewName, null);
```
kevinhan88, line 139: Same here, please call out it's new property name being used.

common/src/main/java/io/pravega/common/util/TypedProperties.java
```diff
@@ -125,27 +125,35 @@ public boolean getBoolean(Property<Boolean> property) throws ConfigurationExcept
...
         }
-        if (value == null) {
-            // 2. Nothing in the configuration for this Property.
+
+        // A property with neither old not new name was defined, so the property value is still null
+        if (propValue == null) {
             if (property.hasDefaultValue()) {
                 return property.getDefaultValue();
```
kevinhan88, line 145: Also call out in log that default value is being used.

andreipaduroiu: No, there is no need to pollute the logs with this information. Both the Segment store and controller log the values so you should be able to see them if you parse the logs. 

No point in adding noise to the logs.

common/src/main/java/io/pravega/common/util/TypedProperties.java
kevinhan88: Consider adding code to handle the scenario that both legacy and current properties are being defined.
This is the most confusing scenario, so please log it in details, e.g. "both legacy and current properties defined as blah blah, and legacy properties is used."

andreipaduroiu: I disagree. See below.

Just log the most important thing, which is when you are still using something that will not be supported in the future.

common/src/main/java/io/pravega/common/util/TypedProperties.java
ravisharda: Yes, that makes sense. I've added this comment now.

common/src/main/java/io/pravega/common/util/TypedProperties.java
ravisharda: I agree with @andreipaduroiu here.

ravisharda:
A system test run was triggered and the job passed. Test Pravega Continuous Job id: `2944`.

PR #4888, opened 2020-06-22
add unit-test for RG notifications

ravibeta:
**Change log description**  
We were missing a unit-test case for the different notifications we have per stream policy. This merely adds that unit-test

**Purpose of the change**  
test hole coverage for reader group notification discovered from investigation

**What the code does**  
It adds a test for the case when we can have more than one notifications.

**How to verify it**  
The unit-test is run as part of the command-line option to run tests and the successful output is posted on the commit history.


ravibeta:
The failed tests
```
io.pravega.segmentstore.server.store.StreamSegmentServiceTests > testEndToEndWithFencing PASSED
io.pravega.segmentstore.server.containers.StreamSegmentContainerTests > testWriteFenceOut PASSED

BUILD SUCCESSFUL in 13s
```
are not due to this change and work fine on local workspace.

medvedevigorek:
@ravibeta, you need to fix DCO error. I suggest you close this PR, create a new branch following the naming convention as stated in [the guidelines](https://github.com/pravega/pravega/wiki/Contributing#issuesandprs), cherry-pick related commits from your original branch and commit them with the proper signing. Then open a new PR for the new branch.

ravibeta:
Yes Igor, the DCO checks are successful now. The commits are also appearing as verified and with proper sign off. The branch is temporary and will be deleted when this pr is merged. May I keep the same PR now that all the DCO errors are fixed in this PR itself? Thank you very much. Cheers, Ravi

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,24 +196,107 @@ public void testSegmentNotifications() throws Exception {
         assertEquals(2, segmentNotificationPostScale.getNumOfSegments());
     }
 
+    @Test(timeout = 40000)
+    public void testTargetRateNotifications() throws Exception {
```
shrids, line 200: Isn't this test case similar to `io.pravega.test.integration.ReaderGroupNotificationTest#testSegmentNotifications` ? 
The only difference is that the new test waits for Stream to auto-scaled by the controller while the existing test manually increases the number of segments by invoking controller.scaleAPI(). 
Let me know if I missed anything 🤔

ravibeta: Yes Sandeep, the difference is in the number of notifications. The former is fixed and the latter is based on rate. Thanks.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,24 +196,107 @@ public void testSegmentNotifications() throws Exception {
...
+                    try {
+                        initialSegmentNotification = notificationResults.take();
+                    } catch (InterruptedException e) {
+                        log.error("Exception:", e);
+                        throw new IllegalStateException(e);
+                    }
+                    assertNotNull(initialSegmentNotification);
+                    assertTrue(initialSegmentNotification.getNumOfSegments() > 2);
+                    if (streamSegments.getSegments().size() > 2) {
+                        System.err.println("Success");
```
shrids, line 268: please remove system.err at all places.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,24 +196,107 @@ public void testSegmentNotifications() throws Exception {
...
+                    segmentNotifier.pollNow();
+                    SegmentNotification initialSegmentNotification = null;
+                    try {
+                        initialSegmentNotification = notificationResults.take();
+                    } catch (InterruptedException e) {
+                        log.error("Exception:", e);
+                        throw new IllegalStateException(e);
+                    }
+                    assertNotNull(initialSegmentNotification);
+                    assertTrue(initialSegmentNotification.getNumOfSegments() > 2);
```
shrids, line 266: the initial number of segments in the stream is 3, so the initial value should be definitely 3.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,24 +196,107 @@ public void testSegmentNotifications() throws Exception {
...
+                    if (streamSegments.getSegments().size() > 2) {
```
shrids, line 267: if the initial number of segments is 3 then the test would check the currentSegments once and would be successful.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,6 +196,87 @@ public void testSegmentNotifications() throws Exception {
...
+        SegmentNotifier segmentNotifier = (SegmentNotifier) readerGroup.getSegmentNotifier(executor);
+        segmentNotifier.registerListener(l1);
+
+        long start = System.currentTimeMillis();
+        CompletableFuture.runAsync(() -> {
+            while (System.currentTimeMillis() - start < Duration.ofMinutes(3).toMillis()) {
+                try {
+                    writer1.writeEvent(routingKeys.get(0), "data1").get();
+                } catch (Throwable e) {
+                    log.error("test exception: ", e);
```
andreipaduroiu, line 245: This will not fail the test. Throw the exception.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,6 +196,87 @@ public void testSegmentNotifications() throws Exception {
...
+        Listener<SegmentNotification> l1 = notification -> {
+            log.info("Number of Segments: {}, Number of Readers: {}", notification.getNumOfSegments(), notification.getNumOfReaders());
+            notificationResults.add(notification);
+        };
+        SegmentNotifier segmentNotifier = (SegmentNotifier) readerGroup.getSegmentNotifier(executor);
+        segmentNotifier.registerListener(l1);
+
+        long start = System.currentTimeMillis();
+        CompletableFuture.runAsync(() -> {
+            while (System.currentTimeMillis() - start < Duration.ofMinutes(3).toMillis()) {
```
andreipaduroiu, line 241: So you're writing an integration test that deliberately takes 3 minutes to run? The only thing this will do is make our build take longer for everyone who wants to run it.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,6 +196,87 @@ public void testSegmentNotifications() throws Exception {
...
+                            assertNotNull(initialSegmentNotification);
+                            assertTrue(initialSegmentNotification.getNumOfSegments() == 3);
+                            if (streamSegments.getSegments().size() == 3) {
+                                log.info("Success");
+                            } else {
+                                throw new IllegalStateException("expected number of segments to be 3");
+                            }
+                        }), executor)
+                .exceptionally(e -> {
+                    log.error("Failure: ", e);
```
andreipaduroiu, line 274: This will not fail the test.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,6 +196,87 @@ public void testSegmentNotifications() throws Exception {
...
+                    return null;
+                }).get();
+    ExecutorServiceHelpers.shutdown(executor);
```
andreipaduroiu, line 277: This is done for your in `teardown()`.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,6 +196,87 @@ public void testSegmentNotifications() throws Exception {
...
+        //Add segment event listener
+        Listener<SegmentNotification> l1 = notification -> {
+            log.info("Number of Segments: {}, Number of Readers: {}", notification.getNumOfSegments(), notification.getNumOfReaders());
+            notificationResults.add(notification);
+        };
+        SegmentNotifier segmentNotifier = (SegmentNotifier) readerGroup.getSegmentNotifier(executor);
+        segmentNotifier.registerListener(l1);
+
+        long start = System.currentTimeMillis();
+        CompletableFuture.runAsync(() -> {
```
andreipaduroiu, line 240: You are not executing this future on any explicit executor nor are you making sure you stop its execution when your test is done. As such, this will continue running for long after your test is done.

andreipaduroiu:
BTW, you are adding an integration test, not unit test.

ravibeta:
Yes Andrei

ravibeta:
``` Should manual scaling of streams generate notifications ? Currently auto scaling of streams do ?
In this integration test case, I attempt to generate a lot of segments.

Segments do not overlap in both horizontal and vertical directions and need not align on the same time boundaries. Each segment can be started and sealed independently so that no more events are appended to it. The number of segments for a stream depends on the scaling policy specified at the time of stream creation. A scaling policy can be fixed, data-based or event-based to accommodate varying size and count of segments according to anticipated load.  This logical representation of a stream is hard for a user to visualize when the data is being read while it is arriving by the order of millions from a device.  The user has the guarantee that the events will be written in order of their arrival and none of the events will be lost during write and subsequent read but the user does not know the progress made in terms of reading the segments or if all the data has been read. If the user knew the progress, the readers could be scaled accordingly to keep up or complete reading at a faster rate. 
```

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,130 @@ public void testSegmentNotifications() throws Exception {
...
+                                }
+                                segmentNotifier.pollNow();
+                                SegmentNotification initialSegmentNotification = notificationResults.poll();
+                                log.info("notification={}", initialSegmentNotification);
+                                readerGroup.resetReaderGroup(readerGroupConfig);
+                                readerGroup.readerOffline("readerId", null);
+                                long segmentSize = streamSegments.getSegments().size();
+                                log.info("segment_size={}", segmentSize);
+                                if (segmentSize > 3) {
+                                    log.info("success");
```
shrids, line 310: This implies as soon as the segment size grows beyond 3 the test would stop, right?

ravibeta: Yes Sandeep.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -265,5 +398,57 @@ public void testEndOfStreamNotifications() throws Exception {
...
+    Pair<List<Long>, Map<Double, Double>> getScaleInput(ArrayList<Segment> sortedCurrentSegments) {
+        return new ImmutablePair<>(getSegmentsToSeal(sortedCurrentSegments), getNewRanges());
+    }
 
+    private List<Long> getSegmentsToSeal(ArrayList<Segment> sorted) {
+        return sorted.stream()
+                .map(Segment::getSegmentId).collect(Collectors.toList());
+    }
+
+    private Map<Double, Double> getNewRanges() {
```
andreipaduroiu, line 443: Please explain what you're doing here.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -265,5 +398,57 @@ public void testEndOfStreamNotifications() throws Exception {
...
+    
+    private List<List<Segment>> scale(Controller controller, String scopeName, String streamName, int numSegments, int scalesToPerform, ScheduledExecutorService executorService) {
+        Stream stream = new StreamImpl(scopeName, streamName);
+        AtomicInteger counter = new AtomicInteger(0);
+        List<List<Segment>> listOfEpochs = new LinkedList<>();
+
+        CompletableFuture<Void> scaleFuture = Futures.loop(() -> counter.incrementAndGet() <= scalesToPerform,
+                () -> controller.getCurrentSegments(scopeName, streamName)
+                        .thenCompose(segments -> {
+                            ArrayList<Segment> sorted = Lists.newArrayList(segments.getSegments().stream()
```
andreipaduroiu, line 410: 1. Why do you make a copy of a list that you have just created and you only use for one purpose?
2. Why do you use `Lists.newArrayList`? What's wrong with `new ArrayList<>(...)`?
3. I don't understand what this sorting is trying to achieve.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -265,5 +398,57 @@ public void testEndOfStreamNotifications() throws Exception {
...
+                            Map<Double, Double> newRanges = scaleInput.getValue();
+
+                            return controller.scaleStream(stream, segmentsToSeal, newRanges, executorService)
+                                    .getFuture()
+                                    .thenAccept(scaleStatus -> {
+                                        log.info("scale stream for epoch {} completed with status {}", counter.get(), scaleStatus);
+                                    });
+                        }), executorService);
+
+        scaleFuture.join();
```
andreipaduroiu, line 429: Instead of writing a future and then immediately joining it, it may be simpler to write a classing `while` loop and join on individual async futures. That's OK for writing tests.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+        } catch (InterruptedException e) {
+            log.error("Interrupted: ", e);
+        }
+        Retry.withExpBackoff(10, 10, 100, 10000)
+                .retryingOn(IllegalStateException.class)
+                .throwingOn(RuntimeException.class)
+                .runAsync(() -> controller.getCurrentSegments(SCOPE, streamName)
+                        .thenAccept(streamSegments -> {
+                            try {
+                                Thread.sleep(1000);
```
andreipaduroiu, line 284: No sleeping

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+                                EventStreamReader<String> reader1 = clientFactory.createReader("readerId", "reader", new JavaSerializer<>(),
```
andreipaduroiu, line 285: use `readerGroup.getName` instead of hardcoding the name.

andreipaduroiu: Also put `readerId` in a constant to avoid copy-pasting everywhere.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
andreipaduroiu: This pattern of ending the method with throwing an exception is counter-intuitive.
Use `AssertExtensions.assertGreaterThanOrEquals(...)` to validate.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+                                readerGroup.resetReaderGroup(readerGroupConfig);
+                                readerGroup.readerOffline("readerId", null);
+                                long segmentSize = streamSegments.getSegments().size();
+                                log.info("segment_size={}", segmentSize);
+                                if (segmentSize > 3) {
+                                    log.info("success");
+                                    return;
+                                }
+                                throw new IllegalStateException("expected number of segments to be > 3");
+                            } catch (InterruptedException e) {
```
andreipaduroiu, line 312: This is an incorrect way of handling `InterruptedException`.

Use `Exceptions.handleInterrupted(...)` on whichever call declares that it's throwing this and remove this catch clause.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+                                log.error("Exception:", e);
+                                throw new IllegalStateException(e);
+                            }
+                        }), threadPoolExecutor)
+                .whenComplete((r, ex) -> {
+                        if (ex != null) {
+                            log.error("Failure: ", ex);
```
andreipaduroiu, line 319: This is superfluous. Let the exception bubble up and the test will be automatically failed.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+                            Assert.fail(String.format("Unexpected error occurred running this test. %s", ex));
+                        }
+                }).get();
```
andreipaduroiu, line 322: You should be able to simplify this by replacing `runAsync` with `run` (then call `join` on whatever Future you need to wait). It will make your code look much simpler.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+        @Cleanup
+        ReaderGroup readerGroup = groupManager.getReaderGroup("reader");
+
+        long start = System.currentTimeMillis();
+        log.info("Segments found as: {}", controller.getCurrentSegments(SCOPE, streamName).get());
+        CompletableFuture.runAsync(() -> {
+            while (System.currentTimeMillis() - start < Duration.ofMinutes(1).toMillis()) {
+                try {
+                    writer1.writeEvent(routingKeys.get(random.nextInt(3)), "data").get();
+                    Thread.sleep(0);
```
andreipaduroiu, line 260: Really?

andreipaduroiu:
> The failed tests
> 
> ```
> io.pravega.segmentstore.server.store.StreamSegmentServiceTests > testEndToEndWithFencing PASSED
> io.pravega.segmentstore.server.containers.StreamSegmentContainerTests > testWriteFenceOut PASSED
> io.pravega.segmentstore.server.reading.ContainerReadIndexTests > testStorageReadsConcurrentWithOverwrite PASSED
> 
> BUILD SUCCESSFUL in 13s
> ```
> 
> are not due to this change and work fine on local workspace.

@ravibeta Just because they "work fine on local workspace", it doesn't mean they are correct. Travis is a much, much slower machine with fewer resources than your "local workspace" which means your code will be subject to delays beyond its control (GC, slow cpu, etc). If something fails on Travis but not locally then you need to look at both your production code and test code and see why that happens. It could be too aggressive timeouts, incorrect assumption about async method executions, not properly waiting on async Futures to complete or others. In "others", I include all those `Thread.sleep` calls I see throughout this PR. While in general these are just time wasters, whatever problem they "solve" on your "local workspace" may not be that well resolved in other build environments.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -265,5 +396,57 @@ public void testEndOfStreamNotifications() throws Exception {
...
+                                        log.info("scale stream for epoch {} completed with status {}", counter.get(), scaleStatus);
+                                    });
+                        }), executorService);
+
+        scaleFuture.join();
+
+        return listOfEpochs;
+    }
+
+    Pair<List<Long>, Map<Double, Double>> getScaleInput(ArrayList<Segment> sortedCurrentSegments) {
```
tkaitchuck, line 432: Here and below, rather than using two doubles, why not use Range?

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+                    writer1.writeEvent(routingKeys.get(random.nextInt(3)), "data").get();
+                    Thread.sleep(0);
+                } catch (Throwable e) {
+                    log.error("test exception: ", e);
+                    throw new IllegalStateException(e);
+                }
+            }
+        }, threadPoolExecutor)
+                .exceptionally(e -> {
+                    log.error("Failure: ", e);
```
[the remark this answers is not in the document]
ravibeta: Yes Andre

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -105,6 +105,7 @@
      * Add this reader to the reader group so that it is able to acquire segments
      */
     void initializeReader(long initialAllocationDelay) {
+        sync.fetchUpdates();
```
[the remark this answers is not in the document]
ravibeta: Yes Andre this is required

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -77,12 +98,14 @@
     private ControllerWrapper controllerWrapper;
     private ServiceBuilder serviceBuilder;
     private ScheduledExecutorService executor;
+    private ScheduledExecutorService threadPoolExecutor;
```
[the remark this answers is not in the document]
ravibeta: Yes Andre fixed

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -77,12 +98,14 @@
...
     private AtomicBoolean listenerInvoked = new AtomicBoolean();
     private ReusableLatch listenerLatch = new ReusableLatch();
 
     @Before
     public void setUp() throws Exception {
         executor = Executors.newSingleThreadScheduledExecutor();
```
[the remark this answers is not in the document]
ravibeta: Yes Andre

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,128 @@ public void testSegmentNotifications() throws Exception {
...
+        @Cleanup
+        ClientFactoryImpl clientFactory = new ClientFactoryImpl(SCOPE, controller, connectionFactory);
+        EventStreamWriter<String> writer1 = clientFactory.createEventWriter(streamName, new JavaSerializer<>(),
+                EventWriterConfig.builder().build());
+        List<String> routingKeys = new ArrayList<>();
+        routingKeys.add("0");
+        routingKeys.add("1");
+        routingKeys.add("2");
+        routingKeys.add("3");
+        writer1.writeEvent("0", "data").get();
```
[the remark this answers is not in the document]
ravibeta: Yes Andre

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Andre

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Andre

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Andre

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Andre it yields

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,126 @@ public void testSegmentNotifications() throws Exception {
...
+        Controller controller = controllerWrapper.getController();
+        controllerWrapper.getControllerService().createScope(SCOPE).get();
+        controller.createStream(SCOPE, streamName, config).get();
+        @Cleanup
+        ConnectionFactory connectionFactory = new ConnectionFactoryImpl(ClientConfig.builder()
+                .controllerURI(URI.create("tcp://localhost"))
+                .build());
+        @Cleanup
+        ClientFactoryImpl clientFactory = new ClientFactoryImpl(SCOPE, controller, connectionFactory);
+        EventStreamWriter<String> writer1 = clientFactory.createEventWriter(streamName, new JavaSerializer<>(),
```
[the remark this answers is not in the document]
ravibeta: Yes Tom

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,126 @@ public void testSegmentNotifications() throws Exception {
...
+            }
+        }, threadPoolExecutor)
+                .exceptionally(e -> {
+                    log.error("Failure: ", e);
+                    throw new IllegalStateException(e);
+                });
+        try {
+            scale(controller, SCOPE, streamName, 1, 1, threadPoolExecutor);
+        } catch (Exception e) {
+            log.error("Scaling Interrupted: ", e);
```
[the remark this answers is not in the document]
ravibeta: Yes Tom

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Tom, i did it only because it was required as parameter to controller api scaleStream

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,126 @@ public void testSegmentNotifications() throws Exception {
...
+                .throwingOn(RuntimeException.class)
+                .runAsync(() -> controller.getCurrentSegments(SCOPE, streamName)
+                        .thenAccept(streamSegments -> {
+                            try {
+                                Thread.sleep(0);
+                                EventStreamReader<String> reader1 = clientFactory.createReader("readerId", "reader", new JavaSerializer<>(),
+                                        ReaderConfig.builder().initialAllocationDelay(0).build());
+                                for (int i = 0; i < NUM_EVENTS; i++) {
+                                    EventRead<String> result =  reader1.readNextEvent(1000);
+                                    assertNotNull(result);
```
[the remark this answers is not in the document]
ravibeta: Yes Tom I did

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,126 @@ public void testSegmentNotifications() throws Exception {
...
+        } catch (Exception e) {
+            log.error("Scaling Interrupted: ", e);
+        }
+        Retry.withExpBackoff(10, 10, 100, 10000)
+                .retryingOn(IllegalStateException.class)
+                .throwingOn(RuntimeException.class)
+                .runAsync(() -> controller.getCurrentSegments(SCOPE, streamName)
+                        .thenAccept(streamSegments -> {
+                            try {
+                                Thread.sleep(0);
```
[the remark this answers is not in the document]
ravibeta: Yes Tom

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,126 @@ public void testSegmentNotifications() throws Exception {
...
+                                EventStreamReader<String> reader1 = clientFactory.createReader("readerId", "reader", new JavaSerializer<>(),
```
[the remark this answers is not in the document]
ravibeta: Yes Tom

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -189,19 +213,126 @@ public void testSegmentNotifications() throws Exception {
...
+                    log.error("Failure: ", e);
+                    throw new IllegalStateException(e);
+                });
+        try {
+            scale(controller, SCOPE, streamName, 1, 1, threadPoolExecutor);
+        } catch (Exception e) {
+            log.error("Scaling Interrupted: ", e);
+        }
+        Retry.withExpBackoff(10, 10, 100, 10000)
+                .retryingOn(IllegalStateException.class)
```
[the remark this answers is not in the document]
ravibeta: Yes Tom np

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Andre

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Andrei

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
ravibeta: Yes Andre, manually scaling the stream to NUM_SEGMENTS.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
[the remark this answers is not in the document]
tkaitchuck: Right below this, it is doing an update which implicitly fetches updates if it does not have all of the updates. So how is this line needed?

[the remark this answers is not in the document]
ravibeta: Yes Tom, the implicit fetch update is not always invoked. This adds the update.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -106,26 +125,31 @@ public void setUp() throws Exception {
 
     @After
     public void tearDown() throws Exception {
-        executor.shutdownNow();
+        executorService().shutdownNow();
```
andreipaduroiu, line 128: This is not needed. If you look at the base class, it is shut down for you using the proper shutdown sequence.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
andreipaduroiu: I was suggesting you write a comment explaining what this does so somebody reading the code later on can understand it

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -65,9 +83,12 @@
 import static org.junit.Assert.assertTrue;
 
 @Slf4j
-public class ReaderGroupNotificationTest {
+public class ReaderGroupNotificationTest extends LeakDetectorTestSuite {
 
     private static final String SCOPE = "test";
+    private static final int NUM_SEGMENTS = 10;
```
andreipaduroiu, line 89: For these 2 fields you use `NUM_` to indicate the count yet for the next one you use the `_COUNT` suffix. Please be consistent.

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -188,19 +206,93 @@ public void testSegmentNotifications() throws Exception {
...
+        routingKeys.add("0");
+        routingKeys.add("1");
+        routingKeys.add("2");
+        routingKeys.add("3");
+        writer1.writeEvent(routingKeys.get(0), "data").get();
+        writer1.writeEvent("0", "data").get();
+        long start = System.currentTimeMillis();
+        log.info("Segments before scale found as: {}", controller.getCurrentSegments(SCOPE, streamName).get());
+        while (System.currentTimeMillis() - start < Duration.ofSeconds(20).toMillis()) {
+            writer1.writeEvent(routingKeys.get(random.nextInt(3)), "data").get();
```
andreipaduroiu, line 237: You are continuously writing events for 20 seconds with no delay in between. How many events are you writing in this interval?  If it's too high, then would that be appropriate for an integration test?

ravibeta: Yes Andrei

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -65,9 +78,12 @@
 import static org.junit.Assert.assertTrue;
 
 @Slf4j
-public class ReaderGroupNotificationTest {
+public class ReaderGroupNotificationTest extends LeakDetectorTestSuite {
```
shrids, line 81: As I understand we can remove `LeakDetectorTestSuite`...

test/integration/src/test/java/io/pravega/test/integration/ReaderGroupNotificationTest.java
```diff
@@ -188,19 +206,93 @@ public void testSegmentNotifications() throws Exception {
...
+        writer1.writeEvent(routingKeys.get(0), "data").get();
+        writer1.writeEvent("0", "data").get();
+        long start = System.currentTimeMillis();
+        log.info("Segments before scale found as: {}", controller.getCurrentSegments(SCOPE, streamName).get());
+        while (System.currentTimeMillis() - start < Duration.ofSeconds(20).toMillis()) {
+            writer1.writeEvent(routingKeys.get(random.nextInt(3)), "data").get();
+        }
+
+        // scale
+        List<List<Segment>> segmentsList = scale(controller, SCOPE, streamName, 1, 1, executorService());
```
shrids, line 241: The scale method here performs a single manual scale operation by invoking   `controller.scaleStream(...)` and does not rely on auto-scaling by the controller. 
{ Even though the stream is configured with an auto-scaling policy(`ScalingPolicy.byEventRate(1, 2, 3)`) }
Hence this test is similar to the already existing test `io.pravega.test.integration.ReaderGroupNotificationTest#testSegmentNotifications`.

If the intent is to verify notifications with auto-scaling of the stream by the controller then I think it would be better if it is implemented as a system test. We should not invoke the manual scale API of the controller for such a test.

PR #4905, opened 2020-06-26
Issue 4900: Fix configuration related issues in templates and docs

### LINKED ISSUE #4900: New configuration documentation and naming needs further refinement
**Problem description**
There are couple of remaining issues in naming/documentation w.r.t. changes made to convert config names (in PR #4713). 

1. AWS and Swarm templates uses old config names. Also, the old `controller.zk.url` config doesn't work. 
2. `autoScale.authEnabled` has not been defined in [Pravega Security Configs](https://github.com/pravega/pravega/blob/master/documentation/src/docs/security/pravega-security-configurations.md) doc. 

**Problem location**

ravisharda:
**Change log description**  
Fix configuration related issues in example templates and docs.

**Purpose of the change**  
Fixes #4900 . 

**What the code does**  
* Replaces the use of old config names in templates, with new names. 
* Removes a phantom/unused `metricenableCSVReporter` config from the templates. 
* Modifies the Docker compose templates to use the same Zookeeper image version as that used in [Pravega Zookeeper Operator](https://github.com/pravega/zookeeper-operator/blob/c80add8305c25f2a85289db011a3c9494246ae99/docker/Dockerfile). 

**Note:** I've removed `controller.zk.url` config from everywhere. That will leave the AWS template backward incompatible in this respect. But, I believe it's better to do that as that config is not defined in Controller's config class - `Config.java`. 

**How to verify it**  


docker/compose/docker-compose-extendeds3.yml
```diff
@@ -10,7 +10,7 @@
 version: '2'
 services:
   zookeeper:
-    image: zookeeper:3.5.4-beta
+    image: zookeeper:3.6.1
```
RaulGracia, line 13: I thought that we were on 3.5.5, but maybe my understanding is outdated already.

deployment/aws/installer/entry_point_template.yml
```diff
@@ -32,8 +32,7 @@
     JAVA_OPTS: |
       -Dcontroller.service.rpc.listener.port=9090
       -Dcontroller.service.rest.listener.port=10080
-      -Dcontroller.zk.url=ZKNODE:2181
-      -Dconfig.controller.metricenableCSVReporter=false
+      -Dcontroller.zk.connect.uri=ZKNODE:2181
```
RaulGracia, line 35: Thanks, the CSV Reported was removed as an option since the introduction of micrometer, but it looks like there were still leftovers from that work.

docker/compose/docker-compose-extendeds3.yml
ravisharda: I checked this and noticed it was `3.6.1` now. 

https://github.com/pravega/zookeeper-operator/blob/db0e7467853123ed75053633d859b246dcefaf05/docker/Dockerfile#L17

andreipaduroiu: `gradle.properties` still says `3.5.5`.

ravisharda: Upgrading Zookeper client will require code changes in Pravega (in the class below).  In this PR, I've only changed the Zookeeper server version in templates to reflect the same version used in Zookeeper operator. 
```
pravega/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/ZooKeeperServiceRunner.java
```
It is not the intent of this PR to upgrade the Zookeeper client and I think we should do it separately.

RaulGracia:
@ravisharda something is wrong with codecov check, maybe you need to push another commit to make it work.

ravisharda:
Started system test run for this PR at pravega-branch-builder/2570/.

ravisharda:
test-pravega-continuous/2967/ had passed for this branch on 6th July, FYI.

ravisharda:
@andreipaduroiu ping

PR #4906, opened 2020-06-26
Issue 4893: Add Scopes Metrics

### LINKED ISSUE #4893: Add metrics for scope creation and deletion
**Problem description**
We have metrics for all operations on Streams, but not for Scopes. It would be great for completeness to also provide metrics about the creation and deletion of scopes, as well as any failure in these operations.

**Problem location**
Controller metrics.

Add the metrics for scopes.

RaulGracia:
**Change log description**  
Adds metrics for created and deleted scopes.

**Purpose of the change**  
Fixes #4893.

**What the code does**  
This PR extends `StreamMetrics` class to also take care of reporting a couple of new metrics related to scopes creation and deletion.

**How to verify it**  
New test checking that the new metrics (and some former ones) are reported.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
+++ b/controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
@@ -308,6 +308,8 @@ public final class StreamMetrics extends AbstractControllerMetrics {
             INSTANCE.get().sealStreamLatency.close();
             INSTANCE.get().updateStreamLatency.close();
             INSTANCE.get().truncateStreamLatency.close();
+            INSTANCE.get().createScopeLatency.close();
+            INSTANCE.get().deleteScopeLatency.close();
             INSTANCE.set(null);
         }
     }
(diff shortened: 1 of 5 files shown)


controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
```diff
@@ -93,6 +104,16 @@ public void createStream(String scope, String streamName, int minNumSegments, Du
         createStreamLatency.reportSuccessValue(latency.toMillis());
     }
 
+    /**
+     * This method increments the global counter of Scope creations and reports the latency of the operation.
+     *
+     * @param latency           Latency of the createStream operation.
+     */
+    public void createScope(Duration latency) {
+        DYNAMIC_LOGGER.incCounterValue(CREATE_SCOPE, 1);
```
shiveshr, line 113: this is counting how many scopes this controller instance has created so far..

RaulGracia: The values from multiple Controllers will be aggregated at the InfluxDB level when collecting the metrics.

PR #4911, opened 2020-06-30
Issue 4910: Added new metrics to report the distribution of client appends and reads

### LINKED ISSUE #4910: Create new metric to keep track of the size of client appends
**Problem description**
Currently, we know the incoming aggregated bytes and number appends from clients, but we do not have information about the size of these appends. This may be a useful information to have for various reasons, such as to understand better the size of appends and the behavior of batching at the client side. 

**Problem location**
`AppendProcessor` metrics.

Add a new metric that captures the size distribution of appends.

RaulGracia:
**Change log description**  
Adds two new metrics to report the size distribution of appends and reads.

**Purpose of the change**  
Fixes #4910.

**What the code does**  
This PR adds a couple of `OpStatsLogger` metrics in `SegmentStatsRecorderImpl` to report the size distribution of appends and reads. This information can give us important insights about the kind of writes and reads that Pravega is serving.

**How to verify it**  
All tests should be working as before.

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
@@ -33,9 +33,11 @@ import lombok.Getter;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
+import static io.pravega.shared.MetricsNames.SEGMENT_APPEND_SIZE;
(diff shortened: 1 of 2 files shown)


andreipaduroiu:
How is this different than `SEGMENT_WRITE_BYTES`?

RaulGracia:
@andreipaduroiu `SEGMENT_WRITE_BYTES` is a `DynamicLogger` so it only publishes counters of the amount of bytes and appends. What we pursue here is something similar to what we already have for append latency, but for the sizes of appends (and reads, by the way); that is, having a distribution (with percentiles) of sizes of appends issued by clients, so we can understand various things, including the behavior of client batching.

PR #4916, opened 2020-07-02
Issue 4915: Add timeout to add requests from Controller to Segment Store

### LINKED ISSUE #4915: Add timeout to all calls to segment store from controller
**Problem description**
We have encountered several issues where segment store either took very long or failed to reply while the connection was not dropped. This keeps the controller and in turn client requests being stuck either for very long or indefinitely. We should add timeout to the calls into segment store and fail the call with connection failed exception so that the retries can be performed wherever applicable. 

Make sure that the timeout is configurable such that we can set very large timeouts in our system tests so as to not mask any issues. 

**Problem location**
SegmentHelper.java

Add a timeout to calls into segment store

shiveshr:

**Change log description**  
Adds a timeout to requests being made to segment store using segment helper. 

**Purpose of the change**  
Fixes #4915 

**What the code does**  
Adds a timeout to the future for the call to segment store. when the future is timedout, the request is failed with connection failed exception which is a retryable exception and all code paths handle it. 

**How to verify it**  
Unit test added.


controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -635,17 +651,34 @@ private void closeConnection(Reply reply, RawClient client) {
     }
 
     private <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(RawClient connection, long requestId, T request) {
-        return connection.sendRequest(requestId, request)
+        CompletableFuture<Reply> future = Futures.futureWithTimeout(timeout.get(), executorService);
+        AtomicReference<CompletableFuture<Reply>> f = new AtomicReference<>();
+        Futures.completeAfter(() -> {
```
shrids, line 656: Can we create a method on lines of the below pattern? This would eliminate the need for extra code.
 ```
 public static <T> CompletableFuture<T> futureWithTimeout(CompletableFuture<T> cf, Duration timeout, ScheduledExecutorService executorService) {
      
        ScheduledFuture<Boolean> sf = executorService.schedule(() -> cf.completeExceptionally(new TimeoutException(tag)), timeout.toMillis(), TimeUnit.MILLISECONDS);
        result.whenComplete((r, ex) -> sf.cancel(true));
        return cf;
    }
```

Comment (shrids) [reply controller/src/main/java/io/pravega/controller/server/SegmentHelper.java:656]: Interesting JAVA9 APIS support this https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/CompletableFuture.html#completeOnTimeout-T-long-java.util.concurrent.TimeUnit-

File: common/src/main/java/io/pravega/common/concurrent/Futures.java [hunk 2]
```diff
@@ -543,6 +543,32 @@
...
+     *
+     * @param futureSupplier  Supplier of the future. 
+     * @param timeout         The timeout for the future.
+     * @param executorService An ExecutorService that will be used to invoke the timeout on.
+     * @param <T>             The Type argument for the CompletableFuture to create.
+     * @return The result.
+     */
+    public static <T> CompletableFuture<T> futureWithTimeout(Supplier<CompletableFuture<T>> futureSupplier, 
+                                                             Duration timeout, ScheduledExecutorService executorService) {
+        CompletableFuture<T> result = new CompletableFuture<>();
```
shrids, line 559: Is there a reason why we need to create a new CompletableFuture object? Can we avoid it ?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -635,17 +651,27 @@ private void closeConnection(Reply reply, RawClient client) {
...
+        return future
                 .exceptionally(e -> {
-                    Throwable unwrap = Exceptions.unwrap(e);
-                    if (unwrap instanceof ConnectionFailedException || unwrap instanceof ConnectionClosedException) {
-                        log.warn(requestId, "Connection dropped");
+                    if (Exceptions.unwrap(e) instanceof TimeoutException) {
+                        log.warn(requestId, "Request timedout.");
                         throw new WireCommandFailedException(request.getType(), WireCommandFailedException.Reason.ConnectionFailed);
-                    } else if (unwrap instanceof AuthenticationException) {
-                        log.warn(requestId, "Authentication Exception");
```
shrids, line 645: we should retain the auth failure logs.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -635,17 +651,27 @@ private void closeConnection(Reply reply, RawClient client) {
     }
 
     private <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(RawClient connection, long requestId, T request) {
-        return connection.sendRequest(requestId, request)
+        CompletableFuture<Reply> future = Futures.futureWithTimeout(
+                () -> connection.sendRequest(requestId, request)
+                                .exceptionally(e -> {
```
shrids, line 656: nit: Can we have all the exception handling in one block? it will reduce the creation of a CompletionStage and all the error handling will be done in one block (e.g: right before returning the future below)

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: authentication exception cannot come here ever. this is an unreachable code. 

this is because auth failure will always come in as a response from segment store. 
which is handled and translated in `handleReply` method. 

I can retain it but its a code that doesnt get covered and decreases the coverage.

common/src/main/java/io/pravega/common/concurrent/Futures.java
shiveshr: i kept as separate future so that we can issue a cancel on the future received from futureSupplier. 
so at least any downstream processing that was chained may get cancelled.

shiveshr: I have made the change to not create a new future. I have also added a method in futures to add timeout to an existing future.

shrids: Thanks Shivesh.
cancel() does not cancel the computation as per the java doc.
``` 
/**
     * If not already completed, completes this CompletableFuture with
     * a {@link CancellationException}. Dependent CompletableFutures
     * that have not already completed will also complete
     * exceptionally, with a {@link CompletionException} caused by
     * this {@code CancellationException}.
     *
     * @param mayInterruptIfRunning this value has no effect in this
     * implementation because interrupts are not used to control
     * processing.
     *
     * @return {@code true} if this task is now cancelled
     */
    public boolean cancel(boolean mayInterruptIfRunning)
```

common/src/main/java/io/pravega/common/concurrent/Futures.java
```diff
@@ -567,6 +567,47 @@
...
+     * Creates a new CompletableFuture that either holds the result of future from the futureSupplier
+     * or will timeout after the given amount of time.
+     *
+     * @param futureSupplier  Supplier of the future. 
+     * @param timeout         The timeout for the future.
+     * @param executorService An ExecutorService that will be used to invoke the timeout on.
+     * @param <T>             The Type argument for the CompletableFuture to create.
+     * @return A CompletableFuture which is either completed within given timebound or failed with timeout exception.
+     */
+    public static <T> CompletableFuture<T> futureWithTimeout(Supplier<CompletableFuture<T>> futureSupplier,
```
andreipaduroiu, line 581: please rewrite the `futureWithTimeout` above to make use of this method.
Ex:
`return futureWithTimeout(CompletableFuture::new, timeout, executorService)`

common/src/main/java/io/pravega/common/concurrent/Futures.java
```diff
@@ -567,6 +567,47 @@
...
+     * Add a timeout to the CompletableFuture. If the future does not complete within the specified timeout,
+     * this would complete the future with timeout exception. 
+     *
+     * @param future          Completable future to add timeout to. 
+     * @param timeout         The timeout for the future.
+     * @param executorService An ExecutorService that will be used to invoke the timeout on.
+     * @param <T>             The Type argument for the CompletableFuture to create.
+     */
+    public static <T> void addTimeout(CompletableFuture<T> future, Duration timeout, ScheduledExecutorService executorService) {
+        ScheduledFuture<Boolean> sf = executorService.schedule(() -> future.completeExceptionally(new TimeoutException()), timeout.toMillis(), TimeUnit.MILLISECONDS);
```
andreipaduroiu, line 602: This can be rewritten as
```
futureWithTimeout(() -> future, timeout, executorService)
```

common/src/test/java/io/pravega/common/concurrent/FuturesTests.java
```diff
@@ -21,23 +24,40 @@
...
+import org.junit.After;
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 
 /**
  * Unit tests for the Futures class.
  */
 public class FuturesTests {
+    private ScheduledExecutorService executor;
```
andreipaduroiu, line 49: Please change this class to extend from `ThreadPooledTestSuite` and override `getThreadPoolSize` to return 1.

That base class will take care of your executor.

While you're at it, there is at least one other test in this class that makes its own executor (and never shuts it down!). Please change that too to use `executorService()` instead.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -605,7 +622,9 @@ private void closeConnection(Reply reply, RawClient client) {
     }
 
     private <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(RawClient connection, long requestId, T request) {
-        return connection.sendRequest(requestId, request)
+        CompletableFuture<Reply> future = connection.sendRequest(requestId, request);
+        Futures.addTimeout(future, timeout.get(), executorService);
```
andreipaduroiu, line 626: `Futures.futureWithTimeout(() -> connection.sendRequest(...), ...)` ?

That will preclude the need for `Futures.addTimeout`

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -614,6 +633,9 @@ private void closeConnection(Reply reply, RawClient client) {
                     } else if (unwrap instanceof AuthenticationException) {
                         log.warn(requestId, "Authentication Exception");
                         throw new WireCommandFailedException(request.getType(), WireCommandFailedException.Reason.AuthFailed);
+                    } else if (Exceptions.unwrap(e) instanceof TimeoutException) {
```
andreipaduroiu, line 636: `unwrap` contains the unwrapped `e`. Use that.

controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -128,6 +128,9 @@
     public static final Property<String> PROPERTY_TLS_ENABLED_FOR_SEGMENT_STORE = Property.named(
             "segmentstore.connect.channel.tls", "", "auth.segmentStoreTlsEnabled");
 
+    public static final Property<Integer> PROPERTY_SEGMENT_STORE_REQUEST_TIMEOUT_MINUTES = Property.named(
+            "segmentstore.connect.channel.timeoutMinutes", 2, "");
```
andreipaduroiu, line 132: I would suggest you express this in seconds. We may want to add 30s timeouts. 2 minutes seems too long.

controller/src/test/java/io/pravega/controller/server/SegmentHelperTest.java
```diff
@@ -65,19 +70,29 @@
     private final byte[] value = "v".getBytes();
     private final ByteBuf token1 = wrappedBuffer(new byte[]{0x01});
     private final ByteBuf token2 = wrappedBuffer(new byte[]{0x02});
-
+    private ScheduledExecutorService executorService;
```
andreipaduroiu, line 73: Use `ThreadPooledTestSuite`

PR #4921, opened 2020-07-06
Issue 4920: From noteTime api, throw exception rather than set the response code

### LINKED ISSUE #4920: NoteTime API failure is not retried.
**Problem description**

The log snippet observed. 

```
Client log
2020-07-01 15:26:04,488 [grpc-default-executor-938] WARN  io.pravega.client.stream.impl.ControllerImpl                  - Writer 2c339ca2-248e-400f-bac8-5fdd6b4ea45f-0 failed to note time because: INTERNAL_ERROR time was: 298460542

Controller log:
2020-07-01 15:25:55,344 457945863 [grpc-default-executor-2219] INFO  i.p.c.s.r.g.v.ControllerServiceImpl - noteWriterMark called for stream long-0/long-0, writer=2c339ca2-248e-400f-bac8-5fdd6b4ea45f-0 time=298460542

// this might be the reason why it failed.
2020-07-01 15:25:55,349 457945868 [epollEventLoopGroup-2-1] INFO  i.p.controller.server.SegmentHelper - Closing connection as a result of receiving: Conditional table update failed since the key version is incorrect : _system/_tables/long-0/long-0/writersPositions.#.94b95844-b766-4d93-a78a-988a04e5aff1
{code}

```
 The controller client does not retry in case of an internal error ref: https://github.com/pravega/pravega/blob/b81cf9e69bb53b0e0860f703f3a47a371ad80cc6/client/src/main/java/io/pravega/client/stream/impl/ControllerImpl.java#L1115-L1117

**Problem location**
Controller error handling.

Fix the issue , we can ensure the automatic retry mechanism on the controller client is invoked during error.

shiveshr:

**Change log description**  
Throw exception from noteTime api rather than sending WriterTimeResponse.InternalError.  

**Purpose of the change**  
Fixes #4920 

**What the code does**  
noteTime api currently responds with an internal error response code for any error. Which is not translated into a retryable exception on controller client.

**How to verify it**  
Unit test added.


controller/src/test/java/io/pravega/controller/server/v1/InMemoryControllerServiceImplTest.java
```diff
@@ -85,9 +86,10 @@ public void setup() throws Exception {
 
         Cluster mockCluster = mock(Cluster.class);
         when(mockCluster.getClusterMembers()).thenReturn(Collections.singleton(new Host("localhost", 9090, null)));
-        controllerService = new ControllerServiceImpl(
```
eolivelli, line 88: Why can't you *spy* directly this *controllerService* variable?

shiveshr: controllerService which is an instance of `ControllerServiceImpl` is what we are verifying in this test.

controller/src/test/java/io/pravega/controller/server/v1/InMemoryControllerServiceImplTest.java
```diff
@@ -85,12 +83,11 @@ public void setup() throws Exception {
...
-                new ControllerService(streamStore, StreamStoreFactory.createInMemoryBucketStore(), streamMetadataTasks, streamTransactionMetadataTasks,
-                                      SegmentHelperMock.getSegmentHelperMock(), executorService, mockCluster), GrpcAuthHelper.getDisabledAuthHelper(), requestTracker, true, 2);
+        return new ControllerService(streamStore, StreamStoreFactory.createInMemoryBucketStore(), streamMetadataTasks, streamTransactionMetadataTasks,
+                SegmentHelperMock.getSegmentHelperMock(), executorService, mockCluster);
     }
 
-    @Override
```
shrids, line 93: Override should be fine here since io.pravega.controller.server.v1.ControllerServiceImplTest#tearDown method already has @After annotation right ? https://github.com/pravega/pravega/blob/b81cf9e69bb53b0e0860f703f3a47a371ad80cc6/controller/src/test/java/io/pravega/controller/server/v1/ControllerServiceImplTest.java#L86-L87

controller/src/test/java/io/pravega/controller/server/v1/InMemoryControllerServiceImplTest.java
eolivelli: I am not sure this is the right way. But unfortunately I don't have much time to suggest another approach. 
So I am fine with keeping it.

Maybe one day we could introduce PowerMock, that allows you to rewrite entire methods and constructors without introducing this intermediate objects.

Now we are introducing this layer between the streamStore and the ControllerService, I don't have enough context, so I don't know if it may introduce some sort of difference from the behavior of the tests  and the execution at runtime in production (the risk is to not test the right thing).

controller/src/test/java/io/pravega/controller/server/v1/InMemoryControllerServiceImplTest.java
shiveshr: its not overridden because its not defined in the base class.

PR #4929, opened 2020-07-08
Issue 4919: Use delomboked source code in javadoc distribution

### LINKED ISSUE #4919: Lombok obfuscates Javadocs, @Data classes don't document methods
**Problem description**
Some Javadocs are devoid of class method information due to use of Lombok `@Data` annotation.

**Problem location**
http://pravega.io/docs/v0.7.1/javadoc/clients/io/pravega/client/admin/StreamInfo.html
vs.
https://github.com/pravega/pravega/blob/v0.7.1/client/src/main/java/io/pravega/client/admin/StreamInfo.java

I'm not sure... Lombok `@Data` annotation is very helpful for generating boilerplate code that doesn't necessarily need to be reified in the source itself. I don't know if there's a general solution here, but there ought to be. I've seen this same problem with Swagger & auto-generating OpenAPI specs off of JAX-RS resources, where Swagger can't lift the Javadocs off of annotated methods for inclusion in the OpenAPI spec. In this case, Lombok can't lift the Javadoc off of wrapped member variables and stage the generated source for use by the Javadoc generator.

medvedevigorek:
Signed-off-by: Igor Medvedev <medvedevigorek@gmail.com>

**Change log description**  
Javadoc task for all projects operates on delomboked source code and produces the complete javadoc including methods generated by Lombok. However the same is not true for javadoc distribution, it works on the source code as-is that includes Lombok annotations and produces incomplete javadoc bundle. 

The change is to ensure the javadoc distribution uses delomboked source code. 

**Purpose of the change**  
Fixes #4919 

**What the code does**  
Javadoc distribution task is updated to use delomboked source code for javadoc generation.

**How to verify it**  
Running `javadocDistZip` or `javadocDistTar` tasks would produce a javadoc bundle which would have method generated by Lombok included as well

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1021,12 +1021,14 @@ task javadocs(type: Javadoc) {
     ext.projects = [':client']
     title = "Pravega API"
     destinationDir = file("${buildDir}/javadocs")
-    source = files(projects.collect {
-        project(it).sourceSets.main.allJava
-    })
-    classpath = files(projects.collect {
-        project(it).sourceSets.main.output + project(it).sourceSets.main.compileClasspath
-    })
+
+    projects.collect {
+        dependsOn project(it).delombok
+
+        source += project(it).delombok.outputDir
+        classpath += project(it).sourceSets.main.output + project(it).sourceSets.main.compileClasspath
+    }
+
     failOnError = true
     exclude "**/impl/**"
     options.addBooleanOption("Xdoclint:all,-reference", true)


andreipaduroiu:
Did we verify that this isn't going to add unwanted Javadoc comments in the Javadoc that we publish with our Client?

medvedevigorek:
There are no additional Javadoc comments added besides those which were missing before. For example, **Builders**  are now also include in the final Javadoc bundle.

PR #4935, opened 2020-07-09
Issue 4933: (Key-Value Tables) Enforcing Key and Value serialization limits

### LINKED ISSUE #4933: Key-Value Tables: Key and Value serialization limits
**Problem description**
Ensure the serialization limits are enforced for all `KeyValueTable` APIs.
Ensure that no operation may exceed the maximum WireCommand length.
Ensure that multi-key retrievals from a single segment may not exceed the WireCommand length (on the reply side).

**Problem location**
Key Value Tables. Both on the client and server-side.

Define and document proper limits. Enforce them.

andreipaduroiu:
**Change log description**  
- Enforcing the maximum Key Length, maximum Value Length and maximum Key Family Length for all KeyValueTable operations.
    - Max Key Length is 4KB, Max Value Length is 1MB-1KB, Max Key Family Length is 2KB.
- Ensuring no KeyValueTable wire command exceeds the maximum wire command size.

**Purpose of the change**  
Fixes #4933.

**What the code does**  
- KeyValueTable
    - Exposed constants that define the maximum key length and maximum value length
    - Exposed a constant that defines the maximum number of items that can be updated, removed or retrieved at any given time.
    - Exposed a constant that defines the maximum serialization size for any multi-key operations (even though each key and value are subject to their own constraints, we enforce that a batch of these will not exceed pre-set limits).
        - This value is required so that we do not exceed the maximum Wire Command size and inherently the Netty frame size.
- TableSegment
    - If a Get request exceeds a predefined number of keys (calculated to around 15, which ensures that no response exceeds the max wire command size), then that get request is split into non-overlapping sub-requests (of at most 15 keys) and issued in parallel to the Segment Store.
    - The requests are "throttled" on the client side by ensuring no more than 5 concurrent calls per request can execute at any given time. 
    - The response is compiled and returned to the upstream code as if no split happened.
- OrderedProcessor
    - Renamed from OrderedItemProcessor and moved to a new package (where it belongs)
    - Since this was no longer used by anything else, I repurposed it to the needs of the TableSegmentImpl, by simplifying its behavior.
    - It now acts as a simple "capacity limiter" that only allows a certain number of tasks to execute concurrently and queues the rest.

**How to verify it**  
All unit and integration tests must pass. New tests have been added to verify these limits and new TableSegmentImpl.get behavior.

### DIFF
--- a/client/src/main/java/io/pravega/client/tables/impl/KeyFamilySerializer.java
+++ b/client/src/main/java/io/pravega/client/tables/impl/KeyFamilySerializer.java
@@ -24,8 +24,17 @@ import org.apache.commons.lang3.SerializationException;
  * Serializer for Key-Value Table Key Families.
  */
 class KeyFamilySerializer {
+    /**
+     * Maximum length for a Key Family serialization using {@link #ENCODING}.
+     * If this value is changed, care must be taken such that the sum of it, {@link #PREFIX_LENGTH} and
+     * {@link TableSegment#MAXIMUM_KEY_LENGTH} do not exceed the server-side limits.
+     */
     @VisibleForTesting
-    static final int MAX_KEY_FAMILY_LENGTH = 1024; // It can't be longer than TableSegment.MAX_KEY_LENGTH
+    static final int MAX_KEY_FAMILY_LENGTH = 2096;
+    /**
+     * A prefix added to every Key Family serialization. This is encoded in the key itself. See
+     * {@link #MAX_KEY_FAMILY_LENGTH} for constraints.
+     */
     @VisibleForTesting
     static final int PREFIX_LENGTH = 2; // [0, 65535]. First 6 bits will be empty.
     @VisibleForTesting
--- a/client/src/main/java/io/pravega/client/tables/impl/KeyFamilySerializer.java
+++ b/client/src/main/java/io/pravega/client/tables/impl/KeyFamilySerializer.java
@@ -30,7 +30,7 @@ class KeyFamilySerializer {
      * {@link TableSegment#MAXIMUM_KEY_LENGTH} do not exceed the server-side limits.
      */
     @VisibleForTesting
-    static final int MAX_KEY_FAMILY_LENGTH = 2096;
+    static final int MAX_KEY_FAMILY_LENGTH = 2048;
     /**
      * A prefix added to every Key Family serialization. This is encoded in the key itself. See
      * {@link #MAX_KEY_FAMILY_LENGTH} for constraints.
--- a/client/src/main/java/io/pravega/client/tables/impl/TableSegmentImpl.java
+++ b/client/src/main/java/io/pravega/client/tables/impl/TableSegmentImpl.java
@@ -159,7 +159,6 @@ class TableSegmentImpl implements TableSegment {
                 final int sliceStart = index;
                 final int sliceLength = Math.min(MAX_GET_KEY_BATCH_SIZE, wireKeys.size() - sliceStart);
                 futures.add(processor.execute(() -> fetchSlice(resultBuilder.slice(sliceStart, sliceStart + sliceLength))));
-                //futures.add(fetchSlice(resultBuilder.slice(index, index + sliceLength)));
                 index += sliceLength;
             }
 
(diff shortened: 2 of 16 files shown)


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -710,8 +717,11 @@ public void readTable(final WireCommands.ReadTable readTable) {
         val timer = new Timer();
         tableStore.get(segment, keys, TIMEOUT)
                 .thenAccept(values -> {
-                    connection.send(new WireCommands.TableRead(readTable.getRequestId(), segment, getTableEntriesCommand(keys, values)));
-                    this.tableStatsRecorder.getKeys(readTable.getSegment(), keys.size(), timer.getElapsed());
+                    // NOTE: getTableEntriesCommand will truncate the result to prevent it from exceeding MAX_TABLE_RESPONSE_SIZE.
+                    // In such a situation, not all requested items will be returned to the caller; only the head of the list will.
+                    WireCommands.TableEntries response = getTableEntriesCommand(readTable.getRequestId(), keys, values);
```
tkaitchuck, line 722: This seems particularly inefficient. Supposing there are N items to be retrieved and a fixed number fit in a batch this is O(n^2) because the first call gets all N and throws all but the first batch away. The second fetches N-batch and throws all but the next batch away etc.
We should figure out a way to pass down a limit to the get call, even if it is not exact. It's better to reply with more smaller messages which are efficient than maximizing our wirecommand size out in the reply.

client/src/main/java/io/pravega/client/tables/impl/KeyFamilySerializer.java
```diff
@@ -24,8 +24,17 @@
...
  */
 class KeyFamilySerializer {
+    /**
+     * Maximum length for a Key Family serialization using {@link #ENCODING}.
+     * If this value is changed, care must be taken such that the sum of it, {@link #PREFIX_LENGTH} and
+     * {@link TableSegment#MAXIMUM_KEY_LENGTH} do not exceed the server-side limits.
+     */
     @VisibleForTesting
-    static final int MAX_KEY_FAMILY_LENGTH = 1024; // It can't be longer than TableSegment.MAX_KEY_LENGTH
+    static final int MAX_KEY_FAMILY_LENGTH = 2096;
```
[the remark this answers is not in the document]
andreipaduroiu: Duh! Fixed.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
andreipaduroiu: I agree with this. However I am not sure how easy it is to do this with the current `RawClient` that expects a single response for a request.

I filed #4936 to research a way to handle these situations. This may also help iterators too (i.e., to avoid a flurry of request-replies for each iterator iteration).

andreipaduroiu: I reworked this part. Check out the updated description for details.

client/src/main/java/io/pravega/client/tables/impl/TableSegmentImpl.java
```diff
@@ -138,11 +145,37 @@ public void close() {
...
+        } else {
+            // The request has to be split into multiple calls and then combined.
+            val processor = new OrderedProcessor<Void>(MAX_GET_CONCURRENT_REQUESTS, this.connectionFactory.getInternalExecutor());
+            val futures = new ArrayList<CompletableFuture<Void>>();
+            int index = 0;
+            while (index < wireKeys.size()) {
+                final int sliceStart = index;
+                final int sliceLength = Math.min(MAX_GET_KEY_BATCH_SIZE, wireKeys.size() - sliceStart);
+                futures.add(processor.execute(() -> fetchSlice(resultBuilder.slice(sliceStart, sliceStart + sliceLength))));
+                //futures.add(fetchSlice(resultBuilder.slice(index, index + sliceLength)));
```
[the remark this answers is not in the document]
andreipaduroiu: Done.

PR #4939, opened 2020-07-10
Issue 4162: Add TLS mode for system tests

### LINKED ISSUE #4162: Enable running of System Tests with TLS Enabled
**Problem description**
Add an additional TLS enabled mode to System test, so all of them can be executed with only TLS enabled or TLS+Auth enabled on a Distributed Pravega Cluster.

Sinilar support for running System Tests with Authentication enabled was added by PR: 
https://github.com/pravega/pravega/pull/3662

For details on enabling TLS/Auth, see:
https://github.com/pravega/pravega/blob/master/documentation/src/docs/security/pravega-security-configurations.md
https://gist.github.com/ravisharda/5ef97136e06615776ffd95a03b669c3f

ravibeta:
**Change log description**  
* Adds option to the framework to run system tests with TLS enabled. 

**Purpose of the change**  
Fixes #4162

**What the code does** 
System tests use the Pravega deployment in distributed mode which provides the controller URI as `tcp://controller-service-ip:9090` for gRPC. This change enables the system test to support a deployment with TLS mode on so that the controller URI can be changed to `tls://controller-service-ip:9090`. Later, more specific TLS test can be added and this change only adds support in the library.  All tests can take advantage of this mode by pointing to one of the two controller URI and specifying the boolean parameter when creating the service.

Both insecure and secure form of Pravega are deployed on K8s by the K8s pravega-operator, which is invoked by the system test framework. TLS mode is added via a configuration option that is passed to the operator as an environment variable in the configuration option.

**How to verify it**  
This change is run through the build and test routine.

The build failed on a single test but that was not due to this change.

Update:
[TestingTLS.docx](https://github.com/pravega/pravega/files/5035281/TestingTLS.docx)


ravisharda:
@ravibeta have you verified this to be working in system test environment? 

Asking that because when someone last looked at the issue, it was realized that we'd also need to support turning off Controller hostname verification for the same. Currently, it can be turned off only for client-to-segment-store communications, but not for client-to-controller communications. I don't see that part of the change in the PR, and that is why asking if this works in system test environment.

andreipaduroiu:
Please fix your build. You have Spotbugs failures.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -89,23 +89,26 @@ public static Service createBookkeeperService(final URI zkUri) {
         }
     }
 
-    public static Service createPravegaControllerService(final URI zkUri, String serviceName) {
+    public static Service createPravegaControllerService(final URI zkUri, String serviceName, Boolean secure) {
```
andreipaduroiu, line 92: `boolean`, not `Boolean`

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -106,13 +106,17 @@ public String getID() {
     }
 
     CompletableFuture<Object> deployPravegaUsingOperator(final URI zkUri, int controllerCount, int segmentStoreCount, int bookieCount, ImmutableMap<String, String> props) {
+        return deployPravegaUsingOperator(zkUri, controllerCount, segmentStoreCount, bookieCount, props, false);
+    }
+
+    CompletableFuture<Object> deployPravegaUsingOperator(final URI zkUri, int controllerCount, int segmentStoreCount, int bookieCount, ImmutableMap<String, String> props, Boolean secure) {
```
andreipaduroiu, line 112: and here

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -388,6 +392,10 @@ private V1beta1ClusterRoleBinding getPravegaOperatorClusterRoleBinding() {
     }
 
     private V1Deployment getPravegaOperatorDeployment() {
+         return getPravegaOperatorDeployment(false);
+    }
+
+    private V1Deployment getPravegaOperatorDeployment(Boolean secure) {
```
andreipaduroiu, line 398: and here

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/PravegaControllerK8sService.java
```diff
@@ -27,12 +27,20 @@
 
     private final URI zkUri;
     private final ImmutableMap<String, String> properties;
+    private final Boolean secure;
```
andreipaduroiu, line 30: and here

test/system/src/test/java/io/pravega/test/system/AbstractScaleTests.java
```diff
@@ -56,8 +64,8 @@ private ControllerImpl createController(final ClientConfig clientConfig) {
                                   getConnectionFactory().getInternalExecutor());
     }
 
-    private URI createControllerURI() {
-        Service conService = Utils.createPravegaControllerService(null);
+    private URI createControllerURI(Boolean secure) {
```
andreipaduroiu, line 67: and here

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -89,23 +89,26 @@ public static Service createBookkeeperService(final URI zkUri) {
         }
     }
 
-    public static Service createPravegaControllerService(final URI zkUri, String serviceName) {
+    public static Service createPravegaControllerService(final URI zkUri, String serviceName, boolean secure) {
         switch (EXECUTOR_TYPE) {
```
ravisharda, line 93: `isSecure` or `enableSecurity`, etc.

Also, curious about this: In this PR you seem to be adding TLS only (as auth is already there). So, this variable and the same variable in other classes in this PR might be referring to TLS only. So, wouldn't it make sense to just call it `enableTls` or the like. Otherwise, people will be misled to believe its setting `Auth` too.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -89,23 +89,26 @@ public static Service createBookkeeperService(final URI zkUri) {
...
             case REMOTE_SEQUENTIAL:
                 return new PravegaControllerService(serviceName, zkUri);
             case DOCKER:
                 return new PravegaControllerDockerService(serviceName, zkUri);
             case KUBERNETES:
             default:
-                return new PravegaControllerK8sService(serviceName, zkUri, getPravegaProperties());
+                return new PravegaControllerK8sService(serviceName, zkUri, getPravegaProperties(), secure);
```
ravisharda, line 100: Same here

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
ravisharda, line 93: Does it make sense to log the inputs here, for making it easier to debug system test issues? That is, if its not already logged in the call hierarchy.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -106,13 +106,17 @@ public String getID() {
...
+
+    CompletableFuture<Object> deployPravegaUsingOperator(final URI zkUri, int controllerCount, int segmentStoreCount, int bookieCount, ImmutableMap<String, String> props, boolean secure) {
     return k8sClient.createCRD(getPravegaCRD())
                         .thenCompose(v -> k8sClient.createRole(NAMESPACE, getPravegaOperatorRole()))
                         .thenCompose(v -> k8sClient.createClusterRole(getPravegaOperatorClusterRole()))
                         .thenCompose(v -> k8sClient.createRoleBinding(NAMESPACE, getPravegaOperatorRoleBinding()))
                         .thenCompose(v -> k8sClient.createClusterRoleBinding(getPravegaOperatorClusterRoleBinding()))
                         //deploy pravega operator.
-                        .thenCompose(v -> k8sClient.createDeployment(NAMESPACE, getPravegaOperatorDeployment()))
+                        .thenCompose(v -> k8sClient.createDeployment(NAMESPACE, getPravegaOperatorDeployment(secure)))
```
ravisharda, line 119: Same as earlier

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -388,6 +392,10 @@ private V1beta1ClusterRoleBinding getPravegaOperatorClusterRoleBinding() {
     }
 
     private V1Deployment getPravegaOperatorDeployment() {
+         return getPravegaOperatorDeployment(false);
+    }
+
+    private V1Deployment getPravegaOperatorDeployment(boolean secure) {
```
ravisharda, line 398: Same here

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/PravegaControllerK8sService.java
```diff
@@ -27,12 +27,20 @@
 
     private final URI zkUri;
     private final ImmutableMap<String, String> properties;
+    private final boolean secure;
```
ravisharda, line 30: Same here

ravisharda:
@ravibeta this PR appears to have a severe problem. 

* It shows 16k+ lines of new code added, which looks odd to me. 
* There is a DCO check failing
* There are conflicting files.

ravibeta:
Yes Ravi, I will fix the change so it is easier to view the diff and I'm happy to share the following:
1) You are right this is an old and unsolved problem. It cannot do without disabling hostname verification.
2) Testing in the Kubernetes environment also shows a lot of difference from standalone tls deployment.
3) First, the pre-existing certificates don't work without hostname because of the error `curl: (51) SSL: certificate subject name 'server.pravegastack.io' does not match target host name 'a.b.c.d'`
4) The pravega-operator treats segmentstore differently from controller for tls related activities even though both take separate secrets
5) The system tests deploy pravega-operator for any test.
6) The exceptions encountered listed in the attached document shows pem format working better than jks format for kubernetes
7)  I worked around 2,3,4,5 by not requiring anything K8s specific since the standalone controller and segment store work with TLS enabled. I used the same Pravega configurations as for standalone by passing them through java_opts and reloading the pravega cluster which can be seen from new AbstractService methods.
8)  The only exception that remains to be solved is the 
java.lang.IllegalArgumentException: File does not contain valid certificates: /opt/pravega/conf/client.truststore.jks
        at io.netty.handler.ssl.SslContextBuilder.trustManager(SslContextBuilder.java:182)
which I believe we can solve with including trusted certificates in config and disabling this test until then.

ravibeta:
Closed in favor of https://github.com/pravega/pravega/pull/5059

PR #4941, opened 2020-07-12
Issue 4940: Throw TxnFailedException instead of mixed

### LINKED ISSUE #4940: TxnFailedException not RuntimeException on transaction aborts
**Problem description**
The TransactionalEventStreamWriter throws both TxnFailedExceptions and RuntimeExceptions. The latter is used in a mixed manner where the former is exclusively defined for the purpose of the TransactionalEventStreamWriter. This puts the client in a dilemma when using syntax like the following:
        Retry.withExpBackoff(DELAY, MULTIPLIER, ATTEMPTS, MAXDELAY)
             .retryingOn(TxnFailedException.class)
             .throwingOn(RuntimeException.class)
             .run(() -> {stateHelper.afterWrite(); return true;});
because it has to be selectively thrown on a case by case basis. 

**Problem location**
The problem occurs in 
client/src/main/java/io/pravega/client/stream/TransactionalEventStreamWriter.java
client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java

This is very similar in nature to Issue 3003 that tries to unify logging for transaction writers while this is unifying exceptions. Other than that similarity, these are two different issues.

ravibeta:
**Change log description**  
Raises and handles `TxnFailedException` instead of `RuntimeException` on Txn operations

**Purpose of the change**  
Fixes #4940 

**What the code does**  
The `TransactionalEventStreamWriter` throws both `TxnFailedException` and `RuntimeException`. The latter is used in a mixed manner where the former is exclusively defined for the purpose of the `TransactionalEventStreamWriter`. This puts the client in a dilemma when using syntax like the following:

```
        Retry.withExpBackoff(DELAY, MULTIPLIER, ATTEMPTS, MAXDELAY)
             .retryingOn(TxnFailedException.class)
             .throwingOn(RuntimeException.class)
             .run(() -> {stateHelper.afterWrite(); return true;});
````
because it has to be selectively thrown on a case by case basis. 

This change makes the `Exception` raised consistent across `io.pravega.client.stream` txn operations.

**How to verify it**  
All unit-tests, integration tests and system tests

### CODE REVIEW

File: client/src/main/java/io/pravega/client/stream/Transaction.java [hunk 1]
```diff
@@ -92,13 +92,16 @@
...
+     *
+     * @throws TxnFailedException Error aborting transaction: FAILURE
      */
-    void abort();
+    void abort() throws TxnFailedException;
 
     /**
      * Gets the status of the transaction.
      *
+     *  @throws TxnFailedException This exception could be thrown on checkStatus
```
eolivelli, line 103: This comment does not add value. Can you please explain in which case, as an example, this method will throw this exception?

client/src/main/java/io/pravega/client/stream/Transaction.java
```diff
@@ -92,13 +92,16 @@
...
      *  @return Current status of the transaction
      */
-    Status checkStatus();
+    Status checkStatus() throws TxnFailedException;
```
eolivelli, line 106: This is weird from the user point of view.
TxnFailedException sounds like the transaction is broken or you broke it by calling this method.

Here I would expect some new TxNotValidException

client/src/main/java/io/pravega/client/stream/TransactionalEventStreamWriter.java
```diff
@@ -26,17 +26,19 @@
      * Start a new transaction on this stream. This allows events written to the transaction be written an committed atomically.
      * Note that transactions can only be open for {@link EventWriterConfig#transactionTimeoutTime}.
      * 
+     * @throws TxnFailedException This exception could be thrown on beginTxn
```
eolivelli, line 29: This kind of comments are not useful to the user

client/src/main/java/io/pravega/client/stream/TransactionalEventStreamWriter.java
```diff
@@ -26,17 +26,19 @@
...
      * @return A transaction through which multiple events can be written atomically.
      */
-    Transaction<Type> beginTxn();
+    Transaction<Type> beginTxn() throws TxnFailedException;
```
eolivelli, line 32: How can a transaction that does not exist can fail?
Maybe it is fine to have this error here but it deserves a little explanation in javadocs

test/integration/src/main/java/io/pravega/test/integration/selftest/Actor.java
```diff
@@ -81,10 +82,18 @@ protected void doStart() {
...
-                .thenCompose(v -> run());
+                .thenCompose(v -> runWithException());
         this.runTask.whenComplete((r, ex) -> stopAsync());
     }
 
+    private CompletableFuture<Void> runWithException() {
+        try {
+            return run();
+        } catch (TxnFailedException ex) {
+            throw new RuntimeException(ex);
```
eolivelli, line 93: Methods that return a CP should not throw but return a failed CompletableFuture

test/integration/src/main/java/io/pravega/test/integration/selftest/Actor.java
```diff
@@ -119,7 +128,7 @@ protected void doStop() {
     /**
      * Executes the role of this Actor.
      */
-    protected abstract CompletableFuture<Void> run();
+    protected abstract CompletableFuture<Void> run() throws TxnFailedException;
```
eolivelli, line 131: The same here

test/integration/src/main/java/io/pravega/test/integration/selftest/StreamProducerDataSource.java
```diff
@@ -258,5 +261,13 @@ private void postStreamDeletion(String name) {
         this.state.recordDeletedStream(name);
     }
 
+    private CompletableFuture<Void> deleteStreamThrowing(String name, Duration timeout) {
+        try {
+            return this.store.deleteStream(name, this.config.getTimeout());
+        } catch (TxnFailedException ex) {
+            throw new CompletionException(ex);
```
eolivelli, line 268: return a failed future  here and in other points

client/src/main/java/io/pravega/client/stream/TransactionalEventStreamWriter.java
```diff
@@ -26,17 +26,19 @@
...
 
     /**
      * Returns a previously created transaction.
      * 
      * @param transactionId The result retained from calling {@link Transaction#getTxnId()}
+     * @throws TxnFailedException The Transaction is no longer in state {@link Transaction.Status#OPEN}
```
tkaitchuck, line 38: I don't think we want this behavior. 
It seems legitimate to me to do:
```
getTxn(id).checkStatus()
```
In this case I would expect the result to return whatever the status was even if it is not `OPEN`. 
If the code doesn't work this way. we should open an issue.

client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java
```diff
@@ -144,15 +144,15 @@ public void abort() {
                         log.debug("Got exception while writing to transaction on abort: {}", e.getMessage());
                     }
                 }
-                getAndHandleExceptions(controller.abortTransaction(stream, txId), RuntimeException::new);
+                getAndHandleExceptions(controller.abortTransaction(stream, txId), TxnFailedException::new);
```
tkaitchuck, line 147: I think what we really want if for the controller client to dictate the exception type. 
So instead of converting this, I think we should instead call `getThrowingException` which will just pass up whatever the underlying error was.

client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java
```diff
@@ -176,9 +176,9 @@ private void throwIfClosed() throws TxnFailedException {
     }
 
     @Override
-    public Transaction<Type> beginTxn() {
+    public Transaction<Type> beginTxn() throws TxnFailedException {
         TxnSegments txnSegments = getAndHandleExceptions(controller.createTransaction(stream, config.getTransactionTimeoutTime()),
-                RuntimeException::new);
+                TxnFailedException::new);
```
tkaitchuck, line 181: Same here

client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java
```diff
@@ -198,16 +198,16 @@ private void throwIfClosed() throws TxnFailedException {
     }
 
     @Override
-    public Transaction<Type> getTxn(UUID txId) {
+    public Transaction<Type> getTxn(UUID txId) throws TxnFailedException {
         // check if the transaction is open.
-        Status status = getAndHandleExceptions(controller.checkTransactionStatus(stream, txId), RuntimeException::new);
+        Status status = getAndHandleExceptions(controller.checkTransactionStatus(stream, txId), TxnFailedException::new);
```
tkaitchuck, line 203: same here

ravibeta:
Thank you for your comments. I will revisit these. Cheers

ravibeta:
All code review comments addressed. Thank you very much.

client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java
```diff
@@ -145,15 +144,15 @@ public void abort() {
                         log.debug("Got exception while writing to transaction on abort: {}", e.getMessage());
                     }
                 }
-                getAndHandleExceptions(controller.abortTransaction(stream, txId), RuntimeException::new);
+                Futures.getThrowingException(controller.abortTransaction(stream, txId));
```
shrids, line 147: Observation : The abort Transaction operations on the controller are idempotent, hence calling an abortTransaction() successively will not throw an error.(TnxFailedException or any other error).  
The controller client will throw a TxnFailedException if the txn id is not found by the controller during abort().

client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java
```diff
@@ -145,15 +144,15 @@ public void abort() {
...
                 closed.set(true);
             }
         }
 
         @Override
         public Status checkStatus() {
             log.info("Check transaction status {}", txId);
-            return getAndHandleExceptions(controller.checkTransactionStatus(stream, txId), RuntimeException::new);
+            return Futures.getThrowingException(controller.checkTransactionStatus(stream, txId));
```
shrids, line 155: checkTransactionStatus will never throw a TxnFailedException ; as I understand the intent here is to remove redundant RuntimeException wrapping.

ravibeta: Yes Sandeep

client/src/main/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterImpl.java
ravibeta: Yes Sandeep

test/integration/src/main/java/io/pravega/test/integration/selftest/StreamProducerDataSource.java
ravibeta: Yes Enrico

test/integration/src/main/java/io/pravega/test/integration/selftest/Actor.java
ravibeta: Yes Enrico

test/integration/src/main/java/io/pravega/test/integration/selftest/Actor.java
ravibeta: Yes Enrico

client/src/main/java/io/pravega/client/stream/Transaction.java
ravibeta: Yes Enrico it does not throw TxnFailedException anymore.

client/src/main/java/io/pravega/client/stream/TransactionalEventStreamWriter.java
ravibeta: Yes Enrico

client/src/main/java/io/pravega/client/stream/TransactionalEventStreamWriter.java
ravibeta: Yes Enrico

PR #4950, opened 2020-07-17
Issue 4751: ListScopes, CheckStreamExists and CheckScopeExists

### LINKED ISSUE #4751: Create a list scopes API in stream manager
**Problem description**
Flink connector is currently trying to develop some POC on Pravega [catalog](https://ci.apache.org/projects/flink/flink-docs-stable/dev/table/catalogs.html) which can manage the metadata of Pravega in Flink SQL/Table API. Such implementation is abstracting scopes as Flink databases and streams as Flink tables.
We are trying to use `StreamManager` to manage the metadata, but `listScopes` API is missing to complete the catalog implementation. (https://github.com/apache/flink/blob/master/flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/Catalog.java#L102)
As `StreamManager` provides `createScope` and `deleteScope` API, I don't think we have any security issues to add such an API. 

Update:
The `streamExists` and `ScopeExists` API is also a needed for this purpose.

**Problem location**
`io.pravega.client.admin.StreamManager` and impl class

List scopes should return the scopes that are authorized to be seen concerning the user.

shiveshr:
**Change log description**  
Implementation for listScopes, CheckScopeExists and CheckStreamExists

**Purpose of the change**  
Fixes #4751 

**What the code does**  
Adds grpc implementation for listScopes which is a paginated call. 
Also provides implementation for CheckScopeExists and CheckStreamExists.
There is not implementation for Zookeeper based store for listScopes 

**How to verify it**  
Unit test updated


client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
```diff
@@ -107,13 +107,26 @@ public boolean deleteStream(String scopeName, String streamName) {
         return  Futures.getThrowingException(controller.deleteStream(scopeName, streamName));
     }
 
+    @Override
+    public Iterator<String> listScopes() {
+        log.info("Listing scopes");
+        AsyncIterator<String> asyncIterator = controller.listScopes();
+        return new BlockingAsyncIterator<>(asyncIterator);
```
andreipaduroiu, line 114: `asyncIterator.asIterator()` will do the same, but you won't need to explicitly use `BlockingAsyncIterator`. This will help in case we want to change the implementation later.

client/src/test/java/io/pravega/client/stream/mock/MockStreamManager.java
```diff
@@ -181,6 +193,30 @@ public boolean deleteStream(String scopeName, String toDelete) {
         return Futures.getAndHandleExceptions(controller.deleteStream(scopeName, toDelete), RuntimeException::new);
     }
 
+    @Override
+    public Iterator<String> listScopes() {
+        AsyncIterator<String> asyncIterator = controller.listScopes();
```
andreipaduroiu, line 198: Why not use `asyncIterator.asIterator`?

You implementation below is incorrect. Both `hasNext` and `next` will invoke `load` which will cause the wrapped iterator to advance. This means you will be skipping over iterator items as you iterate through. Also `next` is supposed to throw if `hasNext() == false`

controller/src/main/java/io/pravega/controller/server/eventProcessor/LocalController.java
```diff
@@ -68,6 +70,45 @@ public LocalController(ControllerService controller, boolean authorizationEnable
         this.authorizationEnabled = authorizationEnabled;
     }
 
+    @Override
+    public CompletableFuture<Boolean> checkScopeExists(String scopeName) {
+        return this.controller.getScope(scopeName)
```
andreipaduroiu, line 75: use `Futures.exceptionallyExpecting` to make this a one-liner.

controller/src/main/java/io/pravega/controller/server/eventProcessor/LocalController.java
```diff
@@ -68,6 +70,45 @@ public LocalController(ControllerService controller, boolean authorizationEnable
...
+        final Function<String, CompletableFuture<Map.Entry<String, Collection<String>>>> function = token ->
+                controller.listScopes(token, PAGE_LIMIT)
+                          .thenApply(result -> new AbstractMap.SimpleEntry<>(result.getValue(), result.getKey()));
+
+        return new ContinuationTokenAsyncIterator<>(function, "");
+    }
+
+    @Override
+    public CompletableFuture<Boolean> checkStreamExists(String scopeName, String streamName) {
+        return this.controller.getStream(scopeName, streamName)
```
andreipaduroiu, line 99: `Futures.exceptionallyExpecting`

client/src/test/java/io/pravega/client/stream/mock/MockStreamManager.java
shiveshr: yes, but the load doesnt call asynciterator.getNext unless it has exhausted all existing entries. 
so wrapper iterator will only advance when its getNext gets called and if it returns items then hasNext returns true, false otherwise.

shiveshr: And next throws exception if you call next when hasNext may have written false.

andreipaduroiu: The point is that you are returning an `Iterator` object which does not conform to the `Iterator` [contracts](https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html). If you want to implement a different behavior you will have to return something that does not implement `java.util.Iterator`

shiveshr: i dont completely follow how the behaviour is different for BlockingAsyncIterator from the iterator contract. 
Can you please illustrate with an example.

It upholds following contracts (does not provide an implementation for `remove` method as that is unsupported for this iterator).

- if there are elements to iterate over, `hasNext` will return true. 
- `next` returns next element in the iteration "without skipping over any element".

Following is the implementation for load method which is called from both hashNext and next which loads next and canHaveNext values into fields `T next` and `boolean canHaveNext`.
then hasNext simply returns canHaveNext and next returns next or throws if canHaveNext is false and next is called.
```
    private void load() {
        if (next == null && canHaveNext) {
            next = asyncIterator.getNext().join();
            if (next == null) {
                canHaveNext = false;
            }
        }
    }
```

`forEachRemaining` would work correctly as it performs `hasNext` followed by `next` call in a loop.

shiveshr: Oh. My Bad.

 I didnt realize that this mock class also has an implementation.. yes, that is perhaps incorrect. 
i was talking about BlockingAsyncIterator.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -532,6 +532,126 @@ public void createScope(ScopeInfo request, StreamObserver<CreateScopeStatus> res
...
+                .listScopes(request.getContinuationToken().getToken(), pageLimit)
+                .thenApply(response -> {
+                    log.debug("All scopes in scope with continuation token: {}", response);
+                    return Controller.ScopesResponse
+                            .newBuilder().addAllScopes(response.getKey())
+                            .setContinuationToken(Controller.ContinuationToken.newBuilder()
+                                                                              .setToken(response.getValue()).build())
+                            .build();
+                });
+        authenticateExecuteAndProcessResults(stringSupplier, scopesFn, responseObserver, requestTag);
```
pbelgundi, line 567: Can the auth filter out some scopes, causing us to return less number of scopes than limit?

shiveshr: We could, but i kept the authorization part here consistent with the equivalent api in the REST which requires users to have READ permission on "/" to be able to invoke list scopes. 

If we want a different approach for authorization, we should do it consistently on both interfaces. 
I have created a separate issue for the same here:
https://github.com/pravega/pravega/issues/4955

Issue #4907, opened 2020-06-29
Separate bindings from Pravega


ravibeta:
**Problem description**
This issue tracks the initiative to separate bindings from Pravega source

**Problem location**
The source will be moved from project('bindings') to its own repository

**Suggestions for an improvement**
A draft for the corresponding PDP is included below.

Associated code changes:
https://github.com/pravega/pravega-bindings/pull/1
https://github.com/pravega/pravega/pull/4908

**Status:** Draft
# Summary
This PDP discusses a change to not require S3, hdfs and file-system bindings to be build with the Pravega because they incur a lot of technical debt.
Once implemented, this proposal will allow the bindings to be discovered at runtime and used independently.
The project(bindings) will be removed from build.gradle and will appear in its own repository.
# Shortcomings of the current implementation
1. **External dependency**. The current implementation of Bindings is at the mercy of all defects from external service providers such as HDFS, extended S3 and file-system implementors. Frequently, they change their methods and properties in their apis and cause break-in changes in Pravega while the bindings are not always required for the execution of the stream store.
2. **Incorrect usages**. Usually the bindings are loaded to serve the standalone mode of deployment. However, they are subject to misuse. For example, hdfs binding is used to launch an distributed file system cluster over ephemeral storage all within a single node. Such deployments should not be in production.
# Key Design Changes
## Proposal
Below is the summary of this proposal.
#### *Step 1*: Implement the bindings in a separate repository.
* The bindings project will be completely removed from the Pravega source. It serves to provide implementations for SyncStorage which will be resolved dynamically.
* It will be hosted in a separate repository where it can be serviced more frequently as and when the external dependencies provided updates.
#### *Step 2*: Re-Implement Bindings to load the class libraries.
* NFS
* HDFS
* Extended S3
#### *Step 3*: Re-implement segment store to load the class libraries:
* ChunkManager
* Metadata table store and
* Chunk Storage provider
Earlier these used to correspond to
* No-op storage
* Asynchronous storage
* Rolling storage
Use ChunkStorageProvider not just in a simulated environment but also in the actual deployment corresponding to No-op, asynchronous and Rolling storage.
## API Changes
*  There are no user level API changes to ChunkStorageProvider.
*  There are no changes to how tier-2 is configured.
## API Testing
*  All usages of ChunkStorageProvider from segmentstore corresponding to No-op, async and rolling storages will be exercised.
[...]

PR #4954, opened 2020-07-21
add new docker method

ravibeta:
**Change log description**  
A new way to do docker for Igor's request to include jars outside maven references

**Purpose of the change**  
#4907 

**What the code does**  
packages the docker and dockerPush

**How to verify it**  
The changes are part of build integration.


ravibeta:
This is in draft mode because we want it to go into a feature branch that is coming.

PR #4957, opened 2020-07-22
Issue 4956: Fix bugs in list streams and list scope api when resources are filtered

### LINKED ISSUE #4956: Authorization Filtering in list streams can return empty results if everything is filtered out, while the next page may have valid results
**Problem description**
List streams should return valid results in a page as long as there are valid results available. we filter out results based on authorization rules, but that could mean that all results in a page get filtered out, in which case we would send an empty response making the caller believe the request has ended while there may still be data in subsequent pages. 

So instead of returning the empty list, call the list method to retrieve the values and if some values are filtered out, call the list method recursively with reduced limits. 

Also same method for kvt has two minor issues:
1. recursive call is made in a future who is not chained to the request's response.
2. the recursive method call condition presumes that if fetched records are less than requested limit, then we should make the recursive call. Luckily it sends 0 as number of records to fetch which causes termination of the request. but the condition is incorrect.
**Problem location**
ControllerServiceImpl

Fix the issue

shiveshr:
**Change log description**  
If resources are filtered due to authorization, keep calling list resource api on store until we have collected desired number of results to fill the page or entries are exhausted. 

**Purpose of the change**  
Fixes #4956  

**What the code does**  
Make recursive call if the resources are filtered out due to authorization until we either collect "limit" number of records or exhaust all entries in the store. 
This is performed for both streams and KVT.
KVT already had the implementation but there were bugs in it which are also fixed here. 

Also, segment store returns an empty continuation token if we request from the last element and there are no more elements to read. This is handled in PravegaTableStoreHelper to return the caller's token if response from segment store is empty. 

**How to verify it**  
unit tests added


controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -696,49 +696,65 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request, StreamO
...
-                                AuthResourceRepresentation.ofScope(scopeName),
-                                AuthHandler.Permissions.READ,
-                                ctx);
-                        log.debug("Result of authorization for [{}] and READ permission is: [{}]",
-                            AuthResourceRepresentation.ofScope(scopeName), result);
-                        return result;
-                },
-                delegationToken -> controllerService
-                        .listStreams(scopeName, request.getContinuationToken().getToken(), pageLimit)
+        Function<String, CompletableFuture<Controller.StreamsInScopeResponse>> streamsFn = delegationToken ->
```
pbelgundi, line 699: The AuthContext if..else block could be expressed more concisely using a ternary operator as done in listKeyValueTables :
https://github.com/pravega/pravega/blob/bf961c4e3b39353b9881dbe60252bfbd1fbdae66/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java#L158

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -696,49 +696,65 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request, StreamO
...
-                                        .setContinuationToken(Controller.ContinuationToken.newBuilder()
-                                                                .setToken(response.getValue()).build())
+                                        .newBuilder().addAllStreams(response.getKey())
+                                        .setContinuationToken(Controller.ContinuationToken.newBuilder().setToken(response.getValue()).build())
                                         .setStatus(Controller.StreamsInScopeResponse.Status.SUCCESS).build();
                             }
-                        }), responseObserver, requestTag);
+                        });
+
+        authenticateExecuteAndProcessResults(
```
pbelgundi, line 717: Passing streamsFn as an argument to authenticateExecuteAndProcessResults is a much cleaner way to organize code. Could we also do the same for listKeyValueTablesInScope?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -696,49 +696,65 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request, StreamO
...
+    }
+
+    private CompletableFuture<Pair<List<StreamInfo>, String>> listStreamsInternal(String scope, String continuationToken, int limit, AuthContext ctx) {
+        List<StreamInfo> streams = new ArrayList<>();
+        return controllerService.listStreams(scope, continuationToken, limit)
+                                .thenCompose(response -> {
+                                    log.debug("All streams in scope with continuation token: {}", response);
+                                    // filter unauthorized scopes. 
+                                    // fetch recursively if scopes are filtered out.
+                                    AtomicInteger filteredCount = new AtomicInteger();
```
pbelgundi, line 736: We could do with an AtomicBoolean here instead of Integer.... all we need to know is if something got filtered out or no...

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -428,13 +430,15 @@ public void invalidateCache(String table, String key) {
                                      List<String> items = result.getItems().stream().map(x -> new String(getArray(x.getKey()), Charsets.UTF_8))
                                                                 .collect(Collectors.toList());
                                      log.trace("get keys paginated on table {} returned items {}", tableName, items);
-                                     return new AbstractMap.SimpleEntry<>(Unpooled.wrappedBuffer(result.getState().toBytes()), items);
+                                     // if the returned token and result are empty, return the incoming token so that 
+                                     // callers can resume from that token. 
+                                     return new AbstractMap.SimpleEntry<>(getNextToken(continuationToken, result), items);
```
pbelgundi, line 435: What is this for?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -696,49 +696,65 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request, StreamO
...
+
+                                    response.getKey().stream().filter(x -> {
```
pbelgundi, line 738: Perhaps putting it this way may make it easier to read and we also won't need the Atomic variable
```
List<String> streamNames = response.getKey().stream().filter(x -> grpcAuthHelper.isAuthorized(AuthResourceRepresentation.ofStreamInScope(scope, x), AuthHandler.Permissions.READ, ctx)).collect(Collectors.toList())
if (streamNames.size() < response.getKey()) {
//recurse...
} else {
//return ...
}
```

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shiveshr: The promise of controller vis a vis the continuation token is that it is an artifact from which if you make a query to read next set of items, it will retrieve them from the store and if there are no more results to be found, it would return empty list but the same token.
Basically token indicates the position till which the results have been consumed. 

if they have exhausted the whole result set, this should point to the last element in the set.
however, the read all keys/entries api on segment store doesnt return the same token back, instead it returns an empty token if there are no more items to return. 
for example: if you issue `readAllKeys(empty)` and `get {List<Keys>, Token}` .. lets say all keys are read.. now if you issue readAllKeys(token), segment store returns {empty List, empty token}`
This makes sure we return the same token back instead so that any time user makes a subsequent call from the token, it gets a response from which it can continue without reading the whole result again.

shiveshr:
@pbelgundi  i have refactored the code to create a common method that is used for all three -- list scopes, list streams in scope, and list tables in scope. 
the refactored code includes all your other suggestions as well.

PR #4958, opened 2020-07-22
Issue 4927 deleteKeyValueTableTest fails for InMemoryKVTableStore when using KVTOperationContext

### LINKED ISSUE #4927: deleteKeyValueTableTest fails for InMemoryKVTableStore when using KVTOperationContext
**Problem description**
deleteKeyValueTableTest() in InMemoryControllerImplServiceTest.java fails with timeout exception for InMemoryKVTableStore when using KVTOperationContext.
The same when invoked for PravegaTablesStore works fine.
This could be due to some issue with the test code or with InMemoryKVStore implementation.

**Problem location**
io/pravega/controller/server/v1/InMemoryControllerImplServiceTest.java --> deleteKeyValueTableTests()

pbelgundi:
**Change log description**  
The root cause was that in case of InMemoryKVTableMetadataStore, when KVTOperationContext was used to get a reference to the KeyValueTable, it kept pointing to a deleted KVTable object. As a result, the state of the KVTable never changed from DELETING to UNKNOWN as was expected after a deleteKVTable operation.

**Purpose of the change**  
Fixes #4927, #4960 

**What the code does**  
In case of an InMemoryKVTableMetadataStore, using KVTOperationContext is unnecessary as all data is already stored in Memory. Removed use of KVTOperationContext in getKVTable() method of InMemoryKVTableMetadataStore and now the test works as expected.

**How to verify it**  
All Unit and Integration tests should pass.


controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -57,20 +51,12 @@ KeyValueTable newKeyValueTable(String scope, String kvTableName) {
 
     @Override
     public KeyValueTable getKVTable(String scope, final String name, KVTOperationContext context) {
-        KeyValueTable kvt;
```
shiveshr, line 60: why is this being changed? what was wrong with taking the kvt from the context?

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -57,20 +51,12 @@ KeyValueTable newKeyValueTable(String scope, String kvTableName) {
...
-            assert kvt.getScopeName().equals(scope);
-            assert kvt.getName().equals(name);
-        } else {
-            if (!scopes.containsKey(scope)) {
-                return new InMemoryKVTable(scope, name);
-            }
-            InMemoryScope kvtScope = scopes.get(scope);
-            Optional<InMemoryKVTable> kvTable = kvtScope.getKVTableFromScope(name);
-            kvt = kvTable.orElse(new InMemoryKVTable(scope, name));
+        if (!scopes.containsKey(scope)) {
```
shiveshr, line 54: why is the check on scopes? 
shouldnt the check be on table.. if the table doesnt exist, create it.. else return an already created table.. and this should be concurrency protected.

shiveshr: and why are we creating kvt object if the scope doesnt exist?
this logic seems incorrect.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -40,14 +39,9 @@
     @GuardedBy("$lock")
     private Map<String, InMemoryScope> scopes = new HashMap<>();
 
-    private final AtomicInt96 counter;
-
-    private final Executor executor;
 
     public InMemoryKVTMetadataStore(Executor executor) {
```
andreipaduroiu, line 43: Remove `executor` from the argument list.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: There is no advantage in taking KVT from context in case of InMemoryStore as all data is already in memory.
When using KVT from context, the KVTable object is available in the context though it is removed from the scope. Besides it has State="Deleting" which causes the isDeleted() method to never compete.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: Ok. Sure.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: This was added because scopes created in StreamStore are not available inside KVTStore. Will fix this as part of fix for issue #4960.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
shiveshr: what is the delete method doing? perhaps it should be implemented such that it marks the entry as deleted and removes the entry from the map. 

that way context will be able to get the correct value as well.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
shiveshr: still, the condition being added here with scopes doesnt exist seems to be introducing an incorrectness. Perhaps we should fix 4960 first before we fix this.

pbelgundi: Sure. I can raise the PR for 4960 and merge this after that.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: It should all fall in place once Scopes are implemented correctly in InMemoryKVTMetadataStore.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: Fixed as part of this PR.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -36,48 +36,44 @@
...
-            kvt = context.getKvTable();
-            assert kvt.getScopeName().equals(scope);
-            assert kvt.getName().equals(name);
-        } else {
-            if (!scopes.containsKey(scope)) {
-                return new InMemoryKVTable(scope, name);
+    @Synchronized
+    KeyValueTable newKeyValueTable(String scope, String name) {
+        if (this.streamStore.getScopes().containsKey(scope)) {
+            InMemoryScope kvtScope = this.streamStore.getScopes().get(scope);
```
shiveshr, line 52: streamstore.getScopes is returning a map which is not synchronization protected in its access here.. 
instead expose a syncrhonized method on the stream store for getting the scope.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -36,48 +36,44 @@
...
     @Override
     public Scope newScope(String scopeName) {
         return getScope(scopeName);
     }
 
+    @Override
+    @Synchronized
+    public KeyValueTable getKVTable(String scope, final String name, KVTOperationContext context) {
+        if (this.streamStore.getScopes().containsKey(scope)) {
+            InMemoryScope kvtScope = this.streamStore.getScopes().get(scope);
```
shiveshr, line 69: same here.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -98,20 +94,26 @@ public Scope newScope(String scopeName) {
...
+        return Futures.completeOn(CompletableFuture.completedFuture(this.streamStore.getScopes().containsKey(scope)), executor);
     }
 
     @Override
     @Synchronized
     public CompletableFuture<Boolean> checkTableExists(String scopeName, String kvt) {
-        return CompletableFuture.completedFuture((InMemoryScope) getScope(scopeName)).thenApply(scope -> scope.checkTableExists(kvt));
+        return Futures.completeOn(checkScopeExists(scopeName).thenCompose(exists -> {
+            if (exists) {
+                return CompletableFuture.completedFuture((InMemoryScope) getScope(scopeName))
```
shiveshr, line 105: you dont need to do `CompletableFuture.completedFuture().thenApply`.
Instead you can compute the value and then do CompletableFuture.completedFuture(value)

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -98,20 +94,26 @@ public Scope newScope(String scopeName) {
...
+                        .thenApply(scope -> scope.checkTableExists(kvt));
```
shiveshr, line 106: isnt scope.checkTableExists returning a future?
if so, shouldnt this be thenCompose?

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: Sure.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: sure.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
shiveshr: creating an immutable map each time its accessed here is costly as it will make a copy. 
instead you can use `Collections.unmodifiableMap` which actually creates a wrapper decoration on the map which prohibits any changes to the map

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
```diff
@@ -41,7 +41,7 @@
     private HashMap<String, Integer> streamsPositionMap;
 
     @GuardedBy("$lock")
-    private TreeMap<String, InMemoryKVTable> kvTablesMap;
+    private final TreeMap<String, InMemoryKVTable> kvTablesMap =  new TreeMap<String, InMemoryKVTable>();
```
shiveshr, line 44: this can be reverted except the `final` keyword.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
```diff
@@ -57,7 +57,6 @@ public String getName() {
     public CompletableFuture<Void> createScope() {
         this.sortedStreamsInScope = new TreeMap<>(Integer::compare);
         this.streamsPositionMap = new HashMap<>();
-        this.kvTablesMap =  new TreeMap<String, InMemoryKVTable>();
```
shiveshr, line 60: it is more consistent to initialize this in the contructor as all other fields are also initialized in the constructor.

shiveshr: also, all other fields are non final. as we set them to null in deleteScope. though logically that doesnt really give us anything. but since that is the pattern being used for all other objects, i would suggest we keep the same pattern for this field too.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
shiveshr: actually i dont see you returning an immutable map.

immutable map is one in which you cannot add/remove/replace any entries. 

setting the field as final doesnt make it immutable, it just means that field cannot be assigned a different value. but the object that the field refers to can still be immutable, as is the case here.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
```diff
@@ -175,4 +165,7 @@ public void refresh() {
         return CompletableFuture.completedFuture(null);
     }
 
+    public KeyValueTable getKeyValueTable(String name) {
```
shiveshr, line 168: kvTablesMap is guarded by $lock so add synchronized annotation for this method

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStreamMetadataStore.java
```diff
@@ -74,7 +75,7 @@ public InMemoryStreamMetadataStore(Executor executor) {
     }
 
     public Map<String, InMemoryScope> getScopes() {
```
shiveshr, line 77: please add synchronized here as well as this is reading the `scopes` map which is concurrency protected.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
pbelgundi: I don't like the idea of having data members point to null values and wouldn't want to repeat the same just for consistency.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: Changed as discussed.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: changed as discussed.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
pbelgundi: Why?

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
pbelgundi: done

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStreamMetadataStore.java
pbelgundi: removed this method as discussed.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
shiveshr: no i mean init in constructor

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
shiveshr: we should never be accessing a "deleted" scope object. 
The purpose of setting all fields in the scope object to null should not cause NPE. And if it does, it would point to a bug that should be fixed.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
pbelgundi: Construction initialization is good for usecases where the values used in building the object need to be passed to the constructor. Initializing at the time of declaration also takes away the need to synchronize the constructor and I don't see any downsides with initializing it this way.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
pbelgundi: Whenever possible, I'd want to avoid null references to class data members in the code, as a good coding practice...

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
shiveshr: no downside, just consistency as other objects are initialized in the constructor.

Also constructor is implicitly synchronized and you dont need to synchronize a constructor.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
```diff
@@ -98,20 +97,25 @@ public Scope newScope(String scopeName) {
...
 
     @Override
     @Synchronized
     public Scope getScope(final String scopeName) {
-        if (scopes.containsKey(scopeName)) {
-            return scopes.get(scopeName);
+        if (this.streamStore.scopeExists(scopeName)) {
+            return this.streamStore.getScope(scopeName);
         } else {
             return new InMemoryScope(scopeName);
```
shiveshr, line 120: is there a check that ensures that "create" is called on the scope object before it is used. 
this is returning the scope object, but where does it get added to the store?

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
pbelgundi: Ok.

controller/src/main/java/io/pravega/controller/store/kvtable/InMemoryKVTMetadataStore.java
pbelgundi: This is implemented in same lines as similar method in InMemoryMetadataStore:
https://github.com/pravega/pravega/blob/master/controller/src/main/java/io/pravega/controller/store/stream/InMemoryStreamMetadataStore.java#L117

shiveshr: yes, but it looks incorrect in both places. we need to see if getScope should return scope not found if the scope does not exist in the store.

pbelgundi: I agree, but it may need significant code refactoring as there are multiple dependent methods that would break.
Filed: https://github.com/pravega/pravega/issues/5021

andreipaduroiu:
@pbelgundi Looks like I can't merge this due to codecov complaining ....

PR #4959, opened 2020-07-22
Issue-4937: Update ZooKeeper and BookKeeper pre-requisites in the documentation

### LINKED ISSUE #4937: Enhance Manual Install Page
**Problem description**

Current [Manual installation page](http://pravega.io/docs/latest/deployment/manual-install/) is to be improved:
- Upgrade BooKeeper and ZooKeeper versions (#4832)
- it should tell about installing a single node Pravega cluster with FILESYSTEM storage (standalone mode is not useful for testing real applications in single mode, as it it easy to break because mostly everything runs in memory)
- It should tell about the requirement to have a working HDFS cluster
- it should tell about how to create a scope/stream (even with _curl_ is fine)

This change may be coordinated with changing default values for:
- bookkeeper replication (use 1 instead of 3)
- tier2 settings (set to FILESYSTEM and set an explicit path for the directory)

Detailed changes:
- Update BK and ZK versions
- Add section about single node with FILESYSTEM storage
- Add section about HDFS
- Document more about FILESYSTEM storage (for instance the default configuration assumes that the filesystem is at '/fs' and that directory should be already present and with the right permissions)
- Maybe suggest to run on Java 11 LTS instead of Java 8 (tell that Java 8 is still supported)

For users that are going to try Pravega it better to have a way to run it locally and in a way that data survive after a reboot.

**Problem location**
http://pravega.io/docs/latest/deployment/manual-install/

ravibeta:
**Change log description**  
This updates the documentation to include the proper versions and references as detailed in #4937 

**Purpose of the change**  
Fixes #4937 

**What the code does**  
There is no change to code. This improves documentation.

**How to verify it**  
The steps were tried manually as part of a repackaging fix for another issue.


documentation/src/docs/deployment/manual-install.md
```diff
@@ -19,27 +19,34 @@ Setup a HDFS storage cluster running **HDFS version 2.7+**. HDFS is used as Tier
 sufficient capacity to store the contents of all the streams. The storage cluster is recommended to be run
 alongside Pravega on separate nodes.
 
+### Filesystem
+
+HDFS is the default Tier 2 storage which is also called Long Term Storage. If it is easier to mount an nfs share, then Filesystem can be used in place of HDFS. The following configuration options are necessary to configure the Filesystem as Long Term Storage.
```
andreipaduroiu, line 24: 1. Do not define Tier 2 <-> Long Term Storage here. There is another place in the doc where we should do this.
2. "nfs" -> "NFS".
3. "Filesystem" -> "FILESYSTEM"

documentation/src/docs/deployment/deployment.md
```diff
@@ -28,9 +28,10 @@ The following prerequisites are required for running Pravega in all modes.
 The following prerequisites are required for running in production. These are only required for running in distributed mode.
 
 - External HDFS 2.7
-- Zookeeper 3.5.4-beta
-- Bookkeeper 4.7.3
+- Zookeeper 3.6.1
+- Bookkeeper 4.9.2
 
+Any version onwards from these versions can be used. These versions are also used by [zookeeper-operator](https://github.com/pravega/zookeeper-operator) and [bookkeeper-operator](https://github.com/pravega/bookkeeper-operator)
```
andreipaduroiu, line 34: `Any version onwards from these versions can be used`

I would remove this altogether. There is no guarantee that a new version of BK will work without issues with the current code of Pravega. If we cannot guarantee that with the current code, we shouldn't be stating this.

documentation/src/docs/deployment/run-local.md
```diff
@@ -60,6 +61,11 @@ Run Pravega Standalone:
 $ pravega-<version>/bin/pravega-standalone
 ```
 
+
+##  With Long Term Storage
+
+When the streams are required to be persisted between server restarts, a tier 2 storage is required by Pravega. For Example, a mounted nfs-share can be used can be used a tier2 storage. Only the Segment store requires its host to be configured with this storage. The controller and segment store are run as different processes. The long term storage is set up only on the latter. These steps are described in [Manual Installation](manual-install.md).
```
andreipaduroiu, line 67: Please use proper capitalization for tier 2, nfs, segment store, controller.

documentation/src/docs/deployment/deployment.md
ravibeta: Yes Andrei

documentation/src/docs/deployment/manual-install.md
ravibeta: Yes Andrei

PR #4963, opened 2020-07-22
Issue 4962: (SegmentStore) Enabling Netty Paranoid Leak Detection if needed

### LINKED ISSUE #4962: Enable Netty Leak Detector
**Problem description**
Add a new config value to enable paranoid Netty Leak detector.

**Problem location**
Segment Store ServiceStarter.

Add it.

andreipaduroiu:
**Change log description**  
Added a setting to enable Netty Paranoid Leak Detection. Disabled by default.

**Purpose of the change**  
Fixes #4962.

**What the code does**  
Added `pravegaservice.netty.leakDetection.enable` config value that enables aggressive Netty ByteBuf leak detection. If enabled, sets the mode to PARANOID.

**How to verify it**  
Build must pass. If any leak detections do occur as a result of this, they should be addressed in a separate PR.

### DIFF
--- a/config/config.properties
+++ b/config/config.properties
@@ -168,6 +168,11 @@ pravegaservice.dataLog.impl.name=BOOKKEEPER
 # information to clients, which may raise security concerns.
 #pravegaservice.request.replyWithStackTraceOnError.enable=false
 
+# Whether to enable PARANOID Netty ByteBuf Leak detection.
+# Recommended values: false for production environments.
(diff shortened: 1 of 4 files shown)


andreipaduroiu:
Looks like this is not necessary. Netty has a system property that it listens to in order to do the same thing:
```
-Dio.netty.leakDetection.level={level}
```

PR #4965, opened 2020-07-22
Issue 4964: Prevent a busy loop while waiting for the segments to be updated

### LINKED ISSUE #4964: Eliminate busy loop in EventStreamWriterImpl
**Problem description**
Currently in EventStreamWriterImpl.flush() if a segment it sealed and in the process of being replaced by its successor, it breaks out of the loop and restarts the flush process. This is the correct behavior, however because it doesn't actually wait for the flush process to complete it can happen multiple times. This is OK, and wasn't previously a problem because it performed IO to send a keep alive on each connection which meant this fairly slow. 

However recently the keep alive was removed and the flush call was optimized to an immediate throw if the segment is already known to be sealed. This is more efficient, but it has the effect of making the flush call into a busy loop.

**Problem location**
EventStreamWriterImpl

Actually coordinating perfect synchronization would be difficult and disruptive. The simplest solution would be to add a sleep, but this can probably be improved upon.

tkaitchuck:

**Change log description**  
Prevent a busy loop while waiting for the segments to be updated

**Purpose of the change**  
Fixes #4964

**What the code does**  
Adds a sleep in flush in the event that a segment is sealed and being processed by a background thread. This is optimized somewhat by using it's lock if the process has already begun. 

**How to verify it**  
We should run system tests on this PR before merging.


client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -224,12 +224,32 @@ public void flush() {
...
+     * This is used by flush to optimistically wait for the `handleLogSealed` work to be completed
+     * to prevent a busy loop while waiting for the segments to be updated following a sealed
+     * segment. Please note there are no guarantees about when this method returns. It can be
+     * thought of as a sleep.
+     */
+    @GuardedBy("writeFlushLock")
+    private void tryWaitForSuccessors() {
+        if (sealedSegmentQueue.isEmpty()) {
+            synchronized (writeSealLock) {
+                // Nothing needs to be done here.
```
eolivelli, line 245: Are  you sure that the JVM (or even javac) won't drop this block? AFAIK the JVM is quite aggressive in skipping useless stuff and monitor related code

andreipaduroiu: Yeah this could be a possibility. 

Could we try to use a ReusableLatch here or some Semaphore (or any other synchronization primitive)?

eolivelli: probably if you add a metric here (just a counter that counts how many times we fall into this case) the JVM will be happy as it thinks that we are doing something useful

andreipaduroiu: I think SpotBugs doesn't like this either. The build failed due to that.

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -224,12 +224,33 @@ public void flush() {
...
+                // When the lock is released the sealing should be complete.
+                log.debug("Segment sealing completed, retrying flush.");
+            }
+        } else {
+            // A background thread should be waking up to process things. Give it a moment.
+            Exceptions.handleInterrupted(() -> Thread.sleep(100));
```
shrids, line 251: Is 100ms too high? This change implies that no new writes can happen via the EventStreamWriter for a max of  100ms since it holds the `writeFlushLock`.

RaulGracia: @shrids you mean no writes to "any" segment can happen while the sleep is executed? If so, this would be quite disruptive for write performance when scales are frequent. And the fact of making the sleep lower I don't know if is going to improve much the situation, right? Could the "sleep" be out of the synchronized block so writes to other segments can make progress?

shrids: Yes, if at time `t0` the sleep is invoked then writes can happen only at time`t0 + 100ms` when the `writeFlushLock `is released. This is the case even if the sealed segment handling is completed at `t0 + 1ms`.

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -224,12 +224,32 @@ public void flush() {
...
+     * This is used by flush to optimistically wait for the `handleLogSealed` work to be completed
+     * to prevent a busy loop while waiting for the segments to be updated following a sealed
+     * segment. Please note there are no guarantees about when this method returns. It can be
+     * thought of as a sleep.
+     */
+    @GuardedBy("writeFlushLock")
+    private void tryWaitForSuccessors() {
+        while (!sealedSegmentQueue.isEmpty()) {
+            // A background thread should be waking up to process things. Give it a moment.
+            Exceptions.handleInterrupted(() -> Thread.sleep(10));
```
fpj, line 245: A clarification question: what would we need to do to wait on the processing of the sealed segment rather than set a fixed amount of time to wait in each loop iteration?

tkaitchuck: This loop is blocking on the `sealedSegmentQueue` becoming empty. Once the queue is empty and the `writeSealLock` is released we should return. However there is no built in way to block on a queue becoming empty. Hence it is polling. We could probably make a custom data structure to provide this capability if we want.

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -250,12 +250,32 @@ public void flush() {
                         // available writers.
                         success = false;
                         log.warn("Flush on segment {} failed due to {}, it will be retried.", writer.getSegmentName(), e.getMessage());
+                        tryWaitForSuccessors();
```
shrids, line 253: The same issue can happen during the invocation of `io.pravega.client.stream.impl.EventStreamWriterImpl#close`

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -250,12 +250,32 @@ public void flush() {
...
+     * segment. Please note there are no guarantees about when this method returns. It can be
+     * thought of as a sleep.
+     */
+    @GuardedBy("writeFlushLock")
+    private void tryWaitForSuccessors() {
+        while (!sealedSegmentQueue.isEmpty()) {
+            // A background thread should be waking up to process things. Give it a moment.
+            Exceptions.handleInterrupted(() -> Thread.sleep(10));
+        }
+        synchronized (writeSealLock) {
```
shrids, line 273: No new writes are allowed once `writeFlushLock` is held by flush(),this implies that writer need not try to acquire the `writeSealLock ` to verify that retransmit to sealed segments are complete, right ?

Also, `sealedSegmentQueue.isEmpty()` check would suffice in the scenario where retransmit of inflight events face a sealed segment exception..

shrids (changes requested):

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
tkaitchuck: We want to make sure that when `flush` is called it does not return until all the segments have been flushed. It aquires the `writeFlushLock` to prevent future writes. However there may be a concurrent seal. We just need to keep retrying calling flush on the writers until all have flushed. The previous logic was "correct" however in the event there was an ongoing seal it was wasteful because it would keep trying to flush a sealed segment over and over. The solution here it to attempt to block on the seal being completed. `sealedSegmentQueue.isEmpty()` Would tell us that no seal operation is pending. This is a necessary but not a sufficient criteria, because there could be still one in progress. The `writeSealLock` is not needed to do anything here, (Notice nothing is done inside of the lock) it's just done to queue the flush thread on an ongoing seal if one is indeed in progress.

shrids: ok 👍

shrids (commented):
The changes look good apart from one comment regarding applying the same optimization during close() API invocation ...

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
tkaitchuck: Added

shrids (approved):
LGTM.

shrids:
System test run is successful with this changeset. ref: `/test-pravega-continuous-1/1313/`

fpj:
Looks a lot better, thanks for changing it, @tkaitchuck . This test case failed:

```
io.pravega.test.integration.endtoendtest.EndToEndTruncationTest > testWriteDuringScaleAndTruncation FAILED
```

and I also noticed that @RaulGracia had this PR merged recently: https://github.com/pravega/pravega/issues/5599

Do we need to reopen the issue or is it the case that the PR is not merged to this PR branch?

fpj:
I restarted the build and I'm seeing three integration tests failing:

```
io.pravega.test.integration.ClusterWrapperTest > writeAndReadBackAMessageWithTlsAndAuthOn FAILED
io.pravega.test.integration.endtoendtest.EndToEndTruncationTest > testWriteDuringScaleAndTruncation FAILED
io.pravega.test.integration.StreamMetricsTest > testStreamsAndScopesBasicMetricsTests FAILED
```

I'm not following other builds, are the integration tests known to be this flakey?

RaulGracia:
System tests passed `test-pravega-continuous-1#1389`.

PR #4971, opened 2020-07-24
Issue 4972: Set addLombokGeneratedAnnotation in lombok.config

### LINKED ISSUE #4972: Lombok-generated methods are not exempted from code coverage
**Problem description**
Lombok-generated methods not being explicitly tested artificially reduces code coverage metrics.

**Problem location**
Wherever `@Data`, `@Builder` and other Lombok annotations are used.

Set `lombok.addLombokGeneratedAnnotation` in lombok.config.

derekm:
**Change log description**  
Adds `lombok.addLombokGeneratedAnnotation = true` to lombok.config

**Purpose of the change**  
Improve code coverage. Fixes #4972.

**What the code does**  
Adds `@LombokGenerated` to all lombok methods, so these methods are exempted from code coverage metrics.

**How to verify it**  
This PR should improve code coverage.


derekm:
0.41% coverage improvement overall is a small improvement overall, but this setting may help devs focus on testing the right things in PRs.

RaulGracia:
@tkaitchuck @andreipaduroiu @fpj just wondering if it would be better merging the NoNetty PR before this one, or otherwise the coverage target could increase, which can delay that PR from being merged as more test would require to be added.

andreipaduroiu:
@RaulGracia We should not be prioritizing PRs based on "coverage mergers". If we have a problem with the code coverage, we should be fixing codecov's configuration so it doesn't get in our way so much.

derekm:
@RaulGracia @andreipaduroiu -- @tkaitchuck can use a variant of this PR to set a patch threshold (how much percents are allowed to drop): https://github.com/pravega/pravega/pull/4974 If he wants to add this changes to his branch to test for success & then merge, then we can close my PR. Or we can update my PR to include a patch threshold (allowable drop percentage), and Tom can rebase his PR onto it.

I think Codecov isn't taking the aggregate percent, but is judging all component percentages, some of which are negative.

PR #4974, opened 2020-07-24
Issues 4973, 4997: Make .codecov.yml valid, set 80% patch target

### LINKED ISSUE #4973: .codecov.yml file is invalid
**Problem description**
.codecov.yml doesn't validate against Codecov tools

**Problem location**
`$ curl -X POST --data-binary @.codecov.yml https://codecov.io/validate`

> Path: coverage->status->project->default
> Wrong key 'against' in {'against': 'parent', 'target': 'auto', 'threshold': '0.5%', 'branches': ['master']}

Remove invalid key.

### LINKED ISSUE #4997: .codecov.yml needs sensible project & patch targets or thresholds
**Problem description**
Builds occasionally fail when PR status check `codecov/patch` doesn't meet automatic targets.

There is an existing `codecov/project` threshold of 0.5%, but it is being ignored because .codecov.yml is invalid (see #4973).

**Problem location**
https://github.com/pravega/pravega/blob/master/.codecov.yml#L11

We can increase the threshold that allows percentage drops against the automatic target, or we can use a fixed target of, say, 80%.

derekm:
**Change log description**  
Remove invalid `against` key. Also remove keys set to redundant default values. Add a 'patch' stanza to enforce an 80% code coverage target per branch.

**Purpose of the change**  
Fixes #4973, fixes #4997

**What the code does**  
.codecov.yml passes validation, which activates project threshold, and sets patch target to 80%

**How to verify it**  
`curl -X POST --data-binary @.codecov.yml https://codecov.io/validate`
This PR should reflect any side-effects to making the file valid again.


derekm:
> You need to amend the commit to add the sign-off. Also, for some reason this PR did not receive the report. I can't tell whether it is related to the changes here.

Connection timed out when uploading to codecov.io: https://travis-ci.org/github/pravega/pravega/jobs/711340911#L7104

andreipaduroiu:
@fpj: @derekm has fixed the DCO issue, which is why you blocked this PR. Can you please unblock it so we can get it merged in?

andreipaduroiu:
@derekm The problem I see with allowing drops of 0.5% is that it takes only 10 PRs to drop it by (just under) 5%. 

Is there a way to enforce a minimum coverage per PR patch, such as 80%? That will make sure our coverage will never drop below that. 

It took us a long time to get to our coverage today, and I do not want to see that wasted.

.codecov.yml
```diff
@@ -11,11 +11,10 @@ coverage:
...
     project:
       default:
-        against: parent
-        target: auto
         threshold: 0.5%
-        branches:
-          - master
+    patch:
+      default:
+        threshold: 0.5%
```
fpj, line 17: What about considering some of the parameters in the `coverage.status` section here:

https://docs.codecov.io/docs/codecovyml-reference

fpj:
Looking more closely, I think this PR is only proposing changes to get the validation to pass. If that's the case, then we should create separate issues for the other configuration changes we are proposing. That would help with unblocking these changes.

.codecov.yml
derekm: That's what this PR does. Any others besides "threshold"?

derekm:
> @derekm The problem I see with allowing drops of 0.5% is that it takes only 10 PRs to drop it by (just under) 5%.
> 
> Is there a way to enforce a minimum coverage per PR patch, such as 80%? That will make sure our coverage will never drop below that.
> 
> It took us a long time to get to our coverage today, and I do not want to see that wasted.

@andreipaduroiu - I've set codecov/patch target parameter to 80%.

.codecov.yml
derekm: I've set `target` to 80%.

derekm:
> we should create separate issues for the other configuration changes we are proposing. That would help with unblocking these changes.

@fpj, I created Issue #4997 for covering the threshold or target configs. I've set patch target to 80% as per @andreipaduroiu's suggestion.

andreipaduroiu:
@fpj Can you take another look at this PR please?

derekm:
> Am I correct in my understanding that this would allow any change as long as it does not bring the whole project's coverage under 80%? If so I don't think that is acceptable as it could allow a large reduction in coverage without failing the build

@tkaitchuck - `codecov/patch` check is supposed to be the coverage of the lines affected by the PR: https://docs.codecov.io/docs/commit-status#patch-status

Assuming the patch is entirely new code, it will require at least 80% coverage of that new code. Otherwise the target is the current overall project coverage (new code must be covered as much as all other code is).

I have a hard time understanding what `codecov/project` is, though, because project and patch statuses for a fully-rebased branch should have the same results, but they almost never do! PRs that don't even touch code and are missing not commits in master will still say project coverage has changed, which I don't get.

derekm:
In other words, if I understand it correctly, if all new code only ever tests the minimum of 80%, then the overall project coverage will approach 80% (from our current 84.7%). In my understanding, this is preferable to @andreipaduroiu rather than allowing drops in percentage from the current target, where subsequent drops could bring project coverage below 80%.

andreipaduroiu:
@derekm  To be clear, what I want is something like:
- **Diff** coverage at least 80% (this is the number of lines that have been touched)
    - This helps small PRs that may not be able to test every file. These PRs also are unlikely to make large dents in the overall coverage.
- **Overall coverage** must not decrease by more than 0.5%
    - This makes sure that large PRs do not significantly decrease coverage


Combined, these rules should keep our coverage stable and also prevent us from going overboard in overtesting just for the sake of coverage.

derekm:
> * **Diff** coverage at least 80% (this is the number of lines that have been touched)
> * **Overall coverage** must not decrease by more than 0.5%

That's what this PR achieves. However, notice that the `codecov/project` check has stopped being expected in recent PRs. Maybe @fpj turned that check off elsewhere. (My Actions CI PR #4994 uses the Codecov Action to post `codecov/project` statuses, but you can also see they're not a required check anymore.)

derekm:
Previously, we were running two buids, and I suspect that was causing confusion as increases looked like decreases across the two builds.

@tkaitchuck also points out due to timing issues, we have non-determinism in our tests with different code paths being traversed across test runs, causing which lines of code are considered covered to change slightly from build to build. This explains why we've see PRs that don't touch code and aren't missing commits from master still show coverage changes at the project level.

derekm:
To reduce and eliminate non-determinism in our project coverage reports, we need to increase coverage of the "Impacted Files" in PRs such as this one. If unit tests can cover affected code paths, then it won't matter whether larger integration tests in the same suite cover those paths or not.

tkaitchuck:
Ok, sounds good. We might consider decreasing the .5% because it won't do much as is. It would take a **very** large PR with 80% coverage to bring our average of 84 down half a percent. (Doing the math: that PR would have to touch at minimum 15% of all the lines in our codebase.) So perhaps .05% would be more appropriate.

derekm:
> We might consider decreasing the .5% ... perhaps .05% or .02% would be more appropriate.

Perhaps we should revisit `codecov/project` settings after we re-enable it and watch it for a bit.

As for PRs failing right now, this PR https://github.com/pravega/pravega/pull/4958 needs either a 0.22% patch threshold or the 80% patch target proposed here.
image

.codecov.yml
```diff
@@ -11,11 +11,10 @@ coverage:
...
     project:
       default:
-        against: parent
-        target: auto
         threshold: 0.5%
-        branches:
-          - master
+    patch:
+      default:
+        target: 80%
```
shrids, line 17: +1; This implies the diff coverage should be 80 % (auto implies PR base branch coverage is used )  and overall coverage can be reduced by 0.5%

derekm:
> Could you please address this comment and reduce the threshold accordingly?
> 
> [#4974 (comment)](https://github.com/pravega/pravega/pull/4974#issuecomment-669482209)

Sure... I suggested we should mess with that number after we re-enable the codecov/project check (which is presently disabled). But we can mess with that number now. The larger of the two suggested numbers seems safer (0.05%). We can always increase the number later if it gives us problems when codecov/project check is re-enabled (it will have no effect until then).

PR #4979, opened 2020-07-24
Issue 4978: DeltaIterator NullPointerException.

### LINKED ISSUE #4978: DeltaEntryIterator NullException on Empty TableSegment.
**Problem description**

In a request to read the delta of a TableSegment is called on an empty segment, a NullPointerException can occur. This is because the state is only set if there is some table entries to iterate over. If no entries are found the state is not set, but still called upon in the `thenAccept` block.

**Problem location**

`PravegaRequestProcessor.readTableEntriesDelta`.

Change `DeltaIteratorResult` to accept a default value for `S state`.

co-jo:
**Change log description**  

* This PR adds logic to prevent a NullPointException when a TableSegment has no entries to iterate over (relative to the provided fromPosition). This can occur during two situations: an empty segment with length == 0 or when the fromPosition provided in the request exceeds the segment length.

**Purpose of the change**  

Fixes #4978 

**What the code does**  

* This code makes sure that the `result.state` field is initialized before being accessed in any way. Add a new method that (given the table segment to iterate over and the from position) sets up the state, assuming that no iteration has taken place yet.

The following semantics of the response has changed as follows:
* If the `PravegaRequestProcessor` is provided a `fromPosition` value that exceeds the segment length, it will return a `lastPosition` value equal to the length of the segment. 
  * It may also return `shouldClear` as true (defined using the same condition as before) even though it did not iterate over any entries. 
* An empty segment is treated the same as above.

**How to verify it**  

* Run unit tests. Added new unit tests for both situations.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -838,6 +839,12 @@ public void readTableEntriesDelta(WireCommands.ReadTableEntriesDelta readTableEn
         val timer = new Timer();
         val result = new DeltaIteratorResult<BufferView, Map.Entry<WireCommands.TableKey, WireCommands.TableValue>, DeltaIteratorState>(
                 segment.getBytes().length + WireCommands.TableEntriesRead.HEADER_BYTES);
+        // Guards against two cases: an empty segment, or a starting position that exceeds the segment length.
+        segmentStore.getStreamSegmentInfo(segment, TIMEOUT)
+                .thenAccept(properties -> {
+                    result.setState(TableEntryDeltaIterator.initialState(properties, fromPosition));
```
co-jo, line 845: I'm not quite sure on the performance impact (if any) but certain properties about the Segment are needed to define some `DeltaIteratorState` instance. It did not seem feasible to be able to use the information acquired in `ContainerTableExtensionImpl.newDeltaIterator` as that would require several changes about the semantics of the iteration.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2339,6 +2339,75 @@ void releaseInternal() {
...
+            return new ErrorMessage(in.readLong(), in.readUTF(), ErrorCode.valueOf(in.readInt()));
+        }
+
+        public enum ErrorCode {
+            UNSPECIFIED(-1, null),                                         // indicates un-specified (for backward compatibility
+            ILLEGAL_ARGUMENT_EXCEPTION(0, IllegalArgumentException.class), // indicates an IllegalArgumentException
+            ILLEGAL_STATE_EXCEPTION(1, IllegalStateException.class);       // indicates an IllegalStateException
+
+            private static final Map<Integer, ErrorCode> OBJECTS_BY_CODE = new HashMap<>();
+            private static final Map<Class, ErrorCode> OBJECTS_BY_CLASS = new HashMap<>();
```
co-jo, line 2375: I chose to include a Class -> ErrorCode map because presumably we will need some translation logic to turn a given exception into some error code. Including this logic in here prevents a (potentially) continuously growing if/else-if block in `PravegaRequestProcessor.handleException`.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2339,6 +2339,75 @@ void releaseInternal() {
...
+            out.writeInt(errorCode.getCode());
+            out.writeUTF(message);
+        }
+
+        public static WireCommand readFrom(EnhancedByteBufInputStream in, int length) throws IOException {
+            return new ErrorMessage(in.readLong(), in.readUTF(), ErrorCode.valueOf(in.readInt()));
+        }
+
+        public enum ErrorCode {
+            UNSPECIFIED(-1, null),                                         // indicates un-specified (for backward compatibility
```
co-jo, line 2370: While this may increase the risk of a `NullPointerException`, `null` appears to be a valid return for unknown causes (see: https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#getCause--).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -975,8 +975,9 @@ private Void handleException(long requestId, String segment, long offset, String
             invokeSafely(connection::send, new WireCommands.TableKeyBadVersion(requestId, segment, clientReplyStackTrace), failureHandler);
         } else {
             logError(requestId, segment, operation, u);
-            connection.close(); // Closing connection should reinitialize things, and hopefully fix the problem
-            throw new IllegalStateException("Unknown exception.", u);
+            invokeSafely(connection::send,
```
andreipaduroiu, line 978: Here you are changing the behavior for this class. You should make sure that `ErrorMessage` only captures those exceptions that it can properly send a message back for, and let the rest fall down to this default behavior. We have a lot of code (on the other side of the Wire Protocol) that relies on these connections to be closed when something unexpected happens.

It is true that we will eventually want to "fix" that code to understand proper error codes, but that is not a task I believe you are willing to undertake now, are you? It's OK to file  follow up issue (not necessarily for yourself) to look at this new Wire Command and make any proper changes (like handling specific errors, etc.) - we can do that later.

co-jo: Sure thing, thanks for pointing that out. I'll create an issue for addressing this.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -973,6 +973,11 @@ private Void handleException(long requestId, String segment, long offset, String
         } else if (u instanceof BadKeyVersionException) {
             log.warn(requestId, "Conditional update on Table segment '{}' failed due to bad key version.", segment);
             invokeSafely(connection::send, new WireCommands.TableKeyBadVersion(requestId, segment, clientReplyStackTrace), failureHandler);
+        } else if (errorCodeExists(u)) {
+            log.warn(requestId, "Operation on Table segment '{}' failed due to a {}.", segment, u.getClass());
```
andreipaduroiu, line 977: This may not necessarily be on a `TableSegment`. I'd remove it (leave the rest).

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2339,6 +2339,75 @@ void releaseInternal() {
...
+            ILLEGAL_ARGUMENT_EXCEPTION(0, IllegalArgumentException.class), // indicates an IllegalArgumentException
+            ILLEGAL_STATE_EXCEPTION(1, IllegalStateException.class);       // indicates an IllegalStateException
```
andreipaduroiu, line 2372: Is there a reason (for this PR) that `IllegalStateException` is also included here? The old behavior was that it would close the connection in this case.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessorTest.java
```diff
@@ -1017,6 +1017,72 @@ public void testGetTableEntries() throws Exception {
         assertTrue(keyVersions.containsAll(getTableEntriesIteratorsResp.getEntries().getEntries().stream().map(e -> e.getKey().getKeyVersion()).collect(Collectors.toList())));
     }
 
+    @Test
+    public  void testReadTableEntriesDeltaEmpty() throws  Exception {
```
andreipaduroiu, line 1021: You have 2 extra spaces here (before `void` and after `throws`. Same with the methods below.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
co-jo: That's a good question.. No.. Will remove.

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -80,6 +83,18 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
             log.warn("Auth token check failed on segment {} with {}", segmentId, authTokenCheckFailed);
             closeConnection(new AuthenticationException(authTokenCheckFailed.toString()));
         }
+
+        @Override
+        public void errorMessage(WireCommands.ErrorMessage errorMessage) {
```
andreipaduroiu, line 88: It should be enough to implement this once on `FailingReplyProcessor`; if you observe - all the classes where you overrode this method do more or less the same thing. You can just write it once.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1644,6 +1644,88 @@ public int getCode() {
...
+            out.writeLong(requestId);
+            out.writeUTF(message == null ? "" : message);
+            out.writeInt(errorCode.getCode());
+        }
+
+        public static WireCommand readFrom(EnhancedByteBufInputStream in, int length) throws IOException {
+            return new ErrorMessage(in.readLong(), in.readUTF(), ErrorCode.valueOf(in.readInt()));
+        }
+
+        public RuntimeException getThrowableException() {
```
andreipaduroiu, line 1674: Looks like this is not covered by unit tests. Please add unit tests for this class.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1644,6 +1644,88 @@ public int getCode() {
...
+            }
+        }
+
+        @Override
+        public boolean isFailure() {
+            return true;
+        }
+
+        public enum ErrorCode {
+            UNSPECIFIED(-1, RuntimeException.class),                                         // indicates un-specified (for backward compatibility
```
andreipaduroiu, line 1689: Please fix the comment indentation.

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -59,6 +60,8 @@ public void process(Reply reply) {
                 }
             } else if (reply instanceof WireCommands.WrongHost) {
                 closeConnection(new ConnectionFailedException(reply.toString()));
+            } else if (reply instanceof WireCommands.ErrorMessage) {
+                    errorMessage((ErrorMessage) reply);
```
andreipaduroiu, line 64: Please indent this line like the other ones above and below.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -983,6 +988,11 @@ private Void handleException(long requestId, String segment, long offset, String
         return null;
     }
 
+    private boolean errorCodeExists(Throwable e) {
+        val errorCode = WireCommands.ErrorMessage.ErrorCode.valueOf(e.getClass());
+        return (errorCode == WireCommands.ErrorMessage.ErrorCode.UNSPECIFIED) ? false : true;
```
andreipaduroiu, line 993: You don't need `? :` here. Change `==` to `!=` and return the result.

andreipaduroiu (approved):

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
co-jo: 🤦.. rookie mistake.

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -59,6 +60,8 @@ public void process(Reply reply) {
                 }
             } else if (reply instanceof WireCommands.WrongHost) {
                 closeConnection(new ConnectionFailedException(reply.toString()));
+            } else if (reply instanceof WireCommands.ErrorMessage) {
+                errorMessage((ErrorMessage) reply);
```
tkaitchuck, line 64: This is not going to propagate up to the caller.

client/src/test/java/io/pravega/client/connection/impl/RawClientTest.java
```diff
@@ -78,6 +79,24 @@ public void testRequestReply() throws InterruptedException, ExecutionException,
...
+        PravegaNodeUri endpoint = new PravegaNodeUri("localhost", -1);
+        @Cleanup
+        MockConnectionFactoryImpl connectionFactory = new MockConnectionFactoryImpl();
+        @Cleanup
+        MockController controller = new MockController(endpoint.getEndpoint(), endpoint.getPort(), connectionFactory, true);
+        ClientConnection connection = Mockito.mock(ClientConnection.class);
+        connectionFactory.provideConnection(endpoint, connection);
+        Segment segment = new Segment("scope", "testHello", 0);
+        @Cleanup
+        RawClient rawClient = new RawClient(controller, connectionFactory, segment);
```
tkaitchuck, line 93: This is not used. A request should be sent and the reply returned should be verified.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/FailingReplyProcessor.java
```diff
@@ -204,4 +204,15 @@ public void tableEntriesRead(WireCommands.TableEntriesRead tableEntriesRead) {
     public void tableEntriesDeltaRead(WireCommands.TableEntriesDeltaRead tableEntriesDeltaRead) {
         throw new IllegalStateException("Unexpected operation: " + tableEntriesDeltaRead);
     }
+
+    @Override
+    public void errorMessage(WireCommands.ErrorMessage errorMessage) {
```
tkaitchuck, line 209: All of the classes that extend FailingReplyProcessor should be modifed to properly handle the error.

tkaitchuck (changes requested):

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/FailingReplyProcessor.java
co-jo: What constitutes 'properly handling the error?

client/src/test/java/io/pravega/client/connection/impl/RawClientTest.java
co-jo: Instantiation the `RawClient` is required as it sets up the connection to provided the `ReplyProcessor`. I chose not to actually send a request because of the complications of implicitly relying on segmentstore logic to throw back the expected error.

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
co-jo: I might be misunderstanding, but it seems to propogate properly on like 97 in `RawClientTests`.

tkaitchuck (approved):

PR #4983, opened 2020-07-27
Issue 4982: Fix batch calculation in client

### LINKED ISSUE #4982: Revert adding appendsInTime in batch size calculation
**Problem description**
In PR #4873, there are some changes related to the batching algorithm. During our performance experiments, we mentioned that the change in the calculation `appendsInBatch` was inducing higher latency for writes with low throughput rates and multiple segments. The actual change to address this comment was similar but not exactly that what was proposed, and it looks like the issue still persists in `master` now.

**Problem location**
Client batching algorithm.

Use the previous approach to calculate `appendsInBatch`.

RaulGracia:
**Change log description**  
Reverted calculation of `appendsInBatch` as it was before to fix the latency problem with multiple segments and low throughput rates.

**Purpose of the change**  
Fixes #4982.

**What the code does**  
This PR addressed the comment in https://github.com/pravega/pravega/pull/4873#discussion_r456594915 as it was initially validated; that is, completely reverting the calculation of `appendsInBatch` as it was before.

Note that while the batching algorithm can be improved in the future, this provides us a good baseline for 0.8.

**How to verify it**  
Validated that latency now is good for multiple segments. All other tests should be passing as before.


client/src/test/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerTest.java
```diff
@@ -68,7 +68,7 @@ public void testReasonableLowerBound() {
                 tracker.recordAck(i-100);
             }
         }
-        assertTrue(tracker.getAppendBlockSize() >= size * 50);
```
tkaitchuck, line 71: If this test is failing something is wrong.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -89,8 +89,7 @@ public int getAppendBlockSize() {
         }
         double nanosPerAppend = nanosBetweenAppends.getCurrentValue();
         double appendsInMaxBatchTime = Math.max(1.0, (MAX_BATCH_TIME_MILLIS * NANOS_PER_MILLI) / nanosPerAppend);
-        double appendsInTime = Math.max(1.0, BASE_TIME_NANOS / nanosPerAppend);
```
tkaitchuck, line 92: If the problem here is the max of 1 then why not remove that rather than deleting the value which means that BASE_TIME_NANOS cannot be configured?

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
```diff
@@ -89,8 +89,7 @@ public int getAppendBlockSize() {
...
-        double appendsInBatch = MathHelpers.minMax(appendsOutstanding.getCurrentValue() * OUTSTANDING_FRACTION + appendsInTime, 1.0, appendsInMaxBatchTime);
```
tkaitchuck, line 93: If 1 event is a problem, perhaps a batch size of 0 would be better in these cases. So maybe change this 1 to 0.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
RaulGracia: Done, I kept previous structure but changed `appendsInTime` to be effectively 0 by default, which should be the same as it was before.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
RaulGracia: With the previous change, we do not need to touch this line.

client/src/test/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerTest.java
RaulGracia: As the calculation has slightly changed, it looks like the append block size for this test is a bit lower than 50x, but still pretty close to it (40x or more).

tkaitchuck: Does 45 work?

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
tkaitchuck: Yes, but only because your test sends the exact same event size every time. If there were some slight variation we would want this to be 0, otherwise we would have the same problem you observed.

client/src/test/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerTest.java
RaulGracia: Yes, changed.

client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java
RaulGracia: As I shown, the change you propose seems to do not have a big impact on latency, but the max throughput for small events is significantly reduced. For this reason, I would rather keep the algorithm in its current form (similar to what was before), as it has been tested as a good baseline in a variety of situations. We can address further improvements on it on separate PRs, as the purpose of this PR is only to solve the write latency problem for low event rates and multiple segments.

PR #4984, opened 2020-07-28
Issue 4896: Integration tests for Key Value Tables

### LINKED ISSUE #4896: Add new Integration tests for Key Value Store feature
**Problem description**
Add System test for Key Value Store feature

**Problem location**
io.pravega.test.system

SomeshJoshi19:
Signed-off-by: SomeshJoshi19 <SomeshJoshi19@gmail.com>

**Change log description**
This PR has system test to verify single & multiple table entries over conditional update and removal operations along with delete KVT test.

**Purpose of the change**
Fixes #4896

**What the code does**

* retrieve KVP after deleting KVT: manger.deleteKeyValueTable()
* Create a Stream and KVTable with same name inside the same scope. 
* Create KVTable with 0 partitions.
* Create-Delete-ReCreate a KVTable with same name inside a given scope. 
* Listing valid/invalid KVT and Stream entries under same scope.
* Combined tests for validation of deleting stream and kvt with same name under same scope and 
rechecking IO read/write operations over the same entities.

**How to verify it**  
All the integration tests present under this file must pass

### DIFF
--- a/test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
+++ b/test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
@@ -1,0 +1,769 @@
+/**
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.test.integration;
+
+import io.grpc.stub.StreamObserver;
+import io.pravega.client.ClientConfig;
+import io.pravega.client.KeyValueTableFactory;
+import io.pravega.client.admin.KeyValueTableInfo;
+import io.pravega.client.admin.ReaderGroupManager;
+import io.pravega.client.admin.impl.ReaderGroupManagerImpl;
+import io.pravega.client.connection.impl.ConnectionFactory;
+import io.pravega.client.connection.impl.ConnectionPool;
+import io.pravega.client.connection.impl.ConnectionPoolImpl;
+import io.pravega.client.connection.impl.SocketConnectionFactoryImpl;
+import io.pravega.client.control.impl.Controller;
+import io.pravega.client.control.impl.ModelHelper;
+import io.pravega.client.stream.Stream;
+import io.pravega.client.stream.ScalingPolicy;
+import io.pravega.client.stream.Serializer;
+import io.pravega.client.stream.EventWriterConfig;
+import io.pravega.client.stream.EventStreamWriter;
+import io.pravega.client.stream.EventStreamReader;
+import io.pravega.client.stream.ReaderGroupConfig;
+import io.pravega.client.stream.ReaderConfig;
+import io.pravega.client.stream.StreamConfiguration;
+import io.pravega.client.stream.impl.ClientFactoryImpl;
+import io.pravega.client.stream.impl.JavaSerializer;
+import io.pravega.client.tables.KeyValueTable;
+import io.pravega.client.tables.KeyValueTableClientConfiguration;
+import io.pravega.client.tables.KeyValueTableConfiguration;
+import io.pravega.client.tables.Version;
+import io.pravega.client.tables.TableEntry;
+import io.pravega.client.tables.impl.KeyValueTableFactoryImpl;
+import io.pravega.client.tables.impl.KeyValueTableTestBase;
+import io.pravega.common.concurrent.Futures;
+import io.pravega.common.tracing.RequestTracker;
+import io.pravega.common.util.ByteArraySegment;
+import io.pravega.controller.server.rpc.auth.GrpcAuthHelper;
+import io.pravega.controller.stream.api.grpc.v1.Controller.DeleteKVTableStatus;
+import io.pravega.controller.server.rpc.grpc.v1.ControllerServiceImpl;
+import io.pravega.segmentstore.contracts.StreamSegmentNotExistsException;
+import io.pravega.segmentstore.contracts.StreamSegmentStore;
+import io.pravega.segmentstore.contracts.tables.TableStore;
+import io.pravega.segmentstore.server.host.handler.PravegaConnectionListener;
+import io.pravega.segmentstore.server.store.ServiceBuilder;
+import io.pravega.segmentstore.server.store.ServiceBuilderConfig;
+import io.pravega.test.common.AssertExtensions;
+import io.pravega.test.common.TestUtils;
+import io.pravega.test.common.TestingServerStarter;
+import io.pravega.test.integration.demo.ControllerWrapper;
+
+import lombok.SneakyThrows;
+import lombok.Cleanup;
+import lombok.extern.slf4j.Slf4j;
+import lombok.val;
+import org.apache.curator.test.TestingServer;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.time.Duration;
+import java.util.Random;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.NoSuchElementException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.ExecutionException;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.apache.commons.lang.RandomStringUtils.randomAlphanumeric;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Integration test for {@link KeyValueTable}s along with {@link Stream}s to check proper functioning
+ * of both primitives when used together using real Segment Store, Controller and connection.
+ */
+@Slf4j
+public class KVTableWithStreamTest extends KeyValueTableTestBase {
+
+    private static final String ENDPOINT = "localhost";
+    private static final String SCOPE = "Scope";
+    private static final KeyValueTableConfiguration DEFAULT_CONFIG = KeyValueTableConfiguration.builder()
+            .partitionCount(5).build();
+    private static final Duration TIMEOUT = Duration.ofSeconds(30);
+    private ServiceBuilder serviceBuilder;
+    private TableStore tableStore;
+    private StreamSegmentStore streamSegmentStore;
+    private PravegaConnectionListener serverListener = null;
+    private ConnectionPool connectionPool;
+    private TestingServer zkTestServer = null;
+    private ControllerWrapper controllerWrapper = null;
+    private Controller controller;
+    private ControllerServiceImpl controllerService;
+    private RequestTracker requestTracker = new RequestTracker(true);
+    private KeyValueTableFactory keyValueTableFactory;
+    private final int controllerPort = TestUtils.getAvailableListenPort();
+    private final String serviceHost = ENDPOINT;
+    private final int servicePort = TestUtils.getAvailableListenPort();
+    private final int containerCount = 4;
+    private final int startKeyRange = 10000;
+    private final int endKeyRange = 99999;
+    private final int valueSize = 10312;
+    private Random random = new Random();
+
+    @Before
+    public void setup() throws Exception {
+        super.setup();
+
+        // 1. Start ZK
+        this.zkTestServer = new TestingServerStarter().start();
+
+        // 2. Start Pravega SegmentStore service.
+        this.serviceBuilder = ServiceBuilder.newInMemoryBuilder(ServiceBuilderConfig.getDefaultConfig());
+        serviceBuilder.initialize();
+        this.tableStore = serviceBuilder.createTableStoreService();
+        this.streamSegmentStore = serviceBuilder.createStreamSegmentService();
+
+        this.serverListener = new PravegaConnectionListener(false, servicePort, this.streamSegmentStore,
+                this.tableStore, executorService());
+        this.serverListener.startListening();
+
+        // 3. Start Pravega Controller service
+        this.controllerWrapper = new ControllerWrapper(zkTestServer.getConnectString(), false,
+                controllerPort, serviceHost, servicePort, containerCount);
+        this.controllerWrapper.awaitRunning();
+        this.controller = controllerWrapper.getController();
+
+        //4. Create Scope
+        this.controller.createScope(SCOPE).get();
+        ClientConfig clientConfig = ClientConfig.builder().build();
+        SocketConnectionFactoryImpl connectionFactory = new SocketConnectionFactoryImpl(clientConfig);
+        this.connectionPool = new ConnectionPoolImpl(clientConfig, connectionFactory);
+        this.keyValueTableFactory = new KeyValueTableFactoryImpl(SCOPE, this.controller, this.connectionPool);
+
+        //5. setting up controller service instance
+        this.controllerService = new ControllerServiceImpl(controllerWrapper.getControllerService(),
+                GrpcAuthHelper.getDisabledAuthHelper(), requestTracker, true, 2);
+    }
+
+
+    @After
+    public void tearDown() throws Exception {
+        this.controller.close();
+        this.connectionPool.close();
+        this.controllerWrapper.close();
+        this.serverListener.close();
+        this.serviceBuilder.close();
+        this.zkTestServer.close();
+        this.keyValueTableFactory.close();
+    }
+
+
+    /**
+     * Create Stream and KVTable with same name under same scope.
+     */
+    @Test
+    public void testCreateStreamKVTable() throws ExecutionException, InterruptedException {
+        String newScope = "NewScope";
+        KeyValueTableInfo newKVT = newKeyValueTableName(newScope);
+        String streamName = newKVT.getKeyValueTableName();
+        int streamSegments = 1;
+        String readerGroupName = "readerGroup";
+        String readerName = "reader";
+        String routingKey = "routingKey";
+        String streamEvent = "Stream event";
+        int numEvents = 100;
+        int writeCount = 0;
+        int readCount = 0;
+
+        final ScalingPolicy scalingPolicy = ScalingPolicy.fixed(streamSegments);
+        final StreamConfiguration config = StreamConfiguration.builder()
+                .scalingPolicy(scalingPolicy)
+                .build();
+
+        assertTrue("The requested Scope must get created", this.controller.createScope(newScope).join());
+
+        assertTrue("The requested Stream must get created", this.controller.createStream(newScope,
+                streamName, config).join());
+
+        assertTrue("The requested KVT must get created",
+                this.controller.createKeyValueTable(newScope, newKVT.getKeyValueTableName(),
+                        DEFAULT_CONFIG).join());
+
+        assertEquals("Table partition value should be same as passed", DEFAULT_CONFIG.getPartitionCount(),
+                Futures.getAndHandleExceptions(this.controller.getCurrentSegmentsForKeyValueTable(newScope, newKVT.getKeyValueTableName()),
+                        RuntimeException::new).getSegments().size());
+
+        assertEquals("The stream segment count should be same as passed", streamSegments,
+                Futures.getAndHandleExceptions(this.controller.getCurrentSegments(newScope, streamName),
+                        RuntimeException::new).getSegments().size());
+
+        try (ConnectionFactory connectionFactory = new SocketConnectionFactoryImpl(ClientConfig.builder().build());
+             ClientFactoryImpl clientFactory = new ClientFactoryImpl(newScope, this.controller, connectionFactory);
+             ReaderGroupManager readerGroupManager = new ReaderGroupManagerImpl(newScope, this.controller, clientFactory)) {
+
+            log.info("Creating writer");
+            @Cleanup
+            final EventStreamWriter<String> writer = clientFactory.createEventWriter(streamName,
+                    new JavaSerializer<>(), EventWriterConfig.builder().build());
+
+            // start writing events to the stream.
+            for (int i = 0; i < numEvents; i++) {
+                log.info("Writing event {}", streamEvent);
+                writer.writeEvent(routingKey, streamEvent);
+                writeCount++;
+                writer.flush();
+            }
+            log.info("Closing writer {}", writer);
+            writer.close();
+
+            // create a reader group.
+            log.info("Creating Reader group : {}", readerGroupName);
+
+            readerGroupManager.createReaderGroup(readerGroupName, ReaderGroupConfig.builder()
+                    .stream(Stream.of(newScope, streamName)).build());
+            log.info("Reader group name {} ", readerGroupManager.getReaderGroup(readerGroupName).getGroupName());
+            log.info("Reader group scope {}", readerGroupManager.getReaderGroup(readerGroupName).getScope());
+
+            log.info("Creating reader");
+            @Cleanup
+            final EventStreamReader<String> reader = clientFactory.createReader(readerName,
+                    readerGroupName,
+                    new JavaSerializer<>(),
+                    ReaderConfig.builder().build());
+
+            // start reading events from the stream.
+            while (!(readCount == writeCount)) {
+                final String eventRead = reader.readNextEvent(SECONDS.toMillis(2)).getEvent();
+                log.info("Reading event {}", eventRead);
+                if (eventRead != null) {
+                    // update if event read is not null.
+                    readCount++;
+                }
+            }
+            log.info("Closing reader {}", reader);
+            reader.close();
+
+            //delete readergroup
+            log.info("Deleting readergroup {}", readerGroupName);
+            readerGroupManager.deleteReaderGroup(readerGroupName);
+        }
+        log.info("All writers have stopped. Event Written Count:{}, Event Read " +
+                "Count: {}", writeCount, readCount);
+        assertEquals("Number of events written & read should be equal", writeCount, readCount);
+
+        log.info("Read write test on stream succeeds");
+
+        KeyValueTableFactory newKeyValueTableFactory = new KeyValueTableFactoryImpl(newScope, this.controller, this.connectionPool);
+        KeyValueTable<Integer, String> newKeyValueTable = newKeyValueTableFactory.forKeyValueTable(newKVT.getKeyValueTableName(),
+                KEY_SERIALIZER, VALUE_SERIALIZER, KeyValueTableClientConfiguration.builder().build());
+
+        String keyFamily = "KeyFamily1";
+        LinkedHashMap<Integer, String> entryMap1 = new LinkedHashMap<>();
+        Integer[] keyArray = new Integer[10];
+        for (int i = 0; i < keyArray.length; i++) {
+            keyArray[i] = getKeyID();
+        }
+        String[] valueArray1 = new String[10];
+        for (int i = 0; i < valueArray1.length; i++) {
+            valueArray1[i] = getValue();
+        }
+
+        // Adding 10 new entries to the Table only if they are not already present.
+        for (int i = 0; i < keyArray.length; i++) {
+            entryMap1.put(keyArray[i], valueArray1[i]);
+        }
+
+        List<Version> versionList1 = newKeyValueTable.putAll(keyFamily, entryMap1.entrySet()).join();
+        log.info("Version list1: {}", versionList1);
+        assertNotNull("Version list1 should not be empty", versionList1);
+        assertEquals("Version list1 size should be same as entry list1 size",
+                versionList1.size(), entryMap1.size());
+
+        List<Integer> keyList = new ArrayList<>();
+        for (int i = 0; i < keyArray.length; i++) {
+            keyList.add(keyArray[i]);
+        }
+        log.info("Key list: {}", keyList);
+        List<TableEntry<Integer, String>> getEntryList1 = newKeyValueTable.getAll(keyFamily, keyList).join();
+        log.info("Get Entry List1: {}", getEntryList1);
+        assertNotNull("Get Entry List1 should not be empty", getEntryList1);
+        assertEquals("Get Entry List1 size should be same as get entry list1 size",
+                getEntryList1.size(), entryMap1.size());
+        List<Integer> entryKeyList = new ArrayList<>();
+        for (Integer key: entryMap1.keySet()) {
+            entryKeyList.add(key);
+        }
+        log.info("entryKeyList: {}", entryKeyList);
+        for (int i = 0; i < keyArray.length; i++) {
+            assertEquals("Corresponding key for getEntryList should be as inserted",
+                    keyList.get(i), getEntryList1.get(i).getKey().getKey());
+            assertEquals("Corresponding key for entryList should be as inserted",
+                    entryKeyList.get(i), getEntryList1.get(i).getKey().getKey());
+        }
+
+        // Comparing each & every entry values of both entrylist and getEntryList.
+        for (int i = 0; i < keyArray.length; i++) {
+            assertEquals("All string values of both entryList1 and getEntryList1 should be equal",
+                    entryMap1.get(entryKeyList.get(i)), getEntryList1.get(i).getValue());
+        }
+        log.info("KVPair operations succeeds over KVTable");
+
+        // seal the stream.
+        CompletableFuture<Boolean> sealStreamStatus = controller.sealStream(newScope, streamName);
+        log.info("Sealing stream {}", streamName);
+        assertTrue(sealStreamStatus.get());
+        // delete the stream.
+        CompletableFuture<Boolean> deleteStreamStatus = controller.deleteStream(newScope, streamName);
+        log.info("Deleting stream {}", streamName);
+        assertTrue(deleteStreamStatus.get());
+        // delete the  scope.
+        CompletableFuture<Boolean> deleteScopeStatus = controller.deleteScope(newScope);
+        log.info("Deleting scope {}", newScope);
+        assertTrue(deleteScopeStatus.get());
+
+        log.info("Successfully completed creating Stream & KVTable with same " +
+                "name under same scope with event read write test on both entities");
+    }
+
+    /**
+     * Tests Creation and deletion of KVTables and Streams +
+     * under different scopes
+     */
+    @Test
+    public void testDeleteKVTableStream() throws ExecutionException, InterruptedException {
+        String newScope1 = "TestScope1-" + randomAlphanumeric(5);
+        String newScope2 = "TestScope2-" + randomAlphanumeric(5);
+        KeyValueTableInfo newKVT1 = newKeyValueTableName(newScope1);
+        KeyValueTableInfo newKVT2 = newKeyValueTableName(newScope2);
+        KeyValueTableConfiguration config = KeyValueTableConfiguration.builder().partitionCount(5).build();
+        String streamName1 = newKVT1.getKeyValueTableName();
+        String streamName2 = newKVT2.getKeyValueTableName();
+        String readerGroupName = "readerGroup";
+        String readerName = "reader";
+        String routingKey = "routingKey";
+        String streamEvent = "Stream event";
+        int numEvents = 100;
+        int writeCount = 0;
+        int readCount = 0;
+
+        assertTrue("The requested Scope must get created", this.controller.createScope(newScope1).join());
+        // Creating KVT1 under Scope1.
+        assertTrue("The requested KVT must get created",
+                this.controller.createKeyValueTable(newScope1, newKVT1.getKeyValueTableName(),
+                        config).join());
+
+        assertEquals("Table partition value should be same as passed", DEFAULT_CONFIG.getPartitionCount(),
+                Futures.getAndHandleExceptions(this.controller.getCurrentSegmentsForKeyValueTable(newScope1,
+                        newKVT1.getKeyValueTableName()), RuntimeException::new).getSegments().size());
+
+        val tableSegments1 = this.controller.getCurrentSegmentsForKeyValueTable(newScope1,
+                newKVT1.getKeyValueTableName()).join();
+
+        for (val s : tableSegments1.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT).join();
+        }
+
+        // Deleting KeyValueTable where Scope does not exist.
+        ResultObserver<DeleteKVTableStatus> result1 = new ResultObserver<>();
+        this.controllerService.deleteKeyValueTable(ModelHelper.createKeyValueTableInfo(newScope2,
+                newKVT1.getKeyValueTableName()), result1);
+        assertEquals("The requested table to delete must not be found for invalid Scope name",
+                DeleteKVTableStatus.Status.TABLE_NOT_FOUND, result1.get().getStatus());
+
+        // Delete KeyValueTable where KVTable does not exist within the Scope.
+        ResultObserver<DeleteKVTableStatus> result2 = new ResultObserver<>();
+        assertTrue("The requested Scope must get created", this.controller.createScope(newScope2).join());
+        this.controllerService.deleteKeyValueTable(ModelHelper.createKeyValueTableInfo(newScope1,
+                newKVT2.getKeyValueTableName()), result2);
+        assertEquals("The requested table to delete must not be found for invalid Table name",
+                DeleteKVTableStatus.Status.TABLE_NOT_FOUND, result2.get().getStatus());
+
+        final ScalingPolicy scalingPolicy = ScalingPolicy.fixed(5);
+        final StreamConfiguration strmConfig = StreamConfiguration.builder()
+                .scalingPolicy(scalingPolicy)
+                .build();
+
+        // Creating Stream1 under Scope1.
+        assertTrue("The requested Stream must get created", this.controller.createStream(newScope1,
+                streamName1, strmConfig).join());
+
+        val strmSegments1 = this.controller.getCurrentSegments(newScope1, streamName1).join();
+        for (val s : strmSegments1.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.streamSegmentStore.getStreamSegmentInfo(s.getScopedName(), TIMEOUT).join();
+        }
+
+        // KVTable and its table segments should get deleted but Stream and its stream segments should exist and +
+        // we should be able to read/write from them.
+        ResultObserver<DeleteKVTableStatus> result3 = new ResultObserver<>();
+        this.controllerService.deleteKeyValueTable(ModelHelper.createKeyValueTableInfo(newScope1,
+                newKVT1.getKeyValueTableName()), result3);
+        assertEquals("The requested table to delete must not be found for invalid Scope name",
+                DeleteKVTableStatus.Status.SUCCESS, result3.get().getStatus());
+
+        for (val s : tableSegments1.getSegments()) {
+            AssertExtensions.assertSuppliedFutureThrows(
+                    "Segment " + s + " has not been deleted.",
+                    () -> this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new
+                            ByteArraySegment(new byte[1])), TIMEOUT),
+                    ex -> ex instanceof StreamSegmentNotExistsException);
+        }
+
+        for (val s : strmSegments1.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.streamSegmentStore.getStreamSegmentInfo(s.getScopedName(), TIMEOUT).join();
+        }
+
+        try (ConnectionFactory connectionFactory = new SocketConnectionFactoryImpl(ClientConfig.builder().build());
+             ClientFactoryImpl clientFactory = new ClientFactoryImpl(newScope1, this.controller, connectionFactory);
+             ReaderGroupManager readerGroupManager = new ReaderGroupManagerImpl(newScope1, this.controller, clientFactory)) {
+
+            log.info("Creating writer");
+            @Cleanup
+            final EventStreamWriter<String> writer = clientFactory.createEventWriter(streamName1,
+                    new JavaSerializer<>(), EventWriterConfig.builder().build());
+
+            // start writing events to the stream.
+            for (int i = 0; i < numEvents; i++) {
+                log.info("Writing event {}", streamEvent);
+                writer.writeEvent(routingKey, streamEvent);
+                writeCount++;
+                writer.flush();
+            }
+            log.info("Closing writer {}", writer);
+            writer.close();
+
+            // create a reader group.
+            log.info("Creating Reader group : {}", readerGroupName);
+
+            readerGroupManager.createReaderGroup(readerGroupName, ReaderGroupConfig.builder()
+                    .stream(Stream.of(newScope1, streamName1)).build());
+            log.info("Reader group name {} ", readerGroupManager.getReaderGroup(readerGroupName).getGroupName());
+            log.info("Reader group scope {}", readerGroupManager.getReaderGroup(readerGroupName).getScope());
+
+            log.info("Creating reader");
+            @Cleanup
+            final EventStreamReader<String> reader = clientFactory.createReader(readerName,
+                    readerGroupName,
+                    new JavaSerializer<>(),
+                    ReaderConfig.builder().build());
+
+            // start reading events from the stream.
+            while (!(readCount == writeCount)) {
+                final String eventRead = reader.readNextEvent(SECONDS.toMillis(2)).getEvent();
+                log.info("Reading event {}", eventRead);
+                if (eventRead != null) {
+                    // update if event read is not null.
+                    readCount++;
+                }
+            }
+            log.info("Closing reader {}", reader);
+            reader.close();
+
+            //delete readergroup
+            log.info("Deleting readergroup {}", readerGroupName);
+            readerGroupManager.deleteReaderGroup(readerGroupName);
+        }
+        log.info("All writers have stopped. Event Written Count:{}, Event Read " +
+                "Count: {}", writeCount, readCount);
+        assertEquals("Number of events written & read should be equal", writeCount, readCount);
+        log.info("Read write test on stream succeeds");
+
+        // Creating Stream2 under Scope2.
+        assertTrue("The requested Stream must get created", this.controller.createStream(newScope2,
+                streamName2, strmConfig).join());
+
+        val strmSegments2 = this.controller.getCurrentSegments(newScope2, streamName2).join();
+        for (val s : strmSegments2.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.streamSegmentStore.getStreamSegmentInfo(s.getScopedName(), TIMEOUT).join();
+        }
+
+        // Creating KVT2 under Scope2.
+        assertTrue("The requested KVT must get created",
+                this.controller.createKeyValueTable(newScope2, newKVT2.getKeyValueTableName(),
+                        config).join());
+
+        assertEquals("Table partition value should be same as passed", config.getPartitionCount(),
+                Futures.getAndHandleExceptions(this.controller.getCurrentSegmentsForKeyValueTable(newScope2,
+                        newKVT2.getKeyValueTableName()), RuntimeException::new).getSegments().size());
+
+        val tableSegments2 = this.controller.getCurrentSegmentsForKeyValueTable(newScope2,
+                newKVT2.getKeyValueTableName()).join();
+
+        for (val s : tableSegments2.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT).join();
+        }
+
+        // seal the stream.
+        CompletableFuture<Boolean> sealStreamStatus = controller.sealStream(newScope2, streamName2);
+        log.info("Sealing stream {}", streamName2);
+        assertTrue(sealStreamStatus.get());
+        // delete the stream.
+        CompletableFuture<Boolean> deleteStreamStatus = controller.deleteStream(newScope2, streamName2);
+        log.info("Deleting stream {}", streamName2);
+        assertTrue(deleteStreamStatus.get());
+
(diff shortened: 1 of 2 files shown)


test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -63,11 +118,19 @@
...
+    private RequestTracker requestTracker = new RequestTracker(true);
     private KeyValueTableFactory keyValueTableFactory;
     private final int controllerPort = TestUtils.getAvailableListenPort();
     private final String serviceHost = ENDPOINT;
     private final int servicePort = TestUtils.getAvailableListenPort();
     private final int containerCount = 4;
+    private final int startKeyRange = 10000;
+    private final int endKeyRange = 99999;
+    private final int valueSize = 10312;
+    private ScheduledExecutorService requestPool = ExecutorServiceHelpers.newScheduledThreadPool(5,
```
andreipaduroiu, line 131: Remove this. Use `executorService()` to get access to the Thread pool.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -63,11 +118,19 @@
     private TestingServer zkTestServer = null;
     private ControllerWrapper controllerWrapper = null;
     private Controller controller;
+    private ControllerServiceImpl controllerService;
```
andreipaduroiu, line 121: You already have Controllers instantiated here. You don't need another one.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -96,6 +160,10 @@ public void setup() throws Exception {
         SocketConnectionFactoryImpl connectionFactory = new SocketConnectionFactoryImpl(clientConfig);
         this.connectionPool = new ConnectionPoolImpl(clientConfig, connectionFactory);
         this.keyValueTableFactory = new KeyValueTableFactoryImpl(SCOPE, this.controller, this.connectionPool);
+
+        //5. setting up controller service instance
+        this.controllerService = new ControllerServiceImpl(controllerWrapper.getControllerService(),
```
andreipaduroiu, line 165: The controller is already set up. Remove this .

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
 
     }
 
+    /**
+     * Test case-42: Insert Same Key into multiple KeyFamilies
+     */
+    @Test
+    public void testSameKeyMutipleKF() {
```
andreipaduroiu, line 268: This test is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        log.info("Table entry1, key: {} and value: {}", tableEntry1.getKey(), tableEntry1.getValue());
+        log.info("Table entry2, key: {} and value: {}", tableEntry2.getKey(), tableEntry2.getValue());
+        log.info("Successfully completed test Insert Same Key into multiple KeyFamilies");
+    }
+
+    /**
+     * Test case-43: Conditional insert KVP, if key not exists already
+     */
+    @Test
+    public void testInsertIfExists() {
```
andreipaduroiu, line 305: This is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+                1, getAllEntry.size());
+        log.info("Only one entry is present from the Fetched list of 'if-exists-test': {}", getAllEntry);
+        log.info("Successfully completed test Conditional insert KVP, if key not exists already");
+    }
+
+    /**
+     * Test case-44: Conditional update single TableEntry value using its version
+     */
+    @Test
+    public void testConditionUpdateKeyValue() {
```
andreipaduroiu, line 342: This is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        assertNotEquals("Corresponding value for entry4 should not remian as entry1",
+                value1, tableEntry4.getValue());
+        log.info("Successfully completed test Conditional update Key value using its version");
+    }
+
+    /**
+     * Test case-45: Conditional delete single TableEntry using its version
+     */
+    @Test
+    public void testConditionDeleteKey() {
```
andreipaduroiu, line 416: This is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        );
+        log.info("Found NoSuchKeyException on attempt to delete already removed key entry");
+        log.info("Successfully completed test Conditional delete single Key using its version");
+    }
+
+    /**
+     * Test case-46: Get Keys version List
+     */
+    @Test
+    public void testGetKeyVersionList() {
```
andreipaduroiu, line 474: This is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+                    versionList.get(i), getEntryList.get(i).getKey().getVersion());
+        }
+        log.info("Successfully completed test Get Key versions List");
+    }
+
+    /**
+     * Test case-47: Conditional insert multiple table entries with KeyFamily
+     */
+    @Test
+    public void testConditionInsertMultiKeys() {
```
andreipaduroiu, line 528: This is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        }
+        log.info("Successfully completed test Conditional insert multiple entries with KeyFamily");
+    }
+
+    /**
+     * Test case-48: Conditional update multiple TableEntry values with KF
+     * using its version
+     */
+    @Test
+    public void testConditionUpdateMultiKeys() {
```
andreipaduroiu, line 583: This is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        }
+        log.info("Successfully completed test Conditional update multiple Key values with KF using its version");
+    }
+
+    /**
+     * Test case-49: Conditional delete multiple TableEntries with KeyFamily
+     * using its version
+     */
+    @Test
+    public void testConditionDeleteMultiKeys() {
```
andreipaduroiu, line 683: This is already covered in the base class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        );
+        log.info("Found NoSuchKeyException on attempt to delete already removed key entries");
+        log.info("Successfully completed test Conditional delete Key with KeyFamily using its version");
+    }
+
+    /**
+     * Test case-50: Delete KVT and check TableEntries are not accessible
+     */
+    @Test
+    public void testDeleteKVTRetriveKVP() {
```
andreipaduroiu, line 762: This test is good, but you can simplify it . You don't need to test all the keys. Doing one is sufficient.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+                    ex -> ex instanceof StreamSegmentNotExistsException);
+        }
+        log.info("Successfully completed test Delete KVT and check KVPs are not accessible");
+    }
+
+    /**
+     * Creates & tests Stream and KVTable with same name under same scope
+     */
+    @Test
+    public void testCreateStreamKVTable() throws ExecutionException, InterruptedException {
```
andreipaduroiu, line 839: This is a Controller-specific test. Please remove it from here or move it somewhere else where we test controller-specific features.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        assertTrue("The requested KVTable creation must fail",
+                this.controller.createKeyValueTable(SCOPE, newKVT.getKeyValueTableName(),
+                        tableConfig).join());
+
+        val segments = this.controller.getCurrentSegmentsForKeyValueTable(newKVT.getScope(), newKVT.getKeyValueTableName()).join();
+        log.info("Zero partition segemnts: {}", segments);
+        for (val s : segments.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT);
```
andreipaduroiu, line 1023: This returns a CompletableFuture. Where are you waiting for it? If you don't do that, you will never catch exceptions.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+                } else {
+                    tableMap.put(kvtInfo.getKeyValueTableName(), 1);
+                }
+            } else {
+                log.error("The iterated Table {} must be under the requested scope rather than {}",
+                        kvtInfo.getKeyValueTableName(), SCOPE, kvtInfo.getScope());
+            }
+        }
+
+        log.info("TableMap obtained from ietrator: {}", tableMap);
```
andreipaduroiu, line 1052: typo

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        log.info("Successfully completed creating Stream & KVTable with same " +
+                "name under same scope with event read write test on both entities");
+    }
+
+    /**
+     * Test creation of KVTable with zero (0) partition in their config value
+     */
+    @Test
+    @Ignore
+    public void testKVTablePartitions() {
```
andreipaduroiu, line 1007: This looks like a Controller-specific test. Please move it.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        log.info("Version list: {}", versionList1);
+
+        List<Integer> keyList1 = new ArrayList<>();
+        for (int i = 0; i < keyArray.length; i++) {
+            keyList1.add(keyArray[i]);
+        }
+        log.info("Key list: {}", keyList1);
+        List<TableEntry<Integer, String>> getEntryList1 = newKeyValueTable1.getAll(keyFamily, keyList1).join();
+        assertNotNull("Get Entry List 1 should not be empty", getEntryList1);
+        assertEquals("Get Entry List 1 size should be same as get entry list size",
```
andreipaduroiu, line 1098: Why didn't you use `AssertExtensions.assertListEquals`?

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+                return observerList.get(index);
+            }, requestPool);
+            requestFutures.thenApply(x -> resultMap.put(index, x.get().getStatus()));
+            requestFutures.thenRun(() -> log.info("Thread request: {}", index));
+            requestList.add(requestFutures);
+        }
+
+        //wait for create Table requests completion
+        Futures.allOf(requestList).get();
+        ExecutorServiceHelpers.shutdown(this.requestPool);
```
andreipaduroiu, line 1202: Why are you shutting down your thread pool in the middle of the test?

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        for (int i = 0; i < 3; i++) {
+            ResultObserver<CreateKeyValueTableStatus> result = new ResultObserver<>();
+            observerList.add(result);
+        }
+
+        HashMap<Integer, CreateKeyValueTableStatus.Status> resultMap = new HashMap<>(3);
+        List<CompletableFuture<ResultObserver<CreateKeyValueTableStatus>>> requestList = new ArrayList<>();
+        for (int i = 0; i < 3; i++) {
+            final int index = i;
+            CompletableFuture<ResultObserver<CreateKeyValueTableStatus>> requestFutures = CompletableFuture.supplyAsync(() -> {
```
andreipaduroiu, line 1190: There is no guarantee that these requests execute in parallel. For all we know, they could execute serially or in parallel. IMO, this test is not valid.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        assertEquals("The same created table should be present in the table list",
+                newKVT.getKeyValueTableName(), tableList.get(0).getKeyValueTableName());
+    }
+
+    /**
+     * Tests Creation and deletion of KVTables and Streams +
+     * under different scopes
+     */
+    @Test
+    public void testDeleteKVtableStream() throws ExecutionException, InterruptedException {
```
andreipaduroiu, line 1243: This test is out of scope for KVTs. What does deleting Streams have to do with KVTs? Please remove the whole test or move it to another class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+
+        for (val s : tableSegments1.getSegments()) {
+            AssertExtensions.assertSuppliedFutureThrows(
+                    "Segment " + s + " has not been deleted.",
+                    () -> this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT),
+                    ex -> ex instanceof StreamSegmentNotExistsException);
+        }
+
+        for (val s : strmSegments1.getSegments()) {
+            this.tableStore.get(s.getScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT);
```
andreipaduroiu, line 1321: Please fix this everywhere in your tests. You are not waiting on the returned Future to complete so this loop isn't providing any value.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        }
+
+        try (ConnectionFactory connectionFactory = new SocketConnectionFactoryImpl(ClientConfig.builder().build());
+             ClientFactoryImpl clientFactory = new ClientFactoryImpl(newScope1, this.controller, connectionFactory);
+             ReaderGroupManager readerGroupManager = new ReaderGroupManagerImpl(newScope1, this.controller, clientFactory)) {
```
andreipaduroiu, line 1326: Are you closing these anywhere?

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+        }
+        log.info("KVPair operations succeeds over KVTable");
+    }
+
+    /**
+     * Test concurrent requests to delete KVTable with same name +
+     * but different config values
+     */
+    @Test
+    @Ignore
```
andreipaduroiu, line 1487: Delete this whole test.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +261,1427 @@ public void testCreateDeleteKeyValueTable() {
...
+                    ex -> ex instanceof StreamSegmentNotExistsException);
+        }
+        log.info("Successfully completed testing concurrent delete table requests for same KVTable");
+    }
+
+    /**
+     * Test listing KVTables under a scope with only Streams and under empty Scope
+     */
+    @Test
+    public void testListKVTables() throws ExecutionException, InterruptedException {
```
andreipaduroiu, line 1567: Do not test Streams in this class. Remove anything Stream-related from here.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Yes, I do have Controller instance but ControllerService instance is also required here as this has overridden methods of creating, deleting and updating KVTables along with ResponseObserver<> parameter which I can use in my tests to get the return status for the requested APIs.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Supporting the above comment to make use of ControllerServiceImp instance.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Okay, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Yes, at line 1373.

SomeshJoshi19: }
        log.info("All writers have stopped. Event Written Count:{}, Event Read " +

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll remove this test as its a duplicate.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Thanks Andrei. Sure, I'll simplify this test by checking the functioning of KVT with single entry instead of 10 entries.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, I'll correct this part of code where ever it is used.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sorry, will correct the typo.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Will do it in the `teardown`

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure, will fix this at all places. Thanks

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Still keeping the same 10 entries as wanted to test whether a KVTable with few entries in it can be deleted directly without any issue.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: If only I use this separate pool of threads then the Concurrent create request for same KVTable test works fine.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+    private RequestTracker requestTracker = new RequestTracker(true);
+    private KeyValueTableFactory keyValueTableFactory;
+    private final int controllerPort = TestUtils.getAvailableListenPort();
+    private final String serviceHost = ENDPOINT;
+    private final int servicePort = TestUtils.getAvailableListenPort();
+    private final int containerCount = 4;
+    private final int startKeyRange = 10000;
+    private final int endKeyRange = 99999;
+    private final int valueSize = 10312;
+    private Random random = new Random();
```
andreipaduroiu, line 120: seed this with a constant

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+
+            log.info("Creating writer");
+            @Cleanup
+            final EventStreamWriter<String> writer = clientFactory.createEventWriter(streamName,
+                    new JavaSerializer<>(), EventWriterConfig.builder().build());
+
+            // start writing events to the stream.
+            for (int i = 0; i < numEvents; i++) {
+                log.info("Writing event {}", streamEvent);
+                writer.writeEvent(routingKey, streamEvent);
```
andreipaduroiu, line 221: Are you sure you don't want to wait on this operation to complete? Check the return value.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+                }
+            }
+            log.info("Closing reader {}", reader);
+            reader.close();
+
+            //delete readergroup
+            log.info("Deleting readergroup {}", readerGroupName);
+            readerGroupManager.deleteReaderGroup(readerGroupName);
+        }
+        log.info("All writers have stopped. Event Written Count:{}, Event Read " +
```
andreipaduroiu, line 259: You can probably fit this on one line.

andreipaduroiu: But I doubt its usefulness. You are checking this on the next line so what's the point of logging them if you check them immediately afterwards?

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+
+        log.info("Read write test on stream succeeds");
+
+        KeyValueTableFactory newKeyValueTableFactory = new KeyValueTableFactoryImpl(newScope, this.controller, this.connectionPool);
+        KeyValueTable<Integer, String> newKeyValueTable = newKeyValueTableFactory.forKeyValueTable(newKVT.getKeyValueTableName(),
+                KEY_SERIALIZER, VALUE_SERIALIZER, KeyValueTableClientConfiguration.builder().build());
+
+        String keyFamily = "KeyFamily1";
+        LinkedHashMap<Integer, String> entryMap1 = new LinkedHashMap<>();
+        Integer[] keyArray = new Integer[10];
```
andreipaduroiu, line 271: You can combine these 3 loops in 1.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+        // Comparing each & every entry values of both entrylist and getEntryList.
+        for (int i = 0; i < keyArray.length; i++) {
+            assertEquals("All string values of both entryList1 and getEntryList1 should be equal",
+                    entryMap1.get(entryKeyList.get(i)), getEntryList1.get(i).getValue());
+        }
+        log.info("KVPair operations succeeds over KVTable");
+
+        // seal the stream.
+        CompletableFuture<Boolean> sealStreamStatus = controller.sealStream(newScope, streamName);
+        log.info("Sealing stream {}", streamName);
```
andreipaduroiu, line 322: Put these logs before initiating the call

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+        // delete the stream.
+        CompletableFuture<Boolean> deleteStreamStatus = controller.deleteStream(newScope, streamName);
+        log.info("Deleting stream {}", streamName);
+        assertTrue(deleteStreamStatus.get());
+        // delete the  scope.
+        CompletableFuture<Boolean> deleteScopeStatus = controller.deleteScope(newScope);
+        log.info("Deleting scope {}", newScope);
+        assertTrue(deleteScopeStatus.get());
+
+        log.info("Successfully completed creating Stream & KVTable with same " +
```
andreipaduroiu, line 333: This is useless. If the test completes, then you already know this. There is no point in logging it.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+
+        for (val s : tableSegments1.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT).join();
+        }
+
+        // Deleting KeyValueTable where Scope does not exist.
+        ResultObserver<DeleteKVTableStatus> result1 = new ResultObserver<>();
+        this.controllerService.deleteKeyValueTable(ModelHelper.createKeyValueTableInfo(newScope2,
```
andreipaduroiu, line 379: Is this an async call? Should you wait?

andreipaduroiu: Same below

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+        for (val s : strmSegments1.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.streamSegmentStore.getStreamSegmentInfo(s.getScopedName(), TIMEOUT).join();
+        }
+
+        // KVTable and its table segments should get deleted but Stream and its stream segments should exist and +
+        // we should be able to read/write from them.
+        ResultObserver<DeleteKVTableStatus> result3 = new ResultObserver<>();
+        this.controllerService.deleteKeyValueTable(ModelHelper.createKeyValueTableInfo(newScope1,
```
andreipaduroiu, line 411: async?

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+
+            log.info("Creating writer");
+            @Cleanup
+            final EventStreamWriter<String> writer = clientFactory.createEventWriter(streamName1,
+                    new JavaSerializer<>(), EventWriterConfig.builder().build());
+
+            // start writing events to the stream.
+            for (int i = 0; i < numEvents; i++) {
+                log.info("Writing event {}", streamEvent);
+                writer.writeEvent(routingKey, streamEvent);
```
andreipaduroiu, line 442: async

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+                    ex -> ex instanceof StreamSegmentNotExistsException);
+        }
+
+        for (val s : tableSegments2.getSegments()) {
+            // We know there's nothing in these segments. But if the segments hadn't been created, then this will throw
+            // an exception.
+            this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT).join();
+        }
+
+        KeyValueTableFactory newKeyValueTableFactory = new KeyValueTableFactoryImpl(newScope2, this.controller, this.connectionPool);
```
andreipaduroiu, line 538: This class implements `AutoCloseable`. If so, you must close it. Add `@Cleanup` to all these.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+        KeyValueTable<Integer, String> newKeyValueTable = newKeyValueTableFactory.forKeyValueTable(newKVT2.getKeyValueTableName(),
+                KEY_SERIALIZER, VALUE_SERIALIZER, KeyValueTableClientConfiguration.builder().build());
+
+        String keyFamily = "KeyFamily1";
+        LinkedHashMap<Integer, String> entryMap1 = new LinkedHashMap<>();
+        Integer[] keyArray = new Integer[10];
```
andreipaduroiu, line 544: This test has a lot of code that is identical to the test above. Can you factor it out and reuse it?

The same comment as above holds: merge these 3 loops in 1.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+            assertEquals("Corresponding key for entryList should be as inserted",
+                    entryKeyList.get(i), getEntryList1.get(i).getKey().getKey());
+        }
+
+        // Comparing each & every entry values of both entrylist and getEntryList.
+        for (int i = 0; i < keyArray.length; i++) {
+            assertEquals("All string values of both entryList1 and getEntryList1 should be equal",
+                    entryMap1.get(entryKeyList.get(i)), getEntryList1.get(i).getValue());
+        }
+        log.info("KVPair operations succeeds over KVTable");
```
andreipaduroiu, line 591: remove

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+        deleteScopeStatus = controller.deleteScope(newScope2);
+        log.info("Deleting scope2 {}", newScope2);
+        assertTrue(deleteScopeStatus.get());
+
+        log.info("Successfully completed testing list KVTables under a scope with " +
+                "only Streams and under empty Scope ");
+    }
+
+    private Integer getKeyID() {
+        Integer keyId = random.nextInt(endKeyRange - startKeyRange) + startKeyRange;
```
andreipaduroiu, line 684: You do not need to declare a variable here.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,769 @@
...
+
+    private String getValue() {
+        String alphaNumericString = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                + "0123456789"
+                + "abcdefghijklmnopqrstuvxyz";
+
+        StringBuilder sb = new StringBuilder(valueSize);
+
+        for (int i = 0; i < valueSize; i++) {
+            int index = (int) (alphaNumericString.length() * Math.random());
```
andreipaduroiu, line 696: Why can't you use the `random` field you have in this class??

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +218,276 @@ public void testCreateDeleteKeyValueTable() {
...
+     */
+    @Test
+    public void testDeleteKVTRetrieveKVP() {
+        final KeyValueTableInfo newKVT = newKeyValueTableName();
+        final KeyValueTable<Integer, String> newKeyValueTable = getKeyValueTable(newKVT);
+        final String keyFamily = "KeyFamily1";
+
+        LinkedHashMap<Integer, String> entryMap1 = new LinkedHashMap<>();
+        Integer[] keyArray = new Integer[10];
+        for (int i = 0; i < keyArray.length; i++) {
```
andreipaduroiu, line 232: combine loops

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +218,276 @@ public void testCreateDeleteKeyValueTable() {
...
+        assertFalse("Request to remove already deleted KVT must not succeed",
+                this.controller.deleteKeyValueTable(SCOPE, newKVT.getKeyValueTableName()).join());
+
+        for (val s : segments.getSegments()) {
+            AssertExtensions.assertSuppliedFutureThrows(
+                    "Segment " + s + " has not been deleted.",
+                    () -> this.tableStore.get(s.getKVTScopedName(), Collections.singletonList(new ByteArraySegment(new byte[1])), TIMEOUT),
+                    ex -> ex instanceof StreamSegmentNotExistsException);
+        }
+        log.info("Successfully completed test Delete KVT and check KVPs are not accessible");
```
andreipaduroiu, line 295: delete

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +218,276 @@ public void testCreateDeleteKeyValueTable() {
...
+    }
+
+    /**
+     * Test creation of KVTable with zero (0) partition in their config value
+     */
+    @Test
+    @Ignore
```
andreipaduroiu, line 302: Why?

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
andreipaduroiu: I don't see this as done.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +218,276 @@ public void testCreateDeleteKeyValueTable() {
...
+            }
+        }
+
+        log.info("TableMap obtained from iterator: {}", tableMap);
+        // checking for duplicate table names within same scope.
+        for (String s: tableMap.keySet()) {
+            assertEquals("There shouldn't exist any duplicate table names within a scope",
+                    1, tableMap.get(s).intValue());
+        }
+        log.info("Successfully tested creation of KVTable with zero (0) partition in their config value");
```
andreipaduroiu, line 354: delete

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +218,276 @@ public void testCreateDeleteKeyValueTable() {
...
+     */
+    @Test
+    public void testRecreateKVTable() {
+        KeyValueTableInfo newKVT = newKeyValueTableName();
+        KeyValueTable<Integer, String> newKeyValueTable1 = getKeyValueTable(newKVT);
+
+        String keyFamily = "KeyFamily1";
+        List<TableEntry<Integer, String>> entryList = new ArrayList<>();
+        Integer[] keyArray = new Integer[10];
+        for (int i = 0; i < keyArray.length; i++) {
```
andreipaduroiu, line 368: combine

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -192,6 +218,276 @@ public void testCreateDeleteKeyValueTable() {
...
+            assertEquals("Corresponding key for entryList should be as inserted",
+                    entryList.get(i).getKey().getKey(), getEntryList3.get(i).getKey().getKey());
+        }
+
+        // Comparing each & every entry values of both entryList and getEntryList.
+        for (int i = 0; i < keyArray.length; i++) {
+            assertEquals("All string values of both entryList and getEntryList3 should be equal",
+                    entryList.get(i).getValue(), getEntryList3.get(i).getValue());
+        }
+        log.info("Successfully completed testing recreation of KVTable with same name");
```
andreipaduroiu, line 455: delete

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: No, not an async call just trying to check the deletion status in case where a KVT does not exists in the requested scope.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: No, not an async call.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: okay, I'll add `@Cleanup` for all `KeyValueTableFactory` objects used in the methods.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: Sure, I'll refactor the code here.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: Its just 1 writer appending some 100 events in 1 stream, do you suggest here to make this operation async ?

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: I'll refactor the code here. Thanks

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: Sure, I'll refactor the code here. Thanks

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: okay.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: okay.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: @andreipaduroiu This test is maintained here to validate KeyValueTable creation with 0(zero) as partition size in order to validate an internal open issue once it gets fixed and merged into the master.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: Sure.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: Sure, I'll refactor the code here. Thanks

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: Sure, I'll refactor the code here. Thanks

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: Sure, I'll correct the code here. Thanks

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: okay.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: okay

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: okay

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
andreipaduroiu: Is there a GitHub Issue to track when this test will be reenabled? Otherwise there is a good chance this test will be forgotten forever.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
andreipaduroiu: No. I am saying that `writer.writeEvent` is an async call. It returns a `CompletableFuture`. This indicates that the event itself is not going to be (guaranteed to be) written until that Future is completed. In your code, you are just initiating the Event write but never waiting for it so there's no guarantee as to what gets written - if anything. Also you will not be notified of any errors during the write (since errors will complete the returned Future in an exceptional way).

Please go through your entire PR and look for any APIs that you invoke which return a `CompletableFuture`. Unless you have a very good reason (in which case please document using a comment), you need to wait on that future to complete, otherwise your test may not be testing anything after all.

This is the reason why pointed out several places in this PR where you are not waiting for returned futures to complete.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
andreipaduroiu: `deleteKeyValueTable` returns a CompletableFuture. See my other comment about async calls.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
andreipaduroiu: see above

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
SomeshJoshi19: @andreipaduroiu There was an internal Jira I created to track this but now a new test addition request has already been raised, so I'll remove my test from here. Thanks
https://github.com/pravega/pravega/issues/5100

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: @andreipaduroiu It is not the same `deleteKeyValueTable` api which we often use rather it is this one `public void deleteKeyValueTable(KeyValueTableInfo request, StreamObserver<DeleteKVTableStatus> responseObserver)` which does not returns anything as there is already a `responseObserver` to check for the completion of request from return status through enums. Thanks

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: same reason as above.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
SomeshJoshi19: okay, got it. I'll complete all the `CompletableFuture` API calls wherever used to make them work correctly as suggested. Thanks @andreipaduroiu

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Integration test for {@link KeyValueTable}s along with {@link Stream}s to check proper functioning
+ * of both primitives when used together using real Segment Store, Controller and connection.
+ */
+@Slf4j
+public class KVTableWithStreamTest extends KeyValueTableTestBase {
```
ravisharda, line 95: Please use non-abbreviated "KeyValue..." here, so that this test class shows up together with the other `KeyValueTableTest` class, which is there in the same package.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+    private final String serviceHost = ENDPOINT;
+    private final int servicePort = TestUtils.getAvailableListenPort();
+    private final int containerCount = 4;
+    private final int startKeyRange = 10000;
+    private final int endKeyRange = 99999;
+    private final int valueSize = 10312;
+    private Random random = new Random(100);
+
+    @Before
+    public void setup() throws Exception {
```
ravisharda, line 123: You can simplify the whole setup code significantly by using the `ClusterWrapper` instead. The `ClusterWrapper` encapsulates the server bootstrap logic and provides a simplified interface. See an example here: 

https://github.com/pravega/pravega/blob/22996fc928d762f886265a3dee75157fac944357/test/integration/src/test/java/io/pravega/test/integration/ClusterWrapperTest.java#L50

For your tests, you could disable Auth and TLS.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+        this.serviceBuilder.close();
+        this.zkTestServer.close();
+        this.keyValueTableFactory.close();
+    }
+
+    /**
+     * Create Stream and KVTable with same name under same scope.
+     */
+    @Test
+    public void testCreateStreamKVTable() throws ExecutionException, InterruptedException {
```
ravisharda, line 172: Please use a more descriptive test name. Something like `createStreamAndKeyValueTableUnderTheSameScope`.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+        String routingKey = "routingKey";
+        String streamEvent = "Stream event";
+        int numEvents = 100;
+
+        final ScalingPolicy scalingPolicy = ScalingPolicy.fixed(streamSegments);
+        final StreamConfiguration config = StreamConfiguration.builder()
+                .scalingPolicy(scalingPolicy)
+                .build();
+
+        assertTrue("The requested Scope must get created", this.controller.createScope(newScope).join());
```
ravisharda, line 188: You could use the utility method in `TestUtils` to create scope and stream: 

https://github.com/pravega/pravega/blob/ede135d5cc72aceb2add3ab6d6cef2f1ac8238fa/test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java#L56

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+        }
+
+        // Case-1: Deleting KeyValueTable where Scope does not exist.
+        ResultObserver<DeleteKVTableStatus> result1 = new ResultObserver<>();
+        this.controllerService.deleteKeyValueTable(ModelHelper.createKeyValueTableInfo(newScope2,
+                newKVT1.getKeyValueTableName()), result1);
+        assertEquals("The requested table to delete must not be found for invalid Scope name",
+                DeleteKVTableStatus.Status.TABLE_NOT_FOUND, result1.get().getStatus());
+
+        // Case-2: Delete KeyValueTable where KVTable does not exist within the Scope.
```
ravisharda, line 263: Different test cases should go into different test methods, so that the failure in one doesn't mask failure in others and each test verifies largely a single behavior, among other things. 

A good mental model of distributing tests according to the single responsibility principle is: `UnitOfWork_StateUnderTest_ExpectedBehavior` (even though we don't really use [Roy Oshreve](https://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html)'s naming convention in Pravega). 

IMO this test case should be broken down into these smaller test cases:
* deleteTableReturnsTableNotFoundWhenTableDoesn'tExist
* deleteTableReturnsTableNotFoundWhenScopeDoesn'tExist

Also, it makes it much easier for others to understand the test when its name specified not only the unit of work (`deleteTable`), but also the `state under test` and `expected outcome`. Reasonably long names for tests (unlike in main classes) is not considered bad, as long as they add value.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+        for (int i = 0; i < keyArray.length; i++) {
+            assertEquals("All string values of both entryList1 and getEntryList1 should be equal",
+                    entryMap1.get(entryKeyList.get(i)), getEntryList1.get(i).getValue());
+        }
+        log.info("KVPair operations succeeds over KVTable");
+    }
+
+    private void cleanUpStream(String scope, String stream)
+            throws ExecutionException, InterruptedException {
+        // seal the stream.
```
ravisharda, line 486: This comment is redundant, as the operation name `sealStream` is self-explanatory. Besides, there is also a log message that says the same thing.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+        log.info("Sealing stream {}", stream);
+        CompletableFuture<Boolean> sealStreamStatus = controller.sealStream(scope, stream);
+        assertTrue(sealStreamStatus.get());
+        // delete the stream.
```
ravisharda, line 490: Same here (and other places). Use comments only to explain what is not easily understood from the code itself.

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+        log.info("Deleting stream {}", stream);
+        CompletableFuture<Boolean> deleteStreamStatus = controller.deleteStream(scope, stream);
+        assertTrue(deleteStreamStatus.get());
```
ravisharda, line 493: In this case, inlining the call would be useful. It'd save lines without reducing the code's readability. 

assertTrue(controller.deleteStream(scope, stream).get());

test/integration/src/test/java/io/pravega/test/integration/KVTableWithStreamTest.java
```diff
@@ -0,0 +1,644 @@
...
+
+        // To read and write few sample events on the stream having same name as the KVTable.
+        readWriteOnStream(newScope, streamName, numEvents, routingKey, streamEvent, readerGroupName, readerName);
+
+        String keyFamily = "KeyFamily1";
+        // inserts and fetches few Table entries for validate basic functioning of KVTable
+        insertAndGetFromKVTable(newScope, newKVT, keyFamily);
+
+        // Cleans up stream resources by sealing existing stream segments & then deletes the stream.
+        cleanUpStream(newScope, streamName);
```
ravisharda, line 213: Do we really need to do these? We run cluster cleanup after each test in `@after` so, IMO it is unnecessary to delete the scope and the streams, as each test is using its own cluster, and it may be reducing the tests' performance.

ravisharda (changes requested):

RaulGracia:
@SomeshJoshi19 is this PR still valid? If not, I can close it.

Issue #4790, opened 2020-05-13
Improvements to Pravega standalone


fpj:
**Problem description**
Pravega standalone (run with `./gradlew startStandalone`) does not provide a good experience to new users for the following reasons:

- It outputs log messages at `DEBUG` level by default and it is really a large number of messages
- It outputs several stack traces, which are false positives, but confuse new users as they don't know whether that's a problem or not
- It doesn't clearly indicate when it has started and if it has started correctly.

**Problem location**
Pravega standalone

**Suggestions for an improvement**
The main goal is to create a good experience for a new user and have the output as clean as possible. The three tasks I suggest:
- [ ]  Increase the log level or even turn off logging to stdout by default.
- [ ] Clean up spurious stack traces (false positives)
- [ ] Make sure to indicate clearly when it has started, it entails detecting that it has started and printing a clear message indicating it

co-jo:
Can you elaborate on observed stack traces and or how to prompt them?

derekm:
Standard output is improved for `./gradlew startStandalone`, but if you download the distribution as instructed in [the Quick Start](https://pravega.io/docs/nightly/getting-started/quick-start/) and run `./bin/pravega-standalone`, then you get the barrage of output again.

abhinb:
@derekm   #6234  fixes the output in ./bin/pravega-standalone and the docker image

PR #5000, opened 2020-07-30
Issue 4790: Pravega Standalone Enhancements

co-jo:
**Change log description**  
* Separates logging into two destinations to provide more visibility to essential information (stdout and now `standalone.log`).
* Highlights services details/config and increases visibility of start message.

**Purpose of the change**  
Partial progress towards #4790.

**What the code does**  
* Makes minor changes to logging in the standalone version.

**How to verify it**  
* No semantic changes.


config/logback.xml
```diff
@@ -11,14 +11,35 @@ You may obtain a copy of the License at
...
+
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
             <charset>UTF-8</charset>
             <Pattern>%d %-4relative [%thread] %-5level %logger{35} - %msg%n</Pattern>
         </encoder>
     </appender>
 
-    <root level="DEBUG">
+    <logger name="io.pravega.local" level="INFO">
```
co-jo, line 29: The idea with this change was to bring forward two things: readiness of various services (zookeeper, controller, segmentstore, etc) and highlight the configuration options of those services. It was separated from normal logging (which now is directed towards `standalone.log`) to ensure visibility of these things.

standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
```diff
@@ -125,12 +126,12 @@ public void run() {
 
             log.info("Starting Pravega Emulator with ports: ZK port {}, controllerPort {}, SegmentStorePort {}",
                     conf.getZkPort(), conf.getControllerPort(), conf.getSegmentStorePort());
-
             localPravega.start();
-
-            System.out.println(
-                    String.format("Pravega Sandbox is running locally now. You could access it at %s:%d", "127.0.0.1",
-                            conf.getControllerPort()));
+            String msg = String.format("# Pravega Sandbox is running locally now. You could access it at %s:%d. #", "127.0.0.1", conf.getControllerPort());
```
co-jo, line 130: A change that is less needed with the changes performed in `logback.xml` but this further brings attention to pravega-standalone is in a ready state.

standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
```diff
@@ -125,12 +126,11 @@ public void run() {
...
                     conf.getZkPort(), conf.getControllerPort(), conf.getSegmentStorePort());
-
             localPravega.start();
-
-            System.out.println(
-                    String.format("Pravega Sandbox is running locally now. You could access it at %s:%d", "127.0.0.1",
-                            conf.getControllerPort()));
+            log.info("");
+            log.info("Pravega Sandbox is running locally now. You could access it at {}:{}.", "127.0.0.1", conf.getControllerPort());
+            log.info("For more detailed logs, see: {}", String.format("%s/%s", System.getProperty("user.dir"), "standalone/standalone.log"));
```
andreipaduroiu, line 132: You don't need to do String format here. the `log.info` arg is a formatted string. Use `{}` instead of `%s`.

config/logback.xml
medvedevigorek: Wouldn't be better to define all the necessary changes to logging for standalone in `logback-standalone.xml` file and use it whenever Standalone starts (like in a gradle file)?  

otherwise these changes will be taking an effect in production as well and I am not sure how they fit K8s environments

andreipaduroiu: Good point. There are a number of different logback files spread throughout our codebase. We have to be sure that the correct one is picked up by Logback at runtime. We should also make sure this change doesn't get picked up when we publish the production jars for releases.

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
```diff
@@ -238,7 +238,9 @@ private void cleanUpZK() {
     }
 
     private void startLocalHDFS() throws IOException {
-        localHdfs = LocalHDFSEmulator.newBuilder().baseDirName("temp").build();
+        String baseDir = "temp";
+        log.info("Starting HDFS Emulator @ {}/", baseDir);
```
eolivelli, line 242: If this is a directory on disk, can we write an absolute path?

config/logback.xml
co-jo: I thought that production environments use the `logback.xml` from `dist/conf/`? Although it does seem good to have a more explicitly named `logback.xml` file.

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
co-jo: The baseDir is used within the context of a temp file so it should be absolute. I will alter the log message to make this clear.

PR #5012, opened 2020-08-01
Issue 5003: (SLTS) - Use per segment locks in BaseMetadataStore.

### LINKED ISSUE #5003: PDP-34 - SLTS : Deadlock in BaseMetadataStore
**Problem description**
After about 5 minutes of run, we hit the deadlock where BaseMetadataStore.get call on key results in a table segment reads which in turn cause the re-entry into BaseMetadataStore.get and waits on the same lock. That essentially locks two thread in a deadlock.
This problem is made worse by clients retrying the same requests again and again and that basically exhausts thread pool.
ReaderWriter lock should be used here.

**Problem location**
BaseMetadataStore.get should use reader lock.

Short term convert lock into reader/writer lock. Use reader lock for get .

sachin-j-joshi:
**Change log description**  
Use per segment lock in BaseMetadataStore to increase read/write parallelism and avoid deadlocks.

**Purpose of the change**  
Fixes #5003 

**What the code does**  

- All transactions in BaseMetadataStore must be serialized for consistency. Currently all transactions serialize on a global lock.

- We can take advantage of domain specific fact that individual segments don't share any chunks between them and operation/transaction on one segment should not affect transactions on other segments in flight. This allows us to not bother locking on chunk keys. Therefore now the transactions are synchronized on per segment basis. In other words transactions on two separate unrelated segments do not need to synchronize.

- The `metadataStore.beginTransaction()` takes an array of string keys which it will use to synchronize. All methods of `ChunkedSegmentStorage` use segment name as key to synchronize/serialize transactions on.

- MetadataStore keeps track on mapping from segment names to locks. It also implements LockWrapper class to acquire locks for a group of segments.

**How to verify it**  
System tests and integration tests should pass.


sachin-j-joshi:
Work in progress.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -527,6 +538,138 @@ protected void setVersion(long version) {
...
     }
 
+    /**
+     * Helper class to encapsulates lock over group of keys.
+     */
+    private static class LockWrapper {
+        /**
+         * Global map to track active locks.
+         */
+        private static final HashMap<String, LockData> LOCKS = new HashMap<>();
```
eolivelli, line 548: Why this is static?
With a 'static' holder we can have bad side effects, especially on tests.
Please consider using am instance field for this and so having an instance field for LockWrapper inside the enclosing class

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -527,6 +538,138 @@ protected void setVersion(long version) {
...
+         */
+        static LockWrapper writeLock(String... keysToLock) {
+            return new LockWrapper(false, keysToLock);
+        }
+
+        /**
+         * Acquires all required locks.
+         */
+        void lock() {
+            for (val key: keysToLock) {
```
eolivelli, line 604: Is there any guarantee on the ordering of the elements of this array?

If we do not set a specific order we can fall into deadlocks

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: Good idea. I'll move it into outer class as instance member.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: In reality almost all operations use only one key. 
And the only operation that uses two keys is Storage::concat which is the last operation on source segment.
So in practice unlikely.
But yes will enforce some order.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -527,6 +542,158 @@ protected void setVersion(long version) {
...
+         * @param keysToLock Keys to lock.
+         * @return LockWrapper instance.
+         */
+        MultiKeyLock writeLock(String... keysToLock) {
+            return new MultiKeyLock(this, false, keysToLock);
+        }
+
+        private LockData addReference(String key) {
+            LockData lockData;
+            synchronized (keyToLockMap) {
```
eolivelli, line 574: If we use a ConcurrentHashMap this block can be expressed as a 'compute' operation,l with the main benefit that we will increase parallelism because in this case the lock is per key (actually per bucket).
With the current approach you have a single global lock

sachin-j-joshi: One problem with that approach is that we are also keeping track of reference counts so that we can efficiently delete unused keys.  Now we could use AtomicInteger for atomic update to the count. But the problem is that removeReference is not atomic anymore. Between decrementing + checking counter value and actually deleting it , another thread could increment reference count and that causes key being deleted incorrectly.

Let me see if double checking solves the issue
Assuming count is AtomicInteger
```
if (0== count.decrementAndGet()) {
    synchronized {
          if (0== count.decrementAndGet()) {
              remove()...
          }
    }
}
```
But otherwise we'll have to build something more complicated. We'll do that only if absolutely necessary.

eolivelli: You can use ConcurrentHashMap.merge, if the computation returns null the entry is remove by the map (IIRC)
see
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#merge-K-V-java.util.function.BiFunction-

eolivelli: no need for an AtomicInteger

sachin-j-joshi: If the key is present and multiple threads access it then we still need to keep track of ref counter. That operation needs to be atomic or we'll have some leaks or worse - premature deletion. 

Enter -
1. Get Value for key, if not present insert.
2. Increase ref count.

Leave 
1. Get Value for key.
2. Decrease ref count, if 0 then delete key.

Thread 1 :  GetValue , doesn't find insert
Thread 2 :  Get Value, finds existing value
Thread 2:  Increment counter
Thread 2: Done . Decrement counter
Thread 2 : Delete key as count ==0;
**Thread 1 : Dealing with value that  is not present in map 
**Thread 3: .GetValue , doesn't find insert...essentially new threads will use new locks not existing one. The key is no more protected against concurrent access.
** Thread 1 : Null Exceptions when releasing 

I couldn't figure out an elegant way to achieve this without use of synchronized block in both addReference and releaseReference. 
So decided to do the simplest possible thing .

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -143,13 +145,14 @@ public BaseMetadataStore() {
     /**
      * Begins a new transaction.
      *
+     * @param keysToLock Array of keys to lock for this transaction.
```
andreipaduroiu, line 148: You need to explain what is the behavior of this method with respect to locks. Is it going to block synchronously? Is it going to throw?

What if only one of the locks can't be acquired? Are the others going to be held while waiting for that one to be acquired?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -527,6 +542,158 @@ protected void setVersion(long version) {
...
+         * Acquires all required locks.
+         */
+        void lock() {
+            for (val key: keysToLock) {
+                // Get existing lock.
+                LockData lockData = lockManager.addReference(key);
+
+                // Lock.
+                Lock lock = isReadOnly ? lockData.lock.readLock() : lockData.lock.writeLock();
+                lock.lock();
```
andreipaduroiu, line 653: What happens if you fail to lock one key? Are you rolling back the locks for already acquired keys?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -527,6 +542,158 @@ protected void setVersion(long version) {
...
+    static class LockData {
+        /**
+         * Underlying ReadWriteLock lock.
+         */
+        final ReadWriteLock lock;
+
+        /**
+         * Reference count.
+         */
+        int count;
```
andreipaduroiu, line 686: A best practice is not to expose this field to the outside (this class isn't even private). You should encapsulate whatever logic you need in this class and expose it via methods.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/MetadataTransaction.java
```diff
@@ -97,15 +97,23 @@
...
      *
      * @param store   Underlying metadata store.
      * @param version Version number of the transactions.
+     * @param keysToLock Array of keys to lock for this transaction.
      */
-    public MetadataTransaction(ChunkMetadataStore store, long version) {
+    public MetadataTransaction(ChunkMetadataStore store, long version, String... keysToLock) {
         this.store = Preconditions.checkNotNull(store, "store");
         this.version = version;
+        this.keysToLock = Preconditions.checkNotNull(keysToLock, "keys");
```
andreipaduroiu, line 116: If you invoke this as `new MetadataTransaction(store, version)` then `keysToUnlock` will be an empty array and never be null.

You should accept is as a String array (`String[] keysToLock`)

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -121,9 +119,13 @@
      * Buffer for reading and writing transaction data entries to underlying KV store.
      * This allows lazy storing and avoiding unnecessary load for recently/frequently updated key value pairs.
      */
-    @GuardedBy("lock")
     private final ConcurrentHashMap<String, TransactionData> bufferedTxnData;
 
+    /**
+     * {@link MultiKeyLockManager} instance that manages locks.
+     */
+    private final MultiKeyLockManager lockManager = new MultiKeyLockManager();
```
andreipaduroiu, line 127: I am not fully convinced that we need both ReadWriteLock and ReentrantReadWriteLock. Why can't you just create a dummy `Object`, map it to a `String` and use `synchronized` on it?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -209,7 +212,9 @@ public void commit(MetadataTransaction txn, boolean lazyWrite, boolean skipStore
         }
         // Step 2 : Check whether transaction is safe to commit.
         // This check needs to be atomic, with absolutely no possibility of re-entry
```
tkaitchuck, line 214: Why? It seems like this class is concerned about cycles somehow, but how is that possible isn't there only one of these?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -209,7 +212,9 @@ public void commit(MetadataTransaction txn, boolean lazyWrite, boolean skipStore
...
-        synchronized (lock) {
+        val writeLock = lockManager.writeLock(txn.getKeysToLock());
+        try {
+            writeLock.lock();
```
tkaitchuck, line 217: This is performing IO while holding a lock which is generally considered bad form, and in our contributing guidelines it says:

> If a blocking operation is performed under lock, it should be very obvious and clear that it is absolutely necessary.

From my point of view this criteria has not been met.

Additionally, can't there be concurrent modification from other processes?

tkaitchuck: For example: Did you consider making this a 'service' which enqueues callso onto a queue and returns completable futures while a background pool performs the IO?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: The underlying TableStore stores data in table segments and any get and put on table segments might result in related pages being read/written from LTS which ends up reentering the same code.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: Serializing transactions is necessary for consistency and also persisting these transactions in actual modifications in TableStore or as records in SystemJournal is also essential, transaction can not succeed if external operation fails. 

I do plan to convert methods in this class into async API in 0.9. Some of this code will change then.
But the main purpose of  this PR is to remove global lock and instead synchronize writes based on per segment lock.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: yes.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: 1. The bigger issue is that reads block each other (if reading metadata about segment ends up in page read it is a different request on a possibly different thread. This will result in a deadlock. )  If there are multiple parallel read requests on same segment (especially possible with metadata segments) then we don't want the read to block each other.

2.  We'll have to use array of objects and the structured as
```
 synchronized (ob1) {
     synchronized (ob2) {
           synchronized (ob3) {
                      ....  {
                             ....  {
                        // Do work
                           }
                      }
           }
      }
}
```
We can do some lambda magic but really synchronized does not solve the problem for multiple concurrent readers

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: No acquiring locks in predefined order (segment name sort order).  Releasing in reverse order.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/MetadataTransaction.java
sachin-j-joshi: added check.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: Updated.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
eolivelli: This is what we have in HerdDB

https://github.com/diennea/herddb/blob/c44c80d8326d2280d207bfa5447f3db4d024742c/herddb-utils/src/main/java/herddb/utils/LocalLockManager.java#L65

sachin-j-joshi: Got it. Thanks
I'll try that.

sachin-j-joshi: I tried that approach but didn't quite get it to work as expected. I got into some weird race conditions that I couldn't reason with (eg Unexpected releasing of non-existent key).  

While synchronize is slower, I think for this specific purpose it is "good enough".
This code is going to be rewritten to be fully async in 0.9 ...so I do not want to invest too much in this part.

eolivelli: Ok.
Better to have something that work now.

Thanks for checking

sachin-j-joshi:
https://github.com/pravega/pravega/issues/5067 
To enable incremental approach and to avoid complexity a few APIs were implemented as synchronous APIs in initial implementation of SLTS.
This is a tracking issue for planned work to convert SLTS APIs to async version.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: https://github.com/pravega/pravega/issues/5067

tkaitchuck (commented):
I don't see any issues, but I don't know the context well enough. @andreipaduroiu  Assigning to you.

tkaitchuck (approved):
Removing my block. (Please don't submit without another reviewer)

sachin-j-joshi:
Closing as this is change will be replaced by https://github.com/pravega/pravega/pull/5200

sachin-j-joshi:
Looks like we might have to resurrect this change.

fpj:
Why did we have to reopen this PR? Could you clarify, @sachin-j-joshi ?

sachin-j-joshi:
I opened it while we were researching solutions to dealock.
This needs to be closed now.

PR #5018, opened 2020-08-03
Issue 5009: (SegmentStore) Fixed Segment Attribute Index Truncated Root Pointer

### LINKED ISSUE #5009: Attribute Index Root Pointer may point to truncated offset
**Problem description**
Every Attribute Index update works along the following steps:
1. Write new data.
2. Truncate head (if compacted)
3. Store Root Pointer in Segment Core Attributes (this is because step 1 above cannot be atomic in LTS and we need to know where to begin reading the BTreeIndex from).

The problem is that we may perform steps 1 and 2 but not 3. As such, the Root Pointer may very well point to a truncated portion of the Attribute Index.

Furthermore there are no retries on this logic. However, even with retries, the root pointer is still pointing to an invalid offset, so we cannot make progress. 

**Problem location**
SegmentAttributeBTreeIndex.java

1. Perform Retries.
2. Work around the invalid Root Pointer. TBD

andreipaduroiu:
**Change log description**  
Fixed a bug in SegmentAttributeBTreeIndex where it would attempt to initialize the BTreeIndex from a truncated offset.

Added logback-test.xml for Segment Store Integration tests. This helps debugging.

**Purpose of the change**  
Fixes #5009.

**What the code does**  
Checking the retrieved Root Pointer against the Attribute Segment's Start Offset (Truncation Offset). If less, it is ignored.
Please refer to the comment(s) in the code for more explanations.

**How to verify it**  
New unit test added to verify fix.
Un-ignored E2E integration test that was sporadically failing due to this problem.


segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
```diff
@@ -771,13 +837,39 @@ public void close() {
...
 
             TestStorage(SyncStorage syncStorage, Executor executor) {
                 super(syncStorage, executor);
                 this.wrappedStorage = syncStorage;
+                this.startOffsets = new ConcurrentHashMap<>();
+            }
+
+            @Override
+            public CompletableFuture<Void> truncate(SegmentHandle handle, long offset, Duration timeout) {
+                // We need to simulate the ChunkStorageProvider (correct) behavior for truncating segments. While the
```
sachin-j-joshi, line 857: Nit:  The new name is now - `ChunkedSegmentStorage`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
```diff
@@ -466,7 +466,20 @@ private long deserializeValue(ByteArraySegment value) {
...
+                    if (rootPointer != BTreeIndex.IndexInfo.EMPTY.getRootPointer() && rootPointer < segmentInfo.getStartOffset()) {
+                        // The Root Pointer is invalid as it points to an offset prior to the Attribute Segment's Start Offset.
+                        // The Attribute Segment is updated in 3 sequential steps: 1) Write new BTree pages, 2) Truncate and
+                        // 3) Update root Pointer.
+                        // The purpose of the Root Pointer is to provide a location of a consistently written update in case
+                        // step 1) above fails (it is not atomic). However, if both 1) and 2) complete but 3) doesn't, then
+                        // it's possible that the existing Root Pointer has been truncated out. In this case, it should be
+                        // safe to ignore it and let the BTreeIndex read the file from the end (as it does in this case).
+                        log.warn("{}: Root Pointer ({}) is below Attribute Segment's StartOffset ({}). Ignoring.", this.traceObjectId, rootPointer, segmentInfo.getStartOffset());
```
eolivelli, line 479: Shall we add something like 'this is usually not a problem' ?

If this is not a problem then probably we could log it at info  level

andreipaduroiu: Good point. Making it `info`.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
```diff
@@ -627,6 +631,68 @@ public void testRecoveryAfterIncompleteUpdate() {
...
+            val value = nextValue++;
+            expectedValues.put(attributeId, value);
+            val updateBatch = Collections.singletonMap(attributeId, value);
+
+            val rootPointer = idx.update(updateBatch, TIMEOUT).join();
+            val startOffset = context.storage.getStreamSegmentInfo(attributeSegmentName, TIMEOUT).join().getStartOffset();
+            if (previousRootPointer >= 0) {
+                context.containerMetadata.getStreamSegmentMetadata(SEGMENT_ID)
+                        .updateAttributes(Collections.singletonMap(Attributes.ATTRIBUTE_SEGMENT_ROOT_POINTER, previousRootPointer));
+                invalidRootPointer |= previousRootPointer < startOffset;
```
sachin-j-joshi, line 676: A comment explaining how root pointer is always set to a previous value to simulate scenario and where truncation happens would be helpful.

sachin-j-joshi: Thanks for adding comment.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
andreipaduroiu: Fixed.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
```diff
@@ -771,13 +841,39 @@ public void close() {
...
                 this.wrappedStorage = syncStorage;
+                this.startOffsets = new ConcurrentHashMap<>();
+            }
+
+            @Override
+            public CompletableFuture<Void> truncate(SegmentHandle handle, long offset, Duration timeout) {
+                // We need to simulate the ChunkedSegmentStorage (correct) behavior for truncating segments. While the
+                // legacy RollingStorage would approximate a StartOffset to an offset at most equal to the requested
+                // Truncation Offset, the ChunkedSegmentStorage is very strict about that, so it will deny any read prior
+                // to that offset. In addition, the ChunkStorageProvide also returns the correct StartOffset as part of
```
sachin-j-joshi, line 860: ChunkStorageProvide  ->  ChunkedSegmentStorage

PR #5044, opened 2020-08-07
Issue 4476: Truncate internal streams to reclaim storage space

### LINKED ISSUE #4476: Retention policy is not configured for internal streams.
**Problem description**

Internal streams (`_system`) do not have a retention policy configured. This causes increased space usage on tier2 
https://github.com/pravega/pravega/blob/b8c18b2ad8118d35f950da08b4edca6314b70941/controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java#L227-L237

```
/mnt/tier2 # date
Tue Jan  7 10:13:39 UTC 2020
/mnt/tier2 # du -sh *
7.3G    _system
1.4T    flink-longevity-0
36.9G   longevity-0
```

**Problem location**
ControllerEventProcessors.

Fix the issue.

shiveshr:

**Change log description**  
Truncate internal streams used by controller periodically using the event position from the event that has been successfully processed. 

**Purpose of the change**  
Fixes #4476 

**What the code does**  
When controller event processors are initialized, we also initialize a periodic loop for each of the event processor groups for truncating the stream by deriving a streamcut from the checkpointed positions of all readers in the readergroup. 

**How to verify it**  
Unit tests added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java
+++ b/controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java
@@ -90,7 +90,7 @@ public class ControllerEventProcessors extends AbstractIdleService implements Fa
     private static final long DELAY = 100;
     private static final int MULTIPLIER = 10;
     private static final long MAX_DELAY = 10000;
-    private static final long TRUNCATION_INTERVAL_MILLIS = Duration.ofMinutes(2).toMillis();
+    private static final long TRUNCATION_INTERVAL_MILLIS = Duration.ofMinutes(10).toMillis();
 
     private final String objectId;
     private final ControllerEventProcessorConfig config;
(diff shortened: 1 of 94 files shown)


controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java
```diff
@@ -288,9 +300,67 @@ public boolean isReady() {
             streamMetadataTasks.initializeStreamWriters(clientFactory, config.getRequestStreamName());
             streamTransactionMetadataTasks.initializeStreamWriters(clientFactory, config);
             tableMetadataTasks.initializeStreamWriters(clientFactory, config.getKvtStreamName());
+
+            long delay = truncationInterval.get();
+            Futures.loop(this::isRunning, () -> Futures.delayedFuture(
```
andreipaduroiu, line 305: You need to add a check inside `truncate` to verify if `isRunning() == true`. Otherwise we can stop this service, but since this still has a 10-second delay, it will execute one call of `truncate` and then cancel it. This may not be the intended behavior.

controller/src/test/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessorsTest.java
```diff
@@ -50,18 +59,14 @@
...
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
 
 public class ControllerEventProcessorsTest {
     ScheduledExecutorService executor;
 
     @Before
     public void setUp() {
-        executor = Executors.newSingleThreadScheduledExecutor();
+        executor = Executors.newScheduledThreadPool(10);
```
andreipaduroiu, line 69: Why don't you inherit from `ThreadPoolTestBase` which will manage the executor for you?

controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java
shiveshr: Added a Precondition on truncate method. 

Although its not really going to cause any incorrect behaviour even if an extra truncation ran (and would run until the executor is shutdown or curator is shutdown), as it is equivalent to us being in the middle of truncation when the service shutdown.

Issue #5035, opened 2020-08-06
PDP 42: New Resource String Format for Authorization


ravisharda:
**Problem description**
Discuss [PDP 42: New Resource String Format for Authorization](https://github.com/pravega/pravega/wiki/PDP-42:-New-Resource-String-Format-for-Authorization)

**Problem location**
See the [PDP](https://github.com/pravega/pravega/wiki/PDP-42:-New-Resource-String-Format-for-Authorization).

**Suggestions for an improvement**

fpj:
It looks really good, @ravisharda , thanks for putting this together. A few questions and comments:

1- Does it need to be hierarchical? For example, could I use:

```
prn::/scope:myscope/stream:mystream
```

and

```
prn::/stream:mystream/scope:myscope
```

I understand that for Pravega it matters, but as far as resource declaration goes, it conveys the same information.

2- Can I use the wildcard `*` anywhere, like can I do this `*::*` or even use only `*` to represent everything in the system?

3- I'm not sure about the use of the backslash, perhaps we could stick only to colons like:

```
prn::stream:mystream::scope:myscope
```

but I don't know, maybe that's less clear and I don't have a strong opinion about that.

ravisharda:
> 1- Does it need to be hierarchical? For example, could I use:
> 
> ```
> prn::/scope:myscope/stream:mystream
> ```
> 
> and
> 
> ```
> prn::/stream:mystream/scope:myscope
> ```
> 
> I understand that for Pravega it matters, but as far as resource declaration goes, it conveys the same information.

From an auth handler perspective, the hierarchical representation makes it easier to process the string during authorization (using regex, for instance). Also, it makes it easier to support positional wildcards like these:
* `prn::/*` - All resources under the root resource)
* `prn::/scope:abc/*` - All resources that are children of scope, such as streams, reader groups, key-value tables, etc. 

While Pravega shall be using a fixed format, ACLs may be specified in any format that the Auth Handler implementation supports.  So, if a new auth handler plugin supports non-hierarchical format and can authorize a resource `prn::/scope:myscope/stream:mystream` using an ACL defined using its own custom format, it is up to it.

ravisharda:
> 2- Can I use the wildcard `*` anywhere, like can I do this `*::*` or even use only `*` to represent everything in the system?
> 

@fpj Yes, it could be anywhere in the ACL definitions of the built-in Password Auth Handler). A few examples: 

* `prn::*`  - All resources
[...]

PR #5051, opened 2020-08-10
Issue 5035: New Resource String Format for Authorization

ravisharda:
**Change log description**  
Replace the old resource string format with the new resource string format defined by [PDP-42](https://github.com/pravega/pravega/wiki/PDP-42:-New-Resource-String-Format-for-Authorization). 

**Purpose of the change**  
Implements #5035 ([PDP-42](https://github.com/pravega/pravega/wiki/PDP-42:-New-Resource-String-Format-for-Authorization)).

**What the code does**  
The main changes made in this PR are:
* Use the new resource string format for authorization.
* Modify the Password Auth Handler to interpret resource strings and patterns specified according to the new format. 

Here are the ancillary changes:
* Add unit tests for some of the existing code. Use the same tests to verify workings after the changes were made. 
* Add unit tests for newly added code.
* Refactor (rename, reorganize) some of the existing auth related code to make it easier to understand and make the required changes changes. 

**How to verify it**  
All unit, integration and system tests must pass with and without auth.


controller/src/test/java/io/pravega/controller/rest/v1/SecureStreamMetaDataTests.java
```diff
@@ -11,40 +11,50 @@
...
-
+        File file = File.createTempFile("SecureStreamMetaDataTests", ".txt");
         StrongPasswordProcessor passwordEncryptor = StrongPasswordProcessor.builder().build();
 
         try (FileWriter writer = new FileWriter(file.getAbsolutePath())) {
             String passwd = passwordEncryptor.encryptPassword("1111_aaaa");
 
             // Admin has READ_WRITE permission to everything
-            writer.write("admin:" + passwd + ":*,READ_UPDATE\n");
+            addAuthFileEntry(writer, "admin", passwd, Arrays.asList("prn::*,READ_UPDATE"));
```
sarlaccpit, line 41: Just curious, is `prn::*` equivalent to `prn::/*`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -75,6 +75,14 @@ public JsonWebToken verifyToken(@NonNull String resource, String token, @NonNull
      * @return
      */
     private boolean resourceMatchesClaimKey(String claimKey, String resource) {
+        log.trace("claimKey = {}, resourceKey = {}", claimKey, resource);
+        if (claimKey.startsWith("prn::")) {
+            claimKey = claimKey.replace("prn::/", "")
+                    .replace("prn::", "")
```
sarlaccpit, line 81: This is just to confirm my understanding:   the fact that the delegation token deals with resource strings that are in the same format as what goes on in the controller side authorization is just a coincidence, correct?

Meaning, in theory what goes on in the delegation token _could_ have been completely different (though why re-invent the wheel basically).  is that correct?

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
```diff
@@ -75,6 +76,7 @@
     private final RESTAuthHelper restAuthHelper;
     private final LocalController localController;
     private final ConnectionFactory connectionFactory;
+    private final AuthorizationResource authorizationResource = new AuthorizationResourceImpl();
```
sarlaccpit, line 79: Just curious where the toggle is to use the LegacyAuthorizationImpl() ?

Might be useful to implementors of the auth handler interface could take the new Pravega code and have a way to stick to the old format while they're updating their implementation to handle the new format ( else they'll break right away once Pravega starts asking about "prn://scope:scope1" and they really only understand  "/scope1" right ?).

I saw there IS a purposely implemented LegacyAuthorizationResourceImpl class but other than in tests, I didn't see it getting used (there are a lot of files in this PR though so I might have missed a subtlety somewhere :))

controller/src/test/java/io/pravega/controller/rest/v1/SecureStreamMetaDataTests.java
ravisharda: No, they are different. 
* `prn::*` gives the user access to all resources.
* `prn::/*` gives the user access to all scopes and resources underneath it. It doesn't provide the user access to the root resource `/`, the user cannot create a scope for instance.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: To provide a bit more context: For Segment store operations like reading from stream and appending to a stream, the user must have corresponding read/write access. The delegation token generated by the Controller has a "claim" of the form `<resource string>=<READ | READ_WRITE>`  that tells the segment store that the bearer of the token is allowed to perform the specified action. 

To answer your question, yes the delegation token could use a completed different format for the authorization key, but why re-invent the wheel.

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
ravisharda: No, you are right - there is no toggle built for the `LegacyAuthorizationImpl`. I've kept the code for ease of comparison between the new and old formats, as we transition from the old to the new format. 

With this code, the resource strings passed to the Auth Handler implementation will all be in the new format. Also, the built-in Password Auth Handler has been modified to take in ACLs containing patterns that match the new format. IMO, supporting both old and new formats would make the code a bit too complex (I had attempted it already) and cause problems in upgrade scenarios, without adding enough value to justify it. 

What do you say?

controller/src/test/java/io/pravega/controller/rest/v1/SecureStreamMetaDataTests.java
sarlaccpit: Got it.  So in order to create a scope, I'd need to either have:  READ_UPDATE on `prn::*` or READ_UPDATE on `prn::/`, right?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
sarlaccpit: Got it. Thanks for confirming.

controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java
sarlaccpit: I understand.  Cost vs reward.  It just means that implementors of the AuthHandler interface can't upgrade to Pravega 0.9 until they've updated their plugin code.  That said the changes imposed by the new format are not a big deal.  Also the auth handler implementation doesn't have to change the way IT stores its version of acls.  It simply has to adapt to the new way Pravega will "ask about" the resources being authorized and adjust its mapping. ie. "/scope1" is now "prn::/scope:scope1"

There are probably not many of those implementations out there at this time so that's probably ok.    

I can see though that in the password file impl, you've revamped also the format of the acls themselves since they follow closely the resource format that Pravega itself uses.

One thing I saw is that you have a LegacyAclAuthorizer; that is so somehow one can still use their old password file, correct?    The new PasswodFileHandler will still deal with new resources name being queried on, but it'll be able to map those to old and new ACLs.    Is there a setting/path in the code on the controller side that specifically allows setting that "oldFormat" flag? https://github.com/pravega/pravega/blob/a000a66776c32c3e475cfd7671eca45a1fbb5b69/controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/PasswordAuthHandler.java#L52

The documentation for the format of the acls in the password file will need to be updated (I'm thinking in particular the pravega cli in the tools repo that produces that password file)

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -101,9 +102,19 @@
     private final boolean replyWithStackTraceOnError;
 
     private final Supplier<Long> requestIdGenerator = RandomFactory.create()::nextLong;
-
+    /**
```
andreipaduroiu, line 105: Why do you have this copyright in here?

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.NonNull;
+
+class AclAuthorizerImpl extends AclAuthorizer {
+
+    @Override
+    public AuthHandler.Permissions authorize(@NonNull AccessControlList accessControlList, @NonNull String resource) {
+        AuthHandler.Permissions result = AuthHandler.Permissions.NONE;
+
+        String resourceDomain;
+        if (resource.indexOf("::") > 0) {
```
andreipaduroiu, line 22: You're doing a double-lookup here. Get the index of `"::"` and check for negative values, then reuse it in your `substring` call.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
```diff
@@ -0,0 +1,41 @@
...
+            resourceDomain = resource.substring(0, resource.indexOf("::"));
+        } else {
+            resourceDomain = "prn"; // default
+        }
+
+        for (AccessControlEntry accessControlEntry : accessControlList.getEntries()) {
+            // Replaces any `*` with `.*`, if it's not already preceded by `.`, for regex processing.
+            // So, `pravega:://*` becomes `pravega:://.*` and `pravega:://scope:*` becomes `pravega:://scope:.*`
+            String aclResourcePattern = accessControlEntry.getResourcePattern().replaceAll("(?!=.)\\*", ".*");
```
andreipaduroiu, line 31: The code inside the `for` loop does a lot of regex pattern compilation and matching, which is relatively expensive to process. Anyway we can simplify this? Can we precompute something (here or in `AccessControlEntry`)?

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
```diff
@@ -23,4 +28,26 @@ public static String credentialsAndAclAsString(String username, String password,
         // This will return a string that looks like this:"<username>:<pasword>:acl\n"
         return String.format("%s:%s:%s%n", username, password, acl);
     }
+
+    public static String createAclString(@NonNull List<String> aceStrings) {
+        StringBuilder builder = new StringBuilder();
```
andreipaduroiu, line 33: `String.join`?

or `aceStrings.stream().collect(Collectors.joining(...))`?

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
```diff
@@ -23,4 +28,26 @@ public static String credentialsAndAclAsString(String username, String password,
...
+        for (String aceString: aceStrings) {
+            builder.append(aceString).append(";");
+        }
+        return builder.toString();
+    }
+
+    @SneakyThrows
+    public static void addAuthFileEntry(@NonNull FileWriter writer, @NonNull String entryAsString) {
+        StringBuilder builder = new StringBuilder(entryAsString);
```
andreipaduroiu, line 42: For this simple case it's probably easier to do a direct string concatenation. StringBuilder won't help much.

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
```diff
@@ -23,4 +28,26 @@ public static String credentialsAndAclAsString(String username, String password,
...
+        writer.write(builder.append("\n").toString());
+    }
+
+    @SneakyThrows
+    public static void addAuthFileEntry(@NonNull FileWriter writer, @NonNull String username,
+                                           @NonNull String hashedPwd, @NonNull List<String> aceStrings) {
+        StringBuilder builder = new StringBuilder();
```
andreipaduroiu, line 49: Recommend using `String.format`. Easier to read.

controller/src/test/java/io/pravega/controller/rest/v1/SecureStreamMetaDataTests.java
andreipaduroiu, line 41: `Arrays.asList` -> `Collections.singletonList`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
```diff
@@ -75,6 +75,14 @@ public JsonWebToken verifyToken(@NonNull String resource, String token, @NonNull
      * @return
      */
     private boolean resourceMatchesClaimKey(String claimKey, String resource) {
+        log.trace("claimKey = {}, resourceKey = {}", claimKey, resource);
+        if (claimKey.startsWith("prn::")) {
+            claimKey = claimKey.replace("prn::/", "")
```
andreipaduroiu, line 80: Since you know your replaced string is at the beginning, use `substring`. Much faster to process.

controller/src/test/java/io/pravega/controller/rest/v1/SecureStreamMetaDataTests.java
ravisharda: Yes, that is right.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
ravisharda: That happened by mistake. I've removed it now.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
ravisharda: Yeah. I've now modified the code to ensure that is done once.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
ravisharda: Sure, that's a very good idea. I've pushed this logic up to when the built-in Auth handler's accounts DB file is loaded during ACE construction time, which happens just once.

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
ravisharda: Done.

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
ravisharda: Done.

controller/src/test/java/io/pravega/controller/auth/AuthFileUtils.java
ravisharda: Sure, I've modified it and it looks better.

controller/src/test/java/io/pravega/controller/rest/v1/SecureStreamMetaDataTests.java
ravisharda: Modified.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/delegationtoken/TokenVerifierImpl.java
ravisharda: IMO, the code looks simpler with the `replace` call and reveals its intention better. 

Besides, resources can be both of the form `prn::/...` and `prn::*`, which is why I'm replacing the former first and then attempting to replace `prn::`.

ravisharda: I have modified the code with a much more efficient logic. Please check out the new logic.

controller/src/main/java/io/pravega/controller/server/security/auth/AuthorizationResource.java
```diff
@@ -46,10 +42,7 @@ public static String ofScopes() {
      * @throws NullPointerException if {@code scopeName} is null
      * @throws IllegalArgumentException if {@code scopeName} is empty
      */
-    public static String ofScope(String scopeName) {
-        Exceptions.checkNotNullOrEmpty(scopeName, "scopeName");
-        return scopeName;
-    }
+    String ofScope(String scopeName);
```
shiveshr, line 45: is there any specific advantage in making these methods non static and accessing it on object?

controller/src/main/java/io/pravega/controller/server/security/auth/LegacyAuthorizationResourceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+        return scopeName;
+    }
+
+    public String ofStreamsInScope(String scopeName) {
+        return Exceptions.checkNotNullOrEmpty(scopeName, "scopeName");
+    }
+
+    public String ofStreamInScope(String scopeName, String streamName) {
+        Exceptions.checkNotNullOrEmpty(streamName, "streamName");
+        return String.format("%s/%s", ofStreamsInScope(scopeName), streamName);
```
shiveshr, line 34: same here and elsewhere

controller/src/main/java/io/pravega/controller/server/security/auth/StrongPasswordProcessor.java
```diff
@@ -7,7 +7,7 @@
  *
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
-package io.pravega.controller.server.rpc.auth;
+package io.pravega.controller.server.security.auth;
```
shiveshr, line 10: Should this class be in controller or be moved to "auth" module. 
This has nothing specific to controller's auth.

controller/src/main/java/io/pravega/controller/server/security/auth/UserPrincipal.java
```diff
@@ -7,7 +7,7 @@
  *
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
-package io.pravega.controller.server.rpc.auth;
+package io.pravega.controller.server.security.auth;
```
shiveshr, line 10: same comment here - should we take the opportunity to move this to auth module as this is generic User Principal which is serializable.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AccessControlEntry.java
```diff
@@ -0,0 +1,81 @@
...
+    @VisibleForTesting
+    AccessControlEntry(@NonNull String aceResource, @NonNull AuthHandler.Permissions permissions) {
+        this(aceResource, permissions, false);
+    }
+
+    @VisibleForTesting
+    AccessControlEntry(@NonNull String aceResource, @NonNull AuthHandler.Permissions permissions, boolean isLegacyFormat) {
+        // Replaces any `*` with `.*`, if it's not already preceded by `.`, for regex processing.
+        // So, `pravega:://*` becomes `pravega:://.*` and `pravega:://scope:*` becomes `pravega:://scope:.*`
+        this.resourcePattern = isLegacyFormat ?  aceResource :
```
shiveshr, line 42: so legacy pattern here does not have the "domain". and we do not have .* there. right?
so in the new pattern, why do we allow for `pravega:://scope:*` at all if this is the new pattern and we want ".*"

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AccessControlEntry.java
```diff
@@ -0,0 +1,81 @@
...
+                PATTERN_STAR_NOT_PRECEDED_BY_DOT.matcher(aceResource).replaceAll(".*");
+        this.permissions = permissions;
+    }
+
+    static AccessControlEntry fromString(String ace) {
+        String[] splits = null;
+        if (Strings.isNullOrEmpty(ace) || (splits = ace.split(",")).length != 2) {
```
shiveshr, line 49: in which situations can "ace" split not be of length 2? and is that an illegal argument or is ace = null legit if the number of tokens in it are 3 or 1?

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizer.java
```diff
@@ -0,0 +1,48 @@
...
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.controller.server.security.auth.handler.impl;
+
+import io.pravega.auth.AuthHandler;
+
+/**
+ * Authorizes resources based on supplied ACLs.
+ */
+abstract class AclAuthorizer {
```
shiveshr, line 17: is it possible that some ACLs are defined in legacy format while others are in new format? or is it expected to be consistent for all ACLs?

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.NonNull;
+
+class AclAuthorizerImpl extends AclAuthorizer {
+
+    @Override
+    public AuthHandler.Permissions authorize(@NonNull AccessControlList accessControlList, @NonNull String resource) {
+        AuthHandler.Permissions result = AuthHandler.Permissions.NONE;
+
+        String resourceDomain;
+        int indexOfPartsSeparator = resource.indexOf("::");
```
shiveshr, line 22: what is the reason to use `:://` instead of `://` ?

shiveshr: and should you simply split it by ":://" ?

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        if (indexOfPartsSeparator > 0) {
+            resourceDomain = resource.substring(0, indexOfPartsSeparator);
+        } else {
+            resourceDomain = "prn"; // default
```
shiveshr, line 26: when we are using the new authorization scheme, then should we even allow for resource domain to be unspecified?

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        for (AccessControlEntry accessControlEntry : accessControlList.getEntries()) {
+            // You could have a null ACE in the ACL if you had a malformed entry such as `prn::/scope:readresource`
+            // having no permissions set.
+            if (accessControlEntry != null && accessControlEntry.resourceStartsWith(resourceDomain)) {
```
shiveshr, line 32: nit: two nested if conditions. can it be simplified?

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
```diff
@@ -0,0 +1,41 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.controller.server.security.auth.handler.impl;
+
+import io.pravega.auth.AuthHandler;
+import lombok.NonNull;
+
+class AclAuthorizerImpl extends AclAuthorizer {
```
shiveshr, line 15: Should we move this class to a common module like "auth" such that schema registry and other services can leverage the auth module

controller/src/main/java/io/pravega/controller/server/security/auth/LegacyAuthorizationResourceImpl.java
ravisharda: Specifying the format directly makes it easier to understand what's happening here, IMO.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AccessControlEntry.java
ravisharda: Since ACEs are specified manually via the Password Auth Handler input file, they could be ill-formed. 

A couple of examples of the split being of a length != 2 are as follows: 
* `prn::/` (No comma - permission missing, splits = 1)
* `,,READ`  (3 splits)

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
ravisharda: As discussed, I've created a separate issue #5118 for the reorganization of the security-related code in general.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizer.java
ravisharda: As per current implementation, all ACLs must be specified in the new format. Any ACLs specified in legacy format won't work.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AccessControlEntry.java
ravisharda: I added support for `*` in lieu of the regex `.*` for usability. Admins/operators may find it easier to remember to use wildcards without the preceding dot. Both'd work exactly the same way as we replace `*` with `.*`  if it is not preceded by the dot.

controller/src/main/java/io/pravega/controller/server/security/auth/AuthorizationResourceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public String ofStreamsInScope(String scopeName) {
+        return ofScope(scopeName);
+    }
+
+    @Override
+    public String ofStreamInScope(String scopeName, String streamName) {
+        Exceptions.checkNotNullOrEmpty(scopeName, "scopeName");
+        Exceptions.checkNotNullOrEmpty(streamName, "streamName");
+        return String.format("%s/%s:%s", ofScope(scopeName), TAG_STREAM, streamName);
```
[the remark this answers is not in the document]
ravisharda: IMO specifying the format directly here makes it easier to understand the output returned by this line of code.

controller/src/main/java/io/pravega/controller/server/security/auth/StrongPasswordProcessor.java
ravisharda: I've created a separate issue #5118 that should include these changes plus other organizational changes required in the security-related code.

controller/src/main/java/io/pravega/controller/server/security/auth/UserPrincipal.java
ravisharda: https://github.com/pravega/pravega/pull/5051/#discussion_r478259935

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
ravisharda: Yes, the code above assumes the default `prn` domain. When this code is used as-is in other domains like `prn.schema-registry`, the domain part will automatically be mandatory.

I have two opposite opinions about this. 
* On one hand, allowing for this flexibility could make life a tad easier for the admins, improving usability.
* On the other hand, it's slightly simpler if admins always had to specify it and they didn't have to make an explicit choice (to specify the domain or not). 

In this case, I've favored the former. What do you think?

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
ravisharda: Actually, the resource looks like this: `prn::/scope:MarketData/stream:Prices`. There was an inaccurate comment elsewhere that right have misled you about the format.

More examples can be found [here](https://github.com/pravega/pravega/wiki/PDP-42:-New-Resource-String-Format-for-Authorization#examples).

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
ravisharda: On second thought, we don't actually need to add a default resource domain. This is not the ACE, where domain might be missing. This is really the resource string which is in our control and should always have the domain. I will fix it.

ravisharda: I've now pushed the modified code that assumes domain is always present in the resource string used for authorization.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizerImpl.java
ravisharda: Hmm... can't think of a simpler means. Do you have an idea?

controller/src/main/java/io/pravega/controller/server/security/auth/LegacyAuthorizationResourceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.controller.server.security.auth;
+
+import io.pravega.common.Exceptions;
+
+/**
+ * A legacy implementation that constructs resource strings in the old format.
+ */
+public final class LegacyAuthorizationResourceImpl implements AuthorizationResource {
```
fpj, line 17: I suppose we are deprecating the previous format, how are we going to indicate that it is deprecated? Via documentation? We can certainly annotate this class, but that's not visible to users.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AccessControlEntry.java
fpj: A minor point, but shouldn't it be an exception rather returning null? We expect a valid string, and if the string is not valid, then I'd say it is an exception.

fpj (changes requested):
No major concern, only a couple of small comments.

controller/src/main/java/io/pravega/controller/server/security/auth/LegacyAuthorizationResourceImpl.java
ravisharda: Actually, we are removing support for the old format altogether. This class is there only so that it is easy to see what was the old format and compare it with the new format.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AccessControlEntry.java
ravisharda: @fpj Returning `null` here makes the calling code in the `PasswordAuthhandler` much simpler, which is why I took this approach here. This method returns a non-null `AccessControlEntry` object for valid strings and null for invalid strings.

ravisharda:
> A minor point, but shouldn't it be an exception rather returning null? We expect a valid string, and if the string is not valid, then I'd say it is an exception.

@fpj Returning null here makes the calling code in the PasswordAuthhandler much simpler, which is why I took this approach here. This method returns a non-null AccessControlEntry object for valid strings and null for invalid strings.

shiveshr (approved):
LGTM, few minor comments

andreipaduroiu (approved):

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizer.java
ravisharda: Yes, that is right.

controller/src/main/java/io/pravega/controller/server/security/auth/LegacyAuthorizationResourceImpl.java
ravisharda: I wanted to keep it at least for some time so that its easier to inspect the differences and compare the two at development time. I shall remove them once this change has been well tested through the normal course of testing.

ravisharda: In the latest comment that I have just pushed, I've added a deprecation note in the class JavaDoc. (I couldn't use the deprecation annotation, as the current build configuration fails build if classes decorated with that annotation are used.) 

Also, these classes are used internally for representing resources and authorization and aren't meant to be used by external parties.

controller/src/main/java/io/pravega/controller/server/security/auth/AuthorizationResource.java
```diff
@@ -7,12 +7,10 @@
...
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
-package io.pravega.controller.server;
-
-import io.pravega.common.Exceptions;
+package io.pravega.controller.server.security.auth;
 
 /**
- * A utility class with methods for preparing string representations of auth-protected resources.
+ * Returns string representations of auth-protected resources.
```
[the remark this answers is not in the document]
ravisharda: I've modified the comment as suggested.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/impl/AclAuthorizer.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final static AclAuthorizerImpl AUTHORIZER_FOR_NEW_FORMAT = new AclAuthorizerImpl();
+    private final static LegacyAclAuthorizerImpl AUTHORIZER_FOR_LEGACY_FORMAT = new LegacyAclAuthorizerImpl();
+
+    /**
+     * Returns a cached instance of the legacy implementation.
+     *
+     * @return an instance
+     */
+    static AclAuthorizer legacyAuthorizerInstance() {
```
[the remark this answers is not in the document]
ravisharda: I've kept it so that the tests can instantiate it.

controller/src/main/java/io/pravega/controller/server/security/auth/LegacyAuthorizationResourceImpl.java
fpj: I'm not sure this makes sense. If we need to revert or anything, the we can go back to a previous version, that's a reason for using version control. If this code is only being kept for reference, then let's please remove it. If it is being used anywhere, then keep it.

ravisharda: Ok. I have removed the two classes that show legacy logic in the latest commit that I just pushed.

fpj (approved):

PR #5053, opened 2020-08-10
Issue 5039: (SegmentStore) RollingStorage can trim Header Chunk on FileSystemStorage

### LINKED ISSUE #5039: RollingStorage: Trim Header file upon Segment Chunk deletion
**Problem description**
This applies to Rolling Storage (not SLTS).

The header file is only appended to and is never head-truncated when we delete Segment Chunks. This causes it to grow unboundedly.

**Problem location**
RollingStorage

Devise a mechanism for head-truncating it.

andreipaduroiu:
**Change log description**  
- Added generic API to "atomically replace" a Storage segment contents with provided data. 
- Making use of this API in `RollingStorage` to trim Header Chunk after a Segment Truncation that removes one or more chunks.
- This has only been implemented for `FileSystemStorage` and only if `filesystem.replace.enable` is set to `true`.

**Purpose of the change**  
Fixes #5039.

**What the code does**  
- `SyncStorage`: added some APIs to explicitly check for and enable atomic replace support.
    - The `withReplaceSupport` has been added to make sure that only code that actually requires this feature uses it and to prevent any associated overhead and/or side effects from spilling over to other types of segment chunks
- `RollingStorage`
    - Using base `SyncStorage` for all data chunk manipulation (same as before).
    - Using `SyncStorage.withStorageSupport` (if available) for Header Chunk manipulation.
    - Replacing Header Chunk with updated one after each Segment Truncation that deletes chunks.
- `FileSystemStorage`
    - Implemented a `FileSystemStorageWithReplaceSupport` derived class that enables atomic replace. Please refer to this class' documentation for methodology.
    - Since this operation is multi-step and non-atomic, we need to handle the fallout from any partially executed operation and recover that the next time we use the touched segment. 
    - It is due to this recovery process that the best place to encapsulate this feature is at the LTS Storage adapter level, and it is also due to this that we should restrict it to Header Chunks alone (this extra code may add slight runtime overhead for frequently used data chunks - Header LTS operations are a fraction of total LTS operations).

**THIS DOES NOT AFFECT `ChunkStorageProvider`. It is restricted to `RollingStorage` alone.**

**How to verify it**  
All tests must pass.
New tests added to verify functionality.


sachin-j-joshi:
I am not sure why we need to modify SyncStorage interface instead of adding additional logic in RollingStorage itself.  
I do not think we necessarily need to modify FileSystemStorage a lot. The code seems more complex than necessary.

andreipaduroiu:
| I am not sure why we need to modify SyncStorage interface instead of adding additional logic in RollingStorage itself.

There is no way to expose the rename operation outside of FileSystemStorage. This method is extensible to non-FileSystemStorage implementation as well should we decide that we need it for ExtendedS3Storage, for example.

| I do not think we necessarily need to modify FileSystemStorage a lot

FileSystemStorage hasn't been modified too much except for the 3 methods that create the `FileSystemWithReplaceSupport` class. That latter one has all the logic for replacing encapsulated within it, including auto-recovery support, which cannot be done outside of that class.


| The code seems more complex than necessary

Would you mind elaborating please?

sachin-j-joshi:
1. I was expecting only `rename` to be added to `SyncStorage`, instead we have 3 methods. Rename is simple enough premitive operation and useful in other contexts as well (eg DR tool). 
Header rename/replace fnctionality can then be implemented in RollingStorage without making any further changes to SyncStorage and it's implementations.

2. Segment operations can not be done without first opening the handle. So I was expecting that header manipulation changes would be limited to code that opens the handle and the truncate. (In RollingStorage). Other mnethods do not need to manipulate header.

3. However my biggest concern is that I am still not sure how we are dealing with zombie writes. For idempotent stores the logic was we used to write/overwrite same data at same offset, so the old zombie storage could never corrupt the data. However with truncated header it is quite possible that zombie just overwrites outdated data at now wrong offset, completely corrupting the header file.

andreipaduroiu:
> I was expecting only rename to be added to SyncStorage, instead we have 3 methods. Rename is simple enough premitive operation and useful in other contexts as well (eg DR tool).
Header rename/replace fnctionality can then be implemented in RollingStorage without making any further changes to SyncStorage and it's implementations.

Rename is not as simple as you may think. We do not need to rename a file A to B.  We need to rename A to B, where B already exists. As part of this, the rename operation needs to atomically replace B with A. While some file system implementations provide this guarantee, not all do. Windows does not (hence the exceptional case in the code). NFS does not guarantee that.

If you look at the three methods added, they are as follows:
1. The method doing the hard work itself.
2. Some boolean method indicating whether this feature is implemented. I did not want to sign up to do this for all existing LTS implementations. RollingStorage must know which LTS bindings supports this.
3. Whether using RollingStorage or directly in the binding, we do not want to enable this for every access to LTS. I decided to only enable this feature for those LTS segment chunks that actually need it. In this case, that is only the header file. No other chunks require it so I didn't want to touch that code.

> Segment operations can not be done without first opening the handle. So I was expecting that header manipulation changes would be limited to code that opens the handle and the truncate. (In RollingStorage). Other mnethods do not need to manipulate header.

You need this for create and delete too. You need this for concat, since concat does not take a handle for the source segment. The other ones are up for debate but if we detected the operation it's simpler to complete the partially executed operation instead of retrying using the temp file. You mention complexity - this approach reduces it.

> However my biggest concern is that I am still not sure how we are dealing with zombie writes. For idempotent stores the logic was we used to write/overwrite same data at same offset, so the old zombie storage could never corrupt the data. However with truncated header it is quite possible that zombie just overwrites outdated data at now wrong offset, completely corrupting the header file.

The zombie cannot just overwrite outdated data. First of all, the likelihood of 2 segment container instances fighting for the same header file at the same time is much smaller than fighting for a data file. Furthermore, that problem existed before - if 2 instances want to both write to the header file at the same offset, nothing will stop them.

But if you think about it, there is an additional reason why this won't be a problem. If by any chance one of the instances tries to append to a "replaced" file, it will:
1. Open a file channel (handle)
2. Check the length
3. perform the write.

If the replace was done by the other instance prior to step 1, then step 2 will throw an exception. If the replace was done between 1 and 2, then step 3 will fail with invalid handle - since the target file has been replaced by another.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -458,5 +466,225 @@ private Void doDelete(SegmentHandle handle) throws IOException {
...
+            boolean baseExists = super.doExists(segmentName);
+            boolean shouldReseal = baseExists && super.getStreamSegmentInfo(segmentName).isSealed();
+
+            // Check temp file already exists.
+            String tmpSegmentName = getTempSegmentName(segmentName);
+            if (super.doExists(tmpSegmentName)) {
+                if (baseExists) {
+                    // We have both a temp file and a segment file. This is most likely the result of an incomplete replace,
+                    // however we still have the original segment file around. It is safe to delete the temp file now.
+                    log.debug("Incomplete replace operation detected for '{}'. Deleting temp file before new replace attempt.", segmentName);
```
fpj, line 615: Since this case is not the regular, expected path, I'd say this log message is at least `info`.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -458,5 +466,225 @@ private Void doDelete(SegmentHandle handle) throws IOException {
...
+                    super.doDelete(super.doOpenWrite(tmpSegmentName));
+                } else {
+                    // Temp file exists, but the segment file does not. This may be the result of an incomplete replace,
+                    // in which case we need to finalize that one to prevent deleting (what could be) the only persisted
+                    // copy of our data.
+                    log.debug("Incomplete replace operation detected for '{}'. Finalizing before new replace attempt.", segmentName);
```
fpj, line 621: Same suggestion about the severity of the log message.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -458,5 +466,225 @@ private Void doDelete(SegmentHandle handle) throws IOException {
...
+            return null;
+        }
+
+        @SneakyThrows(StreamSegmentReplaceException.class)
+        private void finalizeRename(String fromSegmentName, String toSegmentName) {
+            val fromFile = getPath(fromSegmentName).toFile().getAbsoluteFile();
+            val toFile = getPath(toSegmentName).toFile().getAbsoluteFile();
+            boolean renamed = fromFile.renameTo(toFile);
+            if (!renamed) {
+                // On Windows, renameTo does not replace.
```
fpj, line 659: When we say `On Windows...`, is it referring to the Windows implementation of the JVM or a Windows file system (FAT, NTFS)? I ask because it is the LTS implementation that needs to implement rename, right?

Also, is the behavior reflected here documented anywhere we can point to or was this validated manually?

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
```diff
@@ -458,5 +466,225 @@ private Void doDelete(SegmentHandle handle) throws IOException {
...
+            // Rename file. After this is done, the replace operation is complete.
+            finalizeRename(tmpSegmentName, segmentName);
+
+            // Sanity check #2 (only executes in tests).
+            assert super.doGetStreamSegmentInfo(segmentName).getLength() == contents.getLength();
+            return null;
+        }
+
+        @SneakyThrows(StreamSegmentReplaceException.class)
+        private void finalizeRename(String fromSegmentName, String toSegmentName) {
```
fpj, line 654: Maybe add a debug or trace log to record the rename.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/SyncStorage.java
```diff
@@ -216,6 +216,40 @@ default SegmentHandle create(String streamSegmentName, SegmentRollingPolicy roll
...
+        throw new UnsupportedOperationException("replace() is not implemented");
+    }
+
+    /**
+     * Returns a new {@link SyncStorage} instance for the same Storage type as this one, but with {@link #replace} support
+     * enabled. If there is no such implementation, this instance is returned.
+     *
+     * @return Either this instance or a new {@link SyncStorage} instance based on this one that can perform replaces.
+     */
+    default SyncStorage withReplaceSupport() {
```
fpj, line 249: This is the only call that feels odd when comparing to the other calls in the interface. To be able to get rid of it, we would need to make `FileSystemStorage` support it directly rather than have a derived class `FileSystemStorageWithReplace`, which seems possible and it is not clear to me whether it is incorrect or simply undesirable.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
andreipaduroiu: The Javadoc for `File.renameTo` states that this operation may not be atomic or may not succeed if a file with the destination name already exists, so that is enough to warrant the `if` block and the subsequent 2-step rename.
 
With respect to Windows, I am unaware of the inner details on what the requirements are. For this code, I got inspired from the AtomicFileOutputStream (both in the [HDFS repository](https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/util/AtomicFileOutputStream.java) and [ZooKeeper repository](https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/common/AtomicFileOutputStream.java) - they look nearly identical save for a few diffs).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/SyncStorage.java
```diff
@@ -216,6 +216,40 @@ default SegmentHandle create(String streamSegmentName, SegmentRollingPolicy roll
...
+     * respect to behavior, atomicity and recovery mechanisms.
+     *
+     * @param segment A {@link SegmentHandle} representing the Segment to replace.
+     * @param contents A {@link BufferView} representing the new contents of the Segment.
+     * @throws StreamSegmentNotExistsException When the given Segment does not exist in Storage.
+     * @throws StorageNotPrimaryException      When this Storage instance is no longer primary for this Segment (it was
+     *                                         fenced out).
+     * @throws UnsupportedOperationException   If {@link #supportsReplace()} returns false.
+     */
+    default void replace(SegmentHandle segment, BufferView contents) throws StreamSegmentException {
```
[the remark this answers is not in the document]
andreipaduroiu: But I am not renaming a file here. What I need is to set the contents of a Segment to a given buffer. I have that buffer in `contents` and this method is expected to set that as the actual contents of `segment`. Whether or not a `rename` operation happens behind the scenes to implement this is a different story.

For example, in `ExtendedS3Storage` this may be accomplished by a `putObjectRequest` that overwrites a target object - no need for rename.

For this fix, we do not need or care about renaming segments. We care about "atomically" replacing the contents of a segment with some other contents that we have already generated (in memory).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/SyncStorage.java
andreipaduroiu: `FileSystemWithReplace` does a bit of extra work for every file access and catches certain exceptions. To prevent that overhead from happening **on every single LTS call** which could negatively affect performance (and open the door for more hacks in the future (since it's easy to use such functionality once there), I decided to explicitly offer this as a separate piece of functionality. The only place where this functionality is required is in `RollingStorage` for the header file. Furthermore, `SyncStorage` is the internal LTS contract, not the external one (that one is `Storage`) so nobody else outside of `RollingStorage` would even have access to this. 

This is not a long-term solution. We are doing it as a short-term fix until SLTS will replace `RollingStorage` (SLTS does not have the problem that this PR is fixing). 

In other words, I'm deliberately trying to make it hard for people to accidentally use this feature and not slip through future code reviews unnoticed. We do not plan to support this functionality long term.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
andreipaduroiu: Done.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
andreipaduroiu: Added some clarifying comment on this.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorage.java
andreipaduroiu: Done.

bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorageConfig.java
```diff
@@ -24,6 +24,7 @@
     //region Config Names
 
     public static final Property<String> ROOT = Property.named("root", "/fs/");
+    public static final Property<Boolean> REPLACE_ENABLED = Property.named("replace.enabled", false);
```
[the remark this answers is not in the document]
andreipaduroiu: > is the full property name filesystem.replace.enable

Yes.

> are we omitting this configuration in config.properties intentionally?

Yes. This is to support a stop-gap fix for RollingStorage which will not be needed once SLTS will replace it. I do not want to advertise it to potential users and stir confusion when it will not be there anymore.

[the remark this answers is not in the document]
andreipaduroiu: > it should be enable rather than enabled.

Fixed.

PR #5079, opened 2020-08-17
Issue 5080: Adding character length checks for scope and stream name.

### LINKED ISSUE #5080: Creation of stream is failing with 500 error when the scope name has more than 256 characters in it.
**Problem description**
Creation of stream is failing with 500 error when the scope name has more than 256 characters in it.
I have tested scope names up to 501 characters and scope creation is working fine.
This is not a regression test scenario and we have executed this first time in Pravega build 0.8.0.
Creation of stream worked fine when we had 256 characters in the scope name.
Creation of scope is working fine with more than 256 characters also, but creation of stream is failing with 500 error.

**Problem location**
NameUtils.java

We should restrict the creation of scope with more than 256 characters if we have an upper limit when we create stream in that scope.

ldr7:
Signed-off-by: Atharva Joshi <Atharva.Joshi@dell.com>

**Change log description**  
Character length limit check is added for scope and stream name.

**Purpose of the change**  
Fixes #5080 

**What the code does**  
Adds a precondition check to the validateStreamName & validateUserStreamName methods in NameUtils.java class to ensure that stream and scope name do not exceed 255 characters.

**How to verify it**  
Attempt to add a scope or stream with a name having more than 255 characters.

### DIFF
--- a/shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
+++ b/shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
@@ -624,6 +624,7 @@ public final class NameUtils {
      */
     public static String validateUserStreamName(String name) {
         Preconditions.checkNotNull(name);
+        Preconditions.checkArgument(name.length()<256, "Name must have less than 256 characters");
         Preconditions.checkArgument(name.matches("[\\p{Alnum}\\.\\-]+"), "Name must be a-z, 0-9, ., -.");
         return name;
     }
@@ -648,6 +649,7 @@ public final class NameUtils {
 
         // In addition to user stream names, pravega internally created stream have a special prefix.
         final String matcher = "[" + INTERNAL_NAME_PREFIX + "]?[\\p{Alnum}\\.\\-]+";
+        Preconditions.checkArgument(name.length()<256, "Name must have less than 256 characters");
         Preconditions.checkArgument(name.matches(matcher), "Name must be " + matcher);
         return name;
     }
--- a/shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
+++ b/shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
@@ -624,7 +624,7 @@ public final class NameUtils {
      */
     public static String validateUserStreamName(String name) {
         Preconditions.checkNotNull(name);
-        Preconditions.checkArgument(name.length()<256, "Name must have less than 256 characters");
+        Preconditions.checkArgument(name.length() < 256, "Name must have less than 256 characters");
         Preconditions.checkArgument(name.matches("[\\p{Alnum}\\.\\-]+"), "Name must be a-z, 0-9, ., -.");
         return name;
     }
@@ -649,7 +649,7 @@ public final class NameUtils {
 
         // In addition to user stream names, pravega internally created stream have a special prefix.
         final String matcher = "[" + INTERNAL_NAME_PREFIX + "]?[\\p{Alnum}\\.\\-]+";
-        Preconditions.checkArgument(name.length()<256, "Name must have less than 256 characters");
+        Preconditions.checkArgument(name.length() < 256, "Name must have less than 256 characters");
         Preconditions.checkArgument(name.matches(matcher), "Name must be " + matcher);
         return name;
     }
--- a/shared/protocol/src/test/java/io/pravega/shared/NameUtilsTest.java
+++ b/shared/protocol/src/test/java/io/pravega/shared/NameUtilsTest.java
@@ -1,21 +1,19 @@
 /**
  * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
- * <p>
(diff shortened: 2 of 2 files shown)


RaulGracia:
@ldr7 please, create a public Git issue to link this PR.

shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
```diff
@@ -624,6 +624,7 @@ public static String getStreamForReaderGroup(String groupNameName) {
      */
     public static String validateUserStreamName(String name) {
         Preconditions.checkNotNull(name);
+        Preconditions.checkArgument(name.length()<256, "Name must have less than 256 characters");
```
derekm, line 627: How does this work with multi-byte unicode character strings?

`String#length()` counts code units, not code points.

We may want to use `String#codePointCount(int, int)`. This will allow languages in higher unicode planes to also use 255 characters (which can consume up to 4 bytes per character in UTF8).

I believe the limiting factor in a scope name or stream name is the 1MB ZooKeeper ZNode size limit.

Will 255 character ever be too limiting? 255 characters is like 7 UUIDs.

derekm: Some languages use multiple code points for a single character or grapheme. See: http://illegalargumentexception.blogspot.com/2009/05/java-rough-guide-to-character-encoding.html#javaencoding_strnlen

andreipaduroiu: We don't store Stream names in ZK. We stash them inside Table Segment Key-Value Pairs, either as Keys (max 8KB) or values (max 1MB - but may contain more than just a stream name). So the main constraint here will be Key - max 8KB.

Stream names are also part of the Segment names (`<Scope-Name>/<Stream-Name>/<Segment-Name>`. Given that we use Segment names as FileSystem paths (for NFS) or Object Keys (for S3/ExtS3), and so on, the constraint might be even higher than 8KB.

If you look at the line immediately following this line you'll see a Regex that enforces the name to be alphanumeric. As such, the character count is good enough because ASCII should be enough to encode this. I don't think we should worry about higher-level unicode chars in this case).

ldr7: In case of HDFS, the maximum component length for a file path is 255 bytes in UTF-8 encoding. 
[https://issues.apache.org/jira/browse/HDFS-6055](url)

Since these names are to only contain alphanumeric characters, I do not think using `String#codePointCount(int, int)` would offer significant benefits.

andreipaduroiu: If the maximum length for a path is 256, then your change isn't going to fix anything. See my previous comment why.

andreipaduroiu:
@ldr7 Please fix the title of this PR in accordance with our contribution guidelines.

shared/protocol/src/test/java/io/pravega/shared/NameUtilsTest.java
```diff
@@ -73,6 +74,23 @@ public void testStreamNameVerifier() {
         NameUtils.validateStreamName("1.2.3");
     }
 
+    @Test
+    public void testStreamNameLimit() {
+        int leftLimit = 48; // numeral '0'
+        int rightLimit = 122; // letter 'z'
+        int targetStringLength = 256;
+        Random random = new Random();
```
andreipaduroiu, line 82: seed with 0

shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
derekm: @andreipaduroiu
> If the maximum length for a path is 256, then your change isn't going to fix anything. See my previous comment why.

I think this means each path part, not the entire path. He is sizing the max length to the limits of HDFS path parts.

@ldr7
> Since these names are to only contain alphanumeric characters, would using String#codePointCount(int, int) offer any benefits?

Yeah, my suggestion is invalid given the \p{Alnum} character class constraint. Apologies for the distraction! I can be a stickler about i18n, but I didn't notice it doesn't really apply in this context.

Now the only thing bugging me is we can't throw all constraint violations at once! ;)

shared/protocol/src/test/java/io/pravega/shared/NameUtilsTest.java
derekm: Aside: It might be cool to bring in the *jqwik* test engine so we can get property generators such as string length and character classes: https://jqwik.net/docs/current/user-guide.html#string-length

Cf., https://github.com/junit-team/junit5-samples/tree/main/junit5-multiple-engines

ldr7: Since the test is written around a hard coded limit, I think bringing in jqwik for this might be a bit of an overkill.

PR #5084, opened 2020-08-19
Issue 5072: Single routing key batch writes

### LINKED ISSUE #5072: PDP 44 - Lightweight Transactions
**Problem description**
Lightweight transactions proposal as described in https://github.com/pravega/pravega/wiki/PDP-44:-Single-Routing-Key-Transactions.

**Problem location**
Transactions

Implement light weight transaction

shiveshr:

**Change log description**  
Implementation for writing batch events using append block protocol.

**Purpose of the change**  
Fixes #5072 

**What the code does**  
Updates the `PendingEvent` object to **also** be a composite object of a batch of events apart from supporting single event. 
SegmentOutputStreamImpl uses this pending event which can have more than one event and advances the event sequence number based on the number of events.

**How to verify it**  
Unit tests added


client/src/main/java/io/pravega/client/stream/impl/SingleRoutingKeyTransactionWriterImpl.java
```diff
@@ -0,0 +1,204 @@
...
+
+    @Synchronized
+    private void segmentSealedCallback(Segment segment) {
+        // instead of traversing thru successors we will call current segments and fetch the appropriate segment.
+        // by setting segmentOutputStream to null we will force the caller to get the replacement segmentoutputstream.
+        // we will call resend while holding the lock so that successor segment outputstream is setup and events are
+        // replayed to it before new writeevents into successor are triggered. 
+        List<PendingEvent> toResend = segmentOutputStream.getUnackedEventsOnSeal();
+        segmentOutputStream = null;
+        for (PendingEvent event : toResend) {
```
shrids, line 180: I do not see how there can be multiple PendingEvents here. Am I missing something here?

shiveshr: so conceptually there could be multiple "transactions" created on the same txn writer. so there could be more than one pending event (one per transaction)

client/src/main/java/io/pravega/client/stream/impl/PendingEvent.java
```diff
@@ -38,26 +43,48 @@
...
+        return eventBuf;
     }
-    
+
+    public static PendingEvent withHeader(@NonNull String routingKey, @NonNull List<ByteBuffer> batch, @NonNull CompletableFuture<Void> ackFuture) {
+        Preconditions.checkArgument(!batch.isEmpty(), "Batch cannot be empty");
+        ByteBuf batchBuff = Unpooled.EMPTY_BUFFER;
+        for (int i = 0; i < batch.size(); i++) {
+            ByteBuf eventBuf = getByteBuf(batch.get(i));
+            batchBuff = Unpooled.wrappedUnmodifiableBuffer(batchBuff, eventBuf);
```
tkaitchuck, line 77: Rather than making an N level deep structure, it would be better to add all the buffers into a list and then create a single wrapping buffer.

client/src/main/java/io/pravega/client/EventStreamClientFactory.java
```diff
@@ -85,6 +86,18 @@ static EventStreamClientFactory withScope(String scope, ClientConfig config) {
...
+     * Creates a new single routing key transaction writer that can write a batch of events with atomic guarantees. 
+     * There are limits on the size of each event in the batch and the overall size of the batch.
+     *
+     * @param streamName The name of the stream to write to.
+     * @param routingKey Routing key to be used while writing the events.
+     * @param s The Serializer.
+     * @param <T> The type of events.
+     * @return Newly created writer object
+     */
+    <T> SingleRoutingKeyTransactionWriter<T> createSingleRoutingKeyWriter(String streamName, String routingKey, Serializer<T> s);
```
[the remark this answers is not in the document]
shiveshr: This is a good suggestion. Though based on the feedback and suggestion from @fpj and others, i am removing the transaction decoration at the moment. 
We are going to attempt to improve the performance of distributed transaction and see if we can incorporate "single" routing key transaction by using the transient segment and performing segment-merges.

shrids (approved):
+1

tkaitchuck (approved):

PR #5092, opened 2020-08-21
Issue 5058: Cache Manager Metrics Fix

### LINKED ISSUE #5058: Failing Test: SegmentStoreMetricTests.testCacheManagerMetrics
**Problem description**

The newly added metric validates itself using the wrong metric type (`Timer` over `Gauge`).

**Problem location**

`SegmentStoreMetricsTest.testCacheManagerMetrics`

Use proper metric type to verify correctness.

co-jo:
**Change log description**  

* Make  the metric an instance field instead of a static field.
* Clean up meters alongside registries during `startWithoutExporting`

**Purpose of the change**  
* Fixes #5058 

**What the code does**  
It addresses two problems without our metric implementation:

* It does not seem `static` metrics can work reliably, as these metrics will often resolve to a `NullStatsLogger` as `MetricsProvider.initialize()` may not have been called before the class has been loaded. I believe that the metric defined above will also be susceptible to this issue, but is not tested against.
* As noted in the `io.micrometer.Meter` comments, removing a registry from a `CompositeRegistry` (`Metrics.globalRegistry`) does **not** remove it from the `meterMap` of said `CompositeRegistry`. The problem is that the internal `meterMap` of the `CompositeRegistry` is what the search (`Metrics.globalRegistry.find`) is performed on (and will return a 0 value (in the case of a counter). This may be expected, but seemingly can be problematic as we cannot differentiate between a 'DNE' 0 and a reported 0 value.

**How to verify it**  

### DIFF
--- a/segmentstore/server/src/test/java/io/pravega/segmentstore/server/SegmentStoreMetricsTests.java
+++ b/segmentstore/server/src/test/java/io/pravega/segmentstore/server/SegmentStoreMetricsTests.java
@@ -161,22 +161,22 @@ public class SegmentStoreMetricsTests {
 
     @Test
     public void testCacheManagerMetrics() {
-        int storedBytes = 10000;
-        int usedBytes = 1000;
-        int allocatedBytes = 100;
+        long storedBytes = 10000;
+        long usedBytes = 1000;
+        long allocatedBytes = 100;
         int generationSpread = 10;
-        int managerIterationDuration = 1;
+        long managerIterationDuration = 1;
 
         int containerId = new Random().nextInt(Integer.MAX_VALUE);
         @Cleanup
         SegmentStoreMetrics.CacheManager cache = new SegmentStoreMetrics.CacheManager();
         cache.report(new CacheState(storedBytes, usedBytes, 0, allocatedBytes, storedBytes), generationSpread, managerIterationDuration);
 
-        assertEquals(storedBytes, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_STORED_SIZE_BYTES).value());
-        assertEquals(usedBytes, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_USED_SIZE_BYTES).value());
-        assertEquals(allocatedBytes, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_ALLOC_SIZE_BYTES).value());
+        assertEquals(storedBytes, (long) MetricRegistryUtils.getGauge(MetricsNames.CACHE_STORED_SIZE_BYTES).value());
+        assertEquals(usedBytes, (long) MetricRegistryUtils.getGauge(MetricsNames.CACHE_USED_SIZE_BYTES).value());
+        assertEquals(allocatedBytes, (long) MetricRegistryUtils.getGauge(MetricsNames.CACHE_ALLOC_SIZE_BYTES).value());
         assertEquals(generationSpread, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_GENERATION_SPREAD).value());
-        assertEquals(managerIterationDuration, (int) MetricRegistryUtils.getTimer(MetricsNames.CACHE_MANAGER_ITERATION_DURATION).mean(TimeUnit.MILLISECONDS));
+        assertEquals(managerIterationDuration, (long) MetricRegistryUtils.getTimer(MetricsNames.CACHE_MANAGER_ITERATION_DURATION).mean(TimeUnit.MILLISECONDS));
 
         cache.close();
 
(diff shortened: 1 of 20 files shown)


andreipaduroiu:
The build for this failed twice on the same unit test, which has something to do with metrics. This is highly likely due to the change made to the metrics framework in this PR.

co-jo:
> 
> 
> The build for this failed twice on the same unit test, which has something to do with metrics. This is highly likely due to the change made to the metrics framework in this PR.

I think it may be due to the removal of Meters in `StatsProviderImpl`. This is not *required* for the fix, so will remove.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/SegmentStoreMetricsTests.java
```diff
@@ -161,30 +161,30 @@ public void testThrottlerMetrics() {
...
 
         int containerId = new Random().nextInt(Integer.MAX_VALUE);
         @Cleanup
         SegmentStoreMetrics.CacheManager cache = new SegmentStoreMetrics.CacheManager();
         cache.report(new CacheState(storedBytes, usedBytes, 0, allocatedBytes, storedBytes), generationSpread, managerIterationDuration);
 
-        assertEquals(storedBytes, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_STORED_SIZE_BYTES).value());
-        assertEquals(usedBytes, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_USED_SIZE_BYTES).value());
-        assertEquals(allocatedBytes, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_ALLOC_SIZE_BYTES).value());
+        assertEquals(storedBytes, (long) MetricRegistryUtils.getGauge(MetricsNames.CACHE_STORED_SIZE_BYTES).value());
```
kevinhan88, line 175: Please wait sufficient time before assertion (using assert utility tools), otherwise when the testing env is slow, the test will fail again randomly.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/SegmentStoreMetricsTests.java
```diff
@@ -161,30 +161,30 @@ public void testThrottlerMetrics() {
...
+        assertEquals(usedBytes, (long) MetricRegistryUtils.getGauge(MetricsNames.CACHE_USED_SIZE_BYTES).value());
+        assertEquals(allocatedBytes, (long) MetricRegistryUtils.getGauge(MetricsNames.CACHE_ALLOC_SIZE_BYTES).value());
         assertEquals(generationSpread, (int) MetricRegistryUtils.getGauge(MetricsNames.CACHE_GENERATION_SPREAD).value());
-        assertEquals(managerIterationDuration, (int) MetricRegistryUtils.getTimer(MetricsNames.CACHE_MANAGER_ITERATION_DURATION).mean(TimeUnit.MILLISECONDS));
+        assertEquals(managerIterationDuration, (long) MetricRegistryUtils.getTimer(MetricsNames.CACHE_MANAGER_ITERATION_DURATION).mean(TimeUnit.MILLISECONDS));
 
         cache.close();
 
         assertNull(MetricRegistryUtils.getGauge(MetricsNames.CACHE_STORED_SIZE_BYTES));
```
kevinhan88, line 183: Similarly, wait sufficient time (or retry) before the null check - this is the reason why the test always pass locally but sometimes fail in a loaded env.

co-jo: The failure was due to the metric being declared as static as mentioned in the description. Is there anything about this specific metric that warrants a timeout (as opposed to others). I feel like we should be consistent in defining how we perceive success/failure across all the metrics. I don't notice many metrics relying on retries or timeouts.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/SegmentStoreMetricsTests.java
co-jo: Same response as below.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/SegmentStoreMetricsTests.java
kevinhan88: The failure was caused by not closing metrics after CacheManager is closed, as the code shows.
Static metrics registry is a design dilemma of Micrometer, which might be enhanced, however it's not the direct cause.

We've experienced the random assertion problem (not the random number which was caused by not closing metrics properly), due to the test code talked to registry too soon, before the registry was able to finish the processing. Since we do have assertion tools to address the concern, then why not take this opportunity to make the test robust? - Isn't this fix part of the effort trying to recover from a recent test disaster due to hostile test env?

Remember in production code, metrics registry doesn't need to talk back to application code. So it is test code's responsibility to ensure things happen as expected.

co-jo: No, that was not the cause of the failure. The direct cause (for the issue introduced with this new metric) was because the metric was declared as static.

kevinhan88: Ok, I thought you were talking about "static" metric registry.

co-jo: There were two issues with the PR that updated the CacheManager metric:
* Declaring the metric as `static`. Depending on various timings, the logger likely resolves to a `NullStatsLogger` before the metrics provider is initialized, preventing any updates to the metric.
* The CacheManager metrics was not closed, leading to potential residual metrics being reported between disjoint tests.

This PR is not necessarily to make the test as robust as possible. I would prefer to address any timing issues with the `assert` statements in a separate issue as it is not related to issues related to this PR (or the PR prompting this one). Many other metrics would have the same problem and it should be addressed with a new issue.

PR #5094, opened 2020-08-24
Issue 5093: validate internal and external names

### LINKED ISSUE #5093: validate names with prefixes or suffixes
**Problem description**
Issue #5080 required the stream name and scope name to be restricted as per an external defect. This ensures that we treat the internal and external names differently.

**Problem location**
NameUtils.java - validateStreamName, validateUserStreamName

validate length

ravibeta:
**Change log description**  
Issue 5080 put restriction on the length of the name. This adds different restrictions to internal and external names.

**Purpose of the change**  
Fixes #5093

**What the code does**  
takes prefix or suffix into consideration when internal names are created.

**How to verify it**  
added unit-tests

### DIFF
--- a/shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
+++ b/shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
@@ -35,6 +35,21 @@ public final class NameUtils {
     public static final String READER_GROUP_STREAM_PREFIX = INTERNAL_NAME_PREFIX + "RG";
 
     /**
+     * Size of the prefix or suffix included with the user stream name.
+     */
+    public static final int MAX_PREFIX_OR_SUFFIX_SIZE = 5;
+
+    /**
+     * Size of the overall name as permitted by the host.
+     */
+    public static final int MAX_NAME_SIZE = 255;
+
+    /**
+     * Size of the name that can be specified by user.
+     */
+    public static final int MAX_GIVEN_NAME_SIZE = MAX_NAME_SIZE - MAX_PREFIX_OR_SUFFIX_SIZE;
+
+    /**
      * This is used for composing metric tags.
      */
     static final String TAG_SCOPE = "scope";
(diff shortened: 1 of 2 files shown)


shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
```diff
@@ -624,7 +638,7 @@ public static String getStreamForReaderGroup(String groupNameName) {
      */
     public static String validateUserStreamName(String name) {
         Preconditions.checkNotNull(name);
-        Preconditions.checkArgument(name.length() < 256, "Name must have less than 256 characters");
+        Preconditions.checkArgument(name.length() < MAX_GIVEN_NAME_SIZE + 1, String.format("Name cannot exceed %d characters", MAX_GIVEN_NAME_SIZE));
```
andreipaduroiu, line 641: Use `<=` to avoid having to do arithmetic every time this executes.
Do not use String.format. If you check the syntax of `Preconditions.checkArgument`, you'll see that it already takes a formatted string as arg.

shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
```diff
@@ -649,7 +663,7 @@ public static String validateStreamName(String name) {
 
         // In addition to user stream names, pravega internally created stream have a special prefix.
         final String matcher = "[" + INTERNAL_NAME_PREFIX + "]?[\\p{Alnum}\\.\\-]+";
-        Preconditions.checkArgument(name.length() < 256, "Name must have less than 256 characters");
+        Preconditions.checkArgument(name.length() < MAX_NAME_SIZE + 1, String.format("Name cannot exceed %d characters", MAX_NAME_SIZE));
```
andreipaduroiu, line 666: same here

ldr7:
Since the prefixes and suffixes are added to the stream, we can have scope names that are 255 characters long (UTF-8) and keep the stream name character length limit at 250. Have tried out this scenario. Although, this would require changes in the scope validation method.

PR #5101, opened 2020-08-25
Issue 5100: In Integration Tests, KVTable Creation should fail when partition count is 0

### LINKED ISSUE #5100: Integration tests pass when creating a KeyValueTable with 0 partitions
**Problem description**
Pravega Integration Tests, create a KeyValueTable successfully with 0 partitions in the KeyValueTable configuration.
This not expected as '0' is an invalid value for number of partitions and so KVTable creation should fail with "IllegalArgumentException"

**Problem location**
Integration tests for creating a KVTable with 0 partitions pass

Integration tests use "LocalController" implementation of "Controller" interface, which is **not** used by client. 
The Client uses ControllerImpl implementation of Controller interface which has a check for partitionCount and hence the table creation errors out when using ControllerImpl with '0' partitionCount.
As such this is not a real code issue but only an issue in the Integration Tests.

pbelgundi:

**Change log description**  
Integration tests use "LocalController" implementation of "Controller" interface, which is not used by the client.
The Client uses ControllerImpl implementation of Controller interface which has a check for partitionCount and hence the table creation errors out when using ControllerImpl with '0' partitionCount.
However, LocalController does not have this check and so creates the KVTable successfully even with a parititonCount=0.

**Purpose of the change**  
Fixes #5100

**What the code does**  
Added a check to ControllerService to check for partitionCount > 0 and throw and IllegalArgumentException if it is <=0.

**How to verify it**  
Added a new UT to ControllerServiceImplTest to cover the 0 partitionCount scenario.
Added a new test case to Integration Tests for creating a KeyValueTable with partitionCount 0.
Both above tests should return IllegalArgumentException for partitionCount=0.
All other unit and integration tests should pass.

### DIFF
--- a/client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
+++ b/client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
@@ -1898,6 +1898,11 @@ public class ControllerImplTest {
                 KeyValueTableConfiguration.builder().partitionCount(1).build());
         AssertExtensions.assertFutureThrows("Server should throw exception",
                 createKVTableStatus, Throwable -> true);
+
+        createKVTableStatus = controllerClient.createKeyValueTable("scope1", "kvtable6",
+                KeyValueTableConfiguration.builder().partitionCount(0).build());
+        AssertExtensions.assertFutureThrows("Server should throw exception",
+                createKVTableStatus, Throwable -> true);
     }
 
     @Test
--- a/client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
+++ b/client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
@@ -1901,8 +1901,8 @@ public class ControllerImplTest {
 
         createKVTableStatus = controllerClient.createKeyValueTable("scope1", "kvtable6",
                 KeyValueTableConfiguration.builder().partitionCount(0).build());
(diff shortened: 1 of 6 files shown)


client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
```diff
@@ -1898,6 +1898,11 @@ public void testCreateKeyValueTable() throws Exception {
                 KeyValueTableConfiguration.builder().partitionCount(1).build());
         AssertExtensions.assertFutureThrows("Server should throw exception",
                 createKVTableStatus, Throwable -> true);
+
+        createKVTableStatus = controllerClient.createKeyValueTable("scope1", "kvtable6",
+                KeyValueTableConfiguration.builder().partitionCount(0).build());
+        AssertExtensions.assertFutureThrows("Server should throw exception",
+                createKVTableStatus, Throwable -> true);
```
shiveshr, line 1905: it is better to have a stronger predicate check which looks for expected exception

shiveshr (approved):
LGTM, one minor comment

client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
pbelgundi: Sure. Fixed.

PR #5120, opened 2020-08-27
Issue 5119: (SegmentStore) Copy-on-Read for Table Segment Compaction

### LINKED ISSUE #5119: Segment Store Read Index should return a copy of cached data if requested
**Problem description**
This can cause problems if the user of the data is hanging on to the BufferViews for too long and we get a cache eviction.

**Problem location**
Read Index.

Add API (optional) to request data to be copied out of the cache if needed.

andreipaduroiu:
**Change log description**  
- Updated the `ReadResult` interface with the ability to specify "copy-on-read" behavior for any data that is returned from the cache. If set, any data that is returned from the cache is copied into a heap buffer and served from there. Default is not set.
- Updated `TableCompactor` to make use of this functionality.

**Purpose of the change**  
Fixes #5119.

**What the code does**  
- `ReadResult`
    - If copy-on-read is false (default), behavior is as before. Any cache data will be returned as a view over cache buffers/blocks that are involved. This is the desired behavior for streaming scenarios (tail reads, LTS reads, Tier1->LTS data copies) as these are not prone to unexpected cache eviction.
    - If copy-on-read is true, then any time a `ReadResultEntry` is returned that points to the cache, the cache data is copied into a heap buffer and that buffer is served to the caller. This is done in the Read Index so even if the underlying data is evicted (and eventually overwritten), the result will not be affected.
- `TableCompactor`
    - Setting copy-on-read for candidate reading. We need to decouple this data from the actual cache because there could be a significant delay between the time of reading and the time it is actually used again (data must be re-appended to the segment, so it's subject to Tier 1 write delay, queuing and throttling). This increases the likelihood of the original data being evicted from the cache, especially under cache pressure situations. 

**How to verify it**  
All tests must pass.


medvedevigorek:
@sachin-j-joshi  there is a successful run of system tests so let's proceed with merge. Thanks.

PR #5121, opened 2020-08-29
Issue 5082: Initial Java 11 Migration

### LINKED ISSUE #5082: Java 11 Migration - Compile using JDK11.
**Problem description**

Currently the Pravega project is built upon JDK8. Upgrading to JDK11 will provide various new APIs and tools to benefit the project. Enumerating and updating to use these new features is out of scope for this issue. The one caveat is that in order to continue support for _Flink 1.8_ , we must compile the `client` project and it's dependencies using JDK8.

**Problem location**

Largely `gradle.build`.

* Default the project builds to use `JavaVersion.VERSION_11`.
* Supply a `--release 8` flag in the `compilerArgs` for the `client`.
  * Do the same for it's dependencies.

co-jo:
**Change log description**  
Makes the following minimal set of changes to accomplish a passing compilation phase and unit testing phase. It keeps the client and all its dependencies building JDK8 jars.

**Purpose of the change**  
Fixes #5082 

**What the code does**  
* Updates TravisCI to deploy using JDK11.
* Dockerfile defaults to use JDK 11.
* Compile `client` using JDK 8 and it's dependencies.
* Compile everything else using JDK 11.
Note: This PR makes no Java API changes.

Not too many complications were encountered, but the following changes *had* to be made due to the bump to JDK11.

| Change | Reason |
| ------------- | ------------- |
| Bump to Gradle Version  | JDK 11 incompatible with Gradle 4.10.2, >= 5.0 required. |
| Bump Spotbugs Version | JDK 11 incompatible with SpotBugs 3.2.0 |
| Added `@SuppressFBWarnings` | Spotbugs does not parse JDK 11 bytecode correctly. See: https://github.com/spotbugs/spotbugs/issues/811 |
| Bump Lombok Version | Issue with versions <= 6.4. See: https://github.com/franzbecker/gradle-lombok/pull/77 |
| Add JVM flags | The `system:test` project deploys a pravega cluster using the operators, which use invalid JVM flags as of Java 10. |
| Segregate client, sever builds | To improve developer experience, add a flag that will control which Java version the client is built. |

**How to verify it**  


.travis.yml
```diff
@@ -12,7 +12,7 @@ git:
 language: java
 install: true
 jdk:
- - openjdk8
+ - openjdk11
```
eolivelli, line 15: The client must run well also on jdk8 how can we test it?

build.gradle
```diff
@@ -150,6 +142,10 @@ project('common') {
         //Do NOT add any additional dependencies here.
     }
 
+    tasks.withType(JavaCompile) {
+        options.compilerArgs.addAll(['--release', '8'])
```
medvedevigorek, line 146: can a build property be defined to control what Java version to use for Client and its dependencies? This is for the case if someone wants to checkout the code and build everything using JDK11. Currently there is a need to make edits in multiple places in `biuild.gradle`. It would be a much better experience if there will be a single property that controls it.

co-jo: I looked into that, but couldn't find any sort of function/option that can programmatically apply some config to dependent tasks. The best we can do is have some sort of flag and a check in each of these `task.withType(JavaCompile)` blocks.

medvedevigorek: That should work. As long as there is one place to turn Java11 on for Client and its dependencies - it's good enough

build.gradle
```diff
@@ -868,9 +884,9 @@ project('test:system') {
         systemProperty "bookkeeperImageName", System.getProperty("bookkeeperImageName", "bookkeeper") // bookkeeper image name
         systemProperty "zookeeperImageName", System.getProperty("zookeeperImageName", "zookeeper") // zookkeeper image name
         systemProperty "zookeeperImageVersion", System.getProperty("zookeeperImageVersion", "latest") // zookeeper image version
-        systemProperty "imageVersion", System.getProperty("imageVersion") // pravega version.
+        systemProperty "imageVersion", System.getProperty("imageVersion", "latest") // pravega version.
```
ravisharda, line 887: Should this be `latest`? Shouldn't it be using a specific named version instead?

ravisharda (changes requested):

build.gradle
co-jo: Likely -- I mostly followed suit from the `zookeeperImageVersion` property above. Any suggestions to a sain default?

.travis.yml
co-jo: Can you clarify a little bit? Would relying on the unit/integration tests be insufficient?

eolivelli: I can't see here where we are testing with jdk8

build.gradle
medvedevigorek: I don't think we should use `latest` as Pravega image here. Ideally `imageVersion` should be pointing to an image that was built from the current source code as this is the intend in most cases - run system tests for the current commit/branch.  If no `imageVersion` provided then system test run should fail

build.gradle
```diff
@@ -150,6 +142,12 @@ project('common') {
         //Do NOT add any additional dependencies here.
     }
 
+    if (System.getProperty("buildJDK11Client") == null) {
```
medvedevigorek, line 145: It's bit inconsistent to call this property this way as when default JDK is changed to the next version this property should be updated as well. Suggest to use a better name, something that will reflect the fact that Client will be build for the same release version as Server code.

medvedevigorek: Also please don't forget to update repository README file to reflect the changes in default JDK

medvedevigorek (changes requested):

.travis.yml
tkaitchuck: I believe we can set the `-source` and `-target` javac flags for the client which should preserve compatibility,

build.gradle
```diff
@@ -91,18 +90,11 @@ allprojects {
 
     gradle.projectsEvaluated {
         tasks.withType(JavaCompile) {
-            sourceCompatibility = JavaVersion.VERSION_1_8
-            targetCompatibility = JavaVersion.VERSION_1_8
+            sourceCompatibility = JavaVersion.VERSION_11
+            targetCompatibility = JavaVersion.VERSION_11
```
tkaitchuck, line 94: Is there a way to override this for the client?

.travis.yml
co-jo: I have kept all the system test pod defaults as is, so we can still rely on running those for some level of verification. I'm not sure if there is a standard convention (for the variable name) but now if a user has a `JAVA_8` env variable defined, it will attempt to run the unit tests using the executable that `JAVA_8` points to.

build.gradle
co-jo: I don't believe it is sufficient (or even possible) to supply overrides for the client (and its dependencies). For example if I supply 

```
    tasks.withType(JavaCompile) {
        sourceCompatibility = JavaVersion.VERSION_1_8
        targetCompatibility = JavaVersion.VERSION_1_8
    }
```

in any other project, it still builds Java 11 classes. Even if this was just a matter of ordering this configuration, I don't this it is sufficient. See: https://bugs.openjdk.java.net/browse/JDK-8058150

The way we currently override it is through the `--release` flag, which basically does the above anyway.

build.gradle
co-jo: Instead I added a user configurable property to set the client java version accordingly: `clientJavaVersion`. I opted for this instead of say, a `compileUsingSameJDK` flag because it essentially requires the same amount of user effort, but is more informative to what versions are being targeted.

build.gradle
co-jo: Should I just remove the default instead then? Obviously its not possible to determine the exact build tag ahead of time, but if you run the gradle docker task, it produces some sort of version based snapshot `<version>-<hash>-SNAPSHOT` but that build also gets tagged as `latest`, so I think it does cover the main use case.

medvedevigorek: I think it comes to where the build is published after the gradle docker task. If it's published to the same repository as the default one to pull image from for the purpose of  the system test run then it may work. But as I understand it's not always true hence it's better to require the user to explicitly mention what image to use. With that in mind the default image version should be removed and if the user doesn't provide it the system test run will fail.

build.gradle
medvedevigorek: okay, makes sense.

build.gradle
medvedevigorek: My understanding is that `--release` is not an equivalent of `--source` and `--target` , you should always use `--release` starting with Java9+. With that in mind, do we really need to specify `--source` and `--target` explicitly for the components we want to build with the current JDK (e.g. JDK11)?

For Client and its dependencies we should use `--release` explicitly as we do now.

co-jo: We don't need to explicitly include them no. I do however think it is a good idea to also include release flags for the non-client components. This should protect any users who are running some JDK version >11 from running into issues.

.travis.yml
eolivelli: @tkaitchuck  The best way is to set "-release 8" (as it is the current approach), otherwise javac will pick wrong overloaded versions in newer JDKs.

README.md
```diff
@@ -15,7 +15,9 @@ To learn more about Pravega, visit https://pravega.io
 
 ## Prerequisites
 
-- Java 8+
+- Java 11+
+
+The `client` project is expected to be built using *Java 8* bytecode. If you have a local *JDK 8* installation, the `client`'s unit tests can be deployed by setting the `JAVA_8` environment variable to point the *JDK  8* home directory.
```
eolivelli, line 20: " If you have a local *JDK 8* installation, the `client`'s unit tests can be deployed by setting the `JAVA_8` environment variable to point the *JDK  8* home directory."

I would drop this line, it does not add value.
It is important to state that Pravega Client and all its dependendecies is expected to work on Java 8 compatible runtimes

build.gradle
```diff
@@ -996,6 +1020,10 @@ def getProjectVersion() {
     return ver
 }
 
+def getDefaultJavaVersion() {
+    return JavaVersion.VERSION_11.getMajorVersion()
```
eolivelli, line 1024: why this complexity ? istn't "11" enough ? you are still hardcoding "VERSION_11" here

build.gradle
```diff
@@ -259,6 +261,19 @@ project('client') {
         testCompile group: 'io.netty', name: 'netty-transport-native-epoll', version: nettyVersion, classifier: 'linux-x86_64'
     }
 
+    tasks.withType(JavaCompile) {
+        options.compilerArgs << "--release" << clientJavaVersion
+    }
+
+    // If the user has the client to build with a different version and there exists an appropriate JDK,
```
eolivelli, line 268: Are we adding this block in order to run tests of client on jdk8 ?
I think this is not really needed. 
If we set --release = 8 in the client and in all dependencies it should be enough.

It is better to have some external project / integration tests suite that uses the client.
Having here such support looks to me overwhelming

eolivelli (changes requested):
we are on our way.
I left a few last suggestions.
IMHO there is no need to set up that JAVA_8 machinery, using --release 8 is enough.
The best way to test the client is with a separate test suite, that can run on CI as well

build.gradle
co-jo: Yeah, I debated that. Definitely unnecessary.

build.gradle
co-jo: This functionality was introduced in response to your comment here: https://github.com/pravega/pravega/pull/5121#discussion_r482239853. I'm happy with removing it if you think it is overkill.

eolivelli: @co-jo yes, to me is overkill. Let's remove this complexity.

gradle.properties
```diff
@@ -66,3 +66,5 @@ bookkeeperBaseTag=pravega/bookkeeper
 signing.keyId=
 # This will be defaulted to ~/.gnupg/secring.gpg if not provided as a command line property
 signing.secretKeyRingFile=
+# The version of Java to build the 'client' project in.
```
eolivelli, line 69: The version of Java to build the 'client' project and all dependencies in

README.md
```diff
@@ -15,7 +15,9 @@ To learn more about Pravega, visit https://pravega.io
 
 ## Prerequisites
 
-- Java 8+
+- Java 11+
+
+The `client` project is expected to be built into *Java 8* bytecode.
```
eolivelli, line 20: I would say something like that:
In spite of the requirements of using JDK11+ to build this project, Pravega client artifacts are expected to be compatible with Java 8 runtime, only the server side components need JDK11+ to run.

build.gradle
```diff
@@ -150,6 +140,10 @@ project('common') {
         //Do NOT add any additional dependencies here.
     }
 
+    tasks.withType(JavaCompile) {
```
shrids, line 143: nit: Can we have all the compile options at a single place ? ref: https://github.com/pravega/pravega/blob/4f7be177996bd5d3a05e2ab6f117d5a40b1e2207/gradle/java.gradle#L20-L30

`java.gradle `seems to be the place were this change can be added.

co-jo: We could, but we would have to surround the `compileJava` closure with inside a call to `afterEvaluate`. That doesn't seem like a good practice to me, but its doable.

build.gradle
```diff
@@ -584,10 +595,15 @@ project('shared:controller-api') {
         compile "io.grpc:grpc-protobuf:" + grpcVersion
         compile "io.grpc:grpc-stub:" + grpcVersion
         compile group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: nettyBoringSSLVersion
-         // Since Java 10 javax.annotation is no more bundled with the JRE
+        // Since Java 10 javax.annotation is no more bundled with the JRE
```
andreipaduroiu, line 598: nit: "no more" -> "no longer"

gradle.properties
```diff
@@ -22,12 +22,12 @@ commonsioVersion=2.6
 commonsLang3Version=3.7
 dockerClientVersion=8.16.0
 ecsObjectClientVersion=3.0.6
-spotbugsVersion=4.0.3
-spotbugsAnnotationsVersion=4.0.3
-spotbugsPluginVersion=1.6.9
+spotbugsVersion=4.0.6
+spotbugsAnnotationsVersion=4.0.6
+spotbugsPluginVersion=4.4.4
```
andreipaduroiu, line 27: LOL. This is a big jump 👍

gradle/java.gradle
```diff
@@ -98,3 +104,7 @@ plugins.withId('java') {
         testCompile "org.mockito:mockito-core:${mockitoVersion}"
     }
 }
+
+def getDefaultJavaVersion() {
```
andreipaduroiu, line 108: I see you defined this in 2 places. Since this file is imported into main `build.gradle` file, does it suffice to just declare it here?

gradle.properties
co-jo: Full disclosure, I basically referenced what Tom did in his `faster-build` branch for these exact versions.  @tkaitchuck Any thoughts?

gradle/java.gradle
co-jo: We can pull it out of `java.gradle` but not `build.gradle`. Will remove this declaration.

test/system/kubernetes/setupTestPod.sh
```diff
@@ -92,7 +92,7 @@ spec:
         - mountPath: "/data"
           name: task-pv-storage
 EOF
-kubectl wait --timeout=1m --for=condition=Ready pod/task-pv-pod
+kubectl wait --timeout=5m --for=condition=Ready pod/task-pv-pod
```
sachin-j-joshi, line 95: Is this change intentional ?

co-jo: It was. Five minutes might seem excessive but I was running into issues with the system tests from this pod timing out. I can lower if if you think its too high, but I think its a better experience to have a longer timeout than having to redeploy the tests.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -288,6 +290,16 @@ private static boolean isOperatorVersionAbove040() {
         }));
     }
 
+    // Removal of the JVM option 'UseCGroupMemoryLimitForHeap' is required with JVM environments >= 10. This option
+    // is supplied by default by the operators. We cannot 'deactive' it using the XX:- counterpart as it is unrecognized.
```
medvedevigorek, line 294: @co-jo can you please file an issue for Pravega Operator to remove usage of `UseCGroupMemoryLimitForHeap` for Pravega 0.9+

medvedevigorek (approved):
Looks good to me. Thanks

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
co-jo: > 
> 
> @co-jo can you please file an issue for Pravega Operator to remove usage of `UseCGroupMemoryLimitForHeap` for Pravega 0.9+

Sure.

test/system/kubernetes/setupTestPod.sh
sachin-j-joshi: ok. makes sense.

sachin-j-joshi (approved):
LGTM.

ravisharda (approved):

ravisharda:
@co-jo please run systems tests against the latest code in this branch, if you haven't already done so.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -288,6 +290,16 @@ private static boolean isOperatorVersionAbove040() {
...
+    private String[] getSegmentStoreJVMOptions() {
+        return new String[]{"-XX:+UseContainerSupport", "-XX:+IgnoreUnrecognizedVMOptions"};
+    }
+
+    private String[] getControllerJVMOptions() {
+        return new String[]{"-XX:+UseContainerSupport", "-XX:+IgnoreUnrecognizedVMOptions"};
```
tkaitchuck, line 300: Why do we need IgnoreUnrecognizedVMOptions?

tkaitchuck (approved):

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
co-jo: The Pravega Operator supplies certain default JVM options (`+UseCGroupMemoryLimitForHeap`), which are not recognized by later JVMs. Any kubernetes deployment wouldn't work until the Pravega Operator is adjusted accordingly.

shrids (approved):

andreipaduroiu (approved):

andreipaduroiu:
@eolivelli Do you have any further comments on this? Every other reviewer approved the PR but I want to make sure you're good with it.

eolivelli (approved):
LGTM

thank you @andreipaduroiu  for checking

great work @co-jo

andreipaduroiu (changes requested):
All (still) looks good, however holding off in order to figure out some potential issues with JRE11.

andreipaduroiu (approved):

Issue #5122, opened 2020-08-29
Improving Windows Compatibility


co-jo:
**Problem description**

After using Windows as my main programming environment, I've come across several issues specific to Windows environments. This will be used as a general target for submitting improvements to support better cross-platform development.

**Problem location**

`pravega/**/*`

**Suggestions for an improvement**

Fix any Windows compatibility issues as needed.

co-jo:
Observed an issue (as described in [1]) where the `javadocs` tasks constructs `javadoc.exe` commands that exceed the limit of 8192 characters [2]. This due to the excessive length of the classpath. 

[1] https://stackoverflow.com/questions/10519558/createprocess-error-206-the-filename-or-extension-is-too-long-when-running-main
[2] https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation

co-jo:
> 
> 
> Observed an issue1 where the `javadocs` tasks constructs `javadoc.exe` commands that exceed the limit of 8192 characters2. This due to the excessive length of the classpath.
> 
> I specifically encountered this during the `:standalone:delombok` task, but in principle other projects/tasks are susceptible to this issue.
> 
>     1. https://stackoverflow.com/questions/10519558/createprocess-error-206-the-filename-or-extension-is-too-long-when-running-main
> 
>     2. https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation

Looks to be resolved as of Gradle 6.0, which we will get once the JDK 11 migration work is merged in.

* https://github.com/gradle/gradle/pull/10544

co-jo:
The Wiki repository uses invalid characters (for Windows files) in the naming of their files.

PR #5123, opened 2020-08-29
Issue 5122: Windows Compatibility Enhancements

co-jo:
**Change log description**  

1. Added some documentation recognizing potentially slow unit tests (those involved in touching a large amount of files) and a possible fix.
2. Target the proper platform specific temp directory within `S3ProxyImpl`.
3. Surround a file clean procedure with a try/catch block to absorb the exception.

**Purpose of the change**  
Makes progress towards: #5122 

**What the code does**  

2. There are two benefits for this change. `S3ProxyImpl`  will save files relative to the root of the drive. Firstly for aesthetic reasons, we should avoid dumping files at the top level of someone's `C:` drive, and place them into the proper `Temp` folder. More importantly, because it is placed in a `tmp` folder at the root, Windows will place extra permissions on this folder, which can prevent the creation (and deletion) of files under it, causing the test to fail. 
3. This procedure often fails because Windows (I believe unlike Unix systems) automatically places a FileLock on files, which we later attempt to delete. 

It does not look like we do any extra work to try and recover from `IOException`s here, so it may be appropriate simply to log the exception instead of passing it along. One issue with this is that these temporary files may accumulate over time (Well, this happens anyways..).

**How to verify it**  

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
       script: ./gradlew javadocs
 
     - stage: snapshot
-      if: (branch = master OR branch =~ /^r[0-9]+\.[0-9]+/) AND NOT (type = pull_request)
+      if: branch = master AND NOT (type = pull_request)
       env:
          - secure: "EJGO5B4XxT6NpZ4+St4EG/Y3+UmQ2XNkbnJz/5vv7HZziC4OXFVFCigOsjqqRVjJaWmDmYf1a/U7QwJa+esgA9oBchg+DVWYeqkMOEbZkEJzBgHeA5Bmgp+3Q5/2ouOe6M2cUFhSNXN+lZaLri8xYis25ndK6E0Y6MIQ7CbCNyhA/edIIHSttJNHoGkaVxkfEDWNvIvdUbxt7MTTdAGIvT+Xj/EF8clT349N5UfZ8DdYy8Q6i4TKiy5557dj6nfegbD6qD0mGxWt8dqKBSIqmPMoiCjGsy+3LyfeMNhZiNji+7LRmwJtWD87fewlYeU5xkKZIaxXlCnd9JPrkIFH62FzXctn0wc18mDTD+c4mnc2efKHHrmvRcrvQrRD447iFt17lEMCROriHVZsB4m6igxrEg0i2paCBajMqhQG5LEwSqU7D8CQaW+ffC2+9agFR0a6uhW7n9GahU2AOVf9RhvLU7yFWtVJltOP2mBY0s8Oowxzy5OXSfqjxwSJeKKGgWVaTbBY14Y1Y2pCx9CuInLswz4oIO3LqHHfKJasKwHp9nHRSa6flh8JFR8+jGA7vo1MqGAPH3N4rrMQqpYs9DjYD50TPAh2h4asQj5gSjtYkKpvEu58dRRRzansjtSpiwo6TNfaVp9jhzYI3O3vnssyDyJvHjHjCnrvCB/KYrU="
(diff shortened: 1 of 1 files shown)


segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperServiceRunner.java
```diff
@@ -258,9 +258,13 @@ private void cleanupDirectories() throws IOException {
     }
 
     private void cleanupDirectories(HashMap<?, File> toDelete) throws IOException {
-        for (File dir : toDelete.values()) {
-            log.info("Cleaning up " + dir);
-            FileUtils.deleteDirectory(dir);
+        try {
+            for (File dir : toDelete.values()) {
+                log.info("Cleaning up " + dir);
+                FileUtils.deleteDirectory(dir);
```
eolivelli, line 264: You'd better wrap with a try/catch block this line and not the whole loop

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperServiceRunner.java
```diff
@@ -258,9 +258,13 @@ private void cleanupDirectories() throws IOException {
...
+            }
+        } catch (IOException e) {
+            log.info("{}", e);
```
eolivelli, line 267: Please report the directory

eolivelli (changes requested):

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperServiceRunner.java
```diff
@@ -260,7 +260,11 @@ private void cleanupDirectories() throws IOException {
     private void cleanupDirectories(HashMap<?, File> toDelete) throws IOException {
         for (File dir : toDelete.values()) {
             log.info("Cleaning up " + dir);
-            FileUtils.deleteDirectory(dir);
+            try {
+                FileUtils.deleteDirectory(dir);
+            } catch (IOException e) {
+                log.info("Failed deleting directory: {} - {}", dir, e);
```
eolivelli, line 266: Sorry, I missed to tell that you don't need '{}' for the exception. The logger will print the message and the stacktrace anyway.

eolivelli (approved):
LGTM

I left just one last nit

tkaitchuck (approved):

build.gradle
```diff
@@ -625,7 +625,7 @@ project('controller') {
     }
 
     applicationDistribution.into('') {
-        def pluginDirBase = new File('/tmp/dummy-dir')
+        def pluginDirBase = new File(System.getProperty("java.io.timpdir", "/tmp") + "/dummy-dir")
```
andreipaduroiu, line 628: typo

andreipaduroiu (changes requested):

common/src/main/java/io/pravega/common/io/filesystem/FileOperations.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.commons.io.FileUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+
+@Slf4j
+public class FileOperations {
+
+    public static void cleanupDirectories(HashMap<?, File> toDelete) throws IOException {
```
co-jo, line 22: Some context: a way was needed to add coverage to the catch block added in `cleanupDirectories`, but moved somewhat nuanced. We needed a platform independent way to force an `IOException`. Given the attachment to the `BookKeeperServiceRunner` we were quite limited in what we could do.

Refactoring this method out was not ideal, but it was motivated by the following:
* It allowed us to avoid modifying any of the visibility rules of field/methods under `BookKeeperServiceRunner`. 
* It also allows us to avoid adding a completely contrived and seemingly random test under `BookKeeperServiceRunnerTests`.

PR #5127, opened 2020-08-31
Issue 5126: disabled metrics by default to prevent misconfigs

### LINKED ISSUE #5126: Turn off Metrics by default
**Problem description**

When we disabled StatsD by default, Controller in System test started failing due to no metrics exporter defined by metrics is enabled.

Here is the troubleshooting & finding:
1. we have two sets of configs, one for segmentstore, and one for controller, the two sets are sort of independent conceptually
2. however, metrics configs, among a few others, are different - they could be shared by segmentstore and controller in almost all the cases, long time ago we designed a mechanism to propagate those shareable configs - just discovered this mechanism never worked as expected. 
3. Why almost all deployments worked in the past? because we almost always define those shared configs in deployment yaml, so the broken propagation issue didn't surface.
4. Why system tests worked without overwriting? the controller metrics was always enabled due to the bug even though we turned it off in config file; coincidentally StatsD was also enabled in the past, so thing looked alright to the controller in system test.
5. Why it's broken now? we just disabled statsD by default to prevent accidental turning on of it. Then for controller in system test, metrics is enabled, but no exporter is defined, which is an obvious illegal situation, the code reported the error and quit, which is great!

Short term solution is to turn off metrics by default for controller. This issue is for short solution.
Long term solution is to fix the bug in config propagation mechanism, which is to be addressed in a separate ticket.

**Problem location**
MetricsConfig.java

Disable "statistics.enable" inside MetricsConfig.java

kevinhan88:
Signed-off-by: Kevin Han <kevinhan88@gmail.com>

**Change log description**  
We disabled metrics inside config.properties, expecting metrics is off for scenario such as system tests, unless it's enabled explicitly.
This config only works for Segmentstore but not controller, due to a bug in Controller's config propagation mechanism.
We recently disabled StatsD by default, hence Controller in system test hit an obvious illegal situation - metrics is enabled, but no exporter is defined.
This change is to disable metrics by default inside source code MetricsConfig.java, so the bug above will not have negative impact.

Long term solution is to be addressed separately.

**Purpose of the change**  
Fixes #5126 

**What the code does**  
Disabled ENABLE_STATISTICS by default.

**How to verify it**  
Unit tests should pass;
Controller in system test should stop complaining No metrics exporter defined when metrics is enabled.

### DIFF
--- a/controller/src/test/java/io/pravega/controller/task/Stream/IntermittentCnxnFailureTest.java
+++ b/controller/src/test/java/io/pravega/controller/task/Stream/IntermittentCnxnFailureTest.java
@@ -99,6 +99,7 @@ public class IntermittentCnxnFailureTest {
     @Before
     public void setup() throws Exception {
         MetricsConfig metricsConfig = MetricsConfig.builder()
+                .with(MetricsConfig.ENABLE_STATISTICS, true)
                 .with(MetricsConfig.ENABLE_STATSD_REPORTER, false)
                 .build();
         metricsConfig.setDynamicCacheEvictionDuration(Duration.ofSeconds(60));
--- a/shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java
+++ b/shared/metrics/src/main/java/io/pravega/shared/metrics/MetricsConfig.java
@@ -22,7 +22,7 @@ import java.time.Duration;
  */
 public class MetricsConfig {
     //region Config Names
-    public final static Property<Boolean> ENABLE_STATISTICS = Property.named("statistics.enable", true, "enableStatistics");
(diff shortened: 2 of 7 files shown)


sachin-j-joshi:
This seems not related with the change
```
io.pravega.controller.task.Stream.IntermittentCnxnFailureTest > failedScopeOperationsTest FAILED
        at io.pravega.controller.task.Stream.IntermittentCnxnFailureTest.failedScopeOperationsTest(IntermittentCnxnFailureTest.java:161)
```

kevinhan88:
Yes, it's fixed

medvedevigorek:
@sachin-j-joshi please have this PR merged as system tests ran with no issues on this branch. Thanks.

PR #5129, opened 2020-08-31
Issue 5128: refactor a test source so that the bindings can be tested uniformly

### LINKED ISSUE #5128: Refactor test code for easier sharing across consumers
**Problem description**
The LTS storage bindings are tested uniformly across bindings so the test artifacts are declared as interfaces or abstract classes and shared across bindings. Only one of these artifacts happens to be in the wrong place and could be included with the others. @sachin-j-joshi  prefers this.

**Problem location**
IdempotentStorageTestBase.java is used to test filesystem and extendeds3 binding and it should appear along with the StorageTestBase instead of the bindings test source.

This is a one time request for refactor and specific to Pravega storage bindings.

ravibeta:
**Change log description**  
This moves a test for use with bindings. Please refer issue #5128 for context. The location of the IdempotentStorageTestBase posed inconvenience in separating bindings and testing them. This fix moves the IdempotentStorageTestBase from bindings to segmentstore storage where we use other files just like that. @sachin-j-joshi prefers this.

**Purpose of the change**  
Fixes #5128 

**What the code does**  
This change affects the use of IdempotentStorageTestBase file since it was the only code that required to be shared from the bindings test. It could not be used from where it is without additional changes on the consumer side.

**How to verify it**  
These steps can be verified by running the gradle command to run tests.


segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/IdempotentStorageTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.concurrent.CompletableFuture;
+import lombok.val;
+import org.junit.Assert;
+import org.junit.Test;
+
+import static io.pravega.test.common.AssertExtensions.assertMayThrow;
+import static io.pravega.test.common.AssertExtensions.assertSuppliedFutureThrows;
+
+/**
+ * Common Unit tests for FileSystemStorage and ExtendedS3Storage.
```
sachin-j-joshi, line 25: While we are at it, let's also fix the comment w. 
Should read ` Common Unit tests for Storage implementations that allow idempotent writes.`

ravibeta: Yes Sachin

PR #5133, opened 2020-09-01
Issue 5132: segmentstore should use host name when TLS is enabled in standalone

### LINKED ISSUE #5132: Standalone doesn't support hostname for segmentstore
**Problem description**
When TLS is enabled in standalone, segmentstore endpoints fetched from client controller API should be `hostname:port` in order to support client side hostname verification. 

**Problem location**
standalone

Tristan1900:
**Change log description**  
When TLS is enabled in standalone, endpoints fetched from client controller API should be in `hostname:port` form.

**Purpose of the change**  
Fix #5132


standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
```diff
@@ -278,6 +278,7 @@ private void startLocalSegmentStore(int segmentStoreId) throws Exception {
                         .with(ServiceConfig.KEY_FILE, this.keyFile)
                         .with(ServiceConfig.CERT_FILE, this.certFile)
                         .with(ServiceConfig.ENABLE_TLS_RELOAD, this.enableTlsReload)
+                        .with(ServiceConfig.LISTENING_IP_ADDRESS, this.enableTls ? "localhost" : "")
```
ravisharda, line 281: Curious, shouldn't this be `PUBLISHED_IP_ADDRESS` instead? 

I understand that you are trying to change this from IP address `127.0.1.1` that the Segment store listens on to `localhost`, so that hostname verification can be turned on for client/controller-to-segment store communication.

ravisharda (changes requested):

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
Tristan1900: Thanks Ravi, yes, your understanding is correct, this is for hostname verification for segmentstore. 

Actually I want to ask what's the difference between Published and Listening address? The reason I use Listening address is from this [comment](https://github.com/pravega/pravega/blob/c7ac009970787df2633163644691aa3d72fc1a4e/segmentstore/server/src/test/java/io/pravega/segmentstore/server/store/ServiceConfigTests.java#L28), looks like if we don't specify published address, it will default to listening address.

ravisharda: Yes, if published IP and port are not registered with the Controller for the Segment Store service, the Controller returns the service's listening IP and port to the clients afaik. cc @andreipaduroiu

ravisharda:
@Tristan1900 Did you still want to get this in? Checking as this PR has been open and left unattended for long.

Tristan1900:
@ravisharda yes, I will check with Andrei with the issue tomorrow. Thanks for the notice.

Tristan1900:
@ravisharda Andrei says we might need to set both. My understanding is that `listening IP` is the one that must to be set, if `published IP` is the same as `listening IP`, then we can choose not to explicitly set it because it will default to `listening IP`. But I set them both in this PR anyway. 
Thanks

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
```diff
@@ -278,6 +278,8 @@ private void startLocalSegmentStore(int segmentStoreId) throws Exception {
...
+                        .with(ServiceConfig.PUBLISHED_IP_ADDRESS, this.enableTls ? "localhost" : "")
```
ravisharda, line 282: @Tristan1900 one more ask: could you please extract the `localhost` string into a constant and use that here and above, so that it can be easily changed in one place (to say loopback address `127.0.0.1` or another interface during testing).

ravisharda: Also, curious isn't the condition used here wrong? It sets `LISTENING_IP_ADDRESS` and `PUBLISHED_IP_ADDRESS` to empty string, if TLS is not enabled. Am I missing something?

Tristan1900: Sure, added. Thanks for pointing out.

Empty string is the default value for those two parameters. The constructor will replace the empty string with the right ip address in [here](https://github.com/pravega/pravega/blob/0701daa7213eb8617603d2df725700c5c1ad51bb/segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java#L321)

andreipaduroiu (approved):

ravisharda (approved):
LGTM.

PR #5145, opened 2020-09-04
Issue 5144: Force token refresh when segment store indicates token expiry

### LINKED ISSUE #5144: Unsynchronized clock times across client and the server can prevent token renewal
**Problem description**
If Segment Store is running on a host that is ahead of the client w.r.t. system time, segment store may reject clients' requests due to delegation tokens expiry. However, due to time difference, the client may evaluate that the token hasn't yet expired/nearing expiry. As a result, the client may end up reusing the preexisting token in retries causing continuous failures for some time.

**Problem location**

When the segment store informs the client of token expiry, the client must renew the token mandatorily, rather than evaluate token expiry on its own.

ravisharda:
**Change log description**  
Force token refresh when the Segment Store indicates token expiry.

**Purpose of the change**  
Fixes #5144 

**What the code does**  
Implements a signaling mechanism such that the client can signal the DelegationTokenProvider of the token expiry so that it refreshes the token upon the next call to `retrieveToken()`. 

**How to verify it**  
All tests must pass.

### DIFF
--- a/client/src/main/java/io/pravega/client/connection/impl/RawClient.java
+++ b/client/src/main/java/io/pravega/client/connection/impl/RawClient.java
@@ -10,6 +10,7 @@
 package io.pravega.client.connection.impl;
 
 import io.pravega.auth.AuthenticationException;
+import io.pravega.auth.TokenExpiredException;
 import io.pravega.client.control.impl.Controller;
 import io.pravega.client.segment.impl.Segment;
 import io.pravega.client.stream.impl.ConnectionClosedException;
@@ -78,7 +79,11 @@ public class RawClient implements AutoCloseable {
         @Override
         public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenCheckFailed) {
             log.warn("Auth token check failed on segment {} with {}", segmentId, authTokenCheckFailed);
-            closeConnection(new AuthenticationException(authTokenCheckFailed.toString()));
+            if (authTokenCheckFailed.isTokenExpired()) {
+                closeConnection(new TokenExpiredException(authTokenCheckFailed.getServerStackTrace()));
+            } else {
+                closeConnection(new AuthenticationException(authTokenCheckFailed.toString()));
+            }
         }
     }
 
--- a/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
+++ b/client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
@@ -18,6 +18,7 @@ import io.pravega.common.util.ConfigurationOptionsExtractor;
 import java.time.Instant;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import lombok.AccessLevel;
 import lombok.Getter;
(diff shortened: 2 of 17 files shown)


client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -132,17 +134,22 @@ private JwtTokenProviderImpl(Controller controllerClient, String scopeName, Stri
...
     public CompletableFuture<String> retrieveToken() {
         DelegationToken currentToken = this.delegationToken.get();
-
+        final CompletableFuture<String> result;
         if (currentToken == null) {
-            return this.refreshToken();
+            result = this.refreshToken();
         } else if (currentToken.getExpiryTime() == null) {
-            return CompletableFuture.completedFuture(currentToken.getValue());
+            result = CompletableFuture.completedFuture(currentToken.getValue());
```
shrids, line 141: TokenExpirySignal has a higher priority and hence should be checked before returning a token right ?

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -78,7 +79,11 @@ public void processingFailure(Exception error) {
         @Override
         public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenCheckFailed) {
             log.warn("Auth token check failed on segment {} with {}", segmentId, authTokenCheckFailed);
-            closeConnection(new AuthenticationException(authTokenCheckFailed.toString()));
+            if (authTokenCheckFailed.isTokenExpired()) {
+                closeConnection(new TokenExpiredException(authTokenCheckFailed.getServerStackTrace()));
```
shrids, line 83: would this cause a change to the SegmentHelper logic at https://github.com/pravega/pravega/blob/4f7be177996bd5d3a05e2ab6f117d5a40b1e2207/controller/src/main/java/io/pravega/controller/server/SegmentHelper.java#L624-L641 ?

shrids (changes requested):

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
ravisharda: If the token has no expiry time set on it, the signal wouldn't be interpreted as token having been expired. Therefore, we do this check before the check for the signal.

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
ravisharda: Thanks. I've pushed the fix in the latest commit.

shrids (approved):
Please run system tests on this changeset.

ravisharda:
There was a successful system test run with test-pravega-continuous-1/942/, using both security enabled and disabled.

shiveshr (approved):

client/src/main/java/io/pravega/client/security/auth/JwtTokenProviderImpl.java
```diff
@@ -132,17 +134,22 @@ private JwtTokenProviderImpl(Controller controllerClient, String scopeName, Stri
...
+        } else if (this.tokenExpirySignal.get()) {
+            log.debug("Token was signaled as expired for scope/stream {}/{}", this.scopeName, this.streamName);
+            result = refreshToken();
+            this.tokenExpirySignal.compareAndSet(true, false);
```
tkaitchuck, line 145: This should be done inside of refreshToken in case on of the other branches occurs right before this is set.

PR #5146, opened 2020-09-05
Issue 5140: Ensure ControllerResolverFactory ensure the scheduled task is interrupted during shutdown.

### LINKED ISSUE #5140: Does ControllerResolverFactory always properly shutdown?
**Problem description**
ControllerResolverFactory has a shutdown method which is not being invoked by ControllerImpl

**Problem location**
ControllerImpl.close()

Call shutdown in close.

shrids:

**Change log description**  
- Ensure ControllerNameResolver#shutdown cancels the future the scheduled future with `mayInterruptIfRunning `flag as true.
- Wait for upto 20 seconds for channel termination to complete.
- Ensure watermarking thread pool is shutdown with clientFactory.close().

**Purpose of the change**  
Fixes #5140 

**What the code does**  
- Ensure ControllerNameResolver#shutdown cancels the future the scheduled future with `mayInterruptIfRunning `flag as true.
- Wait for upto 20 seconds for channel termination to complete.
- Ensure watermarking thread pool is shutdown with clientFactory.close().

**How to verify it**  
All the existing tests should pass.


client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -159,7 +159,9 @@ static WireCommand readCommand(InputStream in, IoBuffer buffer) throws IOExcepti
         }
         
         public void stop() {
-            stop.set(true);
+            if (stop.getAndSet(true)) {
```
ravisharda, line 162: Should the get and set separated out here? 
* This method returns if `get` `stop` returns true. 
* Set is called after the callback is complete, using compareAndSet(false, true). 

This might be especially useful if `callback.connectionDropped() can fail with an exception. 

On the other hand, if the failure of `connectionDropped()` call can be safely discounted, the existing code'd be OK.

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
```diff
@@ -186,6 +186,7 @@ public void pruneUnusedConnections() {
     }
 
     @Override
+    @Synchronized
```
ravisharda, line 189: Why is this change required?

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
```diff
@@ -244,8 +244,10 @@ private Segment getSegmentForRevisionedClient(String scope, String streamName) {
 
     @Override
     public void close() {
-        controller.close();
+        // wait for 5 seconds before forcibly terminating the watermarkReader threads.
+        ExecutorServiceHelpers.shutdown(watermarkReaderThreads);
```
ravisharda, line 248: nit: There is no timeout duration specified, unlike what the comment suggests. This comment might be referring on the default value used in the `shutdown` method. If someone changes that value, this comment would go stale. It might be useful rather to say something like `Wait for default timeout duration before...`

ravisharda (changes requested):

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
shrids: yeah, the idea is the to ensure that we do not stop and already stopped/closed connection.

client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java
shrids: updated it.

client/src/main/java/io/pravega/client/connection/impl/ConnectionPoolImpl.java
shrids: Yeah, there can be a `getClientConnection()` invocation which would update the map after a connection creation while this close() method is invoked.

ravisharda (approved):

shrids:
System test with this changeset completed successfully: https://asd-nautilus-jenkins.isus.emc.com/job/test-pravega-continuous-1/941/

client/src/main/java/io/pravega/client/control/impl/ControllerResolverFactory.java
```diff
@@ -201,7 +201,7 @@ public void start(Listener listener) {
         public void shutdown() {
             shutdown = true;
             if (scheduledFuture != null) {
-                scheduledFuture.cancel(false);
+                scheduledFuture.cancel(true);
```
andreipaduroiu, line 204: FYI, this has no effect. It is not implemented in the JRE.

andreipaduroiu (approved):

client/src/main/java/io/pravega/client/control/impl/ControllerResolverFactory.java
shrids: This is a ScheduledFuture on which cancel() is invoked.
` public ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit);`
As I understand your comment applies only on a CompletableFuture. let me know...

andreipaduroiu: Ah I missed that one. Nevermind then :)

tkaitchuck (commented):
Looks good

PR #5148, opened 2020-09-06
Issue 5147: Added check for service in init_kubernetes.sh script

### LINKED ISSUE #5147: Modify init_kubernetes.sh script to handle race condition in case ss pod is created before service
**Problem description**
init_kubernetes.sh script  is not handling the case where ss pod might come up before the service when External Access is enabled and that's causing problem when domainName is specified in the manifest.

**Problem location**
docker/pravega/scripts/init_kubernetes.sh

init_kubernetes.sh script should not allow to execute https://github.com/pravega/pravega/blob/4f7be177996bd5d3a05e2ab6f117d5a40b1e2207/docker/pravega/scripts/init_kubernetes.sh#L50, before verifying that the service exsists.

Prabhaker24:
Signed-off-by: prabhaker24 <prabhaker.saxena@dell.com>

**Change log description**  
init_kubernetes.sh script is not handling the case where ss pod might come up before the service when External Access is enabled and that's causing a problem when domainName is specified in the pravega manifest as in that case ss pods might get the wrong publishedIPAddress. 

**Purpose of the change**  
Fixes #5147 

**What the code does**  
Script checks for the presence of service in case we deploy pravega with external service enabled and when it finds the service then only it will set the publishedIPAddress.
When a service doesn't exists segment store pod will go into **CrashLoopBackOff** state and will be in that state only till the External service comes up post the service starts segment store pod will also recover and will set the correct publishedIPAddress.

**How to verify it**  
1) Deploy Pravega with external access enabled and delete service continuously and delete the pod once when the new pod comes up it will go into **CrashLoopBackOff** state and will wait continuously for the service to be there and will have the following message in its logs
```
Failed to get External Service. Exiting...
```
2) now stop deleting the service and the service will come up post that ss pod will recover and will get the correct publishedIPAddress.
I have checked both of the above scenarios.

### DIFF
--- a/docker/pravega/scripts/init_kubernetes.sh
+++ b/docker/pravega/scripts/init_kubernetes.sh
@@ -49,12 +49,12 @@ init_kubernetes() {
         export PUBLISHED_PORT=""
         local service=$( k8 "${ns}" "services" "${podname}" .kind )
 
-        while [[  "$service" != "Service" ]]; 
-        do
-            echo "Trying to get service"     
-            service=$( k8 "${ns}" "services" "${podname}" .kind )
-        done
-	
+        if [[  "$service" != "Service" ]];
+        then 
+            echo "Failed to get Service"
+            exit 1     
+        fi
+
 	export PUBLISHED_ADDRESS=$( k8 "${ns}" "services" "${podname}" ".metadata.annotations[\"external-dns.alpha.kubernetes.io/hostname\"]" )
 
         if [[ -n ${PUBLISHED_ADDRESS} && "${PUBLISHED_ADDRESS:${#PUBLISHED_ADDRESS}-1}" == "." ]];
--- a/docker/pravega/scripts/init_kubernetes.sh
+++ b/docker/pravega/scripts/init_kubernetes.sh
@@ -47,7 +47,7 @@ init_kubernetes() {
         local podname=${POD_NAME}
         export PUBLISHED_ADDRESS=""
         export PUBLISHED_PORT=""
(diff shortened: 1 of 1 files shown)


gaddamas (commented):
indentation of lines 52-56 appears in correct, need to align with line 50

Prabhaker24:
> indentation of lines 52-56 appears incorrect, need to align with line 50

Done.

docker/pravega/scripts/init_kubernetes.sh
```diff
@@ -47,6 +47,13 @@ init_kubernetes() {
         local podname=${POD_NAME}
         export PUBLISHED_ADDRESS=""
         export PUBLISHED_PORT=""
+        local service=$( k8 "${ns}" "services" "${podname}" .kind )
+
+        while [[  "$service" != "Service" ]];
```
shrids, line 52: as discussed with Ravi, I see this as a Precondition check before starting Segment store. And the startup  on k8s should not happen if the precondition fails.

shrids (changes requested):
One comment. 
Also, please ensure you have verified that the system tests which use k8s to deploy pravega work fine too...

docker/pravega/scripts/init_kubernetes.sh
Prabhaker24: I have removed the loop and have added exit condition now.

shrids (approved):
LGTM by inspection.
Please merge it once @gaddamas confirms this issue is resolved.

gaddamas (approved):

PR #5149, opened 2020-09-07
Issue 5002: Delete scope api with delete streams flag 

### LINKED ISSUE #5002: The streamManager API should have a way to delete a scope that contains streams in any state
**Problem description**

In practical uses of the streamManager api, we are able to create scopes, streams, list streams, delete streams, seal streams etc.  A given scope can at any time contain any number of streams, in different states.  Sealed, not sealed, some maybe stuck in creating state.  The scope can also contain reader groups, materialized as internal streams that can't normally be directly deleted. 

When the caller needs to delete the scope, it has become difficult to sort out the correct order and logic in which to clean up the scope to prepare for final deletion.  

**Problem location**

streamManager GRPC api.

Provide a flag to the _deleteScope_ call which allows for a "force" or "purge" action which tells Pravega to do the needful to prepare the scope for deletion, then delete it.

Likely the correct default is for that flag to be assumed as false, which is backward compatible and safe.  (fail the deletion of the scope by default if it contains streams)

For reference, this is an example of code that callers needs to do today to delete a scope.  Note that the code below contains a flaw, outlined after the sample:

```
        try {
            List<Stream> userStreams = new ArrayList<>();
            List<Stream> readerGroupStreams = new ArrayList<>();
            Iterator<Stream> it = streamManager.listStreams(instance.getScopeName());
            if (it != null) {
                it.forEachRemaining((stream) -> {
                    if (streamIsUserStream(stream)) {
                        userStreams.add(stream);
                    } else if (streamIsReaderGroup(stream)) {
                        readerGroupStreams.add(stream);
                    } else {
                        LOG.info("Found unknown internal stream {}", stream.getStreamName());
                    }
                });
            }
            // Delete all the ReaderGroup first.
            ReaderGroupManager readerGroupManager = ReaderGroupManager.withScope(instance.getScopeName(), PravegaAutoConfiguration.controllerURI);
            for (Stream s : readerGroupStreams) {
                LOG.info("Deleting reader group {}", readerGroupNameFromStream(s));
                readerGroupManager.deleteReaderGroup(readerGroupNameFromStream(s));
            }
            readerGroupManager.close();
            // Then delete all the user streams.
            for (Stream s : userStreams) {
                LOG.info("Deleting stream {}", s.getStreamName());
                streamManager.sealStream(instance.getScopeName(), s.getStreamName());
                streamManager.deleteStream(instance.getScopeName(), s.getStreamName());
            }
            LOG.info("Deleting scope {}", instance.getScopeName());
            streamManager.deleteScope(instance.getScopeName());

```

The flaw in the code above is if there is a stream that was stuck in creating state somehow, it cannot be sealed.  So the _sealStream_ call throws an exception, does not get to be deleted and fails the overall _deleteScope_ operation.

It's just one nuance that the caller needs to know, an example of the type of internal complexity that would probably be best encapsulated in _deleteScope_ itself

shiveshr:

**Change log description**  
Implements a new api in StreamManager for deleting a scope by internally listing all streams and attempting to delete them. 

**Purpose of the change**  
Fixes #5002 

**What the code does**  
Adds a new Api in stream manager that lists all streams in scope and iteratively seals and deletes each stream. 

Also adds a retry block around submission of seal stream request. Seal stream can fail if the stream is non active -- i.e. scaling or truncating or committing transaction for instance. Such errors are transient and whenever those operations complete, stream can be marked for sealing and then sealed. So we have added a retry in streammetadata tasks to handle such cases. 

**How to verify it**  
Unit test added.


client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
```diff
@@ -141,12 +141,27 @@ public boolean checkStreamExists(String scopeName, String streamName) {
...
+    public boolean deleteScope(String scopeName, boolean deleteStreams) {
         NameUtils.validateUserScopeName(scopeName);
         log.info("Deleting scope: {}", scopeName);
-        return  Futures.getThrowingException(controller.deleteScope(scopeName));
+        
+        if (deleteStreams) {
+            Iterator<Stream> iterator = listStreams(scopeName);
+            while (iterator.hasNext()) {
+                Stream stream = iterator.next();
+                Futures.getThrowingException(controller.sealStream(stream.getScope(), stream.getStreamName()));
```
shrids, line 158: If a "seal" or "deleteStream" operation fails for a single stream should'nt we continue deleting the other streams?

shrids (changes requested):
LGTM apart from one question.

client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
shiveshr: Well, even if we opportunistically delete other streams, the scope can not be deleted if a single stream's seal and delete failed. So we should throw the error back to the user so that they can decide on how to handle it. 
a subsequent attempt to delete the scope using this api will iterate over all available streams and attemp to delete them.

shrids:
The changes LGTM, please find the build failure on the tests due to a Deprecation warning...

shrids (approved):
LGTM.

PR #5150, opened 2020-09-07
Issue 5142: Retry Create stream until it succeeds.

### LINKED ISSUE #5142: In create stream workflow, retry it until it succeeds
**Problem description**
Once we have accepted a create stream request, it should be retried until it completes. Presently we fail the request after few retries which leaves the stream in "creating" state. This makes it impossible to seal and delete such streams until someone requests a create stream again. Instead, by ensuring that once we have started the create, we should retry and create the stream such that it can eventually be sealed and deleted. 

**Problem location**
StreamMetadataTasks

Retry create stream indefinitely

shiveshr:
**Change log description**  
Adds a retry to retry on all retryable failures once create stream is accepted.  

**Purpose of the change**  
Fixes #5142

**What the code does**  
Adds a retry block which retries indefinitely if the create stream is started. This ensures that a stream whose creation is started will be retried until it completes so that we dont leave streams in Creating state. 

**How to verify it**  
All existing tests should pass


controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -738,14 +738,13 @@ public void setRequestEventWriter(EventStreamWriter<ControllerEvent> requestEven
...
-                                                }));
-                                    }, executor)
+                                                }))
                                             .thenApply(z -> status);
                                 });
                     } else {
                         return CompletableFuture.completedFuture(status);
                     }
-                }, executor)
+                }), RetryHelper.RETRYABLE_PREDICATE.or(e -> Exceptions.unwrap(e) instanceof RetriesExhaustedException), Integer.MAX_VALUE, executor)
```
shrids, line 747: Shivesh, I did not understand why we would need Integer.MAX_VALUE retries. Wondering if there a better way instead of retrying it infinitely. Do share you thoughts on this.

shiveshr: so we follow this pattern elsewhere too when we intend to add indefinite retries. Integer.Max = 2 billion.. lets say we retry once every second then this is = ~60+ years.. Which is practically indefinite retry.

shrids: My question was more on the lines of a "fail fast" principle.  The approach in the PR would cause the controller to retry this operation multiple times without us knowing why it failed for so long. Perhaps there is really something wrong in the deployment and it will not evident to the users of Pravega...

shiveshr: the users will fail fast -- both rest and grpc apis have timeouts. So their calls would timeout while controller would continute to retry since it had accepted the create request and started work on it.

PR #5156, opened 2020-09-08
Issue 5155: (SegmentStore) Enabling copy-on-read for all Segment Reads

### LINKED ISSUE #5155: Enable Copy-on-Read for all Segment Reads
**Problem description**
1. An upstream API wants to read offsets O1-O2 from a segment.
    - This can be the PravegaRequestProcessor wanting to read from the segment (due to the Client requesting it) or it could be a Table Segment reader wanting to read a Table Entry
2. We make sure O1-O2 is in the cache (we may load it, whatever). This will occupy 1 or more 4KB cache blocks.
3. We return a view of those blocks (no copying...) to the upstream API
4. The upstream API will eventually copy it onto the wire to send to the requesting client.

If between steps 3 and 4, the cache blocks that hold O1-O2 are evicted and replaced with some other data, then the data copied  as part of step 4 will be corrupted.

Issue #5119 with PR #5120 fixed this for Table Segment Compaction, but there are other paths that haven't been fixed.

**Problem location**
Segment Store.

Enable copy-on-read for:
- Segment reading
- Table Bucket Reader (i.e., when reading Keys or Entries from a Table Segment).

andreipaduroiu:
**Change log description**  
- Enabling copy-on-read for all Segment Reads from the client.
- Enabling copy-on-read for all Segment Reads as a result of Table Segment Entry reads (TableBucketReader calls).

**Purpose of the change**  
Fixes #5155.

**What the code does**  
- Please see #5155 for a description of the bug being resolved that includes the scenario that can lead to it.
- PR #4841 has reduced the number of memory copies that were performed on the read path from a segment, however it has eliminated one too many such copies. 
- #5119 and #5120 have introduced a mechanism that tells the Read Index to return a pointer to a copy of the data in the cache instead of a direct view of the cache data. This would prevent the scenario described in #5155 from happening (i.e., a subsequent eviction of those blocks and reallocation may cause it to read the wrong data), however it has not been applied for all read paths.
- This PR enables this functionality for all Segment Reads. This includes:
    - Client reads
    - Table Segment Key/Entry reads.
- This ensures that the data returned is a single, immutable buffer that will not be modified by any background process.

**How to verify it**  
Unit tests updated to verify the flag is being set.
The actual scenario described in #5155 cannot be reliably tested directly via unit tests, but #5120  did add a new unit test to verify that the copy-on-read works as expected.

### DIFF
--- a/segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/TableCompactorTests.java
+++ b/segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/TableCompactorTests.java
@@ -273,9 +273,6 @@ public class TableCompactorTests extends ThreadPooledTestSuite {
         // In the real world, the IndexWriter will readjust this number as appropriate when reindexing these values.
         Assert.assertEquals("Expecting TOTAL_ENTRY_COUNT to be 0 after a full compaction.",
                 0, context.indexWriter.getTotalEntryCount(context.segmentMetadata));
-
-        Assert.assertEquals("Unexpected number of candidate reads from the segment with copy-on-read enabled.",
-                expectedCopyOnReadCount, context.segment.getCopyOnReadCount());
     }
 
     /**
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
@@ -224,6 +224,7 @@ public class PravegaRequestProcessor extends FailingRequestProcessor implements 
      * an appropriate message is sent back over the connection.
      */
     private void handleReadResult(ReadSegment request, ReadResult result) {
+        result.setCopyOnRead(true); // Get a copy of any data from the cache to avoid losing it due to cache eviction.
         String segment = request.getSegment();
         ArrayList<BufferView> cachedEntries = new ArrayList<>();
         ReadResultEntry nonCachedEntry = collectCachedEntries(request.getOffset(), result, cachedEntries);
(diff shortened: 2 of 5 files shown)


tkaitchuck (approved):
Could this be done the other way, and make the side wishing to free make the copy?

andreipaduroiu:
Unfortunately the copy has to be made inside the Read Index while holding the lock. As soon as the lock is released, the cache eviction is free to happen (if needed) so the returned BufferView could already be dead on arrival.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/TableCompactorTests.java
```diff
@@ -273,9 +273,6 @@ private void testCompaction(int readLength) {
         // In the real world, the IndexWriter will readjust this number as appropriate when reindexing these values.
         Assert.assertEquals("Expecting TOTAL_ENTRY_COUNT to be 0 after a full compaction.",
                 0, context.indexWriter.getTotalEntryCount(context.segmentMetadata));
-
-        Assert.assertEquals("Unexpected number of candidate reads from the segment with copy-on-read enabled.",
-                expectedCopyOnReadCount, context.segment.getCopyOnReadCount());
```
eolivelli, line 278: probably you can drop `expectedCopyOnReadCount` variable

eolivelli (changes requested):
looks good in general.
I left one little comment

RaulGracia (approved):

segmentstore/server/src/test/java/io/pravega/segmentstore/server/tables/TableCompactorTests.java
RaulGracia: Thanks @eolivelli, we couldn't address this comment as we received instructions to cut the 0.8 RC branch in the next hours (the PR author is not likely to be able to address that within this time frame). We will address your comment anyway in 0.9, if this works for you.

eolivelli: @RaulGracia 
Good for me. it is not a big deal to leave or drop such useless variable.

It is also weird that spotbugs did not catch the dead local store.
I thought we were running spotbugs even against tests.

PR #5161, opened 2020-09-09
Issue 5801: When routing randomly stick with a single segment for a while.

### LINKED ISSUE #5801: Optimize batching when not using routing keys.
When no routing key is provided, events are routed randomly. 
For streams with large numbers of segments this can result in small batch sizes on the wire.

Because it doesn't actually matter which segment an event goes to we ought to be able to select the same segment for a small time or size before switching to the next. If this is done well, it could provide better on the wire batching with no additional hotspots on the server.

tkaitchuck:

**Change log description**  
* To preform random segment selection for null routing key workloads stick with one segment for each 256kb and then rotate to another segment.

**Purpose of the change**  
This is follow up to: https://github.com/pravega/pravega/pull/4909 . This fixes #5801

**What the code does**  
Uses the number of bytes written as a routing key.

**How to verify it**  
Performance tests are needed.


client/src/main/java/io/pravega/client/stream/impl/SegmentSelector.java
```diff
@@ -64,24 +74,26 @@
...
     @Synchronized
-    public Segment getSegmentForEvent(String routingKey) {
+    public Segment getSegmentForEvent(String routingKey, int length) {
         if (currentSegments == null) {
             return null;
         }
         if (routingKey == null) {
-            return currentSegments.getSegmentForKey(random.nextDouble());
+            bytesWritten += length;
+            return currentSegments.getSegmentForKey(randomRoutingKeyHasher.hashToRange(bytesWritten / BYTES_PER_SEGMENT));
```
shrids, line 96: This would return the same routing key for every 1GB of data, right ? (2^12 + BYTES_PER_SEGMENT)
(first 1GB of data with random routing key will be written to segment corresponding to routing key as 0.0)

tkaitchuck: I am not sure what you are getting at. bytesWritten is a long, so it should only have the same value every 2^64 bytes written. By dividing we are throwing out the low order bits so that the value hashed only changes every 256kb.

shrids: oops i misread the code. Yeah the routing key changes every `262144` bytes.

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -121,7 +121,8 @@
         CompletableFuture<Void> ackFuture = new CompletableFuture<Void>();
         synchronized (writeFlushLock) {
             synchronized (writeSealLock) {
-                SegmentOutputStream segmentWriter = getSegmentWriter(routingKey);
+                int size = data.remaining();
```
shrids, line 124: We could move this outside the synchronized block right ?

shrids (approved):
LGTM apart from one comment.

client/src/main/java/io/pravega/client/stream/impl/SegmentSelector.java
```diff
@@ -47,6 +48,15 @@
 @RequiredArgsConstructor
 public class SegmentSelector {
 
+    /**
+     * When writing without a routing key, keep writing this many bytes to the same segment before selecting a new one.
+     */
+    private static final long BYTES_PER_SEGMENT = 256 * 1024;
```
claudiofahey, line 54: Can we make this a runtime parameter from a Java property? This will allow users to easily optimize this.

RaulGracia:
I have validated this change on AWS: when no routing keys are used, it leads to 3x-4x higher latency (p95) compared to when routing keys are used (also compared to Pravega 0.9). I think that is better to keep the current behavior of the client, as it achieves low latency both with and without routing keys. I think that the potential per-writer gain in maximum throughput of this change does not compensate the important hit in performance we have observed.

PR #5175, opened 2020-09-11
Issue 5166: In deleteScope(deleteStreams) method, handle seal stream failures

### LINKED ISSUE #5166: In deleteScope(deleteStreams) method, attempt to delete the stream even if seal fails.
**Problem description**
In the new delete scope method which attempts to list and delete the streams, we first attempt to seal followed by deleting the stream. But a seal stream can fail if a stream was partially created and is left in creating state. So we should attempt to delete the stream even if the seal fails. If the stream was in creating or sealed state, delete would succeed, otherwise it would fail, whereby failing deletescope method.  

**Problem location**
StreamManagerImpl

Fix the issue. Throw appropriate exception.

shiveshr:

**Change log description**  
Handle seal stream failures in deleteScope(deleteStreams) method which would attempt to delete the stream even if seal fails. Such attempts may fail unless the stream was stuck in creating state. 

**Purpose of the change**  
Fixes #5166 

**What the code does**  
In delete scope we first attempt to seal the stream and if it fails, we handle the exception and directly proceed to delete the stream. If delete stream also fails we throw delete scope failed exception

**How to verify it**  
Unit test added

### DIFF
--- a/client/src/main/java/io/pravega/client/stream/DeleteScopeFailedException.java
+++ b/client/src/main/java/io/pravega/client/stream/DeleteScopeFailedException.java
@@ -12,7 +12,7 @@ package io.pravega.client.stream;
 /**
  * Delete scope failed, typically because we failed to seal and delete the streams. 
  */
-public class DeleteScopeFailedException extends RuntimeException {
+public class DeleteScopeFailedException extends Exception {
 
     private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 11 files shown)


client/src/main/java/io/pravega/client/stream/DeleteScopeFailedException.java
```diff
@@ -0,0 +1,30 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.client.stream;
+
+/**
+ * Delete scope failed, typically because we failed to seal and delete the streams. 
+ */
+public class DeleteScopeFailedException extends RuntimeException {
```
shrids, line 15: A checked exception would make this more clear to the user right?

shrids (commented):
LGTM ...

shrids (approved):
+1

PR #5179, opened 2020-09-13
Issue 5122: Improving Windows Compatibility

co-jo:
I mistaken messed up the other branch, which ended up automatically closing the other PR (https://github.com/pravega/pravega)/pull/5123.

**Change log description**  

1. Added some documentation recognizing potentially slow unit tests (those involved in touching a large amount of files) and a possible fix.
2. Target the proper platform specific temp directory within `S3ProxyImpl`.
3. Surround a file clean procedure with a try/catch block to absorb the exception.

**Purpose of the change**  
Makes progress towards: #5122 

**What the code does**  

2. There are two benefits for this change. `S3ProxyImpl`  will save files relative to the root of the drive. Firstly for aesthetic reasons, we should avoid dumping files at the top level of someone's `C:` drive, and place them into the proper `Temp` folder. More importantly, because it is placed in a `tmp` folder at the root, Windows will place extra permissions on this folder, which can prevent the creation (and deletion) of files under it, causing the test to fail. 
3. This procedure often fails because Windows (I believe unlike Unix systems) automatically places a FileLock on files, which we later attempt to delete. 

It does not look like we do any extra work to try and recover from `IOException`s here, so it may be appropriate simply to log the exception instead of passing it along. One issue with this is that these temporary files may accumulate over time (Well, this happens anyways..).

**How to verify it**  

Reference: https://github.com/pravega/pravega/pull/5123


common/src/main/java/io/pravega/common/io/filesystem/FileOperations.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.commons.io.FileUtils;
+
+import java.io.File;
+import java.util.HashMap;
+
+@Slf4j
+public class FileOperations {
+
+    public static boolean cleanupDirectories(HashMap<?, File> toDelete) {
+        for (File dir : toDelete.values()) {
```
co-jo, line 22: This change is due to the unfortunate reality of pleasing code coverage. A way was required to (platform independently) test for a failure scenario. The previous solution had several problems which made this difficult:

1. The visibility of `cleanupDirectories` and other `BookKeeperServiceRunner` fields made it hard to unit test `cleanupDirectories` at all in the first place.
2. Unit testing `cleanupDirectories` within any other the `BookKeeper*` tests seems very forced and unnatural.
3. Forcing an `IOException` to be thrown (across platforms) is also quite difficult -- it is much easier to force an `IllegalArgumentException`.
4. We still need some way to determine whether or not an exception was thrown. Previously the exception was only caught and logged. 

At the expensive of creating two small classes, this avoids the above problems.

common/src/main/java/io/pravega/common/io/filesystem/FileOperations.java
```diff
@@ -0,0 +1,33 @@
...
+            log.info("Cleaning up " + dir);
+            if (!FileUtils.deleteQuietly(dir)) {
+                log.info("Failed deleting directory: {}", dir);
+                return false;
+            }
+        }
+        toDelete.clear();
```
andreipaduroiu, line 29: Why do you do this? This will modify the HashMap that you are being given, so you caller may have a nasty surprise after invoking this method? 

I don't think this is needed. If the caller wants to clear, they can do it after invoking this method.

andreipaduroiu (changes requested):
LGTM. One minor issue, though, but we should be good after that.

common/src/main/java/io/pravega/common/io/filesystem/FileOperations.java
co-jo: > 
> 
> Why do you do this? This will modify the HashMap that you are being given, so you caller may have a nasty surprise after invoking this method?
> 
> I don't think this is needed. If the caller wants to clear, they can do it after invoking this method.

This was part of the original method so I just kept it. Will remove.

andreipaduroiu (approved):

PR #5180, opened 2020-09-14
Issue 5169: Data Recovery - Copy the core attributes from old segment metadata to new segment metadata

### LINKED ISSUE #5169: Data Recovery - Copy the core attributes from old segment metadata to new segment metadata
**Problem description**
Right now, in DR integration testing, the old segment metadata is deleted from LTS, and a new segment metadata is created using the other segments in the LTS. While registering the segments to new segment metadata, we only copy segment's length, sealed status and name. We need to copy all other attributes from the old segment metadata to new segment metadata. 

**Problem location**
`ContainerRecoveryUtils.java`

N/A

ManishKumarKeshri:
**Change log description**  
The `registerSegment` method is changed to included core attributes as one of the parameters while registering a segment which exists in LTS, but not in Durable data log.
The tests around the data recovery have also been updated with this.

**Purpose of the change**  
Fixes #5169 

**What the code does**  
Instead of using only sealed status, length and name for registering a segment, now we also include core attributes. 

**How to verify it**  
Build shall pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
@@ -10,7 +10,6 @@
 package io.pravega.segmentstore.server.containers;
 
 import com.google.common.base.Preconditions;
-import io.pravega.common.Exceptions;
 import io.pravega.common.concurrent.Futures;
 import io.pravega.segmentstore.contracts.SegmentProperties;
 import io.pravega.segmentstore.contracts.StreamSegmentNotExistsException;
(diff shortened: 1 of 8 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -114,6 +117,32 @@ public static void recoverAllSegments(Storage storage, Map<Integer, DebugStreamS
...
+                .exceptionally(e -> {
+                    if (Exceptions.unwrap(e) instanceof StreamSegmentNotExistsException) {
+                        log.info("Segment '{}' doesn't exist.", metadataSegmentName);
+                        return null;
+                    }
+                    log.error("Error occurred while copying the segment.");
+                    return null;
+                }).thenAcceptAsync(x -> container.copySegment(storage, attributeSegmentName, "NewAttributeSegment" + timeStamp + "_" +
+                        container.getId(), executorService)).exceptionally(e -> {
+                    if (Exceptions.unwrap(e) instanceof StreamSegmentNotExistsException) {
```
sachin-j-joshi, line 137: Please indent this block , so that it is obvious that if statement is inside exceptionally.

PR #5182, opened 2020-09-14
Issue 5181: Fix Null Pointer Exception in the log message in StreamMetadataTasks

### LINKED ISSUE #5181: Null Pointer Exception in StreamMetadataTasks
**Problem description**
```
Caused by: java.lang.NullPointerException: null
        at io.pravega.controller.task.Stream.StreamMetadataTasks.truncate(StreamMetadataTasks.java:333)
        at io.pravega.controller.task.Stream.StreamMetadataTasks.lambda$null$11(StreamMetadataTasks.java:300)
```

This is caused by new log message that was introduced recently:
```
        log.info("Found truncation record for stream {}/{} newRecord time/size: {}/{}", scope, stream,
                                                                                newRecord.getRecordingTime(), newRecord.getRecordingSize());
```
The record to use here should be `truncationRecord` instead of `newRecord`.

**Problem location**
StreamMetadataTasks.java

Fix the log

shiveshr:

**Change log description**  
Fixes NPE in StreamMetadataTasks.retention method where we may truncate at an older streamcut despite not producing a "new stream cut".

**Purpose of the change**  
Fixes #5181 

**What the code does**  
NPE is caused by a new log message that was introduced recently which logs `newRecord` instead of `truncationRecord`. The new record represents the new streamcut which is generated periodically. It is possible that we do not generate a new streamcut in each cycle if the previous cycle was cut short for some reason and a streamcut was generated. The truncation record represents an older streamcut at which we want to truncate. 
The intent of the log message is to log the identified truncation streamcut. Instead it ends up logging new streamcut which can be null in some cases. 

**How to verify it**  
All tests should pass and there should be no NPE.


pbelgundi (approved):
LGTM

PR #5191, opened 2020-09-15
issue 4162: test pod side changes for tls enabled

ravibeta:
**Change log description**  
The pravega cluster is setup properly with tls enabled option and the controller and segmentstore are able to communicate over tls but the test pod creation fails because the volume mount with the tls secret was not found. This is now fixed along with test pod side changes to allow the requests to go through. This was shown to work in longevity tool but the system tests had a few specific issues: 1) the test pod did not get the tls enabled parameter from the build automation 2) when it did, the pod did not find the tls secret  3) the test pod image version did not support tls.

**Purpose of the change**  
Fixes: #4162 

**What the code does**  
The test pod had a volume mount with an incorrect name. The name used now is the same as the one for the volume builder. The test pod is now using the jre that has the required java libraries.  Sample pod description attached

**How to verify it**  
The system tests are run with this change in build 980 as can be seen with the mounts section in:
[MultiControllerTestPodDescription1.txt](https://github.com/pravega/pravega/files/5232750/MultiControllerTestPodDescription1.txt)
The system test automated run 3185 was run with code similar to this change in issue-4162-propagate-parameters-3 branch except that it hardcoded the tls and auth enabled parameter to true and the logs from the test pod there show that the modified secret now establishes a successful test run from the pod to the store.
[test.log](https://github.com/pravega/pravega/files/5235601/test.log)


build.gradle
```diff
@@ -842,6 +843,7 @@ project('test:system') {
     }
 
     task copyTestsToK8sCluster(type: Exec) {
+        tasks.withType(Exec) { environment "tlsEnabled", System.getProperty("tlsEnabled", "false") }
```
shrids, line 846: this property is set at line 854 again. Is declaration seems redundant?

andreipaduroiu:
Closing this for now. We may reopen if we decide to continue this work.

PR #5193, opened 2020-09-15
Issue 5192: Copying a segment

### LINKED ISSUE #5192: Renaming a segment in FILESYSTEM storage
**Problem description**
For data recovery, we need to save the state of current metadata segment, and recreate a new one. The state of the current metadata segment is saved by naming the segment in storage. 

**Problem location**
`FileSystemStorage`

Implemented a method only under `FileSystemStorage` which will rename any given segment.

ManishKumarKeshri:
**Change log description**  
API to copy the contents of a segment to another segment.

**Purpose of the change**  
Fixes #5192 

**What the code does**  
A method which takes in the names of the `source` & `target` segment as parameters, creates the `target` segment using the `storage` API and then copies the contents of the `source` segment to the `target` segment. Copy is done by filling the buffer with bytes to be read and then writing it. Buffer size is 8MB.

**How to verify it**  
Build shall pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
@@ -8,10 +8,8 @@
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
 package io.pravega.segmentstore.server;
-import io.pravega.segmentstore.storage.Storage;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutorService;
 
 /**
  * Defines debug segment container for stream segments.
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
@@ -29,5 +29,12 @@ public interface DebugSegmentContainer extends SegmentContainer {
      */
     CompletableFuture<Void> registerSegment(String streamSegmentName, long length, boolean isSealed);
 
-    void copySegment(Storage storage, String sourceSegment, String targetSegment) throws ExecutionException, InterruptedException;
+    /**
+     * Creates a segment with target segment name and copies the contents of the source segment to the target segment.
+     * @param storage                   A storage instance to create the segment.
+     * @param sourceSegment             The name of the source segment to copy the contents from.
+     * @param targetSegment             The name of the segment to write the contents to.
+     * @throws Exception                In case of an exception occurred while execution.
+     */
+    void copySegment(Storage storage, String sourceSegment, String targetSegment) throws Exception;
 }
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
@@ -11,6 +11,7 @@ package io.pravega.segmentstore.server;
 import io.pravega.segmentstore.storage.Storage;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
 
 /**
  * Defines debug segment container for stream segments.
@@ -33,7 +34,9 @@ public interface DebugSegmentContainer extends SegmentContainer {
      * @param storage                   A storage instance to create the segment.
      * @param sourceSegment             The name of the source segment to copy the contents from.
      * @param targetSegment             The name of the segment to write the contents to.
-     * @throws Exception                In case of an exception occurred while execution.
+     * @param executor                  An Executor for async operations.
+     * @return                          A CompletableFuture that, when completed normally, will indicate the operation
+     * completed. If the operation failed, the future will be failed with the causing exception.
      */
-    void copySegment(Storage storage, String sourceSegment, String targetSegment) throws Exception;
+    CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor);
 }
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
@@ -30,11 +30,11 @@ public interface DebugSegmentContainer extends SegmentContainer {
     CompletableFuture<Void> registerSegment(String streamSegmentName, long length, boolean isSealed);
 
     /**
-     * Creates a segment with target segment name and copies the contents of the source segment to the target segment.
+     * Creates a target segment with the given name and copies the contents of the source segment to the target segment.
      * @param storage                   A storage instance to create the segment.
      * @param sourceSegment             The name of the source segment to copy the contents from.
      * @param targetSegment             The name of the segment to write the contents to.
-     * @param executor                  An Executor for async operations.
+     * @param executor                  A thread pool for execution.
      * @return                          A CompletableFuture that, when completed normally, will indicate the operation
      * completed. If the operation failed, the future will be failed with the causing exception.
      */
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
@@ -11,7 +11,6 @@ package io.pravega.segmentstore.server;
 import io.pravega.segmentstore.storage.Storage;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 
 /**
  * Defines debug segment container for stream segments.
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
@@ -8,7 +8,10 @@
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
 package io.pravega.segmentstore.server;
+import io.pravega.segmentstore.storage.Storage;
+
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 
 /**
  * Defines debug segment container for stream segments.
@@ -25,4 +28,6 @@ public interface DebugSegmentContainer extends SegmentContainer {
      * completed. If the operation failed, the future will be failed with the causing exception.
      */
     CompletableFuture<Void> registerSegment(String streamSegmentName, long length, boolean isSealed);
+
+    void copySegment(Storage storage, String sourceSegment, String targetSegment) throws ExecutionException, InterruptedException;
 }
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
@@ -28,15 +28,4 @@ public interface DebugSegmentContainer extends SegmentContainer {
      * completed. If the operation failed, the future will be failed with the causing exception.
      */
     CompletableFuture<Void> registerSegment(String streamSegmentName, long length, boolean isSealed);
-
-    /**
-     * Creates a target segment with the given name and copies the contents of the source segment to the target segment.
-     * @param storage                   A storage instance to create the segment.
-     * @param sourceSegment             The name of the source segment to copy the contents from.
-     * @param targetSegment             The name of the segment to write the contents to.
-     * @param executor                  A thread pool for execution.
-     * @return                          A CompletableFuture that, when completed normally, will indicate the operation
-     * completed. If the operation failed, the future will be failed with the causing exception.
-     */
-    CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor);
 }
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import static io.pravega.shared.NameUtils.getMetadataSegmentName;
@@ -43,8 +44,6 @@ import static io.pravega.shared.NameUtils.getMetadataSegmentName;
 public class ContainerRecoveryUtils {
     private static final Duration TIMEOUT = Duration.ofSeconds(30);
     private static final int BUFFER_SIZE = 8 * 1024 * 1024;
-    private static int bytesToRead;
-    private static int offset;
 
     /**
      * Recovers Segments from the given Storage instance. This is done by:
@@ -256,22 +255,24 @@ public class ContainerRecoveryUtils {
      * @return                          A CompletableFuture that, when completed normally, will indicate the operation
      * completed. If the operation failed, the future will be failed with the causing exception.
      */
-    public static CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor) {
+    protected static CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor) {
+        AtomicInteger offset = new AtomicInteger();
+        AtomicInteger bytesToRead = new AtomicInteger();
         return storage.create(targetSegment, TIMEOUT).thenCompose(targetHandle -> {
             return storage.getStreamSegmentInfo(sourceSegment, TIMEOUT).thenCompose(info -> {
                 return storage.openRead(sourceSegment).thenCompose(sourceHandle -> {
-                    offset = 0;
-                    bytesToRead = (int) info.getLength();
+                    offset.set(0);
+                    bytesToRead.set((int) info.getLength());
                     return Futures.loop(
-                            () -> bytesToRead > 0,
+                            () -> bytesToRead.get() > 0,
                             () -> {
-                                byte[] buffer = new byte[Math.min(BUFFER_SIZE, bytesToRead)];
-                                return storage.read(sourceHandle, offset, buffer, 0, buffer.length, TIMEOUT)
+                                byte[] buffer = new byte[Math.min(BUFFER_SIZE, bytesToRead.get())];
+                                return storage.read(sourceHandle, offset.get(), buffer, 0, buffer.length, TIMEOUT)
                                         .thenComposeAsync(size -> {
-                                            bytesToRead -= size;
-                                            return (size > 0) ? storage.write(targetHandle, offset, new
+                                            bytesToRead.addAndGet(-size);
+                                            return (size > 0) ? storage.write(targetHandle, offset.get(), new
                                                     ByteArrayInputStream(buffer, 0, size), size, TIMEOUT).thenAcceptAsync(r -> {
-                                                offset += size;
+                                                offset.addAndGet(size);
                                             }, executor) : null;
                                         }, executor);
                             }, executor);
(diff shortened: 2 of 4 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/rolling/RollingStorage.java
```diff
@@ -442,6 +442,11 @@ public void concat(SegmentHandle targetHandle, long targetOffset, String sourceS
         LoggerHelpers.traceLeave(log, "concat", traceId, target, targetOffset, sourceSegment);
     }
 
+    @Override
+    public void rename(SegmentHandle handle, String newName) throws StreamSegmentException {
```
sachin-j-joshi, line 446: Rolling storage is implemented using concept of chunks and a header file that points to them.

Proper implementation will involve following steps
```

Find name of the header file (see functions in nameutils)
Deserialize the header file
Find out list of chunks
 For each chunk in list { 
          parse chunk name to extract offset, 
          using offset+new segment name generate new chunk name
          rename by calling baseStorage.rename
          update name in the header data structure
}
serialize updated header (so that it points to the recently renamed chunks)
now rename the header itself to new name.
```

- Just calling this.baseStorage.rename won't work because there is no such segment in base segment.
- Just renaming the header won't work because then we are still pointing to the old chunks. When new segment is created with old name, it will try to create a chunk at offset 0 which will be already existing.

sachin-j-joshi (changes requested):
Need to implement rename for Rolling storage

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
```diff
@@ -58,4 +62,20 @@
         ArrayView segmentInfo = MetadataStore.SegmentInfo.recoveredSegment(streamSegmentName, length, isSealed);
         return metadataStore.createSegment(streamSegmentName, segmentInfo, new TimeoutTimer(TIMEOUT));
     }
+
+    @Override
+    public void copySegment(Storage storage, String sourceSegment, String targetSegment)
+            throws Exception {
+        storage.create(targetSegment, TIMEOUT);
```
andreipaduroiu, line 69: This is an async method. And so is `getStreamSegmentInfo`.

Please rewrite this entire method to be aynchronous:
1. `public CompletableFuture<Void> copySegment(...)` (remove the `throws Exception`).
2. Use `CompletableFuture` composition (`thenCompose`, `thenApply`, etc.) to chain your futures together. 
3. Use `Futures.loop`
4. return the resulting `CompletableFuture`.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
```diff
@@ -17,17 +17,21 @@
...
 
+import java.io.ByteArrayInputStream;
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ScheduledExecutorService;
 
 @Slf4j
 public class DebugStreamSegmentContainer extends StreamSegmentContainer implements DebugSegmentContainer {
     private static final Duration TIMEOUT = Duration.ofMinutes(1);
+    private static final int BUFFER_SIZE = 8 * 1024;
```
andreipaduroiu, line 34: This buffer is too small. I suggest 4/8/16MB or so.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
```diff
@@ -58,4 +62,20 @@
...
+    @Override
+    public void copySegment(Storage storage, String sourceSegment, String targetSegment)
+            throws Exception {
+        storage.create(targetSegment, TIMEOUT);
+        val info = storage.getStreamSegmentInfo(sourceSegment, TIMEOUT);
+        int bytesToRead = (int) info.get().getLength();
+        int offset = 0;
+        while (bytesToRead > 0) {
+            byte[] buffer = new byte[Math.min(BUFFER_SIZE, bytesToRead)];
+            int size = storage.read(storage.openRead(sourceSegment).join(), offset, buffer, 0, buffer.length, TIMEOUT).join();
```
andreipaduroiu, line 75: Here, you invoke `openRead` on every loop call. This is highly inefficient. Invoke it once outside of the loop and then reuse the returned handle.

andreipaduroiu: also verify that `size > 0`. If it's 0, then you risk getting into an infinite loop.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
```diff
@@ -58,4 +62,20 @@
...
+            bytesToRead -= size;
+            storage.write(storage.openWrite(targetSegment).join(), offset, new ByteArrayInputStream(buffer, 0, size), size, TIMEOUT).join();
```
andreipaduroiu, line 77: `openWrite` outside of the loop.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
```diff
@@ -25,4 +27,13 @@
...
     CompletableFuture<Void> registerSegment(String streamSegmentName, long length, boolean isSealed);
+
+    /**
+     * Creates a segment with target segment name and copies the contents of the source segment to the target segment.
+     * @param storage                   A storage instance to create the segment.
+     * @param sourceSegment             The name of the source segment to copy the contents from.
+     * @param targetSegment             The name of the segment to write the contents to.
+     * @throws Exception                In case of an exception occurred while execution.
+     */
+    void copySegment(Storage storage, String sourceSegment, String targetSegment) throws Exception;
```
andreipaduroiu, line 38: Have we thought whether it would be better to put `copySegment` into the `Storage` interface? That way we could take advantage of LTS-specific copy features (so that we do not need to read everything and write it back). @sachin-j-joshi what do you think?

andreipaduroiu (changes requested):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/DebugSegmentContainer.java
sachin-j-joshi: Ideally we want rename method implemented in Storage. But that also means implementing  it for RollingStorage with all its corner cases.
Given that the rename functionality is needed only in DR tool and only once for a container metadata segment. I suggested this short term work around of copying that segment.
In near future we'll implement rename in Storage. But for now I think this is good enough to unblock us.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
ManishKumarKeshri: made it to 8MB.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
```diff
@@ -58,4 +66,27 @@
         ArrayView segmentInfo = MetadataStore.SegmentInfo.recoveredSegment(streamSegmentName, length, isSealed);
         return metadataStore.createSegment(streamSegmentName, segmentInfo, new TimeoutTimer(TIMEOUT));
     }
+
+    @Override
+    public CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor) {
+        val targetHandle = storage.create(targetSegment, TIMEOUT).thenCompose(v -> storage.openWrite(targetSegment)).join();
```
sachin-j-joshi, line 72: Need to get rid of these joins as well.
It should look something like below .
Also you don't need to openWrite again after create.
```
return storage.create(targetSegment, TIMEOUT).thenCompose(targetHandle -> {
           return storage.getStreamSegmentInfo(sourceSegment, TIMEOUT).thenCompose(info -> {
               return storage.openRead(sourceSegment).thenCompose(sourceHandle  ->  {
                      offset = 0;
                      return Futures.loop( .....
              }
        })
}})
```

sachin-j-joshi (changes requested):

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainerTests.java
```diff
@@ -247,6 +251,85 @@ public void testEndToEnd() throws Exception {
...
+        log.info("Created a storage instance");
+
+        String sourceSegmentName = "segment-" + RANDOM.nextInt();
+        String targetSegmentName = "segment-" + RANDOM.nextInt();
+
+        // Create source segment
+        s.create(sourceSegmentName, TIMEOUT).join();
+        val handle = s.openWrite(sourceSegmentName).join();
+
+        // do some writing
```
sachin-j-joshi, line 276: Not sure why we need this in for loop.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainerTests.java
```diff
@@ -247,6 +251,85 @@ public void testEndToEnd() throws Exception {
...
+        for (offset = 0; offset < expectedData.length / 2; offset++) {
+            int length = (int) (expectedData.length - 2 * offset);
+            byte[] readBuffer = new byte[length];
+            int bytesRead = s.read(readHandle, offset, readBuffer, 0, readBuffer.length, TIMEOUT).join();
+            Assert.assertEquals(String.format("Unexpected number of bytes read from offset %d.", offset),
+                    length, bytesRead);
+            AssertExtensions.assertArrayEquals(String.format("Unexpected read result from offset %d.", offset),
+                    expectedData, (int) offset, readBuffer, 0, bytesRead);
+        }
+    }
```
sachin-j-joshi, line 321: Not sure why this needs to be in loop

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainer.java
ManishKumarKeshri: Done.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainerTests.java
ManishKumarKeshri: Removed loop. Writing 10 MB data at once.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainerTests.java
ManishKumarKeshri: Removed data. Wrote 10 MB data, and reading all of it at once and verifying.

sachin-j-joshi (changes requested):
LGTM

sachin-j-joshi (changes requested):
You'll have to create a table segment outside of copySegment using table extension createsegment.
And then copy the contents.

So instead of create in copySegment need to use openWrite.

ManishKumarKeshri:
Moved the `copySegment` method from `DebugSegmentContainer.java` to `ContainerRecoveryUtils.java`, so that a `DebugSegmentContainer` instance is not required for executing the method. 
From the integration test, it looks like creating the segment as `tableSegment` is not required. This PR is only for copying the segment contents. Other PR is coming up in which iterating over the copied segment is required. If necessary, We can have the segment created as `TableSegment` there.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -242,4 +245,39 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     * @return                          A CompletableFuture that, when completed normally, will indicate the operation
+     * completed. If the operation failed, the future will be failed with the causing exception.
+     */
+    protected static CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor) {
+        AtomicInteger offset = new AtomicInteger();
+        AtomicInteger bytesToRead = new AtomicInteger();
+        return storage.create(targetSegment, TIMEOUT).thenCompose(targetHandle -> {
+            return storage.getStreamSegmentInfo(sourceSegment, TIMEOUT).thenCompose(info -> {
+                return storage.openRead(sourceSegment).thenCompose(sourceHandle -> {
+                    offset.set(0);
```
andreipaduroiu, line 264: This can  be done in 259 above, when you initialize the variable.

Actually you can move the declaration of both `offset` and `bytesToRead` here.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -242,4 +245,39 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+                    bytesToRead.set((int) info.getLength());
+                    return Futures.loop(
+                            () -> bytesToRead.get() > 0,
+                            () -> {
+                                byte[] buffer = new byte[Math.min(BUFFER_SIZE, bytesToRead.get())];
```
andreipaduroiu, line 269: Instead of allocating this buffer every time, do it outside of the loop. Pre-allocate one of `BUFFER_SIZE` length.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -242,4 +245,39 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+                                return storage.read(sourceHandle, offset.get(), buffer, 0, buffer.length, TIMEOUT)
+                                        .thenComposeAsync(size -> {
+                                            bytesToRead.addAndGet(-size);
```
andreipaduroiu, line 272: You can move this in the `thenAccept` below.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -242,4 +245,39 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+                                            return (size > 0) ? storage.write(targetHandle, offset.get(), new
+                                                    ByteArrayInputStream(buffer, 0, size), size, TIMEOUT).thenAcceptAsync(r -> {
```
andreipaduroiu, line 274: For readability, please be consistent in how you format and indent these callbacks. For best readability, consider moving the `.thenXYZ` methods on a new line. Then you won't have these weird line wrap-arounds that break the read flow.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -242,4 +245,39 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     * @param sourceSegment             The name of the source segment to copy the contents from.
+     * @param targetSegment             The name of the segment to write the contents to.
+     * @param executor                  A thread pool for execution.
+     * @return                          A CompletableFuture that, when completed normally, will indicate the operation
+     * completed. If the operation failed, the future will be failed with the causing exception.
+     */
+    protected static CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor) {
+        AtomicInteger offset = new AtomicInteger();
+        AtomicInteger bytesToRead = new AtomicInteger();
+        return storage.create(targetSegment, TIMEOUT).thenCompose(targetHandle -> {
```
andreipaduroiu, line 261: What happens when the segment copy is interrupted in the middle (by whatever reason, including a crash)?

From what I can tell, the target segment will be left in an inconsistent state. It will either be empty or have some partial data in it. 

Should we change this method to either:
1. Delete and recreate the segment (i.e., overwrite)
2. Throw (like now)
3. Resume from where it left off.

If think #3 is off the table given that it may already contain some other, unrelated data. I would choose between 1 and 2. Does the upstream code handle an exception gracefully? Can it recover from it?

andreipaduroiu (changes requested):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: This method will be used in an upcoming PR(#5180). Since, delete & re-copy can be handled there. 
I will be working on renaming the segments, then this method will not be used.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

andreipaduroiu (approved):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -242,4 +245,37 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     * @param sourceSegment             The name of the source segment to copy the contents from.
+     * @param targetSegment             The name of the segment to write the contents to.
+     * @param executor                  A thread pool for execution.
+     * @return                          A CompletableFuture that, when completed normally, will indicate the operation
+     * completed. If the operation failed, the future will be failed with the causing exception.
+     */
+    protected static CompletableFuture<Void> copySegment(Storage storage, String sourceSegment, String targetSegment, ExecutorService executor) {
+        byte[] buffer = new byte[BUFFER_SIZE];
+        return storage.create(targetSegment, TIMEOUT).thenCompose(targetHandle -> {
+            return storage.getStreamSegmentInfo(sourceSegment, TIMEOUT).thenCompose(info -> {
```
sachin-j-joshi, line 261: Suggest that you use composeAsync here and previous line and pass in executor.

sachin-j-joshi (changes requested):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

sachin-j-joshi (approved):
LGTM

PR #5198, opened 2020-09-17
Issue 5197: Limit number of transactions in a commit batch

### LINKED ISSUE #5197: Limit number of transactions in a batch while processing commits
**Problem description**
Add an upper limit on number of transactions that will be batched together and committed. 

**Problem location**
Commit Transaction workflow

Add an upper limit on number of transactions to be processed in a batch.

shiveshr:

**Change log description**  
Limit number of transactions in a commit batch.

**Purpose of the change**  
Fixes #5197 

**What the code does**  
We add a new "limit" paramter to store.startCommitTransaction method which only starts a batch equal to the size of commitBatch. 

**How to verify it**  
Unit test added


controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1673,7 +1677,11 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
             txnIds.add(x.getValue());
         });
 
-        return getTransactionRecords(epoch, txnIds).thenAccept(txns -> {
+        AtomicInteger from = new AtomicInteger(0);
+        Function<Integer, Integer> tillFunc = f -> f + limit < txnIds.size() ? f + limit : txnIds.size();
```
shrids, line 1681: Math.min () can be used here ?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1683,6 +1691,9 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
                         if (txnRecord.getCommitOrder() == order) {
                             // if entry matches record's position then include it
                             transactionsMap.put(txnId, txnRecord);
+                            if (transactionsMap.size() >= limit) {
```
shrids, line 1694: do we require this check given that the txnIds are not greater than the passed limit ?

shrids:
System test run for master @ `job/test-pravega-continuous/3186/`

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: transactionsMap is actually accumulating filtered txns from the txnIds across all iterations and the limit s really to be imposed on the number of entries we put in this map as that will be sent back to the caller to use as the batch of transactions

shrids (approved):
+1 LGTM on code inspection. 
This code should be merged post the system test run.

Issue #5110, opened 2020-08-26
Controller API for updating truncation StreamCuts on Controller.


pbelgundi:
**Problem description**
Add a new Controller API that will be used by Pravega Client to update truncation Stream-Cuts on Controller.
Both user generated and automated checkpoints can be updated to Controller using this Controller API.
These stream cuts will be used by Controller to arrive at the common truncation Stream-Cut. on a Stream with Consumption based Retention policy.

**Problem location**
Controller

**Suggestions for an improvement**

> boolean updateTruncationStreamCut(String RGName, Map<ScopedStreamName, StreamCut sc> streamCutPerStream)

PR #5204, opened 2020-09-18
Issue:5109 Add and Remove Subscriber API on Controller

### LINKED ISSUE #5109: Controller APIs for Adding/Removing Subscribers to a Stream
**Problem description**
For Streams that support Consumption based retention, Controller needs to know which reader groups have "subscribed" to the Stream, so it can arrive at a common "truncation Stream-Cut" based on truncation stream-cuts submitted by each subscribing RG.
The purpose of this change is to add new APIs to Controller to store the list of subscribing reader groups for a Stream or remove a Reader Group from the list of Subscribers.

**Problem location**
Controller

Add the following 2 APIs to Controller:
> void addSubscriber (String RGName, String scopeName, String streamName)
> void removeSubscriber (String RGName, String scopeName, String streamName)

and these should add/remove a reader group from the list of Subscribers for the Stream.

pbelgundi:

**Change log description**  
This PR adds the following 2 APIs to ControllerService:

```
void addSubscriber (String RGName, String scopeName, String streamName)
void removeSubscriber (String RGName, String scopeName, String streamName)
void updateSubscriberStreamCut(String RGName, String scopeName, String streamName, Streamcut streamCut)
```

**Purpose of the change**  
Fixes #5109 #5110 

**What the code does**  

Add/Remove a subscriber to a Stream

_Newly added classes:_
StreamSubscriber.java

_Classes with changes:_
StreamMetadataTasks.java
StreamMetadataStore.java
AbstractStreamMetadataStore.java
Stream.java
PersistentStreamBase.java
PravegaTablesStream.java

**How to verify it**  
Unit and Integration tests should pass.


controller/src/main/java/io/pravega/controller/server/eventProcessor/ControllerEventProcessors.java
```diff
@@ -156,6 +158,8 @@ public ControllerEventProcessors(final String host,
                 new SealStreamTask(streamMetadataTasks, streamTransactionMetadataTasks, streamMetadataStore, executor),
                 new DeleteStreamTask(streamMetadataTasks, streamMetadataStore, bucketStore, executor),
                 new TruncateStreamTask(streamMetadataTasks, streamMetadataStore, executor),
+                new AddSubscriberTask(streamMetadataTasks, streamMetadataStore, executor),
```
shiveshr, line 161: do you need to add a subscriber asynchronously? that can be done inline for the call as it should be a single metadata update -- a new subscriber id as a record in some table, right?

pbelgundi: A 'StreamSubscriberRecord' has the 'subscriberId' and 'SubscriberConfiguration' for multiple subscribers.
And addSubscriber and updateStreamCutForSubscriber, updates the same metadata record, hence it is necessary to handle this asynchronously.

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
```diff
@@ -324,6 +325,21 @@ public static final StreamConfig decode(String scope, String streamName, final S
...
+    /**
+     * Converts StreamConfiguration into StreamConfig.
+     *
+     * @param scope the stream's scope
+     * @param streamName The Stream Name
+     * @param subscriberId Id of the subscriber for this stream.
+     * @return StreamSubscriberInfo instance.
+     */
+    public static final StreamSubscriberInfo decode(String scope, String streamName, final String subscriberId) {
+        Preconditions.checkNotNull(subscriberId, "subscriber");
```
shiveshr, line 337: nit: preconditions for scope and stream as well.

controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
```diff
@@ -14,40 +14,7 @@
...
-import static io.pravega.shared.MetricsNames.UPDATE_STREAM_LATENCY;
-import static io.pravega.shared.MetricsNames.CREATE_KVTABLE_LATENCY;
-import static io.pravega.shared.MetricsNames.CREATE_KVTABLE;
-import static io.pravega.shared.MetricsNames.KVTABLE_SEGMENTS_COUNT;
-import static io.pravega.shared.MetricsNames.CREATE_KVTABLE_FAILED;
-import static io.pravega.shared.MetricsNames.DELETE_KVTABLE_LATENCY;
-import static io.pravega.shared.MetricsNames.DELETE_KVTABLE;
-import static io.pravega.shared.MetricsNames.DELETE_KVTABLE_FAILED;
-import static io.pravega.shared.MetricsNames.globalMetricName;
+import static io.pravega.shared.MetricsNames.*;
```
shiveshr, line 17: nit: revert this please

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -158,6 +160,37 @@
                 }, executor);
     }
 
+    public CompletableFuture<AddSubscriberStatus> addSubscriber(String scope, String stream, final String subscriber) {
+        Preconditions.checkNotNull(scope, "scopeName is null");
+        Preconditions.checkNotNull(stream, "streamName is null");
+        Preconditions.checkNotNull(subscriber, "subscriber is null");
+        Timer timer = new Timer();
+        return streamMetadataTasks.addSubscriber(scope, stream, subscriber, null)
```
shiveshr, line 168: do you still need to route it through stream metadata tasks? you can perhaps directly write using the store apis.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -158,6 +160,37 @@
...
+
+        return streamMetadataTasks.getSubscribersForStream(scope, stream, null);
+    }
+
+    public CompletableFuture<RemoveSubscriberStatus> removeSubscriber(String scope, String stream, final String subscriber) {
+        Preconditions.checkNotNull(scope, "scopeName is null");
+        Preconditions.checkNotNull(stream, "streamName is null");
+        Preconditions.checkNotNull(subscriber, "subscriber is null");
+        Timer timer = new Timer();
+        return streamMetadataTasks.removeSubscriber(scope, stream, subscriber, null)
```
shiveshr, line 187: same here

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -823,6 +824,21 @@ private void handleStreamMetadataExists(final long timestamp, CompletableFuture<
         return result;
     }
 
+    @Override
+    CompletableFuture<Void> createSubscribersDataIfAbsent(StreamSubscribersRecord data) {
+        return null;
```
shiveshr, line 829: please implement or throw notimplemented exception

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -823,6 +824,21 @@ private void handleStreamMetadataExists(final long timestamp, CompletableFuture<
...
+    }
+
+    @Override
+    CompletableFuture<VersionedMetadata<StreamSubscribersRecord>> getSubscribersData(boolean ignoreCached) {
+        return null;
```
shiveshr, line 834: here and everywhere else in this class.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -156,6 +158,50 @@ public String getScopeName() {
         return deleteStream();
     }
 
+    /**
+     * Create a subscribersRecord with provided subscriber for this Stream.
+     * @return future of operation.
+     */
+    @Override
+    public CompletableFuture<Void> createSubscribersRecord(String subscriber) {
```
shiveshr, line 166: we can maybe always create this subscriber record for a stream as part of stream creation.. 
that way it only needs to be updated if a subscriber registers or unregisters.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -156,6 +158,50 @@ public String getScopeName() {
...
+        ImmutableMap<String, SubscriberConfiguration> subscribers
+                = new ImmutableMap.Builder<String, SubscriberConfiguration>()
+                .put(subscriber, SubscriberConfiguration.EMPTY).build();
+        return createSubscribersDataIfAbsent(new StreamSubscribersRecord(subscribers));
+    }
+
+    public CompletableFuture<Void> updateSubscribers(final String subscriber,
```
shiveshr, line 173: please add javadoc.. also mention that it can throw write conflict exceptions.

controller/src/main/java/io/pravega/controller/store/stream/StreamMetadataStore.java
```diff
@@ -287,6 +289,67 @@
                                                                                            final OperationContext context,
                                                                                            final Executor executor);
 
+
+    /**
+     * Updates the subscribers metadata for an existing stream.
```
shiveshr, line 294: nit: create. 
also store can expose AddSubscriber and RemoveSubscriber apis. 
internally the subscriber record can be created at the time of stream creation and is updated whenver a new subscriber is added/removed by doing a get and set.
and this can fail with write conflict to the caller may need to handle and retry the operation.
we follow optimistic concurrency.. the assumption here is that subscriber registration is a rare occurrence and there will not be many conflicts.

controller/src/main/java/io/pravega/controller/store/stream/records/StreamSubscribersRecord.java
```diff
@@ -0,0 +1,115 @@
...
+                                                                .collect(Collectors.toMap(p -> p.getKey(), p -> p.getValue()));
+        return new StreamSubscribersRecord(ImmutableMap.copyOf(streamSubscribers));
+    }
+
+    private static class StreamSubscribersRecordBuilder implements ObjectBuilder<StreamSubscribersRecord> {
+    }
+
+    @SneakyThrows(IOException.class)
+    public static StreamSubscribersRecord fromBytes(final byte[] record) {
+        //InputStream inputStream = new ByteArrayInputStream(record, 0, record.length);
```
shiveshr, line 73: nit: remove commented code

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * The configuration of a Stream Subscriber.
+ */
+@Data
+@Builder
+public class SubscriberConfiguration implements Serializable {
+
+    public static final SubscriberConfigurationSerializer SERIALIZER = new SubscriberConfigurationSerializer();
+    public static final SubscriberConfiguration EMPTY = new SubscriberConfiguration(0L, ImmutableMap.of());
+    private static final long serialVersionUID = 1L;
```
shiveshr, line 35: why do we need this?

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Time when this configuration was updated.
+     */
+    private final long updateTime;
+
+    /**
+     * Truncation Stream cut published by this subscriber
+     */
+    private final ImmutableMap<Long, Long> streamCut;
```
shiveshr, line 44: please dont put the position/streamcut inside the subscriber record. since this record is actually written under the `StreamSubscribersRecord` this will be problematic on two dimensions:
1. with large number of subscribers, the size of this record can exceed the 1mb limit of kvt record if there are large number of segments in the streamcut. its better to divide and rule. 
2. this will mean StreamSubscribersRecord is upadted with every "report" from a subscriber which means the write conflicts can occur lot more whereby making optimistic concurrency undesirable.

my suggestion is to do as follows:
- StreamSubscribersRecord should only contain a list of subscriber ids. 
- create separate record for each subscriber for their reported position.

addSubscriber/removeSubscriber updates the StreamSubscribersRecord with the subscriber's id. 
then for each subscriber there is a separate key called say `subscriber_<subscriberId>`.
when subscriber with `id` reports its streamcut, its streamcut is updated in `subscriber_id` record. 

getSubscribers will first fetch the StreamSubscribersRecord  and then fetch all their individual records corresponding to their last reported positions. 

removesubscriber can perform batch update of StreamSubscribersRecord and subscriber_id record by updating the former and removing the later via an atomic kvt update.

shiveshr (changes requested):

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -525,6 +525,116 @@ private RetryAndThrowConditionally createRetryConfig(final ControllerImplConfig
...
+            RPCAsyncCallback<AddSubscriberStatus> callback = new RPCAsyncCallback<>(requestId, "addSubscriber", scope, streamName, subscriber);
+            new ControllerClientTagger(client, timeoutMillis).withTag(requestId, "addSubscriber", scope, streamName)
+                    .addSubscriber(ModelHelper.decode(scope, streamName, subscriber), callback);
+            return callback.getFuture();
+        }, this.executor);
+        return result.thenApply(x -> {
+            switch (x.getStatus()) {
+                case FAILURE:
+                    log.warn(requestId, "Failed to update stream: {}", streamName);
+                    throw new ControllerFailureException("Failed to add subscriber: " + subscriber);
```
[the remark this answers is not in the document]
pbelgundi: This throws ControllerFailureException in line with other APIs that also throw the same exception on getting a FAILURE status.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -525,6 +525,116 @@ private RetryAndThrowConditionally createRetryConfig(final ControllerImplConfig
...
+                case SCOPE_NOT_FOUND:
+                    log.warn(requestId, "Scope not found: {}", scope);
+                    throw new IllegalArgumentException("Scope does not exist: " + scope);
+                case STREAM_NOT_FOUND:
+                    log.warn(requestId, "Stream does not exist: {}", streamName);
+                    throw new IllegalArgumentException("Stream does not exist: " + streamName);
+                case SUBSCRIBER_EXISTS:
+                    log.warn(requestId, "Subscriber {} for stream {}/{} already exists {}.", subscriber, scope, streamName);
```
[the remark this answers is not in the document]
pbelgundi: 2 unrelated users could try to add the same subscriber name/Id and if we just return success on each addition they'll never know they are overwriting each other's data.
So it is important to error out if the subscriber already exists.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -525,6 +525,116 @@ private RetryAndThrowConditionally createRetryConfig(final ControllerImplConfig
...
+                    log.warn(requestId, "Failed to update stream: {}", streamName);
+                    throw new ControllerFailureException("Failed to add subscriber: " + subscriber);
+                case SCOPE_NOT_FOUND:
+                    log.warn(requestId, "Scope not found: {}", scope);
+                    throw new IllegalArgumentException("Scope does not exist: " + scope);
+                case STREAM_NOT_FOUND:
+                    log.warn(requestId, "Stream does not exist: {}", streamName);
+                    throw new IllegalArgumentException("Stream does not exist: " + streamName);
+                case SUBSCRIBER_NOT_FOUND:
+                    log.warn(requestId, "Subscriber does not exist: {} for stream {}/{}", subscriber, scope, streamName);
```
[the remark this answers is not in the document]
pbelgundi: same as above

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
pbelgundi: sure, will add these.

controller/src/main/java/io/pravega/controller/metrics/StreamMetrics.java
pbelgundi: wonder why checkstyle did not complain here... will revert....

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
pbelgundi: ok... this got missed. will implement.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
pbelgundi: will do

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: The reason I did not want to do it is because a lot of Streams may never need CBR, in which case that additional metadata is an unnecessary overhead that they can do without.
Also for existing Pravega deployments, that move to the new version with CBR this metadata would need to be explicitly added when the Stream moves to CBR.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: ok. sure

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
pbelgundi: done

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberConfiguration.java
pbelgundi: 'SubscriberConfiguration' encapsulates all and any data that needs to be stored about a Stream Subscriber. Today it has only the truncation StreamCut but could have more stuff in the future.

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberConfiguration.java
pbelgundi: > 1. with large number of subscribers, the size of this record can exceed the 1mb limit of kvt record if there are large number of segments in the streamcut. its better to divide and rule.

The 1 MB limit would be exceeded even when we maintain separate records for subscribers and truncation stream-cuts. Assuming each 'Subscriber --> SC' map entry takes up 100 bytes, it would take 10K subscribers to hit that limit....which is a pretty large number.... Having Subscriber name as part of the same record only marginally impacts the table segment record limitation.....

> 2. this will mean StreamSubscribersRecord is updated with every "report" from a subscriber which means the write conflicts can occur lot more whereby making optimistic concurrency undesirable.

The downside of maintaining 2 separate records is the need to maintain consistency between them.
Optimistic concurrency concerns due to frequent update of truncation stream-cuts are very valid, but it has nothing to do with Subscribers being part of the same metadata, as subscriber addition is a much infrequent operation.

One way to deal with this can be to have a separate record in Stream Metadata per subscriber and update just that record on stream-cut updates.

controller/src/main/java/io/pravega/controller/store/stream/records/StreamSubscribersRecord.java
pbelgundi: done

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberConfiguration.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -525,6 +525,116 @@ private RetryAndThrowConditionally createRetryConfig(final ControllerImplConfig
...
+        Preconditions.checkNotNull(scope, "scopeName");
+        Preconditions.checkNotNull(streamName, "stream");
+        final long requestId = requestIdGenerator.get();
+        long traceId = LoggerHelpers.traceEnter(log, "getSubscribersForStream", streamName);
+        final CompletableFuture<StreamSubscribers> result = this.retryConfig.runAsync(() -> {
+            RPCAsyncCallback<StreamSubscribers> callback = new RPCAsyncCallback<>(requestId, "getSubscribersForStream", scope, streamName);
+            client.withDeadlineAfter(timeoutMillis, TimeUnit.MILLISECONDS).getSubscribersForStream(ModelHelper.createStreamInfo(scope, streamName), callback);
+            return callback.getFuture();
+        }, this.executor);
+        return result.thenApply(subscribers -> {
```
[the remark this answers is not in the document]
pbelgundi: I primarily added this API for testing. Client does not invoke it ever. Hence did not add a separate status for this.
But it can be added if you thing this could be invoked by the Client.

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberConfiguration.java
pbelgundi: removed now. fixed.

client/src/main/java/io/pravega/client/control/impl/Controller.java
```diff
@@ -118,6 +118,46 @@
...
 
+    /**
+     * API to add a Subscriber to the Stream.
+     * @param scope Scope name
+     * @param streamName Stream name
+     * @param subscriber Name/Id that uniquely identifies a Stream Subscriber.
+     * @return A future which will throw if the operation fails, otherwise returning a boolean to
+     *         indicate that the subscriber was updated in Stream Metadata.
+     */
+    CompletableFuture<Boolean> addSubscriber(final String scope, final String streamName, final String subscriber);
```
shiveshr, line 129: why is it a boolean? is boolean helpful to user if the subscriber was already present?

client/src/main/java/io/pravega/client/control/impl/Controller.java
```diff
@@ -118,6 +118,46 @@
...
+
+    /**
+     * API to remove a Subscriber from list of Subscribers for the Stream.
+     * @param scope Scope name
+     * @param streamName Stream name
+     * @param subscriber Name/Id that uniquely identifies a Stream Subscriber..
+     * @return A future which will throw if the operation fails, otherwise returning a boolean to
+     *         indicate that the subscriber was updated in Stream Metadata.
+     */
+    CompletableFuture<Boolean> removeSubscriber(final String scope, final String streamName, final String subscriber);
```
shiveshr, line 139: same question: what is the reason to return a boolean here?

client/src/main/java/io/pravega/client/control/impl/Controller.java
```diff
@@ -118,6 +118,46 @@
...
+     * API to update the truncation StreamCut for a particular Subscriber on Controller.
+     * Used when Stream has Consumption Based Retention Policy configured.
+     * @param scope Scope name
+     * @param streamName Stream name
+     * @param subscriber Name/Id that uniquely identifies a Stream Subscriber..
+     * @param streamCut StreamCut at which Stream can be Truncated for a Consumption based retention policy
+     * @return A future which will throw if the operation fails, otherwise returning a boolean to
+     *         indicate that the subscriber was updated in Stream Metadata.
+     */
+    CompletableFuture<Boolean> updateTruncationStreamCut(final String scope, final String streamName, final String subscriber, final StreamCut streamCut);
```
shiveshr, line 159: nit: since this api is in the context of a subscriber, the name should reflect it.. perhaps a better name for this API would be `reportSubscriberStreamCut` or `reportSubscriberPosition`.

client/src/main/java/io/pravega/client/control/impl/Controller.java
```diff
@@ -118,6 +118,46 @@
...
+
+    /**
+     * API to update the truncation StreamCut for a particular Subscriber on Controller.
+     * Used when Stream has Consumption Based Retention Policy configured.
+     * @param scope Scope name
+     * @param streamName Stream name
+     * @param subscriber Name/Id that uniquely identifies a Stream Subscriber..
+     * @param streamCut StreamCut at which Stream can be Truncated for a Consumption based retention policy
+     * @return A future which will throw if the operation fails, otherwise returning a boolean to
+     *         indicate that the subscriber was updated in Stream Metadata.
```
shiveshr, line 157: nit: subscriber`'s position` was updated

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
shiveshr: ok, but we need better exception handling for these cases. A pravega user is not expected to be aware of controller and controller failure exceptions. typically any exception can be propagated to the user apis and if they are actionable, then they should be specific, and perhaps even checked. I agree that we havent done this consistently across APIs. But we have the opportunity here to evaluate what is right behaviour for this API. it really depends on how the error is translated when its presented to the user, esp if we expect them to `handle` the error in a meaningful way.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
shiveshr: i see your point. however let me ask a counter question - is the subscriber id same as readergroup id or is it chosen by the user?
if its former, then it doesnt matter if two users were attempting to perform this action simultaneously. 
However, if its implicit and handled at a readergroup level,

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
shiveshr: remove is different from above case. since we are not doing conditional remove, but is predicated on the subscriber's name. it should perhaps not throw the exception. 

in fact i would say that this should not even be a response code for remove api from idempotent case's perspective.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
shiveshr: the response structure for this api is a defined as a new protobuf message. so you have the opportunity to include a status and a response list. 
and you have pointed out an important thing -- that the pattern for streamnotfound is not consistently followed across apis stream apis.. please create a new issue in backlog for these (unrelated to current change set) to include a status with each such api where we can set the appropriate status or make use of grpc status runtime to indicate such failures..

client/src/main/java/io/pravega/client/control/impl/Controller.java
shiveshr, line 159: also add "throws" javadoc for exceptions that can be thrown from this if user needs to handle those exceptions. for instance - stream cut not valid should be propagated to the user.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -275,6 +282,184 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                        return CompletableFuture.completedFuture(subscribersData.keySet().stream().collect(Collectors.toList()));
+                    }).exceptionally(ex -> {
+                              log.warn(requestId, "Exception thrown in trying to get list of Stream subscribers. {}",
+                                      ex.getMessage());
+                              Throwable cause = Exceptions.unwrap(ex);
+                              if (cause instanceof TimeoutException) {
+                                  throw new CompletionException(cause);
+                              } else {
+                                  log.warn(requestId, "getSubscribersForStream failed due to {}", cause);
+                                  return emptySubscribersList;
```
shiveshr, line 410: why return empty subscriber list when there is an exception?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -275,6 +282,184 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+        final long requestId = requestTracker.getRequestIdFor("updateTruncationStreamCut", scope, stream);
+
+        return RetryHelper.withRetriesAsync(() -> streamMetadataStore.checkStreamExists(scope, stream)
+                .thenCompose(exists -> {
+                    // 1. check Stream exists
+                    if (!exists) {
+                        return CompletableFuture.completedFuture(UpdateSubscriberStatus.Status.STREAM_NOT_FOUND);
+                    }
+                    // 2. check if StreamCut is valid
+                    return streamMetadataStore.isStreamCutValid(scope, stream, truncationStreamCut, context, executor)
```
shiveshr, line 439: we should also make use of "greater" than property to see if the new streamcut is strictly greater than equal to previously reported truncation streamcut for the given subscriber so that it always move ahead.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -212,10 +224,12 @@ private String getWritersTableName(String id) {
             String metadataTable = getMetadataTableName(id);
             String epochWithTxnTable = getEpochsWithTransactionsTableName(id);
             String writersPositionsTable = getWritersTableName(id);
+            String subscribersTable = getSubscribersTableName(id);
             return CompletableFuture.allOf(storeHelper.createTable(metadataTable),
-                    storeHelper.createTable(epochWithTxnTable), storeHelper.createTable(writersPositionsTable))
-                                    .thenAccept(v -> log.debug("stream {}/{} metadata tables {}, {} & {} created", getScope(), getName(), metadataTable,
-                                            epochWithTxnTable, writersPositionsTable));
+                    storeHelper.createTable(epochWithTxnTable), storeHelper.createTable(writersPositionsTable),
+                    storeHelper.createTable(subscribersTable))
```
shiveshr, line 230: although this approach is fine, but should we put the list of subscribers in the stream metadata table itself? I mean in principle it would be better if we avoid creating an additional table segment for each stream. 
for example:
key: subscribers value: List<subscriberIds>
key: subscriber-<id> value: <stream-cut>

that way we avoid creating a separate table. and with this model, we dont necessarily need to iterate over entries in a separate table.. instead use two queries - 1 to get list of subscribers and then the streamcut for each subscriber while performing the truncation.

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -237,7 +241,32 @@
         return getId().thenCompose(id -> store.getCachedData(creationPath, id, x -> BitConverter.readLong(x, 0))
                 .thenApply(VersionedMetadata::getObject));
     }
-    
+
+    @Override
+    public CompletableFuture<Void> createSubscriber(String subscriber) {
+        return null;
```
shiveshr, line 247: Please throw NotImplementedException or UnsupportedException

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
```diff
@@ -237,7 +241,32 @@
...
+    }
+
+    @Override
+    public CompletableFuture<VersionedMetadata<StreamSubscriber>> getSubscriber(String subscriber) {
+        return null;
```
shiveshr, line 252: same for all apis in this class

shiveshr (changes requested):

client/src/main/java/io/pravega/client/control/impl/Controller.java
pbelgundi: All APIs in this interface return boolean indicating the operation either failed or succeeded...The ControllerImpl implementation returns a true/false based on weather the operation was successful or not...
For "addSubscriber" the operation is successful if subscriber was not present and got added, in every other case its a failure....

client/src/main/java/io/pravega/client/control/impl/Controller.java
pbelgundi: This is a client interface and every operation on Controller is expected to fail or succeed.
There is no _return status_ for other operations as well.
Only the getXXX operations return non-boolean values.

client/src/main/java/io/pravega/client/control/impl/Controller.java
pbelgundi: sure will fix.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
pbelgundi: sure. can take up as part of a separate refactoring issue.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
pbelgundi: same as readergroup id - it is fine for 2 users to do this simultaneously, but only 1 would succeed and the other would fail.... a subsequent getSubscribers() can be used to determine of the subscriber actually got added.....

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
pbelgundi: All other delete APIs return response codes. I think it is important to keep APIs consistent for better user experience...

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
pbelgundi: ok. will do

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: The DataNotFoundException here indicates there is no subscriber data present for this Stream, which can happen if users query subscriber data for a Stream that does not have "subscribers". So this is not really an exceptional case and hence returning empty list here.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: There may be "recovery" scenarios where a Reader Group restarts from an earlier Stream-Cut and hence publishes an older StreamCut as a truncation Stream-Cut later.... so I would not want to add that check here...

client/src/main/java/io/pravega/client/control/impl/Controller.java
pbelgundi: Ok will change.

pbelgundi: ok

controller/src/main/java/io/pravega/controller/server/ControllerService.java
pbelgundi: I'd prefer encapsulating the various steps like checking for streamExists etc inside StreamMetadataTasks..

controller/src/main/java/io/pravega/controller/store/stream/ZKStream.java
pbelgundi: done

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
shiveshr: i am not convinced if we should throw an exception in this case, because its not an exceptional but idempotent case. return a boolean to indicate if it exists.. throwing illegal argument is certainly misleading.

client/src/main/java/io/pravega/client/control/impl/Controller.java
shiveshr: semantics of true false and error cases could be consistent with create stream api:
true - if its newly created
false - if it already exists
exception - specific to the error occurred.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: but can empty list also be returned if the stream did not exist?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: but can stream cut be "reset" back to an older streamcut? 
what if controller had already used the more recent one for truncation.
we should strictly move the streamcuts forward..
it also takes care of scenarios where multiple different controllers received different requests and if newer update is already set, the stale value should not override it.
take for example:

T0: rg ->SC1 -> controller1
t1: connection failure
t2: rg ->SC2 -> controller2

controller1 is attempting to update SC1 to segment store and controller 2 is attempting to update SC2.. 

if SC2 is updated, we should reject SC1.

pbelgundi: The validation you are suggesting can be added at a later point, if needed.
Do we do this check with manual truncation API today?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
pbelgundi: Fixed this and added new keys to existing stream metadata table:
K                            --->    V
SUBSCRIBERS        ----> sub1, sub2, sub3.... subn
SUBSCRIBER_sub1 ----> sub1, ts1, StreamCut1
SUBSCRIBER_sub2 -----> sub2, ts2, StreamCut2
. . .

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
pbelgundi: Ok changed return status for SUBSCRIBER_NOT_FOUND to false.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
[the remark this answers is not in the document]
pbelgundi: changed return status for SUBSCRIBER_EXISTS to false instead of throwing exception.

client/src/main/java/io/pravega/client/control/impl/Controller.java
pbelgundi: Changed as per createStream.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: yes we do, we check that the streamcut is actually moving forward. 
for each range, if segment number in one streamcut is same or higher, then that stream cut is greater than.

pbelgundi: Added this validation.

shiveshr (approved):

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -671,6 +672,20 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
                       });
     }
 
+    @Override
+    public CompletableFuture<Boolean> isStreamCutValidForTruncation(final Map<Long, Long> streamCut, Map<Long, Long> previousStreamCut) {
```
shiveshr, line 676: we dont necessarily need a public method here. this can be done as part of precondition check for updating a subscriber streamcut internally. that will allow for get and update post validation. 
otherwise you are validating. and then refetchng subscriber's record and updating it which may mean the condition may have changed. 
Also the implementation of this method can be optimized slightly. 

PS: I have made those changes in my PR.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -275,6 +282,205 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                                                return streamMetadataStore.isStreamCutValid(scope, stream, truncationStreamCut, context, executor)
+                                                        .thenCompose( isValid -> {
+                                                         if (isValid) {
+                                                           return updateStreamCut(scope, stream, subscriber, truncationStreamCut, context, requestId);
+                                                         } else {
+                                                           return CompletableFuture.completedFuture(UpdateSubscriberStatus.Status.STREAMCUT_NOT_VALID);
+                                                         }
+                                                 });
+                                            } else {
+                                                return streamMetadataStore.isStreamCutValidForTruncation(scope, stream, truncationStreamCut,
```
shiveshr, line 451: here you are fetching subscriber record, checking its validity and then updating the subscriber record but the update is not predicated on previous subscriber record that you had fetched. this could mean if the  subscriber record was updated while you performed the check, it could get overwritten incorrectly.
this check should be done in the storage layer where you do a get followed by a check and then conditional update using the entity tag from the previous record.

pbelgundi: I;m fetching the subscriberRecord again in the StorageLayer and only then updating the streamcut on top. But I agree  this double fetch can be avoided.
Will change this to do the validity check and get/update directly in storage layer as done in case of RetentionSet

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: i am making the change in my PR to make the optimization to this method and its invocation in streamMetadataTasks.

shiveshr (changes requested):
Requesting one important change that i noticed. I have asked you to simply remove the incorrect check and i will include the check correctly in my PR so that we can expedite this.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: Ok

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: Fixed.

pbelgundi:
> Requesting one important change that i noticed. I have asked you to simply remove the incorrect check and i will include the check correctly in my PR so that we can expedite this.

I have corrected the incorrect check now, Please check.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: changed this to a private method

shiveshr (approved):

PR #5208, opened 2020-09-21
Issue 5207: (SegmentStore) Read Index Bug Fixes

### LINKED ISSUE #5207: Enable copy-on-read for all cache reads
**Problem description**
Set `ReadResult.CopyOnRead` to `true` by default for all reads. This will avoid situations where the cache blocks get evicted and replaced with other data before upstream code gets a chance to process that data.

**Problem location**
Segment Store.

Set it to true.

andreipaduroiu:
**Change log description**  
1. Setting `ReadResult.isCopyOnRead` to `true` by default. This will cause any upstream code reading from the cache to get a a copy of that data instead of a view. 
2. Fixed a bug in `StreamSegmentReadIndex.insertEntriesToCacheAndIndex` where it would incorrectly insert an entry if an existing entry that is a subset of the new one already exists.

**Purpose of the change**  
Fixes #5207.

**What the code does**  
- Copy-on-Read
   - Enable Copy-On-Read for any reads that end up getting data from the Cache. #5155 missed one case (in `WriterTableProcessor.readFromInMemorySegment`, which could lead to data corruption for Table Segments upon indexing).
    - By setting it to true by default, upstream code will have to explicitly opt-out (instead of opt-in), so any other cases (and future ones) should be covered.
    - The only exception to this rule is copying data to LTS. It uses a different API which provides a direct view of the underlying cache data, but that should be OK since by contract that data should not have been evicted in the first place.
- `StreamSegmentReadIndex.insertEntriesToCacheAndIndex`
    - This method was not incrementing the `segmentOffset` variable if it had to partially insert an entry due to another one that overlaps it exists. This caused the wrong slice of that new data to be inserted in that case.

**How to verify it**  
New unit test added for second bug.
Build must pass.

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
@@ -224,7 +224,6 @@ public class PravegaRequestProcessor extends FailingRequestProcessor implements 
      * an appropriate message is sent back over the connection.
      */
     private void handleReadResult(ReadSegment request, ReadResult result) {
-        result.setCopyOnRead(true); // Get a copy of any data from the cache to avoid losing it due to cache eviction.
         String segment = request.getSegment();
         ArrayList<BufferView> cachedEntries = new ArrayList<>();
         ReadResultEntry nonCachedEntry = collectCachedEntries(request.getOffset(), result, cachedEntries);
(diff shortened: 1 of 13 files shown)


eolivelli (approved):
I see that this is still a draft

But overall LGTM

Do you know if is there any significant performance impact ?

andreipaduroiu:
@eolivelli There shouldn't be an impact. This is just _partially_ reverting some of the 0.8 changes we made (i.e., when we reduced the number of memory copies on the read path from 3 to 1 (or even 0 in internal cases)) by making sure that whatever we read from the cache itself is a copy and not a view of the raw buffers inside (which are at risk of eviction and thus replacement while you're using them). This does not affect appends or tail reads. It's mostly on "catch-up"/historical reads that we need it.


The second bug fix (haven't updated the text yet) is required as it may corrupt the data in memory. In my local tests, it actually improved performance in certain situations (i.e., cache full/ cache pressure), so overall I do not expect any sort of impact.

eolivelli:
Good

RaulGracia (approved):
LGTM, I have done an entire round of performance experiments and it is not inducing any noticeable performance problem.

tkaitchuck (approved):

PR #5209, opened 2020-09-22
Issue 5218: Ensure client closes a connection in case of a read timeout.

### LINKED ISSUE #5218: Ensure client closes a connection in case of a read timeout
**Problem description**
When one of the segment stores(SSS) was brought down in a k8s environment it was observed that the client detects that the connection is dropped only after 15-16 mins. During this period we can only see ARP packets trying to figure the mac of the downed SSS.
```
09:45:08.014523 ARP, Request who-has IP_KILLED_SSS tell IP_WHERE_CLIENT_IS_RUNNING, length 28
//this repeats for this duration of 15-16 mins..
```
Only after this, the client observes a "No route to host" exceptions which causes the client to re-establish the connection with the newer SSS IP provided by the controller.

**Problem location**
Read timeout would mitigate this issue in case of a half TCP-connection drop.

Enhance the client to have a read timeout.

shrids:
**Change log description**  
- Ensure the client terminates a connection in the case of a socket read timeout.

**Purpose of the change**  
Fixes #5218 

**What the code does**  
The Pravega client already sends a KeepALive message over a network connection if there has been no message sent over a given connection. Any exception observed during this KeepAlive causes the client to terminate/close the network connection. The client then re-establishes a network connection to the segment store based on the URI provided by the controller.

During one of tests in a K8s cluster we observed that an abrupt deletion of the SSS instance can sometimes cause a scenario where client realizes that the connection is down only after a period of 15 mins. During this period the TCP dump just shows a flood of ARP packets where it is attempting to figure the MAC of the now missing SSS instance. The arp timeout is node/host specific and on linux it depends on the `base_reachable_time_ms ` configuration.

By ensuring the client has a socket read timeout this PR ensures the client forcefully closes the connection in the scenario of Half-dropped tcp connection described above. It is in the order of minutes to ensure we do not cause unnecessary connection closures. 

Note: This change also ensures read on the Socket by `io.pravega.client.connection.impl.IoBuffer` is not potentially blocked forever and an IOException will be thrown in the event of read timeout.

**How to verify it**  
(It also needs to be validated on the DR test bed.)


tkaitchuck (approved):
This is fine. But I don't think it is strictly needed given how keepalive works.

eolivelli (approved):
It looks like a good candidate for a cherry pick to 0.8

shiveshr (approved):

andreipaduroiu (approved):

PR #5224, opened 2020-09-29
Issue 5168: Remove Controller Dependencies

### LINKED ISSUE #5168: Remove controller dependencies from System Test(s).
**Problem description**

The system test project (`test:system`) pulls in several dependencies from the controller. This is problematic because it prevents the system test framework from being build and run using the same JDK version as the client.

**Problem location**

1. `PravegaControllerService.java` -> `io.pravega.controller.util.Config`
2. `ControllerRestApiTest.java` -> `io.pravega.controller.server.rest.generated.model.*`
3. `WatermarkingTest.java` -> `io.pravega.controller.store.stream.StoreException`

1. While a little 'messy', only configuration properties are pulled in. It is sufficient to declare hard coded properties.
2. The best avenue would seem to be to relocate where the generated models are kept, e.g., in `shared:controller-api`.
    * For now, we can retire the test, and change it to an integration test.
3. The use of the `StoreException` is based on an outdated assumption (watermark streams are no longer created lazily), making it entirely erroneous. Simply remove the referencing call.

co-jo:
**Change log description**  

The purpose of this PR is to remove all references of the `controller` project from the `test:system` project.
* This will enable us to properly deploy the system tests using JDK 8, instead of JDK 11 (as this is the JDK version the controller uses).

**Purpose of the change**  
Fixes #5168 

**What the code does**  
* Remove reference to `io.pravega.controller.util.Config` class, and replace with hard coded values.
* Remove outdated/unnecessary check (`WatermarkingTests`)
* Completely remove `ControllerRestApiTest` due to similarity of existing integration test.  

**How to verify it**  


test/system/src/main/java/io/pravega/test/system/framework/services/marathon/PravegaControllerService.java
```diff
@@ -122,18 +121,18 @@ private App createPravegaControllerApp() {
 
         String componentCode = "controller";
 
-        //set env
+        // These properties are defined in "io.pravega.controller.util.Config".
```
andreipaduroiu, line 124: This is not consistent with the PR description. 

In the description you say "remove Controller dependencies".

Here, you are removing references **from the System test package** to the Controller.

A few issues that I'd like resolved:
1. Make the PR description consistent with the contents of the PR.
2. Remove the dependency on the controller project from this project's `build.gradle` section.
3. I am concerned that this change in particular (swapping `Config.*` with hardcoded strings) will break if we make a change to those properties at one point.

andreipaduroiu (changes requested):

test/system/src/main/java/io/pravega/test/system/framework/services/marathon/PravegaControllerService.java
co-jo: In response to #3. Certainly a possibility. I took inspiration from `PravegaSegmentStoreService.java` which also uses hard-coded values to set config options. Can certainly refactor the controller's `Config` class into a shared package if that's a better approach.

tkaitchuck (approved):

shiveshr (approved):

ravisharda (changes requested):
We seem to be removing Controller REST API system tests in this PR, presumably to remove dependencies on classes from the `Controller` module. 

Rather than removing the tests, can we modify them to invoke REST API's dynamically, so that the tests remain? The removal of tests should be avoided to the extent possible.

test/system/src/main/java/io/pravega/test/system/framework/services/marathon/PravegaControllerService.java
ravisharda, line 124: nit: This comment is unnecessary and slightly problematic: 
* The target class may be refactored with a change in name and namespace, and this comment will then go stale. Ideally, classes should be referenced using a `@link` annotation (when used in JavaDocs) or the like. 
* A simple text search will easily locate the class where these properties are located. 


*

test/system/src/main/java/io/pravega/test/system/framework/services/marathon/PravegaControllerService.java
```diff
@@ -122,18 +121,18 @@ private App createPravegaControllerApp() {
...
         String controllerSystemProperties = "-Xmx512m" +
-                setSystemProperty(Config.PROPERTY_ZK_URL.getFullName(componentCode), zk) +
-                setSystemProperty(Config.PROPERTY_RPC_HOST.getFullName(componentCode), this.id + ".marathon.mesos") +
-                setSystemProperty(Config.PROPERTY_RPC_PORT.getFullName(componentCode), String.valueOf(CONTROLLER_PORT)) +
-                setSystemProperty(Config.PROPERTY_SERVICE_PORT.getFullName(componentCode), String.valueOf(CONTROLLER_PORT)) +
-                setSystemProperty(Config.PROPERTY_REST_PORT.getFullName(componentCode), String.valueOf(REST_PORT)) +
+                setSystemProperty(String.format("%s.%s", componentCode, "zk.connect.uri"), zk) +
+                setSystemProperty(String.format("%s.%s", componentCode, "service.rpc.published.host.nameOrIp"), this.id + ".marathon.mesos") +
```
ravisharda, line 127: There is a lot of code repetition here. Extracting this out to a method that takes only the config name and the value and assumes the component code can simplify this:

```
setConfig("zk.connect.uri", zk)
```

p.s.: The name of the method `setSystemProperty()` is also misleading. It doesn't really "set the system property", but rather returns a newly created string representing the system property (`-Dprop=value`). But, that's preexisting code, and you may or may not want to  change that too.

test/system/src/test/java/io/pravega/test/system/WatermarkingTest.java
```diff
@@ -153,13 +151,6 @@ public void watermarkingTests() throws Exception {
         Stream streamObj = Stream.of(SCOPE, STREAM);
         scale(controller, streamObj);
 
-        // wait until mark stream is created
```
ravisharda, line 156: You remove statements in lines 157-161. What is the impact on the test? Given the lines below, it looks like the test is excepted to function without these lines now. But then the question is why were they needed in the first place. 

Am I missing something?

ravisharda (changes requested):

test/system/src/test/java/io/pravega/test/system/WatermarkingTest.java
co-jo: @shiveshr Would be able to explain better, but he mentioned two points to me about this test:

1. The `WatermarkingStream` was originally created lazily so the test had to wait for it to be created, but this is not the case anymore.
2. A store exception is never thrown back to the client, so the exception being compared to has no meaning.

co-jo:
> We seem to be removing Controller REST API system tests in this PR, presumably to remove dependencies on classes from the `Controller` module.
> 
> Rather than removing the tests, can we modify them to invoke REST API's dynamically, so that the tests remain? The removal of tests should be avoided to the extent possible.

I completely removed the test due to the almost exact copy from the integration test (`ControllerRestApiTest`). I would rather someone more familiar with those tests comment about this issue.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+    private String resourceURl;
+
+    public DynamicRestApiTest() {
+
+        org.glassfish.jersey.client.ClientConfig clientConfig = new org.glassfish.jersey.client.ClientConfig();
+        clientConfig.register(JacksonJsonProvider.class);
+        clientConfig.property("sun.net.http.allowRestrictedHeaders", "true");
+
+        if (Utils.AUTH_ENABLED) {
+            HttpAuthenticationFeature feature = HttpAuthenticationFeature.basic("admin", "1111_aaaa");
```
ravisharda, line 68: The credentials should be coming from a properties file instead of being hard-coded. As far as I remember, there is a properties file used in system tests that contain properties such as these.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+    private final Client client;
+    private WebTarget webTarget;
+    private String restServerURI;
+    private String resourceURl;
+
+    public DynamicRestApiTest() {
+
+        org.glassfish.jersey.client.ClientConfig clientConfig = new org.glassfish.jersey.client.ClientConfig();
+        clientConfig.register(JacksonJsonProvider.class);
+        clientConfig.property("sun.net.http.allowRestrictedHeaders", "true");
```
ravisharda, line 65: The reason why these lines are needed is not immediately apparent. Could you please add some comments that explain them.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+    public static void initialize() throws MarathonException {
+        URI zkUri = startZookeeperInstance();
+        startBookkeeperInstances(zkUri);
+        URI controllerUri = ensureControllerRunning(zkUri);
+        ensureSegmentStoreRunning(zkUri, controllerUri);
+    }
+
+    @Test
+    public void listScopes() {
+        Service conService = Utils.createPravegaControllerService(null);
```
ravisharda, line 92: Why don't we just call this `controllerService`? That'd reduce the work needed by readers of the code where this variable is used.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+        List<URI> ctlURIs = conService.getServiceDetails();
```
ravisharda, line 93: `controllerURIs`'d work better here.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+        URI controllerRESTUri = ctlURIs.get(1);
+        Invocation.Builder builder;
+        Response response;
+        String protocol = Utils.TLS_AND_AUTH_ENABLED ? "https://" : "http://";
+        restServerURI = protocol + controllerRESTUri.getHost() + ":" + controllerRESTUri.getPort();
+        log.info("REST Server URI: {}", restServerURI);
+
+        assertFalse(true);
```
ravisharda, line 101: What do you intend to do here? It looks like a faulty statement to me.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+        // TEST REST server status, ping test
```
ravisharda, line 103: This comment can be better phrased.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+        ensureSegmentStoreRunning(zkUri, controllerUri);
+    }
+
+    @Test
+    public void listScopes() {
+        Service conService = Utils.createPravegaControllerService(null);
+        List<URI> ctlURIs = conService.getServiceDetails();
+        URI controllerRESTUri = ctlURIs.get(1);
+        Invocation.Builder builder;
+        Response response;
```
ravisharda, line 96: The response should be closed too so that the underlying message entity input stream (if available and open) is closed and other resources associated with the response are released. You could perhaps use `@Cleanup` Lombok annotation for that.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+        log.info("REST Server URI: {}", restServerURI);
+
+        assertFalse(true);
+
+        // TEST REST server status, ping test
+        resourceURl = new StringBuilder(restServerURI).append("/ping").toString();
+        webTarget = client.target(resourceURl);
+        builder = webTarget.request();
+        response = builder.get();
+        assertEquals("Ping test", OK.getStatusCode(), response.getStatus());
```
ravisharda, line 108: The message won't add much value as it does not tell what is wrong. Ideally, the problem should be immediately apparent when someone looks at the test failure. 

Could you please change the message (for example: `"Received unexpected status code " + response.getStatus + " in response to ping request"`

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+        log.info("REST Server is running. Ping successful.");
```
ravisharda, line 109: This log statement is redundant, the control'd only move forward if the previous assertion passes. Regardless of whether the next assertion succeeds or fails, one'd know that the ping was successful. 

Having said that, it's alright if you want to keep it anyway.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,139 @@
...
+        boolean isStreamCreated = streamManager.createStream(scope1, stream1, StreamConfiguration.builder().scalingPolicy(ScalingPolicy.fixed(1)).build());
+        assertTrue("Failed to create stream", isScopeCreated);
+
+        // Query for an existing scope.
+        webTarget = client.target(restServerURI).path("v1").path("scopes");
+        builder = webTarget.request();
+        response = builder.get();
+        assertEquals("Get scopes failed.", OK.getStatusCode(), response.getStatus());
+        responseAsString = response.readEntity(String.class);
+        assertTrue(responseAsString.contains("\"scopeName\":\"_system\""));
```
ravisharda, line 137: Since you create a scope and stream earlier, you should also check if that shows up in the response too? Otherwise creating the scope/stream previously adds no value in this particular test.

ravisharda (changes requested):

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
co-jo: Sure.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
co-jo: Most of the lines commented on are copied strictly from the now removed `ControllerApiRestTest` file, so the original author would need to comment. I'll remove it and see if its still functional.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
co-jo: This was a mistake, did not mean to commit this.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
co-jo: The properties file suggest that there is token based authentication used with `io.pravega.client.ClientConfig`, but not this glassfish `ClientConfig`.

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+        final String scope1 = RandomStringUtils.randomAlphanumeric(10);
+        final String stream1 = RandomStringUtils.randomAlphanumeric(10);
+
+        Client client = ClientBuilder.newClient();
+        String responseAsString = null;
+
+        ClientConfig clientConfig = ClientConfig.builder()
+                .controllerURI(controllerGRPCUri)
+                .credentials(new DefaultCredentials("1111_aaaa", "admin"))
```
ravisharda, line 118: Credentials are hard-coded. Can we obtain these from the properties file instead?

ravisharda (changes requested):

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
co-jo: Fixed.

build.gradle
```diff
@@ -854,6 +855,9 @@ project('test:system') {
         compile group: 'org.glassfish.jersey.connectors', name:'jersey-apache-connector', version:jerseyVersion
         compile group: 'com.spotify', name: 'docker-client', version: dockerClientVersion
         compile group: 'io.kubernetes', name: 'client-java', version: k8ClientVersion
+        runtime group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: jerseyVersion
+        runtime group: 'org.apache.curator', name: 'curator-framework', version: apacheCuratorVersion
+        testCompile group: 'org.apache.curator', name: 'curator-framework', version: apacheCuratorVersion
```
andreipaduroiu, line 860: Why do we need this both for `runtime` and `testCompile`? Shouldn't `compile` be enough?

test/system/src/test/java/io/pravega/test/system/DynamicRestApiTest.java
andreipaduroiu: Unless I'm missing something, these are still hardcoded.

co-jo: Oops, forgot to push local changes.

build.gradle
co-jo: The `runtime` configuration is required to pull in the dependency into the `test-collection.jar` (below). It is however sufficient just to use `compile` instead of `testCompile`.

ravisharda (approved):

build.gradle
ravisharda: If you use `compile` dependency config, the dependency is available for both `runtime` and `testCompile` (as well as `testRuntime`). However, since `compile` is more permissive than what you may need here, IMO it'd be better to specify `runtime` and `testCompile` (or better still `testRuntime` if that's what you need) to limit the scope of these dependencies. That being said, it's not a big deal using just `compile` too. Just a thought.

andreipaduroiu (approved):

PR #5227, opened 2020-09-30
Issue: Suppress InterruptedException while closing Controller client.

shrids:
**Change log description**  
Suppress InterruptedException while closing the Controller client.

**Purpose of the change**  
TBD

**What the code does**  
Ensure InterruptedException is not thrown while closing the Controller client.

**How to verify it**  
Testing in progress.

### DIFF
--- a/client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
+++ b/client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
@@ -1238,10 +1238,13 @@ public class ControllerImpl implements Controller {
 
     private void closeChannel() {
         this.channel.shutdownNow(); // Initiates a shutdown of channel. Although forceful, the shutdown is not instantaneous.
-        Exceptions.handleInterrupted(() -> {
+        try {
             boolean shutdownStatus = channel.awaitTermination(20, TimeUnit.SECONDS);
             log.debug("Controller client shutdown has been initiated. Channel status: channel.isTerminated():{}", shutdownStatus);
-        });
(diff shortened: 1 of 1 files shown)


client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -1238,10 +1238,14 @@ public void close() {
...
         this.channel.shutdownNow(); // Initiates a shutdown of channel. Although forceful, the shutdown is not instantaneous.
-        Exceptions.handleInterrupted(() -> {
+        try {
             boolean shutdownStatus = channel.awaitTermination(20, TimeUnit.SECONDS);
             log.debug("Controller client shutdown has been initiated. Channel status: channel.isTerminated():{}", shutdownStatus);
-        });
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            log.info("Channel shutdown has been initiated. InterruptedException observed while awaiting termination of channel, retrying. Current termination status is {}", channel.isTerminated());
+            Exceptions.handleInterrupted(() -> channel.awaitTermination(20, TimeUnit.SECONDS));
```
tkaitchuck, line 1247: This is inside an catch interrupted block. The line above sets the interrupted state so the `awaitTermination` on this line is guaranteed to throw InteruptedException. (Which will be caught, re-set and rethrown by `handleInterrupted`) so this line is effectively equivalent to `throw e;` Which makes this whole change the same as the existing code except for the log statement.

shrids:
closing this as it does not improve behaviour.

PR #5239, opened 2020-10-07
Issue 5233: (SegmentStore) Flush-to-Storage API

### LINKED ISSUE #5233: Segment Container Flush Admin API
**Problem description**
Add an API in the `SegmentContainer` class that flushes everything in memory and Tier 1 into LTS/Tier 2. This includes:
- All segment data
- All Table Segment data and indices (the flush itself may trigger additional data to be moved over).
- All in-memory Segment metadata written to the Metadata Segment
    - This addition to the metadata segment must also be itself indexed and flushed to LTS.

This API:
- Should not be exposed externally.
- Should be clearly documented that it is for test purposes only. Not for production use.
    - This means that we do not need to worry (yet) about how this will work if the segment container fails over during execution.
    - However, it should not have adverse side effects (i.e., this should be invoke-able without interfering with the normal segment container lifecycle).

**Problem location**
Segment Container.

See above.

andreipaduroiu:
**Change log description**  
- Added an API on the `SegmentContainer` to fully flush all Tier 1 and in-memory data to LTS/Storage.
- This API is **experimental** and should only be used to aid testing. It is not yet supported for production use.

**Purpose of the change**  
Fixes #5233.

**What the code does**  
- `SegmentContainer.flushToStorage`
    - This method invokes `LogFlusher` to expedite the movement of all outstanding Tier 1 data into LTS. 
- `LogFlusher`
    - Works in 3 steps, with each step making sure that all data from previous steps are persisted to LTS.
    - Step 1: Creates a Metadata Checkpoint, extracts the Sequence Number for it, and instructs `StorageWriter` to flush all data up to, and including that Sequence Number.
        - This is executed in a loop as long as the `StorageWriter` reports that it had data to flush. The loop is required because each iteration of the `StorageWriter` may have, as effect, the generation of additional Tier 1 operations (ex: Attribute Aggregator, Table Segment Indexing, etc.)
        - The repeated invocations of the checkpoint-writer flush should converge into a stable state where a writer iteration has no further effects; as such, the subsequent iteration will report it had nothing to do.
    - Step 2: Persist all in-memory Segment Metadata to the Container Metadata Table Segment. 
        - This will include the effects of any modifications done at Step 1.
    - Step 3: repeat Step 1 to ensure that all modifications done at Step 2 are also persisted to LTS.
- `StorageWriter`
    - Added a new API `forceFlush` which sets a flag on the `WriterState` pointing to the desired "flush-to-sequence-number".
    - The StorageWriter will work (as usual) but pass the flag down to its segment processors instructing them to fully flush whatever outstanding data they have (i.e., ignore the time or size-based flush thresholds).
    - When the Storage Writer has flushed everything past the Operation with given sequence number, it notifies the requestor that the flush has completed.

**How to verify it**  
All unit tests must pass.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -100,6 +101,24 @@ protected void doStop() {
...
+     * interfere with the regular operation of {@link #runOnce()}.
+     *
+     * @param timeout Timeout for the operation.
+     * @return A CompletableFuture that, when completed, indicates that the operation completed.
+     */
+    CompletableFuture<Void> persistAll(Duration timeout) {
+        val tasks = this.metadata.getAllStreamSegmentIds().stream()
+                .map(this.metadata::getStreamSegmentMetadata)
+                .filter(Objects::nonNull)
+                .filter(sm -> !sm.isDeleted() && !sm.isMerged())
```
sachin-j-joshi, line 116: Not sure why we don't want to flush deleted and merged segment.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/WriterState.java
```diff
@@ -122,6 +129,57 @@ long getLastReadSequenceNumber() {
...
+    /**
+     * Indicates the fact that the {@link StorageWriter} has completed a flush stage.
+     *
+     * @param result The {@link WriterFlushResult} summarizing the flush stage.
+     */
+    void recordFlushComplete(WriterFlushResult result) {
+        val ffc = this.forceFlushContext.get();
+        if (ffc != null && ffc.flushComplete(result)) {
+            this.forceFlushContext.set(null);
+            ffc.getCompletion().complete(ffc.isAnythingFlushed());
```
sachin-j-joshi, line 154: What happens when there is nothing to flush?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
andreipaduroiu: The previous step in `LogFlusher` (before this is invoked) tells the `StorageWriter` to apply all the outstanding changes to Storage. That includes any deletions or mergers. Those segments' metadata may still linger in memory (not evicted yet) and should be treated as garbage. No need to write them down here. 

If you look in the `runOnceInternal` method in this class (the one that does evictions), you'll see similar logic.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/WriterState.java
andreipaduroiu: The StorageWriter may execute multiple iterations until it is able to complete this. When requested to force flush, it sets a context, and as long as that context is present, the Storage Writer will forcefully try to flush everything, read more data, flush, etc., until the requested "flush-to" sequence number has been processed.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/SegmentAggregatorTests.java
```diff
@@ -683,6 +683,55 @@ public void testFlushEmptyAppend() throws Exception {
         verifySegmentData(expectedData, context);
     }
 
+    /**
+     * Tests the flush() method with the force flag set.
+     * Verifies both length-based and time-based flush triggers, as well as flushing rather large operations.
+     */
+    @Test
+    public void testFlushForced() throws Exception {
+        final WriterConfig config = DEFAULT_CONFIG;
```
ManishKumarKeshri, line 692: One question: why didn't you use `INFREQUENT_FLUSH_WRITER_CONFIG`?

ManishKumarKeshri (commented):
Very very minor comments.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/WriterSegmentProcessor.java
```diff
@@ -54,9 +54,22 @@
     /**
      * Flushes the contents of the Processor.
      *
+     * @param force   If true, force-flushes everything accumulated in the {@link WriterSegmentProcessor}, regardless of
+     *                the value returned by {@link #mustFlush()}.
```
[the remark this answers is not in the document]
andreipaduroiu: Both are acceptable for Javadoc generation. The former will link to any method matching that name, while the latter will link to a specific method overload (looking at parameters, signature, etc.).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/LogFlusher.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private CompletableFuture<Void> flushAll(TimeoutTimer timer) {
+        // 1. Queue a checkpoint and get its SeqNo. This is poor man's way of ensuring all initiated ops are in.
+        // 2. Tell StorageWriter to flush all (new API, with seqNo). Includes: segment data and table segment indexing.
+        // 3. Repeat 1+2 until StorageWriter claims there is nothing more to flush.
+        val flushAgain = new AtomicBoolean(true);
+        val attemptNo = new AtomicInteger(0);
+        return Futures.loop(
+                () -> flushAgain.get() && attemptNo.getAndIncrement() < MAX_FLUSH_ATTEMPTS,
```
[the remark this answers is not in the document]
andreipaduroiu: MAX_FLUSH_ATTEMPTS is set to 10. At each attempt, it records a checkpoint into the Tier 1 log, gets its sequence number, and tells the Storage Writer to flush everything up to and including that seq no. The Storage Writer will not complete until it has finished doing so. This qualifies as one "attempt".

However, a Storage Writer iteration may produce side effects that get added to Tier 1 (i.e., for Table Segments, new data and/or attributes may be written) - we must make sure that all these get persisted to Storage as well. These should take at most one or two extra iterations, but to add a bit of buffer, I gave it up to 10 attempts to converge.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/AttributeAggregator.java
```diff
@@ -190,14 +190,16 @@ public boolean mustFlush() {
     /**
      * Flushes the contents of the Aggregator to the Storage.
      *
+     * @param force   If true, force-flushes everything accumulated in the {@link AttributeAggregator}, regardless of
+     *                the value returned by {@link #mustFlush()}.
```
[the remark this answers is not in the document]
andreipaduroiu: See above.

ManishKumarKeshri (approved):

sachin-j-joshi (approved):
Looks okay to me.

PR #5241, opened 2020-10-07
Issue 5240: (SegmentStore) Fixed a possible infinite loop in SegmentAggregator

### LINKED ISSUE #5240: Possible infinite loop in SegmentAggregator.deletedUnmergedSourceSegments
**Problem description**
`SegmentAggregator.deletedUnmergedSourceSegments` can enter an infinite loop if there is at least one non `MergeSegmentOperation` in the Segment Aggregator's list.

**Problem location**
SegmentAggregator.java.

Fix the infinite loop.

andreipaduroiu:
**Change log description**  
Fixed a bug in `SegmentAggregator.deleteUnmergedSourceSegments` where it may enter an infinite loop if the outstanding operation list contains at least one non-`MergeSegmentOperation` in it.

**Purpose of the change**  
Fixes #5240.

**What the code does**  
Reworked the `deleteUnmergedSourceSegments` to :
- Simply iterate through the entire list and pick the operations it cares about. 
- Do not try to alter this list. It will be cleared immediately after this method exits.

Context:
- `deleteUnmergedSourceSegments` is invoked when a Segment is deleted in order to delete any (transaction) segments that have recently been merged into it but whose merger has not yet been applied to LTS. 
- Not deleting this would leave a lot of garbage around in LTS in situations where transactions were merged and then the target segment is immediately deleted.

**How to verify it**  
Unit tests updated with this case. All tests must pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
@@ -43,7 +43,7 @@ import io.pravega.segmentstore.storage.Storage;
 import java.io.InputStream;
 import java.time.Duration;
 import java.util.ArrayDeque;
-import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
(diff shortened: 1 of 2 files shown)


sachin-j-joshi (approved):
Looks good to me.

Issue #5108, opened 2020-08-26
Pravega Consumption Based Retention


pbelgundi:
**Problem description**

PDP : https://github.com/pravega/pravega/wiki/PDP-47-(Pravega-Streams-Consumption-Based-Retention)

Parent ticket for the following issues:
- [x]  #5109: Add/Remove Subscriber Controller API
- [x]  #5110: Update Truncation Stream-Cut Controller API
- [x]  #5111: Truncate a Stream with CBR Policy - Controller 
- [x]  #5112: Client API to Create Subscriber reader groups
- [x]  #5114: Client API to UpdateTruncationStreamCut to Controller
- [x]  #5203 : Stream-Cuts get saved as "Checkpoints" in RGState.
- [x]  #5324 : addSubscriber and deleteSubscriber APIs on Controller should support generations
- [ ]  #5350 : User documentation for Consumption based retention.
- [x]  #5356 : Make CBR an implicit retention type for a Stream based on Reader Groups publishing Stream-Cuts.
- [x]  #5384 : Add Create, Update and Remove Reader Group APIs on Controller.
- [ ]  #5553:  System Tests for validating CBR functionality.

**Problem location**
Client, Controller, System Test Framework, Documentation

**Suggestions for an improvement**
Implement and fix all

RaulGracia:
@pbelgundi @shshashwat @bhargav-gulavani is this issue actually resolved? I remember that we have one system tests about consumption based retention, and I don't really know the state of documentation for this. It would be great to wrap up this one, as it has been there for some time and the feature is actually developed.

shshashwat:
All sub-issues fixed and merged with master

PR #5244, opened 2020-10-08
Issue 5203: Stream-Cuts get saved as "Checkpoints" in Reader Group State

### LINKED ISSUE #5203: Stream-Cuts get saved as "Checkpoints" in Reader Group State
**Problem description**
Stream-Cuts can be generated on a Stream by invoking the _generateStreamCut()_ API on a ReaderGroup.
Internally the mechanism used to generate a Stream-Cut as same as the mechanism for generating a checkpoint. 
However, a "Stream-Cut" is conceptually a different thing from a checkpoint.
Stream-Cuts generated by _generateStreamCut()_ API should be returned to the user and there is no need to store them as a checkpoint in the RGState
This is what happens internally today as we don't distinguish between Stream-Cuts and Checkpoints.
This would interfere with the use of checkpoints for truncation when implementing Consumption Based Retention on Client.

**Problem location**
Client Code:
https://github.com/pravega/pravega/blob/6acd317344424bbc92c3f870d394f68d93ce1085/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java#L373
https://github.com/pravega/pravega/blob/e3d71202d057be75fbcca27cbef1963195626d60/client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
https://github.com/pravega/pravega/blob/ca0cbdd6622eaebfc309e03cea489a563880bcba/client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java#L258

Stream-Cut generation should not have any impact on Checkpoint generation and generated Stream-Cuts should not be stored as checkpoints in RGState.

anirudhkovuru:
**Change log description**  
Stream-cuts/silent checkpoints are handled differently after all the readers have persisted their positions. 

**Purpose of the change**  
Fixes #5203. 

**What the code does**  
Added a new field within `CheckpointState` called `lastStreamCutPosition`.

- During checkpointing, the call made to `readerCheckpointed` in the `CheckpointState` allows for the reader to persist its positions. On the last reader's call, the field `lastCheckpointPositions` is updated to the positions of the completed checkpoint.
- This is done for silent checkpoints/stream-cuts also which will cause issues when implementing CBR (#5108)
- So, `lastStreamCutPosition` will hold the positions of the completed stream-cut instead so as to not mix stream-cuts with checkpoints.

**How to verify it**  


client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -246,6 +255,7 @@ private void read00(RevisionDataInput input, CheckpointStateBuilder builder) thr
             builder.uncheckpointedHosts(input.readMap(stringDeserializer, in -> in.readCollection(stringDeserializer, ArrayList::new)));
             builder.checkpointPositions(input.readMap(stringDeserializer, in -> in.readMap(segmentDeserializer, longDeserializer)));
             builder.lastCheckpointPosition(input.readMap(segmentDeserializer, longDeserializer));
+            builder.lastStreamCutPosition(input.readMap(segmentDeserializer, longDeserializer));
```
shrids, line 258: we should be backward compatible here, this line would throw exceptions when trying to read from data from an older client.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -256,6 +266,7 @@ private void write00(CheckpointState object, RevisionDataOutput output) throws I
             output.writeMap(object.uncheckpointedHosts, stringSerializer, (out, hosts) -> out.writeCollection(hosts, stringSerializer));
             output.writeMap(object.checkpointPositions, stringSerializer, (out, map) -> out.writeMap(map, segmentSerializer, longSerializer));
             output.writeMap(object.lastCheckpointPosition, segmentSerializer, longSerializer);
+            output.writeMap(object.lastStreamCutPosition, segmentSerializer, longSerializer);
```
shrids, line 269: Please add tests to verify this change...

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -64,23 +64,26 @@
      */
     private final Map<String, Map<Segment, Long>> checkpointPositions;
 
+    private Map<Segment, Long> lastStreamCutPosition;
```
tkaitchuck, line 67: This is not being used anywhere.

tkaitchuck (changes requested):
Please make the change backward and forward compatible with existing serialized data. We need to be able to inter-operate with mixed versions to support rolling upgrades and rolling downgrades.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
anirudhkovuru: This field will store the positions of the last stream-cut/silent checkpoint so that it does not interfere with `lastCheckpointPosition` where the positions of the actual last checkpoint will be stored. So, instead of having one field store the value of both which was the case before, we have two different fields `lastStreamCutPosition` and `lastCheckpointPosition`.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
pbelgundi: + Please add UTs and integration tests for the new code...

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
anirudhkovuru: I've added a new revision and made changes to `read00` ensure backward compatibility. I've also added a test for the same in `SerializationTest`.

anirudhkovuru: The field has been removed so there shouldn't be an issue in regards to backward compatibility.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
anirudhkovuru: Added two tests for the new code. Please have a look

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
anirudhkovuru: I've removed the field and added two tests to verify the change

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -138,8 +138,12 @@ void readerCheckpointed(String checkpointId, String readerName, Map<Segment, Lon
             positions.putAll(position);
             if (readers.isEmpty()) {
                 uncheckpointedHosts.remove(checkpointId);
-                //checkpoint operation completed for all readers, update the last checkpoint position.
-                lastCheckpointPosition = checkpointPositions.get(checkpointId);
+
```
pbelgundi, line 141: nit: remove unnecessary newline addition

pbelgundi (approved):
LGTM

shrids (approved):
LGTM.

tkaitchuck (approved):

PR #5248, opened 2020-10-09
Issue 4775: (R0.8) Cherry-pick #4776

### LINKED ISSUE #4775: Add snapshot stage for all release branches

crazyzhou:

**Change log description**  
Cherrypick #4776 into 0.8

- Issue 4775: Add snapshot stage for all release branches

**Purpose of the change**  
Fixes #4775 
This will benefit to have `0.8.1-SNAPSHOT` artifacts.

**What the code does**  
Cherrypick #4776 into 0.8

**How to verify it** 
Build passes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
       script: ./gradlew javadocs
 
     - stage: snapshot
-      if: branch = master AND NOT (type = pull_request)
+      if: (branch = master OR branch =~ /^r[0-9]+\.[0-9]+/) AND NOT (type = pull_request)
       env:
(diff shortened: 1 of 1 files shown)


crazyzhou:
Note this change violates the [current release process](https://github.com/pravega/pravega/wiki/How-to-release) , because we now remove `-SNAPSHOT` from the version early in the release cycle and the travis build will fail on the publish stage with a non-SNAPSHOT version.

For the new release process, I would suggest we follow other Apache projects such as Apache zookeeper. Like this: https://github.com/apache/zookeeper/commit/803c7f1a12f85978cb049af5e4ef23bd8b688715
We only remove `-SNAPSHOT` from the version each time we start a release candidate or a formal release, and this commit will just be tagged but not go into the release branch. Only in this case, we can guarantee this change will not break the travis build.

medvedevigorek:
I don't think we should take this change as is, we know it's going to break Travis builds once we enter RC phase. And looks to me that adjusting the release process alone may not be the desired solution to that. I've started a discussion in [#release-process](https://pravega-io.slack.com/archives/C01CFQ9556F ) Slack channel., let's see if we can find a solution.

ravisharda (approved):

ravisharda:
@medvedevigorek regarding your comments below, with the recent discussion in Slack are you OK with having this change in now? Please confirm, so that we can unblock this PR. 

> I don't think we should take this change as is, we know it's going to break Travis builds once we enter RC phase. And looks to me that adjusting the release process alone may not be the desired solution to that. I've started a discussion in #release-process Slack channel., let's see if we can find a solution.

crazyzhou:
Close this to cherrypick the correct commit and keep a cleaner history

PR #5250, opened 2020-10-13
Issue 5221: Port CLI security changes from Pravega Tools to Pravega core

### LINKED ISSUE #5221: Port CLI security changes from Pravega Tools to Pravega core
**Problem description**
PR https://github.com/pravega/pravega-tools/pull/75 added the necessary changes to the Pravega CLI to work with security enabled. Once we merge PR #5065, we need to port these changes to the new CLI location. Also, we need to extend the security functionality to the new User CLI.

**Problem location**
Pravega CLI.

Port security changes to Admin CLI and extend them to User CLI.

anirudhkovuru:
**Change log description**  
Porting the changes made in the pravega tools repo ([PR 75](https://github.com/pravega/pravega-tools/pull/75)) to allow for the admin CLI to function in security enabled environments

**Purpose of the change**  
Closes #5221 

**What the code does**  
- Uses new properties from the `admin-cli.properties` file to determine if the command needs to be run in TLS mode or not.
- The commands now when running in a TLS enabled environment have a trust-store which is used to verify and validate the server.
    - The protocol of the URI's, i.e., http or https , is determined based on the properties in the properties file.
    - The location of the trust-store file can also be mentioned as part of the properties in the properties file.
- Added tests for the above functionality.

**How to verify it**  


RaulGracia:
@anirudhkovuru please review the checkstyle, it is making the build to fail ` Task :cli:admin:checkstyleTest FAILED`

RaulGracia:
@anirudhkovuru please, fix the checkstyle check, it is failing now:
```
> Task :cli:admin:checkstyleTest FAILED
[ant:checkstyle] [ERROR] /home/travis/build/pravega/pravega/cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java:31:5: Variable access definition in wrong order. [DeclarationOrder]
```

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -50,7 +61,39 @@ protected Context createContext() {
...
+            KeyStore ks = null;
+            InputStream trustStore = null;
+            try {
+                trustStore = new FileInputStream(new File(getCLIControllerConfig().getTruststore()));
+                ks = KeyStore.getInstance("JKS");
+                ks.load(trustStore, null);
+                trustStore.close();
+
+            } catch (KeyStoreException e) {
+                output("The keystore file is invalid, the keystore type is not supported: " + e.toString());
```
andreipaduroiu, line 78: `output` supports formatted text. Please use that api.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -50,7 +61,39 @@ protected Context createContext() {
...
-        Client client = ClientBuilder.newClient(clientConfig);
+
+        Client client;
+
+        // If tls parameters are configured, set them in client
+        if (getCLIControllerConfig().isTlsEnabled()) {
+            KeyStore ks = null;
+            InputStream trustStore = null;
+            try {
+                trustStore = new FileInputStream(new File(getCLIControllerConfig().getTruststore()));
```
andreipaduroiu, line 72: I suggest using try-with-resources or annotating with with `@Cleanup` to ensure this is closed. If you get an error between here and line 75, your input stream will not be closed.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AbstractTlsAdminCommandTest {
+
+    // Security related flags and instantiate local pravega server.
+    protected static final AtomicReference<AdminCommandState> STATE = new AtomicReference<>();
+
+    private static final Integer CONTROLLER_PORT = 9090;
```
andreipaduroiu, line 31: Use `TestUtils` to get randomized ports. Hardcoding ports in unit tests is a recipe for failure due to ports already being used.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final Integer SEGMENT_STORE_PORT = 6000;
+    private static final Integer REST_SERVER_PORT = 9091;
+
+    @Rule
+    public final Timeout globalTimeout = new Timeout(80, TimeUnit.SECONDS);
+
+    protected boolean authEnabled = false;
```
andreipaduroiu, line 38: Any reason why these shouldn't be `static` or `final`?

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.net.URI;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AbstractTlsAdminCommandTest {
+
+    // Security related flags and instantiate local pravega server.
+    protected static final AtomicReference<AdminCommandState> STATE = new AtomicReference<>();
```
andreipaduroiu, line 29: Why is this static? 

Do not use static fields that you set during your execution. This will prevent us from eventually running tests in parallel.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
```diff
@@ -0,0 +1,122 @@
...
+        pravegaProperties.setProperty("cli.security.auth.credentials.password", SecurityConfigDefaults.AUTH_ADMIN_PASSWORD);
+        pravegaProperties.setProperty("cli.security.tls.enable", Boolean.toString(tlsEnabled));
+        pravegaProperties.setProperty("cli.security.tls.trustStore.location", "../../config/" + SecurityConfigDefaults.TLS_CLIENT_TRUSTSTORE_NAME);
+
+        STATE.get().getConfigBuilder().include(pravegaProperties);
+
+        localPravega.start();
+
+        // Wait for the server to complete start-up.
+        TimeUnit.SECONDS.sleep(20);
```
andreipaduroiu, line 99: Do not do this. Find a different way to figure out when the server has started.

20 seconds may be too long in most cases and may not be sufficient if Travis is super-slow when your test executes.

cli/admin/src/test/java/io/pravega/cli/admin/controller/AuthEnabledControllerCommandsTest.java
```diff
@@ -0,0 +1,252 @@
...
+        Parser.Command pc = Parser.parse(inputCommand);
+        Assert.assertNotNull(pc.toString());
+        CommandArgs args = new CommandArgs(pc.getArgs(), state);
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        TestingDescribeStreamCommand cmd = new TestingDescribeStreamCommand(args);
+        try (PrintStream ps = new PrintStream(baos, true, "UTF-8")) {
+            cmd.setOut(ps);
+            cmd.execute();
+        }
+        return new String(baos.toByteArray(), StandardCharsets.UTF_8);
```
andreipaduroiu, line 197: Please be consistent in how you reference UTF8. Here you use `StandardCharset`, yet 4 lines above you use a string.

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.net.URI;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AbstractTlsUserCommandTest {
+
+    // Security related flags and instantiate local pravega server.
+    protected static final AtomicReference<InteractiveConfig> CONFIG = new AtomicReference<>();
+
+    private static final Integer CONTROLLER_PORT = 9090;
```
andreipaduroiu, line 31: randomize

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.rules.Timeout;
+
+import java.net.URI;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AbstractTlsUserCommandTest {
+
+    // Security related flags and instantiate local pravega server.
+    protected static final AtomicReference<InteractiveConfig> CONFIG = new AtomicReference<>();
```
andreipaduroiu, line 29: same comment about static shared variables.

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static final Integer CONTROLLER_PORT = 9090;
+    private static final Integer SEGMENT_STORE_PORT = 6000;
+    private static final Integer REST_SERVER_PORT = 9091;
+
+    @Rule
+    public final Timeout globalTimeout = new Timeout(120, TimeUnit.SECONDS);
+
+    protected boolean authEnabled = false;
```
andreipaduroiu, line 38: final

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+                    .jksKeyFile("../../config/" + SecurityConfigDefaults.TLS_SERVER_KEYSTORE_NAME)
+                    .jksTrustFile("../../config/" + SecurityConfigDefaults.TLS_CLIENT_TRUSTSTORE_NAME)
+                    .keyPasswordFile("../../config/" + SecurityConfigDefaults.TLS_PASSWORD_FILE_NAME);
+        }
+
+        localPravega = emulatorBuilder.build();
+        localPravega.start();
+
+        // Wait for the server to complete start-up.
+        TimeUnit.SECONDS.sleep(20);
```
andreipaduroiu, line 79: No sleeping please.

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * Utility functions for creating the test setup.
+ */
+@Slf4j
+@NotThreadSafe
+public final class SecureSetupUtils {
+
+    // The different services.
+    @Getter
```
andreipaduroiu, line 63: Make all these fields `final`. Initialize them in the class constructor, and start the services only when you invoke `startAllServices`

andreipaduroiu: Also make this class implement `AutoCloseable` and in the `close` method make sure you shut down and close all these services. Otherwise you risk leaving them running in the background when your test is done.

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
```diff
@@ -0,0 +1,249 @@
...
+        this.controller = new ControllerImpl(ControllerImplConfig.builder().clientConfig(clientConfig).build(),
+                executor);
+        this.clientFactory = new ClientFactoryImpl(scope, controller, clientConfig);
+
+        // Start zookeeper.
+        this.zkTestServer = new TestingServerStarter().start();
+        this.zkTestServer.start();
+
+        // Start Pravega Service.
+        ServiceBuilder serviceBuilder = ServiceBuilder.newInMemoryBuilder(ServiceBuilderConfig.getDefaultConfig());
```
andreipaduroiu, line 138: You never shut this down. You need to keep a pointer to it and close it in your `close` method.

andreipaduroiu (changes requested):

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
anirudhkovuru: Used the api for formatted text.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
anirudhkovuru: Used `@Cleanup` on the `InputStream` as suggested.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
anirudhkovuru: Used `TestUtils` for randomized ports.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
anirudhkovuru: `@Rule` requires the variable to be `final` only and not `static`. The variables `authEnabled` and `tlsEnabled` cannot be `final` as these flags are set before calling `super.setUp()` in the test classes based on whether we test with authentication or TLS enabled or both.
Also static non final variables cannot be declared according to the checkstyle.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
anirudhkovuru: I've made this field non-static as suggested.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
anirudhkovuru: Will try to look for a better way to identify if the server has started.

cli/admin/src/test/java/io/pravega/cli/admin/controller/AuthEnabledControllerCommandsTest.java
anirudhkovuru: The references are consistent now.

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
anirudhkovuru: done

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
anirudhkovuru: made non-static

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
anirudhkovuru: `authEnabled` and `tlsEnabled` cannot be `final` as they will be changed according to the tests in which they are used. Explained above

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
anirudhkovuru: Will look for another way.

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
anirudhkovuru: The class implements `AutoCloseable` and the `close` method. Will look into making the fields final.

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
anirudhkovuru: I've added this as a field and close it in the `close` method

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+import java.net.URI;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AbstractTlsAdminCommandTest {
+
+    // Security related flags and instantiate local pravega server.
+    private static final Integer CONTROLLER_PORT = TestUtils.getAvailableListenPort();
```
andreipaduroiu, line 30: Do not make these static. Keeping state (available ports is a form of state) in static variables prevents these tests from being run in parallel.

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
andreipaduroiu: Ping! Please make the fields final. If you only set them in the constructor, then `final` is how they should be.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractAdminCommandTest.java
```diff
@@ -22,28 +23,34 @@
 public abstract class AbstractAdminCommandTest {
 
     // Setup utility.
-    protected static final SetupUtils SETUP_UTILS = new SetupUtils();
-    protected static final AtomicReference<AdminCommandState> STATE = new AtomicReference<>();
+    protected static final SecureSetupUtils SETUP_UTILS = new SecureSetupUtils();
```
andreipaduroiu, line 26: Same here. Do not make this thing static.

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.Rule;
+import org.junit.rules.Timeout;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AbstractTlsUserCommandTest {
+
+    // Security related flags and instantiate local pravega server.
+    private static final Integer CONTROLLER_PORT = TestUtils.getAvailableListenPort();
```
andreipaduroiu, line 27: not static

cli/user/src/test/java/io/pravega/cli/user/AbstractUserCommandTest.java
```diff
@@ -22,26 +23,32 @@
 public abstract class AbstractUserCommandTest {
 
     // Setup utility.
-    protected static final SetupUtils SETUP_UTILS = new SetupUtils();
-    protected static final AtomicReference<InteractiveConfig> CONFIG = new AtomicReference<>();
+    protected static final SecureSetupUtils SETUP_UTILS = new SecureSetupUtils();
```
andreipaduroiu, line 26: not static

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
```diff
@@ -0,0 +1,238 @@
...
+            return;
+        }
+
+        this.controllerWrapper.close();
+        this.server.close();
+        this.zkTestServer.close();
+        this.serviceBuilder.close();
+        this.clientFactory.close();
+        this.controller.close();
+        this.executor.shutdown();
```
andreipaduroiu, line 169: This is the wrong way to shut down an executor. Use `ExecutorServiceHelpers.shutdown(...)`

andreipaduroiu (changes requested):

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
anirudhkovuru: Using the method mentioned above.

cli/user/src/test/java/io/pravega/cli/user/AbstractUserCommandTest.java
anirudhkovuru: same

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
anirudhkovuru: same

cli/admin/src/test/java/io/pravega/cli/admin/AbstractAdminCommandTest.java
anirudhkovuru: Made it non static.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
anirudhkovuru: Made these non static.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
anirudhkovuru: This has been removed.

cli/user/src/test/java/io/pravega/cli/user/AbstractTlsUserCommandTest.java
anirudhkovuru: Removed this.

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
anirudhkovuru: I've made all these fields final as suggested.

cli/user/src/test/java/io/pravega/cli/user/SecureUserCommandsTest.java
```diff
@@ -0,0 +1,137 @@
...
+        Assert.assertTrue("StreamAppendCommand failed.", commandResult.contains("Done"));
+        commandResult = TestUtils.executeCommand("stream read " + stream + " 5", config.get());
+        Assert.assertTrue("StreamReadCommand failed.", commandResult.contains("Done"));
+        Assert.assertNotNull(StreamCommand.Read.descriptor());
+
+        commandResult = TestUtils.executeCommand("stream delete " + stream, config.get());
+        Assert.assertTrue("StreamDeleteCommand failed.", commandResult.contains("deleted successfully"));
+        Assert.assertNotNull(StreamCommand.Delete.descriptor());
+
+        // TODO: Test KVT commands in the secure scenario (auth+TLS).
```
andreipaduroiu, line 77: Are you going to resolve this TODO?

What about the commented out code below?

RaulGracia: @anirudhkovuru please, review the guidelines to add TODOs in the code: https://github.com/pravega/pravega/wiki/Contributing#todos, which requires to add a reference to the issue the TODO is related to.
@andreipaduroiu  it looks like this test cannot work today due to a bug that ahs been found: https://github.com/pravega/pravega/issues/5374. Until the bug is fix, this test cannot work combining Auth+TLS.

anirudhkovuru: I've removed the commented code and mentioned only the issue.

RaulGracia (changes requested):
This PR is progressing well. I have some comments to address, though.

cli/admin/src/main/java/io/pravega/cli/admin/utils/ControllerHostnameVerifier.java
```diff
@@ -0,0 +1,20 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.cli.admin.utils;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLSession;
+
+public class ControllerHostnameVerifier implements HostnameVerifier {
```
ravisharda, line 15: Curious, why do you need a custom hostname verifier here? You could use one of the standard ones. 

Here's how we do it in the client (for the call to segment store): 

https://github.com/pravega/pravega/blob/899a82d5c4055007311adf01bf58d1629f4e2f6e/client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java#L243-L251

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+                .enableAuth(authEnabled)
+                .enableTls(tlsEnabled);
+
+        // Since the server is being built right here, avoiding delegating these conditions to subclasses via factory
+        // methods. This is so that it is easy to see the difference in server configs all in one place. This is also
+        // unlike the ClientConfig preparation which is being delegated to factory methods to make their preparation
+        // explicit in the respective test classes.
+
+        if (authEnabled) {
+            emulatorBuilder.passwdFile("../../config/" + SecurityConfigDefaults.AUTH_HANDLER_INPUT_FILE_NAME)
```
ravisharda, line 62: The String "../../config" gets repeated many times below. Create a String constant and use it here and below.

cli/admin/src/test/java/io/pravega/cli/admin/controller/AuthEnabledControllerCommandsTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    @Test
+    public void testDescribeScopeCommand() throws Exception {
+        String commandResult = TestUtils.executeCommand("controller describe-scope _system", state.get());
+        Assert.assertTrue(commandResult.contains("_system"));
+        Assert.assertNotNull(ControllerDescribeStreamCommand.descriptor());
+    }
+
+    @Test
+    public void testListStreamsCommand() throws Exception {
```
ravisharda, line 87: In general, you should avoid throwing the base `Exception`. If you must throw, throw the specific exceptions. Here, in the tests, throwing exceptions is meaningless, so you could simplify the code a bit by adding a `@SneakyThrows` exception, if you will.

ravisharda (changes requested):

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -50,7 +62,38 @@ protected Context createContext() {
...
+                    .withConfig(clientConfig)
+                    .trustStore(ks)
+                    .hostnameVerifier(controllerHostnameVerifier)
+                    .build();
+        } else {
+            client = ClientBuilder.newClient(clientConfig);
+        }
+
         // If authorization parameters are configured, set them in the client.
         if (getCLIControllerConfig().isAuthEnabled()) {
```
ravisharda, line 98: You seem to be using `getCLIControllerConfig()` multiple times. (The name of that method is misleading - it actually does a bit of processing as opposed to returning a cached value that one'd usually expect from a method having a name starting with `get`). You might want to call it once, store the value in a variable and reuse it.

cli/admin/src/test/java/io/pravega/cli/admin/controller/TLSEnabledControllerCommandsTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @After
+    @Override
+    public void tearDown() throws Exception {
+        super.tearDown();
+    }
+
+    @Test
+    public void testAllCommands() throws Exception {
```
ravisharda, line 47: This test is too ambitious. It tests multiple behaviors in a single test. If one behavior fails later if something is changed, it'll mask failures in other behaviors. Refactor this into multiple tests.

ravisharda (changes requested):

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -50,7 +62,38 @@ protected Context createContext() {
...
+
+        // If tls parameters are configured, set them in client
+        if (getCLIControllerConfig().isTlsEnabled()) {
+            KeyStore ks = null;
+            try {
+                @Cleanup
+                InputStream trustStore = new FileInputStream(new File(getCLIControllerConfig().getTruststore()));
+                ks = KeyStore.getInstance("JKS");
+                ks.load(trustStore, null);
+
```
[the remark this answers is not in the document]
anirudhkovuru: Using `createTrustStore` now.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerDescribeStreamCommand.java
```diff
@@ -72,7 +72,12 @@ public void execute() {
                 store = StreamStoreFactory.createZKStore(zkClient, executor);
             } else {
                 segmentHelper = instantiateSegmentHelper(zkClient);
-                GrpcAuthHelper authHelper = GrpcAuthHelper.getDisabledAuthHelper();
+                GrpcAuthHelper authHelper;
+                if (getCLIControllerConfig().isAuthEnabled()) {
+                    authHelper = new GrpcAuthHelper(true, "secret", 300);
```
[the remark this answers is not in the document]
anirudhkovuru: `tokenSigningKey` can be subject to change. I've added a new property in the properties file for this.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Rule
+    public final Timeout globalTimeout = new Timeout(80, TimeUnit.SECONDS);
+
+    protected final AtomicReference<AdminCommandState> state = new AtomicReference<>();
+    protected boolean authEnabled = false;
+    protected boolean tlsEnabled = false;
+    LocalPravegaEmulator localPravega;
+
+    // Security related flags and instantiate local pravega server.
```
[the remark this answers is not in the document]
anirudhkovuru: done

cli/admin/src/test/java/io/pravega/cli/admin/controller/AuthEnabledControllerCommandsTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+public class AuthEnabledControllerCommandsTest extends AbstractAdminCommandTest {
+
+    @Before
+    @Override
+    public void setUp() throws Exception {
+        authEnabled = true;
+        super.setUp();
+    }
+
```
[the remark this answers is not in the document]
anirudhkovuru: done

cli/admin/src/test/java/io/pravega/cli/admin/controller/AuthEnabledControllerCommandsTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    @Test
+    public void testListScopesCommand() throws Exception {
```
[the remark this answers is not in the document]
anirudhkovuru: I've shifted the common tests to a single class so for example all the tests in `ControllerCommandsTest.java` are run by `AuthEnabledControllerCommandsTest.java` as it extends `ControllerCommandsTest.java`. Same changes have been made throughout admin and user cli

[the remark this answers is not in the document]
anirudhkovuru: Yeah ass explained above now these tests are in `ControllerCommandsTest`, this file extends it so the tests transfer too.

cli/admin/src/test/java/io/pravega/cli/admin/controller/AuthEnabledControllerCommandsTest.java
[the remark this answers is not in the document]
anirudhkovuru: done

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -50,7 +62,38 @@ protected Context createContext() {
...
+            } catch (IOException e) {
+                output("The keystore file is invalid, check if the file exists: %s", e.toString());
+            } catch (NoSuchAlgorithmException e) {
+                output("The keystore file is invalid, the keystore file might be in the wrong format: %s", e.toString());
+            } catch (CertificateException e) {
+                output("The keystore file is invalid, check if the certificates are valid: %s", e.toString());
+            }
+
+            HostnameVerifier controllerHostnameVerifier = new ControllerHostnameVerifier();
+            client = ClientBuilder.newBuilder()
```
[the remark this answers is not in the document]
anirudhkovuru: The credentials are registered in the next lines.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
anirudhkovuru: Moved the return of `getCLIControllerConfig()` to a single variable which is used everywhere.

RaulGracia (approved):
LGTM assuming that i) all the comments from reviewers are addressed, and ii) the last version of this PR is tested on a secure Pravega cluster (admin/user commands).

cli/admin/src/test/java/io/pravega/cli/admin/controller/AuthEnabledControllerCommandsTest.java
anirudhkovuru: Using `@SneakyThrows` now.

cli/admin/src/test/java/io/pravega/cli/admin/AbstractTlsAdminCommandTest.java
anirudhkovuru: Stored in a local variable.

cli/admin/src/main/java/io/pravega/cli/admin/utils/ControllerHostnameVerifier.java
anirudhkovuru: Instead of having a `HostnameVerfier`, using `SSLContext` as suggested.

cli/admin/src/main/java/io/pravega/cli/admin/utils/ControllerHostnameVerifier.java
```diff
@@ -0,0 +1,20 @@
...
+    @Override
+    public boolean verify(String s, SSLSession sslSession) {
+        return s.equalsIgnoreCase(sslSession.getPeerHost());
```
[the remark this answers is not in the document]
anirudhkovuru: This file is not used anymore as the connection to TLS is made through the use of `SSLContext` instead.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -50,7 +62,38 @@ protected Context createContext() {
...
+                output("The keystore file is invalid, the keystore type is not supported: %s", e.toString());
+            } catch (IOException e) {
+                output("The keystore file is invalid, check if the file exists: %s", e.toString());
+            } catch (NoSuchAlgorithmException e) {
+                output("The keystore file is invalid, the keystore file might be in the wrong format: %s", e.toString());
+            } catch (CertificateException e) {
+                output("The keystore file is invalid, check if the certificates are valid: %s", e.toString());
+            }
+
+            HostnameVerifier controllerHostnameVerifier = new ControllerHostnameVerifier();
```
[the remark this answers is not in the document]
anirudhkovuru: I've added a message. Printed the stack trace. I can't throw an exception as I do not want to change the method signature. Instead I return `null`. Is there a better way than this?

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -50,7 +62,38 @@ protected Context createContext() {
         ClientConfig clientConfig = new ClientConfig();
         clientConfig.register(JacksonJsonProvider.class);
         clientConfig.property("sun.net.http.allowRestrictedHeaders", "true");
-        Client client = ClientBuilder.newClient(clientConfig);
+
+        Client client;
+
+        // If tls parameters are configured, set them in client
```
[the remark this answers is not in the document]
anirudhkovuru: done

RaulGracia:
@anirudhkovuru maybe some recent change has broken the build again:
```
io.pravega.cli.admin.controller.AuthEnabledControllerCommandsTest > testListReaderGroupsCommand FAILED
    java.lang.IllegalStateException: Expected the service ControllerServiceStarter [FAILED] to be RUNNING, but the service has FAILED
        at com.google.common.util.concurrent.AbstractService.checkCurrentState(AbstractService.java:366)
...
io.pravega.cli.admin.controller.AuthEnabledControllerCommandsTest > testDescribeStreamCommand FAILED
    org.junit.runners.model.TestTimedOutException: test timed out after 60 seconds
        at java.base@11.0.2/jdk.internal.misc.Unsafe.park(Native Method)
...
io.pravega.cli.admin.controller.AuthEnabledControllerCommandsTest > testListScopesCommand FAILED
    org.junit.runners.model.TestTimedOutException: test timed out after 60 seconds
```
Apart from that, please @ravisharda and @andreipaduroiu please have another look to see if your comments have been addressed.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -47,14 +58,39 @@
...
+            SSLContext tlsContext;
+            try {
+                KeyStore ks = createTrustStore(config.getTruststore());
+                TrustManagerFactory tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+                tmf.init(ks);
+                tlsContext = SSLContext.getInstance("TLS");
+                tlsContext.init(null, tmf.getTrustManagers(), null);
+            } catch (KeyStoreException | IOException | NoSuchAlgorithmException | CertificateException | KeyManagementException e) {
+                output("Encountered exception while trying to use the given truststore: %s", config.getTruststore());
+                e.printStackTrace();
```
ravisharda, line 79: Please remove printStackTrace.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -47,14 +58,39 @@
...
+        if (config.isTlsEnabled()) {
+            SSLContext tlsContext;
+            try {
+                KeyStore ks = createTrustStore(config.getTruststore());
+                TrustManagerFactory tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+                tmf.init(ks);
+                tlsContext = SSLContext.getInstance("TLS");
+                tlsContext.init(null, tmf.getTrustManagers(), null);
+            } catch (KeyStoreException | IOException | NoSuchAlgorithmException | CertificateException | KeyManagementException e) {
+                output("Encountered exception while trying to use the given truststore: %s", config.getTruststore());
```
ravisharda, line 78: Without the additional context, it'd be very difficult to debug this issue. Please output the exception message/trace too. 

Curious, why don't we add some logs to these methods too?

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
```diff
@@ -47,14 +58,39 @@
...
+                    .build();
+        } else {
+            client = ClientBuilder.newClient(clientConfig);
+        }
+
         // If authorization parameters are configured, set them in the client.
-        if (getCLIControllerConfig().isAuthEnabled()) {
-            HttpAuthenticationFeature auth = HttpAuthenticationFeature.basic(getCLIControllerConfig().getUserName(),
-                    getCLIControllerConfig().getPassword());
+        if (config.isAuthEnabled()) {
```
ravisharda, line 91: nit: The logic here'll be easier to follow with a slight rearrangement. Something like:

```
builder = ClientBuilder.newBuilder().withConfig(clientConfig); 
if (tls enabled) {
     builder.sslContext(...)
} 
Client client = builder.build(); 

if (auth enabled) {
     // do auth-specific stuff
}
Client client = builder.build(); 

```

cli/admin/src/main/java/io/pravega/cli/admin/utils/CLIControllerConfig.java
```diff
@@ -24,11 +24,14 @@
         SEGMENTSTORE, ZOOKEEPER
     }
 
-    private static final Property<String> CONTROLLER_REST_URI = Property.named("controller.rest.uri", "http://localhost:9091");
-    private static final Property<String> CONTROLLER_GRPC_URI = Property.named("controller.grpc.uri", "tcp://localhost:9090");
+    private static final Property<String> CONTROLLER_REST_URI = Property.named("controller.rest.uri", "localhost:9091");
```
ravisharda, line 27: Some of these config names are inconsistent with the scheme we use elsewhere in Pravega. See the following: 
* [Configuration Naming Guidelines](https://github.com/pravega/pravega/wiki/Configuration-Naming-Guidelines)
* [Existing examples](https://github.com/pravega/pravega/wiki/Configuration-Name-Changes) 

Here are the suggested changes: 

* `controller.rest.uri` -> `controller.connect.rest.uri` (because it is connecting to the Controller's REST URI)
* `controller.grpc.uri` -> `controller.connect.grpc.uri`
* `security.auth.credentials.username` -> `controller.connect.credentials.username`
* `security.auth.credentials.password` -> `controller.connect.credentials.pwd` (we avoid the term `password` so that scanners and other tools don't get confused)
* `security.auth.token.signingKey` -- ? (what is this for, here?)

Etc.

As for `security.auth.enable` (and similar items), it indicates you are enabling auth for CLI. If you are using it for specifying that auth be enabled for CLI to controller communication, the config name'd be different.

ravisharda (changes requested):

cli/user/src/main/java/io/pravega/cli/user/stream/StreamCommand.java
```diff
@@ -65,7 +63,7 @@ public Create(@NonNull CommandArgs commandArgs) {
         public void execute() {
             ensureMinArgCount(1);
             @Cleanup
-            val sm = StreamManager.create(URI.create(getConfig().getControllerUri()));
+            val sm = StreamManager.create(getClientConfig());
```
ravisharda, line 66: Please avoid using abbreviations like this one. A `streamMgr` or `streamManager`'d help reduce the cognitive burden on the person who reads the code. Same for `sc` below.

cli/user/src/main/java/io/pravega/cli/user/stream/StreamCommand.java
```diff
@@ -247,7 +245,7 @@ public void execute() {
             val readerGroup = UUID.randomUUID().toString().replace("-", "");
             val readerId = UUID.randomUUID().toString().replace("-", "");
 
-            val cc = ClientConfig.builder().controllerURI(getControllerUri()).build();
+            val cc = getClientConfig();
```
ravisharda, line 248: same here.

cli/user/src/main/java/io/pravega/cli/user/stream/StreamCommand.java
```diff
@@ -183,7 +181,7 @@ public void execute() throws Exception {
             }
 
             @Cleanup
-            val factory = EventStreamClientFactory.withScope(scopedStream.getScope(), ClientConfig.builder().controllerURI(getControllerUri()).build());
+            val factory = EventStreamClientFactory.withScope(scopedStream.getScope(), getClientConfig());
```
ravisharda, line 184: nit: The term `factory` is a bit abstract. A `clientFactory` is more precise, and `eventStreamClientFactory` even more so. In general, variable names should be as precise as possible (for reasons like reducing the work that a code reader needs to perform so that the person may focus more on the logic than the syntax). Here, it's not so much an issue as the `factory` is used right under this statement, so you may choose to leave it as-is.

ravisharda (changes requested):

cli/user/src/main/java/io/pravega/cli/user/stream/StreamCommand.java
anirudhkovuru: done

cli/user/src/main/java/io/pravega/cli/user/stream/StreamCommand.java
anirudhkovuru: done

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
anirudhkovuru: Following a pattern like this

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
anirudhkovuru: I print the error message.

cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerCommand.java
anirudhkovuru: Removed

cli/admin/src/test/java/io/pravega/cli/admin/controller/TLSEnabledControllerCommandsTest.java
anirudhkovuru: Resolved all ambitious tests

cli/admin/src/main/java/io/pravega/cli/admin/utils/CLIControllerConfig.java
anirudhkovuru: I've used the guidelines.

cli/user/src/test/java/io/pravega/cli/user/kvs/KVTCommandsTest.java
```diff
@@ -22,7 +23,8 @@
 public class KVTCommandsTest extends AbstractUserCommandTest {
 
     @Test(timeout = 10000)
-    public void testCreateKVT() throws Exception {
+    @SneakyThrows
+    public void testCreateKVT() {
```
ravisharda, line 27: nit: Please avoid abbreviations to the extent possible. `testCreateKeyValueTable()` sounds better. Same for other tests below.

config/admin-cli.properties
```diff
@@ -9,13 +9,16 @@
...
 
 # Config for CLI credentials (at the moment, Controller auth is only supported).
-cli.security.auth.enable=false
-cli.security.auth.credentials.username=admin
-cli.security.auth.credentials.password=1111_aaaa
+cli.controller.connect.channel.auth=false
+cli.controller.connect.channel.tls=false
+cli.controller.connect.credentials.username=admin
+cli.controller.connect.credentials.pwd=1111_aaaa
+cli.controller.connect.delegationToken.signingKey.basis=secret
```
ravisharda, line 16: What is this used for in CLI?

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.UUID;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Utility functions for creating the test setup.
+ */
+@Slf4j
+@NotThreadSafe
+public final class SecureSetupUtils implements AutoCloseable {
```
ravisharda, line 56: Please use the `ClusterWrapper` instead, as it already supports running clusters both with security (TLS and auth) on and off.

ravisharda (changes requested):

config/admin-cli.properties
anirudhkovuru: When we run the CLI against the standalone in auth scenario, we use this token to set up a `StreamMetadataStore`.
[In these lines.](https://github.com/anirudhkovuru/pravega/blob/ac8067b0e14bed7dffab81a2c11032d6e3705e0a/cli/admin/src/main/java/io/pravega/cli/admin/controller/ControllerDescribeStreamCommand.java#L69-L82)
We generate an `authHelper` using the token to make the store.

cli/user/src/test/java/io/pravega/cli/user/kvs/KVTCommandsTest.java
anirudhkovuru: done

cli/user/src/main/java/io/pravega/cli/user/stream/StreamCommand.java
anirudhkovuru: done

test/integration/src/main/java/io/pravega/test/integration/utils/SecureSetupUtils.java
anirudhkovuru: Using the `ClusterWrapper` now.

anirudhkovuru:
I have updated a lot of the earlier tests to now use `ClusterWrapper`. I've also made sure to streamline the code make sure there is no code duplication.

RaulGracia:
@anirudhkovuru noticed that this test failed in the previous build:
```
io.pravega.cli.user.kvs.AuthEnabledKVTCommandsTest$SecureKVTCommandsTest > testDeleteKVT FAILED
    org.junit.runners.model.TestTimedOutException: test timed out after 20000 milliseconds
        at java.base@11.0.2/jdk.internal.misc.Unsafe.park(Native Method)
```

cli/user/src/test/java/io/pravega/cli/user/AbstractUserCommandTest.java
```diff
@@ -19,29 +20,33 @@
...
 import java.util.concurrent.atomic.AtomicReference;
 
+import static io.pravega.cli.user.TestUtils.createPravegaCluster;
+import static io.pravega.cli.user.TestUtils.setInteractiveConfig;
+
 public abstract class AbstractUserCommandTest {
 
-    // Setup utility.
-    protected static final SetupUtils SETUP_UTILS = new SetupUtils();
+    protected static final AtomicReference<ClusterWrapper> CLUSTER = new AtomicReference<>();
```
ravisharda, line 28: Why are you using `AtomicReference` for `CLUSTER` and `CONFIG`?

Since this `CLUSTER` variable is static, it will be shared across all classes that use it. This will cause a synchronization issue across different classes that use this or call `setUpCluster()`. You should instantiate this in the respective classes instead so that you don't prevent tests from different classes from running parallelly. Or, alternatively, you can have the `setUpCluster()` return a new instance.

cli/user/src/test/java/io/pravega/cli/user/AbstractUserCommandTest.java
```diff
@@ -19,29 +20,33 @@
...
     @BeforeClass
-    public static void setUp() throws Exception {
-        SETUP_UTILS.startAllServices();
-        InteractiveConfig interactiveConfig = InteractiveConfig.getDefault();
-        interactiveConfig.setControllerUri(SETUP_UTILS.getControllerUri().toString());
-        interactiveConfig.setDefaultSegmentCount(4);
-        interactiveConfig.setMaxListItems(100);
-        interactiveConfig.setTimeoutMillis(1000);
-        CONFIG.set(interactiveConfig);
+    public static void start() {
```
ravisharda, line 41: See comment earlier.

cli/user/src/test/java/io/pravega/cli/user/TestUtils.java
```diff
@@ -35,11 +38,60 @@ public static String executeCommand(String inputCommand, InteractiveConfig confi
         CommandArgs args = new CommandArgs(pc.getArgs().getArgs(), config);
         Command cmd = Command.Factory.get(pc.getComponent(), pc.getName(), args);
         final ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        try (PrintStream ps = new PrintStream(baos, true, "UTF-8")) {
+        try (PrintStream ps = new PrintStream(baos, true, StandardCharsets.UTF_8)) {
             assert cmd != null;
```
ravisharda, line 42: This assertion of assumption should be done right after the `Command` is initialized.

cli/user/src/test/java/io/pravega/cli/user/TestUtils.java
```diff
@@ -35,11 +38,60 @@ public static String executeCommand(String inputCommand, InteractiveConfig confi
...
             cmd.setOut(ps);
             cmd.execute();
         }
         return new String(baos.toByteArray(), StandardCharsets.UTF_8);
```
ravisharda, line 46: nit: this'd be shorter: `baos.toString(StandardCharsets.UTF_8)`

cli/user/src/test/java/io/pravega/cli/user/TestUtils.java
```diff
@@ -35,11 +38,60 @@ public static String executeCommand(String inputCommand, InteractiveConfig confi
...
+     *
+     * @param authEnabled whether accessing the cluster require authentication or not.
+     * @param tlsEnabled whether accessing the cluster require TLS or not.
+     * @return A local Pravega cluster
+     */
+    public static ClusterWrapper createPravegaCluster(boolean authEnabled, boolean tlsEnabled) {
+        return ClusterWrapper.builder()
+                .authEnabled(authEnabled)
+                .tlsEnabled(tlsEnabled)
+                .tlsServerCertificatePath(pathToConfig() + SecurityConfigDefaults.TLS_SERVER_CERT_FILE_NAME)
```
ravisharda, line 69: This and the next two lines are unnecessary if TLS is disabled. Why don't you call them only if TLS is enabled?

config/admin-cli.properties
ravisharda: I still don't understand it. `signing key` is a server secret, and it is only supposed to be used across the Controller and the Segment Store. Could you please help me understand the intent better?

ravisharda (changes requested):

cli/user/src/test/java/io/pravega/cli/user/TestUtils.java
anirudhkovuru: Rearranged accordingly.

cli/user/src/test/java/io/pravega/cli/user/AbstractUserCommandTest.java
anirudhkovuru: As discussed, avoiding the use of `AtomicReference`.

ravisharda (approved):

config/admin-cli.properties
ravisharda: The signing key was removed, so this issue is resolved.

PR #5252, opened 2020-10-13
Issue 5162: (SegmentStore) Table Segment Throttling and bug fixes

### LINKED ISSUE #5162: Table Segment Throttling
**Problem description**
We need a way to throttle external Table Segment requests. Currently there is no way to protect against a flood of requests which can overwhelm a Segment Store.

**Problem location**
Segment Store, Client.

One option is to do what #3700 wanted to do (there's a branch for it as well).

Another option is along these lines:
- Do not throttle internal segment store requests. Only throttle requests that come in via `PravegaRequestProcessor` (this includes KVTs and Controller metadata table segments).
    - Note that the Controller does not use `TableSegmentImpl` - whatever we do over the wire will likely need to be handled in 2 places (until #4647 is implemented).
- Determine a maximum outstanding request count. Perhaps use a counter equivalent to how we do Appends (in AppendProcessor).
    - This should be both on a per-segment basis (maximum indexing backlog) and per-process basis (all segments).
- If we exceed this request count threshold, pause the connection (if possible) or send back an error message indicating the server is busy.
    - If the latter (server busy), we need handling on the client to retry.
    - If the former (connection pause), we may need handling on the client for this.
- We need to stop accepting new requests on the Segment Store, otherwise, if we overwhelm the ingestion pipeline (and indexer), those will kick back using their own throttling, but we will end up accumulating new requests in memory, thus leading to OOMs.

### LINKED ISSUE #5251: Sorted Table Segments may block StorageWriter
**Problem description**
Sorted Table Segments hold their index as a B+TreeSet inside the Table Segment. Every time we persist changes to this sorted index, we need to update the B+TreeSet. 

It is possible that, for an update containing a very large number of keys (size of keys is irrelevant), where these keys span the whole spectrum of the B+TreeSet, the Sorted Index update would try to update more than 32MB of entries at once (it will exceed the max allowed batch size). If this is the case, the update will fail. To make matters worse, the storage writer will retry this operation over and over again, failing to make any progress due to the failure -> blockage.

**Problem location**
Sorted Key Index.

Upon updating, the B+TreeSet orders the changes in such a way that it is possible to break them down into smaller pieces. 
By applying them one-by-one, we can circumvent this limit. The fact that the B+TreeSet orders these updates in such a way is important - given that this may be a non-atomic update, it is possible that some parts (from the beginning) may be accepted, while the rest may not (i.e., due to a system crash). This will not leave the index in an inconsistent or corrupted state.

andreipaduroiu:
**Change log description**  
- Added per-segment throttling for each Table Segment. Limiting to 128MB max unindexed data. This is supported by an Async Semaphore construct that can allow/queue tasks with given number of credits/permits.
- Updated TableSegmentImpl to use two connections: one for writing and one for reading (initialized lazily on demand).
- Fixed a bug in SortedKeyIndex where it would be possible to block the StorageWriter by continuous failed index updates.
- Added KVT support to the Self Tester tool.

**Purpose of the change**  
Fixes #5162.
Fixes #5251.

**What the code does**  
- **`AsyncSemaphore`**
    - Similarly to the Semaphore synchronization primitive the `AsyncSemaphore` defines a maximum number of allowed credits (permits) and each async task that is run though it can request a number of such credits to be used. 
        - If there are sufficient credits, it is allowed through and the credits recorded.
        - If insufficient credits, the task is queued (in order) and will be eligible for execution once there are sufficient credits available.
    - The credits are restored via a separate API  (and not when the task completes). See below why this was done this way.
    - The task also has a "force" mode, where a task can be forced through. If set, that task is allowed to execute (regardless of whether there are credits or not) and it will still cause credits to increase.
- **Table Segment Throttling**
    - Each Table Segment has an `AsyncSemaphore` attached to it. 
        - The maximum number of credits is 128*1024*1024 (128MB outstanding).
    - Each external update (update or removal) will require a number of credits equal to that update's serialization length.
        - Credits are requested when the update is about to be processed (in the index), and the task will be queued up (in the `AsyncSemapthore`) if not enough credits are available.
    - The `WriterTableProcessor` (that performs background indexing; driven by the StorageWriter) will return credits back to the `AsyncSemaphore` once it has indexed said updates.
    - Internal updates (the Sorted Index B+TreeSet) are exempt from this credit quota. While the index updates do acquire credits, they will not be subject to blocking.
         - The  Sorted Index is updated from within the `WriterTableProcessor` - if it had been subject to throttling, the indexer would essentially block itself. Hence we need this exemption, only for this purpose. It is up to the `ContainerTableExtensionImpl` to properly label each update as Internal or External - depending on where it was coming from.
- **Connection Pausing/Resuming**
    - Included the TableSegment updates and removals to the global `ConnectionTracker` (that was used for appends). It is now tracking both outstanding appends and updates/removals to Table Segments.
    - The same limits as before apply: 512MB per process limit, with 128MB per connection limit.
    - No changes were required on the client since the TCP backpressure that results due to the connections being paused correctly causes the Client to block synchronously when sending requests.
- **`TableSegmentImpl`**
    - Due to connection pausing/resuming, it is quite plausible that read requests may be unintentionally blocked from being sent through if the same `TableSegmentImpl` instance is used for both reads and updates, and the updates are throttled.
    - This has been alleviated by establishing a connection for writing ("writeContext") used by `put` and `remove` and one for reading ("readContext") used for `get` and `iterate`. Both are lazily created on the first request and request different connections from the connection pool.
        - Note: since this uses a connection pool, it may well be that both will end up using the same connection; however the Connection Pool configuration is out of scope for `TableSegmentImpl` as it is set on the `ConnectionFactory` constructor.
- **Issue #5251 - Sorted Index Updates too large**
    - Added a special, internal-only API to the `ContainerTableExtensionImpl` that enables non-atomic, multi-key updates. 
    - If an update is rejected because it's too large, it will be broken down in equal halves and reattempted, in sequential order (first half first, followed by the second half). Please see #5251 as to why this won't cause the sorted index to become inconsistent if we get a system failure somewhere in the middle and this is only persisted partially.
- Other changes
    - Updated SelfTester to include tests for client-side KeyValue Tables.

**How to verify it**  
Lots of new unit tests added to verify new functionality.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableExtensionConfig.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+@Data
+@Builder
+class TableExtensionConfig {
+    /**
+     * The maximum unindexed length ({@link SegmentProperties#getLength() - {@link TableAttributes#INDEX_OFFSET}}) of a
+     * Segment for which {@link ContainerKeyIndex} {@code triggerCacheTailIndex} can be invoked.
+     */
+    @Builder.Default
+    private int maxTailCachePreIndexLength = EntrySerializer.MAX_BATCH_SIZE * 4;
```
sachin-j-joshi, line 35: did not understand why multiplied by 4

sachin-j-joshi (approved):
This looks okay to me.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/TableExtensionConfig.java
andreipaduroiu: MaxBatchSize == 32MB (32 * MaxEntrySize).
I wrote 128MB as a function of MaxBatchSize, and implicitly as a function of MaxEntrySize.

common/src/main/java/io/pravega/common/concurrent/AsyncSemaphore.java
```diff
@@ -0,0 +1,243 @@
...
+    //region Constructor
+
+    /**
+     * Creates a new instance of the {@link AsyncSemaphore} class.
+     *
+     * @param totalCredits Total number of available credits.
+     * @param usedCredits  Initial number of used credits.
+     * @param logId        A log-friendly identifier for this {@link AsyncSemaphore}.
+     */
+    public AsyncSemaphore(long totalCredits, long usedCredits, String logId) {
```
tkaitchuck, line 61: I feel like this should take a threadpool to run the tasks on. I seems a bit strange that they get run in the thread that calls release.

common/src/main/java/io/pravega/common/concurrent/AsyncSemaphore.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /**
+     * Releases a number of credits back into the pool and initiates the execution of any pending tasks that are now
+     * eligible to run.
+     *
+     * @param credits The number of credits to release. This number will be capped at the number of currently used
+     *                credits ({@link #getUsedCredits()}).
+     */
+    public void release(long credits) {
```
tkaitchuck, line 159: It would be good if we could automate calling this so that the caller cannot forget. IE by chaining it's invocation on the future returned for example.

tkaitchuck (approved):

andreipaduroiu:
This PR should be good for merging now in 0.10.0.

We should not cherrypick this into 0.9.x as it is technically a new feature, not a bug fix.

cc @tkaitchuck @sachin-j-joshi

Issue #1644, opened 2017-07-24
Enable increasing the number of containers


tkaitchuck:
**Problem description**
Right now we are mapping segments to containers by simple consistent hashing. We should modify this so that we can increase the number of containers.

**Problem location**
SegmentToContainerMapper

**Suggestions for an improvement**
Use a inverse integer mapping. IE: Randomize the location by hashing the string, and make to each segment by placing them on the unit circle based on the inverse of the container number.

fpj:
I think the bulk of the changes in this proposal are in the controller. I'm not sure what this "SegmentName-to-ContainerId check" is, but if we are not moving segments across containers, then the check seems fine.

andreipaduroiu:
@fpj I agree. Most of this is Controller work.

> I'm not sure what this "SegmentName-to-ContainerId check"
When we receive a request into the Segment Store, we only have the segment name. But the Segment Store still needs to route that request to the Container that handles the Segment, hence the need for:
1. An additional config on the Segment Store side for number of containers
2. An additional hash on the Segment Store side to route to a container.

We could change this by either sending the Container Id along with the Segment Name via the Client (virtually every Wire Protocol command will need a change) or have the Segment Store gain access to the Controller's epoch metadata so it can do the internal routing.

PR #5253, opened 2020-10-14
Issue 5169: Data Recovery - Copy the core attributes from old segment metadata to new segment metadata

### LINKED ISSUE #5169: Data Recovery - Copy the core attributes from old segment metadata to new segment metadata

ManishKumarKeshri:
**Change log description**  
Instead of directly deleting the container metadata and its attribute segments, we now keep a back up copy of them and then delete them. This back up copy is used to update the attributes of the recovered segments. 

**Purpose of the change**  
Fixes #5169 

**What the code does**  
The code has got methods: 
`backUpMetadataAndAttributeSegments` to make copies of metadata and its attribute segment. This is performed for all containers. 
and `updateCoreAttributes`, in which we iterate through the segments in the back up copy of the metadata segment and update the attributes of it in the newly created container metadata segment. 
 
**How to verify it**  
Build shall pass.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
                 ex -> ex instanceof StreamSegmentNotExistsException, null);
     }
 
+
+    /**
+     * Copies the contents of container metadata segment and its attribute segment to new segments. The new names to the segments
```
andreipaduroiu, line 254: "the new names of the segments"

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     * are passed as parameters.
+     * @param storage                    A {@link Storage} instance where segments are stored.
+     * @param containerId                A Container Id to get the name of the metadata segment.
+     * @param backUpMetadataSegmentName  A name to the back up metadata segment.
```
andreipaduroiu, line 258: "Name of the ..."
below too

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     * @param backUpAttributeSegmentName A name to the back attribute segment.
+     * @param executorService            A thread pool for execution.
+     * @return                           CompletableFuture which when completed will have the segments' contents copied to another segments.
```
andreipaduroiu, line 261: "will indicate the operation has completed"

(the future itself won't have anything in it)

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+        Preconditions.checkNotNull(storage);
+        String metadataSegmentName = NameUtils.getMetadataSegmentName(containerId);
+        String attributeSegmentName = NameUtils.getAttributeSegmentName(metadataSegmentName);
+        return copySegment(storage, metadataSegmentName, backUpMetadataSegmentName, executorService)
+                        .thenAcceptAsync(x -> copySegment(storage, attributeSegmentName, backUpAttributeSegmentName,
+                                executorService));
+    }
+
+    /**
+     * Given the back up of metadata segments, this method iterates through all segments in it and gets their attributes
```
andreipaduroiu, line 276: Rephrase as:
```
Updates Core Attributes for all Segments for the given Containers.
This method iterates through all the Container Metadata Segments for the given Containers, interprets all entries as Segment-SegmentInfo mappings and extracts the Core Attributes for each. These Core Attributes are then applied to the same segments in the new Containers.
```

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     * and updates it in the container metadata segment of the given {@link DebugStreamSegmentContainer} instance.
+     * @param metadataSegments  A map of back of metadata segments along with their container Ids.
+     * @param containersMap     A map of {@link DebugStreamSegmentContainer} instances with their container Ids.
+     * @param executorService   A thread pool for execution.
+     * @throws Exception        If an exception occurred. This could be one of the following:
```
andreipaduroiu, line 281: Why not `@throws TimeoutException` and `@throws IOException`? That is the standard Javadoc we should use.

The format you used was our convention for async methods that return CompletableFuture.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     *                              * TimeoutException:     If the calls for computation(used in the method)
+     *                                                      didn't complete in time.
+     *                              * IOException     :     If a general IO exception occurred.
+     */
+    public static void updateCoreAttributes(Map<Integer, String> metadataSegments,
+                                            Map<Integer, DebugStreamSegmentContainer> containersMap,
+                                            ExecutorService executorService) throws Exception {
+        val args = IteratorArgs.builder().fetchTimeout(TIMEOUT).build();
+        SegmentToContainerMapper segToConMapper = new SegmentToContainerMapper(containersMap.size());
```
andreipaduroiu, line 290: You assume that your input is correct. Can you please do the following validations:
- metadataSegments.size == containersMap.size
- All container ids are accounted for

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+
+        ArrayList<CompletableFuture<Void>> futures = new ArrayList<>();
+        for (val metadataEntry : metadataSegments.entrySet()) {
+            val containerForSegmentsContained = containersMap.get(metadataEntry.getKey());
+            val containerForBackUpMetadataSegment = containersMap.get(segToConMapper.getContainerId(metadataEntry.getValue()));
+            log.info("container id: {}, segment Name: {}", containerForBackUpMetadataSegment.getId(), metadataEntry.getValue());
```
andreipaduroiu, line 296: "Container id: {}, Segment Name: {}" (capitalization)

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+
+                    // skip, if this is a metadata segment
+                    if (metadataSegmentName.equals(properties.getName())) {
+                        continue;
+                    }
+
+                    List<AttributeUpdate> attributeUpdates = properties.getAttributes().entrySet().stream()
+                            .map(e -> new AttributeUpdate(e.getKey(), AttributeUpdateType.Replace, e.getValue()))
+                            .collect(Collectors.toList());
+                    log.info("Updating attributes = {} for segment '{}'", attributeUpdates, properties.getName());
```
andreipaduroiu, line 316: "Segment Name: {}, Attribute Updates: {}`

(put attribute updates second since it's going to be a long list)

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +249,80 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     *                                                      didn't complete in time.
+     *                              * IOException     :     If a general IO exception occurred.
+     */
+    public static void updateCoreAttributes(Map<Integer, String> metadataSegments,
+                                            Map<Integer, DebugStreamSegmentContainer> containersMap,
+                                            ExecutorService executorService) throws Exception {
+        val args = IteratorArgs.builder().fetchTimeout(TIMEOUT).build();
+        SegmentToContainerMapper segToConMapper = new SegmentToContainerMapper(containersMap.size());
+
+        ArrayList<CompletableFuture<Void>> futures = new ArrayList<>();
```
andreipaduroiu, line 292: To make your life easier, I suggest you create the futures and wait for them for each container. That way if one of your containers fails to migrate, at least you know where you left off and what finished executing.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -265,6 +342,7 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
                     return Futures.loop(
                             () -> bytesToRead.get() > 0,
                             () -> {
+                                log.info("Reading");
```
andreipaduroiu, line 345: Add more info. What container id?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentContainerTests.java
```diff
@@ -2158,7 +2158,7 @@ private void mergeTransactions(HashMap<String, ArrayList<String>> transactionsBy
         Futures.allOf(mergeFutures).get(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
     }
 
-    private RefCountByteArraySegment getAppendData(String segmentName, int appendId) {
+    public static RefCountByteArraySegment getAppendData(String segmentName, int appendId) {
```
andreipaduroiu, line 2161: Why do you need this? I don't see where you use this or why you need a `RefCountByteArraySegment`. That class is meant to track append leaks, which you do not need to do.

All you need to do in your tests is generate a basic `ByteArraySegment`, so you don't need these changes here.

andreipaduroiu (changes requested):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Can there be a situation like this:
- We are given a storage instance, in pravega was running with just two containers.
- Migrate the storage instance to new pravega with 4 containers.
So, with the given storage instance with two container metadata segments, we get their back up copies. Now we iterate through 2 back up copies and update the segments in 4 containers. So, a precheck will fail in this case.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentContainerTests.java
ManishKumarKeshri: Fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Removed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
andreipaduroiu: No we cannot have that. We do not currently support changing the number of Segment Containers in a cluster (we may do this - #1644).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +254,112 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+                            properties.getName()));
+
+                    // Update attributes for the current segment
+                    futures.add(Futures.exceptionallyExpecting(
+                            container.updateAttributes(properties.getName(), attributeUpdates, TIMEOUT),
+                            ex -> ex instanceof StreamSegmentNotExistsException, null));
+                }
+
+                // Waiting for update attributes for all segments in each back up metadata segment.
+                Futures.allOf(futures).join();
```
andreipaduroiu, line 358: What I meant is to add this outside of the `entryIterator.forEachRemaining` but still withinin the outer for loop.

And when you wait on it, use `.get(with timeout)` so we don't end up waiting forever.

In that case we also need to make sure the Timeout is something reasonable. If we set it to 5s, then that may not be sufficient to restore all the segments in a container (there could be millions of them).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -278,4 +392,42 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+     * @param storage                   A {@link Storage} instance to get the segments from.
+     * @param containerCount            The number of containers for which renaming of container metadata segment and its
+     *                                  attributes segment has to be performed.
+     * @param executorService           A thread pool for execution.
+     * @return                          A Map of Container Ids to new container metadata segment names.
+     * @throws InterruptedException     If the operation was interrupted while waiting.
+     * @throws TimeoutException         If the timeout expired prior to being able to complete the operation.
+     * @throws ExecutionException       When execution of the opreations encountered an error.
+     */
+    public static Map<Integer, String> getBackUpMetadataSegments(Storage storage, int containerCount, ExecutorService executorService)
```
andreipaduroiu, line 410: Your Javadoc says "creates a backup" yet this method is named `get`. Consider renaming to `createBackupMetadataSegments`

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainerTests.java
```diff
@@ -250,6 +258,149 @@ public void testEndToEnd() throws Exception {
...
+        String backUpAttributeSegment = "segment-" + RANDOM.nextInt();
+
+        s.create(metadataSegment, TIMEOUT).join();
+        s.create(attributeSegment, TIMEOUT).join();
+
+        ContainerRecoveryUtils.backUpMetadataAndAttributeSegments(s, CONTAINER_ID, backUpMetadataSegment, backUpAttributeSegment,
+                executorService()).join();
+
+        // back up metadata segment should exist
+        Assert.assertTrue("Unexpected result for existing segment (no files).", s.exists(backUpMetadataSegment, null).join());
```
andreipaduroiu, line 286: why do you pass null for timeout here?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Thanks for telling that. I am now, taking timeOut as a parameter to the method. The tests which use this method, will pass their timeOut values here. 
I will do similar thing for other methods in this class like `recoverAllSegments` & `copySegment` in this class.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
ManishKumarKeshri: Done.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/DebugStreamSegmentContainerTests.java
ManishKumarKeshri: Fixed.

andreipaduroiu (approved):

sachin-j-joshi (changes requested):
Minor fixes . Otherwise looks okay

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -278,4 +388,44 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+        Map<Integer, String> backUpMetadataSegments = new HashMap<>();
+
+        val futures = new ArrayList<CompletableFuture<Void>>();
+
+        for (int containerId = 0; containerId < containerCount; containerId++) {
+            String backUpMetadataSegment = NameUtils.getMetadataSegmentName(containerId) + fileSuffix;
+            String backUpAttributeSegment = NameUtils.getAttributeSegmentName(backUpMetadataSegment);
+            log.debug("Created '{}' as a back of metadata segment of container Id '{}'", backUpAttributeSegment, containerId);
+
+            int finalContainerId = containerId;
```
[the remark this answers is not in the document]
ManishKumarKeshri: Fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerRecoveryUtils.java
```diff
@@ -246,6 +254,108 @@ private static void validateContainerIds(Map<Integer, DebugStreamSegmentContaine
...
+
+            // Get the container for segments inside back up metadata segment
+            val container = containersMap.get(backUpMetadataSegmentEntry.getKey());
+
+            // Get the iterator to iterate through all segments in the back up metadata segment
+            val tableExtension = containerForBackUpMetadataSegment.getExtension(ContainerTableExtension.class);
+            val entryIterator = tableExtension.entryIterator(backUpMetadataSegment, args)
+                    .get(timeout.toMillis(), TimeUnit.MILLISECONDS);
+
+            ArrayList<CompletableFuture<Void>> futures = new ArrayList<>();
```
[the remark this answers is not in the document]
ManishKumarKeshri: Used val.

sachin-j-joshi (approved):
LGTM

Issue #5172, opened 2020-09-11
Revert the changes for issue 4775


crazyzhou:
**Problem description**
#4775 is raised to add a snapshot release to other release branch, but then the release process is going on, it will block the travis CI to succeed. It will fail because we cannot upload the release version artifact in the snapshot artifactory.

**Problem location**
`.travis.yml`

**Suggestions for an improvement**
Revert the PR from `master` branch and `0.8` branch.

PR #5273, opened 2020-10-21
Issue 4775: Add snapshot stage for all release branches

### LINKED ISSUE #4775: Add snapshot stage for all release branches

crazyzhou:

**Change log description**  
Add the release branches into the travis snapshot stage.

**Purpose of the change**  
Fixes #4775 

**What the code does**  
Add the release branches into the travis snapshot stage with regex `/^r[0-9]\.[0-9]/`.

**How to verify it**  
Verified before with #4776 , however it is reverted #5172 for breaking the 0.8 release branch build. We now bring it back because we have a finalized new process for release which can make this work.


.travis.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       script: ./gradlew javadocs
 
     - stage: snapshot
-      if: branch = master AND NOT (type = pull_request)
+      if: (branch = master OR branch =~ /^r[0-9]+\.[0-9]+/) AND NOT (type = pull_request)
```
ravisharda, line 37: This won't work for `r0.10` or `r0.11`. You might want to take care of that too.

ravisharda: Actually, this'd work, as there are `+` symbols in the pattern.

ravisharda (approved):

medvedevigorek (approved):
Looks good. Thank you.

PR #5283, opened 2020-10-22
Issue 4884: BufferView equals() and compare() optimizations for JDK11

### LINKED ISSUE #4884: Optimize BufferView implementations
**Problem description**
Consider the following classes and/or methods:
- `BitConverter.readInt`, `BitConverter`readLong` (and all associated overloads and methods)
- `ByteArraySegment`
- `BufferView.equals`
- `ByteArrayComparator`

Explore ways to have these use as much native code/HW instructionset as possible in order to make them more efficient.

**Problem location**
Common.

1. Consider using `ByteBuffer` inside `ByteArraySegment`.
2. Consider upgrading to Java 11 and make use of the intrisic methods added in the `java.util.Arrays` class.
3. Rename `ByteArrayComparator` to `BitwiseComparator` and expose 3 final members: `ByteArrayComparator`, `ArrayViewComparator` and `BufferViewComparator`.

andreipaduroiu:
**Change log description**  
- Created project `common_server` - everything that is common but needs to be compiled with JDK11.
- Moved `io.pravega.common.util.btree` from `common` to `common_server`. Package name stays the same.
- Deleted `ByteArrayComparator.java`. Replaced with `BufferViewComparator.java`, which now sits in `common_server`.
- Updated all callsites to use new classes.

**Purpose of the change**  
Fixes #4884.

**What the code does**  
- Created `common_server` as a top-level project. As opposed to `common` (which needs to compile with JDK8), `common_server` compiles with the default JDK version configured in `gradle.properties` (currently 11).
    - Referenced this from `segmentstore:server`, since that's the only place which needs it at the moment.
    - Moved `io.pravega.util.btree` package into it.
- `BufferView`
    - Added `iterateBuffers` which returns an `Iterator<ByteBuffer>`. Every `BufferView` implementation is a collection of 0 or more ordered `ByteBuffers`; this iterator will return each one of them.
    - Reworked `ByteArraySegment.equals()` to use `ByteBuffer.equals` instead of custom comparison. This API exists in all versions of Java (that we care about), so the runtime implementation may change freely (JRE8 vs JRE9+) without needing for code changes on our end. This is safe to be used in all Pravega components.
    - Reworked `AbstractBufferView.equals` to use `BufferView.iterateBuffers` for the two involved buffers. This does "slice-by-slice" comparisons using the underlying `BufferView` implementation of `equals`, which should be the most efficient way of comparing for equality (i.e., a `CompositeBufferView` made of `ByteArraySegment`s or `ByteBufWrapper`s will delegate to those implementations of `equals` to compare slices).
- `BufferViewComparator`
    - Renamed `ByteArrayComparator` to `BufferViewComparator` (more appropriate name, given what it does).
    - Moved from `common` to `common_server` (same package name)
    - Provided 2 implementations: `Legacy` (working as before) and `Intrinsic` (using JDK9+ APIs)
    - Providing a `create()` method that returns the appropriate implementation instance based on what the platform supports.

**How to verify it**  
Build, tests and system tests must pass.


andreipaduroiu:
Comparison:
I tested `equals()` and `compare` using 256 100KB randomly generated buffers. I compared each buffer against every other one and repeated the whole thing 100 times (~6.5M runs). The table below shows the total runtime (in millis) for each such test (NOTE: your mileage may vary, depending on OS and hardware type - use this table only for comparison purposes):

```
                        .equals()          .compare()
                    Legacy	Intrinsic	Legacy	Intrinsic
ByteArraySegment    165	    30          131     26
ByteBufWrapper	    435	    69          621     92
CompositeBufferView 1613    115         2110    111
```

common/src/main/java/io/pravega/common/util/ByteArraySegment.java
```diff
@@ -171,44 +172,28 @@ public void copyTo(OutputStream stream) throws IOException {
...
 
     @Override
-    @SneakyThrows(IOException.class)
     public boolean equals(BufferView other) {
         if (this.length != other.getLength()) {
             return false;
         } else if (other instanceof ArrayView) {
-            return equals((ArrayView) other);
+            // ByteBuffer-optimized equality check.
+            return this.asByteBuffer().equals(((ArrayView) other).asByteBuffer());
```
tkaitchuck, line 180: I believe this method does the same thing without the wrapping:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Arrays.html#equals(byte%5B%5D,int,int,byte%5B%5D,int,int)

build.gradle
```diff
@@ -377,6 +398,7 @@ project ('bindings') {
 project('segmentstore:server') {
     dependencies {
         compile project(':common')
+        compile project(':common_server')
```
eolivelli, line 401: I am not sure that using the underscore is a good practice.
What about common.server?

common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
```diff
@@ -183,6 +185,31 @@ public CompositeArrayView slice(int offset, int length) {
         assert length == 0 : "Collection finished but " + length + " bytes remaining";
     }
 
+    @Override
+    public Iterator<ByteBuffer> iterateBuffers() {
+        if (this.length == 0) {
+            return Collections.emptyIterator();
+        }
+
+        AtomicInteger arrayOffset = new AtomicInteger(getArrayOffset(0));
```
eolivelli, line 194: Why AtomicInteger? What about using some simpler int holder without concurrency support?

common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
```diff
@@ -183,6 +185,31 @@ public CompositeArrayView slice(int offset, int length) {
         assert length == 0 : "Collection finished but " + length + " bytes remaining";
     }
 
+    @Override
+    public Iterator<ByteBuffer> iterateBuffers() {
+        if (this.length == 0) {
+            return Collections.emptyIterator();
+        }
```
eolivelli, line 192: Does is make sense to have a simpler handling in case of wrapping only 2 buffers?
Probably we can save a few cpu cycles and memory allocations, but I am not sure it is worth

common_server/src/main/java/io/pravega/common/util/BufferViewComparator.java
```diff
@@ -0,0 +1,320 @@
...
+    private static final boolean INTRINSIC_SUPPORTED;
+
+    static {
+        // Figure out if we support JRE9+ Intrinsic comparators. We do compile this code with JDK11, however this is to
+        // have a failover in case we run with a lesser JRE.
+        boolean intrinsicSupported = true;
+        try {
+            ByteBuffer bb = ByteBuffer.wrap(new byte[1]);
+            bb.mismatch(bb);
+        } catch (ExceptionInInitializerError | UnsatisfiedLinkError | NoClassDefFoundError e) {
```
eolivelli, line 74: What about using multirelease jars? It will be more simpler at runtime and less error prone

eolivelli (approved):
Great work.
I left a few suggestions

build.gradle
andreipaduroiu: That was my first choice too. However Gradle completely refused to accept that. Having `common` and `common:server` (which means `common:server` must be in a direct sudirectory of `common`) prevented me from being able to access the packages in `common:server` (that is, when adding a reference in `segmentstore:server` to `common:server`, Gradle complained about a circular dependency and Intellij Idea refused to see those packages. Moving this code to a top-level package (like it is now) did not have this issues.

I couldn't figure out what was wrong there since I did nothing special. We do use that pattern before (`segmentstore:server` and `segmentstore:server:host`) albeit with slight differences - the only difference I could tell was the both `common` and `common:server` share the same root package name `io.pravega.common`. I didn't want to create new/sub packages for this code.

common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
andreipaduroiu: Because I need to use this in a lambda just below which doesn't like me fiddling around with the variable contents (it needs to be immutable). 

The reason for a lambda is because this method needs to return an iterator, and the simplest way to make one on-the-fly is by using `Arrays.stream()...` and then doing some mapping, hence the need for lambda.

common/src/main/java/io/pravega/common/util/CompositeByteArraySegment.java
andreipaduroiu: I tried to optimize both this and `CompositeBufferView` based on when they have only one (or two) buffers. The difference I saw was measured in fractions of nanoseconds so not worth it in my POV.

FYI, `CompositeByteArraySegment` is only used in one place (currently). When we generate writes to BookKeeper, we need to write at most 1MB, but we do not know how much in advance to allocate. To give the impression of an array, this class creates a "sparse array" made of smaller arrays which are only allocated when first written to. 

Most of the time this will have just 1 element (up to 128KB).

common_server/src/main/java/io/pravega/common/util/BufferViewComparator.java
andreipaduroiu: Gradle [recommends against that](https://blog.gradle.org/mrjars), so in my opinion, this is less controversial (and more deterministic - what goes where).

common/src/main/java/io/pravega/common/util/ByteArraySegment.java
andreipaduroiu: Good point. Fixed.

andreipaduroiu: That method is not available in JDK8. This class is part of `common` so it must compile with JDK8 APIs. 

However, instances of this class may be used in JRE11 environments (such as Segment store) where the updated `ByteBuffer.equals` will be used.

eolivelli (approved):
LGTM

andreipaduroiu:
@RaulGracia I don't expect you will see much difference, if any at all. We don't use BufferView equals or compare for the streaming path too much. They're used heavily for Table Segments (any metadata we use) and for Key-Value Tables.

tkaitchuck (approved):

PR #5284, opened 2020-10-26
Data recovery commands in CLI Tool

ManishKumarKeshri:
**Change log description**  
- Added `Tier1-recovery <root>` and `list-segments <root>` commands and their implementations.

**Purpose of the change**  
  Adds data recovery feature to the CLI tool. 

**What the code does**  
- Invokes `list-segments` API to enumerate all segments stored in the LTS
- Tier1-recovery command recovers all the segments from the storage and restores the Tier1.

**How to verify it**  
Command should be able to list segments and perform recovery.


cli/admin/src/main/java/io/pravega/cli/admin/dataRecovery/DataRecoveryCommand.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Creates logging directory and file for the command to be run. The path to the directory can be supplied on the
+     * command run. By default, the path is set as current user path.
+     *
+     * @param commandName   The name of the command to be run.
+     * @return              The path to the directory created.
+     * @throws Exception    In case of a failure in creating the directory or the file.
+     */
+    String setLogging(String commandName) throws Exception {
```
RaulGracia, line 72: This seems to be something not specific to this command, but something that is orthogonal to all commands, right? If you agree, maybe you could implement this in `AdminCommand`, so all other commands can benefit from logging if we configure to do so. What do you think?

cli/admin/src/main/java/io/pravega/cli/admin/dataRecovery/StorageListSegmentsCommand.java
```diff
@@ -0,0 +1,136 @@
...
+        this.csvWriters = new FileWriter[this.containerCount];
+    }
+
+    /**
+     * Creates a csv file for each container. All segments belonging to a containerId have their details written to the
+     * csv file for that container.
+     *
+     * @throws Exception   When failed to create/delete file(s).
+     */
+    private void createCSVFiles() throws Exception {
```
RaulGracia, line 65: Why do we need to write all this information in CSV files? Are they going to be used later on for data migration purposes, right?

cli/admin/src/main/java/io/pravega/cli/admin/dataRecovery/Tier1RecoveryCommand.java
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Loads the storage instance, recovers all segments from there.
+ */
+public class Tier1RecoveryCommand extends DataRecoveryCommand {
+    private static final int CONTAINER_EPOCH = 1;
+    private static final Duration TIMEOUT = Duration.ofMillis(100 * 1000);
+
+    // Configurations for setting the context for running debug segment container(s)
+    private static final DurableLogConfig DURABLE_LOG_CONFIG = DurableLogConfig
```
RaulGracia, line 62: Looks like a lot of hardcoded configuration. Is this ok or the user will need to change these parameters depending on the configuration of Tier 1 on each particular Pravega cluster?

config/admin-cli.properties
```diff
@@ -14,16 +14,48 @@ cli.security.auth.credentials.username=admin
 cli.security.auth.credentials.password=1111_aaaa
 
 # Config for CLI Controller REST/GRPC endpoints.
-cli.controller.rest.uri=http://localhost:9091
-cli.controller.grpc.uri=tcp://localhost:9090
+cli.controller.rest.uri=http://10.100.200.187:9091
```
RaulGracia, line 17: Please, remove IPs and non-default values from this file.

config/admin-cli.properties
```diff
@@ -14,16 +14,48 @@ cli.security.auth.credentials.username=admin
...
+# hdfs.root=
+
+# Root path where NFS shared directory needs to be mounted before segmentstore starts execution.
+# filesystem.root=
+
+# Supported parameters include but are not limited to:
+#   url: where the extended S3 cluster is accessible at, e.g. http://localhost:9020
+#   identity: the access key to access the extended S3 cluster, e.g. identity=user
+#   secretKey: the secret key to access the extended S3 cluster, e.g. secretKey=password
+extendeds3.connect.config.uri=http://10.243.86.64:9020?namespace=enki%26identity=enki%26secretKey=jTjchi5OQ5uwX1K4ND2hkLP39qcnAze4xIH0cscw
```
RaulGracia, line 51: Delete this, looks like something that shouldn't be posted.

RaulGracia (changes requested):

ManishKumarKeshri:
I am testing this branch on a real cluster.

PR #5289, opened 2020-10-30
Issue 5111: Consumption based retention policy implementation

### LINKED ISSUE #5111: Implement Consumption Based Retention on Controller for Streams with CBR policy
**Problem description**
Implement CBR on Controller for Streams with CBR policy.

**Problem location**
Controller

This includes the following:
1. Code for truncating Streams with CBR.
2. Code for generating common truncation Stream-Cut from all RG truncation Stream-Cuts.

shiveshr:
**Change log description**  
Implements consumption based retention policy. 

**Purpose of the change**  
Fixes #5111 

**What the code does**  
This does two things primarily - 
1. In Stream Metadata Tasks it consolidates and computes lower bound on subscriber streamcuts and truncates at the said stream cut. This also applies time/size bounds
2. In stream cut validation in stream metadata store, we assumed stream cuts to be made of non overlapping segments, but a stream cut could have future segments which may not have been read from. These are represented using "-1" as the offset. So we update the logic in stream metadata store for validation as well as size computations for streamcuts. 

**How to verify it**  
Unit tests added.


controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -634,75 +734,183 @@ private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegments
...
             return isStreamCutValid(streamCut)
-                    .thenCompose( isValidStreamCut -> {
+                    .thenCompose(isValidStreamCut -> {
                         if (isValidStreamCut) {
-                            return computeStreamCutSpan(streamCut)
-                                    .thenCompose(span -> computeStreamCutSpan(previousStreamCut)
-                                            .thenApply(previousSpan -> greaterThan(streamCut, span, previousStreamCut, previousSpan)));
+                            CompletableFuture<ImmutableMap<StreamSegmentRecord, Integer>> span1 = computeStreamCutSpan(streamCut);
+                            CompletableFuture<ImmutableMap<StreamSegmentRecord, Integer>> span2 = previousStreamCut.isEmpty() ?
+                                    CompletableFuture.completedFuture(ImmutableMap.of()) : computeStreamCutSpan(previousStreamCut);
```
pbelgundi, line 899: `span2` will never be equal to `CompletableFuture.completedFuture(ImmutableMap.of())` becuase if `previousStreamCut.isEmpty=true`, that case is already handled in the if block.

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
```diff
@@ -680,4 +680,22 @@
...
     CompletableFuture<Integer> getSegmentSealedEpoch(long segmentId);
+
+    /**
+     * Method to compare streamcuts to check if streamcut1 is strictly ahead of streamcut2. 
+     * @param cut1 streamcut to check
+     * @param cut2 streamcut to check against. 
+     *
+     * @return CompletableFuture which, upon completion, will indicate if the streamcut1 is strictly ahead of streamcut2.
+     */
+    CompletableFuture<Boolean> streamCutStrictlyGreaterThan(Map<Long, Long> cut1, Map<Long, Long> cut2);
```
pbelgundi, line 691: method name should be `isStreamCutStrictlyGreaterThan` since it returns a boolean.

controller/src/main/java/io/pravega/controller/store/stream/StreamMetadataStore.java
```diff
@@ -1276,4 +1276,40 @@
...
+     * @param streamCut1 Stream cut to check
+     * @param streamCut2 Streamcut to check against
+     * @param context    operation context.
+     * @param executor   callers executor.
+     *                   
+     * @return A completable future which when completed will hold a boolean which will indicate if streamcut1 is strictly
+     * ahead of streamcut2. 
+     */
+    CompletableFuture<Boolean> streamCutStrictlyGreaterThan(final String scope, final String streamName,
+                                                            Map<Long, Long> streamCut1, Map<Long, Long> streamCut2,
```
pbelgundi, line 1296: What is meant by "strictly" here? Would be good to explain that in the javadoc

shared/controller-api/src/main/proto/Controller.proto
```diff
@@ -298,9 +298,22 @@ message RetentionPolicy {
...
+        CONSUMPTION = 3;
     }
     RetentionPolicyType retentionType = 1;
     int64 retentionParam = 2;
+    ConsumptionLimits consumptionLimits = 3;
+}
+
+message ConsumptionLimits {
+    enum ConsumptionLimitType {
+        UNKNOWN = 0;
```
pbelgundi, line 310: Here, UNKNOWN perhaps means we don;t set a limit for CBR. Could we call it NONE instead of UNKNOWN?

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
```diff
@@ -162,6 +195,29 @@ public static final StreamProperty encodeStreamResponse(String scope, String str
...
+                    switch (cl.getType()) {
+                        case SIZE_BYTES:
+                            consumptionLimits.setType(ConsumptionLimits.TypeEnum.SIZE_MB);
+                            divider = BYTES_TO_MB;
+                            break;
+                        case TIME_MILLIS:
+                            divider = MILLIS_TO_MINUTES;
+                            consumptionLimits.setType(ConsumptionLimits.TypeEnum.TIME_MINUTES);
+                            break;
+                        default:
```
pbelgundi, line 212: Same as above

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
```diff
@@ -70,6 +80,24 @@ public static final StreamConfiguration getCreateStreamConfig(final CreateStream
...
+        switch (consumptionLimits.getType()) {
+            case SIZE_MB:
+                type = RetentionPolicy.ConsumptionLimits.Type.SIZE_BYTES;
+                multiplier = BYTES_TO_MB;
+                break;
+            case TIME_MINUTES:
+                type = RetentionPolicy.ConsumptionLimits.Type.TIME_MILLIS;
+                multiplier = MILLIS_TO_MINUTES;
+                break;
+            default:
```
pbelgundi, line 95: ConsumptionLimit.UNKNOWN is not covered here. That is a valid usecase and should not throw "NotImplementedException"

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/server/rest/generated/model/ConsumptionLimits.java
```diff
@@ -0,0 +1,162 @@
...
+ */
+
+public class ConsumptionLimits   {
+  /**
+   * Indicates if consumption limits is by space or time.
+   */
+  public enum TypeEnum {
+    SIZE_MB("SIZE_MB"),
+    
+    TIME_MINUTES("TIME_MILLIS");
```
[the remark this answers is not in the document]
shiveshr: its a generated class. i wonder how it got changed.. must have been some replace all that i did on some string match.

controller/src/main/java/io/pravega/controller/store/stream/StreamMetadataStore.java
shiveshr: isnt this already explained in the javadoc here. 
```
     * Compares two Stream cuts and returns true if streamcut1 is strictly ahead of streamcut2 else returns false. 
     * This method will return false for both strictly less than and overlapping streamcuts.
     * A streamcut is considered greater than if for all key ranges the segment/offset in one streamcut is ahead of 
     * second streamcut. 
```

shared/controller-api/src/main/proto/Controller.proto
shiveshr: UNKNOWN = 0 in protobuf allows future safety of enums evolution .. if reading protobuf messing written with new enum value using an old schema, protobuf sets new fields to unknown. 

if we want to support NONE, that should be a separate enum value. if you think we want to support "none" i could add that. but logically none is equivalent to providing limits as null.

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
shiveshr: unknown, as explained, is not a valid use case. it is a protobuf enum evolution safety measure to allow for new values to be mapped to unknown. 

if we want NONE, that should be added separately. However, i think a limit with null on min or max or both is sufficient to provide that and we may not need a separate enum type NONE

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
pbelgundi, line 212: Add NONE here to handle the case where user does not want to have a fallback policy.

shared/controller-api/src/main/proto/Controller.proto
pbelgundi: Yes we need to provide a NONE value, because a user could choose to not set any min/max limits..... true CBR...

pbelgundi: > UNKNOWN = 0 in protobuf allows future safety of enums evolution .. when reading a protobuf message using an older enum schema, any enum that is not understood is set to unknown.
> 
> if we want to support NONE, that should be a separate enum value. if you think we want to support "none" i could add that. but logically none is equivalent to providing limits as null.

Okay.

client/src/main/java/io/pravega/client/stream/RetentionPolicy.java
```diff
@@ -57,6 +62,43 @@ public static RetentionPolicy byTime(Duration duration) {
...
+        return RetentionPolicy.builder().retentionType(RetentionType.CONSUMPTION)
+                              .consumptionLimits(ConsumptionLimits.builder().type(type)
+                                                                  .minValue(minLimit)
+                                                                  .maxValue(maxLimit)
+                                                                  .build()).build();
+    }
+    
+    @Data
+    @Builder
+    public static class ConsumptionLimits {
```
pbelgundi, line 88: Here we assume the user always wants to specify both values - min and max for a Stream.
Some users may just want to set a min value but not the max or vice versa.
It may be good to assign some placeholder value for a limit users don't want to exercise (perhaps -1 ??)

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
pbelgundi: The design allows a user to skip specifying the max and min limits and so there should be an ENUM to represent that option, perhaps a NONE.

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
pbelgundi: Yes NONE is same as min=null and max=null so NONE can be skipped.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -78,18 +79,23 @@
...
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
```
pbelgundi, line 96: Using updateStream() API when a user tries to switch the Retention Policy for an existing Stream( Stream created prior to 0.9) from non-CBR to CBR, we need to invoke `createSubscribersRecordIfAbsent` to make sure the SubscribersSet Record is created in the Stream metadata table. Without this add/createSubscriber would fail for existing Streams.

pbelgundi (changes requested):

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -507,32 +514,283 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+            if (lowerBound.containsKey(segment)) {
+                if (lowerBound.get(segment) > offset) {
+                    lowerBound.put(segment, offset);
+                }
+            } else {
+                Map<StreamSegmentRecord, Long> predecessors = new HashMap<>();
+                Map<StreamSegmentRecord, Long> successors = new HashMap<>();
+                lowerBound.forEach((s, o) -> {
+                    if (s.overlaps(segment)) {
+                        if (s.segmentId() < segment.segmentId()) {
```
shrids, line 700: Q: We can use the getCreationEpoch() here right?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -250,7 +351,8 @@ private boolean greaterThan(Map<Long, Long> cut1, Map<StreamSegmentRecord, Integ
      */
     @Override
     public CompletableFuture<Void> startUpdateConfiguration(final StreamConfiguration newConfiguration) {
-        return getVersionedConfigurationRecord()
+        return createSubscribersRecordIfAbsent()
```
pbelgundi, line 354: With this the SubscribersSet record gets added only when updateStream() is invoked.
However,  if the client tries to addSubscriber() to a Stream, without changing the retention policy using updateStream, it will lead to an exception.
Perhaps it may be good to add `createSubscribersRecordIfAbsent()` here:
https://github.com/pravega/pravega/blob/master/controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java#L528

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -242,6 +248,101 @@ private boolean greaterThan(Map<Long, Long> cut1, Map<StreamSegmentRecord, Integ
...
+                                            return 1;
+                                        } else {
+                                            return -1;
+                                        }
+                                    });
+                                });
+                    };
+
+                    // binary search retention set. 
+                    return binarySearch(0, size, index -> {
```
pbelgundi, line 304: Instead of re-implementing `binarySearch` could we use Collections.binarySearch() ?

pbelgundi (changes requested):

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
```diff
@@ -1020,6 +1021,446 @@ public void sizeBasedRetentionStreamTest() throws Exception {
         // endregion
         // endregion
     }
+    
+    @Test(timeout = 30000)
+    public void consumptionBasedRetentionSizeLimitTest() throws Exception {
```
pbelgundi, line 1026: Do these tests cover all cases listed here:
https://github.com/pravega/pravega/blob/bd7badf9bd8d1dd457baf8014f4716b0b468f61c/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java#L608

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
```diff
@@ -1020,6 +1021,446 @@ public void sizeBasedRetentionStreamTest() throws Exception {
...
+        truncationRecord = streamStorePartialMock.getTruncationRecord(SCOPE, stream1, null, executor).join();
+        assertEquals(truncationRecord.getObject().getStreamCut().get(0L).longValue(), 21L);
+        assertEquals(truncationRecord.getObject().getStreamCut().get(1L).longValue(), 19L);
+        assertTrue(truncationRecord.getObject().isUpdating());
+        streamStorePartialMock.completeTruncation(SCOPE, stream1, truncationRecord, null, executor).join();
+        // endregion
+    }
+    
+    @Test(timeout = 30000)
+    public void consumptionBasedRetentionTimeLimitTest() throws Exception {
```
pbelgundi, line 1161: Do these tests cover all cases listed here:
https://github.com/pravega/pravega/blob/bd7badf9bd8d1dd457baf8014f4716b0b468f61c/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java#L659

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
```diff
@@ -1020,6 +1021,446 @@ public void sizeBasedRetentionStreamTest() throws Exception {
...
+        // subscriber lowerbound is 0/22, 1/21 
+        // this overlaps with max. so truncate at streamcut
+        truncationRecord = streamStorePartialMock.getTruncationRecord(SCOPE, stream1, null, executor).join();
+        assertEquals(truncationRecord.getObject().getStreamCut().get(0L).longValue(), 22L);
+        assertEquals(truncationRecord.getObject().getStreamCut().get(1L).longValue(), 21L);
+        assertTrue(truncationRecord.getObject().isUpdating());
+        streamStorePartialMock.completeTruncation(SCOPE, stream1, truncationRecord, null, executor).join();
+        // endregion
+    }
+
```
pbelgundi, line 1301: We need to add tests to cover the following corner cases:
1. CBR Retention Policy does not specify any max/min limit (so they are at their default values)
2. Stream does not have any subscribers (# of subscribers = 0 and/or no subscriber has a valid StreamCut... there have been addSubscriber() calls but no updateSubscriberStreamCut() calls so far)
3. Stream has few subscribers with empty StreamCuts (StreamCuts have not been updated from client so far)

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: fixed

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: creation epoch is extracted from the segment id itself. we create segmentid as "32byte epoch + 32 byte segment number".

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: thats a good suggestion. it keeps the change in context. i will also add the comment that this is only for streams prior to 0.9 being changed to CBR policy

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: this is an asynchronous binary search implementation. its working with futures :)

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
shiveshr: 1. is handled via the CBR policy constructor itself. if nothing is specified the max is long.max and min is defaulted to 0.

2. and 3. 
if one or more subscribers have not updated their streamcut then they are effectively at the mercy of "min". 
should we wait for all subscribers to have reported at least one streamcut? if that is preferred i will make the change.

pbelgundi: 1. I understand that default max==LONG_MAX and default min==0, I'm suggesting having a test case that invokes CBR Stream truncation without these values being explicitly set, just to cover the scenario where customers may not set these limits. I know it should work, but having a test case to validate would be good.
2 & 3 
> if one or more subscribers have not updated their streamcut then they are effectively at the mercy of "min".
should we wait for all subscribers to have reported at least one streamcut? if that is preferred i will make the change.

Yes this is by design, but need to have tests to cover this, so that we know it works as expected.

pbelgundi: >should we wait for all subscribers to have reported at least one streamcut? if that is preferred i will make the change.

No we should not wait for any subscriber to update their Stream cut. Truncation can happen with 0 subscribers/valid-streamcuts ....its just that it will happen at the max value instead of the lowerBound.
We just need to make sure these corner cases are tested

shiveshr: i will add test to cover Long.Max and 0 as max and min respectively.

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
shiveshr: yes

controller/src/test/java/io/pravega/controller/task/Stream/StreamMetadataTasksTest.java
shiveshr: yes

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -507,32 +514,296 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                                     // truncate at the lower bound 
+                                     return CompletableFuture.completedFuture(lowerBound);
+                                 } else {
+                                     return CompletableFuture.completedFuture(null);
+                                 }
+                             }
+                         });
+    }
+
+    private Map<Long, Long> computeSubscribersLowerBound(List<Map<StreamSegmentRecord, Long>> subscribers) {
```
pbelgundi, line 698: In the case where a Stream has only a single Subscriber, there is no need to compute a "lower bound" and we could short circuit this computation and directly return the Stream-Cut  corresponding to that Subscriber.

pbelgundi: It may be good to add a test case covering this scenario.

pbelgundi (approved):
LGTM other than 1 small optimization.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: created an issue to tackle it later. https://github.com/pravega/pravega/issues/5349

pbelgundi: It should be a small change. I think we should do it as part of this PR itself.

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: the reason is "lowerbound" is Map<Long, Long> while subscriber's streamcut is Map<StreamSegmentRecord, Long> so there is one pass needed to convert even the single subscriber to Map<Long, Long>.
so the suggested optimization will not yield anything. 

what we can do instead is optimize the code a bit to not have to deal with these different types of maps but unify it.

pbelgundi: Ok. Got it.

pbelgundi (approved):
LGTM

client/src/main/java/io/pravega/client/stream/RetentionPolicy.java
```diff
@@ -57,6 +63,45 @@ public static RetentionPolicy byTime(Duration duration) {
      * @return Retention policy object.
      */
     public static RetentionPolicy bySizeBytes(long size) {
-        return new RetentionPolicy(RetentionType.SIZE, size);
+        return RetentionPolicy.builder().retentionType(RetentionType.SIZE).retentionParam(size).build();
+    }
+    
+    /**
+     * Create a retention policy to configure a stream to truncate a stream
+     * according to positions of subscribed reader groups.
```
tkaitchuck, line 71: ` subscribed reader groups.` is not a concept that has be introduced.

Issue #666, opened 2017-03-04
Add unit tests for RocksDBCache


andreipaduroiu:
**Problem description**
Add unit tests for RocksDBCache using the same pattern used for Storage (InMemoryStorage/HDFSStorage) and DurableDataLog (InMemory vs DistributedLog).

**Problem location**
Unit tests.

PR #5295, opened 2020-11-03
Issue 5294: Bookkeeper Client may cache old DNS entries for too long

RaulGracia:
**Change log description**  
(2-3 concise points about the changes in this PR. When committing this PR, the committer is expected to copy the content of this section to the merge description box)

**Purpose of the change**  
(_e.g._, Fixes #666, Closes #1234)

**What the code does**  

**How to verify it**  

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
@@ -30,6 +30,8 @@ import io.pravega.segmentstore.storage.mocks.InMemoryDurableDataLogFactory;
 import io.pravega.shared.metrics.MetricsConfig;
 import io.pravega.shared.metrics.MetricsProvider;
 import io.pravega.shared.metrics.StatsProvider;
+
+import java.security.Security;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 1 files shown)


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
```diff
@@ -82,6 +84,8 @@ private ServiceBuilder createServiceBuilder() {
 
     public void start() throws Exception {
         Exceptions.checkNotClosed(this.closed, this);
+        // Make sure that the JVM is not caching DNS entries forever.
+        Security.setProperty("networkaddress.cache.ttl", "10");
```
eolivelli, line 88: I am not sure this would have effect. Probably it should be a command line argument for java

RaulGracia: Thanks @eolivelli, looking at some documentation I have seen that both methods are similar (for instance https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-jvm-ttl.html). We will look whether this takes effect or not, but in theory it should.

RaulGracia:
Closing, as team decided to check this path using JVM options instead. cc/ @pavanbhat21

PR #5299, opened 2020-11-04
Issue 5298: Use Bookkeeper client config to mitigate FrameTooLongException

### LINKED ISSUE #5298: Use Bookkeeper client config to mitigate FrameTooLongException
**Problem description**
Under high pressure situations, we observe Bookies crashing with `FrameTooLongException`s. Apparently, this is a problem in which the Bookkeeper client sends (via Netty) a piece of data larger than what the server expects. The result of this issue is that Bookies permanently crash in most cases.

There is a configuration parameter in the Bookkeeper client that defines the maximum frame size to be send to Bookkeeper. We need to explore whether this flag mitigates the aforementioned problem or eliminates it completely.

**Problem location**
Bookkeeper client config.

Use `setNettyMaxFrameSizeBytes` in Bookkeeper client and experiment whether it improves the original problem.

RaulGracia:
**Change log description**  
Add setting in Bookkeeper client configuration to limit the length of Netty data frames to the max size of messages in Pravega.

**Purpose of the change**  
Fixes #5298.

**What the code does**  
The new configuration setting `setNettyMaxFrameSizeBytes` is supposed to add a hard limit on the client Netty data frames to be sent to Bookkeeper. The purpose of this PR is to align this setting to the maximum size of a Bookkeeper event, so we have more control on the size of events sent to Bookies. The purpose is to reduce crashes on Bookies under high load situations related to messages being inadvertently too long (i.e., `FrameTooLongException`). 

**How to verify it**  
It has been tested in long run experiments with slow drives that no `FrameTooLongException` has been detected so far. Prior to that change, the problem was consistently reproduced.


andreipaduroiu (approved):

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
```diff
@@ -142,7 +142,9 @@ private BookKeeper startBookKeeperClient() throws Exception {
                 .setGetBookieInfoTimeout(readTimeout)
                 .setEnableDigestTypeAutodetection(true)
                 .setClientConnectTimeoutMillis((int) this.config.getZkConnectionTimeout().toMillis())
-                .setZkTimeout((int) this.config.getZkConnectionTimeout().toMillis());
+                .setZkTimeout((int) this.config.getZkConnectionTimeout().toMillis())
+                .setNettyMaxFrameSizeBytes(BookKeeperConfig.MAX_APPEND_LENGTH)
```
andreipaduroiu, line 146: This value is smaller than the actual frame length. The BK Client will wrap our data in an envelope which will increase this size. If this value is passed per-se to Netty, I don't understand how the system could still work. That's unless there's other code in the BK client to increase this value to account for the extra envelope size. 

Would you mind checking in the BK client code to see how this is handled?

andreipaduroiu (changes requested):
See above.

RaulGracia:
Closing, as this mechanism only involves the client decoding, not the encoding. This means that the server is exposed anyway to large messages from the Bookkeeper clients (I could reproduce the `FrameTooLongException` despite this setting).

PR #5304, opened 2020-11-05
Issue 5114: Update the StreamCut for lastCheckpoint to Controller, once reader has moved past it

### LINKED ISSUE #5114: Client code changes for auto publishing truncation StreamCut to Controller when autoTruncateAtastCheckpoint=true
**Problem description**
A Reader Group reading from a Stream, may want its reads to impact data retention in the Stream (consumption based retention) or just do reads  without impacting Stream's data retention
If a Reader Group wants only unconsumed data to be retained in the Stream it needs to notify the Controller of its position in the Stream till which it has competed consuming data. This  position can be manually supplied by the user or inferred from the last checkpoint, once the Reader has attempted reading beyond the last checkpoint in the Stream.

**Problem location**
Client

**Suggestions for improvement**
This issue includes changes to Reader Group Configuration for enabling consumption based retention of data on Stream as well as changes for publishing retention Stream-Cuts to Controller.

pbelgundi:

**Change log description**  
This PR has code changes for publishing truncation Stream-Cut to Controller which can be used for Consumption Based Retention of data in the Stream. The truncation StreamCut can be provided manually by the user by invoking Client API `updateRetentionStreamCut(Map<Stream, StreamCut> scMap)` or auto-published when a Reader reads past the Last Completed Checkpoint.

**Purpose of the change**  
Fixes #5114

**What the code does**  
- Adds a new enum type to ReaderGroupConfig with values as follows:
NONE - Set when the reader's positions of this Reader Group should not impact Stream retention. 
MANUAL_RELEASE_AT_USER_STREAMCUT - User provides a Stream-Cut to mark position in the Stream till which data is consumed using {@link ReaderGroup#updateRetentionStreamCut(java.util.Map) } API.
AUTOMATIC_RELEASE_AT_LAST_CHECKPOINT - Stream-Cut of the last completed checkpoint is used as the retention Stream-Cut.

- Adds a new public API on client : `updateRetentionStreamCut(Map<Stream, StreamCut> scMap)` used to publish user Stream-Cuts to Controller

- Adds a new method `updateTruncationStreamCut()` that publishes the Stream-Cuts corresponding to the last completed checkpoint to Controller as truncation Stream-Cuts, when a Reader tries reading past the last checkpoint.

**How to verify it**  
All Unit and Integration Tests should pass.


client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +49,28 @@
 
     private final int maxOutstandingCheckpointRequest;
 
+    private final ReaderGroupRetentionConfig retentionConfig;
+
+    public enum ReaderGroupRetentionConfig {
```
shiveshr, line 54: this config also seems to be part of anirudh's PR. can we ensure that we consolidate it.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +49,28 @@
...
+        NO_CONSUMPTION_BASED_TRUNCATION(false, false),
```
shiveshr, line 55: i gave the same comment on the other PR -- do we want to make it an enum?
also "truncate at user streamcut" as the enum value is confusion from the end users perspective.. what is a user streamcut?

i suggested that we use two booleans. 
1. `subscribeForRetention`
2. `acknowledge last checkpoint`
and to ensure that `2` is only available to be set when `1` is true, we can use builder methods and hide the setters for individual booleans.

for example:
```
ReaderGroupConfigBuilder {
...
   ReadGroupConfigBuilder subscribe(boolean acknowledgeAtCheckpoints) {
       this.subscribe = true;
       this.acknowledgeAtCheckpoints = acknowledgeAtCheckpoints;
       return this;
   }
}
```

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
```diff
@@ -258,10 +259,12 @@ private PositionInternal getCurrentPosition() {
     private String updateGroupStateIfNeeded() throws ReaderNotInReaderGroupException {
         PositionInternal position = null;
         if (atCheckpoint != null) {
+            // process the checkpoint we're at
             position = refreshAndGetPosition();
             groupState.checkpoint(atCheckpoint, position);
             log.info("Reader {} completed checkpoint {}", groupState.getReaderId(), atCheckpoint);
             releaseSegmentsIfNeeded(position);
+            groupState.updateTruncationStreamCutIfNeeded();
```
shiveshr, line 267: are all readers reporting this streamcut? what happens if this fails 
1. because of intermittent issues
2. because the reader was reporting a stale value and a newer value was already published to controller

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -462,7 +472,27 @@ void checkpoint(String checkpointName, PositionInternal lastPosition) throws Rea
...
+        // here add check if RG is Subscriber and autoPublishTruncationStreamCut = true
+        if (state.getConfig().getRetentionConfig().equals(ReaderGroupConfig.ReaderGroupRetentionConfig.TRUNCATE_AT_LAST_CHECKPOINT)
+        && !state.getCheckpointState().isLastCheckpointPublished()) {
+            // we get here only when this is the reader that has completed the lastCheckpoint
+            Optional<Map<Stream, Map<Segment, Long>>> cuts = state.getPositionsForLastCompletedCheckpoint();
+            if (!cuts.isPresent()) {
+                throw new CheckpointFailedException("Could not get positions for last checkpoint.");
+            }
+            Set<StreamCut> streamCuts = cuts.get().entrySet().stream()
+                    .map(entry -> new StreamCutImpl(entry.getKey(), entry.getValue()))
```
shiveshr, line 488: map followed by foreach is two passes.. 
we can simply do it in one pass with foreach where you do the transformation and then call update subscriber streamcut.

shiveshr (changes requested):

client/src/main/java/io/pravega/client/stream/impl/EventStreamReaderImpl.java
pbelgundi: Every reader would invoke that method, but it executes only of this condition is true:
```
if (state.getConfig().getRetentionConfig().equals(ReaderGroupConfig.RetentionConfig.CBR_LAST_CHECKPOINT)
        && !state.getCheckpointState().isLastCheckpointPublished()) {
```
isLastCheckpointPublished() is true by default and is changed to `false` only by the last reader completing a checkpoint :
https://github.com/pravega/pravega/blob/76e3a5be72635262f95b48d85880fe27954f4e12/client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java#L154

So ideally this would only be executed by the lastReader completing the checkpoint and immediately after the streamcut is published to controller, we change the value of this variable back to "true" in state synchronizer, so no other Reader would attempt to publish again. If the streamcut publish from this reader fails, the variable value remains at false and so another Reader would publish the Stream-Cut.

pbelgundi: > because the reader was reporting a stale value and a newer value was already published to controller.

With regular fetchUpdates() being done I don't see the possibility of a reader trying to publish a stale checkpoint value and eventhat does happen the update would fail, based on streamcut validation checks added on Controller.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Sure. Once his PR is merged I will remove this part and adjust my code to work with his changes.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: I see enum as a superior choice here because of the following reasons:
1. We don't allow users to make invalid choices (ex: subscriber=false, autoTruncate=true) 
2. Indicates to the user that both these flags are linked to the same functionality.
3. User Friendly - rather than asking users to set flags we let them choose a friendly enum name based on their functional requirements the internal implementation details of what flags we use to implement this are hidden from them.

pbelgundi: Changed Enum names to indicate these are for ConsumtionBasedRetention

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: Ok. fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -304,6 +334,29 @@ private void write01(ReaderGroupConfig object, RevisionDataOutput revisionDataOu
             revisionDataOutput.writeMap(object.endingStreamCuts, keySerializer, valueSerializer);
             revisionDataOutput.writeInt(object.getMaxOutstandingCheckpointRequest());
         }
+
+        private void read02(RevisionDataInput revisionDataInput, ReaderGroupConfigBuilder builder) throws IOException {
```
tkaitchuck, line 338: This logic is incorrect. You don't need to re-serialize all the previous fields.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -255,11 +265,35 @@ private void read00(RevisionDataInput input, CheckpointStateBuilder builder) thr
         private void write00(CheckpointState object, RevisionDataOutput output) throws IOException {
             ElementSerializer<String> stringSerializer = RevisionDataOutput::writeUTF;
             ElementSerializer<Long> longSerializer = RevisionDataOutput::writeLong;
+            ElementSerializer<Boolean> boolSerializer = RevisionDataOutput::writeBoolean;
```
tkaitchuck, line 268: Why is the old version being changed?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -462,7 +470,25 @@ void checkpoint(String checkpointName, PositionInternal lastPosition) throws Rea
     boolean isCheckpointSilent(String atCheckpoint) {
         return sync.getState().getCheckpointState().isCheckpointSilent(atCheckpoint);
     }
-    
+
+    @SneakyThrows(CheckpointFailedException.class)
```
tkaitchuck, line 474: This is not a good use of sneakythrows. (See our contributing guidelines)

tkaitchuck (changes requested):
The bor requires additional test coverage

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: This exception is thrown as a SneakyThrows in all places where we attempt to get the StreamCut from a checkpoint. 
As per Contributing guidelines SnealyThrows should be used "_For the specific case where you know with certainty an exception cannot occur_" and this is that case.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Not setting those values causes this test case to fail:
```
io.pravega.client.stream.impl.SerializationTest > testReaderGroupInit FAILED
 java.lang.AssertionError: expected:<ReaderGroupState.CompactReaderGroupState(config=ReaderGroupConfig(groupRefreshTimeMillis=261, automaticCheckpointIntervalMillis=-1, startingStreamCuts={StreamImpl(scope=MnPkB, streamName=Klytg)=UNBOUNDED}, endingStreamCuts={StreamImpl(scope=MnPkB, streamName=Klytg)=UNBOUNDED}, maxOutstandingCheckpointRequest=3, retentionConfig=NONE), checkpointState=CheckpointState { ongoingCheckpoints: [UqObC, dnUVN],  readersBlockingEachCheckpoint: {jEdbi=[]} }, distanceToTail={crzYB=-2509738709635323033}, futureSegments={}, assignedSegments={uiHLL={}}, unassignedSegments={SegmentWithRange(segment=pUzSq/jfyqa/65.#epoch.0, range=SegmentWithRange.Range(low=0.06767117977142334, high=0.7583411931991577))=-8086877361167159070, SegmentWithRange(segment=xGabw/OmZmb/21.#epoch.0, range=SegmentWithRange.Range(low=0.5213870406150818, high=0.6767080426216125))=1153959188627795253}, lastReadPosition={SegmentWithRange(segment=vknHM/DqlWO/28.#epoch.0, range=SegmentWithRange.Range(low=0.16790086030960083, high=0.20448356866836548))=9050953671715088000}, endSegments={uSSnW/OGpLi/75.#epoch.0=3759285871800690071, MYTxV/xCRdJ/18.#epoch.0=-5404048114723376066})> but was:<ReaderGroupState.CompactReaderGroupState(config=ReaderGroupConfig(groupRefreshTimeMillis=261, automaticCheckpointIntervalMillis=-1, startingStreamCuts={StreamImpl(scope=MnPkB, streamName=Klytg)=UNBOUNDED}, endingStreamCuts={StreamImpl(scope=MnPkB, streamName=Klytg)=UNBOUNDED}, maxOutstandingCheckpointRequest=3, retentionConfig=NONE), checkpointState=CheckpointState { ongoingCheckpoints: [UqObC, dnUVN],  readersBlockingEachCheckpoint: {jEdbi=[]} }, distanceToTail={crzYB=-2509738709635323033}, futureSegments={}, assignedSegments={uiHLL={}}, unassignedSegments={SegmentWithRange(segment=xGabw/OmZmb/21.#epoch.0, range=SegmentWithRange.Range(low=0.5213870406150818, high=0.6767080426216125))=1153959188627795253, SegmentWithRange(segment=pUzSq/jfyqa/65.#epoch.0, range=SegmentWithRange.Range(low=0.06767117977142334, high=0.7583411931991577))=-8086877361167159070}, lastReadPosition={SegmentWithRange(segment=vknHM/DqlWO/28.#epoch.0, range=SegmentWithRange.Range(low=0.16790086030960083, high=0.20448356866836548))=9050953671715088000}, endSegments={MYTxV/xCRdJ/18.#epoch.0=-5404048114723376066, uSSnW/OGpLi/75.#epoch.0=3759285871800690071})>
        at io.pravega.client.stream.impl.SerializationTest.verify(SerializationTest.java:330)
        at io.pravega.client.stream.impl.SerializationTest.testReaderGroupInit(SerializationTest.java:266)
```

shiveshr (approved):
LGTM except the serialization change requested by tom.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
tkaitchuck: Oh. I didn't realize it was impossible. In that case, it's fine.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -259,7 +268,28 @@ private void write00(CheckpointState object, RevisionDataOutput output) throws I
             output.writeCollection(object.checkpoints, stringSerializer);
             output.writeMap(object.uncheckpointedHosts, stringSerializer, (out, hosts) -> out.writeCollection(hosts, stringSerializer));
             output.writeMap(object.checkpointPositions, stringSerializer, (out, map) -> out.writeMap(map, segmentSerializer, longSerializer));
+        }
+
+        private void read01(RevisionDataInput input, CheckpointStateBuilder builder) throws IOException {
```
tkaitchuck, line 273: Same situation here.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
```diff
@@ -206,7 +215,7 @@ CheckpointState copy() {
         Map<String, Map<Segment, Long>> cpps = new HashMap<>();
         checkpointPositions.forEach((cp, pos) -> cpps.put(cp, new HashMap<>(pos)));
         Map<Segment, Long> lcp = lastCheckpointPosition == null ? null : new HashMap<>(lastCheckpointPosition);
-        return new CheckpointState(cps, ucph, cpps, lcp);
+        return new CheckpointState(cps, ucph, cpps, lcp, true);
```
shrids, line 218: If there is compaction then the `isLastCheckpointPublished` would be set to true, even though it was never published. This would be an issue right ?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -462,7 +470,25 @@ void checkpoint(String checkpointName, PositionInternal lastPosition) throws Rea
...
+    public void updateTruncationStreamCutIfNeeded() {
+        ReaderGroupState state = sync.getState();
+        // here add check if RG is Subscriber and autoPublishTruncationStreamCut = true
+        if (state.getConfig().getRetentionConfig().equals(ReaderGroupConfig.RetentionConfig.CONSUMPTION_BASED_AT_LAST_CHECKPOINT)
+        && !state.getCheckpointState().isLastCheckpointPublished()) {
+            // we get here only when this is the reader that has completed the lastCheckpoint
+            Optional<Map<Stream, Map<Segment, Long>>> cuts = state.getPositionsForLastCompletedCheckpoint();
+            if (!cuts.isPresent()) {
```
shrids, line 482: nit: cuts.isEmpty()

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
pbelgundi: This was existing code and ideally I should not be changing it as part of this PR. But will fix.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -462,7 +470,25 @@ void checkpoint(String checkpointName, PositionInternal lastPosition) throws Rea
...
+                throw new CheckpointFailedException("Could not get positions for last checkpoint.");
+            }
+            cuts.get().entrySet().stream().forEach(entry ->
```
shrids, line 485: Nit: No need of stream(). `cuts.get().entrySet().forEach(..)`

shrids (commented):
LGTM apart from one question related to compaction.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -177,6 +195,17 @@ public ReaderGroupConfigBuilder startFromCheckpoint(final Checkpoint checkpoint)
            return this;
        }
 
+       /**
+        * Set the retention config for the {@link ReaderGroup}.
```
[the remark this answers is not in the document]
pbelgundi: It applies only if CBR is enabled on the Stream on Controller. Thanks for bringing this up. Will improve the comment.

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
pbelgundi: Good catch. Yes, thats' a mistake. Fixed it now.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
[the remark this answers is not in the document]
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/impl/CheckpointState.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: fixed

shrids (approved):

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,37 @@
 
     private final int maxOutstandingCheckpointRequest;
 
+    private final StreamDataRetention retentionConfig;
+
+    /**
+     * The values of this enum indicate if a ReaderGroup chooses Consumption based data retention for Streams it reads from.
```
tkaitchuck, line 56: I hate to be pedantic, but this is public facing Javadoc, so I really want to nail it.
"this enum indicate if a ReaderGroup chooses Consumption based data retention"
This is not strictly accurate. The setting is actually on the stream so the reader group cannot "choose" consumption based retention. This only configures how this reader group interacts with the setting.

Let's make this explanation as precise as possible, and probably link to the Stream configuration setting so that users can locate it easily.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,37 @@
...
+     * For Consumption based retention to work, in addition to setting the appropriate value here,
```
tkaitchuck, line 57: "For Consumption based retention to work..."
As this is specified on the ReaderGroup and the group may have multiple streams with different settings this should be more explicit. Spell out what happens if it is not enabled on the stream.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,37 @@
...
+     * the Streams' Retention Policy must be set to 'CONSUMPTION'.
+     *
+     * NO_IMPACT - Read Positions of Readers do not impact Stream truncation.
```
tkaitchuck, line 60:
```suggestion
     * NO_IMPACT - Read Positions of Readers do not impact Stream retention.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,37 @@
...
+     * CONSUMPTION_BASED_USER_STREAMCUT - ConsumptionBasedRetention, user provides the StreamCut for truncation.
```
tkaitchuck, line 61: As this setting only affects consumption based retention, having it in the name is somewhat redundant.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -177,6 +204,20 @@ public ReaderGroupConfigBuilder startFromCheckpoint(final Checkpoint checkpoint)
            return this;
        }
 
+       /**
+        * Set the retention config for the {@link ReaderGroup}.
+        * For Consumption based retention of data in the Stream, this field should be set to
```
tkaitchuck, line 209: "should be set to" is not helpful. 
Specify what happens with each setting including NO_IMPACT

pbelgundi: I have addressed that in the javadoc for the enum.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Ok. changed this.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
 
     private final int maxOutstandingCheckpointRequest;
 
+    private final StreamDataRetention retentionConfig;
+
+    /**
+     * If a Stream's Retention Policy {@link RetentionPolicy} is set to 'CONSUMPTION' based,
+     * the Reader Group needs to register with the Stream as a Subscriber
```
tkaitchuck, line 57: `the Reader Group needs to register with the Stream as a Subscriber`
This is simply not true.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
+     * and periodically publish retention StreamCuts either manually or at checkpoints.
+     * This can be done by setting retentionConfig to 'CONSUMPTION_BASED_USER_STREAMCUT' or
+     * 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
+     * If the Stream's RetentionPolicy is not Consumption based, retentionConfig should be 'NO_IMPACT'.
+     *
+     * If the Streams' Retention Policy is set to 'CONSUMPTION', but the ReaderGroup retentionConfig is set at
+     * 'NO_IMPACT', the ReaderGroup will not get registered as a Subscriber with the Stream
```
tkaitchuck, line 64: Please do not use the word "subscriber" in this way, as it will confuse users.
Similarly "register" implies additional follow up questions like "do I need to unregister?", "how do I do that?" etc.
This should only explain what the user needs to know, not how it is implemented.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
 
+    private final StreamDataRetention retentionConfig;
+
+    /**
+     * If a Stream's Retention Policy {@link RetentionPolicy} is set to 'CONSUMPTION' based,
+     * the Reader Group needs to register with the Stream as a Subscriber
+     * and periodically publish retention StreamCuts either manually or at checkpoints.
+     * This can be done by setting retentionConfig to 'CONSUMPTION_BASED_USER_STREAMCUT' or
+     * 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
+     * If the Stream's RetentionPolicy is not Consumption based, retentionConfig should be 'NO_IMPACT'.
```
tkaitchuck, line 61: >  * If the Stream's RetentionPolicy is not Consumption based, retentionConfig should be 'NO_IMPACT'.

"Consumption based" is not clear. Please link to the exact setting explicitly. 
"should be 'NO_IMPACT'" is a meaningless detective. We should say that this setting will have no effect regardless of what it is set to.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
 
     private final int maxOutstandingCheckpointRequest;
 
+    private final StreamDataRetention retentionConfig;
+
+    /**
+     * If a Stream's Retention Policy {@link RetentionPolicy} is set to 'CONSUMPTION' based,
```
tkaitchuck, line 56: "Consumption based" is not clear. Please link to the exact setting explicitly.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
+     * the Reader Group needs to register with the Stream as a Subscriber
+     * and periodically publish retention StreamCuts either manually or at checkpoints.
+     * This can be done by setting retentionConfig to 'CONSUMPTION_BASED_USER_STREAMCUT' or
+     * 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
+     * If the Stream's RetentionPolicy is not Consumption based, retentionConfig should be 'NO_IMPACT'.
+     *
+     * If the Streams' Retention Policy is set to 'CONSUMPTION', but the ReaderGroup retentionConfig is set at
+     * 'NO_IMPACT', the ReaderGroup will not get registered as a Subscriber with the Stream
+     * and retention StreamCuts cannot be published for this ReaderGroup.
+     * This will break the promise of ConsumptionBased Retention.
```
tkaitchuck, line 66: > This will break the promise of ConsumptionBased Retention.

What!?!?

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
+     * If a Stream's Retention Policy {@link RetentionPolicy} is set to 'CONSUMPTION' based,
+     * the Reader Group needs to register with the Stream as a Subscriber
+     * and periodically publish retention StreamCuts either manually or at checkpoints.
+     * This can be done by setting retentionConfig to 'CONSUMPTION_BASED_USER_STREAMCUT' or
+     * 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
+     * If the Stream's RetentionPolicy is not Consumption based, retentionConfig should be 'NO_IMPACT'.
+     *
+     * If the Streams' Retention Policy is set to 'CONSUMPTION', but the ReaderGroup retentionConfig is set at
+     * 'NO_IMPACT', the ReaderGroup will not get registered as a Subscriber with the Stream
+     * and retention StreamCuts cannot be published for this ReaderGroup.
```
tkaitchuck, line 65: Please do not use the word "published" it has other connotations. 

Additionally there is no need to explain the implementation. Stick to explaining the effect on the user's application.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
+
+    /**
+     * If a Stream's Retention Policy {@link RetentionPolicy} is set to 'CONSUMPTION' based,
+     * the Reader Group needs to register with the Stream as a Subscriber
+     * and periodically publish retention StreamCuts either manually or at checkpoints.
+     * This can be done by setting retentionConfig to 'CONSUMPTION_BASED_USER_STREAMCUT' or
+     * 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
+     * If the Stream's RetentionPolicy is not Consumption based, retentionConfig should be 'NO_IMPACT'.
+     *
+     * If the Streams' Retention Policy is set to 'CONSUMPTION', but the ReaderGroup retentionConfig is set at
```
tkaitchuck, line 63: Currently the only retention policies are "time" and "size". Do you mean to modify this file in this PR?

This should link to the actual policy not leave people to track it down themselves.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
+     * 'NO_IMPACT', the ReaderGroup will not get registered as a Subscriber with the Stream
+     * and retention StreamCuts cannot be published for this ReaderGroup.
+     * This will break the promise of ConsumptionBased Retention.
+     *
+     * NO_IMPACT - Read Positions of Readers do not impact Stream truncation/retention.
+     * CONSUMPTION_BASED_USER_STREAMCUT - User provides the StreamCut for truncation using Consumption Based Retention.
```
tkaitchuck, line 69: > User provides the StreamCut for truncation using Consumption Based Retention.

This should link to the API through which it is provided. 
The streamcut is not "for truncation".

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
+     * CONSUMPTION_BASED_AT_LAST_CHECKPOINT - StreamCut corresponding to lastCompletedCheckpoint is published as truncation point for Consumption Based Retention.
```
tkaitchuck, line 70: Not a truncation point.
Here and elsewhere don't say "for Consumption Based Retention" refer to the specific stream configuration policy. Users don't know our PDPs or internal names for things.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -463,7 +470,23 @@ void checkpoint(String checkpointName, PositionInternal lastPosition) throws Rea
     boolean isCheckpointSilent(String atCheckpoint) {
         return sync.getState().getCheckpointState().isCheckpointSilent(atCheckpoint);
     }
-    
+
+    @SneakyThrows(CheckpointFailedException.class)
+    public void updateTruncationStreamCutIfNeeded() {
```
tkaitchuck, line 475: This method does not need to be public.

Shouldn't there be an API method on ReaderGroup that calls this?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -463,7 +470,23 @@ void checkpoint(String checkpointName, PositionInternal lastPosition) throws Rea
...
+        ReaderGroupState state = sync.getState();
+        // here add check if RG is Subscriber and autoPublishTruncationStreamCut = true
+        if (state.getConfig().getRetentionConfig().equals(ReaderGroupConfig.StreamDataRetention.CONSUMPTION_BASED_AT_LAST_CHECKPOINT)
```
tkaitchuck, line 478: Why do we check just `CONSUMPTION_BASED_AT_LAST_CHECKPOINT`
Is it invalid to set `CONSUMPTION_BASED_USER_STREAMCUT` and call this method? If so why? There is no inconsistency.
Additionally if the user is making a mistake if this is set to "no_impact" should it silently do nothing or throw?

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,45 @@
...
+     * */
+    public enum StreamDataRetention {
+        NO_IMPACT(false, false),
+        CONSUMPTION_BASED_USER_STREAMCUT(true, false),
```
tkaitchuck, line 74: "`CONSUMPTION_BASED_USER_STREAMCUT`" reflects how it is being used. This is fine in most code, but this is our public API, so everything should reflect the user's perspective. 
So something like "HOLD_DATA_UNTIL_RELEASED" or similar depending on the name of the api they are calling (this example is assuming the method were to be called "releaseDataBefore(streamCut)" or similar).

tkaitchuck (changes requested):

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: > Is it invalid to set CONSUMPTION_BASED_USER_STREAMCUT and call this method? If so why? 
User needs to set retentionPolicy = 'CONSUMPTION_BASED_USER_STREAMCUT' when user will be providing the truncation StreamCut and does not intend to use the checkpoint StreamCut as truncation point. This has been explained in the enum javadoc

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: That retention type was added by another PR and not available till this morning. I have added a reference to it now.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: That API is not added as part of this PR but as part of another PR. So this javadoc can be updated there.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: I understand users are not expected to read PDPs or design documents.
However, the intent of a Javadoc is to explain how an API works and not to provide documentation for the entire feature. 
For that we need user documentation that explains how Consumption based retention needs to be configured and used in Pravega.
GH issue for the same: https://github.com/pravega/pravega/issues/5350

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: I do not think "HOLD_DATA_UNTIL_RELEASED" is intuitive.
Also it does not bring out the relation of this enum value to Consumption based retention.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: It is true, but an implementation detail. Changed now.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Consumption based has just been explained 2 lines earlier along with the link.
Nevertheless added the link again

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: This has been handled in a separate PR.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: removed word subscriber and register.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: removed the word publish...

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Yes it does break the promise, but I have reworded the same....

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: > Additionally if the user is making a mistake if this is set to "no_impact" should it silently do nothing or throw?

If the user choses NO_IMPACT, when the stream is configured for Consumption Based Retention, the Consumption Based Retention promise is broken and this is made explicit in the enum javadoc

pbelgundi:
> LGTM except the serialization change requested by tom.

Fixed

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: removed 'public' modifier from this method.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: True. Perhaps the enum type ie `StreamDataRetention` could do this? After they all relate to that.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: It is perfectly valid to set `no_impact` and not be a subscriber. So it's not needed.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: > Consumption enum value will be incorrect.

Will it throw an exception? If so document that. If not it's not "invalid", in which case what it does should be described.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: We can't have these in different versions of pervega. It leaves the API in an inconsistent state.
Please combine these PRs.

tkaitchuck: Perhaps these PRs should be going into a feature branch. It would allow us to avoid this sort of inconsistency in master.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: FYI those docs are in the repo https://github.com/pravega/pravega/tree/master/documentation/src/docs
so you can edit them as part of a PR, or in a branch.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
tkaitchuck: >  Consumption Based Retention promise is broken 

Then the API is broken. It needs to either throw an exception because the case is invalid and document that it does this explicitly, or better yet be written in such a way that invalid states cannot be expressed.

tkaitchuck: > We want to publish the stream-cut corresponding to the last checkpoint only when the user has chosen the later option

Why do we want this? It seems like an application could both ask that the retention cut be updated at checkpoints and sometimes provide its own value. It's certainly possible to write such an application. If the user does that now, they will simply get confusing undocumented behavior. A choice needs to be made as to what it does, and that behavior needs to be defined. If it is "invalid" then it should throw, and the exception should stated in the Javadoc. If it is valid then what happens must be specified.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,46 @@
 
     private final int maxOutstandingCheckpointRequest;
 
+    private final StreamDataRetention retentionConfig;
+
+    /**
+     * If a Stream's Retention Policy {@link RetentionPolicy} is set to
+     * 'CONSUMPTION' based retention {@link RetentionPolicy.RetentionType#CONSUMPTION} ,
```
tkaitchuck, line 57:
```suggestion
     * If a Stream's Retention Policy is set to {@link RetentionPolicy.RetentionType#CONSUMPTION} ,
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,46 @@
...
+     * the Reader Group needs to periodically notify StreamCuts to Controller to indicate
+     * the point in the Stream till which it has completed consuming data.
```
tkaitchuck, line 59:
```suggestion
     * the Reader Group may supply StreamCuts which indicate
     * the point through which it no longer requires data be retained.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,46 @@
...
+     * This can be done by setting the retentionConfig in ReaderGroupConfig {@link ReaderGroupConfig}
+     * to 'CONSUMPTION_BASED_USER_STREAMCUT' or 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
```
tkaitchuck, line 61:
```suggestion
     * This can be done automatically at each checkpoint by setting the `retentionConfig` to {@link ReaderGroupConfig#CONSUMPTION_BASED_AT_LAST_CHECKPOINT}.
     * Alternatively the ReaderGroup can request the data be held until {@link  ... } is called by setting `retentionConfig` to {@link ReaderGroupConfig#CONSUMPTION_BASED_USER_STREAMCUT}.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,46 @@
...
+     * If a Stream's Retention Policy is not Consumption based the ReaderGroupConfig should have retentionConfig = 'NO_IMPACT'.
```
tkaitchuck, line 62:
```suggestion
     * If the ReaderGroup does not wish to affect retention {@ReaderGroupConfig#NO_IMPACT} can be used.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,46 @@
...
+     *
+     * Note: It is incorrect to set ReaderGroup retentionConfig = 'NO_IMPACT' when
+     * a Streams' Retention Policy = 'CONSUMPTION' {@link RetentionPolicy.RetentionType#CONSUMPTION},
+     * because with NO_IMPACT the ReaderGroup will not notify consumption Stream-Cuts to Controller
+     * and so Consumption based retention of data in the Stream would not happen.
```
tkaitchuck, line 67:
```suggestion
```
A stream may have multiple reader groups, where one wishes to retain the data explicitly and the other does not care. This is an important usecase.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,46 @@
...
+     *
+     * NO_IMPACT - Read Positions of Readers do not impact Stream truncation/retention.
+     * CONSUMPTION_BASED_USER_STREAMCUT - User provides the StreamCut for truncation using Consumption Based Retention.
+     * CONSUMPTION_BASED_AT_LAST_CHECKPOINT - StreamCut corresponding to lastCompletedCheckpoint is auto-notified as truncation Stream-Cut for Consumption Based Retention.
```
tkaitchuck, line 71: I think we can eliminate these separate bullets if we explain each in the paragraph above.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -177,6 +213,20 @@ public ReaderGroupConfigBuilder startFromCheckpoint(final Checkpoint checkpoint)
            return this;
        }
 
+       /**
+        * Set the retention config for the {@link ReaderGroup}.
+        * For Consumption based retention of data in the Stream, this field should be set to
+        * CONSUMPTION_BASED_USER_STREAMCUT or CONSUMPTION_BASED_AT_LAST_CHECKPOINT.
```
tkaitchuck, line 219:
```suggestion
```
Again it is valid for a stream to enable CBR and for one reader group to not take advantage of it.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -177,6 +213,20 @@ public ReaderGroupConfigBuilder startFromCheckpoint(final Checkpoint checkpoint)
...
+        * Default value NO_IMPACT, is to be used when Stream's RetentionPolicy is not Consumption based.
```
tkaitchuck, line 220:
```suggestion
        * Default value: NO_IMPACT.
```

tkaitchuck (changes requested):

tkaitchuck:
Please update the Javadoc on the StreamConfig to link to the ReaderGroupConfig

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
fpj, line 57: I'm wondering about the use of the term `CONSUMPTION` for retention and whether it is going to be confusing to the user. Would terms like `ACKNOWLEDGED`, `CONFIRMED` or `ASSURED` be any better? I don't feel strongly about it at this point, but I want to have a discussion about it.

fpj (changes requested):

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: What if instead of adding a new type to the RetentionPolicy enum, we just overloaded the methods in the `StreamConfig`?
Instead of: `TIME`, `SIZE`, and `CONSUMPTION` and the methods `bySizeBytes`, `byTimeMillis`, which take a single value and `byConsumption` which takes two longs and a unit which is either `TIME` or `SIZE` we instead have:
just `TIME`, and `SIZE` and then have `withTimeRangeMillis` and `withSizeRangeBytes` which each take a min and max arg for time/size.
In this case calling `bySizeBytes` is equivalent to passing the same value as the min and the max for `withSizeRangeBytes` (except under the hood we can optimize to do fewer RPCs).
This means that all streams are always use CBR (it's just that some have the same max and min) and so we don't have to explain it as a distinct concept, instead we just say the reader group can request data be retained and when they do so it will be retained provided it is less than the stream's configured max.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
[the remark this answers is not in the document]
tkaitchuck: I think of the three options as:
"No need to extend retention for this reader group beyond the configured minimum"
"Extend retention until the readers have read the data (and checkpointed) up to the configured maximum"
"Extend retention up to the maximum unless explicitly released"
The call used by the third option also makes sense in the case of the second. It's just not needed.

In terms of names, we could try to express this as:

`RETAIN_UNTIL_CHECKPOINT` and `RETAIN_UNTIL_RELEASE`
or if not "release" whatever verb we want to associate with the method call, like maybe `UPDATE`

Or we could try to get the idea of "retain data until" out into the enum class name and just have single word alternatives. Like just "CHECKPOINT".

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
pbelgundi: > It seems like an application could both ask that the retention cut be updated at checkpoints and sometimes provide its own value.

This can be disallowed, by the Client API that accepts truncation Stream-Cuts for CBR, by checking  ReaderGroupConfig.retentionConfig == CONSUMPTION_BASED_USER_STREAM_CUT. 

I don;t understand how  the check "retentionConfig == CONSUMPTION_BASED_AT_LAST_CHECKPOINT" breaks anything. It is an absolutely necessary check, because we want to publish that stream-cut only when the user has chosen this option CONSUMPTION_BASED_AT_LAST_CHECKPOINT.

pbelgundi: > If the user does that now, they will simply get confusing undocumented behavior. 

There is nothing undocumented or confusing here.
The system behaves as per the retentionConfig option chosen by the user.
Please read through the PDP if you have questions/concerns:
https://github.com/pravega/pravega/wiki/PDP-47:-Pravega-Streams:-Consumption-Based-Retention

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Yes it is completely valid. That would be a non-subscriber Reader Group.
See: https://github.com/pravega/pravega/wiki/PDP-47:-Pravega-Streams:-Consumption-Based-Retention

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,46 @@
...
+    /**
+     * If a Stream's Retention Policy {@link RetentionPolicy} is set to
+     * 'CONSUMPTION' based retention {@link RetentionPolicy.RetentionType#CONSUMPTION} ,
+     * the Reader Group needs to periodically notify StreamCuts to Controller to indicate
+     * the point in the Stream till which it has completed consuming data.
+     * This can be done by setting the retentionConfig in ReaderGroupConfig {@link ReaderGroupConfig}
+     * to 'CONSUMPTION_BASED_USER_STREAMCUT' or 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
+     * If a Stream's Retention Policy is not Consumption based the ReaderGroupConfig should have retentionConfig = 'NO_IMPACT'.
+     *
+     * Note: It is incorrect to set ReaderGroup retentionConfig = 'NO_IMPACT' when
```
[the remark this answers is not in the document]
pbelgundi: Sorry. My bad. That comment is wrong and I will change it.
A ReaderGroup **can** set retentionConfig=NO_IMPACT, when the Stream has RetentionPolicy set to "CONSUMPTION" (CBR). What that means is the Reader Group is a non-subscriber and does not want its reads to impact Stream truncation.
So we cannot throw an exception when the Stream policy is CBR but RG Config says NO_IMPACT.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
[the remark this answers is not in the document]
pbelgundi: I'd not want to hold back the merge of this PR until we figure out the most "perfect" terms.
Name changes could be handled using another PR.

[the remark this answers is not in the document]
pbelgundi: I would not want to delay the merge of this PR till we've agreed on the perfect names.
Name refactoring could be done using another PR.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: > I was assuming that doing so would be fine and would work as expected on the stream with it enabled and have no effect on the other stream. 

Yes. Thats right. 
It is not possible to provide the explanation for all these varied combinations of Stream and Reader Group options and their impact in the javadoc and still keep it succinct. 
I will add these details in the User Documentation.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Just to be able to reference it in the javadoc, I have added this API to this PR now and added a link in the javadoc.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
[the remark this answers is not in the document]
fpj: It is not a matter of finding the perfect name, it is about providing and describing the right semantics. At the same time, from the title and description, this PR is about updating the controller with a stream cut. As such, updating configuration shouldn't be part of it, and as such, we can remove such changes from this PR and address the configuration change separately.

pbelgundi:
> LGTM apart from one question related to compaction.

Resolved

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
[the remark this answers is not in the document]
pbelgundi: > As such, updating configuration shouldn't be part of it,

What is config update?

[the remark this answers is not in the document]
pbelgundi: As discussed, created a new issue to track this name changes: https://github.com/pravega/pravega/issues/5356

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Filed issue for this: https://github.com/pravega/pravega/issues/5356

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Filed : https://github.com/pravega/pravega/issues/5356 for this

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: I have stated both in the javadoc now.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Keeping these as they are not explained in the paragraph above.

pbelgundi:
> Please update the Javadoc on the StreamConfig to link to the ReaderGroupConfig

updated

fpj (approved):

pbelgundi:
> Please update the Javadoc on the StreamConfig to link to the ReaderGroupConfig

Added a reference to javadoc in StreamConfiguration and RetentionPolicy classes.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,48 @@
 
     private final int maxOutstandingCheckpointRequest;
 
+    private final StreamDataRetention retentionConfig;
+
+    /**
+     * If a Stream's Retention Policy is set to {@link RetentionPolicy.RetentionType#CONSUMPTION},
+     * and the user expects reads from this Reader Group to impact data retention in the Stream,
+     * the retentionConfig in {@link ReaderGroupConfig} should be set to
+     * to 'CONSUMPTION_BASED_USER_STREAMCUT' or 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT'.
```
tkaitchuck, line 59:
```suggestion
     * and the the {@link ReaderGroupConfigBuilder#setRetentionConfig retentionConfig} is set 
     * to 'CONSUMPTION_BASED_USER_STREAMCUT' or 'CONSUMPTION_BASED_AT_LAST_CHECKPOINT',
     * the reader group can extend data retention for the Stream.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,48 @@
...
+     * Setting these options implies the Reader Group will notify the Controller of its consumption position {@link StreamCut}
+     * in the Stream, so only un-consumed data can be retained.
```
tkaitchuck, line 61:
```suggestion
     * Setting these options implies the Reader Group will supply its consumption {@link StreamCut},
     * so only un-consumed data can be retained.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,48 @@
...
+     * This notification can be manual ('CONSUMPTION_BASED_USER_STREAMCUT') or automatic ('CONSUMPTION_BASED_AT_LAST_CHECKPOINT')
+     * If a Stream's Retention Policy is TIME/SPACE based the ReaderGroupConfig should have retentionConfig='NO_IMPACT',
+     * so consumption position notifications are not sent to Controller.
```
tkaitchuck, line 64:
```suggestion
     * If a Stream's Retention Policy is 'TIME' or 'SPACE' based the ReaderGroupConfig will have no effect.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,48 @@
...
+     *
+     * NO_IMPACT - Set when read positions of this Reader Group are not expected to impact Stream truncation/retention.
```
tkaitchuck, line 66:
```suggestion
     * NO_IMPACT - Set when the reader's positions of this Reader Group should not impact Stream retention.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,48 @@
...
+     *             Set this value when :
+     *             a. Stream Retention Policy = "CONSUMPTION" {@link RetentionPolicy.RetentionType#CONSUMPTION}
+     *             but this Reader group does *not* want its reads to impact Stream data retention.
+     *             OR
+     *             b. Stream Retention policy is TIME/SPACE based.
```
tkaitchuck, line 71:
```suggestion

```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,48 @@
...
+     * CONSUMPTION_BASED_USER_STREAMCUT - User provides StreamCut to mark consumption boundary on the Stream using {@link ReaderGroup#updateRetentionStreamCut(java.util.Map) } API.
+     * CONSUMPTION_BASED_AT_LAST_CHECKPOINT - StreamCut corresponding to lastCompletedCheckpoint is auto-notified as truncation Stream-Cut to Controller.
```
tkaitchuck, line 73:
```suggestion
     * CONSUMPTION_BASED_AT_LAST_CHECKPOINT - StreamCut of the last completed checkpoint is used as the retention Stream-Cut.
```

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
tkaitchuck: Ok. Then please remove these two lines.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
tkaitchuck: > This can be disallowed, 

The code here is not disallowing this. It should either work and pass the data to the controller or actually enforce that it is disallowed by throwing an exception.

> The system behaves as per the retentionConfig option chosen by the user.

It is not. The docs do not explain what happens if `CONSUMPTION_BASED_USER_STREAMCUT` is specified and this method is invoked. If that is invalid then this must throw. If it is valid it must work. Silently ignoring the call violates either assumption.

pbelgundi: The updateRetentionStreamCut() call does throw an UnsupportedOperationException when the RG Retention config is not set to "CONSUMPTION_BASED_USER_STREAMCUT".

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: > If a Stream's Retention Policy is 'TIME' or 'SPACE' based the ReaderGroupConfig will have no effect.

Based on discussion with @shiveshr and @fpj today, a Stream with TIME/SPACE based retention policy can be truncated based on RG Stream-Cuts. In this case the TIME/SPACE policy related paramters just serve as min/max limits for Consumption based retention.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: Changed NO_IMPACT to NONE as suggested. fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: fixed

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: I have changed this based on recent discussions. I think this explanation is necessary to give the user a correct understanding of when RG reads will impact truncation and when not.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
pbelgundi: I have modified this suitably

pbelgundi:
> LGTM except the serialization change requested by tom.

Fixed

tkaitchuck (approved):
Approve with one comment

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -49,11 +50,34 @@
 
     private final int maxOutstandingCheckpointRequest;
 
+    private final StreamDataRetention retentionType;
+
+    /**
+     * If a Reader Group wants only unconsumed data to be retained in a Stream,
```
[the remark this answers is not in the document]
pbelgundi: fixed

PR #5307, opened 2020-11-06
Issue 5139: Ensure writer does not retry indefinitely to establish a connection with Pravega.

### LINKED ISSUE #5139: Non blocking call EventStreamWriterImpl#writeEvent blocks current thread in case of unreachability of Segmentstore
**Problem description**
We are using Pravega client 0.7.1 and Pravega server 0.7.1.
The  **EventStreamWriterImpl#writeEvent** is an asynchronous method the returns a CompletableFuture, the caller expects it to not perform significant blocking operations.

In case of Pravega cluster failure (see below), simple this method blocks the current thread.
Blocking the caller thread is bad because it is potentially holding resources and it prevents the system to work.
In the case of the stacktrace below the thread is serving an HTTP request.

It is quite easy to reproduce the problem, just by starting a single machine Pravega cluster (controller + segmentstore on the same machine + FILESYSTEM tier2, BK and ZK cluster are replicated and shared with other components of the application) and then **stopping gracefully the segmentstore service**.

**Problem location**
Pravega client code

This is a stacktrace (Pravega client 0.7.1)

 ```
java.lang.Thread.State: BLOCKED (on object monitor)
	at io.pravega.client.stream.impl.EventStreamWriterImpl.writeEventInternal(EventStreamWriterImpl.java:118)
	- waiting to lock <0x0000000601dcd298> (a java.lang.Object)
	at io.pravega.client.stream.impl.EventStreamWriterImpl.writeEvent(EventStreamWriterImpl.java:104)
	at magnews.gateway.pravega.PravegaOutputCollector.collect(PravegaOutputCollector.java:244)
	at magnews.gateway.logger.OutputCollectorFactoryFacade$OutputCollectorWrapper.collect(OutputCollectorFactoryFacade.java:369)
	at magnews.gateway.logger.InputEventsCollectors.collect(InputEventsCollectors.java:128)
	at magnews.gateway.logger.InputCollector.log(InputCollector.java:540)
	at magnews.gateway.messagemanager.MessageManager.acceptMessage(MessageManager.java:1204)
       ......
        at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:343)
	at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:373)
	at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:65)
	at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:868)
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1589)
	at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)
	- locked <0x000000079cc635f0> (a org.apache.tomcat.util.net.NioEndpoint$NioSocketWrapper)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@13.0.1/ThreadPoolExecutor.java:1128)

```

Do not block the current thread, leave the control as soon as possible and defer the execution to other internal Pravega client thread.

### PR DESCRIPTION
**Pull request by shrids**
**Change log description**  
- Ensure `EventStreamWriter.writeEvent(..)`does not indefinitely try to establish a connection with Pravega.  The future returned by this API will be completed exceptionally incases the client writer is not able to establish a conection with Pravega even after retrying based on the the `EventWriterConfig` configuration.
- Ensure `EventStreamWriter.flush()` throws a RetriesExhaustedException incase the client fails to establish connectivity with Pravega.

**Purpose of the change**  
Fixes #5139 

**What the code does**  
Ensures,
- `EventWriterConfig` decides for how long the client tries to establish a connection with Pravega.
- Ensure the async API does not block indefinitely to establish a connection with Pravega.
- Ensure flush() indicates an exception if there is an issue with connectivity.

**How to verify it**  
Existing and newly added tests should pass.

### CODE REVIEW

File: client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java [hunk 1]
```diff
@@ -557,6 +562,10 @@ public void flush() throws SegmentSealedException {
                 }
             } catch (Exception e) {
                 failConnection(e);
+                if (e instanceof RetriesExhaustedException) {
+                    //throw an exception to the external world that the flush failed due to RetriesExhaustedException
+                    Exceptions.sneakyThrow(e);
```
eolivelli, line 567: what about adding a "return" ?
it won't have effect but it may make the code cleaner

eolivelli (approved):
awesome work.
I will try to test it with EmailSuccess,
unfortunately there are a few dependencies upgrades and the update on EmailSuccess won't be easy.

@hamadodene @diegosalvi please also take a look

shrids:
System test run: `pravega-branch-builder/430/` is in progress.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: The intent here was to ensure the user knows that there was a connectivity issue for this synchronous API. Adding a return with a log message will imply the user will not know that some events were not written due to connectivity issues.

@tkaitchuck is there a better way to communicate this to the user of Pravega?

eolivelli: Exceptions.sneakyThrow(e); throws an exception, and the execution of the method is aborted, so basically it is like we had a "return" after   Exceptions.sneakyThrow(e);

my concern is only about telling explicitly to the reader of the code that the execution is ending up there (also automatic flow execution tools like spotbugs may be happier)

not a big deal, we can keep it as it is now

shrids: Ok, Thanks for clarifying it.
I have improved the code.

eolivelli (approved):
great

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -557,6 +562,11 @@ public void flush() throws SegmentSealedException {
...
```
tkaitchuck, line 567: The typical pattern here is to do:
` throw Exceptions.sneakyThrow(e);`
For this reason the signature of `sneakyThrow` returns an exception even though the function never returns. Calling it in this way makes it clear visually that an exception will be thrown from this line and makes static analysis tools like IDEs better able to parse what is going on.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -586,61 +597,64 @@ void reconnect() {
...
+                              if (exception instanceof NoSuchSegmentException) {
+                                  log.info("Ending reconnect attempts on writer {} to {} because segment is truncated", writerId, segmentName);
+                                  return null;
+                              }
+                              throw Exceptions.sneakyThrow(t1);
+                          });
+
+                      }, connectionPool.getInternalExecutor());
+              }, connectionPool.getInternalExecutor()).exceptionally(t -> {
+                 log.error("Error will attempting to establish connection", t);
```
tkaitchuck, line 653:
```suggestion
                 log.error("Error while attempting to establish connection", t);
```

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -652,6 +666,11 @@ void reconnect() {
         }
     }
 
+    private void failUnackedEvents(Throwable t) {
+        state.getAllInflightEvents().parallelStream().forEach(event -> event.getAckFuture().completeExceptionally(t));
```
tkaitchuck, line 670: If they are completed exceptionally they should probably be removed from inflight as they cannot be subsequently be completed successfully.

tkaitchuck (commented):
Please add a test where the failure occurs but the caller subsequently attempts to write more data.

shrids:
System test run is successful ref: `test-pravega-continuous-1/1058/`

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: fixed it.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: ok, I have updated the code.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: fixed it.

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: fixed it.

shrids:
I have added a test where a failure occurs and the caller subsequently attempts to write data @ https://github.com/pravega/pravega/pull/5307/files#diff-ebc28a0ce857a7312d23034c2f7c3eab917e0418828b7c520daca0d66205b9e6R288
Let me know if it needs to be fortified further.

tkaitchuck (approved):

tkaitchuck:
Build failed due to: https://github.com/pravega/pravega/issues/5321
(Looks to be a bad controller test that somehow was merged.)

PR #5317, opened 2020-11-11
Issue 5294: Bookkeeper Client may cache old DNS entries for too long

### LINKED ISSUE #5294: Bookkeeper Client may cache old DNS entries for too long
**Problem description**
In a Kubernetes environment, we observed that, after inducing a Bookie restart, some Segment Store instances where trying to read from the old IP of the Bookie. While a related problem has been already fixed in Bookkeeper (see https://github.com/apache/bookkeeper/pull/1762), it looks like now the responsibility for refreshing DNS entries is on the shoulders of the JVM, as the Bookkeeper client uses `InetAddress` to resolve the IP of Bookie names.

Also mentioned in the previous issue, it looks like the parameter that dictates the JVM behavior to refresh DNS entries is related to parameter `networkaddress.cache.ttl`. Even more, this parameter may default to `-1`, which means "cache forever". If this is the case, it may justify the behavior we have seen when inducing Bookie restarts.

**Problem location**
Bookkeeper client.

Set `networkaddress.cache.ttl` to `0` in the Segment Store code, so we make sure that DNS are not cached.

RaulGracia:
**Change log description**  
Added a defensive mechanism in which the Segment Store can re-create the Bookkeeper client to force the resolution of Bookie IPs in the case of persistent container recoveries.

**Purpose of the change**  
Fixes #5294.

**What the code does**  
We have found that, in some sporadic cases. the Bookkeeper client may get stuck "connected" with an IP of a dead Bookie for a significant period of time (e.g., 15 min) until an underlying timeout closes the connection. This leads the Segment Store to get stuck as well trying to contact with that Bookie while trying to recover a container. In the long term, the ideal solution would be that the Bookkeeper client correctly detects such a situation, but this is being discussed in this issue: https://github.com/apache/bookkeeper/issues/2482

Meanwhile, we have added a defensive mechanism in `BookkeeperLogFactory`, which is the class holding the Bookkeeper client reference shared across all containers. Concretely, in the case of detecting successive recoveries (i.e., >=2) on a given container within a defined period of time (i.e., 60 seconds), the Segment Store re-creates the Bookkeeper client. This will force the client to resolve the new IP of Bookies.

**How to verify it**  
We have tried to reproduce the original issue multiple times and the system recovers with this fix. All other tests should be working as before.


segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
```diff
@@ -26,18 +35,31 @@
...
+    // Maximum number of log creation attempts for a given container before considering resetting the BK client.
+    private static final int MAX_CREATE_ATTEMPTS_PER_LOG = 2;
+
     private final String namespace;
     private final CuratorFramework zkClient;
     private final AtomicReference<BookKeeper> bookKeeper;
     private final BookKeeperConfig config;
     private final ScheduledExecutorService executor;
+    @GuardedBy("this")
+    @Getter(AccessLevel.PACKAGE)
```
andreipaduroiu, line 58: How can you have `GuardedBy` and a public getter at the same time? You cannot guarantee thread safety like this. You need to provide APIs that access this while holding locks.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
```diff
@@ -26,18 +35,31 @@
...
+    private final Map<Integer, LogInitializationRecord> logInitializationTracker = new HashMap<>();
+    @GuardedBy("this")
+    @Getter(AccessLevel.PACKAGE)
```
andreipaduroiu, line 61: Same here. Never expose `Atomic*` directly; use APIs.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
```diff
@@ -169,5 +193,74 @@ private BookKeeper startBookKeeperClient() throws Exception {
...
+     * Recreate the Bookkeeper client if a given log exhibits MAX_CREATE_ATTEMPTS_PER_LOG creation attempts (as a proxy
+     * for Container recoveries) within the period of time defined in LOG_CREATION_INSPECTION_PERIOD.
+     *
+     * @param logId Id of the log being restarted.
+     */
+    private void tryResetBookkeeperClient(int logId) {
+        synchronized (this) {
+            if (logInitializationTracker.containsKey(logId)) {
+                // Account for a restart of the Bookkeeper log.
+                logInitializationTracker.get(logId).incrementLogCreations();
```
andreipaduroiu, line 206: Take `logInitializationTracker.get(logId)` and stash it in a local variable. Then compare it against null in the `if` above and reuse that variable. This should reduce the number of `get` calls into that HashMap.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
```diff
@@ -169,5 +193,74 @@ private BookKeeper startBookKeeperClient() throws Exception {
...
+                        BookKeeper newClient = startBookKeeperClient();
+                        // If we have been able to create a new client successfully, reset the current one and update timer.
+                        log.info("Successfully created new Bookkeeper client, setting it as the new one to use.");
+                        BookKeeper oldClient = this.bookKeeper.getAndSet(newClient);
+                        lastBookkeeperClientReset.set(new Timer());
+                        // Lastly, attempt to close the old client.
+                        log.info("Attempting to close old client.");
+                        oldClient.close();
+                    } catch (Exception e) {
+                        log.error("Failure resetting the Bookkeeper client: ", e);
```
andreipaduroiu, line 221: This will result in double logging. Whomever is invoking this method will absolutely log it as well.

segmentstore/storage/impl/src/test/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookkeeperLogFactoryTests.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.Test;
+
+import java.net.NetworkInterface;
+import java.util.Enumeration;
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * Test to exercise the re-creation of the Bookkeeper client upon successive log creation attempts.
+ */
+public class BookkeeperLogFactoryTests extends BookKeeperClusterTestCase {
```
andreipaduroiu, line 32: Why do you subclass from this? 

Can't you do it like we test `BookKeeperLog`? We should have everything we need there.

segmentstore/storage/impl/src/test/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookkeeperLogFactoryTests.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private ScheduledExecutorService executorService = ExecutorServiceHelpers.newScheduledThreadPool(2, "test");
```
andreipaduroiu, line 34: If you do not need to subclass from `BookKeeperClusterTestCase`, then you can subclass from `ThreadPooledTestSuite` which will take care of your executor for you.

segmentstore/storage/impl/src/test/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookkeeperLogFactoryTests.java
```diff
@@ -0,0 +1,122 @@
...
+        oldBookkeeperClient = factory.getBookKeeperClient();
+        // The timer for this log should have been updated, so even if there are more initialization attempts, they should
+        // not lead to a new Bookkeeper client re-creation until the inspection period expires.
+        factory.createDebugLogWrapper(0);
+        Assert.assertEquals(oldBookkeeperClient, factory.getBookKeeperClient());
+        factory.createDebugLogWrapper(0);
+        Assert.assertEquals(oldBookkeeperClient, factory.getBookKeeperClient());
+    }
+
+    static class OldTimer extends Timer {
```
andreipaduroiu, line 113: Use `ManualTimer`. You can set the time as you wish.

andreipaduroiu (changes requested):

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
RaulGracia: Ok, moved the `Getter` to a explicit method tagged with `@VisibleForTesting`.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
RaulGracia: Fixed as mentioned above.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
RaulGracia: Done.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
RaulGracia: Removed first log line.

segmentstore/storage/impl/src/test/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookkeeperLogFactoryTests.java
RaulGracia: Moved the test to `BookkeeperLog`.

segmentstore/storage/impl/src/test/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookkeeperLogFactoryTests.java
RaulGracia: Fixed by moving the test to `BookkeeperLogTests`.

segmentstore/storage/impl/src/test/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookkeeperLogFactoryTests.java
RaulGracia: I tried, but this forces me to add another dependency to `segmentstore:storage:impl`, so I think that just adding this mock, tiny class has less impact to the project.

segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/BookKeeperLogFactory.java
```diff
@@ -169,5 +189,84 @@ private BookKeeper startBookKeeperClient() throws Exception {
...
+                    try {
+                        log.info("Start creating Bookkeeper client in reset.");
+                        BookKeeper newClient = startBookKeeperClient();
+                        // If we have been able to create a new client successfully, reset the current one and update timer.
+                        log.info("Successfully created new Bookkeeper client, setting it as the new one to use.");
+                        BookKeeper oldClient = this.bookKeeper.getAndSet(newClient);
+                        lastBookkeeperClientReset.set(new Timer());
+                        // Lastly, attempt to close the old client.
+                        log.info("Attempting to close old client.");
+                        oldClient.close();
```
eolivelli, line 216: this "close()" will fail all of the pending operations 
IIUC with this change we are going to close the BK client even if it is not needed

RaulGracia: @eolivelli correct, this defensive mechanism is going to reset the Bookkeeper client (shared across all containers) if we detect X number of restarts for a given container within a period of time T. I understand that this is not the ideal way to handle this problem, but this may be a temporarily solution to the problem reported in the issue until this is fixed: https://github.com/apache/bookkeeper/issues/2482. In parallel, we are working on another solution related to better set the `TCP_USER_TIMEOUT` as suggested by @fpj. If that works, maybe we do not need to merge this PR.

eolivelli: I see, thanks for your clarification

andreipaduroiu (commented):
All my comments are addressed. Keeping "Request changes" until we decide it's OK to merge this in.

andreipaduroiu (approved):

PR #5319, opened 2020-11-11
Issue 5318: use ByteBuf as payload

### LINKED ISSUE #5318: Use ByteBuf as data in ConditionalAppend
**Problem description**
Rust client needs to use ConditionalAppend for byte stream read/write. However, the current ConditionalAppend can only take Event as payload, which adds an extra 8 bytes header that cannot be interpreted by byte stream reader.

**Problem location**
ConditionalAppend wirecommand

Use ByteBuf as payload in ConditionalAppend just like AppendBlockEnd

Tristan1900:
**Change log description**  
Add ConditionalBlockEnd 

**Purpose of the change**  
Fix #5318 

**What the code does**  
The ConditionalBlockEnd is basically ConditionalAppend wirecommand but using ByteBuf as payload

**How to verify it**  
Tests should pass. Rust client byte stream API should work using this PR.


andreipaduroiu (changes requested):
Isn't @tkaitchuck trying to remove all of this stuff from the client? #5249

Tristan1900:
@andreipaduroiu Thanks for the notice. I see, so it will be using `BufferView` as the raw bytes instead of `ByteBuf`. Anyway we need to change `ConditionlAppend` from using `Event` to either `BufferView` or `ByteBuf` and let client to decide to encode data with the 8 bytes header or not.

tkaitchuck (commented):
I am indeed trying to remove byteBufs. 
So this will be a PR that one way or another is going to conflict with the one I am working on.

This PR needs to increment the wireprotocol version. It is a backwards compatible change, but if a client wishes to depend on a version at least this new, it needs to have a version number to refer to.

tkaitchuck (approved):

andreipaduroiu (approved):

andreipaduroiu:
@Tristan1900 Can you please improve unit test coverage? I am not able to merge until codecov is happy.

tkaitchuck (approved):

PR #5329, opened 2020-11-13
Issue 5324: Add generations to Controller API for Consumption Based Retention

### LINKED ISSUE #5324: Controller add/delete Subscriber APIs should accept Subscriber generation field
**Problem description**
To ensure stale API updates from Client are not applied by Controller, we need to add a Subscriber generation field to the API and ensure Controller only executes the API if the generation value is greater than the generation stored in its metadata.

**Problem location**
Controller

pbelgundi:
**Change log description**  
To ensure stale API updates from Client are not applied by Controller, we need to add a Subscriber generation field to the API and ensure Controller only executes the API if the generation value is greater than the generation stored in its metadata.

**Purpose of the change**  
Fixes #5324

**What the code does**  
Adds a generation field to add and delete subscriber APIs, such that if generation provided in the API is lower than generation stored in the metadata, the operation is rejected.

**How to verify it**  
Unit and Integration Tests should pass.


controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -278,15 +277,32 @@ private String getWritersTableName(String id) {
...
-                                SubscriberSet.add(subscriberSetRecord.getObject(), newSubscriber).toBytes(), subscriberSetRecord.getVersion())
-                                .thenCompose(v -> storeHelper.addNewEntryIfAbsent(metadataTable, getKeyForSubscriber(newSubscriber), newSubscriberRecord.toBytes()))
-                                .thenAccept(v -> storeHelper.invalidateCache(metadataTable, SUBSCRIBER_SET_KEY))
-                                .thenAccept(v -> storeHelper.invalidateCache(metadataTable, getKeyForSubscriber(newSubscriber)))));
+                   .thenCompose(subscriberSetRecord -> {
+                       if (subscriberSetRecord.getObject().getSubscribers().containsKey(newSubscriber)) {
+                           // update Subscriber generation
+                           Long generation = subscriberSetRecord.getObject().getSubscribers().get(newSubscriber);
+                           if (generation.longValue() < operationGeneration) {
+                               storeHelper.updateEntry(metadataTable, SUBSCRIBER_SET_KEY,
```
shiveshr, line 289: this is a hanging future. this should be returned from this thenCompose call

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -278,15 +277,32 @@ private String getWritersTableName(String id) {
...
+                                SubscriberSet.update(subscriberSetRecord.getObject(), newSubscriber, generation).toBytes(),
+                                                                         subscriberSetRecord.getVersion())
+                                .thenAccept(v -> storeHelper.invalidateCache(metadataTable, SUBSCRIBER_SET_KEY));
+                           }
```
shiveshr, line 293: we need to handle the else case and return a completed future

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -278,15 +277,32 @@ private String getWritersTableName(String id) {
...
+                      } else {
+                           // add new Subscriber
+                           storeHelper.updateEntry(metadataTable, SUBSCRIBER_SET_KEY,
```
shiveshr, line 296: same comment.. hanging future.. should be linked to the response being sent from this callback

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -278,15 +277,32 @@ private String getWritersTableName(String id) {
...
+                                   SubscriberSet.add(subscriberSetRecord.getObject(),
+                                           newSubscriber, operationGeneration).toBytes(), subscriberSetRecord.getVersion())
+                                   .thenCompose(v -> storeHelper.addNewEntryIfAbsent(metadataTable,
+                                           getKeyForSubscriber(newSubscriber), newSubscriberRecord.toBytes()))
+                                   .thenAccept(v -> storeHelper.invalidateCache(metadataTable, SUBSCRIBER_SET_KEY))
+                                   .thenAccept(v -> storeHelper.invalidateCache(metadataTable, getKeyForSubscriber(newSubscriber)));
+                       }
+                       return CompletableFuture.completedFuture(null);
```
shiveshr, line 304: whatever computation you are doing above is done asynchronously and then you are returning a completed future which is not linked to the above computation.. this is incorrect. this method will return with a completed future and caller will assume the work is done while the work would still be happening asynchronously and could fail.. the response future should be linked to the actual processing.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -278,15 +277,32 @@ private String getWritersTableName(String id) {
     }
 
     @Override
-    public CompletableFuture<Void> createSubscriber(String newSubscriber) {
+    public CompletableFuture<Void> createSubscriber(String newSubscriber, long operationGeneration) {
```
shiveshr, line 280: we are setting generation in create but we need to do the same in update subscriber

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
```diff
@@ -113,9 +113,10 @@
      * Create subscribers record for storing metadata about Stream Subscribers.
      * Also add this new Subscriber to the Record.
      * @param subscriber first subscriber to be added the SubscribersRecord in Stream Metadata.
+     * @param generation generation of subscriber in Stream Metadata.
      * @return future of operation.
      */
-    CompletableFuture<Void> createSubscriber(String subscriber);
+    CompletableFuture<Void> createSubscriber(String subscriber, long generation);
```
shiveshr, line 119: we need to also update "set subscriber method"

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
pbelgundi: ok. fixed.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -311,18 +312,9 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
-                    //4. If SubscriberRecord does not exist create one...
-                    if (subscribersData == null) {
-                        return streamMetadataStore.createSubscriber(scope, stream, newSubscriber, context, executor)
-                                .thenApply(v -> AddSubscriberStatus.Status.SUCCESS);
-                    } else {
-                        return CompletableFuture.completedFuture(AddSubscriberStatus.Status.SUBSCRIBER_EXISTS);
-                    }
-                    })
-                    .exceptionally(ex -> {
+                return streamMetadataStore.createSubscriber(scope, stream, newSubscriber, generation, context, executor)
```
shiveshr, line 315: question: does create subscriber update the entry if it already existed, but with an older generation?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: If entry already exists, its generation is updated only if the (API generation) > stored generation and not otherwise.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
```diff
@@ -116,6 +116,9 @@
     @GuardedBy("subscribersLock")
     private final List<VersionedMetadata<StreamSubscriber>> streamSubscribers = new ArrayList<>();
 
+    @GuardedBy("subscribersLock")
+    private final Map<String, Long> subscribersSet = new HashMap<String, Long>();
```
shiveshr, line 120: nit: misnomer `set` when its a `map`

shiveshr: also why do we need an additional map here? can we update the StreamSubscriber to have the generation field?

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberSet.java
```diff
@@ -37,37 +37,55 @@
...
+    /**
+     * This method updates the generation of a subscriber in the subscriberSet.
+     * @param subscriberSet Subscriber Set.
+     * @param subscriber subscriber to be added.
+     * @param generation subscriber generation.
+     * @return updated Subscriber Set.
+     */
+    public static SubscriberSet update(@NonNull SubscriberSet subscriberSet, @NonNull String subscriber, long generation) {
+        ImmutableMap.Builder<String, Long> builder = ImmutableMap.builder();
+        Map<String, Long> otherSubscribers = subscriberSet.getSubscribers().entrySet().stream().filter(e -> !e.getKey().equals(subscriber))
```
shiveshr, line 70: this can be improved.. its making two passes (filter and putall) over the map. we can get it down to one pass only by adding all the existing entries and for entry which already exists, we add the new subscriber.

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberSet.java
```diff
@@ -37,37 +37,55 @@
...
      * @param subscriberSet Subscriber Set.
      * @param subscriber subscriber to be removed.
      * @return updated Subscriber Set.
      */
     public static SubscriberSet remove(@NonNull SubscriberSet subscriberSet, @NonNull String subscriber) {
-        ImmutableList.Builder<String> builder = ImmutableList.builder();
-        List<String> otherSubscribers = subscriberSet.getSubscribers().stream().filter(s -> !s.equals(subscriber)).collect(Collectors.toList());
-        builder.addAll(otherSubscribers);
-        return new SubscriberSet(builder.build());
+       ImmutableMap.Builder<String, Long> builder = ImmutableMap.builder();
```
shiveshr, line 84: same here. we can put all and then call remove on the builder.

shiveshr (approved):
LGTM with some minor comments

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
pbelgundi: SubscriberSet was not created for adding generations. 
It exists so we can maintain a separate list of subscribers from the "Subscriber--> StreamCut" mapping. 
This this needed for finding the list of all Subscribers for a Stream (listSubscribers)

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberSet.java
pbelgundi: ok.

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberSet.java
pbelgundi: putAll and Remove is still 2 passes. 
I'm not sure if this would matter as much..... as I'd expect the number of subscribers per Stream to be < 20

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStream.java
pbelgundi: > nit: misnomer `set` when its a `map`

changed to Subscribers

controller/src/main/java/io/pravega/controller/store/stream/records/SubscriberSet.java
pbelgundi: fixed.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -327,15 +344,20 @@ private String getWritersTableName(String id) {
...
-                            .thenCompose(subscriberRecord -> storeHelper.removeEntry(table, getKeyForSubscriber(subscriber)))
-                            .thenAccept(x -> storeHelper.invalidateCache(table, SUBSCRIBER_SET_KEY))
-                            .thenAccept(x -> storeHelper.invalidateCache(table, getKeyForSubscriber(subscriber))));
+                .thenCompose(subscriberSetRecord -> getMetadataTable()
+                .thenCompose(table -> {
+                    if (subscriberSetRecord.getObject().getSubscribers().containsKey(subscriber)
+                        && subscriberSetRecord.getObject().getSubscribers().get(subscriber).longValue() < generation) {
+                        Subscribers subSet = Subscribers.remove(subscriberSetRecord.getObject(), subscriber);
+                        return getSubscriberRecord(subscriber)
+                                .thenCompose(subscriberRecord -> storeHelper.removeEntry(table, getKeyForSubscriber(subscriber)))
```
shiveshr, line 355: this is not done atomically with updateSubscribers. 
so you have two steps:
updateSubscribers + removeSubscriberEntry
the problem is after updateSubscribers if someone added the subscriber back and created the subscriber entry idempotently, you may end up removing it.

shiveshr: A solution could be do to conditional remove.

shiveshr (approved):

PR #5337, opened 2020-11-16
Issue 5320: Metrics and Stats Reporting Optimizations

### LINKED ISSUE #5320: Segment Stats Collection Optimization
**Problem description**
Suspects:
- DynamicProxyImpl is slow
- SegmentStatsRecorderImpl is slow
- May be more.

**Problem location**
Everywhere.

Optimize code.

andreipaduroiu:
**Change log description**  
- Changed `DynamicLoggerImpl` to use a simplified Cache (instead of Guava Cache).
- Optimized `SegmentStatsRecorderImpl` to use fewer resources.
- Replaced usages of Dynamic Logger throughout the Segment Store codebase with the equivalent Counter, Metric and OpStatsLogger implementations.

**Purpose of the change**  
Fixes #5320.

**What the code does**  
- `SimpleCache` a very simple cache that supports a maximum-size and maximum-time eviction policy. This is about 4-8x faster than using Guava Cache (due to lower implementation overhead - as we do not need all features of Guava Cache).
    - Uses a HashMap to store values and a customized linked list to store order of access. If need be, entries will be evicted from the head of this queue.
    - Entries are evicted upon access or upon invocation of `cleanUp`. There is no background thread or async job that maintains the cache (this aids in the simplicity and small resource utilization).
    - An "expired" entry is going to be treated as "inexistent" (w.r.t get, remove, put and putIfAbsent)
- Changed `DynamicLoggerImpl`:
    - Making use of `SimpleCache` instead of GuavaCache. This resulted in a 45-50% performance improvement.
    - This implementation is equivalent to the old one. Unit tests have been updated to verify that the behavior is consistent.
- `SegmentStatsRecorderImpl`
    - Making use of `SimpleCache` to cache values.
    - Reworked `recordAppend` to be async. The caller (`AppendProcessor.handleAppendResult`) will queue a value in a `BlockingDrainingQueue` (very small effort) and a background processor picks items from this queue (in batches) and reports the entire batch at once. 
    - Reduced usages of `DynamicLogger` and using equivalent implementations of `Counter` for its metrics.
- `AutoScaleProcessor`
    - Making use of `SimpleCache` to cache values.
- `AutoScaleMonitor`
    - Reduced the dedicated thread pool priority.
- Metrics Reporting
    - Changed `SegmentStoreMetrics` and `CacheMetrics` to not use DynamicLogger for Counters and Metrics. This is a more consistent implementation and more efficient.
        - Cannot change `Gauge` at this time due to significant API differences. Such a change may be done in the future.
- Other minor changes
    - Changed `MetricsConfig.DynamicCacheSize` default value to 100,000 (down from 10,000,000). Previous value was excessive.
    - Removed a usage of `String.format` in `ContainerMetadataUpdateTransaction`. This was serving very little purpose, yet it's a frequently invoked code on the hot path.

**How to verify it**  
All existing tests must pass. Some new unit tests were added and some existing tests got updated to cover more.
System tests must pass.


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleMonitor.java
```diff
@@ -34,14 +34,14 @@
     @VisibleForTesting
     public AutoScaleMonitor(@NonNull StreamSegmentStore store, @NonNull EventStreamClientFactory clientFactory,
                             @NonNull AutoScalerConfig configuration) {
-        this.executor = ExecutorServiceHelpers.newScheduledThreadPool(configuration.getThreadPoolSize(), "auto-scaler");
+        this.executor = ExecutorServiceHelpers.newScheduledThreadPool(configuration.getThreadPoolSize(), "auto-scaler", Thread.MIN_PRIORITY);
```
tkaitchuck, line 37: Please be aware a min priority thread can be starved.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheMetrics.java
```diff
@@ -10,28 +10,41 @@
...
 import io.pravega.shared.metrics.MetricsProvider;
+import io.pravega.shared.metrics.StatsLogger;
 
 /**
- * Metrics for {@link CacheStorage}
+ * Metrics for {@link DirectMemoryCache}.
  */
-final class CacheMetrics {
-    private static final DynamicLogger DYNAMIC_LOGGER = MetricsProvider.getDynamicLogger();
+final class CacheMetrics implements AutoCloseable {
```
tkaitchuck, line 20: Is this being properly closed everywhere? (This is hard to check outside of an IDE)

tkaitchuck (approved):
Approved with comments

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/cache/CacheMetrics.java
andreipaduroiu: Yes. It's only used in `DirectMemoryCache` and I added a `metrics.close` when closing it.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleMonitor.java
andreipaduroiu: Fully aware of that. But calculating and generating autoscale metrics is of less importance than ingesting data itself. If we reach a situation where our CPUs are fully loaded, then we need to prioritize ingestion over this.

This CPU is mostly used for AutoScaleMonitor/Processor and SegmentStatsImpl cache cleanup and writing events to the Controller's auto scale stream.

In my local tests I have found this thread pool to take up considerable CPU. That is likely because, previously, the code in `SegmentStatsRecorderImpl` was executing one callback per Store append. Now that I've batched them together this should be less of a concern.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/AutoScaleProcessor.java
```diff
@@ -60,10 +56,9 @@
     private static final long TEN_MINUTES = Duration.ofMinutes(10).toMillis();
     private static final long TWENTY_MINUTES = Duration.ofMinutes(20).toMillis();
     private static final int MAX_CACHE_SIZE = 1000000;
-    private static final int INITIAL_CAPACITY = 1000;
 
     private final EventStreamClientFactory clientFactory;
-    private final Cache<String, Pair<Long, Long>> cache;
+    private final SimpleCache<String, Pair<Long, Long>> cache;
```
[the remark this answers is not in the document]
andreipaduroiu: From what I could measure, Guava cache has a significant overhead due to it doing more work. So it makes sense to keep it for those situations where the value to be cached is expensive to retrieve or compute (i.e., requires I/O). However, for small values which are easy to recalculate, it had too much overhead, hence the SimpleCache which is faster. 

AFAIK, there aren't any other places where Guava Cache is used. SLTS uses it, but in that case it's for one of those expensive values.

common/src/main/java/io/pravega/common/util/SimpleCache.java
```diff
@@ -0,0 +1,399 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import lombok.val;
+
+@ThreadSafe
+@Slf4j
+public class SimpleCache<KeyT, ValueT> {
```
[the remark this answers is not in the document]
andreipaduroiu: Good catch. I totally forgot about this aspect.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/TableSegmentStatsRecorderImpl.java
```diff
@@ -67,57 +69,42 @@ public void close() {
...
-        getDeleteSegment().reportSuccessEvent(elapsed);
-        String[] segmentTags = segmentTags(tableSegmentName);
-        getDynamicLogger().freezeCounter(MetricsNames.TABLE_SEGMENT_UPDATE_CONDITIONAL, segmentTags);
-        getDynamicLogger().freezeCounter(MetricsNames.TABLE_SEGMENT_UPDATE, segmentTags);
-        getDynamicLogger().freezeCounter(MetricsNames.TABLE_SEGMENT_REMOVE_CONDITIONAL, segmentTags);
-        getDynamicLogger().freezeCounter(MetricsNames.TABLE_SEGMENT_REMOVE, segmentTags);
-        getDynamicLogger().freezeCounter(MetricsNames.TABLE_SEGMENT_GET, segmentTags);
-        getDynamicLogger().freezeCounter(MetricsNames.TABLE_SEGMENT_ITERATE_KEYS, segmentTags);
-        getDynamicLogger().freezeCounter(MetricsNames.TABLE_SEGMENT_ITERATE_ENTRIES, segmentTags);
+        this.deleteSegment.reportSuccessEvent(elapsed);
```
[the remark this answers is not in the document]
andreipaduroiu: I completely removed these metrics. We were reporting both global metrics and per-segment metrics. However this seems to be overkill, especially for Table Segments. To avoid having to do unnecessary work, I decided to only keep the global metrics and get rid of the per-segment ones. Only the per-segment ones required DynamicLogger, and the other ones are properly closed when this class is closed.

FYI, "Freezing" is only needed for Dynamic Loggers. The "non-dynamic" ones should be closed when they are no longer needed.

RaulGracia (approved):

Issue #5118, opened 2020-08-27
Improve organization of security-related source code


ravisharda:
**Problem description**
Currently, the security-related code is spread out across the following Gradle sub-projects:

| Gradle subproject | Purpose  | 
|:--------------------|:----------|
| `:controller` | Controller-specific code for security | 
| `:segmentstore:server` | Segment Store-specific code for security |
| `:common` | General utilities that are useful for Pravega as well as any other projects. |
| `:shared:authplugin` | Plugin interfaces and other key external-facing interfaces. It is useful for auth plugin and token verifier plugin writers. |
| `:shared:security` | Security-specific code that may be used across Client, Controller and Segmenstore |
|`:client` | `Credentials` interface and the default implementation `DefaultCredentials` |

Issues with the current organization include (but are not limited to): 

1. The code appears to be too spread-out. It can be a bit daunting for developers to identify what to use. 
2. Some of the code is useful in other repositories like the [Schema Registry](https://github.com/pravega/schema-registry/), but is not well organized for reuse. For example, the Password-based Auth handler is used there, but using it required depending on the controller module, among other problems. 
3. The `Credentials` interface and its default implementation is in `:client` module. If another service needs to use it, it needs to include `pravega-client` as a dependency. 

**Problem location**
The Gradle subprojects listed above.

**Suggestions for an improvement**

ravisharda:
Pull request (PR) #5339 resolved this issue apart from issue #5257. Since the PR is merged to master, closing this issue.

PR #5339, opened 2020-11-17
Issue 5257 and 5118: Improve organization of security code for reuse in related projects like Schema Registry

### LINKED ISSUE #5257: Move Password auth handler implementation out of controller
**Problem description**
To facilitate reuse of Password auth handler outside of pravega controller we should move it out of controller and into a separate library. 

**Problem location**
PasswordAuthHandler.java

Refactor it

ravisharda:
**Change log description**  
Reorganizes security code for easier reuse in related projects like Schema Registry.

**Purpose of the change**  
Resolves #5257 and #5118.  

**What the code does**  
Here's what it does: 
1. Moves out the password-based auth handler that supports HTTP Basic Authentication into its own Gradle module `shared:basic-authplugin`. This enables other projects to use the auth handler if needed. 
2. Moves reusable classes from `controller` and `client` into `shared:security`. That makes it possible to use them across projects, without needing to have a dependency on the Controller. 
3. Modifies the Auth Handler plugin interface to inject configurations to plugins via properties, thereby removing their dependency to `Controller` / `GRPCServerConfig` class. Projects that want to use the handler no more need to have a dependency on the Controller for this. (This implies that other projects with Auth handler implementations will need to change their code too.)

**How to verify it**  


settings.gradle
```diff
@@ -31,4 +32,7 @@ include 'common',
         'test:testcommon',
         'test:integration',
         'test:system',
-        'bindings'
\ No newline at end of file
+        'bindings'
+include 'password-authplugin'
```
shiveshr, line 36: what is password-authplugin?
also since its shared:basic-authplugin, what is this include for basic-authplugin?

shiveshr (approved):
LGTM, one minor comment

settings.gradle
ravisharda: Thanks for catching those leftover lines. The IDE had added them when I was experimenting with adding Gradle modules and I forgot to remove them. I've now removed them in the latest commit that I just pushed.

controller/src/main/java/io/pravega/controller/server/security/auth/handler/AuthHandlerManager.java
```diff
@@ -157,7 +157,7 @@ public void registerInterceptors(ServerBuilder<?> builder) {
                 ServiceLoader<AuthHandler> loader = ServiceLoader.load(AuthHandler.class);
                 for (AuthHandler handler : loader) {
                     try {
-                        handler.initialize(serverConfig);
+                        handler.initialize(serverConfig.toAuthHandlerProperties());
```
shiveshr, line 160: as discussed, we should keep the interface for auth handler unchanged so that dependents do not have to update. 
we can add a new method in the serverconfig to get properties.

shared/authplugin/src/main/java/io/pravega/auth/AuthHandler.java
```diff
@@ -69,7 +70,9 @@
     /**
      * Sets the configuration. If the auth handler needs to access the server configuration, it can be accessed though this var.
      *
-     * @param serverConfig The server configuration.
+     * @param properties The auth handler configuration.
      */
-    void initialize(ServerConfig serverConfig);
+    default void initialize(Properties properties) {
```
shiveshr, line 75: lets keep the interface signature same. lets change ServerConfig to add a new API for getting properties.

shiveshr (changes requested):
let us make sure we do not change the auth handler interface

controller/src/main/java/io/pravega/controller/server/security/auth/handler/AuthHandlerManager.java
ravisharda: Sure, I've now modified the implementation, as suggested, to avoid the change in the interface here - handler.initialize() continues to take `ServerConfig` as input.

shared/authplugin/src/main/java/io/pravega/auth/AuthHandler.java
ravisharda: Sure, I've now modified the implementation, as suggested, to avoid the change in the interface here - handler.initialize() continues to take `ServerConfig` as input.

ravisharda:
> let us make sure we do not change the auth handler interface

Sure, that sounds good. The latest commit avoids the interface change.

shiveshr (approved):
LGTM

build.gradle
```diff
@@ -193,10 +193,26 @@ project('shared:authplugin') {
...
+
+    javadoc {
+        title = "AuthHandler supporting HTTP Basic Authentication"
+        dependsOn delombok
+        source = delombok.outputDir
+        failOnError = true
+    }
+}
+
 project('shared:security') {
```
sarlaccpit, line 211: Just to confirm my understanding of the dependency changes by taking a concrete example of an existing implementation:

https://github.com/pravega/pravega-keycloak/blob/master/build.gradle

Is it correct to say that in the above's:

```
 dependencies {
        compileOnly "org.slf4j:slf4j-api:${slf4jVersion}"
        compileOnly "io.pravega:pravega-shared-authplugin:${pravegaVersion}"
        compileOnly "io.pravega:pravega-client:${pravegaVersion}"
        implementation 'org.keycloak:keycloak-authz-client'
...
```

it would now become:

```
 dependencies {
        compileOnly "org.slf4j:slf4j-api:${slf4jVersion}"
        compileOnly "io.pravega:pravega-shared-security:${pravegaVersion}"
        implementation 'org.keycloak:keycloak-authz-client'
...
```
?

The fact we needed the client and authplugin before was because of this and this respectively:

https://github.com/pravega/pravega-keycloak/blob/master/client/src/main/java/io/pravega/keycloak/client/PravegaKeycloakCredentials.java#L13

https://github.com/pravega/pravega-keycloak/blob/master/client/src/main/java/io/pravega/keycloak/client/PravegaKeycloakCredentials.java#L17

But now both would come from the single shared-security module, is that correct?

sarlaccpit (approved):

build.gradle
ravisharda: Yes, that is correct, except that `AuthConstants` which contains constants applicable to auth plugins is still under `"io.pravega:pravega-shared-authplugin`, so you'll need to have a dependency on that module. To remove the dependency, the implementation may choose to use their own String constants instead of the ones in `io.pravega.auth.AuthConstants`.

PR #4878, opened 2020-06-17
Issue 4874: Update hadoopVersion

ravibeta:
**Change log description**  
This touches only one file gradle.properties to update the version of pravega-hdfs.
The hadoopVersion is updated to 2.9.2 as described in the linked issue. However, the guava version had to be lowered slightly. This is not a mistake. All revisions upwards of 2.9.2 up to the latest have had this breaking change which is described by the issues [HDFS-752](https://issues.apache.org/jira/browse/HDFS-752) and [HDFS-7477](https://issues.apache.org/jira/browse/HDFS-7477) and they don't agree with guava. The links for these issues are mentioned here so that the patch associated with these issues can be viewed. They are the only break-in changes we have. These changes are applicable to all recent versions but 2.9.2 as described in the attached issue fixes all the security vulnerabilities cited in the GitHub issue [4874](https://github.com/pravega/pravega/issues/4874) linked to this request.  

**Purpose of the change**  
The purpose of this change is to remove the security vulnerabilities associated with the hadoop version. The Hadoop Version poses a bigger security threat than others to Pravega.

Fixes #4874.

**What the code does**  
This code just changes versions in gradle.properties

**How to verify it**  
This change was verified by travis build.

The exception faced with upgrading Guava is
```
io.pravega.storage.hdfs.HDFSStorageTest > testDelete FAILED
    java.lang.NoSuchMethodError: com.google.common.util.concurrent.Futures.addCallback(Lcom/google/common/util/concurrent/ListenableFuture;Lcom/google/common/util/concurrent/FutureCallback;)V
        at org.apache.hadoop.hdfs.server.datanode.checker.ThrottledAsyncChecker.addResultCachingCallback(ThrottledAsyncChecker.java:167)
        at org.apache.hadoop.hdfs.server.datanode.checker.ThrottledAsyncChecker.schedule(ThrottledAsyncChecker.java:156)
        at org.apache.hadoop.hdfs.server.datanode.checker.StorageLocationChecker.check(StorageLocationChecker.java:164)
        at org.apache.hadoop.hdfs.server.datanode.DataNode.makeInstance(DataNode.java:2762)
        at org.apache.hadoop.hdfs.server.datanode.DataNode.instantiateDataNode(DataNode.java:2677)
        at org.apache.hadoop.hdfs.MiniDFSCluster.startDataNodes(MiniDFSCluster.java:1663)
[...]

PR #5340, opened 2020-11-17
Issue 4874: Upgrade HDFS Client from 2.x to 3.x in order to address security vulnerabilities

### LINKED ISSUE #4874: Fix HDFS security vulnerabilities
**Problem description**
Pravega uses HDFS for tier 2 and it has a set of security vulnerabilities. These include `CVE-2016-6811, CVE-2018-8009, CVE-2018-8029, CVE-2018-11766`. These come from the Apache Hadoop hdfs library along with its transitive dependencies. The versions of these libraries needs to be updated to overcome the security vulnerabilities. The updates to the library requires changes in Pravega source code to remove the methods, exceptions and properties used from the library that are no longer available and to use the replacements for these and other deprecated syntax.

**Problem location**
The version for the Apache hadoop hdfs library is specified in the gradle.properties. Its usage spans HDFS tests and exception helpers. These changes expand to others parts of the code where they are used. 

 The version 2.9.2 of apache hadoop hdfs fixes the above CVEs but there are a few more CVEs that have been addressed in recent versions. These new versions have increased the breaks in compatibility with our existing code. The updates can therefore be progressive increasing from 2.9.2 to more recent versions. The changes to the hdfs library and its transitive dependencies can be made at once.

ravisharda:
**Change log description**  
Upgrade HDFS client version to the latest version, which is currently `3.3.0` as per [Maven Repo](https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-hdfs).  

**Purpose of the change**  
Fixes #4874. 

**What the code does**  

* Updates HDFS client dependency. 
* Modifies code to account for breaking changes. 
* Adds additional dependencies to `bindings` project (since the newer version doesn't brings some of the transitive dependencies that were used by other components of the project)

**How to verify it**  
All unit and integration tests must pass. 

In addition, a distributed mode cluster running HDFS, must be able to run successfullly and services reads and writes. I use Docker-Compose based cluster for this purpose and deployed it using this YAML: 
https://github.com/ravisharda/pravega/blob/issue-4874-upgrade-hdfs/docker/compose/docker-compose.yml

**Note:** 
1. Note that the YAML runs `pravega/hdfs:2.7.7` as the HDFS server, which is:
   * Built out of code here: https://github.com/pravega/docker-hdfs
   * Based on Hadoop v2.7.7

   So, the test mentioned above uses HDFS client `v3.3.0` against a server `v2.7.7`. 

   Upgrade of the server should be taken separately as part of a different PR for the https://github.com/pravega/docker-hdfs repo. 

2. This PR borrows some of the work done in an older and now closed PR #4878 by @ravibeta.


sachin-j-joshi (approved):
Looks okay to me.

sachin-j-joshi (commented):
Are there more changes expected ?

ravisharda:
> Are there more changes expected ?

@sachin-j-joshi Yes, and I've now made those changes with the last commit.

bindings/src/test/java/io/pravega/storage/extendeds3/S3ProxyImpl.java
```diff
@@ -37,7 +37,7 @@
 import lombok.val;
 import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang.ArrayUtils;
+import org.apache.commons.lang3.ArrayUtils;
```
andreipaduroiu, line 40: Unrelated.

ravisharda: This was necessary, as it was depending on a transitive dependency coming via Hadoop client. Once I upgraded Hadoop client, the code compilation failed here. So, I used the replacement that was being brought in.

RaulGracia (approved):
LGTM, but it would be advisable to run system tests with HDFS (docker swarm) against this branch.

ravisharda:
> LGTM, but it would be advisable to run system tests with HDFS (docker swarm) against this branch.

@RaulGracia I've run this branch successfully with HDFS using a Docker Compose cluster instead (as described in the `Description` > `How to Test` section).

andreipaduroiu (approved):

PR #5343, opened 2020-11-17
Issue 5341: Fix thread leak in controller

### LINKED ISSUE #5341: Periodic watermarking is leaking threads from connection factory that it craetes as it is not closed when the object is closed
**Problem description**
In periodic watermarking, we first create a connection factory for the stream's scope followed by creating the revisioned stream client. we keep a reference to revisioned stream client and do close it but not the connection factory. 

Also, a cache limit of 1000 on watermarking for caching revisioned stream client may be too high and can lead to OOM on thread creation as each revisioned stream client has an internal thread and we are seeing that on k8s we get oom at around 700-800 threads on the controller process.

**Problem location**

close the connection factory apart from revisioned stream client.

shiveshr:

**Change log description**  
Fixes thread leak in Periodic watermarking where we create client factory object for each new revisioned client that we create, but when we close the client we do not close the factory which leaks a thread that it had created internally. 

**Purpose of the change**  
Fixes #5341 

**What the code does**  
Closes the factory as part of closing the watermark client.

**How to verify it**  
Unit test added


controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
```diff
@@ -387,14 +391,28 @@ boolean checkExistsInCache(Stream stream) {
...
-            this(stream, SynchronizerClientFactory.withScope(stream.getScope(), clientConfig));
+            this(stream, SynchronizerClientFactory.withScope(stream.getScope(), clientConfig), true);
         }
         
         @VisibleForTesting
         WatermarkClient(Stream stream, SynchronizerClientFactory clientFactory) {
-            this.client = clientFactory.createRevisionedStreamClient(
-                    NameUtils.getMarkStreamForStream(stream.getStreamName()), 
-                    new WatermarkSerializer(), SynchronizerConfig.builder().build());
+            this(stream, clientFactory, false);
```
shrids, line 399: closeClientFactory is set to false only for tests. Hence we could remove the condition at line 411 below.

controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
```diff
@@ -387,14 +391,28 @@ boolean checkExistsInCache(Stream stream) {
         private final ConcurrentHashMap<String, Long> inactiveWriters;
         
         WatermarkClient(Stream stream, ClientConfig clientConfig) {
-            this(stream, SynchronizerClientFactory.withScope(stream.getScope(), clientConfig));
+            this(stream, SynchronizerClientFactory.withScope(stream.getScope(), clientConfig), true);
```
shrids, line 394: Can we avoid creating a new SynchronizerClientFactory for every WatermarkClient? Every factory creates an new thread pool.

shrids (changes requested):

controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
```diff
@@ -387,14 +391,28 @@ boolean checkExistsInCache(Stream stream) {
...
-            this.client = clientFactory.createRevisionedStreamClient(
-                    NameUtils.getMarkStreamForStream(stream.getStreamName()), 
-                    new WatermarkSerializer(), SynchronizerConfig.builder().build());
+            this(stream, clientFactory, false);
+        }
+        
+        @VisibleForTesting
+        WatermarkClient(Stream stream, SynchronizerClientFactory clientFactory, boolean closeClientFactory) {
+            this.synchronizerClientFactory = clientFactory;
+            this.closeClientFactory = closeClientFactory;
```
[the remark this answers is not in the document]
shiveshr: removed this. its an internal constructor which is made visible for testing. so dont need preconditions.

controller/src/main/java/io/pravega/controller/server/bucket/PeriodicWatermarking.java
shiveshr: removed closeclientfactory

tkaitchuck (approved):

shrids (approved):

shrids:
System test run : test-pravega-continuous/3308/

PR #5346, opened 2020-11-17
Issue 5330: SLTS - Handle partially written chunks and snapshots after failures.

### LINKED ISSUE #5330: SLTS - Improve robustness and efficiency of SystemJournal snapshots
**Problem description**
SLTS has two special segments for storing storage metadata. The layout changes to these two segments are written in special system journal chunks as logs of change records.

During SLTS bootstrap, SystemJournal reads latest snapshot and applies entries in the system journal on top of it. 
At the end of the bootstrap it creates new snapshot based on current state and saves it.

There are 3 related problems 
1. SLTS is not deleting any journal files once snapshot is written. This creates garbage.
2. STLS is not skipping over partially written/invalid snapshots.
3. STLS is not snapshotting journal periodically.   

**Problem location**
SystemJournal

1. SLTS should delete any journal files once snapshot is written
2. STLS should skip over partially written/invalid snapshots.
3. STLS is not snapshotting periodically.   Initially size based (i.e when enough changes have been accumulated)

sachin-j-joshi:
During SLTS bootstrap, correctly handle partially written snapshots (during fail over) by skipping over them.

**Change log description**  
Correctly handle partially written chunks and snapshots after failures by skipping over them.

**Purpose of the change**  
Fixes #5330 

**What the code does**  
It fixes following failures

- If segment store failed while it was writing system journal snapshot then that partially written data can not be deserialized and causes EOFException. Code handles it and skips over the partial snapshot.

- When a previous write failure leaves behind partially written chunk it will generate InvalidOffsetException, the code handles it and skips over partial chunk.

**How to verify it**  
I have added the new unit tests.
build should pass


eolivelli (commented):
I am missing one part of the story here.
The new tests you added inject failures but even in spite of those failures the checks at the end of the test are unchanged.
Is it possible to test a broken snapshot with garbage?
How is it possible that losing a snapshot is not meaningful for the system? Probably we miss some test case.
Usually when you miss a snapshot you have to recover from other data, the tests are not testing of this recovery happens.

Probably I am simply missing something.

andreipaduroiu (changes requested):
In addition to the comments above:
1. I am not comfortable with relying on "functional" tests here. You have full control over your tests in these unit tests and can do everything you want. You should be able to corrupt your test data in any way you want. Please either update the language in the PR description or update your tests to be more thorough. Looking at the tests, they already seem to be simulating such corruption. What am I missing?
2. What is the expected behavior and/or outcome if we are unable to read the snapshot? Can we read the previous one? Would that help?
3. From what I remember in how this journal works, you write a new snapshot, then you truncate the old one away. So if we failed to write the new one, no truncation should have happened; as such, the old snapshot should still be there.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -324,6 +333,35 @@ public void testSimpleBootstrapWithMultipleFailovers() throws Exception {
...
     }
 
+    @Test
+    public void testSimpleBootstrapWithIncompleteSnapshot() throws Exception {
+        val containerId = 42;
+        ChunkStorage chunkStorage = getChunkStorage();
+        testSimpleBootstrapWithMultipleFailovers(containerId, chunkStorage, epoch -> {
+            val snapShotFile = NameUtils.getSystemJournalFileName(containerId, epoch, 0);
+            val size = 1;
+            if (chunkStorage.supportsTruncation()) {
```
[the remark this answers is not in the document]
sachin-j-joshi: This code is run against all providers (in bindings project) - not just in memory.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -250,9 +250,14 @@ private long applyLatestSnapshot(MetadataTransaction txn, HashMap<String, Long>
         for (epochToCheck = epoch - 1; epochToCheck >= 0; epochToCheck--) {
             snapshotFile = getSystemJournalChunkName(containerId, epochToCheck, 0);
             if (chunkStorage.exists(snapshotFile).get()) {
```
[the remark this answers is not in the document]
sachin-j-joshi: No we did not change it.

sachin-j-joshi:
@eolivelli  SLTS treats storage metadata segment and container metadata segment ( "storage system segments") in a special way.  To avoid circular dependency, records about **_changes to their chunk layout_** are stored in journal chunk(s) and replayed back at boot time.
To avoid replaying the whole change log. Snapshots (containing results of application of change logs so far) are created so that process can start from the last known point.
Also note that each instance has its own journal chunks and that they not shared between the instances.

Currently we don't yet delete any snapshots or journal files older than current snapshots after creating the new snapshot. we leave them behind. That is why when missing or corrupted snapshot is found we can simply skip over that snapshot and read the previous one and start applying changes from that point onwards. 

The test is a complex one. Each instance writes some valid data and then test "failovers" by starting a new instance. The new instance replays journal and creates a snapshot as part of bootstrap, it also simulates old zombie instance simultaneously continuing to write stale/garbage data after failover.  At the end with multiple failovers and bunch of garbage added to journals we still recover the valid data.

Because we are not yet deleting old journal files - just corrupting/deleting most recent snapshot is good enough.

In next PR - I'm planning to implement garbage collection and compaction of journal (The behavior does not change, unless snapshot is successfully written it should not delete old journal/most recent snapshots). That time I'll need to rewrite some part of these tests.

sachin-j-joshi:
@andreipaduroiu I had forgotten to update the comment after adding unit tests. I updated the comment. You are right the tests are already doing what you outlined above.

sachin-j-joshi:
@andreipaduroiu 

- SystemJournal is made up of series of chunks
- Each instance writes to its own chunks and never reuses/overwrites chunk owned by other instance.
- Currently we are not yet deleting old snapshot and journal chunks. So we can just skip over the corrupted/incomplete chunk and just start applying changes from the previous snapshot .

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -229,10 +230,23 @@ public void commitRecords(Collection<SystemJournalRecord> records) throws ChunkS
         }
         // Persist
         synchronized (lock) {
-            writeToJournal(bytes);
-            // Add a new log file if required.
-            if (!chunkStorage.supportsAppend() || !config.isAppendEnabled()) {
-                startNewJournalFile();
+            boolean done = false;
+            while (!done) {
```
andreipaduroiu, line 234: Can you do this with the `Retry` class (in `common`). We have various retry utilities that can help with your syntax here.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -229,10 +230,23 @@ public void commitRecords(Collection<SystemJournalRecord> records) throws ChunkS
...
+                try {
+                    writeToJournal(bytes);
+                    done = true;
+                } catch (ExecutionException e) {
```
andreipaduroiu, line 238: Why just `ExecutionException`? Can it be `Exception`, `CompletionException`? `Exceptions.unwrap` will handle all of those.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
sachin-j-joshi: I need to retry only if specific (wrapped) exception was thrown. I don't think Retry can do that.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
sachin-j-joshi: I want to catch and handle a very specific exception InvalidOffsetException wrapped in ExecutionException.
Other exception should bubble up and loop should not be retried here.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
andreipaduroiu: `.retryWhen(ex -> Exceptions.unwrap(ex) instanceof xyz)`

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
andreipaduroiu: What if that exception is wrapped in CompletionException?

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
```diff
@@ -719,6 +719,38 @@ public void testWrite() throws Exception {
...
+
+    /**
+     * Test Write after repeated failure.
+     *
+     * @throws Exception Exception if any.
+     */
+    @Test
+    public void testWriteAfterWriteFailure() throws Exception {
+        String testSegmentName = "foo";
+        TestContext testContext = getTestContext();
```
andreipaduroiu, line 731: Just like in the other PR, please add cleanup methods to this class.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -257,6 +258,79 @@ public void testSimpleBootstrapWithTwoFailovers() throws Exception {
...
     }
 
+    /**
+     * Tests a scenario when journal chunks may be partially written during failure.
+     *
+     * @throws Exception Throws exception in case of any error.
+     */
+    @Test
+    public void testSimpleBootstrapWithPartialDataWrite() throws Exception {
+        ChunkStorage chunkStorage = getChunkStorage();
```
andreipaduroiu, line 268: Cleanup

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -257,6 +258,79 @@ public void testSimpleBootstrapWithTwoFailovers() throws Exception {
...
+        if (!chunkStorage.supportsAppend()) {
+            return;
+        }
+
+        ChunkMetadataStore metadataStore = getMetadataStore();
```
andreipaduroiu, line 273: Cleanup

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -257,6 +258,79 @@ public void testSimpleBootstrapWithTwoFailovers() throws Exception {
...
+        ChunkMetadataStore metadataStoreAfterCrash = getMetadataStore();
```
andreipaduroiu, line 274: Cleanup

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -257,6 +258,79 @@ public void testSimpleBootstrapWithTwoFailovers() throws Exception {
...
+        int containerId = 42;
+        String systemSegmentName = SystemJournal.getChunkStorageSystemSegments(containerId)[0];
+        long epoch = 1;
+        val policy = new SegmentRollingPolicy(4);
+        val config = ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder().defaultRollingPolicy(policy).build();
+
+        long offset = 0;
+
+        // Epoch 1
+        ChunkedSegmentStorage segmentStorage1 = new ChunkedSegmentStorage(containerId, chunkStorage, metadataStore, executorService(), config);
```
andreipaduroiu, line 284: Cleanup

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -257,6 +258,79 @@ public void testSimpleBootstrapWithTwoFailovers() throws Exception {
...
+        segmentStorage1.write(h, offset, new ByteArrayInputStream(b2), b2.length, null).join();
+        offset += b2.length;
+
+        checkSystemSegmentsLayout(segmentStorage1);
+
+        // Step 2
+        // Start container with epoch 2
+        epoch++;
+
+        ChunkedSegmentStorage segmentStorage2 = new ChunkedSegmentStorage(containerId, chunkStorage, metadataStoreAfterCrash, executorService(), config);
```
andreipaduroiu, line 317: cleanup

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -267,9 +341,12 @@ public void testSimpleBootstrapWithTwoFailovers() throws Exception {
      */
     @Test
     public void testSimpleBootstrapWithMultipleFailovers() throws Exception {
+        val containerId = 42;
         ChunkStorage chunkStorage = getChunkStorage();
```
andreipaduroiu, line 345: cleanup

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -305,7 +382,12 @@ public void testSimpleBootstrapWithMultipleFailovers() throws Exception {
             oldHandle = h;
         }
 
+        if (null != faultInjection) {
+            faultInjection.accept(epoch);
+        }
+
         epoch++;
+
         ChunkMetadataStore metadataStoreFinal = getMetadataStore();
```
andreipaduroiu, line 391: cleanup to all of these

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -324,6 +406,35 @@ public void testSimpleBootstrapWithMultipleFailovers() throws Exception {
         Assert.assertEquals(expected, actual);
     }
 
+    @Test
+    public void testSimpleBootstrapWithIncompleteSnapshot() throws Exception {
+        val containerId = 42;
+        ChunkStorage chunkStorage = getChunkStorage();
```
andreipaduroiu, line 412: cleanup

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -324,6 +406,35 @@ public void testSimpleBootstrapWithMultipleFailovers() throws Exception {
...
+        testSimpleBootstrapWithMultipleFailovers(containerId, chunkStorage, epoch -> {
+            val snapShotFile = NameUtils.getSystemJournalFileName(containerId, epoch, 0);
+            val size = 1;
+            if (chunkStorage.supportsTruncation()) {
+                chunkStorage.truncate(ChunkHandle.writeHandle(snapShotFile), size).join();
+            } else {
+                val bytes = new byte[size];
+                chunkStorage.read(ChunkHandle.readHandle(snapShotFile), 0, size, bytes, 0).join();
+                chunkStorage.delete(ChunkHandle.writeHandle(snapShotFile)).join();
+                val h = chunkStorage.create(snapShotFile).join();
```
andreipaduroiu, line 422: Are you cleaning up this file anywhere? Leaving files behind is also a leak

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -324,6 +406,35 @@ public void testSimpleBootstrapWithMultipleFailovers() throws Exception {
...
+                chunkStorage.write(h, 0, size, new ByteArrayInputStream(bytes)).join();
+            }
+        });
+    }
+
+    @Test
+    public void testSimpleBootstrapWithMissingSnapshot() throws Exception {
+        val containerId = 42;
+        ChunkStorage chunkStorage = getChunkStorage();
```
andreipaduroiu, line 431: cleanup

andreipaduroiu (changes requested):

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
sachin-j-joshi: added

sachin-j-joshi (commented):
@andreipaduroiu I have added the cleanup.
Thanks

andreipaduroiu (approved):

PR #5355, opened 2020-11-18
Issue 5345: (SegmentStore) Rewiring StorageWriter into DurableLog

### LINKED ISSUE #5345: Improve StorageWriterDataSource Reading
**Problem description**
StorageWriterDataSource.read() performs a O(n) lookup to find the first item it aims to read. 

**Problem location**
DurableLog/StorageWriter/SequenceItemList.

One of these should improve things:
1. Remember the last item read. When reading next, begin the lookup from that last read item. Reading is done in sequence so this should be an easy fix.
2. Replace SequenceItemList with a BlockingDrainingQueue. There is no need to keep these items in memory after serving them to the StorageWriter. 

Benchmark the two approaches and see which one turns out to be better.

andreipaduroiu:
**Change log description**  
Changed the way `DurableLog` holds Tier 1 Operations in memory and how they are read by the Storage Writer. Retired uses of `SequencedItemList` in this case and replaced with a `BlockingDrainingQueue`.

Changed the internals of `BlockingDrainingQueue` and `PriorityBlockingDrainingQueue` to use a variation of `ArrayDeque`  that is more efficient at polling (removing) items in bulk.

**Purpose of the change**  
Fixes #5345.

**What the code does**  
- `SimpleDeque`: `ArrayDeque`-like implementation that has an extra method (`pollFirst(int)`) which efficiently removes a number of items from the head of the queue. This is more efficient than pulling one item out at a time and accumulating the result in a collection.
- Updated `BlockingDrainingQueue` and `PriorityBlockingDrainingQueue` to make use of `SimpleDeque` internally (in lieu of `ArrayDeque`).
    - Both of these classes are meant to extract items in bulk (1000 or 10,000 at once), so such a method should improve runtime performance.
- `DurableLog`
    - Changed the "In-Memory Log" to not be an identical replica of the Operations that exist in Tier 1.
    - This Log now contains only those Operations that have been written to Tier 1 but not yet picked up for processing by the `StorageWriter`. As soon as the `StorageWriter` picks them up, they are removed from this in-memory log.
    - Using a `BlockingDrainingQueue` to implement this in-memory log.
    - This required an update to the `DurableLog.read` method (new contract, new signature) and triggered some changes downstream in the `StorageWriter` (to adapt to the new API).
    - This simplified the internals of the `DurableLog` quite a bit. All the complex reading, including tail reads with timeout, is now delegated to the `BlockingDrainingQueue`, which is much nimbler at this task (that class is used twice in the `OperationProcessor` already).

**How to verify it**  
All tests must pass.


RaulGracia (approved):
LGTM, first performance tests show a great improvement with this change. Anyway, let's wait a bit before merging it to verify that no other issues appear from performance tests.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+        return (T) this.items[this.head];
+    }
+
+    /**
+     * Gets a value indicating the number of items in the {@link SimpleDeque}.
+     *
+     * @return The size of the {@link SimpleDeque}.
+     */
+    public int size() {
+        final int size = this.tail - this.head;
```
tkaitchuck, line 163: The correctness of this function and several others are predicated on the fact that there is always at least one empty position in the array. This appears to be enforced but it probably worth explicitly stating in a comment somewhere.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+        if (newCapacity >= MAX_HALF) {
+            // We double capacity, so if we're already above half of max, then just set it to max.
+            newCapacity = MAX_CAPACITY;
+        } else {
+            // Double capacity.
+            newCapacity = newCapacity * 2;
+        }
+
+        final Object[] newItems = Arrays.copyOf(this.items, newCapacity);
+        boolean flipped = this.tail < this.head || (this.tail == this.head && newItems[this.head] != null);
```
tkaitchuck, line 207: Based on the above definition of addLast I don't believe `this.tail == this.head && newItems[this.head] != null` can ever be true.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+        if (flipped) {
+            final int diff = newCapacity - this.items.length;
+            System.arraycopy(newItems, this.head, newItems, this.head + diff, this.items.length - this.head);
+            int idx = this.head;
+            this.head += diff;
+            while (idx < this.head) {
```
tkaitchuck, line 213: Can't a fill do this? (Or it could be avoided per my above comment)

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+
+        if (newCapacity >= MAX_HALF) {
+            // We double capacity, so if we're already above half of max, then just set it to max.
+            newCapacity = MAX_CAPACITY;
+        } else {
+            // Double capacity.
+            newCapacity = newCapacity * 2;
+        }
+
+        final Object[] newItems = Arrays.copyOf(this.items, newCapacity);
```
tkaitchuck, line 206: Looking at the implementation of this method, it would be more efficient to just allocate a new array, then copy the part after head and then copy the part up until tail if it was wrapped.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+        public T poll() {
+            if (this.nextIndex >= this.items.length) {
+                return null;
+            }
+
+            return (T) this.items[this.nextIndex++];
+        }
+
+        @Override
+        public T element() {
```
tkaitchuck, line 250: Per the documentation here:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Queue.html
Element is a variation on `peek` not on `poll`.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean remove(Object o) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public T remove() {
```
tkaitchuck, line 291: Here `remove` is unimplemented but the function `element` above is doing what remove is supposed to be doing.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+        return String.format("Size = %s, Capacity = %s, Head = %s, Tail = %s", size(), this.items.length, this.head, this.tail);
+    }
+
+    //endregion
+
+    //region WrapQueue
+
+    @RequiredArgsConstructor
+    @NotThreadSafe
+    private static class WrapQueue<T> extends AbstractCollection<T> implements Queue<T> {
```
tkaitchuck, line 233: Javadoc

common/src/test/java/io/pravega/common/util/PriorityBlockingDrainingQueueTests.java
```diff
@@ -155,7 +155,7 @@ public void testMultiplePriorities() {
         Assert.assertEquals("Unexpected size.", expectedItems.size(), q.size());
         Assert.assertSame("Unexpected peek.", expectedItems.first(), q.peek());
 
-        val finalPoll = q.poll(100);
+        val finalPoll = new ArrayDeque<>(q.poll(100)); // Make a copy since we need to modify it.
```
tkaitchuck, line 158: Shouldn't poll remove the item from `q`, in which case it won't matter if you modify it?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/InMemoryLog.java
```diff
@@ -0,0 +1,51 @@
...
+     * @throws OutOfOrderOperationException If item's Sequence Number is out of order.
+     */
+    @Override
+    protected void addInternal(Operation item) {
+        if (this.lastSequenceNumber >= item.getSequenceNumber()) {
+            throw new OutOfOrderOperationException(String.format("Operation '%s' is out of order. Expected sequence number of at least %s.",
+                    item, this.lastSequenceNumber));
+        }
+        super.addInternal(item);
+        this.lastSequenceNumber = item.getSequenceNumber();
```
tkaitchuck, line 43: These two operations are not thread safe but this class extends `BlockingDrainingQueue` which calls itself threadsafe.

There are multiple race conditions:
* There could be two threads the first perform the check and then both call `addInternal` even though the second one is before the first violating the invariant above.
* There could be two threads which insert in one order but set lastSequenceNumber in the other order resulting in the wrong sequence number.

tkaitchuck (changes requested):

common/src/main/java/io/pravega/common/util/SimpleDeque.java
```diff
@@ -0,0 +1,315 @@
...
+
+    //endregion
+
+    //region Constructor
+
+    /**
+     * Creates a new instance of the {@link SimpleDeque} class.
+     */
+    public SimpleDeque() {
+        this(16);
```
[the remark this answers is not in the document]
andreipaduroiu: This class is very similar to Java's `ArrayDeque`. That class also chooses 16 as a default init value.

For our purposes (`BlockingDrainingQueue`) this initial value makes little difference. We only create the queue once and then hang on to it. So it will eventually grow to a "steady-state" length and then reuse the same element array.

common/src/main/java/io/pravega/common/util/AbstractDrainingQueue.java
```diff
@@ -147,6 +152,41 @@ public void add(T item) {
...
+                    if (this.pendingTake == result) {
+                        this.pendingTake = null;
+                    }
+                }
+
+                // Timeout the future after we have unregistered it. A zealous callback may invoke us immediately and be
+                // surprised that their call is rejected since we can only have one take() outstanding at any given time.
+                result.completeExceptionally(new TimeoutException());
+            }, timeout.toMillis(), TimeUnit.MILLISECONDS);
+            result.whenComplete((r, ex) -> sf.cancel(true));
```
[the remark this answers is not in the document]
andreipaduroiu: Yes, I am aware of that. All of our executors have that flag set (see `ExecutorServiceHelpers.java`) precisely for this reason. 

FYI, this code is borrowed from `Futures.futureWithTimeout`. I didn't want to reuse that method since I wanted the leanest possible implementation which was tied into the `take()` future above.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
andreipaduroiu: I updated the class Javadoc with implementation details.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
andreipaduroiu: Yes you are correct. I looked too much at the `ArrayDeque.java` class

common/src/main/java/io/pravega/common/util/SimpleDeque.java
andreipaduroiu: Arrays.fill. does exactly the same thing. But since I used it everywhere else in this class, I might as well do it here too.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
andreipaduroiu: Good idea.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
andreipaduroiu: Fixed

common/src/main/java/io/pravega/common/util/SimpleDeque.java
andreipaduroiu: Fixed.

common/src/main/java/io/pravega/common/util/SimpleDeque.java
andreipaduroiu: Fixed.

common/src/test/java/io/pravega/common/util/PriorityBlockingDrainingQueueTests.java
andreipaduroiu: `poll(100)` returns one of those `SimpleDeque.WrapQueue` instances, to which we cannot add anything. Previously, it returned an `ArrayDeque` to which we could add new items, so the test code was lazily reusing that one. 

Since the `WrapQueue` is closed to additions, I have to make a copy of it. Copying is done using its iterator, so all its elements are contained in the same order. There is no need to test the functionality of `WrapQueue` - that is already done in the `SimpleDequeTests` class.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/InMemoryLog.java
andreipaduroiu: Look at the javadoc for `lastSequenceNumber`. This method can only be invoked from its superclass which holds a lock while doing so. As such, no further synchronization is needed.

tkaitchuck: Ah, ok. Please add the @guardedby annotation on both this and the BlockingDrainingQueue.addInternal

tkaitchuck (approved):

PR #5363, opened 2020-11-23
Issue 4902: (SLTS) - Implement more efficient eviction cache for ReadIndexCache

### LINKED ISSUE #4902: PDP-34 Simplified-LTS : Implement more efficient eviction cache for ReadIndexCache in ChunkManager layer.
**Problem description**
ReadIndexCache is supposed to cache the most recently used `<start offset, chunk name>` tuples in memory using LRU strategy.
During eviction the current implementation scans all cached entries for given segment which is inefficient.

**Problem location**
io.pravega.segmentstore.storage.chunklayer.ReadIndexCache::evictChunks

Use Guava cache as underlying cache implementation. 
Correctly handle cleaning up auxiliary data structures when entries are evicted. (Eg ConcurrentSkipList based per segment index)

sachin-j-joshi:

**Change log description**  
Use Guava cache as underlying cache implementation for chunklayer.ReadIndexCache.
Correctly handle cleaning up auxiliary data structures when entries are evicted. (Eg ConcurrentSkipList based per segment index)
Get rid of per segment limit.

**Purpose of the change**  
Fixes #4902 
ReadIndexCache is supposed to cache the most recently used <start offset, chunk name> tuples in memory using LRU strategy.
During eviction the current implementation scans all cached entries for given segment which is inefficient.

**What the code does**  
Use Guava cache as underlying cache implementation for chunklayer.ReadIndexCache.
It has two caches 
1. Cache that holds segment name to reverse lookup Index <start offset, chunk name> map. 
2. Cache for tracking all chunk index entries.

It correctly handles cleaning up auxiliary data structures when entries are evicted. (Eg ConcurrentSkipList based per segment index)
It also gets rid of per segment limit.

**How to verify it**  
All tests should pass.


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadIndexCache.java
```diff
@@ -86,59 +65,35 @@ public ReadIndexCache(int maxIndexedSegments, int maxIndexedChunksPerSegment, in
...
-        SegmentReadIndex readIndex = segmentsToReadIndexMap.get(streamSegmentName);
-
-        if (null == readIndex) {
-            // Evict segments if required.
-            if (maxIndexedSegments < segmentsToReadIndexMap.size() + 1 || maxIndexedChunks < totalChunkCount.get() + 1) {
-                evictSegmentsFromOldestGeneration();
+    private SegmentReadIndex getSegmentReadIndex(String streamSegmentName, boolean createIfNotPresent) {
+        SegmentReadIndex readIndex = segmentsReadIndexCache.getIfPresent(streamSegmentName);
+        if (null == readIndex && createIfNotPresent) {
+            synchronized (segmentsReadIndexCache) {
```
andreipaduroiu, line 71: Guava cache is synchronized. Why are you re-synchronizing?

andreipaduroiu (changes requested):
Consider using `SimpleCache` in `common`. Based on what I observe you use Guava for (simple, in-memory computations), the `SimpleCache` should be much faster than the Guava  Cache. 

See `DynamicLoggerImpl.java` and `SegmentStatsRecorderImpl.java` for examples of how this has been used to replace Guava Cache. For these instances, the Guava Cache had enough overhead to show up substantially (4%+) in our flame graphs, which is the reason that `SimpleCache` exists now.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadIndexCache.java
sachin-j-joshi: Multiple threads might find that entry does not exist and might race to create the entry for the first time. (This code is invoked only once per segment)

I did not find clean putIfNotPresent equivalent method.

andreipaduroiu: SimpleCache has putIfAbsent. Check it out

sachin-j-joshi:
@andreipaduroiu For now let's keep Guava based cache. 
I'll replace it later with SimpleCache as a separate PR after running benchmark.

andreipaduroiu (approved):
Please file a Github issue for 0.10.0 and assign it to yourself to make use of the SimpleCache.

PR #5364, opened 2020-11-23
Issue 5360: Require permissions on the scope for creating key-value-tables

### LINKED ISSUE #5360: Creating a Key-Value Table (KVT) requires write permissions on the non-existent KVT resource
**Problem description**
Currently, creating a Key-Value Table (KVT) requires the user write permissions on the KVT resource itself, which in turn is axiomatically non-existent at the time of the creation. For other resources like streams and reader-groups, creation requires the user account to have write permissions on the parent resource alone. 

**Problem location**

ravisharda:
**Change log description**  
Modify resource string used for authorizing Key-Value table creation operation from the resource to its parent resource. 

**Purpose of the change**  
Fixes #5360. 

**What the code does**  
Modifies the resource string used for creating Key-Value Tables from the resource itself (e.g., `prn::/scope:scope1/key-value-table:table1`) to its parent resource - the Scope (`prn::/scope:scope1`). 

**How to verify it**  


ravisharda:
Closed in favor of pull request #5366.

PR #5366, opened 2020-11-24
Issue 5360: Require permissions on the scope for creating Key-Value Tables

### LINKED ISSUE #5360: Creating a Key-Value Table (KVT) requires write permissions on the non-existent KVT resource

ravisharda:
**Change log description**  
Modify resource string used for authorizing Key-Value table creation operation from the resource to its parent resource. 

**Purpose of the change**  
Fixes #5360. 

**What the code does**  
Modifies the resource string used for creating Key-Value Tables from the resource itself (e.g., `prn::/scope:scope1/key-value-table:table1`) to its parent resource - the Scope (`prn::/scope:scope1`). 

**How to verify it**  


andreipaduroiu (approved):

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -146,7 +146,7 @@ public void createKeyValueTable(KeyValueTableConfig request, StreamObserver<Crea
                 scope, kvt);
         log.info(requestTag.getRequestId(), "createKeyValueTable called for KVTable {}/{}.", scope, kvt);
         authenticateExecuteAndProcessResults(() -> this.grpcAuthHelper.checkAuthorizationAndCreateToken(
-                authorizationResource.ofKeyValueTableInScope(scope, kvt), AuthHandler.Permissions.READ_UPDATE),
+                authorizationResource.ofKeyValueTablesInScope(scope), AuthHandler.Permissions.READ_UPDATE),
```
pbelgundi, line 149: Not just create but perhaps `listKeyValueTablesInScope` also needs to look at "scope" permissions instead of "scope + kvt" permission.

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
pbelgundi: I see that lisStreamsInScope (ControllerServiceImpl) uses `authorizationResource.ofStreamInScope(scopeName, x)`
where as StreamMetadataResourceImpl.listStreams uses `authorizationResource.ofStreamsInScope(scopeName)`...

I believe create/list operations should be based on permissions for parent resource - scope (in case of Streams/KVTs)

ravisharda: `listKeyValueTablesInScope` appears to be using the same model that we have applied in other listing operations like `listScopes` and `listStreamsInScope`, and looks OK to me. The general rule for applying authorization in listing operations is: 
* Require `READ` permission on the parent resource. (to check whether the user is allowed to invoke the listing operation in the first place)
* Filter in objects for which the user explicitly has `READ` access on. (so that only authorized objects are returned among the population of all objects).

To illustrate the general rule, taking an example of `listStreamsInScope` operation, say:

* `user1` has been assigned `READ` permission on `scope1` and `scope1/stream2`. 
* `user2` has been assigned `READ` on `scope1`.
* `user3` has been assigned `READ on `scope2`.
* The system has 3 streams under the scope - say, `scope1/stream1`,  `scope1/stream2` and  `scope1/stream3`

The `listStreamsInScope` will return: 
* A list containing a single stream `stream2 for `user1`
* An empty list for `user2`
* An authorization exception for `user3`. 

A similar rule seems to have been applied for Key-Value tables listing, which looks good to me. 

Now, switching to `StreamMetadataResourceImpl.listStreams`, it has a similar rule too, which is correct: 
* It first checks if the user has read access: 
https://github.com/pravega/pravega/blob/0e77c81a09a2f6019499259de7fa002398a3dc90/controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java#L530-L531
* It then adds only authorized streams into the list of returned items:
https://github.com/pravega/pravega/blob/0e77c81a09a2f6019499259de7fa002398a3dc90/controller/src/main/java/io/pravega/controller/server/rest/resources/StreamMetadataResourceImpl.java#L545-L546

So, the behaviors of the listing operations look OK to me. Am I missing something?

shared/security/src/main/java/io/pravega/shared/security/auth/AuthorizationResource.java
```diff
@@ -90,6 +90,16 @@
      */
     String ofReaderGroupInScope(String scopeName, String readerGroupName);
 
+    /**
+     * Creates a resource representation for use in authorization of actions pertaining to the collection of Key-Value
+     * tables within the specified scope.
+     *
+     * @param scopeName the name of the scope
+     * @throws NullPointerException if {@code scopeName} or {@code kvtName} are null
```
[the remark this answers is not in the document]
ravisharda: Sure, I've fixed it in the latest commit.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
pbelgundi: Thanks for explaining. It looks fine to me.

pbelgundi (approved):

ravisharda:
Started `Pravega/job/pravega-branch-builder/511/` with system tests (both auth enabled and disabled) for the branch to verify that the changes don't break any system tests.

PR #5368, opened 2020-11-24
Issue 5356 : Make Consumption based retention implicit on Subscriber Stream-Cut publish.

### LINKED ISSUE #5356: Consumption Based Retention Improvements
**Problem description**
Consumption based retention to be implicit based on Reader Groups publishing StreamCuts to controller instead of being explicitly set as a Retention Policy on Controller.

**Problem location**
Controller

https://github.com/pravega/pravega/pull/5304#discussion_r526449191

pbelgundi:
**Change log description**  
Make Consumption based retention implicit on Reader Groups publishing Stream-Cuts to Controller instead of an explicit Stream Retention-Policy on Controller.
Add maxLimit to Time/Space Retention Policies.

**Purpose of the change**  
Fixes #5356

**What the code does**  
- Remove Retention-Policy Type "CONSUMPTION"  and make consumption based retention implicit when Subscribers have published Stream-Cuts for truncation.
- Add a max limit to Time and Space based retention policy.

**How to verify it**  
All unit and Integration Tests should pass.

### DIFF
--- a/client/src/main/java/io/pravega/client/stream/RetentionPolicy.java
+++ b/client/src/main/java/io/pravega/client/stream/RetentionPolicy.java
@@ -40,8 +40,7 @@ public class RetentionPolicy implements Serializable {
     private final long retentionMax;
 
     /**
-     * Create a retention policy to configure a stream to periodically truncated
-     * according to the specified duration
+     * Create a retention policy to configure a stream to be periodically truncated according to the specified duration.
      *
      * @param duration Period to retain data in a stream.
      * @return Retention policy object.
--- a/client/src/main/java/io/pravega/client/stream/RetentionPolicy.java
+++ b/client/src/main/java/io/pravega/client/stream/RetentionPolicy.java
@@ -12,7 +12,6 @@ package io.pravega.client.stream;
 import java.io.Serializable;
 import java.time.Duration;
 
-import com.google.common.base.Preconditions;
 import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.Data;
@@ -33,17 +32,13 @@ public class RetentionPolicy implements Serializable {
         /**
          * Set retention based on the total size of the data in the stream in bytes.
          */
-        SIZE,
-
-        /**
-         * Set retention policy based on Consumption. 
-         */
-        CONSUMPTION
+        SIZE
     }
 
     private final RetentionType retentionType;
-    private final long retentionParam;
-    private final ConsumptionLimits consumptionLimits;
+    private final long retentionMin;
+    private final long retentionMax;
+
     /**
      * Create a retention policy to configure a stream to periodically truncated
      * according to the specified duration.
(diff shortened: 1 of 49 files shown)


client/src/main/java/io/pravega/client/stream/RetentionPolicy.java
```diff
@@ -33,18 +32,13 @@
...
-         * Also see: {@link ReaderGroupConfig.StreamDataRetention}
-         */
-        CONSUMPTION
+        SIZE
     }
 
     private final RetentionType retentionType;
-    private final long retentionParam;
-    private final ConsumptionLimits consumptionLimits;
+    private final long retentionMin;
```
shiveshr, line 39: please dont change the name of the variable otherwise anyone using retention policy in their application would have to recompile their code.

shiveshr: in fact to ensure that this change is backward compatible and we dont break the application, perhaps we can simply have a getter with the original name which returns the min.

pbelgundi: Fixed.

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
```diff
@@ -67,9 +66,6 @@ public static final StreamConfiguration getCreateStreamConfig(final CreateStream
                     retentionPolicy = getRetentionPolicy(createStreamRequest.getRetentionPolicy().getTimeBasedRetention(),
```
shiveshr, line 66: this should also take max and set it. same for size based.

shiveshr (changes requested):
Rest model helper needs to make use of max limits. 
apart from that it looks good.

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
pbelgundi: Fixed

pbelgundi:
> Rest model helper needs to make use of max limits.
> apart from that it looks good.

Fixed model helper now.

controller/src/main/java/io/pravega/controller/server/rest/ModelHelper.java
```diff
@@ -60,15 +59,29 @@ public static final StreamConfiguration getCreateStreamConfig(final CreateStream
...
+                    }
                     break;
                 case LIMITED_DAYS:
-                    retentionPolicy = getRetentionPolicy(createStreamRequest.getRetentionPolicy().getTimeBasedRetention(),
-                            createStreamRequest.getRetentionPolicy().getValue());
-                    break;
-                case CONSUMPTION:
-                    retentionPolicy = getConsumptionRetentionPolicy(createStreamRequest.getRetentionPolicy().getConsumptionLimits());
+                    if (createStreamRequest.getRetentionPolicy().getMaxValue() == 0
+                        && createStreamRequest.getRetentionPolicy().getMaxTimeBasedRetention().getDays() == 0
```
shiveshr, line 74: shouldnt this check simply be:
```
createStreamRequest.getRetentionPolicy().getMaxTimeBasedRetention() != null)
```
otherwise this access attempt itself can throw NPE.
Note: this is json derialization happening here, so the maxTimeBasedRetention will be defaulted to null if its not specified in the json payload.

pbelgundi: Fixed & also added tests to cover max Limit validation

shiveshr (approved):
LGTM

shiveshr:
started system test (500). will merge post successful system test run.

PR #5378, opened 2020-11-26
Issue 5377: Support TLS-enabled clusters for integration testing

### LINKED ISSUE #5377: Integration tests don't support TLS
**Problem description**
Currently, integration testing utilities like `ControllerWrapper` and `ClusterWrapper` that let you run a cluster to execute integration tests, don't support running a cluster with TLS setup. As a result, none of the [integration tests](https://github.com/pravega/pravega/tree/v0.8.0/test/integration/src/test/java/io/pravega/test/integration) under `:test:integration` Gradle project have testing running against a TLS enabled server. 

**Problem location**
https://github.com/pravega/pravega/blob/0e77c81a09a2f6019499259de7fa002398a3dc90/test/integration/src/main/java/io/pravega/test/integration/demo/ClusterWrapper.java

https://github.com/pravega/pravega/blob/1ad51211669ff59a90ca403be823079db6d60720/test/integration/src/main/java/io/pravega/test/integration/demo/ControllerWrapper.java

ravisharda:
**Change log description**  
Add support for TLS-enabled clusters in `ClusterWrapper` and `ControllerWrapper`, which are used for running servers for integration tests. 

**Purpose of the change**  
Resolves #5377. 

**What the code does**  
* Adds support for TLS in `ClusterWrapper` and `ControllerWrapper`. 
* Modify `ClusterWrapper` to support construction via a builder instead of constructors. This became useful as the number of construction parameters increased. Also, a builder makes the intent of each parameter explicit in the test code. 

**How to verify it**  
All integration tests must pass. 

New tests were added in `ClusterWrapperTest` to demonstrate the usage.


test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * A helper class with general-purpose utility methods for integration tests.
+ */
+@Slf4j
+public class TestUtils {
+
+    public static void createStreams(ClientConfig clientConfig, String scopeName, List<String> streamNames) {
```
andreipaduroiu, line 46: It would be nice to add a single-sentence Javadoc to each of these explaining what they do. Others might find them useful in the future;

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,115 @@
...
+            }
+        });
+    }
+
+    @SneakyThrows
+    public static void writeDataToStream(String scope1, String stream1, String message1, ClientConfig writerClientConfig) {
+        @Cleanup final EventStreamClientFactory writerClientFactory = EventStreamClientFactory.withScope(scope1,
+                writerClientConfig);
+
+        @Cleanup final EventStreamWriter<String> writer = writerClientFactory.createEventWriter(stream1,
```
andreipaduroiu, line 68: This is very inefficient. If you call this in a tight loop, then this will take a long time (due to constant reinitialization). Can you have this method take another argument that indicates how many events to write? Even if you don't use it now, someone in the future might find that useful.

PS: if you do that, no need to call `.get()` upon each `.writeEvent`. Collect the futures in a list, and use `Futures.allOf(...).get()`.

PS2: A timeout for that `.get` might be useful too to pass to this method. But your call.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,115 @@
...
+                new JavaSerializer<String>(),
+                EventWriterConfig.builder().build());
+        writer.writeEvent(message1).get();
+        log.info("Wrote a message to the stream {}/{}", scope1, stream1);
+    }
+
+    public static String readAMessageFromStream(String scopeName, String streamName, ClientConfig readerClientConfig,
```
andreipaduroiu, line 75: Same here. If you read just 1 message, this is OK. But as soon as you want to read more, then this method will be very slow.

Consider taking a count argument and returning an (ordered) list of Strings representing the events.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,115 @@
...
+
+        @Cleanup
+        EventStreamReader<String> reader = readerClientFactory.createReader(
+                "readerId", readerGroupName,
+                new JavaSerializer<String>(), ReaderConfig.builder().initialAllocationDelay(0).build());
+        log.debug("Created an event reader");
+
+        // Keeping the read timeout large so that there is ample time for reading the event even in
+        // case of abnormal delays in test environments.
+        String readMessage = reader.readNextEvent(10000).getEvent();
```
andreipaduroiu, line 98: It may surprise you but on Travis, this timeout may not be sufficient. Expect this to be a root cause for random failures.

If the timeout expires, you won't get a `TimeoutException`; rather the `getEvent()` will return null, which will likely manifest as a NullPtrEx upstream.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public static List<PasswordAuthHandlerInput.Entry> preparePasswordInputFileEntries(
+            Map<String, String> entries, String password) {
+        StrongPasswordProcessor passwordProcessor = StrongPasswordProcessor.builder().build();
+        try {
+            String encryptedPassword = passwordProcessor.encryptPassword(password);
+            List<PasswordAuthHandlerInput.Entry> result = new ArrayList<>();
+            entries.forEach((k, v) -> result.add(PasswordAuthHandlerInput.Entry.of(k, encryptedPassword, v)));
+            return result;
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
```
andreipaduroiu, line 111: You already used `SneakyThrows` in this class. You might as well use it here too.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Done.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Sure, I've made the change in the latest commit I just pushed.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Have made this change as suggested.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: I've increased the timeout to account for that.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Done.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,214 @@
...
+        StreamManager streamManager = StreamManager.create(clientConfig);
+        assertNotNull(streamManager);
+
+        boolean isScopeCreated = streamManager.createScope(scope);
+        assertTrue("Failed to create scope", isScopeCreated);
+
+        streams.forEach(s -> {
+            boolean isStreamCreated = streamManager.createStream(scope, s,
+                    StreamConfiguration.builder().scalingPolicy(ScalingPolicy.fixed(1)).build());
+            if (!isStreamCreated) {
```
shiveshr, line 68: false is returned when the stream already exists. 
so in the idempotent case, do you want to throw the exception from here or return the boolean from this method signature to let the caller know so that they can deal with it.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,214 @@
...
+        EventStreamReader<String> reader = readerClientFactory.createReader(
+                "readerId", readerGroup,
+                new JavaSerializer<String>(), ReaderConfig.builder().initialAllocationDelay(0).build());
+        log.debug("Created an event reader");
+
+        // Keeping the read timeout large so that there is ample time for reading the event even in
+        // case of abnormal delays in test environments.
+        List<String> result = new ArrayList<>();
+        for (int i = 0; i < numMessages; i++) {
+            result.add(reader.readNextEvent(20000).getEvent());
```
shiveshr, line 182: in case of a checkpoint event, you could get a null event. 
i think you should add event only when it is not a checkpoint event. 
or disable checkpoint in the readergroup config. 
otherwise that can lead to some tests to fail.

shiveshr (approved):
LGTM, couple of minor comments that should be addressed.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Sure, I've modified the implementation to return a boolean that indicates whether all the objects were newly created or not.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: We are assuming that the checkpoint is disabled in the reader group: 

https://github.com/pravega/pravega/blob/4788434e05b3cafc4dfc7e2bb2c56112ab6ac2ef/test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java#L161-L164

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,212 @@
...
+                writerClientConfig);
+
+        @Cleanup final EventStreamWriter<String> writer = writerClientFactory.createEventWriter(stream,
+                new JavaSerializer<String>(),
+                EventWriterConfig.builder().build());
+        List<CompletableFuture<Void>> futures = new ArrayList<>();
+        for (int i = 0; i < numMessages; i++) {
+            futures.add(writer.writeEvent(message));
+        }
+        Futures.allOf(futures).join();
```
shrids, line 117: we can simply use a writer.flush() here.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,212 @@
...
+                .stream(Stream.of(scope, stream))
+                .disableAutomaticCheckpoints()
+                .build();
+        @Cleanup
+        ReaderGroupManager readerGroupManager = ReaderGroupManager.withScope(scope, readerClientConfig);
+        readerGroupManager.createReaderGroup(readerGroup, readerGroupConfig);
+        log.debug("Created reader group with name {}", readerGroup);
+
+        @Cleanup
+        EventStreamReader<String> reader = readerClientFactory.createReader(
```
shrids, line 171: Since the reader group is not deleted invoking `readDataFromStream(..)` again would cause the method to read the next events from the stream. So we should document this behaviour or perhaps invoke delete reader group or change the name of the method to convey this behaviour perhaps as `readNextEvents(..)`

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Thanks. That change makes the code much simpler. I've made the change as suggested.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Sure, I've renamed the methods, so that they better represent the behavior that these methods exhibit.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
```diff
@@ -0,0 +1,209 @@
...
+            return messages.get(0);
+        }
+    }
+
+    /**
+     * Returns the specified number of unread messages from the given {@code scope}/{@code stream}.
+     *
+     * @param scope the scope
+     * @param stream the stream
+     * @param numMessages the number of event messages to write
```
shrids, line 142: typo.

test/integration/src/main/java/io/pravega/test/integration/demo/ClusterWrapper.java
```diff
@@ -65,52 +72,58 @@
...
-    private boolean isAuthEnabled;
-    private String tokenSigningKeyBasis;
-    private boolean isRGWritesWithReadPermEnabled;
-    private int tokenTtlInSeconds;
+    @Getter
+    @Builder.Default
+    private boolean authEnabled = false;
+
+    @Getter
+    @Builder.Default
```
shrids, line 83: `new ClustuerWrapper()` will ensure the defaults are not used. Should we make the default constructor private to enforce the builder pattern?

test/integration/src/main/java/io/pravega/test/integration/demo/ControllerWrapper.java
```diff
@@ -102,6 +103,21 @@ public ControllerWrapper(final String connectionString, final boolean disableEve
                              boolean enableAuth, String passwordAuthHandlerInputFilePath,
                              String tokenSigningKey, boolean isRGWritesWithReadPermEnabled,
                              int accessTokenTtlInSeconds) {
+        this (connectionString, disableEventProcessor, disableControllerCluster, controllerPort, serviceHost,
+                servicePort, containerCount, restPort, enableAuth, passwordAuthHandlerInputFilePath, tokenSigningKey,
+                isRGWritesWithReadPermEnabled, accessTokenTtlInSeconds, false, "", "", "", "", false);
+    }
+
+    @Builder
```
shrids, line 111: Does this @Builder generate a seperate builder?

shrids (approved):
The changes look good apart from a couple of questions/comments.

test/integration/src/main/java/io/pravega/test/integration/demo/ControllerWrapper.java
ravisharda: It generates a builder for the `ControllerWrapper` based on the parameters used in this Constructor (as opposed to the `Class-level` annotation we use in the `ClusterWrapper`. The idea was to avoid interface changes to `ControllerWrapper` so that the many tests that use it don't need to change.

test/integration/src/main/java/io/pravega/test/integration/utils/TestUtils.java
ravisharda: Great catch - truly a hard to find one! I've fixed it now.

test/integration/src/main/java/io/pravega/test/integration/demo/ClusterWrapper.java
ravisharda: Sure, I've now made the change to prevent instantiation using the default constructor.

PR #5386, opened 2020-11-30
Issue 5385: Upgrade Bookkeeper to last stable version (4.11.1)

### LINKED ISSUE #5385: Upgrade Bookkeeper to last stable version (4.11.1)
**Problem description**
We need to upgrade the Bookkeeper version in Pravega as the current one (4.9.2) is EOL and new versions contain important fixes for Pravega.

**Problem location**
Bookkeeper docker file (and maybe other files).

Upgrade Bookkeeper version.

RaulGracia:
**Change log description**  
Upgrades the version of Bookkeeper in Pravega to 4.11.1 (client, documentation, dockerfile).

**Purpose of the change**  
Fixes #5385.

**What the code does**  
Upgrades the versions of Bookkeeper in Pravega to 4.11.1 (client, documentation, dockerfile).
Apart from that, this PR removes an unnecessary method in `entrypoint.sh` called `fix_bk_ipv6_check()`, as the fix for which that method was intended to is already in Bookkeeper mainline.

**How to verify it**  
JUnit tests (travis/jenkins) and system tests (3/3) are passing. Longevity runs have also passed.


eolivelli:
In BK 4.11 docker image we switched to JDK11,

btw I suggest you to make the Java change in a separate patch if it is possible,
the BookieServer class changed in BK 4.11 in order to support BookieEndpointInfo (BP-38) and probably the code won't build

RaulGracia:
@eolivelli found the issue: the `bindv6only` check that was added in a previous upgrade is not necessary anymore, and actually it was making the script to crash. Fixed it now and also switched to JDK11 (I think it is fine to include this in the current draft).

eolivelli:
good

RaulGracia:
On a first local test (one Bookie on a container using `allowLoopback=true`), it looks like the Bookie started correctly:
```
2020-11-30 15:52:23,861 - INFO  - [main:Environment@109] - Client environment:java.version=11.0.8
2020-11-30 15:52:23,861 - INFO  - [main:Environment@109] - Client environment:java.vendor=N/A
2020-11-30 15:52:23,862 - INFO  - [main:Environment@109] - Client environment:java.home=/usr/lib/jvm/java-11-openjdk-11.0.8.10-0.el7_8.x86_64
2020-11-30 15:52:23,862 - INFO  - [main:Environment@109] - Client environment:java.class.path=/opt/bookkeeper/conf:/opt/bookkeeper/lib/org.apache.bookkeeper-bookkeeper-server-4.11.1.jar
...
2020-11-30 15:52:26,105 - INFO  - [main:ComponentStarter@86] - Started component bookie-server.
2020-11-30 15:52:41,940 - INFO  - [bookie-io-1-2:AuthHandler$ServerSideHandler$AuthHandshakeCompleteCallback@219] - Authentication success on server side
2020-11-30 15:52:41,977 - INFO  - [bookie-io-1-2:BookieRequestHandler@50] - Channel connected  [id: 0x902232e6, L:/127.0.1.1:3181 - R:/127.0.0.1:49262]
2020-11-30 15:52:42,464 - INFO  - [BookieHighPriorityThread-3181-OrderedExecutor-0-0:ReadEntryProcessorV3@208] - Ledger fence request received for ledger: 0 from address: /127.0.0.1:49262
2020-11-30 15:52:42,640 - INFO  - [bookie-io-1-2:BookieRequestHandler@61] - Channels disconnected: [id: 0x902232e6, L:/127.0.1.1:3181 ! R:/127.0.0.1:49262]
2020-11-30 15:52:56,614 - INFO  - [bookie-io-1-3:AuthHandler$ServerSideHandler$AuthHandshakeCompleteCallback@219] - Authentication success on server side
2020-11-30 15:52:56,615 - INFO  - [bookie-io-1-3:BookieRequestHandler@50] - Channel connected  [id: 0xb5fbc90d, L:/127.0.1.1:3181 - R:/127.0.0.1:49266]
```

eolivelli:
good.
The changes I am referring to are to some constructor of the BookieServer, if our testscases are running it means that we are not using those changed functions.
Unfortunately there is no official pragmatic API to start a Bookie and each project has its own code

docker/bookkeeper/entrypoint.sh
```diff
@@ -89,22 +89,6 @@ configure_bk() {
...
-   # an error code `255`. Since, the `bookkeeper/bin/bookeeper` shell script sources the `common.sh` file, and it sets
-   # "exit on error" as true, we encounter this error when it is invoked. To avoid that error, here we modify the
-   # `common.sh` file to execute `/sbin/sysctl -n net.ipv6.bindv6only` only if file `/proc/sys/net/ipv6/bindv6only`
-   # is available.
-
-   # Replace the 22nd line with "if [ -f /sbin/sysctl ] && [ -f /proc/sys/net/ipv6/bindv6only ]; then".
-   sed -i "22s|.*|if [ -f /sbin/sysctl ] \&\& [ -f /proc/sys/net/ipv6/bindv6only ]; then|" ${BK_HOME}/bin/common.sh
-}
-
 initialize_cluster() {
```
ravisharda, line 92: PR #4420 had added many of these functions from Apache/Bookkeeper which didn't work in stock Apache/Bookkeeper image at the time, so that they could be fixed here. The idea was that those fixes are made here, until the fixes are also made on Apache/Bookeeper side. 

Later, I had raised this PR https://github.com/apache/bookkeeper/pull/2219 with the same fixes we made here in Pravega and the PR got merged there, so some of these customized functions aren't needed anymore. As such we can return back to having pravega/bookkeeper entrypoint.sh invoking apache/bookkeeper entrypoint.sh to start bookies (see how this changed in the entrypoint.sh diff in https://github.com/pravega/pravega/pull/4420/files0. 

Now, whether we do that change here while upgrading to Bookkeeper `4.11.x` or do it separately in a different PR is an open question. I'd suggest doing that separately. If you agree, we should raise an issue to track that separate piece of work.

docker/bookkeeper/Dockerfile
```diff
@@ -17,9 +17,9 @@
 # - github.com/apache/bookkeeper/blob/branch-4.9/docker/Dockerfile
 #
 
-FROM apache/bookkeeper:4.9.2
+FROM apache/bookkeeper:4.11.1
```
ravisharda, line 20: This base image downloads `bookkeeper-server-4.11.1-bin.tar.gz` (as per this [layer](https://hub.docker.com/layers/apache/bookkeeper/4.11.1/images/sha256-87f13f6d73d44a01334e3b2bc1fc4796149677fc52c17097f9ccf82358c8ed92?context=explore)) and then we bring in `bookkeeper-all-${BK_VERSION}-bin.tar.gz` and use this new file. As a result, the image has an extra `bookkeeper-server-4.11.1-bin.tar.gz` file, adding to the image size.  This is a preexisting issue, which is present in our older images too. Would it make sense to add a command to delete that redundant binary here?

ravisharda (changes requested):

docker/bookkeeper/entrypoint.sh
RaulGracia: Agree, let's keep changes of this upgrade at minimum, but track this cleanup so we can do that on a separate PR.

docker/bookkeeper/Dockerfile
RaulGracia: @ravisharda I cannot find any extra `bookkeeper-server-4.11.1-bin.tar.gz` inside the container (I'm looking into it and cannot find it). I think that the reason is that this file is already being deleted in the Bookkeeper Dockerfile: https://github.com/apache/bookkeeper/blob/8d3060e925e0fb06c7a1880155a809fe28ea052b/docker/Dockerfile#L48
So, I think that what you are suggesting is already done, right?

ravisharda (approved):

andreipaduroiu (approved):

fpj (approved):

PR #5387, opened 2020-12-01
Issue 5358: Reduce/Optimize number of calls to segment store for metadata operations

### LINKED ISSUE #5358: Optimize metadata operations in Controller's StreamMetadataStore
**Problem description**
Controller's StreamMetadata store exposes apis whose implementation either retrieves the data from the cache or makes a call to table store. And we cache the immutable records. 
For an operation, like a user query or a workflow for instance, that may make multiple calls into the store, could end up fetching the same metadata from the table more than once. This is suboptimal and can be optimized by keeping reference of already retrieved objects in the OperationContext and only retrieve it from the store or cache if its absent in the operation context. 

**Problem location**
StreamMetadataStore

Implement the suggested improvement.

shiveshr:
**Change log description**  
Changes the metadata store implementation for PravegaTablesStream to keep the previously retrieved value in memory for the current operation context so that subsequent calls requesting the same value are served from memory.

**Purpose of the change**  
Fixes #5358 

**What the code does**  
Presently we follow a model that we cache all immutable metadata records and mutable records till a new request comes. And we fetch mutable records from the table store whenever requested to ignore cache. This is optimized slightly by maintaining a global cache of mutable records which were "refreshed" whenever a new request was initiated. So if there was a single request for a stream, then it mutable and immutable records were fetched once. But if there are multiple concurrent requests, each new request would clean up all the previously fetched mutable records from the global cache. 

This is sub optimal for a situation where if the subsequent request (which resulted in cache refresh) was not even interested in refreshed metadata records. it would then unnecessarily clean up the cache where by forcing first request to potentially request the values again. 

By keeping a fetched set of values within the operation context, we actually have the advantage of ensuring that each request is completely isolated from any other request. 
And secondly, each request independently maintains its retrieved records efficiently and only fetches from segment store when the business logic demands it (e.g. retrying on write conflict or if its tracking and watching completion of an outstanding work). 
and we continue to cache the mutable records which are never invalidated.. and the mutable ones are managed in the cache via a "time at cache load" and time for when the store's stream object was created for the request. and the value is taken from cache if the value loaded into cache was greater than the time when the request object was created. 

The way we have implemented it is as follows:
1. Earlier same stream object with different operation contexts were used by all concurrent request on the same stream. We changed it to make sure each request creates its own stream object which stores the retrieved metadata and serves from it locally. 
2. Only immutable records are loaded into and read from the global cache. All mutable records are only present within the operation context. 
3. reduces one call to segment store during commit transaction where we make use of commit offset from merge segment call. 
4. added load time in CacheValue. And when we do a get from cache, we compare the cached value's load time with the time when the streamObj was created, and if the streamobj is newer than cached value, then it fetches the value from the store otherwise it uses the cached value and maintains it in its context for subsequent calls.
**How to verify it**  
Unit tests updated and all tests should pass. and system test should pass.


shiveshr:
System test 3334 passed

controller/src/test/java/io/pravega/controller/store/stream/StreamMetadataStoreTest.java
```diff
@@ -330,7 +330,8 @@ public void partialStreamsInScope() throws Exception {
             return result;
         }).when(streamObjSpied).getConfiguration();
 
-        ((AbstractStreamMetadataStore) store).setStream(streamObjSpied);
+        // TODO: shivesh
```
andreipaduroiu, line 333: Fix this todo

RaulGracia:
@shiveshr do we want to get this one in for 0.9?

shiveshr:
@RaulGracia i would prefer that we get this in for 0.9. 
But its just an improvement/optimization and not a fix for a correctness issue. so even if we include it in 0.10 its also fine.

PR #5391, opened 2020-12-02
Issue 5390: (SLTS) Handle InvalidOffsetException in concat using appends mode and disable that mode when inapplicable.

### LINKED ISSUE #5390: SLTS - Concat small source objects using append mode should correctly handle InvalidOffsetException
**Problem description**
When "concat small source objects using append" mode is active and previous concat operation fails to commit, it may leave partially concatenated target object which has larger length on LTS than in metadata. This causes InvalidOffsetException.
This exception is "sticky" and the system can not make progress anymore.

**Problem location**
DefragementOperation (which is called inline to concat multiple segments from the source segment into target segment.)

Skip over unappendable chunk.

sachin-j-joshi:
**Change log description**  

- Concat small source objects using append mode now correctly handles possible failures with InvalidOffsetException
- Do not enable use of appends for concat of small objects when appends are disabled

**Purpose of the change**  
Fixes #5390 

- When "concat small source objects using append" mode is active and previous concat operation fails to commit, it may leave partially concatenated target object which has larger length on LTS than in metadata. This causes InvalidOffsetException. This exception is "sticky" and the system can not make progress anymore.

**What the code does**  
- Skip over chunk that has larger length on LTS than in metadata.
-  Do not enable use of appends for concat of small objects when appends are disabled

**How to verify it**  
Tests should pass.


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -144,11 +147,29 @@
...
+                                        if (ex instanceof InvalidOffsetException) {
+                                            // Skip ahead by 1 chunk.
+                                            targetChunkName = chunksToConcat.get(1).getName();
+                                            chunksToConcat.clear();
+                                            skipFailed.set(true);
+                                            return null;
+                                        }
+                                        throw new CompletionException(ex);
+                                    }
+                                    return v;
```
andreipaduroiu, line 163: Why `v` and not `vv`?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -144,11 +147,29 @@
...
+                                }, chunkedSegmentStorage.getExecutor());
                             } else {
                                 f = CompletableFuture.completedFuture(null);
                             }
                             return f.thenRunAsync(() -> {
```
andreipaduroiu, line 168: You do not need `Async` here

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -144,11 +147,29 @@
...
+                                if (skipFailed.get()) {
```
andreipaduroiu, line 169: `skipFailed.compareAndSet(true, false)`

andreipaduroiu (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
sachin-j-joshi: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
sachin-j-joshi: leaving it as it is - keeping the old pattern.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
sachin-j-joshi: fixed.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
```diff
@@ -1317,6 +1317,101 @@ public void testBaseConcatWithDefragWithMinMaxLimits() throws Exception {
...
+                                                         long expectedLength) throws Exception {
+        String targetSegmentName = "target";
+        String sourceSegmentName = "source";
+        SegmentRollingPolicy policy = new SegmentRollingPolicy(maxRollingSize); // Force rollover after every 20 byte.
+        ChunkedSegmentStorageConfig config = ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder()
+                .maxSizeLimitForConcat(100)
+                .minSizeLimitForConcat(100)
+                .build();
+
+        TestContext testContext = getTestContext(config);
```
andreipaduroiu, line 1383: Please modify `TestContext` to implement `AutoCloseable`, and close any resources it uses in the `close` method. Then update all usages of this class to clean up (decorate with `@Cleanup`).

andreipaduroiu (changes requested):

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
sachin-j-joshi: I made these changes in other PR, once that PR is merged and pulled into this PR I'll make this change.
Thanks

sachin-j-joshi: Done.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
```diff
@@ -1371,6 +1371,102 @@ public void testBaseConcatWithDefragWithMinMaxLimits() throws Exception {
...
+
+        // Write some garbage at the end.
+        val sourceList = TestUtils.getChunkList(testContext.metadataStore, sourceSegmentName);
+        for (int i : chunksWithGarbageIndex) {
+            // Append some data to the last chunk to simulate partial write during failure
+            val chunkMetadata = TestUtils.getChunkMetadata(testContext.metadataStore, sourceList.get(i).getName());
+            testContext.chunkStorage.write(ChunkHandle.writeHandle(chunkMetadata.getName()), chunkMetadata.getLength(), 1, new ByteArrayInputStream(new byte[1]));
+        }
+        val hTarget = testContext.chunkedSegmentStorage.openWrite(targetSegmentName).get();
+        testContext.chunkedSegmentStorage.concat(hTarget, 5, sourceSegmentName, null).join();
```
RaulGracia, line 1462: Is this validating the `InvalidOffsetException`, right? As it is not validating that such exception is thrown, it means that the  defragmentation will continue working as usual, skipping any number of such failures.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -144,11 +147,28 @@
...
+                                    if (null != ex) {
+                                        ex = Exceptions.unwrap(ex);
+                                        if (ex instanceof InvalidOffsetException) {
+                                            // Skip ahead by 1 chunk.
+                                            targetChunkName = chunksToConcat.get(1).getName();
+                                            chunksToConcat.clear();
+                                            skipFailed.set(true);
+                                            return null;
+                                        }
+                                        throw new CompletionException(ex);
```
RaulGracia, line 161: What other types of exceptions are expected here and why they are bubbled up (conversely to `InvalidOffsetException`)?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -144,11 +147,28 @@
...
+                                f = concatChunks()
+                                .handleAsync((vv, ex) -> {
+                                    if (null != ex) {
+                                        ex = Exceptions.unwrap(ex);
+                                        if (ex instanceof InvalidOffsetException) {
+                                            // Skip ahead by 1 chunk.
+                                            targetChunkName = chunksToConcat.get(1).getName();
+                                            chunksToConcat.clear();
+                                            skipFailed.set(true);
+                                            return null;
```
RaulGracia, line 159: Do we need to add logs reporting that this exception has been captured? Otherwise, how are we going to realize that the defragmentation process is getting `InvalidOffsetException`?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -144,11 +147,28 @@
                             CompletableFuture<Void> f;
                             if (chunksToConcat.size() > 1) {
                                 // Concat
-                                f = concatChunks();
+                                f = concatChunks()
+                                .handleAsync((vv, ex) -> {
+                                    if (null != ex) {
+                                        ex = Exceptions.unwrap(ex);
+                                        if (ex instanceof InvalidOffsetException) {
```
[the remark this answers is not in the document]
sachin-j-joshi: - We accidently encountered this issue while running longevity runs with wrong config. 
    - In SLTS concat operation is 3 step process - first we merge the chunk metadata list ,  then we defragment the chunks in that list and then finally commit all metadata changes.
    -  In this particular instance, the defragment operation was successful but the transaction to commit metadata failed. (because of internal error with table segment code).
    - This resulted in last chunk of target segment having all data from chunks of source segment but metadata still had old information.
    - When storage writer attempted the operation again, we found that for the last chunk of target segment the length on LTS did not match metadata and it correctly threw InvalidOffsetException
    - This results in unrecoverable error. SW can not make any progress and cache becomes full and everything comes to halt.
- Note that we DO NOT delete source chunks unless commit was successful.

- However this is a more generic failure mode and can happen for any of following reason
    - process crashed while doing defrag
    - some I/O error happened while doing defrag
    - the segment store failed over and the zombie instance continued to write to last chunk. (perfectly valid scenario for SLTS)
    - so on.
- Such kind of error is not data corruption - this is a "normal" and valid circumstance .
    - The appropriate handling is to skip over the chunk with garbage data at the end.
    - This kind of situation should never block StorageWriter from making progress.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
sachin-j-joshi: Yes. Here we are writing additional garbage data to chunk directly via `ChunkStorage` without SLTS knowing about it. And it tries to write at the wrong offset.
In this case the `ChunkStorage` contract requires it to throw `InvalidOffsetException` when write operation is called by SLTS with wrong offset - which it throws.

This way the test recreates the exact situation we are trying to fix.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
sachin-j-joshi: We are specifically handling only the `InvalidOffsetException` and nothing else. 
For others we just bubble them up.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
sachin-j-joshi: Good suggestion. I'll add logs.

sachin-j-joshi: added

RaulGracia (approved):

PR #5398, opened 2020-12-03
Issue 5382: Adding Java cacerts to openjdk:11.0.8-jre-slim Base Image

### LINKED ISSUE #5382: ECS certs are not added to java truststore with JDK11 images
**Problem description**
While installing Pravega which uses jdk11 - with ECS as LTS, the ecs certs are not imported into the java trust store and the error can be seen in SS logs
```
Trust this certificate? [no]:  Certificate was added to keystore
keytool error: java.io.FileNotFoundException: /etc/ssl/certs/java/cacerts
```
This is happening as openjdk:11.0.8-jre-slim doesn't have **'/etc/ssl/certs/java/cacerts'**  

**Problem location**
https://github.com/pravega/pravega/blob/master/docker/pravega/Dockerfile

I got the following issue which solves this issue which solves this problem on java 8 link to it https://github.com/docker-library/openjdk/issues/19.

Prabhaker24:
**Change log description**  
Adding ca-certificate to system truststore with openjdk:11.0.8-jre-slim images 

**Purpose of the change**  
Fixes #5382 

**What the code does**  
Following code changes add the ca certs on top of the openjdk:11.0.8-jre-slim image which we are using as Base Image. 

**How to verify it**  
To verify the changes exec into the ss pod and check the following
```
root@nautilus-pravega-segment-store-0:/opt/pravega# ls /etc/ssl/certs/java/cacerts
/etc/ssl/certs/java/cacerts
```
I have also deployed pravega image with these changes and have verified that reads and writes are working by running the benchmark with this pravega deployment.


docker/pravega/Dockerfile
```diff
@@ -9,14 +9,16 @@
 #
 FROM openjdk:11.0.8-jre-slim
 
-RUN apt-get update && apt-get install -y -q \
+RUN mkdir -p /usr/share/man/man1 && apt-get update && apt-get install -y -q \
```
ravisharda, line 12: Please bring `mkdir -p /usr/share/man/man1` to a line before rest of the CA certificates change so that the commands are collocated by functionality. 

Also, it is not intuitive why this change is needed. Could you please add a comment that explains it?

ravisharda (changes requested):

docker/pravega/Dockerfile
Prabhaker24: Done.

docker/pravega/Dockerfile
```diff
@@ -16,7 +16,11 @@ RUN apt-get update && apt-get install -y -q \
     jq \
     curl \
     net-tools \
-    iproute2  \
+    iproute2
+
+# Addidng ca-certificates to the base image
```
ravisharda, line 21: Nit: 
* `Addidng` -> `Adding`. 
* A more useful comment'd be one that explains why you are adding it. Such as "Adding Java system truststore 'cacerts', as it is missing from the Java distribution installed by the base image". 
* Also, explain why "mkdir -p /usr/share/man/man1" is needed.

ravisharda (changes requested):
LGTM.

docker/pravega/Dockerfile
RaulGracia: Agree with @ravisharda, it is unclear why we need `/usr/share/man/man1`.

RaulGracia (commented):
Please, justify the addition of `/usr/share/man/man1` and improve comments as suggested.

docker/pravega/Dockerfile
Prabhaker24: I have added comments explaining the inclusion of /usr/share/man/man1.

ravisharda (approved):
LGTM.

docker/pravega/Dockerfile
```diff
@@ -16,7 +16,16 @@ RUN apt-get update && apt-get install -y -q \
...
     net-tools \
-    iproute2  \
+    iproute2
+
+# Adding Java system truststore 'cacerts', as it is missing from the Java
+# distribution installed by the base image and one of the storage bindings(ECS)
+# depends on its presence.
+#
+# Without `mkdir -p /usr/share/man/man1` we hit the issue mentioned in:
+# https://stackoverflow.com/questions/58160597/docker-fails-with-sub-process-usr-bin-dpkg-returned-an-error-code-1
```
RaulGracia, line 26: I'm not sure that referring to SO is the best approach here. Could we just summarize the reason in 1 or 2 lines?

Prabhaker24: Removed the SO link and added comments.

RaulGracia (approved):

PR #5400, opened 2020-12-03
Issue 5384: Add Controller APIs to Create, Update and Delete Reader Groups

### LINKED ISSUE #5384: Add create, delete and update Reader Group APIs to Controller.
**Problem description**
Add create, delete and update Reader Group APIs to Controller. 
If a Reader Group is a "Subscriber" reader group, the create API along with storing Reader Group metadata will also add it as a  subscriber to Stream metadata. Similarly for delete and update APIs
These APIs will create, delete & update Reader Group information to all Streams that this ReaderGroup subscribes to, if it is a Subscribing reader group.
For non-subscriber reader groups these APIs will only result in changes to Reader Group metadata on Controller and will not cause any changes to Stream metadata.

**Problem location**
Controller.

pbelgundi:
**Change log description**  
This PR has the following code changes:
1. Remove the addSubscriber() and deleteSubscriber() APIs from Controller.
2. Add 4 new APIs to Controller:
   a. createReaderGroup(String scope, String rgName, RGConfig rgConfig)
   b. deleteReaderGroup(String scope, String rgName) and 
   c. updateReaderGroup(String scope, String rgName, RGConfig rgConfig)
   d. getReaderGroupConfig(String scope, String rgName)

3. If the ReaderGroup being created is a "Subscriber" reader group, then corresponding StreamMetadata would also be updated with subscriber entries in the create/update/delete API.

4. Create a new RGMetadataTable to keep track of ReaderGroups in a Pravega Cluster.

**Purpose of the change**  
Fixes #5384 

**How to verify it**  
All Unit and Integration Tests should pass


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
```diff
@@ -0,0 +1,92 @@
...
+                        return streamMetadataStore.getReaderGroupConfigRecord(scope, readerGroup, context, executor)
+                               .thenCompose(configRecord -> {
+                               log.info("Adding metadata to Streams");
+                               if (!ReaderGroupConfig.StreamDataRetention.values()[configRecord.getObject().getRetentionTypeOrdinal()]
+                                   .equals(ReaderGroupConfig.StreamDataRetention.NONE)) {
+                                   // update Stream metadata tables, if RG is a Subscriber
+                                   Iterator<String> streamIter = configRecord.getObject().getStartingStreamCuts().keySet().iterator();
+                                   return Futures.loop(() -> streamIter.hasNext(), () -> {
+                                          Stream stream = Stream.of(streamIter.next());
+                                          return streamMetadataStore.addSubscriber(stream.getScope(),
```
shiveshr, line 78: shouldnt add subscriber include generation number?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
```diff
@@ -0,0 +1,92 @@
...
+                                                       stream.getStreamName(), scopedRGName, null, executor);
+                                      }, executor);
+                                   }
+                               return CompletableFuture.completedFuture(null);
+                               }).thenCompose(v ->
+                                  Futures.toVoid(streamMetadataStore.createStream(scope, NameUtils.getStreamForReaderGroup(readerGroup),
```
shiveshr, line 84: this will only create the metadata for the rg stream.. it is not creating the segments.. 
i suggest you invoke createRg stream method on streamMetadataTasks to do this.

also since create for the RG stream can also come from the users, so there can be locking contentions for creation of the stream. but its not a problem per se except that this event's processing can be delayed until the stream gets created outside this context.
however, older clients will continue to create RG streams directly.. newer clients create rg streams using the new createRG api. 
so there shouldnt be a real contention of the sort.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
```diff
@@ -0,0 +1,79 @@
...
+        Preconditions.checkNotNull(streamMetaTasks);
+        Preconditions.checkNotNull(executor);
+        this.streamMetadataStore = streamMetaStore;
+        this.streamMetadataTasks = streamMetaTasks;
+        this.executor = executor;
+    }
+
+    @Override
+    public CompletableFuture<Void> execute(final DeleteReaderGroupEvent request) {
+      String scope = request.getScope();
```
shiveshr, line 51: do remember, our event processor has at least once processing guarantees. so the events can be replayed. 
which means the delete rg can be replayed. 
we do want to make its processing idempotent and not perform an unwanted delete in a subsequent replay of the event. 

so this delete `event` should include the "generation" or starting segmentsegmentnumber of the RG to be deleted. so that when it is processed we only delete the intended RG.. if it was already deleted previously and the current RG has a higher generation number then we ignore this event and do nothing.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
```diff
@@ -0,0 +1,79 @@
...
+               .thenCompose(configRecord -> {
+               if (!ReaderGroupConfig.StreamDataRetention.values()[configRecord.getObject().getRetentionTypeOrdinal()]
+                       .equals(ReaderGroupConfig.StreamDataRetention.NONE)) {
+               String scopedRGName = NameUtils.getScopedReaderGroupName(scope, readerGroup);
+               log.info("Inside Delete Event, found config");
+               // update Stream metadata tables, if RG is a Subscriber
+               Iterator<String> streamIter = configRecord.getObject().getStartingStreamCuts().keySet().iterator();
+               return Futures.loop(() -> streamIter.hasNext(), () -> {
+                       Stream stream = Stream.of(streamIter.next());
+                       return streamMetadataStore.deleteSubscriber(stream.getScope(),
```
shiveshr, line 66: again the delete subscriber should pass the generation number so that unsubscription is also conditional and we do not unnecessarily remove a subscription unintendedly.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
```diff
@@ -0,0 +1,79 @@
...
+                               stream.getStreamName(), scopedRGName, null, executor);
+                       }, executor);
+               }
+               return CompletableFuture.completedFuture(null);
+               })
+               .thenCompose(v -> {
+                   log.info("Inside Delete Event, deleting metadata and stream");
+                   return streamMetadataStore.deleteStream(scope,
```
shiveshr, line 74: again this is merely updating the metadata and not deleting the actual stream.. 
use streammetadatatasks.deletestream to actually delete the stream.
the caveat there is that this stream will be deleted asynchronously and this processing will wait on that to complete.. i dont particularly like that pattern. 

for that reason it may be preferrable to delete the RG stream outside the context of the event processing.. and then posting the event to perform unsubscriptions.

shiveshr (changes requested):

shared/controller-api/src/main/java/io/pravega/shared/controller/event/CreateReaderGroupEvent.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.concurrent.CompletableFuture;
+
+@Builder
+@Data
+@AllArgsConstructor
+public class CreateReaderGroupEvent implements ControllerEvent {
+    private static final long serialVersionUID = 1L;
+    private final String scopeName;
+    private final String rgName;
+    private final long requestId;
```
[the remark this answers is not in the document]
pbelgundi: For createReaderGroup, the generation is always=0 and this has ben set by the createReaderGroup task in RGMetadata.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
pbelgundi: There should not be a need to maintain generation number in Stream Metadata, if `updateSubscriberStreamCut()` takes a Reader Group `generation` as a parameter and checks the generation against the RGMetadata table before attempting to update the stream-cut in Stream metadata.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
pbelgundi: Ok. Yes sure I'll add createRGStream() to streamMetadataTasks and invoke that.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
pbelgundi: Will add this as discussed.

client/src/main/java/io/pravega/client/control/impl/Controller.java
```diff
@@ -11,13 +11,7 @@
...
 import io.pravega.client.admin.KeyValueTableInfo;
 import io.pravega.client.segment.impl.Segment;
-import io.pravega.client.stream.EventStreamReader;
-import io.pravega.client.stream.EventStreamWriter;
-import io.pravega.client.stream.Stream;
-import io.pravega.client.stream.StreamConfiguration;
-import io.pravega.client.stream.StreamCut;
-import io.pravega.client.stream.Transaction;
-import io.pravega.client.stream.TxnFailedException;
+import io.pravega.client.stream.*;
```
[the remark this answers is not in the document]
pbelgundi: sure

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
pbelgundi: Fixed to invoke streammetadatatasks.deleteStream()

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
```diff
@@ -0,0 +1,104 @@
...
+                                   return Futures.loop(() -> streamIter.hasNext(), () -> {
+                                          Stream stream = Stream.of(streamIter.next());
+                                          return streamMetadataStore.addSubscriber(stream.getScope(),
+                                                       stream.getStreamName(), scopedRGName, configRecord.getObject().getGeneration(),
+                                                  null, executor);
+                                      }, executor);
+                                   }
+                               return CompletableFuture.completedFuture(null);
+                               }).thenCompose(v ->
+                                  streamMetadataTasks.createRGStream(scope, NameUtils.getStreamForReaderGroup(readerGroup),
```
shiveshr, line 84: i would prefer if we moved this call out of event processor.. the problem is it can lead to deadlocks.. one event processing waiting on another asynchronous processing and there may not be sufficient processing slots available to process the new one as the current request holds onto a processing slot. 
that is why i prefer that we create the RG stream and then submit the createRG event to be processed which simply manages the subscriptions. 

PS: i know that presently create stream is performed using task framework. 

an alternative here is to call `createStreamBody` however, that does not guarantee the mutual exclusion.

shiveshr: also I am ok if you open a separate github issue for this and address it immediately after we close this PR.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
```diff
@@ -0,0 +1,84 @@
...
+                              Iterator<String> streamIter = configRecord.getObject().getStartingStreamCuts().keySet().iterator();
+                              return Futures.loop(() -> streamIter.hasNext(), () -> {
+                                     Stream stream = Stream.of(streamIter.next());
+                                     return streamMetadataStore.deleteSubscriber(stream.getScope(),
+                                            stream.getStreamName(), scopedRGName, null, executor);
+                                 }, executor);
+                              }
+                              return CompletableFuture.completedFuture(null);
+                          })
+                          .thenCompose(v -> streamMetadataTasks.deleteStream(scope,
```
shiveshr, line 78: same comment as with create RG... again we can handle this in a separate PR.. or if we move RG creation to a separate event processor in future then it doesnt contest for the processing slots to be available (remember: each instantiation of concurrent event processor processes at most 1000 events concurrently per controller instance)..

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateReaderGroupTask.java
```diff
@@ -0,0 +1,94 @@
...
+                            return Futures.loop(() -> streamIter.hasNext(), () -> {
+                                          Stream stream = Stream.of(streamIter.next());
+                                          return streamMetadataStore.addSubscriber(stream.getScope(),
+                                                       stream.getStreamName(), scopedRGName, rgConfigRecord.getObject().getGeneration(),
+                                                  null, executor);
+                                      }, executor)
+                                    .thenCompose(v -> Futures.loop(() -> removeStreamsIter.hasNext(), () -> {
+                                        Stream stream = Stream.of(removeStreamsIter.next());
+                                        return streamMetadataStore.deleteSubscriber(stream.getScope(),
+                                                stream.getStreamName(), scopedRGName, null, executor);
```
shiveshr, line 86: shouldnt generation be included in deleteSubscriber too? we dont want to be deleting an unintended subscriber just because this event got replayed.

shared/controller-api/src/main/java/io/pravega/shared/controller/event/UpdateReaderGroupEvent.java
```diff
@@ -0,0 +1,87 @@
...
+@Builder
+@Data
+@AllArgsConstructor
+public class UpdateReaderGroupEvent implements ControllerEvent {
+    private static final long serialVersionUID = 1L;
+    private final String scope;
+    private final String rgName;
+    private final long requestId;
+    private UUID readerGroupId;
+    private ImmutableSet<String> removeStreams;
```
shiveshr, line 36: doesnt it also need "add streams" ?
updateRG can have new streams added and some old streams removed.

shiveshr (changes requested):

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateReaderGroupTask.java
pbelgundi: The UUID check in deleteRGEvent should take care of that, but there is no harm in doing the check again in deleteSubscriber, so will add it here.

shared/controller-api/src/main/java/io/pravega/shared/controller/event/UpdateReaderGroupEvent.java
pbelgundi: addSubscriber() is invoked on all Streams in new config.
This takes care of adding (new) streams and updating generation for existing Streams ( present in both old and new config) 
So only streams to be removed need to be passed to the event.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -106,7 +106,9 @@ public void updateRetentionStreamCut(Map<Stream, StreamCut> streamCuts) {
         if (synchronizer.getState().getConfig().getRetentionType()
                 .equals(ReaderGroupConfig.StreamDataRetention.MANUAL_RELEASE_AT_USER_STREAMCUT)) {
             streamCuts.forEach((stream, cut) ->
-                    getThrowingException(controller.updateSubscriberStreamCut(stream.getScope(), stream.getStreamName(), groupName, cut)));
+                    // TODO: replace 0L with getConfig().generation
```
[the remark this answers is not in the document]
pbelgundi: Ok.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CreateReaderGroupTask.java
pbelgundi: Created https://github.com/pravega/pravega/issues/5438

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -481,8 +481,9 @@ void updateTruncationStreamCutIfNeeded() {
             Optional<Map<Stream, Map<Segment, Long>>> cuts = state.getPositionsForLastCompletedCheckpoint();
             cuts.orElseThrow(() -> new CheckpointFailedException("Could not get positions for last checkpoint."))
                  .entrySet().forEach(entry ->
+                        //TODO: replace 0L with rgconfig.generation
```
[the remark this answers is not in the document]
pbelgundi: ok

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
pbelgundi: https://github.com/pravega/pravega/issues/5438

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateReaderGroupTask.java
pbelgundi: fixed

shared/controller-api/src/main/java/io/pravega/shared/controller/event/UpdateReaderGroupEvent.java
shiveshr: again, shouldnt we also include the generation here.. if the event is replayed we should not be unsubscribing from a stream that was added as part of a subsequent message.

shiveshr (changes requested):

shared/controller-api/src/main/java/io/pravega/shared/controller/event/UpdateReaderGroupEvent.java
pbelgundi: Fixed. Added generation check as suggested.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -1588,9 +1519,169 @@ private void closeChannel() {
...
+            }
+        }).whenComplete((x, e) -> {
+            if (e != null) {
+                log.warn(requestId, "createReaderGroup {}/{} failed: ", scope, rgName, e);
+            }
+            LoggerHelpers.traceLeave(log, "createReaderGroup", traceId, rgConfig, requestId);
+        });
+    }
+
+    public CompletableFuture<Boolean> updateReaderGroup(String scope, String rgName, final ReaderGroupConfig rgConfig) {
```
shiveshr, line 1566: please open an issue to return generation back from controller.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -210,7 +256,217 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+    private CompletableFuture<Boolean> isRGDeleted(String scope, String rgName) {
+        return streamMetadataStore.checkReaderGroupExists(scope, rgName).thenApply(exists -> !exists);
+    }
+
+    private CompletableFuture<Void> startReaderGroupDelete(final String scope, final String rgName, VersionedMetadata<ReaderGroupState> currentState,
+                                                           RGOperationContext context) {
+        final String rgStreamName = NameUtils.getStreamForReaderGroup(rgName);
+        return Futures.toVoid(streamMetadataStore.updateReaderGroupVersionedState(scope, rgName,
+           ReaderGroupState.DELETING, currentState, context, executor))
+           .thenCompose(v -> sealStream(scope, rgStreamName, null))
```
shiveshr, line 461: the guarantee of eventhelper.addIndexAndSubmitTask is for one update. 
a failing over of this will simply post the event. it doesnt have the inputs for the future. 
so if you actually do have intent to do multiple updates, they should be done as part of event processing and not here. 
here we should just to do the fist step, start the operation and then post the event for it which will perform the whole execution if the processing has been started.

here you are doing  `updateReaderGroupVersionedState` followed by `sealStream` which in itself does a bunch of things. 
i suggest you move seal stream inside the event processing. but that is a different beast all together. but we have a new issue to take one async processing waiting on another out of the workflow.. we will handle that later. for now, simply move this there.

shiveshr (changes requested):

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: sure. I'll move seal stream inside eventprocessing

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
pbelgundi: Filed: https://github.com/pravega/pravega/issues/5447

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -210,7 +256,210 @@ public void initializeStreamWriters(final EventStreamClientFactory clientFactory
...
+                          log.warn("Reader group update failed as another update is in progress.");
+                          return CompletableFuture.completedFuture(UpdateReaderGroupStatus.Status.FAILURE);
+                        }
+                      });
+               });
+        }, e -> Exceptions.unwrap(e) instanceof RetryableException, READER_GROUP_OPERATION_MAX_RETRIES, executor);
+    }
+
+    private ImmutableSet<String> getStreamsToBeUnsubscribed(final Set<String> currentConfigStreams, final Set<String> newConfigStreams) {
+        return ImmutableSet.copyOf(currentConfigStreams.stream()
```
shiveshr, line 413: this is suboptimal as copyOf will make another pass.. better idea is to create an ImmutableSetBuilder.
then do foreach on each element and those that dont satisfy the filter predicate, add them to the builder.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
```diff
@@ -0,0 +1,85 @@
...
+                                     return streamMetadataStore.deleteSubscriber(stream.getScope(),
+                                            stream.getStreamName(), scopedRGName, configRecord.getObject().getGeneration(), null, executor);
+                                 }, executor);
+                              }
+                              return CompletableFuture.completedFuture(null);
+                          })
+                          .thenCompose(v -> streamMetadataTasks.sealStream(scope, NameUtils.getStreamForReaderGroup(readerGroup), null))
+                                  .thenCompose(v -> streamMetadataTasks.deleteStream(scope,
+                                  NameUtils.getStreamForReaderGroup(readerGroup), null)
+                                  .thenCompose(v1 -> streamMetadataStore.deleteReaderGroup(scope, readerGroup, context, executor)));
```
shiveshr, line 81: should we pass the rg UUID to delete RG method and to a conditonal delete just to be extra safe? i know you check the rg id before processing the event. and the create is also linearized on this processor so they do not race per se. but still a conditional delete may be a safer thing to do. perhaps open a github issue to address this. thanks

shiveshr (approved):
LGTM except two minor comments

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/DeleteReaderGroupTask.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: fixed

PR #5402, opened 2020-12-04
Issue 5401: Fixing resource leaks in tests

### LINKED ISSUE #5401: Fix resource leaks (in tests)
**Problem description**
1. Executors that are not shut down
2. Metrics providers that are started but never shut down.
3. Others.

**Problem location**
Codebase.

1. Ensure that all executors are created using `ExecutorServiceHelpers`. Define a leak detection level in this class. If enabled, record where executor is created from and use the `finalize` method to track leaks.
2. Search & destroy.
3. TBD

andreipaduroiu:
**Change log description**  
- Modified ExecutorServiceHelpers to accept a mode where it can track ExecutorServices it creates that are reclaimed by the GC prior to being shut down.
- Cleaning up all ExecutorService instances that we create, which were found using the above method.
- Added `@Cleanup` to several places in tests where they should have been before.

**Purpose of the change**  
Fixes #5401.

**What the code does**  
- `ExecutorServiceHelpers` delegates new thread pool creation to `ExecutorServiceFactory`
- `ExecutorServiceFactory`
    - Defines a "thread leak detection level" which is automatically picked from System Properties.
        - **None:** No detection. All thread pools are the ones that come from the JDK
        - **Light:** Returns a thread pool derived from the JDK one that overrides the `finalize` method. If the thread pool is still running when this method is invoked (i.e., GC is culling it), it logs a WARN message with a stack trace indicating where the thread pool was created from.
        - **Aggressive:** Same as Light, but additionally shuts down the VM when a thread leak is detected (see below for more details)
- `build.gradle`
    - Setting `ThreadPoolLeakDetectionLevel` to `Aggressive` for all builds (and only for builds). This will ensure that future changes will not reintroduce new such leaks.
- More details and thoughts:
    - The point of **Aggressive** shutting down the VM is to stop (fail) the build if such a leak is detected in unit or integration tests. Since the `finalize` method is run on the GC threads, it is impossible to raise an `AssertionError` that would bubble up to some test (the test that caused it may have even finished long before). To prevent a simple WARN from being lost (and ignored), this option stops the VM and the developer will need to inspect the last few lines of logs to figure out the source of the leak.
        - This method has been used in this PR to hunt down all leaks reported by the build.
    - The thread leak detection level is turned off by default, so production code will not be affected. No code path in this case can even come near the `System.exit(99)` code that halts the VM.
    - In case we want to check for such leaks in production, we can enable the **Light** detection and then watch out for WARNS. Again, there is no risk in this case that the VM will halt (unit tests have been written to ensure this), and this method is non-invasive, so it may be preferred in various functional, system and longevity tests.

**How to verify it**  
Build must pass.


derekm:
@andreipaduroiu -- GitHub Actions build now succeeds with these fixes. Thanks!

client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
```diff
@@ -208,7 +207,7 @@ public StreamInfo getStreamInfo(String scopeName, String streamName) {
     @Override
     public void close() {
         if (this.controller != null) {
-            this.controller.close();
+            Callbacks.invokeSafely(this.controller::close, ex -> log.error("Unable to close Controller client.", ex));
```
derekm, line 210: Should all callers of `close()` use this `invokeSafely()` helper? Such as `KeyValueTableManagerImpl` above?

client/src/test/java/io/pravega/client/stream/impl/ClientFactoryTest.java
```diff
@@ -105,8 +106,10 @@ public void testEventWriterSealedStream() {
                 .thenReturn(CompletableFuture.completedFuture(currentSegments));
 
         EventWriterConfig writerConfig = EventWriterConfig.builder().build();
+        @Cleanup
         EventStreamWriter<String> writer = clientFactory.createEventWriter(stream, new JavaSerializer<String>(), writerConfig);
         assertEquals(writerConfig, writer.getConfig());
+        writer.close();
```
derekm, line 112: Is this `close()` redundant to the `@Cleanup` annotation a few lines above?

client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
andreipaduroiu: Normally `close` shouldn't throw anything. But Java's `AutoCloseable.close` defines a `throws Exception` to handle the general case.

What I found is that the Controller and Controller client have some issues when closing. They use numerous sub-components and when they shut down, some may hang forever or throw an exception during shutdown. This prevents whatever else is supposed to close next to never be closed and stick in memory forever. 

The real fix is to figure out what is hanging or throwing exceptions and dealing with them appropriately. However the Controller is not my area of expertise, so I decided to leave that for another time (otherwise I would not be able to finish this by January).

client/src/test/java/io/pravega/client/stream/impl/ClientFactoryTest.java
andreipaduroiu: If the test fails somwhere before this line, the writer will never be cleaned up. So whatever connections and thread pools it opens (yes, it does that), will stick around. Cleanup is essentially rewriting the code as a `try-finally` to make sure that whatever we have declared as needing cleanup will be closed when the code exits

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceFactory.java
```diff
@@ -0,0 +1,292 @@
...
+
+    //region Constructor
+
+    /**
+     * Creates a new instance of the {@link ExecutorServiceFactory} which extracts the {@link ThreadLeakDetectionLevel}
+     * from system properties and halts the VM if a leak is detected and {@link ThreadLeakDetectionLevel} is set to
+     * {@link ThreadLeakDetectionLevel#Aggressive}.
+     */
+    ExecutorServiceFactory() {
+        this(getDetectionLevel(), () -> System.exit(99));
```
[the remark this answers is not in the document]
andreipaduroiu: At line 256, we print out a `log.error`. Prior to that we dump the stack trace of that exception that we logged. That should be sufficient info for someone to debug.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceFactory.java
```diff
@@ -0,0 +1,292 @@
...
+        }
+    }
+
+    private class LeakDetectorThreadPoolExecutor extends ThreadPoolExecutor {
+        private final Exception stackTraceEx;
+
+        LeakDetectorThreadPoolExecutor(int corePoolSize, int maximumPoolSize, long keepAliveTime, TimeUnit unit, BlockingQueue<Runnable> workQueue,
+                                       ThreadFactory threadFactory, RejectedExecutionHandler handler) {
+            super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue, threadFactory, handler);
+            this.stackTraceEx = new Exception();
```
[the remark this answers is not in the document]
andreipaduroiu: This class is only instantiated if level != `None`. If it's set to `None`, the code path above will create the base (JDK) thread pool so this code is not even executed.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceFactory.java
```diff
@@ -0,0 +1,292 @@
...
+
+        if (!e.isShutdown() || !e.isTerminated()) {
+            log.warn("THREAD POOL LEAK: {} (ShutDown={}, Terminated={}) finalized without being properly shut down.",
+                    e.getThreadFactory(), e.isShutdown(), e.isTerminated(), stackTraceEx);
+            if (this.detectionLevel == ThreadLeakDetectionLevel.Aggressive) {
+                // Not pretty, but outputting this stack trace on System.err helps with those unit tests that turned off
+                // logging.
+                stackTraceEx.printStackTrace(System.err);
+                log.error("THREAD POOL LEAK DETECTED WITH LEVEL SET TO {}. SHUTTING DOWN.", ThreadLeakDetectionLevel.Aggressive);
+                this.onLeakDetected.run();
```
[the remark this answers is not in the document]
andreipaduroiu: I need this for unit testing. I can't verify `System.exit` in unit tests since my JVM would quit and not help much.

common/src/test/java/io/pravega/common/concurrent/ExecutorServiceFactoryTests.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Test;
+
+/**
+ * Unit tests for the {@link ExecutorServiceFactory} class.
+ */
+public class ExecutorServiceFactoryTests {
+    @After
+    public void tearDown() {
+        // Invoke GC and finalization - this will make sure that we catch any leftover thread pools
+        System.gc();
```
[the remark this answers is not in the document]
andreipaduroiu: According to the [Javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/System.html#gc()):

```
Runs the garbage collector. Calling the gc method suggests that the Java Virtual Machine expend effort toward recycling unused objects in order to make the memory they currently occupy available for quick reuse. When control returns from the method call, the Java Virtual Machine has made a best effort to reclaim space from all discarded objects. 
```

tkaitchuck (approved):

andreipaduroiu:
System tests passed with this change.

test/integration/src/test/java/io/pravega/test/integration/controller/server/EventProcessorTest.java
```diff
@@ -94,8 +93,12 @@
     private ServiceBuilder serviceBuilder;
```
shrids, line 93: serviceBuilder needs to be closed in the @After method.

RaulGracia (approved):
LGTM, thanks for the effort on improving the build stability.

test/integration/src/test/java/io/pravega/test/integration/controller/server/EventProcessorTest.java
andreipaduroiu: Good one. Fixed.

client/src/main/java/io/pravega/client/admin/impl/KeyValueTableManagerImpl.java
```diff
@@ -80,17 +80,17 @@ public KeyValueTableManagerImpl(@NonNull ClientConfig clientConfig) {
...
-
             if (this.connectionFactory != null) {
                 this.connectionFactory.close();
             }
 
             if (this.executor != null) {
                 ExecutorServiceHelpers.shutdown(this.executor);
             }
+
+            if (this.controller != null) {
```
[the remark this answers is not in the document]
andreipaduroiu: In some cases, some of these methods may throw exceptions or hang around forever. While I did try to fix most of them, fixing them was out of scope for this effort, so I did my best to try to close executors before the problematic components.

shrids (approved):

shiveshr (commented):
LGTM

PR #5406, opened 2020-12-07
Issue 5403: (SegmentStore) Reducing number of Storage reads from SegmentAttributeBTreeIndex

### LINKED ISSUE #5401: Fix resource leaks (in tests)

### LINKED ISSUE #5403: SegmentAttributeBTreeIndex should piggyback read requests on existing ones if possible
**Problem description**
If multiple page read requests come at the same time, subsequent ones should piggyback on the first one if needing to read from Storage.

**Problem location**
SegmentAttributeBTreeIndex.java

andreipaduroiu:
**Change log description**  
Piggybacking onto existing Storage reads for the same BTree page. 

**Purpose of the change**  
Fixes #5403.

**What the code does**  
- In `SegmentAttributeBTreeIndex`, any BTreePage cache misses results in a direct read from Storage. If this is case, concurrent requests for the same key (or keys that require the same page(s)) would likely end up issuing identical, concurrent Storage read requests.
- This change:
    - Records all (new) read requests and indexes them by offset (which is the BTreePage unique identifier).
    - If another request comes in for the same page (offset) while the first one is running, the second one will not issue a Storage read; instead it will return the completion from the first one (i.e., piggyback onto that one).
    - If the original request fails, so will any that have been piggybacked.
    - When the original request completes (successfully or not), it is unregistered from the pending read collection so any subsequent read may go back to read from Storage (if not a cache hit).

**How to verify it**  
Build must pass.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
```diff
@@ -502,15 +515,63 @@ private long deserializeValue(ByteArraySegment value) {
...
+                    }
+                });
+                this.pendingReads.put(offset, pr);
+            } else if (pr.length < length) {
+                // Somehow the previous request wanted to read less than us. This shouldn't be the case, yet it is
+                // a situation we should handle. In his case, we will not be recording the PendingRead.
+                pr = new PendingRead(offset, length);
+            } else {
+                // Piggyback on the existing read.
+                return pr.completion;
```
sachin-j-joshi, line 556: What happens when page is evicted ?
I think here when page is evicted from cache we need to remove that offset from pendingReads , otherwise next time we need to reload the page again we'll have a stale completed future from the last time and we won't really read anything - leaving cache with missing data.

sachin-j-joshi (changes requested):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
sachin-j-joshi: Oops my bad. 
Never mind - I now see it is removed on line 546. Thanks

sachin-j-joshi (approved):
LGTM

PR #5412, opened 2020-12-09
Issue 5390: SLTS - Improve logging. Shortcut findChunk for reading last chunk.

### LINKED ISSUE #5411: SLTS - Small improvements and better logs in ReadOperator and WriteOperator
**Problem description**
List of small problems in ReadOperator and WriteOperator that need to be fixed
- Incorrectly reported scan counting in ReadOperator
- Incorrect/incomplete log messages.
- Unnecessary chunk lookup when data is in last chunk.

**Problem location**
ReadOperator and WriteOperator

- Fix incorrectly reported scan counting and log messages in ReadOperator
- Fix incorrect/incomplete log messages.
- Skip chunk lookup when data is in last chunk.

sachin-j-joshi:

**Change log description**  
List of small improvements  in ReadOperator and WriteOperator.
- Fix incorrectly reported scan counting and log messages in ReadOperator
- Fix incorrect/incomplete log messages.
- Skip chunk lookup when data is in last chunk.

**Purpose of the change**  
Fixes #5411 

**What the code does**  
- Fix incorrectly reported scan counting and log messages in ReadOperator
- Fix incorrect/incomplete log messages.
- Skip chunk lookup when data is in last chunk.

**How to verify it**  
build should pass.

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
@@ -69,8 +69,8 @@ class ReadOperation implements Callable<CompletableFuture<Integer>> {
     public CompletableFuture<Integer> call() {
         // Validate preconditions.
         checkPreconditions();
-        log.debug("{} read - started op={}, segment={}, offset={}, bytesRead={}.",
-                chunkedSegmentStorage.getLogPrefix(), System.identityHashCode(this), handle.getSegmentName(), offset, totalBytesRead);
+        log.debug("{} read - started op={}, segment={}, offset={}, length={}.",
+                chunkedSegmentStorage.getLogPrefix(), System.identityHashCode(this), handle.getSegmentName(), offset, length);
         val streamSegmentName = handle.getSegmentName();
         return ChunkedSegmentStorage.tryWith(chunkedSegmentStorage.getMetadataStore().beginTransaction(true, streamSegmentName),
                 txn -> txn.get(streamSegmentName)
@@ -90,7 +90,7 @@ class ReadOperation implements Callable<CompletableFuture<Integer>> {
                                         return readData(txn);
                                     }, chunkedSegmentStorage.getExecutor())
                                     .exceptionally(ex -> {
-                                        log.debug("{} read - started op={}, segment={}, offset={}, bytesRead={}.",
+                                        log.debug("{} read - exception op={}, segment={}, offset={}, bytesRead={}.",
                                                 chunkedSegmentStorage.getLogPrefix(), System.identityHashCode(this), handle.getSegmentName(), offset, totalBytesRead);
                                         if (ex instanceof CompletionException) {
                                             throw (CompletionException) ex;
@@ -123,6 +123,7 @@ class ReadOperation implements Callable<CompletableFuture<Integer>> {
         return Futures.loop(
                 () -> bytesRemaining.get() > 0 && null != currentChunkName,
                 () -> {
+                    Preconditions.checkState(null != chunkToReadFrom, "chunkToReadFrom is null");
                     bytesToRead = Math.min(bytesRemaining.get(), Math.toIntExact(chunkToReadFrom.getLength() - (currentOffset.get() - startOffsetForCurrentChunk.get())));
                     if (currentOffset.get() >= startOffsetForCurrentChunk.get() + chunkToReadFrom.getLength()) {
                         // The current chunk is over. Move to the next one.
@@ -132,7 +133,9 @@ class ReadOperation implements Callable<CompletableFuture<Integer>> {
                             return txn.get(currentChunkName)
                                     .thenAcceptAsync(storageMetadata -> {
                                         chunkToReadFrom = (ChunkMetadata) storageMetadata;
-                                        log.debug("{} read - reading from next chunk - segment={}, chunk={}", chunkedSegmentStorage.getLogPrefix(), handle.getSegmentName(), chunkToReadFrom);
+                                        Preconditions.checkState(null != chunkToReadFrom, "chunkToReadFrom is null");
+                                        log.debug("{} read - reading from next chunk - op={}, segment={}, chunk={}", chunkedSegmentStorage.getLogPrefix(),
+                                                System.identityHashCode(this), handle.getSegmentName(), chunkToReadFrom);
                                     }, chunkedSegmentStorage.getExecutor());
                         }
                     } else {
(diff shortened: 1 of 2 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/WriteOperation.java
```diff
@@ -265,8 +265,8 @@ private void collectGarbage() {
                     didSegmentLayoutChange = true;
                     chunksAddedCount.incrementAndGet();
 
-                    log.debug("{} write - New chunk added - segment={}, chunk={}, offset={}.",
-                            chunkedSegmentStorage.getLogPrefix(), handle.getSegmentName(), newChunkName, segmentMetadata.getLength());
+                    log.debug("{} write - New chunk added - op={}, segment={}, chunk={}, offset={}.",
+                            chunkedSegmentStorage.getLogPrefix(), System.identityHashCode(this), handle.getSegmentName(), newChunkName, segmentMetadata.getLength());
```
andreipaduroiu, line 269: how is this helpful to identify it? Also, what's wrong with `this.hashcode()`?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -202,14 +211,15 @@ private void logEnd() {
...
-                        }, chunkedSegmentStorage.getExecutor())
-                        .thenAcceptAsync(v -> {
-                            val elapsed = readIndexTimer.getElapsed();
-                            SLTS_READ_INDEX_SCAN_LATENCY.reportSuccessEvent(elapsed);
-                            log.debug("{} read - chunk lookup - segment={}, offset={}, scanned={}, latency={}.",
-                                    chunkedSegmentStorage.getLogPrefix(), handle.getSegmentName(), offset, cntScanned.get(), elapsed.toMillis());
                         }, chunkedSegmentStorage.getExecutor()),
-                chunkedSegmentStorage.getExecutor());
+                chunkedSegmentStorage.getExecutor())
+                .thenAcceptAsync(v -> {
```
andreipaduroiu, line 216: Why async? Those operations should be quick. No need to add a new task in the executor queue.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/WriteOperation.java
sachin-j-joshi: I needed some unique id for for each request.
Each operation invocation is a new object and that is good enough to for debug purpose.

System.identityHashCode(this) is more "stable", can't be overridden by implementation and more likely to be unique than 
HashCode.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: The java documentation is ambiguous about it whether it is guaranteed to be run on same thread.
It seems it can run on this thread pool or thread pool of the caller. (in this case that would be possibly core thread).
Hence to be careful not to hijack core threads I choose to explicitly provide thread pool I want this to be run on. (which is same thread pool anyway).

Here I'm just maintaining consistency with rest of the code in this file by using Async. 
I do plan to use non async overload if it can be guaranteed that FJ or other thread pools don't get used accidently or a more likely case when storage thread pool is super busy.

sachin-j-joshi:
@andreipaduroiu Are you okay with the changes?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -90,7 +90,7 @@
                                         return readData(txn);
                                     }, chunkedSegmentStorage.getExecutor())
                                     .exceptionally(ex -> {
-                                        log.debug("{} read - started op={}, segment={}, offset={}, bytesRead={}.",
+                                        log.debug("{} read - exception op={}, segment={}, offset={}, bytesRead={}.",
```
RaulGracia, line 93: If this is an exception, why it is in `debug` level? In general, we want to keep track of relevant exceptions in production systems, which requires this to be at least `info` level, if not `warn` or `error`. As SLTS is a new piece of code, let's keep high the log level of exceptions so we can debug problems quickly without needed to repeat experiments with `debug` level logs.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -132,7 +133,9 @@ private void logEnd() {
                             return txn.get(currentChunkName)
                                     .thenAcceptAsync(storageMetadata -> {
                                         chunkToReadFrom = (ChunkMetadata) storageMetadata;
-                                        log.debug("{} read - reading from next chunk - segment={}, chunk={}", chunkedSegmentStorage.getLogPrefix(), handle.getSegmentName(), chunkToReadFrom);
+                                        Preconditions.checkState(null != chunkToReadFrom, "chunkToReadFrom is null");
```
RaulGracia, line 136: Maybe not related to the PR, but the way `chunkToReadFrom` (and other variables) is set and checked here and there looks hard to track in case we face a race conditions. Even tough a variable is declared as `volatile`, the point is that if multiple methods (perhaps executed by different threads) are setting and reading this variable throughout the class, it increases considerably the complexity of debugging it in case of problems. Perhaps this was the only way, I don't know, but maybe in the future we can reconsider this pattern to see if there is an alternative that simplifies the code.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: We do throw this exception which the caller will log appropriately.
This is just additional debug information that exception does not contain.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: Generic answer to your question - 
Unfortunately java does not have [Async/Await syntax suger](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/) that generates all the future based  ([Task parallel library](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/task-parallel-library-tpl)) lambdas for your code.

So we have to do it by hand. Having done lots TPL based programming in the past, I was sort of reluctant to rewrite this code for async knowing it makes very difficult to maintain spaghetti code with lots of lambdas. 
However we absolutely do need it here.

The general pattern that I have here for thread safety is as follows
- Each major function in linear non-async code becomes a _method object_. Each invocation is a separate object. So separate invocations share nothing.
- All members **must** be either **final** or **volatile**
- If you need to use any arithmetic then it is better to use Atomic variables (which actually just volatile variables with additional safety added for atomic addition/subtraction
- The lamdas are executed in series by java sdk (i.e CompletableFuture framework) one lambda at a time.
- The JVM [memory model](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html) guarantees that reads and writes to volatile variable are thread safe.

Writing all this async code by hand and yet keeping it understandable is not easy. But hopefully code gets better, faster and also cleaner as go.

RaulGracia (approved):

PR #5413, opened 2020-12-09
Issue 4903: More robust garbage collection

### LINKED ISSUE #4903: PDP-34 Simplified-LTS : Implement background processing for garbage collection of deleted chunks.
**Problem description**
In current Simplified-LTS implementation attempt is made to delete garbage chunk immediately, however if this fails then warning is logged but that failure is not reattempted in future. This leaves garbage behind. 

**Problem location**
ChunkManager

- Only mark chunks as deleted in metadata.
- Record list of failed attempt to delete chunks in metadata.
- Start a background thread, and periodically reattempt the chunk deletion on this background thread.

sachin-j-joshi:
**Change log description**  
More robust garbage collection with delayed delete action.

**Purpose of the change**  
Fixes #4903 

**What the code does**  
- Only mark chunks as deleted in metadata instead of deleting them immediately.
- Delayed actual deletion .
- Record list of failed attempt to delete chunks in metadata.
- Start a background loop, and periodically perform chunk deletion on this background thread.

**How to verify it**  
All tests should pass.


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -662,14 +634,21 @@ public boolean supportsTruncation() {
 
     @Override
     public void close() {
+        close("metadataStore", this.metadataStore);
+        close("garbageCollector", this.garbageCollector);
+        this.closed.set(true);
+    }
+
+    private void close(String message, AutoCloseable toClose) {
```
andreipaduroiu, line 642: This sounds like a great utility in common. Mind if you add it there?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -662,14 +634,21 @@ public boolean supportsTruncation() {
...
         try {
-            if (null != this.metadataStore) {
-                this.metadataStore.close();
+            log.info("{} Closing {}", logPrefix, message);
+            if (null != toClose) {
+                toClose.close();
             }
+            log.info("{} Closed {}", logPrefix, message);
```
andreipaduroiu, line 648: Let's not pollute the logs with so many messages. Make the "Closing" one `debug`.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -178,7 +174,9 @@ public ChunkedSegmentStorage(int containerId, ChunkStorage chunkStorage, ChunkMe
 
         // Now bootstrap
         log.debug("{} STORAGE BOOT: Started.", logPrefix);
-        return this.systemJournal.bootstrap(epoch).thenRunAsync(() -> log.debug("{} STORAGE BOOT: Ended.", logPrefix), executor);
+        return this.systemJournal.bootstrap(epoch)
+                .thenRunAsync(() -> garbageCollector.initialize())
+                .thenRunAsync(() -> log.debug("{} STORAGE BOOT: Ended.", logPrefix), executor);
```
andreipaduroiu, line 179: Why async on this tiny one? If you don't have debug enabled, this won't even do anything?

Please try to reduce the number of `Async` calls, especially if the code you execute is very simple.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
```diff
@@ -124,6 +128,19 @@
     @Getter
     final private int lateWarningThresholdInMillis;
 
+    /**
+     * Minimum delay in seconds between when garbage chunks are marked for deletion and actually deleted.
+     */
+    @Getter
+    final private int garbageCollectionFrequencyInSeconds;
```
andreipaduroiu, line 135: Why isn't this a `Duration`? Then you don't have to worry about converting to seconds, millis, or whatever

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
```diff
@@ -142,6 +159,8 @@
         long defaultMaxLength = properties.getLong(DEFAULT_ROLLOVER_SIZE);
         this.defaultRollingPolicy = new SegmentRollingPolicy(defaultMaxLength);
         this.lateWarningThresholdInMillis = properties.getInt(SELF_CHECK_LATE_WARNING_THRESHOLD);
+        this.garbageCollectionFrequencyInSeconds = properties.getInt(GARBAGE_COLLECTION_DELAY);
```
andreipaduroiu, line 162: Validate these settings here.
Make sure they're not negative (or too big).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Constructs a new instance.
+     *
+     * @param chunkedSegmentStorage Instance of {@link ChunkedSegmentStorage}.
+     * @param config Configuration to use.
+     */
+    public GarbageCollector(ChunkedSegmentStorage chunkedSegmentStorage, ChunkedSegmentStorageConfig config) {
+        this.chunkedSegmentStorage = chunkedSegmentStorage;
+        this.config = config;
+        this.delayInMillis = Duration.ofSeconds(config.getGarbageCollectionFrequencyInSeconds()).toMillis();
```
andreipaduroiu, line 58: See, this is exactly why keeping that property as a `Duration` is a good thing.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private CompletableFuture<Void> loopFuture;
+
+    /**
+     * Constructs a new instance.
+     *
+     * @param chunkedSegmentStorage Instance of {@link ChunkedSegmentStorage}.
+     * @param config Configuration to use.
+     */
+    public GarbageCollector(ChunkedSegmentStorage chunkedSegmentStorage, ChunkedSegmentStorageConfig config) {
```
andreipaduroiu, line 55: Null checks

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+        this.chunkedSegmentStorage = chunkedSegmentStorage;
+        this.config = config;
+        this.delayInMillis = Duration.ofSeconds(config.getGarbageCollectionFrequencyInSeconds()).toMillis();
+    }
+
+    /**
+     * Initializes this instance.
+     */
+    public void initialize() {
+        loopFuture = run();
```
andreipaduroiu, line 65: Make this class extend from `AbstractThreadPoolService` and follow that pattern. See `StorageWriter.java` for an example.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    public CompletableFuture<Void> run() {
+        return Futures.loop(
+                () -> !closed.get(),
+                () -> CompletableFuture.completedFuture(null)
```
andreipaduroiu, line 71: There is no need for this pattern here. Please get rid of `CompletableFuture.completedFuture().thenCompose`

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+                        .thenComposeAsync(v -> deleteGarbage(true, config.getGarbageCollectionConcurrency()),
+                                chunkedSegmentStorage.getExecutor())
+                        .handleAsync((v, ex) -> {
+                            if (null != ex) {
+                                log.error("{} Error during run {}.", chunkedSegmentStorage.getLogPrefix(), ex);
```
andreipaduroiu, line 76: This will double-log the exception. Remove the second `{}`.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+        suspended.set(value);
+    }
+
+    /**
+     * collect the garbage chunks.
+     *
+     * @param chunksToDelete List of chunks to delete.
+     */
+    CompletableFuture<Void> addToGarbage(Collection<String> chunksToDelete) {
+        val currentTime = System.currentTimeMillis();
```
andreipaduroiu, line 96: How will garbage collection work across failover? This time won't persist in that case.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+        chunksToDelete.forEach(chunkToDelete -> garbageChunks.put(chunkToDelete, currentTime));
+        return CompletableFuture.completedFuture(null);
```
andreipaduroiu, line 98: Why is this a `CompletableFuture`? Change this method to not return this.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Delete the garbage chunks.
```
andreipaduroiu, line 102: Properly document the behavior of this method if it's suspended, background ,etc.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * @param isBackground True if the caller is backgound task else False if called explicitly.
+     * @param maxItems Maximum number of items to delete at a time.
+     * @return CompletableFuture which is completed when garbage is deleted.
+     */
+    CompletableFuture<Void> deleteGarbage(boolean isBackground, int maxItems) {
+        if (suspended.get() && isBackground) {
+            return CompletableFuture.completedFuture(null);
```
andreipaduroiu, line 110: Log it. (Debug)

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+        val currentTime = System.currentTimeMillis();
```
andreipaduroiu, line 112: I don't know how you are planning to unit test this, but I would take a constructor that provides a `Supplier<Long> getTimeMillis` which I can use in my tests. Then your default constructor can pass `System::currentTimeMillis` to that arg.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+        val chunksToDelete = new ArrayList<String>();
+        int count = 0;
+        for (val entry : garbageChunks.entrySet()) {
```
andreipaduroiu, line 115: Why is `garbageChunks` a Map? Are you expecting to insert duplicates (that would smell like a bug).

I suggest you add a sorted list or something, sort by time, and then in this list you only need to loop through those items that need deletion (i.e, do not do an O(n) lookup every time).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+            if (canDelete(entry.getValue(), currentTime)) {
+                chunksToDelete.add(entry.getKey());
+                // Throttle how many chunks are deleted at once.
+                if (++count >= maxItems) {
```
andreipaduroiu, line 119: If you do a sorted list, you may even be able to compress this whole block into a `stream()` with limit and filter.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+                                                log.warn("{} deleteGarbage - Could not delete garbage chunk {}.", chunkedSegmentStorage.getLogPrefix(), chunkToDelete);
+                                            }
+                                            return null;
+                                        });
+                            } else {
+                                log.info("{} deleteGarbage - Chunk is not marked as garbage{}.", chunkedSegmentStorage.getLogPrefix(), chunkToDelete);
+                                return CompletableFuture.completedFuture(null);
+                            }
+                        })
+                    .whenCompleteAsync((v, ex) -> {
```
andreipaduroiu, line 156: Which executor are you running this on? Most likely it will go onto `ForkJoinPool.commonPool()` if you don't explicitly pass one.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,182 @@
...
+        garbageChunks.remove(chunkName);
+    }
+
+    boolean canDelete(long time, long currentTime) {
+        return delayInMillis < (currentTime - time);
+    }
+
+    @Override
+    public void close() throws Exception {
+        closed.set(true);
```
andreipaduroiu, line 179: In this close, you should probably wait for your `run` future to complete as well. The contract for `close` is that when it exits, any activities started by this instance will have ended.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/Constants.java
```diff
@@ -0,0 +1,35 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.segmentstore.storage.metadata;
+
+/**
+ * Class that defines various constants.
+ */
+public final class Constants {
```
andreipaduroiu, line 15: I would not be making these public. Please make sure that any `setFlag`/`getFlag` method is private to the class(es) they are defined in and you expose proper methods to set them. Nobody likes to deal with flags and masks.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageMockTests.java
```diff
@@ -326,13 +326,16 @@ public void testIOExceptionDuringWrite() throws Exception {
     public void testFileNotFoundExceptionDuringGarbageCollection() throws Exception {
         String testSegmentName = "test";
         SegmentRollingPolicy policy = new SegmentRollingPolicy(2); // Force rollover after every 2 byte.
-        val config = ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder().defaultRollingPolicy(policy).build();
+        val config = ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder()
+                .garbageCollectionFrequencyInSeconds(0)
+                .build();
 
         BaseMetadataStore spyMetadataStore = spy(new InMemoryMetadataStore(executorService()));
```
andreipaduroiu, line 333: Add `@Cleanup` to all the objects here that implement `AutoCloseable`

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageMockTests.java
```diff
@@ -345,20 +348,26 @@ public void testFileNotFoundExceptionDuringGarbageCollection() throws Exception
...
     }
 
     @Test
     public void testExceptionDuringGarbageCollection() throws Exception {
         String testSegmentName = "test";
         SegmentRollingPolicy policy = new SegmentRollingPolicy(2); // Force rollover after every 2 byte.
-        val config = ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder().defaultRollingPolicy(policy).build();
+        val config = ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder()
+                .defaultRollingPolicy(policy)
+                .garbageCollectionFrequencyInSeconds(0)
```
andreipaduroiu, line 362: Does this mean that your test takes 8s or more to run?

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
```diff
@@ -1785,15 +1788,20 @@ public void testTruncateWithFailover() throws Exception {
...
         newTestContext.chunkedSegmentStorage.truncate(hWrite, offset, null).get();
+        newTestContext.chunkedSegmentStorage.getGarbageCollector().setSuspended(true);
+        newTestContext.chunkedSegmentStorage.getGarbageCollector().deleteGarbage(false, 100).get();
         //checkDataRead(testSegmentName, testContext, offset, 0);
         TestUtils.checkSegmentBounds(newTestContext.metadataStore, testSegmentName, offset, offset);
 
         AssertExtensions.assertFutureThrows("openWrite() allowed after fencing",
                 oldTestCotext.chunkedSegmentStorage.openWrite(testSegmentName),
                 ex -> ex instanceof StorageNotPrimaryException);
+        /*
```
andreipaduroiu, line 1799: Commented out code. Delete or reinstate.

andreipaduroiu (dismissed):

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageMockTests.java
sachin-j-joshi: No. It is the other way round i want 0 delay.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
```diff
@@ -124,6 +135,32 @@
...
 
+    /**
+     * Minimum delay in seconds between when garbage chunks are marked for deletion and actually deleted.
+     */
+    @Getter
+    final private Duration garbageCollectionDelay;
+
+    /**
+     * Number of chunks deleted concurrently.
+     * This number should be small enough so that it does interfere foreground requests.
```
RaulGracia, line 146: Rather than just advising, maybe you could a test to verify that this is not larger than half of the storage pool size or something along these lines. Then, you can decide whether to throw in this case (so, do not allow large thread pools for garbage collection) or log it as a warning.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Implements simple garbage collector for cleaning up the deleted chunks.
+ * The garbage collector maintains a in memory queue of chunks to delete which is drained by a background task.
+ * This queue is populated by following
+ * 1. Various SLTS operations requesting deletes
+ * 2. Background task that scans all records to find undeleted chunks inside metadata (not yet implemented).
+ * 3. Background task that scans all LTS to find unaccounted chunks that are deemed garbage (not yet implemented).
```
RaulGracia, line 40: Please, create 2 Git issues to keep track of these problems in the future and add them to the comments.
Also, the "scans all records" part sounds scary in terms of efficiency, we need to think of a better approach to do that.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,263 @@
...
+     * collect the garbage chunks.
+     *
+     * @param chunksToDelete List of chunks to delete.
+     */
+    void addToGarbage(Collection<String> chunksToDelete) {
+        val currentTime = System.currentTimeMillis();
+
+        if (queueSize.get() < config.getGarbageCollectionMaxQueueSize()) {
+            chunksToDelete.forEach(chunkToDelete -> garbageChunks.add(new GarbageChunkInfo(chunkToDelete, currentTime)));
+            queueSize.incrementAndGet();
```
RaulGracia, line 145: We discussed that this is a bug, as you are not incrementing the size of the queue inside the loop, right?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,263 @@
...
+        } else {
+            for (val chunkToDelete : chunksToDelete) {
+                log.warn("{} deleteGarbage - Queue full. Could not delete garbage. chunk {}.", chunkedSegmentStorage.getLogPrefix(), chunkToDelete);
```
RaulGracia, line 148: This is dangerous and can flood all the Segment Store logs, leading to lose logs due to their sheer size and/or making debug much harder. Please, add a single log line with the length of `chunksToDelete` if you want.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,263 @@
...
+            val future =
+                    txn.get(infoToDelete.name)
+                    .thenComposeAsync( metadata -> {
+                            val chunkMetadata = (ChunkMetadata) metadata;
+                            // Delete if the chunk is not present at all in the metadata or is present but marked as inactive.
+                            boolean shouldDelete = (null == chunkMetadata) ? true : !chunkMetadata.isActive();
+                            // Check whether the chunk is marked as inactive
+                            if (shouldDelete) {
+                                return chunkedSegmentStorage.getChunkStorage().openWrite(chunkToDelete)
+                                        .thenComposeAsync(chunkedSegmentStorage.getChunkStorage()::delete, executor)
```
RaulGracia, line 210: We discussed that this could lead to another problem: if the chunk is deleted and then something fails, the metadata will continue reflecting that the chunk is there and needs deletion. The same chunk will be picked, which has been deleted from storage, and the deletion will fail, skipping again the deletion in metadata. So this could lead to an infinite loop. Please, either modify this logic to handle the metadata deletion first (so in the worst case if there is a failure after that we would leave one orphan chunk to delete), or handle the case in which the chunk deletion succeeds bu there is an exception just after that, so we can eventually delete it from metadata too and avoid the infinite retries on that chunk.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,263 @@
...
+                                        .handleAsync((v, e) -> {
+                                            if (e != null) {
+                                                val ex = Exceptions.unwrap(e);
+                                                if (ex instanceof ChunkNotFoundException) {
+                                                    // Ignore - nothing to do here.
+                                                    log.debug("{} deleteGarbage - Could not delete garbage chunk {}.", chunkedSegmentStorage.getLogPrefix(), chunkToDelete);
+                                                } else {
+                                                    log.warn("{} deleteGarbage - Could not delete garbage chunk {}.", chunkedSegmentStorage.getLogPrefix(), chunkToDelete);
+                                                    // Queue it back.
+                                                    addToGarbage(Collections.singleton(chunkToDelete));
```
RaulGracia, line 227: If we have, let's say, an `IOError` on that chunk that is persistent, does it mean that we will try to garbage collect it forever?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
RaulGracia: Also, the fact that we haven't found this problem in the test cases of this PR makes me feel that more effort is necessary on testing the potential scenarios that could happen to the garbage collector (fail in chunk deletion but not in metadata, the other way raound, etc.).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
RaulGracia: Please, add a test for the enforcement of queue size limits of Garbage Collector, as the fact that no test is failing due to this makes me feel that it needs better coverage.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/Constants.java
```diff
@@ -0,0 +1,35 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.segmentstore.storage.metadata;
+
+/**
+ * Class that defines various constants.
+ */
+final class Constants {
```
RaulGracia, line 15: The name and description of this class is too generic. Please, refactor its name to its actual scope, like `StorageConstants` or `ChunkedStorageConstants` or something of the sort. Note that if we lookup for "Constants" in the IDE, it will take us more effort to understand what this class is used for, as the name itself is not helping.

RaulGracia (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -178,7 +174,9 @@ public ChunkedSegmentStorage(int containerId, ChunkStorage chunkStorage, ChunkMe
 
         // Now bootstrap
         log.debug("{} STORAGE BOOT: Started.", logPrefix);
-        return this.systemJournal.bootstrap(epoch).thenRunAsync(() -> log.debug("{} STORAGE BOOT: Ended.", logPrefix), executor);
+        return this.systemJournal.bootstrap(epoch)
+                .thenRun(() -> garbageCollector.initialize())
+                .thenRun(() -> log.debug("{} STORAGE BOOT: Ended.", logPrefix));
```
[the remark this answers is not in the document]
sachin-j-joshi: There could be other things added to initialization sequence or reshuffled.
This way the log it is always last.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
```diff
@@ -36,6 +38,11 @@
     public static final Property<Boolean> INLINE_DEFRAG_ENABLED = Property.named("defrag.inline.enable", true);
     public static final Property<Long> DEFAULT_ROLLOVER_SIZE = Property.named("metadata.rollover.size.bytes.max", SegmentRollingPolicy.MAX_CHUNK_LENGTH);
     public static final Property<Integer> SELF_CHECK_LATE_WARNING_THRESHOLD = Property.named("self.check.late", 100);
+    public static final Property<Integer> GARBAGE_COLLECTION_DELAY = Property.named("garbage.collection.delay.seconds", 60);
+    public static final Property<Integer> GARBAGE_COLLECTION_CONCURRENCY = Property.named("garbage.collection.concurrency", 10);
```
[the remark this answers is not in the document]
sachin-j-joshi: We need a name that captures throttling aspect. 
renamed it to `garbage.collection.concurrency.max`

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: yes it was a silly mistake and tests didn't catch it :(
Added new tests.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -254,6 +252,8 @@ public void initialize(long containerEpoch) {
                                 .thenComposeAsync(chunkInfo -> {
                                     Preconditions.checkState(chunkInfo != null, "chunkInfo for last chunk must not be null.");
                                     Preconditions.checkState(lastChunk != null, "last chunk metadata must not be null.");
+                                    // Mark chunk as "not garbage" if present.
+                                    //garbageCollector.removeFromGarbage(lastChunkName);
```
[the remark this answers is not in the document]
sachin-j-joshi: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
sachin-j-joshi: Storage GC is new . I don't know what is a good number here. We'll have to measure and get a sense for it.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
sachin-j-joshi: Fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
sachin-j-joshi: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: added tests

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: yes.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/Constants.java
sachin-j-joshi: done.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: If attempt to delete failed then we should retry.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: ok . made it a debug.

sachin-j-joshi:
@RaulGracia Thanks a lot for reviewing.

I added more tests.

I also made some changes to the underlying data structures.
1. I'm now using DelayQueue to keep track of work items. This is a framework collection and there is no good reason why we shouldn't be using it.
2. However it is a blocking queue so GC now uses its own thread to poll.
3. Concurrent Data structures are weakly consistent, so making timing of various things work reliably in tests was bit tricky but finally got it working with ManualDelay class.

I'll kick off system tests and other tests.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
sachin-j-joshi: done

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageMockTests.java
sachin-j-joshi: done

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
sachin-j-joshi: fixed.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageMockTests.java
sachin-j-joshi: outdated.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/Constants.java
sachin-j-joshi: need to use in two package.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: done.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: done.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: ok.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: made void

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: done.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: changed to priority queue

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: fixed.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: done.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: outdated. But new code doesn't have it.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: out dated . But new code has different structure.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,385 @@
...
+     * @param chunkStorage        ChunkStorage instance to use for writing all logs.
+     * @param metadataStore       ChunkMetadataStore for owner container.
+     * @param config              Configuration options for this ChunkedSegmentStorage instance.
+     * @param executorService     ScheduledExecutorService to use.
+     */
+    public GarbageCollector(int containerId, ChunkStorage chunkStorage,
+                            ChunkMetadataStore metadataStore,
+                            ChunkedSegmentStorageConfig config,
+                            ScheduledExecutorService executorService
+                            ) {
```
RaulGracia, line 106: Nit: Please, move this parenthesis to the end of previous line.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    /**
+     * Gets a value indicating how much to wait for the service to shut down, before failing it.
+     *
+     * @return The Duration.
+     */
+    @Override
+    protected Duration getShutdownTimeout() {
+        return null;
```
RaulGracia, line 152: Is this the expected value to return?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,385 @@
...
+            log.debug("{}: deleteGarbage - Queue full. Could not delete garbage. chunk {}.", traceObjectId, chunkToDelete);
+        }
+    }
+
+    /**
+     * Delete the garbage chunks.
+     *
+     * This method retrieves a few eligible chunks for deletion at a time.
+     * The chunk is deleted only if the metadata for it does not exist or is marked inactive.
+     * If there are any errors then failed chunk is enqueued back.
```
RaulGracia, line 230: > If there are any errors then failed chunk is enqueued back.

To describe the actual behavior, we would need to add `up to a max number of attempts`, right?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -0,0 +1,385 @@
...
+                                // Delete chunk from storage.
+                                if (shouldDeleteChunk) {
+                                    return chunkStorage.openWrite(chunkToDelete)
+                                            .thenComposeAsync(chunkStorage::delete, storageExecutor)
+                                            .handleAsync((v, e) -> {
+                                                if (e != null) {
+                                                    val ex = Exceptions.unwrap(e);
+                                                    if (ex instanceof ChunkNotFoundException) {
+                                                        // Ignore - nothing to do here.
+                                                        log.debug("{}: deleteGarbage - Could not delete garbage chunk={}.", traceObjectId, chunkToDelete);
```
RaulGracia, line 297: Maybe is a personal impression, but in general I feel that the code is heavily influenced by logging. There are many `if/else` branches, computation stages/futures, etc. that exist just to add debug logs. I don't know if this is related to the need of explicitly stating what the program is doing at any time, as there are scenarios that we are unsure of its behavior. If this is the case, maybe we could remove logging by adding more tests and splitting large code blocks into smaller functions that can be heavily tested. Maybe this would give us more certainty and this, in turn, would help us to remove some of the log lines in the PR. But, even in the case that we decide to do not do that now, I wonder if so much logging would be counterproductive, as logs will grow significantly, making them harder to manage.

RaulGracia (changes requested):
Minor comments.

RaulGracia:
There seems to be a resource leek in the tests added according to the last build failure:
```
Execution failed for task ':segmentstore:storage:test'.
> Process 'Gradle Test Executor 13' finished with non-zero exit value 99
  This problem might be caused by incorrect test process configuration.
  Please refer to the test execution section in the User Manual at https://docs.gradle.org/6.5.1/userguide/java_testing.html#sec:test_execution
```

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: There are 5 info level logs which I find important
```
log.info ("{}: Iteration {} started.", traceObjectId, iterationId.get());
log.info("{}: Iteration {} ended.", traceObjectId, iterationId.getAndIncrement());
log.info("{}: deleteGarbage - suspended - sleeping for {}.", traceObjectId, config.getGarbageCollectionDelay());
log.info("{}: deleteGarbage - Chunk is not marked as garbage chunk={}.", traceObjectId, chunkToDelete);
log.info("{}: deleteGarbage - could not delete after max attempts chunk={}.", traceObjectId, chunkToDelete);
```

Following are important to know what went wrong
```
log.warn("{}: deleteGarbage - Queue full. Could not delete garbage. Chunks skipped", traceObjectId);
log.warn("{}: deleteGarbage - Could not delete garbage chunk={}.", traceObjectId, chunkToDelete);
                                                        
```

following are unusual circumstances that need to be recorded

```
log.error("{}: Error during doRun.", traceObjectId, ex);
log.error(String.format("%s deleteGarbage - Could not delete metadata for garbage chunk=%s.",
                                                            traceObjectId, chunkToDelete), e);
log.error(String.format("%s deleteGarbage - Could not find garbage chunk=%s.",
                                            traceObjectId, chunkToDelete), ex);
```

Perhaps last two above could be info or warn.


Other logs are for debug purpose and as such they need to show details of what happened ....
we'll tweak this as we go along ..some of them will become trace as we gain more run time with this code.

RaulGracia:
@sachin-j-joshi I also observed this warning in your build:
```
18:53:18  /var/jenkins_home/workspace/pravega-branch-builder/segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java:1961: warning: You're assigning an auto-cleanup variable to something else. This is a bad idea.
18:53:18          val oldTestCotext = testContext;
```

sachin-j-joshi:
Are you sure you are looking at the correct version?
That method does not mark testcontext with `@cleanup`
https://github.com/sachin-j-joshi/pravega/blob/issue-4903-improve-garbage-collection-2/segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java#L1961

RaulGracia (approved):

RaulGracia:
@sachin-j-joshi I'm approving the PR. Once you confirm that Jenkins/system tests are green, then we can decide the merge procedure for this one.

PR #5417, opened 2020-12-10
Issue 5416: Transaction commit order should take 'limit' elements from ordered list

### LINKED ISSUE #5416: Transaction order regression caused by 5197
**Problem description**
With #5197 we regressed the transaction ordering if the number of transactions in committing state are more than the `limit` (100) as we do not honor the order among them and end up picking 100 of them and then applying order on the chosen list. 
the list should be chosen in order itself 

**Problem location**
PersistentStreamBase

Fix the bug

shiveshr:

**Change log description**  
Fixes transaction commit order regression introduced by #5197 where we took first n elements from an unsorted list whereby breaking the order. 

**Purpose of the change**  
Fixes #5416 

**What the code does**  
There were two regressions introduced by 5197. 
1. We took the "limit" number of elements from an unordered list which could break the order.
2. in subsequent iterations of the loop (if some elements were filtered out), we were reading from wrong index in the loop. 

**How to verify it**  
Unit test added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
@@ -1957,7 +1957,7 @@ public abstract class PersistentStreamBase implements Stream {
                     case ABORTED:
                     case UNKNOWN:
                         // Aborting, aborted, unknown and committed 
-                        log.debug("stale txn {} with status. removing {}", txnId, txnRecord.getTxnStatus(), order);
+                        log.debug("stale txn {} with status {}. removing {}", txnId, txnRecord.getTxnStatus(), order);
                         toPurge.add(order);
                         break;
                 }
--- a/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
@@ -1920,7 +1920,7 @@ public abstract class PersistentStreamBase implements Stream {
         int epoch = nextEpoch.getKey();
         List<Long> orders = new ArrayList<>();
         List<String> txnIds = new ArrayList<>();
-        nextEpoch.getValue().forEach(x -> {
+        nextEpoch.getValue().stream().sorted(Comparator.comparingLong(Map.Entry::getKey)).forEach(x -> {
             orders.add(x.getKey());
             txnIds.add(x.getValue());
         });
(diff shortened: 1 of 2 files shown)


controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1931,8 +1931,9 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
                 () -> getTransactionRecords(epoch, txnIds.subList(from.get(), till.get())).thenAccept(txns -> {
             for (int i = 0; i < txns.size(); i++) {
                 ActiveTxnRecord txnRecord = txns.get(i);
-                UUID txnId = UUID.fromString(txnIds.get(i));
-                long order = orders.get(i);
+                int index = from.get() + i;
```
shrids, line 1934: wow, good catch 👍

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1931,8 +1931,9 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
...
+                UUID txnId = UUID.fromString(txnIds.get(index));
+                long order = orders.get(index);
```
shrids, line 1936: missing parameter in line :1960

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1931,8 +1931,9 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
                 () -> getTransactionRecords(epoch, txnIds.subList(from.get(), till.get())).thenAccept(txns -> {
             for (int i = 0; i < txns.size(); i++) {
                 ActiveTxnRecord txnRecord = txns.get(i);
```
shrids, line 1933: when will `txnRecord.commitOrder` be different from` orders.get(index)`?

shrids (approved):
LGTM apart from the below comments.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: Because we first create znode and then we update the txn record, we could have failed after 1st. Or could have recreated another znode in a different call that then doesn’t update the record.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: I didn’t follow, what param?

RaulGracia:
@shiveshr should this be cherry-picked to 0.8? Let's clarify this to create a new 0.8 RC.

shrids:
The following tests failed 
```
io.pravega.segmentstore.server.store.StreamSegmentServiceNoOpWriteOnlyTests > testEndToEnd FAILED
    java.lang.AssertionError: Unexpected attribute value from getAttributes(). expected:<200> but was:<-9223372036854775808>
io.pravega.segmentstore.server.store.StreamSegmentServiceTests > testEndToEnd FAILED
    java.lang.AssertionError: Segment 'Segment_5' was not deleted. Exception thrown was of unexpected type: io.pravega.segmentstore.contracts.ContainerNotFoundException: [Container 2] Container Id does not exist..
```
Retriggering the build...

shrids:
System test run is successful ref:  `test-pravega-continuous/3380/`

PR #5422, opened 2020-12-11
Issue 5421: Avoid closing pre-existing metrics.

### LINKED ISSUE #5421: Duplicate StatsLoggers Metrics cause the Metric to become unreported.
**Problem description**

If a `StatsLoggerProxy` is requested to create a pre-existing meter (`Counter`, `Gauge`, etc), `StatsLoggerProxy.getOrSet` causes the pre-existing meter to be closed, and therefore removed from the global `CompositeMeterRegitsry`. This means that when exporting metrics to say, an Influx instance, no result will be reported. 

**Problem location**

`StatsLoggerProxy`

Instead of attempting to optimistically create a metric ahead of time, check to see if it exists first. Not only does attempting to create these metrics optimistically destroy program correctness, but is also wasteful as almost all operations will be get operations.

co-jo:
**Change log description**  
In the case a duplicate meter on a `StatsLogger` is ever created, it avoids inadvertently closing the very same metric we are trying to use.

**Purpose of the change**  
Fixes #5421

**What the code does**  
* The old implementation attempted to an optimistic scheme, creating a metric in advance and removing it if it already exists. The problem with this approach is creating another metric (`Meter`) that already exists within the `CompositeMeterRegistry` returns the same instance. If one exists already, we are therefore closing this existing metric, removing it from the `CompositeMeterRegistry` making invisible to any sort of exporter. 

Unless we rework the metrics to a larger degree, it does not seem there is a way of getting around the use of `computeIfAbsent`. Many of the operations of `MetricProxy` have side effects, so the entire `getOrSet` operation must be atomic or we risk using invalidate objects during concurrent accesses.

Almost all `StatsLoggerProxy` method calls are either static or done during initialization of long lived objects so there should be minimal blocking through use of `computeIfAbsent`.


shared/metrics/src/test/java/io/pravega/shared/metrics/StatsLoggerProxyTest.java
```diff
@@ -67,23 +67,20 @@ public void testGetOrCreate() {
         createMetrics(proxy, "", "hostname", "localhost");
         Assert.assertEquals("Unexpected number of metrics registered initially.", METRIC_COUNT * 2, metrics.size());
 
-        // Re-create/request the same metrics. Verify the original ones have not been touched, but we did create (and close)
-        // a new set.
+        // Re-create/request the same metrics. Verify the original ones have not been touched.
         createMetrics(proxy, "");
-        Assert.assertEquals("Unexpected number of metrics registered after re-registering same names.", 3 * METRIC_COUNT, metrics.size());
+        Assert.assertEquals("Unexpected number of metrics registered after re-registering same names.", 2 * METRIC_COUNT, metrics.size());
```
co-jo, line 72: The `expected` argument has been changed because before the expectation that was we would still create the `Metric` and add it to the `metrics` array, but just immediately mark it as closed. Now we just avoid that creation step completely (if it exists).

tkaitchuck (approved):

RaulGracia (approved):
LGTM, but please make sure that system tests pass and that this fixes the original problem reported by users.

RaulGracia:
@co-jo @anirudhkovuru I have seen the build failing twice due to the secure Admin CLI commands:
```
io.pravega.cli.admin.controller.SecureControllerCommandsTest > testDescribeScopeCommand FAILED
io.pravega.cli.admin.controller.SecureControllerCommandsTest > testListStreamsCommand FAILED
io.pravega.cli.admin.controller.SecureControllerCommandsTest > testListReaderGroupsCommand FAILED
io.pravega.cli.admin.controller.SecureControllerCommandsTest > testAuthConfig FAILED
io.pravega.cli.admin.controller.SecureControllerCommandsTest > testListScopesCommand FAILED
```
Any idea if this change could be making these tests flaky or the issue is there irrespective of this PR?

RaulGracia (changes requested):
Blocking until build is fixed.

RaulGracia:
@co-jo build is passing, but it looks that we need some more test coverage for this PR.

RaulGracia (approved):

PR #5434, opened 2020-12-17
Issue 5405: Improve logging on client

### LINKED ISSUE #5405: Logging improvment: ReaderGroup restart process has false SegmentIsTruncated error
**Problem description**
Flink longevity has failed due to some reasons and when TaskManager is restarted, a false SegmentIsTrucated error happens when recreating ReaderGroup. 

I have been able to reproduce this error by using Pravega HelloWorldReader example. The root problem seems to be if internal ReaderGroup segment has been truncated, the recreating process doesn't pick up the latest readable offset at first. 

**Problem location**
https://github.com/pravega/pravega/blob/1ad51211669ff59a90ca403be823079db6d60720/client/src/main/java/io/pravega/client/stream/impl/ClientFactoryImpl.java#L216

Get the latest the starting offset for the internal RG segment when recreating ReaderGroup

Tristan1900:
**Change log description** 
 1. Removes the false alarm of `SegmentIsTruncated` exception by not sending prefetch read request if the startOffset is unknown. (this is done by #5432  )
2. Adds info level log at `resetReaderGroup`. This method is not expected to be called very frequently. 

**Purpose of the change**  
Fix #5405 

**How to verify it**  
Test should pass.


client/src/main/java/io/pravega/client/segment/impl/SegmentInputStreamFactoryImpl.java
```diff
@@ -50,7 +50,7 @@ private EventSegmentReader getEventSegmentReader(Segment segment, Semaphore hasD
                 DelegationTokenProviderFactory.create(delegationToken, controller, segment, AccessOperation.READ), hasData);
         async.getConnection();                      //Sanity enforcement
         bufferSize = MathHelpers.minMax(bufferSize, SegmentInputStreamImpl.MIN_BUFFER_SIZE, SegmentInputStreamImpl.MAX_BUFFER_SIZE);
-        return getEventSegmentReader(async, 0, endOffset, bufferSize);
+        return getEventSegmentReader(async, -1, endOffset, bufferSize);
```
shrids, line 53: This change will affect all the readers, can we localize it?

tkaitchuck (approved):

shrids (approved):

PR #5439, opened 2020-12-17
Issue 5112: Client API to Create Subscriber reader groups

### LINKED ISSUE #5112: Client API to Create Subscriber reader groups
**Problem description**
Client side changes for storing Reader Groups on Controller.

**Problem location**
Client

Make controller calls to create, delete and update the `ReaderGroup on Controller when user invokes create, reset, delete RG on Client.
Keep Reader Group Configuration in State Synchronizer updated with the version on Controller

anirudhkovuru:
**Change log description**  
The `ReaderGroup` operations `createReaderGroup`, `deleteReaderGroup` and `resetReaderGroup` have been modified such that, in the case of the `ReaderGroup` being a subscriber the controller calls are made to ensure consistent subscription/unsubscription to all the necessary streams.

Note: Requires #5400 to build.

**Purpose of the change**  
Fixes #5112, #5543

**What the code does**  
- `ReaderGroupState` now has an `updatingConfig` field that indicates if the `ReaderGroupConfig` is being updated or not.
    - Added a new state update in `ReaderGroupState` to change the above field.
- Changed the methods: `createReaderGroup`, `deleteReaderGroup` and `resetReaderGroup` to make controller calls so that group is consistent with its version on the controller.
- Added a new timer to the `ReaderGroupStateManager` that will help periodically update the config to the one present on the controller to maintain consistency between the client and controller versions of the config.
- Added fix for flaky test StreamMetricsTest.testStreamsAndScopesBasicMetricsTests 

**How to verify it**  


client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -200,8 +201,12 @@ private Checkpoint completeCheckpoint(String checkpointName) {
 
     @Override
     public void resetReaderGroup(ReaderGroupConfig config) {
+        synchronizer.updateState((state, updates) -> {
+            updates.add(new UpdatingConfig(true));
+        });
+        getThrowingException(controller.updateReaderGroup(scope, groupName, config));
         Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, config);
-        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config)));
+        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config), false));
```
tkaitchuck, line 209: This needs to be conditional.
We don't want to perform this update if: 
a) It's already been completed.
b) Another even newer config update has been completed.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -200,8 +201,12 @@ private Checkpoint completeCheckpoint(String checkpointName) {
 
     @Override
     public void resetReaderGroup(ReaderGroupConfig config) {
+        synchronizer.updateState((state, updates) -> {
```
tkaitchuck, line 204: What if updatingConfig is already set to true?

What happens in the following race:
resetReadeGroup(a) - start
resetReadeGroup(b) - start
resetReadeGroup(b) - complete
resetReadeGroup(a) - ???

There is no lock between here and the controller. So an order needs to be defined.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -330,6 +337,21 @@ private void fetchUpdatesIfNeeded() {
             compactIfNeeded();
         }
     }
+
+    void updateConfigIfNeeded() {
+        sync.fetchUpdates();
```
tkaitchuck, line 342: Shouldn't this be `fetchUpdatesIfNeeded()`?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -330,6 +337,21 @@ private void fetchUpdatesIfNeeded() {
...
+        ReaderGroupState state = sync.getState();
+        if (!updateConfigTimer.hasRemaining() && state.isUpdatingConfig()) {
+            log.debug("Update the group config");
```
tkaitchuck, line 345: Be more specific this won't actually help anyone debug a problem if there is one. 
Which reader group? What config? Is an update actually being performed?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -330,6 +337,21 @@ private void fetchUpdatesIfNeeded() {
...
+            ReaderGroupConfig controllerConfig = getThrowingException(controller.getReaderGroupConfig(sync.getSynchronizerScopeName(),
+                    sync.getSynchronizerStreamName().replace(READER_GROUP_STREAM_PREFIX, "")));
```
tkaitchuck, line 347: This precludes cross scope reading, which is an important feature. 
Don't make assumptions about the scope and stream name here. Explicitly obtain them from the state. (Or add them as member variables to this class)

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -330,6 +337,21 @@ private void fetchUpdatesIfNeeded() {
...
+            if (state.getConfig().getGeneration() < controllerConfig.getGeneration()) {
+                Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, controllerConfig);
+                sync.updateStateUnconditionally(new ReaderGroupState.ReaderGroupStateInit(controllerConfig, segments, getEndSegmentsForStreams(controllerConfig), false));
+            }
+            updateConfigTimer.reset(UPDATE_CONFIG_WINDOW);
```
tkaitchuck, line 352: This only resets the timer if a config update was actually performed. However the goal of the timer was to minimise work on a per-event bais. This prevents that because the timer will almost never be not expired if the config updates rarely.

tkaitchuck (changes requested):

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
anirudhkovuru: As you mentioned it would only reset if an update was done. I've separated the timer check and the `updatingConfig` check. So irrespective of whether the update happens or not the timer will reset itself when there is no time remaining.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
anirudhkovuru: The scope name and readergroup name are required to make the controller call. Instead of obtaining it from the sync these are now part of the `ReaderGroupStateManager`. This means we wouldn't need those `StateSynchronizer` API's too.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
anirudhkovuru: Made the debug statement more informative.

client/src/main/java/io/pravega/client/state/StateSynchronizer.java
```diff
@@ -45,6 +45,10 @@
      */
     StateT getState();
 
+    String getSynchronizerScopeName();
+
+    String getSynchronizerStreamName();
```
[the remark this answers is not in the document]
anirudhkovuru: These API's have been removed. They are not needed anymore.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
shiveshr: that should not be a problem because the call to controller should include the generation and hence idempotent. So if RG(b) has already completed the call, RG(a) will simply repeat it with the generation that has already been included in controller.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -200,8 +201,14 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
+            updates.add(new UpdatingConfig(true));
+        });
+        getThrowingException(controller.updateReaderGroup(scope, groupName, config));
         Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, config);
-        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config)));
+        synchronizer.updateState((state, updates) -> {
```
shiveshr, line 209: the problem really is "after" the call to controller when this code is calling `synchronizer.updateState(RGStateInit)`. 

if RG(b) has already successfully reset, then (a) should not be repeating it, rather exit. 
so instead of blindly resetting here, we should actually check if a reset with our desired generation has already completed or not.. and if not, then we should reset, else exit.

anirudhkovuru: Implementing the logic as explained above.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Its been made conditional.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: There is a check now to make sure that we only update `ReaderGroupState` once we have confirmed from the controller that the update was made.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
anirudhkovuru: Changed to use `fetchUpdatesIfNeeded()`.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -200,8 +208,17 @@ private Checkpoint completeCheckpoint(String checkpointName) {
 
     @Override
     public void resetReaderGroup(ReaderGroupConfig config) {
+        synchronizer.updateState((state, updates) -> {
+            updates.add(new UpdatingConfig(true));
+        });
+        getThrowingException(controller.updateReaderGroup(scope, groupName, config));
+        ReaderGroupConfig controllerConfig = getThrowingException(controller.getReaderGroupConfig(scope, groupName));
```
tkaitchuck, line 215: Why is this call needed? Can it be combined with the previous one? (IE have it return this value)

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -200,8 +208,17 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
         Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, config);
-        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config)));
+        synchronizer.updateState((state, updates) -> {
+            if (state.getConfig().getGeneration() < controllerConfig.getGeneration()) {
+                new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config), false);
```
tkaitchuck, line 219: shouldn't it be using `controllerConfig` here?
Also, if so, then a unit test is missing.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
tkaitchuck: Initial config is version 0
resetReadeGroup(a) - Sets flag to 'true'
a is assigned version 1 by the controller.
resetReadeGroup(b)
b is assigned version 2 by the controller.
a unsets the flag and updates the readergroupstate.
b crashes.

Now the controller thinks the version is 2 and the config is b. But the group says A and nobody considers the update ongoing.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -200,8 +208,17 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
     public void resetReaderGroup(ReaderGroupConfig config) {
+        synchronizer.updateState((state, updates) -> {
+            updates.add(new UpdatingConfig(true));
+        });
+        getThrowingException(controller.updateReaderGroup(scope, groupName, config));
+        ReaderGroupConfig controllerConfig = getThrowingException(controller.getReaderGroupConfig(scope, groupName));
         Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, config);
-        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config)));
+        synchronizer.updateState((state, updates) -> {
+            if (state.getConfig().getGeneration() < controllerConfig.getGeneration()) {
```
tkaitchuck, line 218: If the config generation is older it should still disable the flag.

client/src/main/java/io/pravega/client/stream/ReaderGroup.java
```diff
@@ -47,7 +48,14 @@
...
      */
     String getGroupName();
-    
+
+    /**
+     * Returns the UUID of the group.
+     *
+     * @return Reader group UUID.
+     */
+    UUID getGroupId();
```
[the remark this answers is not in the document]
shiveshr: There is no reason for us to expose this on the interface.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
shiveshr: good point. 
i had originally suggested we manage the RG generation number changes on RG. that takes care of the above issue as the reset workflow is started on rg and terminated on rg and controller communication becomes an intermediate step. 

anyway reiterating the logic here:
updateRg: 
```
// if there are ongoing reset, finish it first before starting the new reset
while (true) {
  if rgstate.updating
    update controller with rgGen and config.
    conditionally reset RG state.  
    continue;
  else 
    newGen = rgGen + 1
    contidionally update -> StartRGReset(newGen, newConfig, updating)
    update controller
    conditionally reset RG state (resets readers and segments and changes updating flag)
    break;
}
// any conditional update failure should go back to the loop.

```
the difference above with the implemented logic is that we generate the "next generation" on RG and update it in rg state first and then in controller.. and after updating the controller we complete the reset of the RG. this way only one RG reset can be happening and only after it completes will the subsequent reset be attempted. and any partially completed reset will be left in that state until another operation on rg is attempted which should find rg is updating state and bring it to completion first before doing anything. and the state is only truly reset as the last step so rg state is never left in an inconsistent state or we dont reset it multiple times. 

@pbelgundi @anirudhkovuru can you guys express the pseudo code of how the implemented logic is supposed to be correct?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
shiveshr: yes, i think there is a github issue to address this. 
ideally updateRG should return the generation number itself but it presently returns a boolean.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
shiveshr: https://github.com/pravega/pravega/blob/3497a1129e78ecbea223ed579f631d116036eabd/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java#L207
controller.updateReadergroup is conditional. it will reject the request from (b) if (a) has already incremented the generation to 1. 
and if rg finds the generation on controller ahead of generation in its local state, it should first complete the previously attempted reset and only then should it even attempt the new reset which is to call the controller again to increment the generation.

pbelgundi: Here is how resetReaderGroup() API on Client would look:
```resetReaderGroup(RGConfig config) {
	1. ss.fetchUpdates();
	2. if (ss.updatingConfig==false) {
	  // this is a conditional update and so for 2 processes trying to do this update in parallel, only 1 will succeed.
	   2.1 val updateFlagResult = ss.updateState(ss.updatingConfig = true)
	   2.2 if (updateFlagResult == SUCCESS) {
		    set config.generation = ss.getConfig().generation;
		    set config.readerGroupId = ss.getConfig().readerGroupId;
		    val controllerUpdateResult = controller.updateRG(scope, rgName, config)
		    if (controllerUpdateResult == SUCCESS) {
			   ss.updateUnconditionally(config)
		    } 
		    // we set ss.updatingConfig = false in the following cases:
		    // (i) If both update to controller and SS have succeeded
		    // (ii) update to Controller has failed, & we never tried updating SS and directly landed here
		    // (iii) in the catch block for `controller.updateRg()` call, so if it throws an exception, we again set back "ss.updatingConfig = false"
		    ss.updateState(ss.updatingConfig = false)
	     } 
	     else {
		      // we're here because updateFlagResult==FAILED
		      // goto 1 
	         }
	} 
	else {
	   // here updatingConfig==true 
	  // goto 1 and keep retrying this reset in a while loop waiting for (updatingConfig == false) or bail out
	}
}
```

pbelgundi: Additionally, there would be a separate timer thread that would periodically check if config in SS matches with config on Controller and if it does not update SS with config copy from Controller.

client/src/main/java/io/pravega/client/stream/ReaderGroup.java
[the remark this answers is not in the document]
anirudhkovuru: This method has been moved to the impl instead. It wouldn't make sense to a user so removed it from the API.

shiveshr (approved):

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: This logic has been changed to make sure to account for the case you mentioned [below](https://github.com/pravega/pravega/pull/5439#discussion_r546115261)

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -53,9 +54,11 @@
 
     private final StreamDataRetention retentionType;
 
-    private final long generation;
+    @Setter
+    private long generation;
 
-    private final UUID readerGroupId;
+    @Setter
+    private UUID readerGroupId;
```
tkaitchuck, line 61: This is a public facing object and is immutable. I see no reason to change that. 
https://github.com/pravega/pravega/wiki/Contributing#threading

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -117,6 +119,16 @@ public void updateRetentionStreamCut(Map<Stream, StreamCut> streamCuts) {
                synchronizer.getState().getConfig().getRetentionType().toString());
     }
 
+    public UUID getGroupId() {
+        synchronizer.fetchUpdates();
```
tkaitchuck, line 123: It's strange that this calls fetchUpdates. This is a blocking remote call. This is a getter so it's confusing. It is also not clear when the readerGroupId would change and what consistency is attempting to be enforced by updating first.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -117,6 +119,16 @@ public void updateRetentionStreamCut(Map<Stream, StreamCut> streamCuts) {
...
+        return synchronizer.getState().getConfig().getReaderGroupId();
+    }
+
+    public long getGeneration() {
+        synchronizer.fetchUpdates();
```
tkaitchuck, line 128: Same here.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +215,39 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
+
+    private boolean stateTransition(ReaderGroupState state, ReaderGroupState.ReaderGroupStateUpdate update) {
+        // This boolean will help know if the update actually succeeds or not.
+        AtomicBoolean successfullyUpdated = new AtomicBoolean(true);
+        synchronizer.updateState((s, updates) -> {
+            // If successfullyUpdated is false then that means the current state where this update should
+            // take place (i.e. state with updatingConfig as false) is not the state we are in so we do not
+            // make the update.
+            successfullyUpdated.set(s.getConfig().equals(state.getConfig()));
+            if (successfullyUpdated.get()) {
```
tkaitchuck, line 246: You can use a local

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +215,39 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
-        Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, config);
-        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config)));
+        while (true) {
+            synchronizer.fetchUpdates();
+            val state = synchronizer.getState();
+            // We only move into the block if the state transition has happened successfully.
+            if (stateTransition(state, new UpdatingConfig(true))) {
+                // Use the latest generation and reader group Id.
+                config.setGeneration(state.getConfig().getGeneration());
+                config.setReaderGroupId(state.getConfig().getReaderGroupId());
```
tkaitchuck, line 225: I think what you want here is a way to create a builder from an existing config.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -330,6 +342,24 @@ private void fetchUpdatesIfNeeded() {
...
+    void updateConfigIfNeeded() {
+        if (!updateConfigTimer.hasRemaining()) {
+            fetchUpdatesIfNeeded();
+            ReaderGroupState state = sync.getState();
+            if (state.isUpdatingConfig()) {
+                sync.updateState((s, updates) -> {
+                    ReaderGroupConfig controllerConfig = getThrowingException(controller.getReaderGroupConfig(scope, groupName));
+                    log.debug("Updating the readergroup {} with the new config {} obtained from the controller", groupName, controllerConfig);
+                    if (s.getConfig().getGeneration() < controllerConfig.getGeneration()) {
+                        Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, controllerConfig);
```
tkaitchuck, line 355: This is making multiple calls to the controller inside of the updateState function. This doesn't work well for optimistic concurrency. Perhaps most this outside of the update state, and just rely on the `if` to make the update conditional on the config not having changed.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -330,6 +342,24 @@ private void fetchUpdatesIfNeeded() {
...
         }
     }
+
+    void updateConfigIfNeeded() {
+        if (!updateConfigTimer.hasRemaining()) {
+            fetchUpdatesIfNeeded();
+            ReaderGroupState state = sync.getState();
+            if (state.isUpdatingConfig()) {
+                sync.updateState((s, updates) -> {
+                    ReaderGroupConfig controllerConfig = getThrowingException(controller.getReaderGroupConfig(scope, groupName));
```
tkaitchuck, line 352: Same here. (see below)

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Using `toBuilder` in the Builder annotation instead of setters.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
anirudhkovuru: Removed the Setters

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
anirudhkovuru: Moved the controller calls outside the update state.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
anirudhkovuru: Moved them outside

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Using local as suggested.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Here we do not need to call fetchUpdates for consistency as the groupId never changes.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +214,41 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
+    }
+
+    private boolean stateTransition(ReaderGroupState state, ReaderGroupState.ReaderGroupStateUpdate update) {
+        // This boolean will help know if the update actually succeeds or not.
+        AtomicBoolean successfullyUpdated = new AtomicBoolean(true);
+        synchronizer.updateState((s, updates) -> {
+            // If successfullyUpdated is false then that means the current state where this update should
+            // take place (i.e. state with updatingConfig as false) is not the state we are in so we do not
+            // make the update.
+            boolean updated = s.getConfig().equals(state.getConfig());
```
tkaitchuck, line 245: State is only used for config. So make the parameter 'config' instead. This would also minorly simplify the caller.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +214,41 @@ private Checkpoint completeCheckpoint(String checkpointName) {
 
     @Override
     public void resetReaderGroup(ReaderGroupConfig config) {
-        Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, config);
-        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config)));
+        while (true) {
+            synchronizer.fetchUpdates();
```
tkaitchuck, line 218: The update function in stateTransition will do this.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +214,41 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
+            // We only move into the block if the state transition has happened successfully.
+            if (stateTransition(state, new UpdatingConfig(true))) {
+                // Use the latest generation and reader group Id.
+                ReaderGroupConfig newConfig = config.toBuilder()
+                        .readerGroupId(state.getConfig().getReaderGroupId())
+                        .generation(state.getConfig().getGeneration()).build();
+                boolean success = Futures.getThrowingException(controller.updateReaderGroup(scope, groupName, newConfig));
+                if (success) {
+                    synchronizer.updateState((s, updates) -> {
+                        Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, newConfig);
```
tkaitchuck, line 229: getSegmentsForStreams is making an RPC to the controller. Please move this out of the updateState block.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +214,41 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
-        synchronizer.updateStateUnconditionally(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(config)));
+        while (true) {
+            synchronizer.fetchUpdates();
+            val state = synchronizer.getState();
+            // We only move into the block if the state transition has happened successfully.
+            if (stateTransition(state, new UpdatingConfig(true))) {
+                // Use the latest generation and reader group Id.
+                ReaderGroupConfig newConfig = config.toBuilder()
+                        .readerGroupId(state.getConfig().getReaderGroupId())
+                        .generation(state.getConfig().getGeneration()).build();
```
tkaitchuck, line 225: Doesn't the generation need to change somewhere?

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndReaderGroupTest.java
```diff
@@ -153,6 +156,92 @@ public void testDeleteReaderGroup() throws Exception {
         assertEquals("data1", eventRead.getEvent());
     }
 
+    @Test
```
tkaitchuck, line 159: For these tests, can we also add an assertion that advancing the reader group will correctly result in data being truncated?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: The while loop in `resetReaderGroup` is so that we can retry in case the stateTransition fails.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: The generation is updated on the controller. We update the config's generation on to the controller using the periodic update that we have on the `ReaderGroupStateManager`.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Using config instead.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Moved out.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +214,41 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
+            synchronizer.fetchUpdates();
+            val currentConfig = synchronizer.getState().getConfig();
+            // We only move into the block if the state transition has happened successfully.
+            if (stateTransition(currentConfig, new UpdatingConfig(true))) {
+                // Use the latest generation and reader group Id.
+                ReaderGroupConfig newConfig = config.toBuilder()
+                        .readerGroupId(currentConfig.getReaderGroupId())
+                        .generation(currentConfig.getGeneration()).build();
+                boolean success = Futures.getThrowingException(controller.updateReaderGroup(scope, groupName, newConfig));
+                Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, newConfig);
```
pbelgundi, line 227: `getSegmentsForStreams`  should be inside the `if (success)` block since it is not needed if controller.updateReaderGroup() has failed.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +214,41 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
+    }
+
+    private boolean stateTransition(ReaderGroupConfig config, ReaderGroupState.ReaderGroupStateUpdate update) {
+        // This boolean will help know if the update actually succeeds or not.
+        AtomicBoolean successfullyUpdated = new AtomicBoolean(true);
+        synchronizer.updateState((state, updates) -> {
+            // If successfullyUpdated is false then that means the current state where this update should
+            // take place (i.e. state with updatingConfig as false) is not the state we are in so we do not
+            // make the update.
+            boolean updated = state.getConfig().equals(config);
```
pbelgundi, line 245: I don't see equals() and hashcode() overridden in ReaderGroupConfig, so this would do a reference comparison. That does not look right to me.

pbelgundi (changes requested):

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: It is inside the if block

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Removed fetchUpdates from here.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
tkaitchuck: I think you misunderstand:
Calling `update` on a state synchronizer will be retried can cause the update function to be re-run if it is not up to date. So it is not necessary to call fetchUpdates between calls to `update` in a loop because by definition the state will be up to date. That is the contract of the `update` function.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
```diff
@@ -203,8 +217,41 @@ private Checkpoint completeCheckpoint(String checkpointName) {
...
+            if (stateTransition(currentConfig, new UpdatingConfig(true))) {
+                // Use the latest generation and reader group Id.
+                ReaderGroupConfig newConfig = config.toBuilder()
+                        .readerGroupId(currentConfig.getReaderGroupId())
+                        .generation(currentConfig.getGeneration()).build();
+                boolean success = Futures.getThrowingException(controller.updateReaderGroup(scope, groupName, newConfig));
+                if (success) {
+                    Map<SegmentWithRange, Long> segments = getSegmentsForStreams(controller, newConfig);
+                    synchronizer.updateState((s, updates) -> {
+                        updates.add(new ReaderGroupStateInit(config, segments, getEndSegmentsForStreams(newConfig), false));
```
tkaitchuck, line 233: This is storing `config` which was passed as a parameter. It does not appear to have a generation set on it.

What I see happening now is that resetReaderGroup is called with a config provided by the user. Presumably they don't set the generation. (Should that method be private?) So the default value of -1 is passed. The call ends up here and it passes the config into the state init. But in `updateConfigIfNeeded` it pulls from the controller which updates the generation id every time. So it will always return a new config. Thus we end up reinitializing the group state twice.

tkaitchuck (commented):
> The generation is updated on the controller. We update the config's generation on to the controller using the periodic update that we have on the ReaderGroupStateManager.
https://github.com/anirudhkovuru/pravega/blob/c4c85c732154321d1e598018f6ba4c5de1ee727b/client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java#L346-L362

This does not work because the logic assumes the config stored in the reader group state has the same generation as the one in the controller. (For any given generation). The value passed in is not going to reflect this.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
pbelgundi: Since ReaderGroupConfig has Lombok annotation @Data that takes care of implementing equals and hashcode

pbelgundi (approved):

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
tkaitchuck: I don't see a change. Was it pushed?

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
pbelgundi: `controller.updateReaderGroup` now returns the updated generation( See PR: https://github.com/pravega/pravega/pull/5462), which should be written to StateSynchronizer, but I don't see that happening here.

pbelgundi (changes requested):

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: I have made it such that the updated generation is returned. That generation is what is written into the `StateSynchronizer`.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: The generation that is written into the `StateSynchronizer` is the one returned after `controller.updateReaderGroup` has been called.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: I've moved the `synchronizer.fetchUpdates` to outside the while loop so its called once before the while loop. Then within the while loop `update` will allow for the state to be up to date.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: The updateRG now returns the generation number which is used to update the state too.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: Implementing the pseudocode as described above.

client/src/main/java/io/pravega/client/control/impl/Controller.java
```diff
@@ -141,7 +141,7 @@
      * @return A future which will throw if the operation fails, otherwise returning a boolean to
      *         indicate that the subscriber was updated in Stream Metadata.
      */
-    CompletableFuture<Boolean> updateReaderGroup(final String scopeName, final String rgName, ReaderGroupConfig config);
+    CompletableFuture<Long> updateReaderGroup(final String scopeName, final String rgName, ReaderGroupConfig config);
```
pbelgundi, line 144: IMO, this should return `updateReaderGroupResponse` instead of only returning Long -  generation. Since the return status is lost when we return only generation.

pbelgundi (changes requested):

pbelgundi (approved):
LGTM

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupImpl.java
anirudhkovuru: When calling `deleteReaderGroup` on the controller end we need to pass the generation and readergroupID for which we need their latest versions, hence the `fetchUpdates`. I had initially thought to get the ReaderGroup and then use Impl methods to get these values as I have done above. However, since these getters for the generation and groupID are only being called at one place (in `deleteReaderGroup`) instead of getting the `ReaderGroupImpl` I make the `StateSynchronizer`, call `fetchUpdates` once and then access the latest versions of generation and groupID.

client/src/main/java/io/pravega/client/control/impl/Controller.java
```diff
@@ -141,7 +141,7 @@
      * @return A future which will throw if the operation fails, otherwise returning a boolean to
      *         indicate that the subscriber was updated in Stream Metadata.
```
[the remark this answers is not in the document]
anirudhkovuru: Javadoc is updated accordingly.

test/integration/src/test/java/io/pravega/test/integration/endtoendtest/EndToEndReaderGroupTest.java
anirudhkovuru: Added a test for the case.

tkaitchuck (approved):

PR #5457, opened 2021-01-04
Issue 5456: (SLTS) - Handle possible failure during deletion of metadata keys can cause version mismatch in next attempt

### LINKED ISSUE #5456: (SLTS) - Possible failure during deletion of metadata keys can cause version mismatch in next attempt.
**Problem description**
Currently table store has a severe limitation that it can not update and delete keys in the same transaction.
The current work around for this problem is implemented in two steps in TableBasedMetadataStore.writeAll method. 
It involves first marking the keys as deleted by updating in first transaction and immediately deleting them. 
This works fine as long as both steps succeed, however when second step fails then the whole transaction is aborted. Leaving behind update keys with new version which do not match expected version when new transaction with same keys is reattempted.

**Problem location**
TableBasedMetadataStore.writeAll 

Do not fail transaction if keys are successfully marked as deleted even though deleting them failed.

sachin-j-joshi:

**Change log description**  
Handle possible failure during deletion of metadata keys can cause version mismatch in next attempt by not failing transaction if keys are successfully marked as deleted even though actually deleting them in same method invocation failed.

**Purpose of the change**  
Fixes #5456

**What the code does**  

Problem - Currently table store has a severe limitation that it can not update and delete keys in the same transaction.
The current work around for this problem is implemented in two steps in TableBasedMetadataStore.writeAll method.
It involves first marking the keys as deleted by updating in first transaction and immediately deleting them.
This works fine as long as both steps succeed, however when second step fails then the whole transaction is aborted. Leaving behind update keys with new version which do not match expected version when new transaction with same keys is reattempted.

Solution - Do not fail transaction if keys are successfully marked as deleted even though actually deleting them in same method invocation failed.

**How to verify it**  
Unit tests should pass.


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStore.java
```diff
@@ -164,12 +164,15 @@ public TableBasedMetadataStore(String tableName, TableStore tableStore, Executor
...
+                                                deletedKeyToTxnDataMap.values().stream().forEach(txnData -> txnData.setDbObject(TableKey.NOT_EXISTS));
+                                            }
+                                            return null;
+                                        }, getExecutor());
                             }, getExecutor())
                             .thenRunAsync(() -> {
-                                for (val deletedKey : keysToDelete) {
-                                    deletedKeyToTxnDataMap.get(deletedKey).setDbObject(TableKey.NOT_EXISTS);
-                                }
                                 TABLE_WRITE_LATENCY.reportSuccessEvent(t.getElapsed());
```
andreipaduroiu, line 176: Can we move this into the `handleAsync` callback above so we don't create a task just to report a metric?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStore.java
```diff
@@ -164,12 +164,15 @@ public TableBasedMetadataStore(String tableName, TableStore tableStore, Executor
...
+                                return this.tableStore.remove(tableName, keysToDelete, timeout)
+                                        .handleAsync((v1, ex) -> {
+                                            if (ex == null) {
+                                                deletedKeyToTxnDataMap.values().stream().forEach(txnData -> txnData.setDbObject(TableKey.NOT_EXISTS));
+                                            }
+                                            return null;
+                                        }, getExecutor());
                             }, getExecutor())
                             .thenRunAsync(() -> {
-                                for (val deletedKey : keysToDelete) {
```
andreipaduroiu, line 170: I don't understand how does moving this for loop from here to the `handleAsync` resolve this problem. `thenRunAsync` is executed only on success, which is exactly what you're doing inside `handleAsync`. 

What am I missing?

andreipaduroiu (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStore.java
sachin-j-joshi: We delete keys in two steps
1. Update them by marking them deleted in the value part of K-V pair. Note that we update all updated keys as well as keys marked deleted in a single transaction. After this step version of all keys changes. Record new version. 
2. Delete the "to be deleted" keys by calling remove().

This change deals when step 1  succeeds and 2 fails.
It used to fail the whole commit, but the version of all keys were changed.
What this commit does is that it basically ignores the failure in step 2.  TableBasedMetadata correctly handles when key is found marked deleted.

RaulGracia (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStore.java
sachin-j-joshi: - This has to be done this way only because that table segment cannot delete and update in the same transaction.
- Now, each SLTS metadata is a key-value pair that has `key -> { version, dbVersion, value, flags}`  structure. The version here is not same as dbVersion because of lazy committing.  However every time we write the table segment the dbVersion is used and updated.
- Before each commit the in memory and cached data is validated to make sure transaction does not use stale version. This is done before this writeAll method is called.
- The way delete is done is by first committing an update that sets the value to null. Then as a second step remove the KV pair. First step is conditional update, while the second step is unconditional delete.
- In case when update succeeds but remove fails , the value is already set to null which is interpreted as "not exists". As far as logic is concerned the key is "marked deleted" and that info is safely persisted already.
- Failure to remove should not fail the transaction. The dbVersion of keys correctly match the version returned by table segment. So if in future this key is again modified and committed we'll not have the version mismatch exception (The original bug that this fix is trying to solve).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStore.java
sachin-j-joshi: simplified flow now.

sachin-j-joshi:
@andreipaduroiu @RaulGracia I have updated the code and responded to your comments.
Thanks

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStore.java
andreipaduroiu: @sachin-j-joshi Let me rewrite your changes using abstract symbols.

Case 1 (Before):
```
A
.thenApply(B)
.thenCompose(C)
.thenRun(D)
```

Case 2 (After):
```
A
.thenCompose(B & C.handle( 
                       (no-error)->D; 
		                error-> ...))
```
		 
Situations:

A fails:
- 1: B, C, D not executed
- 2: B, C, D not executed

A succeeds, B Fails:
- 1: C, D not executed
- 2: C, D not executed

A, B succeed, C fails:
- 1: D not executed
- 2: D not executed (since you have `ex == null`)

A, B, C succeed:
1: D executes
2: D executes

That's why @RaulGracia  and I say we don't understand how this change could possibly have altered the program behavior. You reduced the number of CompletableFutures by combining B and C together, but that doesn't alter the behavior.

andreipaduroiu (approved):

RaulGracia (approved):

PR #5459, opened 2021-01-05
Issue 5458: (SLTS) For read operation, read all chunks in parallel.

### LINKED ISSUE #5458: (SLTS) - ReadOperation should read chunks in parallel when multiple chunks need to be read in a single call.
**Problem description**
Currently ReadOperation reads all the relevant chunks sequentially one after another when multiple chunks need to be read in a single call. This is unnecessary and adds to the overall read latency.

**Problem location**
ReadOperation

All the relevant chunks should be read in parallel.

sachin-j-joshi:

**Change log description**  
 ReadOperation reads all the relevant chunks in parallel when multiple chunks need to be read in a single call.

**Purpose of the change**  
Fixes #5458 

**What the code does**  
 ReadOperation reads all the relevant chunks in parallel when multiple chunks need to be read in a single call.

**How to verify it**  
Existing tests should cover this scenario already.
All existing tests should pass.


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -141,23 +143,56 @@ private void logEnd() {
...
+                         long fromOffset,
+                         int bytesToRead,
+                         int bufferOffset) {
+        val chunkBytesRemaining = new AtomicInteger(bytesToRead);
+        val chunkFromOffset = new AtomicLong(fromOffset);
+        val chunkBufferOffset = new AtomicInteger(bufferOffset);
+        // Note that technically it is possible that read actually request reads less than requested bytes, requiring additional reads on the same chunk.
+        // Hence the for loop below.
+        return chunkedSegmentStorage.getChunkStorage().openRead(chunkName)
+                .thenComposeAsync( chunkHandle ->
```
RaulGracia, line 180: Nit: unnecessary space before `chunkHandle`.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -141,23 +143,56 @@ private void logEnd() {
...
+                            bytesRemaining.addAndGet(-bytesToRead);
+                            currentOffset.addAndGet(bytesToRead);
+                            currentBufferOffset.addAndGet(bytesToRead);
+                            totalBytesRead.addAndGet(bytesToRead);
+                        }, chunkedSegmentStorage.getExecutor());
                     }
                     return CompletableFuture.completedFuture(null);
-                }, chunkedSegmentStorage.getExecutor());
+                }, chunkedSegmentStorage.getExecutor())
+                .thenCompose( v -> Futures.allOf(chunkReadFutures));
```
RaulGracia, line 167: Nit: unnecessary space before `v`.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -141,23 +143,56 @@ private void logEnd() {
...
+                                    bytesToRead,
+                                    currentBufferOffset.get()));
+                            log.trace("{} read - reading chunk - op={}, segment={}, chunk={} offset={} length={} bufferOffset={}",
+                                    chunkedSegmentStorage.getLogPrefix(),
+                                    System.identityHashCode(this), handle.getSegmentName(),
+                                    chunkToReadFrom.getName(),
+                                    currentOffset.get() - startOffsetForCurrentChunk.get(),
+                                    bytesToRead,
+                                    currentBufferOffset.get());
+                            bytesRemaining.addAndGet(-bytesToRead);
```
RaulGracia, line 159: The previous logic only modified all these counters when the read request completed. Instead, your are modifying them immediately (i.e., the process waits for all the requests to complete at then, but these atomic variables would have been already incremented/decremented). This is a change in behavior that may have side effects for the `ReadOperation` state, but I don't see any test verifying that the result is still valid.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -141,23 +143,56 @@ private void logEnd() {
...
+                            currentOffset.addAndGet(bytesToRead);
```
RaulGracia, line 160: Also, even if one of the parallel reads fail, the counters would have been changed anyway. Are you testing this scenario?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -141,23 +143,56 @@ private void logEnd() {
...
-                                                .thenAcceptAsync(bytesRead -> {
-                                                    bytesRemaining.addAndGet(-bytesRead);
-                                                    currentOffset.addAndGet(bytesRead);
-                                                    currentBufferOffset.addAndGet(bytesRead);
-                                                    totalBytesRead.addAndGet(bytesRead);
-                                                }, chunkedSegmentStorage.getExecutor()),
-                                        chunkedSegmentStorage.getExecutor());
+                        return CompletableFuture.runAsync(() -> {
+                            // Create parallel requests to read each chunk.
+                            chunkReadFutures.add(readChunk(chunkToReadFrom.getName(),
```
RaulGracia, line 148: Issuing read requests serially is bad for performance, but it seems harder to overload the cache upon the arrival of many read requests of large chunks. That is, what happens if the size of the cache is smaller than a set of read request that are sent in parallel? Could it be the case that the storage reads keep getting `CacheFullException` because this is just reading in parallel as fast as possible from storage without control? (in a similar way in which we got `CacheFullExceptions` when recovering containers in parallel without controlling the number of parallel recoveries). What about having a number of parallel requests > the storage pool size? Is this going to starve data movements from `StorageWriter` to tier 2?

RaulGracia (changes requested):

RaulGracia (changes requested):
While I understand the potential benefits of the PR, I have some questions/concerns:
- How do we ensure that blindly issuing parallel reads to storage is not going to have side effects, like overwhelming the cache? 
- How do we guarantee that this approach is not going to induce importante interferences in the operation of `StorageWriter` if many parallel reads are getting most threads of the storage pool?
- We need tests to validate the above scenarios in some way, and also that the change in updating the atomic counters (e.g., `currentBufferOffset`) is right, even if one of the storage reads fails.

sachin-j-joshi:
**To summarize the change**
- First of all the upper layers are issuing single read request to read from given `offset` data of size `length` and for that they have already allocated and provided a byte array. That remains as it is. This change does not affect anything related to that.
- With SLTS no append mode, it is more likely that a given range may contain more than one chunk to be read to satisfy the request.
- All we are doing here is that, instead of reading and filling up that byte array sequentially by reading chunks one by one, we are now issuing parallel reads for individual chunks and filling up sections of that array in parallel. 
- The various counters are basically used only  to calculate what chunks to read and at what offsets in buffer data gets copied etc. None of that calculation changes because of parallelizing the actual reads. 
- In terms of failures , for Storage.read() request to succeed all chunk reads must succeed. That requirement has not changed. 
- This change should have no effect on cache. It was the cache that issued that single read request on segment we are just parallelizing inside the execution of that  same request.
- We do want reads to have higher priority and lower latency than reads. Reads are often issues as a direct result of client request. Writes can wait, reads shouldn't. However I do not think reads can starve writes. (We'll have to fix that in case we see that in perf/stress tests.)

**Testing**
- Almost each and every SLTS unit test reads data, so this code is exercised by existing tests.  (and that coverage can be definitely improved. )
image

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: It is the cache that issued the original read request. We are just using parallel reads to fulfill that same original resuest. This change should affect nothing as far as cache is concerned.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: The counters are calculating what chunks to read and at what offset in byte array to copy contents. 
This calculation does not change whether actual reads are parallel or sequential.

sachin-j-joshi: readData returns Futures.allOf(chunkReadFutures)

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: The counters are used only to calculate what chunk to read and at what offset in byte array to copy data. This change does not change that logic. It only parallelizes actual reading of individual chunks.

sachin-j-joshi:
Added some additional tests.

Coverage looks better now
image

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -120,6 +121,7 @@ private void logEnd() {
     }
 
     private CompletableFuture<Void> readData(MetadataTransaction txn) {
+        val chunkReadFutures = new ArrayList<CompletableFuture<Void>>();
```
andreipaduroiu, line 124: This collection is not thread safe yet it is passed around from thread to thread. Even if you're not modifying it concurrently, different threads may not pick up the changes from previous threads. Please use `Collections.synchronizedList` to wrap it.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -120,6 +121,7 @@ private void logEnd() {
...
         return Futures.loop(
                 () -> bytesRemaining.get() > 0 && null != currentChunkName,
```
andreipaduroiu, line 126: Is there an upper bound on the number of parallel reads? We should always put a ceiling on such requests to prevent some corner case to bring down the host application or the remote server.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -141,23 +143,56 @@ private void logEnd() {
...
+                         int bufferOffset) {
+        val chunkBytesRemaining = new AtomicInteger(bytesToRead);
+        val chunkFromOffset = new AtomicLong(fromOffset);
+        val chunkBufferOffset = new AtomicInteger(bufferOffset);
+        // Note that technically it is possible that read actually request reads less than requested bytes, requiring additional reads on the same chunk.
+        // Hence the for loop below.
+        return chunkedSegmentStorage.getChunkStorage().openRead(chunkName)
+                .thenComposeAsync(chunkHandle ->
+                    Futures.loop(
+                            () -> chunkBytesRemaining.get() > 0 && null != currentChunkName,
```
andreipaduroiu, line 182: where is `currentChunkName` defined and/or updated?

andreipaduroiu: Ok, I see it's a class-level variable. Maybe let's add some comment explaining this condition and how the loop will eventually terminate.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -141,23 +143,56 @@ private void logEnd() {
...
+                            () -> chunkedSegmentStorage.getChunkStorage().read(chunkHandle,
+                                    chunkFromOffset.get(),
+                                    chunkBytesRemaining.get(),
+                                    buffer,
+                                    chunkBufferOffset.get())
+                                    .thenAccept(n -> {
+                                        Preconditions.checkState(n != 0, "Zero bytes read");
```
andreipaduroiu, line 189: Add more information to this. If you see it in the logs, you will want to have as much info as you can have in order to debug. There's no chunk, no start/end offset, etc. in this message.

andreipaduroiu (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: Here the number of requests is bound by number of thread pool size. Whether there are 1 chunk reads on 200 segments, or 200 chunk reads on 1 segment, the number of outstanding I/O requests can not exceed thread pool size.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: actually `null != currentChunkName` is not needed here is that condition is already checked before calling this method.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: fixed

RaulGracia (approved):

PR #5471, opened 2021-01-08
Issue 5470: Final check for KVTable deletion may cache incorrect value

### LINKED ISSUE #5470: Final check for KVTable deletion may cache incorrect value
**Problem description**
We have seem the tests `io.pravega.cli.user.kvs.SecureKVTCommandsTest:testDeleteKVT` and `io.pravega.cli.user.kvs.KVTCommandsTest:testDeleteKVT` failing very frequently, specially on some environments. These tests are very simple, as they just create a scope, a table segment and delete that table segment immediately, so we discarded the test logic to be the main cause for this problem.

In our investigation, we have seen the following:
1) The test call for KVTable finishes successfully:
```
2021-01-08T15:36:09.486+0100 [DEBUG] [TestEventLogger]     2021-01-08 15:36:09,485 59894 [controllerpool-15] INFO  i.p.c.s.PravegaTablesStoreHelper - table _system/_tables/deleteKVTable1610116568796/kvt1/metadata.#.b9554d30-695d-4373-9028-da2756264026 deleted successfully
```
The flow seems correct, as all the table segments and metadata is deleted.

2) Then, the last step of the deletion workflow requires a final check that confirms that the table has been actually deleted:
```
return eventHelper.addIndexAndSubmitTask(deleteEvent,
    () -> kvtMetadataStore.setState(scope, kvtName, KVTableState.DELETING, context, executor))
    .thenCompose(x -> eventHelper.checkDone(() -> isDeleted(scope, kvtName, context))) <- Final Check before ack the client
    .thenApply(y -> DeleteKVTableStatus.Status.SUCCESS);
``` 

3) Adding more logs to the code in `PravegaTablesKVTable:getStateData`, we found that this check is left retrying indefinitely and always getting `DELETING` as the state of the KVTable. This leads the controller to hold the ack of the client and eventually the test fails due to timeout excaption:
```
2021-01-08T15:36:09.583+0100 [DEBUG] [TestEventLogger]     2021-01-08 15:36:09,581 59990 [controllerpool-15] INFO  i.p.c.t.K.TableMetadataTasks - TableMetadataTasks:isDeleted deleteKVTable1610116568796 kvt1 DELETING
...
2021-01-08T15:36:09.984+0100 [DEBUG] [TestEventLogger]     2021-01-08 15:36:09,984 60393 [controllerpool-15] INFO  i.p.c.t.K.TableMetadataTasks - TableMetadataTasks:isDeleted deleteKVTable1610116568796 kvt1 DELETING
...
2021-01-08T15:36:25.591+0100 [DEBUG] [TestEventLogger]     2021-01-08 15:36:25,591 76000 [controllerpool-15] INFO  i.p.c.t.K.TableMetadataTasks - TableMetadataTasks:isDeleted deleteKVTable1610116568796 kvt1 DELETING
```

The reason seems to be that the call for this final check (`TableMetadataTasks:isDeleted`) needs to fetch the state of the KVTable and it does it as:
```
kvtMetadataStore.getState(scope, kvtName, false, context, executor)
```
That is, `ignoreCache=false`. This may lead to a situation in which the cached state of the table is `DELETING` and the request never issues a new request to the store to check if the actual state is deleted. Setting `ignoreCache=true` solves the problem and the tests pass consistently.

**Problem location**
`TableMetadataTasks:isDeleted`

Always inspect the state in the store (not the cached value) after the deletion process for the final check in `TableMetadataTasks:isDeleted`.

RaulGracia:
**Change log description**  
Ensure that final deletion check of KVTable deletion fetches information from store.

**Purpose of the change**  
Fixes #5470.

**What the code does**  
In some scenarios, we observed consistent failures on tests that create and subsequently delete a KVTable. The reason is that the final check for the deletion operation was relying exclusively on the already cached data in the Controller. In some situations, such data may be stale and the Controller is stuck retrying getting the stale state from the KVTable. We need to always enforce the final deletion check in `TableMetadataTasks:isDeleted` to fetch the state from the store to prevent this situation.

**How to verify it**  
Previously failing tests are now consistently passing (`io.pravega.cli.user.kvs.SecureKVTCommandsTest:testDeleteKVT` and `io.pravega.cli.user.kvs.KVTCommandsTest:testDeleteKVT`). Added test to check the correct result of `isDeleted()` method. All other tests should work as before.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
@@ -194,7 +194,7 @@ public class TableMetadataTasks implements AutoCloseable {
     }
 
     private CompletableFuture<Boolean> isDeleted(String scope, String kvtName, KVTOperationContext context) {
-        return Futures.exceptionallyExpecting(kvtMetadataStore.getState(scope, kvtName, false, context, executor),
+        return Futures.exceptionallyExpecting(kvtMetadataStore.getState(scope, kvtName, true, context, executor),
                 e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException, KVTableState.UNKNOWN)
                 .thenCompose(state -> {
                     if (state.equals(KVTableState.UNKNOWN)) {
--- a/controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
@@ -193,7 +193,8 @@ public class TableMetadataTasks implements AutoCloseable {
                                                 false, delegationToken, requestId), executor));
     }
 
-    private CompletableFuture<Boolean> isDeleted(String scope, String kvtName, KVTOperationContext context) {
+    @VisibleForTesting
+    CompletableFuture<Boolean> isDeleted(String scope, String kvtName, KVTOperationContext context) {
         return Futures.exceptionallyExpecting(kvtMetadataStore.getState(scope, kvtName, true, context, executor),
                 e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException, KVTableState.UNKNOWN)
                 .thenCompose(state -> {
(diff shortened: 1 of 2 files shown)


shiveshr:
LGTM . can we have a unit test covering it please.

RaulGracia:
@shiveshr added test coverage as requested.

andreipaduroiu (approved):

shiveshr (approved):

PR #5473, opened 2021-01-08
Issue: 5467 updateReaderGroup hangs when changing configuration from subscriber to non-subscriber

### LINKED ISSUE #5467: updateReaderGroup hangs when changing ReaderGroup config from subscriber to non-subscriber
**Problem description**
When updating a ReaderGroup on the controller using controller.updateReaderGroup from a config with subscription to a config with no subscription, the program seems to hang and time out with the following stacktrace
```
java.util.concurrent.TimeoutException
java.util.concurrent.CompletionException: java.util.concurrent.TimeoutException
	at io.pravega.controller.util.RetryHelper.lambda$loopWithTimeout$3(RetryHelper.java:72)
	at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:782)
	at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:753)
	at io.pravega.common.concurrent.Futures.runOrFail(Futures.java:644)
	at io.pravega.common.concurrent.Futures$Loop.run(Futures.java:797)
	at java.base/java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:783)
	at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.util.concurrent.TimeoutException
	... 13 more
```
**Problem location**
Controller

pbelgundi:

**Change log description**  
updateReaderGroup() API requires handling stream subscriptions when a ReaderGroup Configuration has stream changes and/or when the Config changes Retention.Type from a non-subscriber or a subscriber or vice-versa. 
These multiple combinations were not handled correctly leading to the 'updating' flag in ReaderGroupConfigRecord staying "true" and never changing to "false" when a ReaderGroup config is changed from subscriber to non-subscriber.

**Purpose of the change**  
Fixes #5467

**What the code does**  
1. Added code to check for transition from subscriber or non-subscriber or vice versa and handle stream subscriptions accordingly in StreamMetadataTasks.updateReaderGroup and updateReaderGroupTask.execute()
2. If the old and new config both have RetentionType=NONE (non-subscriber RG) then there is no need to make any changes with subscriber metadata in StreamMetadata. 
3. In all other cases we need to add the ReaderGroup as a subscriber to Streams in the new config and deleteSubscriber() from Streams present in old config but not in new config.

**How to verify it**  
Added additional unit and integration tests to verify all possible config update scenarios from Subscriber to NonSubscriber and vice versa, changes in Stream without change to RetentionType, etc..
All unit and integration tests should pass.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateReaderGroupTask.java
+++ b/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateReaderGroupTask.java
@@ -57,6 +57,7 @@ public class UpdateReaderGroupTask implements ReaderGroupTask<UpdateReaderGroupE
         long requestId = request.getRequestId();
         long generation = request.getGeneration();
         UUID readerGroupId = request.getReaderGroupId();
+        boolean isTransition = request.isTransitionToFromSubscriber();
         ImmutableSet<String> streamsToBeUnsubscribed = request.getRemoveStreams();
         final RGOperationContext context = streamMetadataStore.createRGContext(scope, readerGroup);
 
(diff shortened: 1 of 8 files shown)


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateReaderGroupTask.java
```diff
@@ -68,30 +69,39 @@ public UpdateReaderGroupTask(final StreamMetadataTasks streamMetaTasks,
...
-                                                stream.getStreamName(), scopedRGName, rgConfigRecord.getObject().getGeneration(), null, executor);
-                                    }, executor))
-                                    .thenCompose(v -> streamMetadataStore.completeRGConfigUpdate(scope, readerGroup, rgConfigRecord, context, executor));
-                            }
+                           if (rgConfigRecord.getObject().getGeneration() != generation) {
+                               log.warn("Skipping processing of Reader Group update request {} as generation did not match.", requestId);
+                               return CompletableFuture.completedFuture(null);
+                           }
+                           if (rgConfigRecord.getObject().isUpdating()) {
+                               if (isTransition) {
```
shiveshr, line 77: i dont fully follow the logic here.. if it is a transition from non subscriber to subscriber then we do the addition.. so shouldnt we remove it when the transition is from subscriber to non subscriber too?

pbelgundi: The only time isTransition = "false" is when the Retention.Type="NONE" in both the old and new config. In this case there is no need to change anything in StreamMetadata (no subscriber adds/deletes)
In all other cases, isTransition = true and so we add/remove RGs to Stream Metadata as needed.

shiveshr: thats the point -- if transition is from NONE to SUBSCRIPTION then the above logic works fine.. 
but when the transition is from subscription to NONE.. do you still end up adding the RG for subscription to streams that are present in the config or are all streams unsubscribed from this RG?

pbelgundi: The if check at ln # 89 guards against that. 
So it works both ways. I have added tests for all these transitions in StreamMetadataTasksTest.

shiveshr (approved):

andreipaduroiu (approved):

PR #5477, opened 2021-01-09
Issue 5476: ReaderGroupConfiguration to ReaderGroupConfig conversion support for UNBOUNDED StreamCuts

### LINKED ISSUE #5476: ReaderGroupConfiguration to ReaderGroupConfig conversion support for UNBOUNDED StreamCuts
**Problem description**
When encoding a `ReaderGroupConfiguration` on the controller end to the client side `ReaderGroupConfig`, the case of `StreamCut.UNBOUNDED` is not taken care of. So these cases lead to an `IllegalArgumentException` when building the `ReaderGroupConfig` as the `StreamCut` is not valid.

**Problem location**
On the controller end in the `ModelHelper` and `CreateReaderGroupTask`

Add support for `UNBOUNDED` stream-cut cases.

anirudhkovuru:
**Change log description**  
Added a check to identify `StreamCut.UNBOUNDED` cases and deal with them separately.

**Purpose of the change**  
Fixes #5476 

**What the code does**  
Currently when a `ReaderGroupConfig` is converted into a controller side `ReaderGroupConfiguration`, the UNBOUNDED cases are stored as empty lists.
In both the `ModelHelper` and `CreateReaderGroupTask`, when an empty stream-cut list is found, we store it as an `UNBOUNDED` instead.

**How to verify it**  
All tests should pass.

### DIFF
--- a/client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
+++ b/client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
@@ -271,26 +271,19 @@ public final class ModelHelper {
                 .generation(rgConfig.getGeneration())
                 .readerGroupId(UUID.fromString(rgConfig.getReaderGroupId()))
                 .startingStreamCuts(rgConfig.getStartingStreamCutsList().stream()
-                        .collect(Collectors.toMap(streamCut -> Stream.of(streamCut.getStreamInfo().getScope(),
-                                streamCut.getStreamInfo().getStream()),
-                                streamCut -> {
-                                    if (streamCut.getCutMap().isEmpty()) {
-                                        return io.pravega.client.stream.StreamCut.UNBOUNDED;
-                                    }
-                                    return new StreamCutImpl(Stream.of(streamCut.getStreamInfo().getScope(),
-                                            streamCut.getStreamInfo().getStream()), getSegmentOffsetMap(streamCut.getStreamInfo().getScope(), streamCut.getStreamInfo().getStream(), streamCut.getCutMap()));
-                                })))
+                        .collect(Collectors.toMap(streamCut -> Stream.of(streamCut.getStreamInfo().getScope(), streamCut.getStreamInfo().getStream()),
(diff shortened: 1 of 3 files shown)


pbelgundi (changes requested):
Please add tests to cover the new code.

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
```diff
@@ -271,15 +271,25 @@ public static ReaderGroupConfig encode(Controller.ReaderGroupConfiguration rgCon
                 .generation(rgConfig.getGeneration())
                 .readerGroupId(UUID.fromString(rgConfig.getReaderGroupId()))
                 .startingStreamCuts(rgConfig.getStartingStreamCutsList().stream()
-                                    .collect(Collectors.toMap(streamCut -> Stream.of(streamCut.getStreamInfo().getScope(),
-                                    streamCut.getStreamInfo().getStream()),
-                                    streamCut -> new StreamCutImpl(Stream.of(streamCut.getStreamInfo().getScope(),
-                                    streamCut.getStreamInfo().getStream()), getSegmentOffsetMap(streamCut.getStreamInfo().getScope(), streamCut.getStreamInfo().getStream(), streamCut.getCutMap())))))
+                        .collect(Collectors.toMap(streamCut -> Stream.of(streamCut.getStreamInfo().getScope(),
+                                streamCut.getStreamInfo().getStream()),
+                                streamCut -> {
```
pbelgundi, line 276: It would be good to encapsulate this lambda in a method and invoke that method here, so we can reuse code instead of repeat, same method can also be invoked in CreateReaderGroupTask.

pbelgundi (changes requested):

pbelgundi (approved):
LGTM.

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
anirudhkovuru: The lambda has been encapsulated into a method which is used in both classes.

anirudhkovuru:
> Please add tests to cover the new code.

I have added a test for the new code.

PR #5479, opened 2021-01-09
Issue 5478: Performance regression due to wrong logging on metrics

### LINKED ISSUE #5478: Performance regression due to wrong logging on metrics
**Problem description**
As part of PR #5422, a new `warn` log message was introduced to detect that an already closed metric was still being used to report values: https://github.com/pravega/pravega/blob/8ca045cf0118a4755efe0cf74e3ad2659164a2b8/shared/metrics/src/main/java/io/pravega/shared/metrics/MetricProxy.java#L83

However, it looks like that it introduces a huge performance penalty. The main reasons could be the following:
- The method is in the hot path
- The logic seems incorrect, as the log happens `if (!closed.get())`, but it should be actually logging in the opposite case. So, basically, this is logging virtually always that `getInstance()` is called.
- The `if` condition just for logging may be called concurrently, but it uses a synchronized object to evaluate whether or not to write the log message. This may add some additional overhead. 
- A `warn` level may be too high for the kind of message being logged.

We tried removing this `if` condition and performance comes back to normal.

**Problem location**
`MetricsProxy`

Remove the `if` block from `getInstace()` method to favor performance.

RaulGracia:
**Change log description**  
Removed logging message that was called upon an incorrect condition.

**Purpose of the change**  
Fixes #5478.

**What the code does**  
As described in #5478, PR #5422 introduced a logging message to warn about the use of a metric after being closed. The main problem is that the logic to write that log message is not correct, which makes the system to log many useless warnings.

While we could reverse the logic condition that logs that message, we propose to completely remove it given that it is called in the hot path (e.g., metrics for number of messages) in order to favor performance.

**How to verify it**  
Just removed a log message, no actual code changed. All tests should be passing as before. We have validated that performance now is similar to what it was in previous commits (e.g., commit 2703).

### DIFF
--- a/shared/metrics/src/main/java/io/pravega/shared/metrics/MetricProxy.java
+++ b/shared/metrics/src/main/java/io/pravega/shared/metrics/MetricProxy.java
@@ -79,9 +79,6 @@ abstract class MetricProxy<T extends Metric> implements AutoCloseable {
     }
 
     protected T getInstance() {
-        if (!closed.get()) {
-            log.warn("This MetricProxy ({}) has already been closed. Further updates to this Metric will not be seen by a MeterRegistry.", proxyName);
-        }
         return this.instance.get();
     }
 }


shared/metrics/src/main/java/io/pravega/shared/metrics/MetricProxy.java
```diff
@@ -79,9 +79,6 @@ void updateInstance(T instance) {
     }
 
     protected T getInstance() {
-        if (!closed.get()) {
```
andreipaduroiu, line 82: I think it would have sufficed to simply fix the condition here. But I agree that probably nobody will look for these warnings in the logs anyway so we might as well get rid of it completely.

andreipaduroiu (approved):

PR #5503, opened 2021-01-10
Issue 5474: Handle NPE in RawClient ReplyProcessor

### LINKED ISSUE #5474: ReplyProcessor NPE
**Problem description**
During testing, `NullPointerException` happened in the `RawClient::ReplyProcessor::connectionDropped`.

It's a race condition caused by `ClientSocketReader` thread. What happened is that during the initialization of `RawClient`, `ClientSocketReader` has an error and is going to shut down and it attempts to close all the Flows on that TCP connection. However at this point, the `FlowId` of this `RawClient` has already been registered to the `FlowHandler` but the creation of `ClientConnection` is not finished. The shundown process of the `FlowHandler` will simply call the callback functions of all registered `FlowId` and that's how `RawClient::ReplyProcessor` is called. And since the `ClientConnection` hasn't been fully created, the `NullPointerException` happened.

**Problem location**
RawClient

RawClient, FlowHandler

Tristan1900:
**Change log description**  
When NPE happens in RawClient ReplyProcessor, we just log the info and return. 
This is safe because anyone who uses this RawClient instance will have `ConnectionFailedException` since the underlying TCP connection is closed.

**Purpose of the change**  
FIx #5474

**What the code does**  
Logs and returns when NPE occurs

**How to verify it**


RaulGracia:
@Tristan1900 please have a look to the codecov report, you may need to add some tests to the new code.

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -135,6 +135,12 @@ private void closeConnection(Throwable exceptionToInflightRequests) {
             log.warn("Closing connection to segment {} with exception", segmentId, exceptionToInflightRequests);
         }
         if (closed.compareAndSet(false, true)) {
+            if (connection == null) {
```
shrids, line 138: `private final CompletableFuture<ClientConnection> connection;`
 I do not see how `connection` can be set to null.
@Tristan1900 could you elaborate here please ?

Tristan1900: sure. It's because the [closeConnection](https://github.com/pravega/pravega/blob/055494a2e2dabfb649e5b7e3c9c4f586dd18940e/client/src/main/java/io/pravega/client/connection/impl/RawClient.java#L130) method is called during the initialization of [connection](https://github.com/pravega/pravega/blob/master/client/src/main/java/io/pravega/client/connection/impl/RawClient.java#L101). The creation of connection requires to register the `Flow` to a `FlowHandler` at [here](https://github.com/pravega/pravega/blob/055494a2e2dabfb649e5b7e3c9c4f586dd18940e/client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java#L88) and that's where the race condition happens. The `Flow` has been registered but the method is not finished so `ClientConnection` does not return and so `RawClient#Connection` is still `Null`. However, at this point, `ClientSocketReader` throws an exception and closes the TcpConnection and the `FlowHandler` is forced to shut down. All registered `Flow` on that FlowHandler is required to shut down and the corresponding callbacks are called, so the `RawClient#closeConnection` is called even when `RawClient#Connection` is not created and that's why `NPE` is thrown.
Hope that makes sense

tkaitchuck: I am not sure we don't have a threading issue here. 
The fact that closeConnection can be called before connection is set means we are working with a partially initialized object, and the fact that a future is involved means we aren't guaranteed this is one thread. 
I think the correct solution is to first assign the future for connection and pass the future into the connect method for it to complete rather than having it return the value. Then there is are no assignments happening, and we don't have to deal with the null case here.

Tristan1900: Thanks @tkaitchuck, that's a good way to do. I have made the change according to that, could you please take a look again to see if it is what you mean? Thanks!

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -98,23 +98,25 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
 
     public RawClient(PravegaNodeUri uri, ConnectionPool connectionPool) {
         this.segmentId = null;
-        this.connection = connectionPool.getClientConnection(flow, uri, responseProcessor)
-        .exceptionally(e -> {
+        this.connection = new CompletableFuture<>();
+        connectionPool.getClientConnection(flow, uri, responseProcessor, this.connection);
+        this.connection.exceptionally(e -> {
```
shrids, line 103: This is different from previous code since `this.connection` will not be completed with `ConnectionFailedException(e)`

```
//previous snip.
this.connection = connectionPool.getClientConnection(flow, uri, responseProcessor)
        this.connection = new CompletableFuture<>();
        .exceptionally(e -> {
```

shrids: The CompletableFuture will not be failed with ConnectionFailedException and it is a change in behaviour for the retry logic

shrids (changes requested):

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -98,23 +98,25 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
-        Futures.exceptionListener(connection, e -> closeConnection(e));
+        Futures.exceptionListener(connection, this::closeConnection);
     }
 
     public RawClient(Controller controller, ConnectionPool connectionPool, Segment segmentId) {
         this.segmentId = segmentId;
-        this.connection = controller.getEndpointForSegment(segmentId.getScopedName())
-                                    .thenCompose((PravegaNodeUri uri) -> connectionPool.getClientConnection(flow, uri, responseProcessor))
+        this.connection = new CompletableFuture<>();
+        controller.getEndpointForSegment(segmentId.getScopedName())
```
shrids, line 113: if controller.getEndpointForSegment(..) complets exceptionally then a ConnectionFailedException will not be thrown...

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
Tristan1900: Thanks! That's a good point!

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -98,23 +98,25 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
-        this.connection = connectionPool.getClientConnection(flow, uri, responseProcessor)
-        .exceptionally(e -> {
+        this.connection = new CompletableFuture<>();
+        connectionPool.getClientConnection(flow, uri, responseProcessor, this.connection);
+        this.connection.exceptionally(e -> {
             log.warn("Exception observed while attempting to obtain a connection to segment store {}", uri, e);
             throw new CompletionException(new ConnectionFailedException(e));
         });
-        Futures.exceptionListener(connection, e -> closeConnection(e));
+        Futures.exceptionListener(connection, this::closeConnection);
```
tkaitchuck, line 107: Shouldn't this and the `exceptionally` be moved above the `getClientConnection` call now that that is possible?

client/src/main/java/io/pravega/client/connection/impl/RawClient.java
```diff
@@ -98,23 +98,25 @@ public void authTokenCheckFailed(WireCommands.AuthTokenCheckFailed authTokenChec
...
-                                    .thenCompose((PravegaNodeUri uri) -> connectionPool.getClientConnection(flow, uri, responseProcessor))
+        this.connection = new CompletableFuture<>();
+        controller.getEndpointForSegment(segmentId.getScopedName())
+                                    .thenAccept((PravegaNodeUri uri) -> connectionPool.getClientConnection(flow, uri, responseProcessor, this.connection))
                                     .exceptionally(e -> {
                                         log.warn("Exception observed while attempting to obtain a connection to segment {}", segmentId, e);
                                         throw new CompletionException(new ConnectionFailedException(e));
                                     });
-        Futures.exceptionListener(connection, e -> closeConnection(e));
+        Futures.exceptionListener(connection, this::closeConnection);
```
tkaitchuck, line 119: Same here.

shrids:
Test `RawClientTest.testExceptionWhileObtainingConnection` is preventing the build from completing here.

tkaitchuck (approved):

common/src/main/java/io/pravega/common/concurrent/Futures.java
```diff
@@ -274,6 +274,7 @@
     public static <T> void exceptionListener(CompletableFuture<T> completableFuture, Consumer<Throwable> exceptionListener) {
         completableFuture.whenComplete((r, ex) -> {
             if (ex != null) {
+                System.out.println(ex);
```
shrids, line 277: this can be removed...

shrids (changes requested):
LGTM apart from one comment.

shrids (approved):

PR #5508, opened 2021-01-11
Issue 5506: Read/Read-Update permissions for the ReaderGroup operations on controller

### LINKED ISSUE #5506: Read/Read-Update permissions for the ReaderGroup operations on controller
**Problem description**
In the `ControllerServiceImpl`, the permissions for operations like `createReaderGroup`, `updateReaderGroup` and `deleteReaderGroup` is `READ_UPDATE` on scope. However for these operations, that is not required as `ReaderGroup`s can be created in scope even without the `READ_UPDATE` permission. This is so that users with only read permissions on stream can still make a `ReaderGroup` and use it. 

**Problem location**
In the `ControllerServiceImpl` class on the controller.

Use the correct permissions for these operations in the `ControllerServiceImpl`.

anirudhkovuru:
Signed-off-by: anirudhkovuru <anirudhkovuru@gmail.com>

**Change log description**  
Added changes to the `ControllerServiceImpl` to ensure the correct permissions are provided when performing `ReaderGroup` operations on the controller.

**Purpose of the change**  
Fixes #5506 

**What the code does**  
Using the `isRGStreamWritesWithReadPermEnabled` flag to ensure the required permissions are correctly imparted.

**How to verify it**  
All tests should pass.


shiveshr (approved):

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -148,8 +148,11 @@ public void createReaderGroup(ReaderGroupConfiguration request, StreamObserver<C
         RequestTag requestTag = requestTracker.initializeAndTrackRequestTag(requestIdGenerator.get(), "createReaderGroup",
                 scope, rgName);
         log.info(requestTag.getRequestId(), "createReaderGroup called for ReaderGroup {}/{}.", scope, rgName);
+        AuthHandler.Permissions requiredPermission = this.isRGStreamWritesWithReadPermEnabled ?
```
ravisharda, line 151: Elsewhere in this class, we have used `StreamAuthParams` to abstract out these details. You might want to use an object of that class to delegate this auth-specific business logic to that abstraction. 

See an example here: https://github.com/pravega/pravega/blob/466d71a06cfb0b37f9163a52073b45f050c91b77/controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java#L305

ravisharda: @anirudhkovuru later found out the `StreamAuthParams` won't work in this case, which only currently handles reader groups coming as internal streams.

ravisharda (approved):

PR #5520, opened 2021-01-14
Issue 5518: (SLTS) Fix Integer overflow.

### LINKED ISSUE #5518: SLTS - ArithmeticException in ReadOperation
**Problem description**

Encountered ArithmeticException in ReadOperation

```
2021-01-13 00:47:19,248 26611977 [storage-io-126] ERROR i.p.s.s.h.h.PravegaRequestProcessor - [requestId=1821066133504] Error (Segment = 'longevity/mediumScale/260.#epoch.161', Operation = 'readSegment')
java.lang.ArithmeticException: integer overflow
        at java.base/java.lang.Math.toIntExact(Unknown Source)
        at io.pravega.segmentstore.storage.chunklayer.ReadOperation.lambda$readData$8(ReadOperation.java:131)
        at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:784)
        at io.pravega.common.concurrent.Futures$Loop.call(Futures.java:753)
        at io.pravega.common.concurrent.Futures.runOrFail(Futures.java:644)
        at io.pravega.common.concurrent.Futures$Loop.run(Futures.java:797)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
        at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
```

The length of the chunk is `3661818825` which is greater than Integer.MAX_VALUE

```
2021-01-13 00:47:19,247 26611976 [storage-io-126] DEBUG i.p.s.s.chunklayer.ReadOperation - ChunkedSegmentStorage[1] read - found chunk to read - op=493405335, segment=longevity/mediumScale/260.#epoch.161,
 chunk=ChunkMetadata(name=longevity/mediumScale/260.#epoch.161.E-2-O-0.a65b9bd9-5593-4396-a717-dcdcdb55bbd5, length=3661818825, nextChunk=null, status=1), startOffset=0, length=3661818825, readOffset=0.
```

**Problem location**
``` java
bytesToRead = Math.min(bytesRemaining.get(), Math.toIntExact(chunkToReadFrom.getLength() - (currentOffset.get() - startOffsetForCurrentChunk.get())));
```
fix it.

sachin-j-joshi:

**Change log description**  
Fix integer overflow errors in ReadOperation and DefragmentOperation.

**Purpose of the change**  
Fixes #5518

**What the code does**  
Fixes integer overflow errors in ReadOperation and DefragmentOperation.

**How to verify it**  
All old and new unit tests must pass.


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -293,7 +293,7 @@
         return Futures.loop(
                 () -> bytesToRead.get() > 0,
                 () -> {
-                    val buffer = new byte[Math.min(chunkedSegmentStorage.getConfig().getMaxBufferSizeForChunkDataTransfer(), bytesToRead.get())];
+                    val buffer = new byte[Math.toIntExact(Math.min(chunkedSegmentStorage.getConfig().getMaxBufferSizeForChunkDataTransfer(), bytesToRead.get()))];
```
RaulGracia, line 296: This is inherently assuming that at least `chunkedSegmentStorage.getConfig().getMaxBufferSizeForChunkDataTransfer()` should be lower than `Integer.MAX_VALUE`. Is the config of `chunkedSegmentStorage` checking this? If not, we need to add this check, or otherwise, we may have still problems because of values being too large.

RaulGracia (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
sachin-j-joshi: yes getMaxBufferSizeForChunkDataTransfer() returns integer.

RaulGracia: Ok, but can you please add a precondition check like in `ChunkedSegmentStorageConfig` to ensure that `getMaxBufferSizeForChunkDataTransfer()` is strictly lower than `Integer.MAX_SIZE` with a message that explains the check failure in the case the user sets that value to something larger? Actually, it would be interesting to also provide a lower bound to this check in `ChunkedSegmentStorageConfig`, because a user could inadvertently set it to 0 or a negative value, which would crash when instiantiating that `buffer`.

RaulGracia (changes requested):
Please, add proper validation of input values for `maxBufferSizeForChunkDataTransfer` in `ChunkedSegmentStorageConfig` to ensure that it is between a reasonable minimum lower bound (i.e., at least 1) and lower than `Integer.MAX_VALUE`.

sachin-j-joshi:
Realized that ChunkedSegmentStorageConfig in general does not throw on negative values.
Instead of fixing unrelated configs etc in this PR I'll raise a new one to take care of all applicable ChunkedSegmentStorageConfig fields and update comments in config file etc. 
https://github.com/pravega/pravega/issues/5536

sachin-j-joshi:
https://github.com/pravega/pravega/pull/5539 for validation.

RaulGracia (approved):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
```diff
@@ -280,7 +280,7 @@
                 () -> {
                     readAtOffset.set(0);
                     val arg = concatArgs[currentArgIndex.get()];
-                    bytesToRead.set(Math.toIntExact(arg.getLength()));
+                    bytesToRead.set(arg.getLength());
```
andreipaduroiu, line 283: Why would we ever want to read more than 2GB at once? We don't have enough memory to store all that data.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStoreTests.java
```diff
@@ -86,6 +86,14 @@ public TestContext getTestContext() throws Exception {
             return new TableBasedMetadataTestContext(executorService());
         }
 
+        @Override
+        public void testReadHugeChunks() {
```
andreipaduroiu, line 90: Why are these empty? If intentional, please add a comment (inside these tests) indicating so and explaining why.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/mocks/InMemorySimpleStorageTests.java
```diff
@@ -76,6 +76,14 @@ protected void checkData(byte[] expected, byte[] output) {
             Assert.assertArrayEquals(expected, output);
         }
 
+        @Override
```
andreipaduroiu, line 79: Same here

andreipaduroiu (changes requested):

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/metadata/TableBasedMetadataStoreTests.java
sachin-j-joshi: These are in memory providers and the test creates chunks that are multiple of Integet.Max_VALUE in length.  
There is no need to allocate and actually test this with in-memory chunk provider.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/mocks/InMemorySimpleStorageTests.java
sachin-j-joshi: same as above

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/DefragmentOperation.java
sachin-j-joshi: No we don't. 
It is the minimum of MaxBufferSizeForChunkDataTransfer and size. 


                    val buffer = new byte[Math.toIntExact(Math.min(chunkedSegmentStorage.getConfig().getMaxBufferSizeForChunkDataTransfer(), bytesToRead.get()))];

andreipaduroiu: So why is `bytesToRead` an `AtomicLong` ? What was wrong with `AtomicInt`?

sachin-j-joshi: The chunk has long length and it may happen it actually is beyond Integer.Max
Casting it to/trying to fit into Int will cause overflow.
Note this is called from concatUsingAppend method.
In case of very long chunk we'll have to copy bytes from it in multiple batches.
After each read we update `bytesToRead` and `readAtOffset `
```
       bytesToRead.addAndGet(-size);
       readAtOffset.addAndGet(size);
```
making bytesToRead AtomicLong is the correct and minimal fix.

Math.toIntExact(Math.min(long, long)) will take care of it because getMaxBufferSizeForChunkDataTransfer is int

andreipaduroiu (approved):

PR #5521, opened 2021-01-14
Issue 5519: (SegmentStore) SegmentAttributeIndex bug fixes.

### LINKED ISSUE #5519: SegmentAttributeBTreeIndex potential data loss
**Problem description**
Storing Segment Attributes in the LTS B+Tree works in the following steps:
1. Read BTree Pages for affected keys.
2. Update those pages in memory
3. Write affected pages (including entire path to the root) to LTS segment and include a footer that points to the root page.
4. Truncate LTS segment as necessary in order to do auto-compaction.
5. Update a Core Attribute on the Segment to point to the footer in LTS.

If any failure (for whatever reason) happens up to and including step 4, there is no issue. However, if we get a failure after step 4 but before step 5 completes successfully, then the following can happen:
1. Upon a recovery, the BTreeIndex is reinitialized by having it read the footer using the Core Attribute on the Segment (that is updated on step 5 above). 
2. If this value is correct (i.e., step 5 completed successfully), all is good.
3. If this value is incorrect (step 5 failed), then an obsolete value will be passed to the BTreeIndex which can point it (directly or indirectly) to offsets within the segment that have already been truncated out. 

At this time, there is no recovery from this situation.

**Problem location**
SegmentAttributeBTreeIndex.java or related.

TBD.

### LINKED ISSUE #5575: Attribute Index Bug Fixes
**Problem description**
1. `SegmentAttributeBTreeIndex.storeInCache` if it fails (due to cache full or whatever), it should not bubble up and interrupt the upstream call. It should log a warn and move on.
2. SegmentContainer does not close AttributeIndex when shutting down.
3. `SegmentAttributeBTreeIndex.updateGenerations/removeFromCache` may pass `-1` as address. This happens if the entry has already been evicted. This leads to log pollution.

**Problem location**
Attribute Index

Fix them.

andreipaduroiu:
**Change log description**  
- Updated SegmentAttributeBTreeIndex to recover a BTreeIndex using solely the information encoded in the index file if the LTS adapter used supports atomic writes.
- Fixed a bug in SegmentAttributeBTreeIndex where it would corrupt in-memory state (and require a process restart) if unable to store entries in the cache after an update.
- Improved SegmentAttributeBTreeIndex to handle cache insertion exceptions (do not bubble up and log accordingly).
- Fixed a bug in StreamSegmentContainer where it would not be closing the Attribute Indices upon shutdown.

**Purpose of the change**  
Fixes #5519.
Fixes #5575.

**What the code does**  
- Root Pointers
    - The Segment Attribute Index (in the format of a BTreeIndex) is highly sensitive to the format of the data written within. Every write to LTS that it makes must be executed atomically by the underlying LTS implementation; partially written data (in the event of a crash) will result in index corruption. #4451 introduced a work-around for then-default `RollingStorage` which did not guarantee atomic writes (especially for FileSystem implementations). 
    - However, as described by #5519, that approach has a major flaw, which can cause more harm than good in certain failure situations. 
    - This PR makes use of the SLTS (`ChunkedSegmentStorage`) guarantees that all writes are atomic, regardless of the capabilities of the underlying storage adapter. It adds a flag (`supportsAtomicWrites`) which is false for `RollingStorage`/`AsyncStorageWrapper` and true for `ChunkedSegmentStorage`. If the flag is indeed set, it will begin recovery in the traditional way, by reading the "footer" as the last 12 bytes of the index file. If the flag is false, then it will execute as before.
    - This fix does not address the problem if `RollingStorage` is the default LTS implementation (as we are aiming to deprecate and ultimately retire that one). It fixes it only for SLTS.
- SegmentAttributeBTreeIndex (other bugs):
    - The state of the index includes the current length of the attribute segment, which is used for length-conditional writes. This state is only updated after successful writes.
    - However, it is possible that an LTS write + truncation occurred, but the rest of the update operation failed due to an inability to store data in the cache. This would leave the state in an inconsistent state.
    - The state would eventually be refreshed/reinitialized upon a new update. However, if the cache is full (which is likely the root cause of the above exceptions), any update would require a read, which would likely try to read from truncated offsets, hence corrupted in-memory state.
    - The fix for this is to handle exceptions coming from `storeInCache` and log them appropriately and perform any necessary index cleanup. It is OK if we can't cache data, as we can always read it later from LTS.

**How to verify it**  
Build shall pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
@@ -279,8 +279,7 @@ class SegmentAttributeBTreeIndex implements AttributeIndex, CacheManager.Client,
         }
 
         Collection<PageEntry> entries = values.entrySet().stream().map(this::serialize).collect(Collectors.toList());
-        return executeConditionally(tm -> this.index.update(entries, tm), timeout)
-                .thenApply(this::maybeIgnoreRootPointer);
+        return executeConditionally(tm -> this.index.update(entries, tm), timeout);
     }
 
     @Override
@@ -481,12 +480,20 @@ class SegmentAttributeBTreeIndex implements AttributeIndex, CacheManager.Client,
         return this.storage.getStreamSegmentInfo(handle.getSegmentName(), timeout)
                 .thenApply(segmentInfo -> {
                     // Get the root pointer from the Segment's Core Attributes.
-                    long rootPointer = getRootPointer(segmentInfo);
+                    long rootPointer = getRootPointerIfNeeded(segmentInfo);
                     return new BTreeIndex.IndexInfo(segmentInfo.getLength(), rootPointer);
                 });
     }
 
-    private long getRootPointer(SegmentProperties segmentInfo) {
+    /**
+     * Extracts the {@link Attributes#ATTRIBUTE_SEGMENT_ROOT_POINTER} from the given {@link SegmentProperties} if necessary.
+     * If {@link Storage#supportsAtomicWrites()} is true for {@link #storage}, then a negative value is returned and
+     * the information from the given {@link SegmentProperties} is ignored.
+     *
+     * @param segmentInfo The {@link SegmentProperties} to extract from.
+     * @return The extracted root pointer or a negative value if not needed.
+     */
+    private long getRootPointerIfNeeded(SegmentProperties segmentInfo) {
         // Get the root pointer from the Segment's Core Attributes.
         long rootPointer = BTreeIndex.IndexInfo.EMPTY.getRootPointer(); // -1;
         if (this.storage.supportsAtomicWrites()) {
@@ -495,7 +502,7 @@ class SegmentAttributeBTreeIndex implements AttributeIndex, CacheManager.Client,
             return rootPointer;
         }
 
-        rootPointer = this.segmentMetadata.getAttributes().getOrDefault(Attributes.ATTRIBUTE_SEGMENT_ROOT_POINTER, BTreeIndex.IndexInfo.EMPTY.getRootPointer());
+        rootPointer = this.segmentMetadata.getAttributes().getOrDefault(Attributes.ATTRIBUTE_SEGMENT_ROOT_POINTER, rootPointer);
         if (rootPointer != BTreeIndex.IndexInfo.EMPTY.getRootPointer() && rootPointer < segmentInfo.getStartOffset()) {
             // The Root Pointer is invalid as it points to an offset prior to the Attribute Segment's Start Offset.
             // The Attribute Segment is updated in 3 sequential steps: 1) Write new BTree pages, 2) Truncate and
@@ -508,17 +515,6 @@ class SegmentAttributeBTreeIndex implements AttributeIndex, CacheManager.Client,
             rootPointer = BTreeIndex.IndexInfo.EMPTY.getRootPointer();
         }
         return rootPointer;
-    }
-
-    /**
-     * If {@link #storage} supports atomic writes ({@link Storage#supportsAtomicWrites()} is true), then this method
-     * returns a negative value. Otherwise it returns the given parameter.
-     *
-     * @param rootPointer The actual root pointer.
-     * @return The {@code rootPointer} if atomic writes are supported, or a negative value otherwise.
-     */
-    private long maybeIgnoreRootPointer(long rootPointer) {
-        return this.storage.supportsAtomicWrites() ? BTreeIndex.IndexInfo.EMPTY.getRootPointer() : rootPointer;
     }
 
     private CompletableFuture<ByteArraySegment> readPage(long offset, int length, Duration timeout) {
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
@@ -24,6 +24,7 @@ import io.pravega.common.util.btree.BTreeIndex;
 import io.pravega.common.util.btree.PageEntry;
 import io.pravega.segmentstore.contracts.Attributes;
 import io.pravega.segmentstore.contracts.BadOffsetException;
+import io.pravega.segmentstore.contracts.SegmentProperties;
 import io.pravega.segmentstore.contracts.StreamSegmentExistsException;
 import io.pravega.segmentstore.contracts.StreamSegmentNotExistsException;
 import io.pravega.segmentstore.contracts.StreamSegmentSealedException;
@@ -278,7 +279,8 @@ class SegmentAttributeBTreeIndex implements AttributeIndex, CacheManager.Client,
         }
 
         Collection<PageEntry> entries = values.entrySet().stream().map(this::serialize).collect(Collectors.toList());
-        return executeConditionally(tm -> this.index.update(entries, tm), timeout);
+        return executeConditionally(tm -> this.index.update(entries, tm), timeout)
+                .thenApply(this::maybeIgnoreRootPointer);
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/Storage.java
```diff
@@ -172,8 +171,21 @@
      */
     boolean supportsTruncation();
 
+    /**
+     * Determines whether this Storage implementation supports atomic writes. Supporting atomic writes means that calls
+     * to {@link #write} will either make all the payload available for reading (via {@link #getStreamSegmentInfo} and
+     * {@link #read}) or none, regardless of whether the underlying Storage binding supports that (i.e., FileSystem does not).
```
sachin-j-joshi, line 177: Nit: 
binding supports that (E.g., FileSystem does not)

segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
```diff
@@ -761,7 +907,12 @@ public void testTruncatedRootPointer() {
             expectedValues.put(attributeId, value);
             val updateBatch = Collections.singletonMap(attributeId, value);
 
+            // Inject a callback to knwo when we finished truncating, as it is an async task.
```
sachin-j-joshi, line 910: Nit: knwo

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
```diff
@@ -450,9 +458,17 @@ synchronized void update(UUID keyHash, long segmentOffset, int currentGeneration
...
-            storeInCache(entryData);
-            this.generation = currentGeneration;
-            this.highestOffset = Math.max(this.highestOffset, segmentOffset);
+            try {
+                storeInCache(entryData);
+                this.generation = currentGeneration;
+                this.highestOffset = Math.max(this.highestOffset, segmentOffset);
+            } catch (CacheEntryEvictedException cex) {
+                // Pass-through exception. If this is the case, the entry has already been evicted so not more we can do.
+                throw cex;
```
sachin-j-joshi, line 467: Not sure I understood why we need to throw here.
Entry doesn't exist as expected. So this should be ok. right?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
```diff
@@ -671,29 +704,47 @@ private void storeInCache(List<Map.Entry<Long, ByteArraySegment>> toAdd, Collect
...
+        }
+
+        // Insert/Update the cache. There is a chance this won't go through, so be able to handle that. It's not the end
+        // of the world if we can't update the cache (as we always have the data in the attribute segment). If we are
+        // having trouble, log the error and move on.
+        int entryAddress = entry == null ? CacheEntry.NO_ADDRESS : entry.getCacheAddress();
+        try {
+            if (entry != null && entry.isStored()) {
+                // Entry exists. Replace it.
+                entryAddress = this.cacheStorage.replace(entry.getCacheAddress(), data);
```
sachin-j-joshi, line 727: Codecov complains this line is not covered.
Seems like an important case to cover.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
```diff
@@ -671,29 +704,47 @@ private void storeInCache(List<Map.Entry<Long, ByteArraySegment>> toAdd, Collect
...
     }
 
     @GuardedBy("cacheEntries")
-    private void storeInCache(CacheEntry entry, ByteArraySegment data) {
-        int newAddress;
+    private void storeInCache(long entryOffset, ByteArraySegment data) {
+        CacheEntry entry = this.cacheEntries.getOrDefault(entryOffset, null);
+        if (entry != null && entry.getSize() == data.getLength()) {
+            // Already cached.
+            return;
```
sachin-j-joshi, line 717: Codecov complains this line is not covered.
Seems like an important case to cover.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
```diff
@@ -672,18 +730,106 @@ public void testIndexCorruption() {
...
+        Assert.assertEquals("Unexpected value read.", value2, (long) actualValue.get(attribute));
+        Assert.assertTrue("Expecting storage reads after reload.", storageRead.get());
+    }
+
+    /**
+     * Tests the behavior of the Attribute Index when it is told that the underlying Storage adapter supports atomic
+     * writes but in fact it does not. It is expected to fail with corrupted data.
+     */
+    @Test
+    public void testRecoveryAfterIncompleteUpdateWithNoRootPointer() {
```
sachin-j-joshi, line 778: I'm not sure I understood how we simulate atomic/non-atomic writes here.

sachin-j-joshi (changes requested):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
andreipaduroiu: I can't cover that case anymore since we only issue one read request per page at any given time (#5406) . There is no chance that two read requests for the same page execute in parallel. 

This is old code that was moved around.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
andreipaduroiu: See explanation above.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/attributes/SegmentAttributeBTreeIndex.java
```diff
@@ -741,15 +797,11 @@ private void ensureInitialized() {
...
-            this.offset = offset;
-            this.size = size;
-            this.generation = currentGeneration;
-            this.cacheAddress = -1;
-        }
+        /**
+         * The {@link CacheStorage} address for this Cache Entry's data.
+         */
+        @Getter
+        private final int cacheAddress;
```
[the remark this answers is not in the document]
andreipaduroiu: It's a final field of a primitive type.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
andreipaduroiu: CacheEntryEvictedException is defined and thrown in this class. It's a way to handle a case when we are attempting to modify an entry while it is being evicted. Part of the process is not done while holding the cache lock so this may occur.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
andreipaduroiu: `   context.storage.supportsAtomicWrites = true;`

followed by lines 795-802 that add a partial write at the end of the segment.

This test verifies that the index gets corrupted if it trusts the Storage to say "it supports atomic writes" but in fact it does not.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/attributes/AttributeIndexTests.java
andreipaduroiu: fixed

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/Storage.java
andreipaduroiu: fixed

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/SegmentKeyCache.java
sachin-j-joshi: ok.

sachin-j-joshi (approved):
LGTM

PR #5532, opened 2021-01-15
Issue 5531: Duplicate layer of AppendBatchSizeTrackers

### LINKED ISSUE #5531: Duplicate layer of AppendBatchSizeTrackers
**Problem description**
The writer uses `AppendBatchSizeTracker` objects to infer the right size of batches to be written to the server. However, currently there are 2 places in which such `AppendBatchSizeTracker` objects are being instantiated: `FlowHandler` and `TcpClientConnection`. In principle, the batching logic should be only in one place, and concretely, at the connection level. This is because, without connection pooling enabled, he number of flows should be equal to the number of connections. With connection pooling enabled, is at the connection level at where we need to track the data being written to the server, which can be a mix of multiple multiplexed flows.

In fact, after a preliminary analysis, it looks like the `AppendBatchSizeTracker` objects instantiated in `FlowHandler` are not used at all, so we can safely remove them.

**Problem location**
Unnecessary `AppendBatchSizeTracker` objects instantiated in `FlowHandler`.

Remove unnecessary code.

RaulGracia:
**Change log description**  
Removed unnecessary instances of AppendBatchSizeTracker objects from FlowHandler.

**Purpose of the change**  
Fixes #5531.

**What the code does**  
In our performance experiments, we realized that `FlowHandler` is instantiating a `AppendBatchSizeTrackerImpl` object per flow. However, nowhere in the code the `getAppendBlockSize()` method of such objects is being invoked. This means that `FlowHandler` is keeping potentially thousands of objects in memory, jointly with their internal state, without any use at all. This PR just removes such an unnecessary code, so we can have efficiency gains and a better clarity of where `AppendBatchSizeTracker` is actually being used (i.e., at the TCP client connection level).

**How to verify it**  
No change in actual logic, just removed unused code from `FlowHandler`. All existing tests should be passing as before.

### DIFF
--- a/client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
+++ b/client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
@@ -13,7 +13,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import io.pravega.common.Exceptions;
 import io.pravega.shared.metrics.MetricNotifier;
-import io.pravega.shared.protocol.netty.AppendBatchSizeTracker;
 import io.pravega.shared.protocol.netty.ConnectionFailedException;
 import io.pravega.shared.protocol.netty.FailingReplyProcessor;
 import io.pravega.shared.protocol.netty.PravegaNodeUri;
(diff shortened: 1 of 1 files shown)


shrids (approved):
+1

RaulGracia:
Closing at the moment, as we may want to bring the AppendBatchSizeTracker to this level properly.

Issue #5469, opened 2021-01-08
Make txn max lease configuration a Long instead of an int


shiveshr:
**Problem description**
Txn max lease configuration is an integer which should be made a long. 

**Problem location**
Config.java

**Suggestions for an improvement**
Make the configuration a long.

RaulGracia:
Fixed via #5542.

PR #5542, opened 2021-01-18
Issue 5468, 5469: Increase max transaction lease range and make max execution time configurable

### LINKED ISSUE #5468: Make txn maximum TTL a configurable param
**Problem description**
Presently txn maximum TTL is derived from txn max_lease. 
We should decouple the two such that we allow for TTL to be much larger without large lease periods. 

**Problem location**
Config.java

make txn TTL configurable

shiveshr:
**Change log description**  
Changes max transaction lease configuration from integer to long to allow for higher leases. 
Make max execution time configurable so that we remove the hard coded max bound of 1 day. 

**Purpose of the change**  
Fixes #5468 #5469

**What the code does**  
1. In Config.java we change the max transaction lease configuration from integer to long to allow for higher leases. 
2. We have added a new configuration for users to specify max execution time. Prior to this change, the value was hardcoded to be upper bound at 1 day. With this change we remove that hard limit. So now its minimum of 1000*lease-period OR configured days. 

**How to verify it**  
Unit tests added.


controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -323,7 +326,7 @@ public void initializeStreamWriters(final EventStreamWriter<CommitEvent> commitW
                                                                                    final OperationContext ctx) {
         // Step 1. Validate parameters.
         CompletableFuture<Void> validate = validate(lease);
-        long maxExecutionPeriod = Math.min(MAX_EXECUTION_TIME_MULTIPLIER * lease, Duration.ofDays(1).toMillis());
+        long maxExecutionPeriod = Math.min(MAX_EXECUTION_TIME_MULTIPLIER * lease, maxTransactionExecutionTimeBound.get());
```
shrids, line 329: `maxExecutionPeriod` represents the maximum time for which the client may extend txn lease. It will always remain 1000 * 30second = 8.33 hours...

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -323,7 +326,7 @@ public void initializeStreamWriters(final EventStreamWriter<CommitEvent> commitW
                                                                                    final OperationContext ctx) {
         // Step 1. Validate parameters.
         CompletableFuture<Void> validate = validate(lease);
```
shrids, line 328: TimeoutServiceConfig#maxlease also needs to be updated from the current 30seconds?

shrids (changes requested):

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
shiveshr: certain connector implementations would actually want to use a much larger lease. maxlease can be increased and the intent is to set it to upto 30 days.. and a 1000 multiplier could take it to several years.. hence we also have a maxTransactionExecutionTimebound and we take min of these two values.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
shiveshr: TimeoutServiceConfig is set to config.max_lease_value in Main.java
```
            TimeoutServiceConfig timeoutServiceConfig = TimeoutServiceConfig.builder()
                    .maxLeaseValue(Config.MAX_LEASE_VALUE)
                    .build();
```
So we dont have a separate configuration for it from the txn max lease.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
shrids: ok, the user would need to change the controller properties `Config.MAX_LEASE_VALUE` and `Config.MAX_TXN_EXECUTION_TIMEBOUND_DAYS ` to ensure his requirements are met.

shrids (approved):
LGTM.

PR #5552, opened 2021-01-20
Issue 5551: Fix CommandEncoder memory leak

### LINKED ISSUE #5551: Client OOM: CommandEncoder session
**Problem description**
A longevity run OOM and after checking the heap dump, a huge number(30k) of `CommandEncoder#Session` has been found in the memory. 

There is no place in the code that removes the old session from the map and that's probably the root cause of this memory leak.

**Problem location**
CommandEncoder

Tristan1900:
**Change log description**  
An attempt to fix the memory leak in `CommandEncoder`.

**Purpose of the change**  
Fix #5551

**What the code does**  
Set a memory limit for `CommandEncoder` and when that limit is reached, shut down the TCP connection and let caller to re-create a new one.

**How to verify it**  
Test should pass


shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -80,7 +79,7 @@
     private static final byte[] LENGTH_PLACEHOLDER = new byte[4];
     private final Function<Long, AppendBatchSizeTracker> appendTracker;
     private final MetricNotifier metricNotifier;
-    private final Map<Map.Entry<String, UUID>, Session> setupSegments = new HashMap<>();
+    private final Map<String, Session> setupSegments = new HashMap<>();
```
RaulGracia, line 82: This is still not deleting the elements that are no longer used, right? Maybe we could replace the hashmap by a `SimpleCache` class in `common`, which is essentially a thread-safe hash map that incorporates eviction logic. For instance, we could discard any entry of this hash map if it has not been used for some time, right?

Tristan1900: Thanks for the review!
Yes it's still not deleting outdated segments, I suspect the main reason is that writer reconnects to the same segment with different writerId multiple times.
The `SimpleCache` sounds great, I will take a look at that. Thanks!

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -77,10 +77,12 @@
 @NotThreadSafe
 @Slf4j
 public class CommandEncoder extends FlushingMessageToByteEncoder<Object> {
+    private static final int INFINITE_SIZE = Integer.MAX_VALUE;
+    private static final Duration EVICTION_TIME = Duration.ofDays(1);
     private static final byte[] LENGTH_PLACEHOLDER = new byte[4];
     private final Function<Long, AppendBatchSizeTracker> appendTracker;
     private final MetricNotifier metricNotifier;
-    private final Map<Map.Entry<String, UUID>, Session> setupSegments = new HashMap<>();
+    private final SimpleCache<String, Session> setupSegments = new SimpleCache<>(INFINITE_SIZE, EVICTION_TIME, null);
```
tkaitchuck, line 85: These values can't be dropped randomly. So a cache is not an appropriate data structure. If one of these goes missing a write will suddenly have all their appends fail.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
```diff
@@ -195,7 +197,7 @@ protected void encode(ChannelHandlerContext ctx, Object msg, ByteBuf out) throws
         log.trace("Encoding message to send over the wire {}", msg);
         if (msg instanceof Append) {
             Append append = (Append) msg;
-            Session session = setupSegments.get(new SimpleImmutableEntry<>(append.segment, append.getWriterId()));
+            Session session = setupSegments.get(append.getSegment());
```
tkaitchuck, line 200: This is incorrect behaviour. Different writers which write on the same segment have different sessions and need  to setup appends individually. This may result in data corruption.

tkaitchuck (changes requested):

tkaitchuck:
Also this file cannot be the source of a client side memory leak, as this version of the encoder is for the server side.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
Tristan1900: got it!

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/CommandEncoder.java
Tristan1900: Got it. Thanks!

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -177,9 +184,18 @@ private void flushAllToBuffer() {
...
         if (msg instanceof SetupAppend) {
             breakCurrentAppend();
             flushAllToBuffer();
+            if (setupSegments.size() >= MAX_SETUP_SEGMENTS_SIZE) {
+                flushBuffer();
+                closed.compareAndSet(false, true);
+                if (callback != null) {
+                    callback.connectionDropped();
+                }
+                throw new IOException("CommandEncoder closed");
```
tkaitchuck, line 197: Include why in the error message. Probably also a connection ID etc.

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -177,9 +184,18 @@ private void flushAllToBuffer() {
     
     @Synchronized
     public void write(WireCommand msg) throws IOException {
+        checkIfClosed();
         if (msg instanceof SetupAppend) {
             breakCurrentAppend();
             flushAllToBuffer();
+            if (setupSegments.size() >= MAX_SETUP_SEGMENTS_SIZE) {
+                flushBuffer();
```
tkaitchuck, line 192: Maybe log at debug

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -422,5 +439,11 @@ public long batchTimeout(long token) {
             closeQuietly(output, log, "Closing output failed");
         }
         return result;
-    }    
+    }
+
+    private void checkIfClosed() throws IOException {
```
tkaitchuck, line 444: Exception.checkNotClosed does this in a generic way.

tkaitchuck (approved):

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -14,12 +14,9 @@
...
 import io.netty.buffer.ByteBufOutputStream;
 import io.netty.buffer.Unpooled;
+import io.pravega.common.Exceptions;
 import io.pravega.shared.metrics.MetricNotifier;
-import io.pravega.shared.protocol.netty.Append;
-import io.pravega.shared.protocol.netty.AppendBatchSizeTracker;
-import io.pravega.shared.protocol.netty.InvalidMessageException;
-import io.pravega.shared.protocol.netty.WireCommand;
-import io.pravega.shared.protocol.netty.WireCommandType;
+import io.pravega.shared.protocol.netty.*;
```
RaulGracia, line 19: This is not permitted by checkstyle.

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
```diff
@@ -56,7 +54,9 @@
     static final int MAX_QUEUED_EVENTS = 500;
     @VisibleForTesting
     static final int MAX_QUEUED_SIZE = 1024 * 1024; // 1MB
-    
+    @VisibleForTesting
+    static final int MAX_SETUP_SEGMENTS_SIZE = 2000;
```
RaulGracia, line 58: Ho do we know that this is the right limit? Could this lead to OOM anyway for clients with low memory resources, or could it limit the parallelism for powerful clients? I assume that this only impacts to real TCP connections and not to Connection Pooling, right?

RaulGracia (changes requested):

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
Tristan1900: Thanks! My IDE automatically did this, I didn't notice that.

andreipaduroiu:
@Tristan1900 Are we still pursuing this PR?

client/src/main/java/io/pravega/client/connection/impl/CommandEncoder.java
Tristan1900: 2k is just a number that I assume that should be fine. 10k should also work. I think making it 2000 limits the memory usage for a single TCP client to be less than 1 MB. The OOM is found when the acutal setup_segment_size being a few millions.
As for the parallelism, it means there are 2k active flows on a single TCP connection, the connection pool has logic to retrieve the connection with least flow, so that means basically every connection in the pool has 2k flows, the current default max connection for segmetstore is 10, so that's 20k active flows to a single segmentstore. I'm not sure if that's enough to use, but we can also increase the max connection for segmentstore number if we are expecting such huge workload. 
Yes it closes the TCP connection when that 2k limit is reached, so connection pool will detect that and prune this connection and perhaps creates a new one.
What do you think? Thanks!

shrids (approved):
LGTM

Tristan1900:
I will trigger a longevity test to verify this PR. please do not merge it right now.

RaulGracia:
@Tristan1900 has this PR passed system tests?

Tristan1900:
@RaulGracia Sorry I didn't see the comment. Yes it has passed the test. I'm seeing the CI not passing now. Are those `testEndToEndWithFencing` `testBoundedStreamWithTruncationTest` flaky tests? I'm rebuilding to see if it works.

PR #5563, opened 2021-01-20
Issue 5523: Enabling GithubActions and Disabling Travis CI

### LINKED ISSUE #5523: Investigate parallel GitHub Actions Build Tasks
**Problem description**
Investigate parallel GitHub Actions Build Tasks

**Problem location**

### LINKED ISSUE #5538: Setup JFrog snapshot publish in GitHub Actions
**Problem description**
We need the `snapshot` stage that is triggered **only** on master branch push builds (not on pull requests) that will publish to JFrog.

See `.travis.yml` (`snapshot` section).

**Problem location**
GitHub/workflows.

1. Get login credentials from JFrog for Pravega (DO NOT PUBLISH THEM ANYWHERE ON GITHUB).
2. Register them as secrets in the GitHub Pravega org (need a user name and a key)
3. Modify/create a GHA workflow that mimics the `snapshot` stage in `.travis.yml`.
3.1. This needs to be triggered on any push to `master` or to any release branch (see the regex in `.travis.yml`).

See this for how secret env vars are handled in Travis: https://docs.travis-ci.com/user/environment-variables/#encrypting-environment-variables

See this for how secret env vars are handled in GHA: https://docs.github.com/en/actions/reference/encrypted-secrets

andreipaduroiu:
**Change log description**  
- Disabled .travis.yml as we will no longer be using Travis CI. This file will be deleted in a subsequent PR.
- Set up GitHub Actions workflow as follows:
   - Triggers on master branch and release branch pushes, on every pull request (regardless of branch) and release publish.
   - Has a compile and static analysis job, followed by a few parallel test jobs, then a check and a snapshot (only for master branch and release branches).
   - Check job is called `CI Complete`.
- Bonus features: 
    - Build output is cached after the `Build` step and reused for subsequent jobs. This speeds up the overall workflow run.
    - Test execution output, history and reports is cached on a per workflow-job basis. If a workflow is re-run for whatever reason (such as one job failing), the successful jobs will execute much faster (by skipping all tests that passed).

**Purpose of the change**  
Fixes #5523.
Fixes #5538.

**What the code does**  
See description.

**How to verify it**  
Please verify that this is functionally equivalent to the previous CI pipeline:
- It compiles the code.
- It performs static analysis.
- It executes all unit tests and all integration tests.
- It uploads code coverage correctly.
- For master and release branches, it publishes to JFrog artifactory.
- Triggers are correct and make sense.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -31,7 +31,10 @@ env:
     gradle-m2-java-
     gradle-m2
   BUILD_CACHE_PATH: ./*
-  TEST_OUTPUT_CACHE_PATH: '**/build/test-results/** **/**/build/test-results/** **/**/build/test-results/**'
+  TEST_OUTPUT_CACHE_PATH: >
+    **/build/test-results
+    **/**/build/test-results
+    **/**/build/test-results
 
 # The workflow begins with a compilation and static analysis job that also caches the build output and source code,
 # followed by a number of parallel test jobs (which make use of that cached artifacts).
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -159,35 +159,6 @@ jobs:
           segmentstore:server:test
           segmentstore:storage:test
           segmentstore:server:host:test
-          --parallel ${{env.GRADLE_OPTS}}
-      - name: Codecov
-        uses: codecov/codecov-action@v1.0.15
-
-  # Bindings should be separate from segmentstore:server:host because of BK and ZK conflicts.
-  unit_bindings:
-    name: Storage Bindings Unit Tests
-    needs: build
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Gradle & Maven Cache
-        uses: actions/cache@v2.1.0
-        with:
-          path: ${{env.GLOBAL_CACHE_PATH}}
-          key: ${{env.GLOBAL_CACHE_KEY}}
-          restore-keys: ${{env.GLOBAL_CACHE_RESTORE_KEYS}}
-      - name: Build Output Cache
-        uses: actions/cache@v2.1.0
-        with:
-          path: ${{env.BUILD_CACHE_PATH}}
-          key: ${{github.run_id}}
-      - name: Test Output Cache
-        uses: actions/cache@v2.1.0
-        with:
-          path: ${{env.TEST_OUTPUT_CACHE_PATH}}
-          key: ${{github.run_id}}-${{github.job}}
-      - name: Storage Bindings Unit tests
-        run: >
-          ./gradlew
           segmentstore:storage:impl:test
           bindings:test
           --parallel ${{env.GRADLE_OPTS}}
@@ -257,7 +228,7 @@ jobs:
   # DO NOT RENAME THIS JOB. Mergers to master branch are gated on this completing successfully.
   build_and_test_complete:
     name: CI Complete
-    needs: [build, integration, unit_other, unit_segment_store, unit_bindings, unit_controller, unit_client]
+    needs: [build, integration, unit_other, unit_segment_store, unit_controller, unit_client]
     runs-on: ubuntu-20.04
     steps:
       - name: Check Build Status
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -34,7 +34,7 @@ env:
   BUILD_CACHE_PATH: ./*
   # We cache test execution (including history) and reports (code coverage) per test job. This helps speed up workflow re-runs.
   TEST_OUTPUT_CACHE_PATH: |
-    .gradle
+    .gradle/**/executionHistory
     **/build/test-results
     **/**/build/test-results
     **/**/build/test-results
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -158,9 +158,13 @@ jobs:
           ./gradlew
           segmentstore:server:test
           segmentstore:storage:test
-          segmentstore:server:host:test
           segmentstore:storage:impl:test
           bindings:test
+          --parallel ${{env.GRADLE_OPTS}}
+      - name: Segment Store Host Tests      # Run this separately since it conflicts with segmentstore:storage:impl:test.
+        run: >
+          ./gradlew
+          segmentstore:server:host:test
           --parallel ${{env.GRADLE_OPTS}}
       - name: Codecov
         uses: codecov/codecov-action@v1.0.15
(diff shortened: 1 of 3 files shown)


.github/workflows/gradle.yml
```diff
@@ -1,68 +1,215 @@
...
+      - published              # On every published release.
+
+# Define a few constants that are shared across all the jobs.
+env:
+  GLOBAL_CACHE_PATH: |
+    .gradle
+    ~/.gradle
+    ~/.m2
+  GLOBAL_CACHE_KEY: gradle-m2-java-11
+  GLOBAL_CACKE_RESTORE_KEYS: |
```
derekm, line 30: typo, GLOBAL_CAC**K**E_RESTORE_KEYS -> GLOBAL_CAC**H**E_RESTORE_KEYS

andreipaduroiu: Fixed.

andreipaduroiu:
Snapshot successfully published to JFrog (using modified condition to push from this branch). Output: https://oss.jfrog.org/artifactory/jfrog-dependencies/io/pravega/pravega-client/0.10.0-2756.ec7d25a-SNAPSHOT/

andreipaduroiu:
@RaulGracia I just fixed the build badge. A search for "travis" in the codebase yielded only the part where we want to run system tests on AWS. IMO, we should just get rid of that since we never used it.

However, we will be deleting `.travis.yml` in a future PR. We can take care of that cleanup then. Right now I still need that file so we can get **this** PR in.

derekm:
@andreipaduroiu -- Did you decide to keep the 6 parallel test jobs instead of reduce them down to 3?
image
 I suppose it is a wash in some ways, if we dispense of containers sooner, they are available sooner, but there is probably still some balance to strike between jobs-queue-to-container-assignment, concurrent builds across PRs, and exhaustion of our 20 container instances.

Client+Controller+Storage Bindings = 21 minutes
Segment Store+All Other = 15 minutes
Integration Tests = 22 minutes
would even things out a bit and give "All Other" room to grow.

andreipaduroiu:
@derekm It's pretty easy to reconfigure that part. After we get our build stable, reduce flaky tests and test duration we can consider rebalancing them and condensing them into 3-ish jobs.

I tried to collate all the Segment Store tasks into one job (and run it using `--parallel`) but I got increased test instability due to `segmentstore:storage:impl` and `segmentstore:server:host` being in the same run. I then realized that this was just transient and that the two could not possibly interfere with each other due the latter depending on the former, so Gradle would run them in sequence vs in parallel. The "flakiness" was present even after I split them again.

I have just re-combined them. Thanks for reminding me.

.github/workflows/gradle.yml
```diff
@@ -1,68 +1,258 @@
...
+      - name: Build Output Cache
+        uses: actions/cache@v2.1.0
+        with:
+          path: ${{env.BUILD_CACHE_PATH}}
+          key: ${{github.run_id}}
+      - name: Test Output Cache
+        uses: actions/cache@v2.1.0
+        with:
+          path: ${{env.TEST_OUTPUT_CACHE_PATH}}
+          key: ${{github.run_id}}-${{github.job}}
```
derekm, line 103: If we had
```
restore-keys: |
  ${{github.run_id}}
```
here, we could combine Build Output Cache and Test Output Cache actions into a single action and use just the `./*` cache path.

If we could run `actions/cache` on failure, then we can also prevent running successful tests again when there is an intermittent test failure. Instructions for forking `actions/cache` and making it run its "post" step in all cases: https://github.com/actions/cache/issues/272#issuecomment-650942160

If we can cache in all cases, then `key: ${{github.run_id}}-${{github.job}}-${{github.run_count}}` in a combined "Build Output Cache" action and
```
restore-keys: |
  ${{github.run_id}}-${{github.job}}-
  ${{github.run_id}}
```
will do the trick.

derekm (approved):
Left a comment about combining the Cache actions and possibly caching in all cases to improve test re-run times further. Leaving it to your discretion.

.github/workflows/gradle.yml
andreipaduroiu: #5568

.github/workflows/gradle.yml
```diff
@@ -1,68 +1,262 @@
...
 
+# Set up when this workflow will run.
 on:
   push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+    branches:
+      - master                 # On every push to the master branch.
+      - 'r[0-9]+.[0-9]+'       # On every push to a release branch.
```
derekm, line 17: We don't get the full benefit of this regex, as can be seen during the Artifactory Snapshot step below: https://github.com/pravega/pravega/pull/5563/files#diff-0d634959c8276ae976e39ba475e63d0b9ec27824470c79f1971f58fe37c46325R241

NOTE: The gate for the Artifactory Snapshot step will need to be revisited when Pravega reached version 2.0.

.github/workflows/gradle.yml
```diff
@@ -1,68 +1,262 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check Build Status
+        run: echo Build, static analysis, unit and integration tests successful.
+
+  snapshot:
+    name: Artifactory Snapshot
+    needs: [build_and_test_complete]
+    # Only run this on PUSH (no pull requests) and only on the master branch and release branches.
+    if: ${{ github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/head/r0.') || startsWith(github.ref, 'refs/head/r1.')) }}
```
derekm, line 245: NOTE: Revisit this line as Pravega reaches version 2.0.

medvedevigorek (approved):

.github/workflows/gradle.yml
andreipaduroiu: This particular wildcard will work for any future release. I think you are referring to the `snapshot` job (at the bottom) where I cannot use wildcard matches. There is no support for that in the `job.if` syntax. Perhaps by the time we reach version `2.x` (if ever - depending on our versioning scheme), GHA will add support for that.

andreipaduroiu:
@fpj Do you want to take a look at this PR? I got 3 approvals and all looks green.

The plan is to:
1. Merge this.
2. Cherry-pick into `r0.9`
3. Update the Pravega repo to gate mergers to master on the `CI Complete` step of GHA.
4. Update Pravega repo to delete Travis hooks and any Travis checks that were configured.

After these are done, we will be able to commit to master and to `r0.9`, but we won't have a CI for `r0.8` and before. My thinking is that we will figure out a solution for `r0.8` (and earlier branches) if we need to come up with new bugfix releases for those.

Keeping the Travis check around (even as optional) is not really a feasible option. Codecov will only upload coverage after the entire CI build completes; so if Travis needs a few hours to complete, then we haven't solved much with this.

fpj (approved):
This is really nice, @andreipaduroiu , thanks for doing it.

PR #5567, opened 2021-01-21
Issue 5562: Wrong outstanding appends calculation in AppendBatchSizeTracker using connection pooling

### LINKED ISSUE #5531: Duplicate layer of AppendBatchSizeTrackers

### LINKED ISSUE #5562: Wrong outstanding appends calculation in AppendBatchSizeTracker using connection pooling
**Problem description**
When connection pooling is disabled, a Pravega writer instantiates as many connections with segment stores as flows are created. In turn, currently the `AppendBatchSizeTrackerImpl` objects are being used on a per-connection basis, so there is a 1:1 relationship between these three entities (connections, batch trackers, flows). 

However, when connection pooling is enabled, the current implementation keeps a limited number of connections and active batch trackers (e.g., 10 per segment store), while the number of flows is still the same as in the case of connection pooling being disabled. This leads to a situation in which a single `AppendBatchSizeTracker` instance is receiving events from multiple flows. The problem is that, currently, the `AppendBatchSizeTracker` is oblivious to the fact that multiple flows can be multiplexed and it does not take into account multiple flows in the batch calculation. This can lead to problems, as in the estimation of `appendsOutstanding`, in which the moving average calculation is polluted with event numbers of many flows without any kind of control: https://github.com/pravega/pravega/blob/11308526048aaed234ef8ee7d3d61f1a72f5ca0e/client/src/main/java/io/pravega/client/connection/impl/AppendBatchSizeTrackerImpl.java#L75

**Problem location**
`AppendBatchSizeTrackerImpl` and `FLowHandler`

There are 2 options: i) either to fix `AppendBatchSizeTrackerImpl` to consider multiple flows (I have a prototype that works with this approach), and ii) make sure that there is a 1:1 relatioship between flows and `AppendBatchSizeTrackerImpl` instances. The latter approach may be better in the sense that it would be a smaller change and would keep intact the logic of `AppendBatchSizeTrackerImpl`.

RaulGracia:
**Change log description**  
Fixed the calculation of append batches in the client if connection pooling is enabled and multiple flows are multiplexed over the same connection.

**Purpose of the change**  
Fixes #5562.
Fixes #5531.

**What the code does**  

To solve #5562, this PR advocates to keep 1 `AppendBatchSizeTracker` object per flow. To this end:

- We introduce a new class (i.e., `FlowToBatchSizeTracker`), to keep the mapping between flows and batch trackers. It is based on `SimpleCache`, which provides us with thread-safety and an eviction mechanism, so this mapping will not grow forever.
- As `AppendBatchSizeTracker` is required at the `TCPClientConnection` and `CommandEncoder` levels, is there where we instantiate `FlowToBatchSizeTracker`. In this sense, this PR also removes the useless mapping of flow ids and batch trackers at the `FlowHandler` level, which in turn fixes #5531.
- Note that both `CommandEncoder` and `TCPClientConnection` fetch the appropriate `AppendBatchSizeTracker` by using `Flow.toFlowID(requestId)`. That is, the `AppendBatchSizeTracker` object is associated to a flow (i.e., most significant bits), irrespective of its sequence number (i.e., least significant bits).
- There is some minor and non-functional cleanup in `CommandEncoder` (e.g., wrong javadoc, redundant castings).

**How to verify it**  
Added new tests for `FlowToBatchSizeTracker`. I have tested this PR on AWS with connection pooling enabled (hundreds of segments and writers) and results are better in terms of latency.

### DIFF
--- a/client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
+++ b/client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
@@ -13,7 +13,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import io.pravega.common.Exceptions;
 import io.pravega.shared.metrics.MetricNotifier;
-import io.pravega.shared.protocol.netty.AppendBatchSizeTracker;
 import io.pravega.shared.protocol.netty.ConnectionFailedException;
 import io.pravega.shared.protocol.netty.FailingReplyProcessor;
 import io.pravega.shared.protocol.netty.PravegaNodeUri;
@@ -30,9 +29,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-
-import static io.pravega.shared.NameUtils.writerTags;
-import static io.pravega.shared.metrics.ClientMetricKeys.CLIENT_OUTSTANDING_APPEND_COUNT;
 
 @Slf4j
 public class FlowHandler extends FailingReplyProcessor implements AutoCloseable {
@@ -52,7 +48,6 @@ public class FlowHandler extends FailingReplyProcessor implements AutoCloseable 
     @VisibleForTesting
     @Getter(AccessLevel.PACKAGE)
     private final ConcurrentHashMap<Integer, ReplyProcessor> flowIdReplyProcessorMap = new ConcurrentHashMap<>();
-    private final ConcurrentHashMap<Integer, AppendBatchSizeTracker> flowIDBatchSizeTrackerMap = new ConcurrentHashMap<>();
     private final AtomicBoolean disableFlow = new AtomicBoolean(false);
 
     private FlowHandler(PravegaNodeUri location, MetricNotifier updateMetric) {
@@ -88,7 +83,6 @@ public class FlowHandler extends FailingReplyProcessor implements AutoCloseable 
         if (flowIdReplyProcessorMap.put(flowID, rp) != null) {
             throw new IllegalArgumentException("Multiple flows cannot be created with the same Flow id " + flowID);
         }
-        createAppendBatchSizeTrackerIfNeeded(flowID);
         return new FlowClientConnection(location.toString(), channel, flowID, this);
     }
 
@@ -103,7 +97,6 @@ public class FlowHandler extends FailingReplyProcessor implements AutoCloseable 
         Preconditions.checkState(!disableFlow.getAndSet(true), "Flows are disabled, incorrect usage pattern.");
         log.info("Creating a new connection with flow disabled for endpoint {}.", location);
         flowIdReplyProcessorMap.put(FLOW_DISABLED, rp);
-        createAppendBatchSizeTrackerIfNeeded(FLOW_DISABLED);
         return new FlowClientConnection(location.toString(), channel, FLOW_DISABLED, this);
     }
 
@@ -115,41 +108,10 @@ public class FlowHandler extends FailingReplyProcessor implements AutoCloseable 
         int flow = clientConnection.getFlowId();
         log.info("Closing Flow {} for endpoint {}", flow, clientConnection.getConnectionName());
         flowIdReplyProcessorMap.remove(flow);
-        flowIDBatchSizeTrackerMap.remove(flow);
         if (flow == FLOW_DISABLED) {
             // close the channel immediately since this connection will not be reused by other flows.
             close();
         }
-    }
-
-    /**
-     * Create a Batch size tracker, ignore if already existing
-     *
-     * @param flowID flow ID.
-     */
-    private void createAppendBatchSizeTrackerIfNeeded(final int  flowID) {
-        if (flowIDBatchSizeTrackerMap.containsKey(flowID)) {
-            log.debug("Reusing Batch size tracker for Flow ID {}.", flowID);
-        } else {
-            log.debug("Creating Batch size tracker for flow ID {}.", flowID);
-            flowIDBatchSizeTrackerMap.put(flowID, new AppendBatchSizeTrackerImpl());
-        }
-    }
-
-    /**
-     * Get a Batch size tracker for requestID.
-     *
-     * @param requestID flow ID.
-     * @return Batch size Tracker object.
-     */
-    public AppendBatchSizeTracker getAppendBatchSizeTracker(final long requestID) {
-        int flowID;
-        if (disableFlow.get()) {
-            flowID = FLOW_DISABLED;
-        } else {
-            flowID = Flow.toFlowID(requestID);
-        }
-        return flowIDBatchSizeTrackerMap.get(flowID);
     }
 
     /**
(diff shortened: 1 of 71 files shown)


client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
```diff
@@ -115,7 +113,7 @@ public void run() {
                     WireCommand command = readCommand(in, buffer);
                     if (command instanceof WireCommands.DataAppended) {
                         WireCommands.DataAppended dataAppended = (WireCommands.DataAppended) command;
-                        batchSizeTracker.recordAck(dataAppended.getEventNumber());
+                        flowToBatchSizeTracker.getAppendBatchSizeTrackerByFlowId(dataAppended.getRequestId()).recordAck(dataAppended.getEventNumber());
```
shrids, line 116: Using `io.pravega.client.connection.impl.Flow#toFlowID` api to get the requestId will ensure  only the flow id is used. This will ensure requests via the RawClient do not create multiple `AppendBatchSizeTrackers`.

client/src/main/java/io/pravega/client/connection/impl/FlowToBatchSizeTracker.java
```diff
@@ -0,0 +1,69 @@
...
+ * Note that if a flow entry is evicted and then the flow is used again, the only consequence will be that its previous
+ * batching information would be lost and the flow will start operating with the batching algorithm defaults.
+ */
+@Slf4j
+public class FlowToBatchSizeTracker {
+
+    /**
+     * Maximum time that the want to keep an inactive flow entry in the cache.
+     */
+    private static final Duration TRACKER_CACHE_EXPIRATION_TIME = Duration.ofSeconds(60);
```
shrids, line 36: This value can be same as `io.pravega.client.connection.impl.TcpClientConnection#SOCKET_TIMEOUT_MS`

shrids (commented):
Great Find 👍 
The changes LGTM apart from a couple of questions...

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
RaulGracia: Done.

client/src/main/java/io/pravega/client/connection/impl/FlowToBatchSizeTracker.java
RaulGracia: Done.

shrids (approved):
LGTM

client/src/main/java/io/pravega/client/connection/impl/TcpClientConnection.java
shrids, line 116: A response to a conditionalAppend would create a AppendBatchSizetracker instance...

RaulGracia: Thanks @shrids, that's correct, but then we need to use `Flow.toFlowID()` in `CommandEncoder` as well as I added in the last commit. Otherwise, the `CommandEncoder` and `TcpClientConnection` refer to different keys for the same `AppendBatchSizeTracker`.

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
```diff
@@ -185,18 +147,6 @@ public void process(Reply cmd) {
...
-        if (cmd instanceof WireCommands.DataAppended) {
-            final WireCommands.DataAppended dataAppended = (WireCommands.DataAppended) cmd;
-            final AppendBatchSizeTracker batchSizeTracker = getAppendBatchSizeTracker(dataAppended.getRequestId());
-            if (batchSizeTracker != null) {
-                long pendingAckCount = batchSizeTracker.recordAck(dataAppended.getEventNumber());
-                // Only publish client side metrics when there is some metrics notifier configured for efficiency.
-                if (!metricNotifier.equals(MetricNotifier.NO_OP_METRIC_NOTIFIER)) {
-                    metricNotifier.updateSuccessMetric(CLIENT_OUTSTANDING_APPEND_COUNT, writerTags(dataAppended.getWriterId().toString()),
-                            pendingAckCount);
-                }
```
tkaitchuck, line 197: Are we loosing this metric?

tkaitchuck (approved):

client/src/main/java/io/pravega/client/connection/impl/FlowHandler.java
RaulGracia: We have never used it since it was introduced as far as I know, and it may reduce the computation overhead of the client when processing the ack. We can bring it back in the future if necessary.

shrids (approved):
+1

client/src/main/java/io/pravega/client/connection/impl/FlowToBatchSizeTracker.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Maximum number of active flow entries in the cache.
+     */
+    private static final int TRACKER_CACHE_MAX_SIZE = 100000;
+
+    @VisibleForTesting
+    @Getter(value = AccessLevel.PACKAGE)
+    private final SimpleCache<Integer, AppendBatchSizeTracker> flowToBatchSizeTrackerMap;
+
+    public FlowToBatchSizeTracker() {
```
fpj, line 46: Should this constructor and this class be package protected rather than public?

client/src/main/java/io/pravega/client/connection/impl/FlowToBatchSizeTracker.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * This method returns the {@link AppendBatchSizeTracker} instance for a given flow id. If the flow id does not exist
+     * yet in the cache, the new entry for the flow is created along with a new {@link AppendBatchSizeTracker} object.
+     *
+     * @param flowId    Identifier for the flow.
+     * @return The {@link AppendBatchSizeTracker} instance associated to the flow id.
+     */
+    public AppendBatchSizeTracker getAppendBatchSizeTrackerByFlowId(int flowId) {
+        if (flowToBatchSizeTrackerMap.putIfAbsent(flowId, new AppendBatchSizeTrackerImpl()) == null) {
+            log.info("Instantiating new batch sze tracker for flow: {}", flowId);
```
fpj, line 65: This is a debug message.

fpj (changes requested):
A couple of minor points.

client/src/main/java/io/pravega/client/connection/impl/FlowToBatchSizeTracker.java
RaulGracia: Done.

client/src/main/java/io/pravega/client/connection/impl/FlowToBatchSizeTracker.java
RaulGracia: Done.

leapsky:
Experiments with `5000 segments, 250MBps, 1KB events` and `connection pooling enabled`, `1` and `100 writers` ran successfully.
Also, I ran experiments with `1000 segments, 250MBps, 1KB events` and `connection pooling disabled` which showed similar results that we had for master.

RaulGracia:
@fpj thanks for the feedback, I have addressed your comments. If you don't have any further comments, please try to give a go to this PR as it would be important to cherry-pick it to `r0.9`.

fpj (approved):

Issue #5587, opened 2021-01-27
BookKeeper Docker Image build failure


AndreyKoltsov1997:
**Problem description**
Unable to build BookKeeper Docker image due to  syntax error in  `get-pip.py` script.
```
...
Traceback (most recent call last):
  File "get-pip.py", line 24226, in <module>
    main()
  File "get-pip.py", line 199, in main
    bootstrap(tmpdir=tmpdir)
  File "get-pip.py", line 82, in bootstrap
    from pip._internal.cli.main import main as pip_entry_point
  File "/tmp/tmp25EwzN/pip.zip/pip/_internal/cli/main.py", line 60
    sys.stderr.write(f"ERROR: {exc}")
...
```

**Problem location**
[BookKeeper's Dockerfile](https://github.com/pravega/pravega/blob/master/docker/bookkeeper/Dockerfile)

**Suggestions for an improvement**
Based on [PIP changelog](https://pip.pypa.io/en/stable/news/#id1), Python 2 is not supported since `pip 20.1 (2020-04-28)`, python 2 and python 3.5 aren't supported since `pip 21.0 (2021-01-23)`.
For python 2.7, it should be considered to change pip source to 2.7:  https://bootstrap.pypa.io/2.7/

PR #5588, opened 2021-01-27
Issue 5587: BookKeeper Docker Image build failure.

AndreyKoltsov1997:
**Change log description**  
 [BookKeeper Dockerfile](https://github.com/pravega/pravega/blob/master/docker/bookkeeper/Dockerfile) uses  `apache/bookkeeper:4.11.1` as a base image. The Dockerfile downloads [pip bootstrap script](https://github.com/pravega/pravega/blob/master/docker/bookkeeper/Dockerfile#L40) whose installation is failing since it uses newer, Python 3 syntax. Thus, BookKeeper image build fails as well.
`pip` is installed in order to install `zk-shell` into the image, however, `zk-shell` is already available within base image `apache/bookkeeper:4.11.1`, thus both pip and zk-shell installations could be removed from the Dockerfile.

**Purpose of the change**  
Fixes https://github.com/pravega/pravega/issues/5587

**What the code does** 
Removes `pip` and `zk-shell` references from [BookKeeper Dockerfile](https://github.com/pravega/pravega/blob/master/docker/bookkeeper/Dockerfile).

**How to verify it**  
Build BookKeeper docker image.

Signed-off-by: Andrey Koltsov <a99920641@gmail.com>

### DIFF
--- a/docker/bookkeeper/Dockerfile
+++ b/docker/bookkeeper/Dockerfile
@@ -36,7 +36,7 @@ RUN set -x \
     && cp -r bookkeeper-all-${BK_VERSION}/* /opt/bookkeeper/ \
     && rm -rf "bookkeeper-all-${BK_VERSION}" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz.sha512" \
     # install zookeeper shell
-    && wget -q https://bootstrap.pypa.io/get-pip.py \
+    && wget -q https://bootstrap.pypa.io/2.7/get-pip.py \
     && python get-pip.py \
     && pip install zk-shell \
     && rm -rf get-pip.py \
--- a/docker/bookkeeper/Dockerfile
+++ b/docker/bookkeeper/Dockerfile
@@ -35,11 +35,6 @@ RUN set -x \
     && tar -xzf "$DISTRO_NAME.tar.gz" \
     && cp -r bookkeeper-all-${BK_VERSION}/* /opt/bookkeeper/ \
     && rm -rf "bookkeeper-all-${BK_VERSION}" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz.sha512" \
-    # install zookeeper shell
-    && wget -q https://bootstrap.pypa.io/2.7/get-pip.py \
-    && python get-pip.py \
-    && pip install zk-shell \
-    && rm -rf get-pip.py \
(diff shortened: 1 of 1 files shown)


docker/bookkeeper/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN set -x \
     && cp -r bookkeeper-all-${BK_VERSION}/* /opt/bookkeeper/ \
     && rm -rf "bookkeeper-all-${BK_VERSION}" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz.sha512" \
     # install zookeeper shell
-    && wget -q https://bootstrap.pypa.io/get-pip.py \
+    && wget -q https://bootstrap.pypa.io/2.7/get-pip.py \
```
RaulGracia, line 39: One question: This image is based on `apache/bookkeeper:4.11.1`, which already has installed `zk-shell`, right? So my question is, can we get rid of install pip at all? `zk-shell` should be already available, if I'm not wrong.

RaulGracia (changes requested):

docker/bookkeeper/Dockerfile
AndreyKoltsov1997: Great catch! You're right - base image `apache/bookkeeper:4.11.1` has zk-shell installed out of the box, thus I think it'd be safe to remove zk-shell from the Dockerfile.
I haven't found any `pip` usage apart from `zk-shell` installation, thus I think `pip` references could be removed from Dockerfile as well.
```
$ docker run -it --entrypoint /bin/bash apache/bookkeeper:4.11.1
...
4.11.1: Pulling from docker.io/apache/bookkeeper
...
Status: Downloaded newer image for docker.io/apache/bookkeeper:4.11.1
[root@e544aa0b1b8c bookkeeper]# zk-shell
Welcome to zk-shell (1.3.2)
(DISCONNECTED) />
```

RaulGracia:
Thanks @AndreyKoltsov1997, the change looks good. Let's execute a branch builder on this branch so we can confirm that everything is ok before merging.

RaulGracia (approved):
Branch builder 745 is successful with this change.

RaulGracia:
@AndreyKoltsov1997 please, can you change the PR description to match the current approach of the PR?

AndreyKoltsov1997:
> @AndreyKoltsov1997 please, can you change the PR description to match the current approach of the PR?

Sure. I've just updated the PR description.

medvedevigorek (approved):

PR #5591, opened 2021-01-27
Issue 5584: Avoid Prematurely closing SegmentStore Write Metrics

### LINKED ISSUE #5584: SegmentStore Write Metrics Unreported After Timeout
**Problem description**

After approximately `DEFAULT_EXPIRY_DURATION`  time has elapsed, `segmentstore.segment.write_bytes/events` metrics stop being exported to `micrometer` registries under transaction workloads. This gives the perception that writes are not happening, even though they are still being reported by the segmentstore and other secondary metrics.

**Problem location**

* `SegmentStatsRecorderImpl.java`

* Avoid pre-mature closure of `segmentstore.segment.write_bytes/events` metrics.

co-jo:
**Change log description**  

* Avoids prematurely closing `SEGMEMT_WRITE_BYTE` metrics on segments due to eviction of  transaction segment `SimpleCache` entries (multiple entries reference the same `MetricProxy`).

**Purpose of the change**  
Fixes #5584 

**What the code does**  
* Prevents this conflict by only adding non-transaction segments to the `SimpleCache`. 

**How to verify it**  
* Run the unit tests.

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
@@ -176,7 +176,9 @@ class SegmentStatsRecorderImpl implements SegmentStatsRecorder {
     public void createSegment(String streamSegmentName, byte type, int targetRate, Duration elapsed) {
         this.createStreamSegment.reportSuccessEvent(elapsed);
         SegmentAggregates sa = SegmentAggregates.forPolicy(ScaleType.fromValue(type), targetRate);
-        cache.put(streamSegmentName, new SegmentWriteContext(streamSegmentName, sa));
+        if (!NameUtils.isTransactionSegment(streamSegmentName)) {
+            cache.put(streamSegmentName, new SegmentWriteContext(streamSegmentName, sa));
+        }
         if (sa.isScalingEnabled()) {
             reporter.notifyCreated(streamSegmentName);
         }
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
@@ -352,6 +352,11 @@ class SegmentStatsRecorderImpl implements SegmentStatsRecorder {
         return context == null ? null : context.getSegmentAggregates();
     }
 
+    @VisibleForTesting
+    protected SimpleCache<String, SegmentWriteContext> getSimpleCache() {
+        return this.cache;
+    }
+
     private static class SegmentWrite {
         long bytes = 0;
         int events = 0;
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
@@ -352,11 +352,6 @@ class SegmentStatsRecorderImpl implements SegmentStatsRecorder {
         return context == null ? null : context.getSegmentAggregates();
     }
 
-    @VisibleForTesting
-    protected SimpleCache<String, SegmentWriteContext> getSimpleCache() {
-        return this.cache;
-    }
-
     private static class SegmentWrite {
         long bytes = 0;
         int events = 0;
--- a/segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderTest.java
+++ b/segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderTest.java
@@ -181,6 +181,31 @@ public class SegmentStatsRecorderTest extends ThreadPooledTestSuite {
         assertEquals(NO_COUNTER_VALUE, getCounterValue(MetricsNames.SEGMENT_READ_BYTES, segmentName));
     }
 
+    @Test(timeout = 10000)
+    public void testTransactionMetrics() throws InterruptedException {
+        val segmentName = getStreamSegmentName();
+
+        long dataLength = 321;
+        int interval = 500;
+        val duration = Duration.ofMillis(2 * interval);
+        @Cleanup
+        val context = new TestContext(segmentName, duration, false);
+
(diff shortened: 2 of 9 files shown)


segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderTest.java
```diff
@@ -181,6 +181,31 @@ public void testMetrics() {
...
+        int interval = 500;
+        val duration = Duration.ofMillis(2 * interval);
+        @Cleanup
+        val context = new TestContext(segmentName, duration, false);
+
+        context.statsRecorder.createSegment(segmentName, ScalingPolicy.ScaleType.BY_RATE_IN_KBYTES_PER_SEC.getValue(), 2, Duration.ofSeconds(1));
+        context.statsRecorder.recordAppend(segmentName, dataLength, 5, Duration.ofSeconds(2));
+        val txnName1 = NameUtils.getTransactionNameFromId(segmentName, UUID.randomUUID());
+        context.statsRecorder.createSegment(txnName1, ScalingPolicy.ScaleType.BY_RATE_IN_KBYTES_PER_SEC.getValue(), 2, Duration.ofSeconds(1));
+        Thread.sleep(duration.toMillis() / 2);
```
co-jo, line 198: From my understanding `Thread.sleep` only provides a lower bound on how much time will have elapsed before waking the thread, meaning it will not _absolutely guarantee_ that `segmentName` won't also be evicted alongside `txnName1` . Obviously `txnName1` can't actually be evicted because its prevented from entering the cache , but this test is setup to emulate the previous conditions and assert they cannot happen.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java
```diff
@@ -176,7 +176,9 @@ private SegmentWriteContext getWriteContext(String streamSegmentName) {
     public void createSegment(String streamSegmentName, byte type, int targetRate, Duration elapsed) {
         this.createStreamSegment.reportSuccessEvent(elapsed);
         SegmentAggregates sa = SegmentAggregates.forPolicy(ScaleType.fromValue(type), targetRate);
-        cache.put(streamSegmentName, new SegmentWriteContext(streamSegmentName, sa));
+        if (!NameUtils.isTransactionSegment(streamSegmentName)) {
+            cache.put(streamSegmentName, new SegmentWriteContext(streamSegmentName, sa));
```
medvedevigorek, line 180: @shiveshr this change will effectively make [SegmentStatsRecorderImpl.policyUpdate()](https://github.com/pravega/pravega/blob/master/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderImpl.java#L214) a no-op for transaction segments. Can you please confirm that there is no side effect of this one some other logic?

shiveshr: this should be inconsequential for txn segments.. this is because there is no policy update on a transaction segment anyway as the segment policy is about scaling and txn segments never scale.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderTest.java
```diff
@@ -181,6 +181,31 @@ public void testMetrics() {
...
+        int interval = 500;
+        val duration = Duration.ofMillis(2 * interval);
+        @Cleanup
+        val context = new TestContext(segmentName, duration, false);
+
+        context.statsRecorder.createSegment(segmentName, ScalingPolicy.ScaleType.BY_RATE_IN_KBYTES_PER_SEC.getValue(), 2, Duration.ofSeconds(1));
+        context.statsRecorder.recordAppend(segmentName, dataLength, 5, Duration.ofSeconds(2));
+        val txnName1 = NameUtils.getTransactionNameFromId(segmentName, UUID.randomUUID());
+        context.statsRecorder.createSegment(txnName1, ScalingPolicy.ScaleType.BY_RATE_IN_KBYTES_PER_SEC.getValue(), 2, Duration.ofSeconds(1));
+        Thread.sleep(interval);
```
andreipaduroiu, line 198: Please do not put Thread.sleep in your tests. Rework your test or use mocks to simulate this.

andreipaduroiu (changes requested):

RaulGracia (approved):
LGTM.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderTest.java
andreipaduroiu: You can use `TestUtils.await` or `AssertExtensions.assertEventuallyEquals`. If not already public, expose (via special getters) some info about your internal state (and mark it as `@VisibleForTesting`); then you can use one of those 2 methods to await some internal event to happen before resuming.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/stat/SegmentStatsRecorderTest.java
```diff
@@ -181,6 +181,28 @@ public void testMetrics() {
...
+        context.statsRecorder.createSegment(segmentName, ScalingPolicy.ScaleType.BY_RATE_IN_KBYTES_PER_SEC.getValue(), 2, Duration.ofSeconds(1));
+        context.statsRecorder.recordAppend(segmentName, dataLength, 5, Duration.ofSeconds(2));
+        val txnName1 = NameUtils.getTransactionNameFromId(segmentName, UUID.randomUUID());
+        context.statsRecorder.createSegment(txnName1, ScalingPolicy.ScaleType.BY_RATE_IN_KBYTES_PER_SEC.getValue(), 2, Duration.ofSeconds(1));
+        // Make sure deletions do not cause side effects.
+        context.statsRecorder.deleteSegment(txnName1);
+        assertEquals(dataLength, getCounterValue(SEGMENT_WRITE_BYTES, segmentName));
+        // All closures of metrics happen through the SimpleCache. Asserting that an entry for `txnName1` does not exist
+        // ensures that there is guaranteed to be a one-one mapping of entries to counters.
+        assertNull(context.statsRecorder.getSimpleCache().get(txnName1));
```
andreipaduroiu, line 202: You can use `context.statsRecorder.getSegmentAggregates(txnName1)` so you don't have to expose the cache.

andreipaduroiu:
PR looks good. Approved.

andreipaduroiu (approved):

PR #5598, opened 2021-01-28
Issue 5597 : ExtendedS3ChunkStorage and ExtendedS3Storage should close S3Client.

### LINKED ISSUE #5597: ExtendedS3ChunkStorage and ExtendedS3Storage do not close S3Client
**Problem description**
ExtendedS3ChunkStorage and ExtendedS3Storage do not close S3Client

**Problem location**
ExtendedS3ChunkStorage
ExtendedS3Storage
ChunkedSegmentStorage

ExtendedS3ChunkStorage and ExtendedS3Storage should close S3Client

sachin-j-joshi:
ExtendedS3ChunkStorage and ExtendedS3Storage should close S3Client

**Change log description**  
ExtendedS3ChunkStorage and ExtendedS3Storage should close S3Client

**Purpose of the change**  
Fixes #5597 

**What the code does**  
ChunkedSegmentStorage closes chunkStorage. 
ExtendedS3ChunkStorage and ExtendedS3Storage close S3Client.

**How to verify it**  
Build should pass.

### DIFF
--- a/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
+++ b/bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
@@ -68,11 +68,7 @@ public class ExtendedS3ChunkStorage extends BaseChunkStorage {
     //endregion
 
     //region constructor
-    public ExtendedS3ChunkStorage(S3Client client, ExtendedS3StorageConfig config, Executor executor) {
-        this(client, config, executor, false);
-    }
-
-    ExtendedS3ChunkStorage(S3Client client, ExtendedS3StorageConfig config, Executor executor, boolean shouldClose) {
+    public ExtendedS3ChunkStorage(S3Client client, ExtendedS3StorageConfig config, Executor executor, boolean shouldClose) {
         super(executor);
         this.config = Preconditions.checkNotNull(config, "config");
         this.client = Preconditions.checkNotNull(client, "client");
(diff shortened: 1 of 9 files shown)


bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3ChunkStorage.java
```diff
@@ -295,6 +298,17 @@ protected void doDelete(ChunkHandle handle) throws ChunkStorageException {
         }
     }
 
+    @Override
+    public void close() {
+        if (!this.closed.getAndSet(true)) {
+            try {
+                this.client.destroy();
```
andreipaduroiu, line 305: While this is going to work, I don't like the pattern. This `client` is given to this class via the constructor so there is no guarantee that nobody else is using it. If you are closing it, then anyone else who might be using it will be in trouble.

Please refactor the factory class and this class to have this class construct the object - which means it owns it wholly.

bindings/src/main/java/io/pravega/storage/extendeds3/ExtendedS3Storage.java
```diff
@@ -545,7 +545,13 @@ private Void doDelete(SegmentHandle handle) {
 
     @Override
     public void close() {
-        this.closed.set(true);
+        if (!this.closed.getAndSet(true)) {
+            try {
+                this.client.destroy();
```
andreipaduroiu, line 550: same here

andreipaduroiu (changes requested):

sachin-j-joshi:
@andreipaduroiu Changed the constructor to take `shouldClose` this way ExtendedS3Storage doesn't have to decide whether to use shared client or not, instead the factory decides and passes correct parameters to `shouldClose` appropriately.

Also we want to move towards optimistic checks. So, instead of checking isClosed on every call we just let the calls fail when client is closed.

andreipaduroiu (approved):

PR #5602, opened 2021-01-28
Issue 3699: Fix for flaky test StoreClientFactoryTest.testZkSessionExpiryRetry

### LINKED ISSUE #3699: Sporadic unit test failure: StoreClientFactoryTest.testZkSessionExpiryRetry
**Problem description**
```
io.pravega.controller.store.client.StoreClientFactoryTest > testZkSessionExpiryRetry FAILED
    java.lang.RuntimeException: Exception of the wrong type. Was expecting class org.apache.zookeeper.KeeperException$SessionExpiredException but got: org.apache.zookeeper.KeeperException$NoNodeException
        at io.pravega.test.common.AssertExtensions.assertThrows(AssertExtensions.java:75)
        at io.pravega.controller.store.client.StoreClientFactoryTest.testZkSessionExpiryRetry(StoreClientFactoryTest.java:84)
        Caused by:
        org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /test/test
            at org.apache.zookeeper.KeeperException.create(KeeperException.java:115)
            at org.apache.zookeeper.KeeperException.create(KeeperException.java:51)
            at org.apache.zookeeper.ZooKeeper.getData(ZooKeeper.java:2019)
            at org.apache.curator.framework.imps.GetDataBuilderImpl$4.call(GetDataBuilderImpl.java:327)
            at org.apache.curator.framework.imps.GetDataBuilderImpl$4.call(GetDataBuilderImpl.java:316)
            at org.apache.curator.connection.StandardConnectionHandlingPolicy.callWithRetry(StandardConnectionHandlingPolicy.java:64)
            at org.apache.curator.RetryLoop.callWithRetry(RetryLoop.java:100)
            at org.apache.curator.framework.imps.GetDataBuilderImpl.pathInForeground(GetDataBuilderImpl.java:313)
            at org.apache.curator.framework.imps.GetDataBuilderImpl.forPath(GetDataBuilderImpl.java:304)
            at org.apache.curator.framework.imps.GetDataBuilderImpl.forPath(GetDataBuilderImpl.java:35)
            at io.pravega.controller.store.client.StoreClientFactoryTest.lambda$testZkSessionExpiryRetry$4(StoreClientFactoryTest.java:84)
            at io.pravega.test.common.AssertExtensions.assertThrows(AssertExtensions.java:64)
            ... 1 more
```

**Problem location**
Controller.

Fix unit test or underlying code.

pbelgundi:
**Change log description**  
This PR fixes the following things about this test:
1. The assertion for `expirationRetryCounter` has been changed to check that retry counter is > 0, which means connection was retried at-least once post expiration. The actual number of retries is not of consequence to us here as it is not controlled by Pravega code.
2. The busy wait for Zookeeper State to be non-Alive is not necessary and can cause test timeouts, hence removed.

**Purpose of the change**  
Fixes #3699

**How to verify it**  
Test StoreClientFactoryTest.testZkSessionExpiryRetry should pass consistently.


controller/src/test/java/io/pravega/controller/store/client/StoreClientFactoryTest.java
```diff
@@ -70,24 +67,10 @@ public void testZkSessionExpiryRetry() throws Exception {
...
 
-        Supplier<Boolean> isAliveSupplier = () -> {
-            try {
-                return client.getZookeeperClient().getZooKeeper().getState().isAlive();
-            } catch (Exception e) {
-                throw new CompletionException(e);
-            }
-        };
-
-        Futures.loop(isAliveSupplier,
```
shiveshr, line 81: we could write this same loop using assertEventuallyEquals.. but why have we removed this? do we no longer need to wait on this?

pbelgundi: My understanding is, with `injectSessionExpiration` the Curator actually fakes a session expiration and so checking the State of the real Zookeeper instance is not needed here and the busy wait can actually cause the test to timeout unnecessarily because the two things are internally unrelated.
Ref:
https://stackoverflow.com/questions/41862575/how-does-apache-curator-try-to-reconnect-when-session-expires
http://curator.apache.org/errors.html ( See under "Notifications" Zookeeper state=LOST)

pbelgundi: Code for ZookeeperTestable.injectSessionExpiration :
```
class ZooKeeperTestable implements Testable {
    private static final Logger LOG = LoggerFactory
            .getLogger(ZooKeeperTestable.class);

    private final ZooKeeper zooKeeper;
    private final ClientCnxn clientCnxn;

    ZooKeeperTestable(ZooKeeper zooKeeper, ClientCnxn clientCnxn) {
        this.zooKeeper = zooKeeper;
        this.clientCnxn = clientCnxn;
    }

    @Override
    public void injectSessionExpiration() {
        LOG.info("injectSessionExpiration() called");

        clientCnxn.eventThread.queueEvent(new WatchedEvent(
                Watcher.Event.EventType.None,
                Watcher.Event.KeeperState.Expired, null));
        clientCnxn.eventThread.queueEventOfDeath();
        clientCnxn.state = ZooKeeper.States.CLOSED;
        clientCnxn.sendThread.getClientCnxnSocket().onClosing();
    }
}
```

pbelgundi: Added assertEventuallyThrows() call  as suggested above to the test to make sure we attempt get the exception multiple times, subject to a timeout.

test/testcommon/src/main/java/io/pravega/test/common/AssertExtensions.java
```diff
@@ -88,6 +88,42 @@
...
+     *
+     * @param run  The Runnable to execute.
+     * @param type The type of exception to expect.
+     * @param checkIntervalMillis The number of milliseconds to wait between two checks.
+     * @param timeoutMillis       The timeout in milliseconds after which an assertion error should be thrown.
+     * @throws Exception If the is an assertion error, and exception from `eval`, or the thread is interrupted.
+     */
+    public static void assertEventuallyThrows(Class<? extends Exception> type, RunnableWithException run, int checkIntervalMillis, long timeoutMillis) throws Exception {
+        try {
+        TestUtils.awaitException(() -> {
```
RaulGracia, line 102: Nit: indentation inside the try block.

RaulGracia (approved):
LGTM, just one minor comment. Have you verified the stability of this test running it in a loop?

shiveshr (approved):

PR #5605, opened 2021-01-28
Issue 5604: Speed up Client Unit Tests

### LINKED ISSUE #5604: Speed up client unit tests
**Problem description**
ControllerImplTests (+Secure)
ConnectionFactoryImplTests (+Secure)

**Problem location**

andreipaduroiu:
**Change log description**  
- ConnectionFactoryImplTests: added short shutdown timeout to prevent waiting 10s for thread pools to shut down.
- ControllerImplTests: split `testKeepAlive` into two tests and adjusted parameters.
- Also fixes their "Secure" counterparts (since they inherit from base).

**Purpose of the change**  
Fixes #5604.

**What the code does**  
- ConnectionFactoryImplTests: added a shutdown timeout to speed up the `tearDown`
- ControllerImplTests`
    - Split `testKeepAlive` into 2 tests as it was testing 2 different things
    - Adjusted Controller configuration (both client-side and server-side) to test the same thing but without having to wait 40s for it (which was likely calculated based on config defaults).
    - The new args are proportional to the old ones, but reduced by a factor of 20.
- The entire runtime of the Client unit tests has been reduced to 90s, down from 4 minutes.

**How to verify it**  
Build must pass.

### DIFF
--- a/client/src/test/java/io/pravega/client/connection/impl/ConnectionFactoryImplTest.java
+++ b/client/src/test/java/io/pravega/client/connection/impl/ConnectionFactoryImplTest.java
@@ -35,6 +35,7 @@ import io.pravega.test.common.SecurityConfigDefaults;
 import io.pravega.test.common.TestUtils;
 import java.io.File;
 import java.net.URI;
+import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLEngine;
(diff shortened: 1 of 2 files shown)


RaulGracia (approved):
LGTM, thanks for taking the time to speed up the tests.

PR #5609, opened 2021-01-29
Issue 5608: Speed up Standalone Unit Tests

### LINKED ISSUE #5608: Speed up standalone unit tests
**Problem description**
They're very slow (50-90s each).

**Problem location**

andreipaduroiu:
**Change log description**  
- Speeding up standalone tests. 
- Made Controller shutdown timeouts configurable (and passing smaller values for standalone unit tests).
- Updated standalone test config to reduce unnecessary internal retries.
- Updated standalone unit tests to use only 1 segment container and fixed a bug in InProcPravegaCluster which wasn't setting this property at all.
- Updated all involved JUnit test timeouts to 30s.
- Overall, reduced execution time by 40s per test.

**Purpose of the change**  
Fixes #5608.

**What the code does**  
See description.

**How to verify it**  
Build must pass.


standalone/src/test/java/io/pravega/local/TlsEnabledInProcPravegaClusterTest.java
```diff
@@ -62,27 +63,29 @@ ClientConfig prepareValidClientConfig() {
                 .build();
     }
 
-    @Override
-    @Test
-    public void testWriteAndReadEventWithValidClientConfig() throws ExecutionException, InterruptedException {
```
andreipaduroiu, line 67: FYI: This was redundant; JUnit will execute this test since it inherits it from the super class.

andreipaduroiu:
Looking at the build execution, the `All Other Unit Tests` duration decreased from ~8m to ~5m40s.

client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
```diff
@@ -48,10 +48,16 @@
     private final ConnectionPool connectionPool;
     private final ScheduledExecutorService executor;
     private final StreamCutHelper streamCutHelper;
-    
+
+
     public StreamManagerImpl(ClientConfig clientConfig) {
+        this(clientConfig, ControllerImplConfig.builder().clientConfig(clientConfig).build());
```
andreipaduroiu, line 54: There are no unit tests that cover this constructor. I had this problem before and it is unfeasible to try to add any meaningful unit tests for this because it creates a real `ControllerImpl` that tries to connect to a real Controller. All the client unit tests use the other constructor that takes in a `Controller` object (which is a mock).

Unfortunately I had to touch this constructor because I needed a way to pass in the controller configuration. A simple review the code can show that I made no functional changes to the existing constructor.

fpj: Ok, I see that this is being tested at least in a standalone integration test. Should we create an issue to see if we can refactor this class to make the constructors testable?

andreipaduroiu: I made #5637 to track this.

shiveshr (approved):

RaulGracia (approved):

PR #5613, opened 2021-02-01
Issue 5393: Logging improvements for Controller

### LINKED ISSUE #5393: Reduce the amount of logging in Controller from Segment Helper class
**Problem description**
Due to frequent logging from Segment helper class, the Controller logs gets overwhelmed thus impacting the cluster storage for maintaining huge logs.

**Problem location**
SegmentHelper

Calls from this class can be avoided at Info level

pbelgundi:

**Change log description**  
Segment Helper and Flow Handler APIs get invoked multiple times during a single Controller operation, esp. during transaction processing. 
Info Logs explaining the details of each Segment Helper / Flow Handler operation cause too many logs to be written for each Controller operation and in some cases cause Controller pods to go out of disk space.

**Purpose of the change**  
Fixes #5393

**What the code does**  
Changed log level for most low-level operations in Segment Helper and Flow Handler, from "INFO" to "DEBUG" since these logs are only needed by developers when debugging.
Changed logback.xml to include a new SizeAndTimeBasedRollingPolicy for logs with a cap on size of all logs written to the Controller pod. That way we can be sure that the disk space occupied by logs on a Controller pod is capped at value specified against `totalSizeCap` in the config.

**How to verify it**  
All tests should pass. 
Size of logs on Controller pod should not exceed the value specified against `totalSizeCap` in logback.xml


controller/src/conf/logback.xml
```diff
@@ -18,16 +18,14 @@
 
     <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>controller-server.log</file>
-        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
```
pbelgundi, line 21: Since `SizeAndTimeBasedFNATP` is now deprecated, changed to `SizeAndTimeBasedRollingPolicy` that also allows specifying a cap on size of all log files put together.

shiveshr (commented):
do we want to reduce a bit of logging for transactions too?
for each transaction info level logs are added for:
1. createtxn. can be changed to debug. 
2. pingtxn. can be changed to debug.
3. committxn. 
4. commit workflow logs several info level messages some of which can be changed to debug.

pbelgundi:
> do we want to reduce a bit of logging for transactions too?
> for each transaction info level logs are added for:
> 
> 1. createtxn. can be changed to debug.
> 2. pingtxn. can be changed to debug.
> 3. committxn.
> 4. commit workflow logs several info level messages some of which can be changed to debug.

I checked all controller log files provided for this issue and I see that ~75% of the printed logs come from these few lines in FlowHandler and SegmentHelper. So in terms of reducing volumes these have the highest impact.
Will take up correction of log levels as part of fix for issue https://github.com/pravega/pravega/issues/5614

controller/src/conf/logback.xml
shrids, line 21: This change will not be used in our docker images. The configuration under dist/conf will be used..

pbelgundi: fixed logback.xml in dist/conf

pbelgundi: Confirmed after logging into docker container  that the updated logback .xml was used.

shiveshr (approved):

dist/conf/logback.xml
```diff
@@ -18,16 +18,14 @@ You may obtain a copy of the License at
...
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
             <!-- daily rollover. Make sure the path matches the one in the file element or else
              the rollover logs are placed in the working directory. -->
             <fileNamePattern>${log.dir:-.}/%d{YYYY-MM-dd, aux}/${log.name:-pravega}_%d{yyyy-MM-dd_HH}.%i.log.gz</fileNamePattern>
-
-            <timeBasedFileNamingAndTriggeringPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedFNATP">
-                <maxFileSize>10MB</maxFileSize>
-            </timeBasedFileNamingAndTriggeringPolicy>
-            <!-- keep 30 days' worth of history -->
+            <!-- each file should be at most 10MB, keep 30 files worth of history, but at most 10GB -->
```
shrids, line 25: Could you confirm that the folder pattern configured in line 24 works with `SizeAndTimeBasedRollingPolicy` ?

shrids (approved):
LGTM apart from one question.
(we can merge it once this is verified...)

dist/conf/logback.xml
pbelgundi: `SizeAndTimeBasedRollingPolicy` is a new policy that replaces the deprecated policy `SizeAndTimeBasedFNATP`. I don't see why FileNamePattern should be impacted by the   API change.

kotlasaicharanreddy: I've tested this and confirm that new policy `SizeAndTimeBasedRollingPolicy` works with the folder pattern configured.

Issue #5553, opened 2021-01-20
Add System Tests for Consumption Based Retention of data on Streams


pbelgundi:
**Problem description**
This issue tracks addition of System tests for the feature Consumption Based Retention of data on Streams.

Parent ticket for the following issues:
- [ ]  #5554 : Test Consumption Based Retention with Single Subscriber and multiple non-subscribers.
- [ ]  #5555 : Test Consumption Based Retention with Multiple Controllers.
- [ ]  #5556 : Test Consumption Based Retention with Controller and Segment Store restarts (failover).
- [ ]  #5557 : Test Consumption Based Retention with Stream Scaling.
- [ ]  #5558 : Test Retention policy change at runtime.
- [ ]  #5559 : Test Consumption based Retention with multiple Subscriber and Non-Subscriber Reader Groups.

**Problem location**
System Tests

**Suggestions for an improvement**
1. List out system tests to be added.
2. Add new tests to System test framework.

RaulGracia:
@shshashwat can we please wrap up this one so we can complete the consumption-based retention story?

ShwethaSNayak:
Added new system tests for all CBR scenario so closing

PR #5617, opened 2021-02-02
Issue 5554: Test Consumption based Stream Retention with Single Subscriber

### LINKED ISSUE #5554: Test Consumption based Stream Retention with Single Subscriber
**Problem description**
_This test verifies that:_
1. Reads of Non-Subscriber Reader Groups do not impact Stream Truncation.
2. Reads of Subscriber Reader Groups impact Stream truncation only when a truncation stream-cut is published to Controller.
3. When truncation stream-cuts are **not** published, stream truncation happens based on min/max limits set in the retention policy.
4. When truncation stream-cuts **are** published by a Subscriber Reader Group, Stream truncation happens based on these stream-cuts.

_Test Steps:_
1. Create a Stream with SPACE based retention policy with min/max limits.
2. Set Controller retention cycle interval to be 1 minute.
3. Create a Non-Subscriber Reader Groups to read data from this Stream. 
4. Create a writer that writes more data to the Stream than the max limit.
5. Non-Subscriber RG starts reading data from the Stream.
6. Verify - After 1 minute, Stream gets truncated at max limit.
7. Create a new Subscriber Reader Group to read from the same Stream and set its RetentionType=MANUAL_RELEASE_AT_USER_STREAMCUT and enable automatic checkpointing.
8. Write more then max_limit data into the Stream and wait for 1 minute.
9. Stream should get truncated at the Stream-Cut that satisfies the max_limit for the Stream's Retention Policy, since we haven't yet published any truncation Stream-Cut.
10. On the Subscriber Reader Group create a new checkpoint using initiateCheckpoint() API
11. Publish this checkpoint as the truncation point for the Stream using updateRetentionStreamCut() API on this ReaderGroup
12. Verify that when the retention cycle runs, the Stream is truncated at this point.
13. Change Subscriber ReaderGroup configuration to RetentionType=AUTOMATIC_RELEASE_AT_LAST_CHECKPOINT.
14. Verify that when the retention cycle runs, the Stream does get truncated at the last-checkpoint stream-cut of the Subscriber ReaderGroup. Note that this will happen only if the published truncation Stream-Cut falls between `min` and `max` space limits set for the Stream. So Stream should be written to accordingly.

**Problem location**
System Tests

Create a new Test Class ConsumptionBasedRetentionTest

anirudhkovuru:
**Change log description**  
Added a new system test to check Consumption based Stream Retention through a single subscriber.

**Purpose of the change**  
Fixes #5554.

**What the code does**  
1. A test stream with a single segment is created with a space based retention policy where the minimum to retain is 30 bytes and the maximum to retain is 300 bytes.
2. An event (e1) of size 30 bytes is written into the stream.
3. A single subscriber, which is subscribed to the test stream, is created. The subscriber reads the event and updates the retention stream-cut with the one after this event (e1) to the controller.
4. Since the size retained by the provided stream-cut would be 0 bytes, truncation does not take place as it is not within the min and max bounds.
5. Two more events (e2 and e3) of size 30 bytes each are written into the stream.
6. Since the size retained by the provided stream-cut would be 60 bytes now, truncation should take place at the provided stream-cut as it is within the min and max bounds.
7. The subscriber reads another event and updates its retention stream-cut to the one after the read event (e2).
8. Since the size retained by the provided stream-cut would be 30 bytes now, truncation should take place at the provided stream-cut as it is within the min and max bounds.

**How to verify it**  
All tests and system tests must pass.


test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,181 @@
...
+                120 * 1000L);
+
+        // Write two more events.
+        log.info("Writing event e2 to {}/{}", SCOPE, STREAM);
+        writer.writeEvent("e2", "data of size 30").join();
+        log.info("Writing event e3 to {}/{}", SCOPE, STREAM);
+        writer.writeEvent("e3", "data of size 30").join();
+
+        // Give the controller a chance to truncate
+        Exceptions.handleInterrupted(() -> Thread.sleep(5 * 60 * 1000));
```
andreipaduroiu, line 158: This makes it a total of 10 minutes.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,181 @@
...
+        // Read next event and update the retention stream-cut.
+        log.info("Reading event e2 from {}/{}", SCOPE, STREAM);
+        reader.readNextEvent(1000);
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        Map<Stream, StreamCut> streamCuts2 = readerGroup.generateStreamCuts(executor).join();
+        log.info("{} updating its retention stream-cut to {}", READER_GROUP, streamCuts2);
+        readerGroup.updateRetentionStreamCut(streamCuts2);
+
+        // Give the controller a chance to truncate
+        Exceptions.handleInterrupted(() -> Thread.sleep(5 * 60 * 1000));
```
andreipaduroiu, line 174: 15 minutes.

andreipaduroiu (changes requested):
Please rewrite your test and/or introduce any necessary config changes into the code to avoid any kind of unnecessary sleeps.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,181 @@
...
+        // Read one event and update the retention stream-cut.
+        log.info("Reading event e1 from {}/{}", SCOPE, STREAM);
+        reader.readNextEvent(1000);
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        Map<Stream, StreamCut> streamCuts = readerGroup.generateStreamCuts(executor).join();
+        log.info("{} updating its retention stream-cut to {}", READER_GROUP, streamCuts);
+        readerGroup.updateRetentionStreamCut(streamCuts);
+
+        // Wait for the retention to kick in.
+        Exceptions.handleInterrupted(() -> Thread.sleep(5 * 60 * 1000));
```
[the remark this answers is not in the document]
anirudhkovuru: I've removed the 5 min timer.

andreipaduroiu (commented):
This looks better. How long does this system test take to run? 

@shiveshr / @pbelgundi  Please take a look at this PR. I cannot remove my Changes Requested without approving it (which will unblock it for merging), so I will await one of your approvals before I approve it too.

RaulGracia:
@anirudhkovuru you will need to change the header for this PR. 
@shiveshr @pbelgundi can you review this one? Looks like we can give a go to it soon.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,213 @@
...
+        // Read one event and update the retention stream-cut.
+        log.info("Reading event e1 from {}/{}", SCOPE, STREAM);
+        reader.readNextEvent(READ_TIMEOUT);
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        Map<Stream, StreamCut> streamCuts = generateStreamCuts(readerGroup);
+        log.info("{} updating its retention stream-cut to {}", READER_GROUP, streamCuts);
+        readerGroup.updateRetentionStreamCut(streamCuts);
+
+        // Check to make sure no truncation happened as the min policy is 30 bytes.
+        AssertExtensions.assertEventuallyEquals(true, () -> controller.getSegmentsAtTime(
```
shiveshr, line 159: this assert will effectively return immediately and we do not give it time to actually truncate. so we dont really verify anything at this step. 

verifying truncation doesnt happen is hard without relying on time.. 
so we might as well remove this check here.

shiveshr: also, anymatch offset=0 is not ideal either if tehre are more than one segments and all data got written to only one of them.. then only that one would have truncated.. 

your intent here should have been `allmatch` offset = 0

shiveshr: i know your configuration has only 1 segment.. but generally allmatch in this case is better representative of the logic in the code.

shiveshr (approved):
One minor comment

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+        @Cleanup
+        ReaderGroupManager readerGroupManager = ReaderGroupManager.withScope(SCOPE, clientConfig);
+        readerGroupManager.createReaderGroup(READER_GROUP, ReaderGroupConfig.builder()
+                .retentionType(ReaderGroupConfig.StreamDataRetention.MANUAL_RELEASE_AT_USER_STREAMCUT)
+                .disableAutomaticCheckpoints()
+                .stream(Stream.of(SCOPE, STREAM)).build());
+        ReaderGroup readerGroup = readerGroupManager.getReaderGroup(READER_GROUP);
+        @Cleanup
+        EventStreamReader<String> reader = clientFactory.createReader(READER_GROUP + "-" + String.valueOf(1),
```
RaulGracia, line 142: `String.valueOf(1)` this is not needed, just use `1` and the concatenation will make it String automatically.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,181 @@
...
+                READER_GROUP, new JavaSerializer<>(), readerConfig);
+
+        // Read one event and update the retention stream-cut.
+        log.info("Reading event e1 from {}/{}", SCOPE, STREAM);
+        reader.readNextEvent(READ_TIMEOUT);
```
RaulGracia, line 147: Why don't you check that what you are reading is the right data? It could be even a checkpoint, right?

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,181 @@
...
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        CompletableFuture<Map<Stream, StreamCut>> futureCuts = readerGroup.generateStreamCuts(executor);
+        EventRead<String> emptyEvent = reader.readNextEvent(100);
```
RaulGracia, line 150: What is this used for?

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,181 @@
...
+        AssertExtensions.assertEventuallyEquals(true, () -> controller.getSegmentsAtTime(
+                new StreamImpl(SCOPE, STREAM), 0L).join().values().stream().anyMatch(off -> off >= 30),
+                5 * 60 * 1000L);
+
+        // Read next event and update the retention stream-cut.
+        log.info("Reading event e2 from {}/{}", SCOPE, STREAM);
+        reader.readNextEvent(READ_TIMEOUT);
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        CompletableFuture<Map<Stream, StreamCut>> futureCuts2 = readerGroup.generateStreamCuts(executor);
+        EventRead<String> emptyEvent2 = reader.readNextEvent(100);
```
RaulGracia, line 171: Again, why do you need this?

RaulGracia (changes requested):
@anirudhkovuru are you still pursuing this one?

anirudhkovuru:
The following system test run for this branch passed: https://asd-nautilus-jenkins.isus.emc.com/job/pravega-branch-builder/1296/

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
anirudhkovuru: changed it to `1`

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
anirudhkovuru: There are checks to make sure the data read is what is expected

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
anirudhkovuru: In order to trigger the silent checkpoint generated during `generateStreamCuts` to complete we read another event after the `ReaderGroupState` updates itself.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
anirudhkovuru: Explained above.

RaulGracia (commented):
I have left some comments, but I'm not blocking the PR. Also, it would be interesting for this test to be a bit more realistic. Wondering if you could add another test method (in the same class) that could write/consume more data and with parallel threads. The current test is super small and unit-like, and for system tests we can try more realistic loads and validate the semantics from an application perspective.

RaulGracia (commented):
Also, it would be interesting for this test to be a bit more realistic. Wondering if you could add another test method (in the same class) that could write/consume more data and with parallel threads. The current test is super small and unit-like, and for system tests we can try more realistic loads and validate the semantics from an application perspective.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,202 @@
...
+@Slf4j
+@RunWith(SystemTestRunner.class)
+public class BasicCBRTest extends AbstractReadWriteTest {
+
+    private static final String SCOPE = "testCBRScope" + RandomFactory.create().nextInt(Integer.MAX_VALUE);
+    private static final String STREAM = "testCBRStream" + RandomFactory.create().nextInt(Integer.MAX_VALUE);
+    private static final String READER_GROUP = "testCBRReaderGroup" + RandomFactory.create().nextInt(Integer.MAX_VALUE);
+
+    private static final int READ_TIMEOUT = 1000;
+    private static final int MAX = 300;
```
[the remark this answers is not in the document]
anirudhkovuru: Using self describing names

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,202 @@
...
+        // Write two more events.
+        log.info("Writing event e2 to {}/{}", SCOPE, STREAM);
+        writer.writeEvent("e2", "data of size 30").join();
+        log.info("Writing event e3 to {}/{}", SCOPE, STREAM);
+        writer.writeEvent("e3", "data of size 30").join();
+
+        // Check to make sure truncation happened after the first event.
+        AssertExtensions.assertEventuallyEquals(true, () -> controller.getSegmentsAtTime(
+                new StreamImpl(SCOPE, STREAM), 0L).join().values().stream().anyMatch(off -> off >= 30),
+                5 * 60 * 1000L);
```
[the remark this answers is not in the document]
anirudhkovuru: The retention cycle takes place periodically every 2 min. By waiting for 5 min we allow the retention cycle to run twice, makes sure that the truncation we wish to see is guaranteed to happen.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,202 @@
...
+        EventRead<String> emptyEvent = reader.readNextEvent(100);
+        assertTrue("Stream-cut generation did not complete", Futures.await(futureCuts, 10_000));
+
+        Map<Stream, StreamCut> streamCuts = futureCuts.join();
+        log.info("{} updating its retention stream-cut to {}", READER_GROUP, streamCuts);
+        readerGroup.updateRetentionStreamCut(streamCuts);
+
+        // Write two more events.
+        log.info("Writing event e2 to {}/{}", SCOPE, STREAM);
+        writer.writeEvent("e2", "data of size 30").join();
```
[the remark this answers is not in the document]
anirudhkovuru: Used a variable

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
anirudhkovuru: Fixed that.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,203 @@
...
+        log.info("Reading event e1 from {}/{}", SCOPE, STREAM);
+        EventRead<String> read = reader.readNextEvent(READ_TIMEOUT);
+        assertFalse(read.isCheckpoint());
+        assertEquals("data of size 30", read.getEvent());
+
+        // Update the retention stream-cut.
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        CompletableFuture<Map<Stream, StreamCut>> futureCuts = readerGroup.generateStreamCuts(streamCutExecutor);
+        // Wait for 5 seconds to force reader group state update.
+        Exceptions.handleInterrupted(() -> TimeUnit.SECONDS.sleep(5));
```
pbelgundi, line 161: I don't like the idea of adding sleeps to tests as it increases possibility of making tests flaky. Why do we need to wait for reader group state update here?

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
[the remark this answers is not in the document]
pbelgundi: The frequency with which Controller executes the retention cycle is a configurable parameter that can be specified in ControllerServiceConfigImpl. See: https://github.com/pravega/pravega/blob/master/controller/src/main/java/io/pravega/controller/server/impl/ControllerServiceConfigImpl.java#L106
When not explicitly set this value defaults to Config.MINIMUM_RETENTION_FREQUENCY_IN_MINUTES which is 1 min...
So there should not be a need to wait for 5 mins...

pbelgundi (changes requested):

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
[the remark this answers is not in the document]
pbelgundi: Another option can be to AssertEventuallyEquals() that head of the Stream is equal to the StreamCut at which we expect truncation to happen. There should be no need to specify any time interval...

pbelgundi:
> t could write/consume more data and with parallel threads. The current test is super small and unit-like, and for system

This is supposed to be a basic CBR test that covers 1 most basic CBR scenario. More tests with multiple reader groups and more complex scenarios would be added later. See: https://github.com/pravega/pravega/issues/5553

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
anirudhkovuru: Calling `generateStreamCuts` leads to the creation of a silent checkpoint. In order for the ReaderGroup to process the checkpoint we need to wait and allow the ReaderGroupState to update and fetch the checkpoint so that it can be processed. This is present in the form of a timer (`checkpointTimer`) in the `ReaderGroupStateManager`.

test/system/src/test/java/io/pravega/test/system/BasicCBRTest.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@Slf4j
+@RunWith(SystemTestRunner.class)
+public class BasicCBRTest extends AbstractReadWriteTest {
```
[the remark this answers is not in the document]
anirudhkovuru: Changed the name accordingly.

anirudhkovuru:
Regarding the 5 minute wait for periodic retention to kick in. This was how it was implemented in other system tests. So the aspect of dealing with that wait can be dealt with in different issue where we can address all such cases at once. The 5 second reader group state update timeout can be dealt with similarly.

anirudhkovuru:
A successful system test run: https://asd-nautilus-jenkins.isus.emc.com/job/pravega-branch-builder/1499/

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
```diff
@@ -0,0 +1,209 @@
...
+    private static final String READER_GROUP = "testCBRReaderGroup" + RandomFactory.create().nextInt(Integer.MAX_VALUE);
+    private static final String SIZE_30_EVENT = "data of size 30";
+
+    private static final int READ_TIMEOUT = 1000;
+    private static final int MAX_SIZE_IN_STREAM = 300;
+    private static final int MIN_SIZE_IN_STREAM = 30;
+
+    private final ReaderConfig readerConfig = ReaderConfig.builder().build();
+    private final ScheduledExecutorService executor = ExecutorServiceHelpers.newScheduledThreadPool(4, "executor");
+    private final ScheduledExecutorService streamCutExecutor = ExecutorServiceHelpers.newScheduledThreadPool(1, "streamCutExecutor");
```
RaulGracia, line 82: Why do you need 2 executors?

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
```diff
@@ -0,0 +1,209 @@
...
+        assertTrue("Creating stream", streamManager.createStream(SCOPE, STREAM,
+                StreamConfiguration.builder()
+                        .scalingPolicy(ScalingPolicy.fixed(1))
+                        .retentionPolicy(RetentionPolicy.bySizeBytes(MIN_SIZE_IN_STREAM, MAX_SIZE_IN_STREAM)).build()));
+    }
+
+    @After
+    public void tearDown() {
+        streamManager.close();
+        ExecutorServiceHelpers.shutdown(executor);
```
RaulGracia, line 123: You have actually missed to shut down the second executor.

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
```diff
@@ -0,0 +1,209 @@
...
+        EventRead<String> read = reader.readNextEvent(READ_TIMEOUT);
+        assertFalse(read.isCheckpoint());
+        assertEquals("data of size 30", read.getEvent());
+
+        // Update the retention stream-cut.
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        CompletableFuture<Map<Stream, StreamCut>> futureCuts = readerGroup.generateStreamCuts(streamCutExecutor);
+        // Wait for 5 seconds to force reader group state update. This will allow for the silent
+        // checkpoint event generated as part of generateStreamCuts to be picked and processed.
+        Exceptions.handleInterrupted(() -> TimeUnit.SECONDS.sleep(5));
```
RaulGracia, line 162: Let's do not add more sleeps if possible. If the update takes 6 seconds this test will fail, and if it takes 1 second we are wasting 4 seconds. Please, use `assertEventuallyEquals` if possible, as it would check the state of the reader group until the condition you want is true.

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
```diff
@@ -0,0 +1,209 @@
...
+        // Write two more events.
+        log.info("Writing event e2 to {}/{}", SCOPE, STREAM);
+        writer.writeEvent("e2", SIZE_30_EVENT).join();
+        log.info("Writing event e3 to {}/{}", SCOPE, STREAM);
+        writer.writeEvent("e3", SIZE_30_EVENT).join();
+
+        // Check to make sure truncation happened after the first event.
+        // The timeout is 5 minutes as the retention period is set to 2 minutes. We allow for 2 cycles to fully complete
+        // and a little longer in order to confirm that the retention has taken place.
+        AssertExtensions.assertEventuallyEquals(true, () -> controller.getSegmentsAtTime(
```
RaulGracia, line 179: I think that this is checking every few milliseconds by default. Maybe you can pick a less aggressive evaluation interval, given that this is going to take at least a minute.

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
```diff
@@ -0,0 +1,209 @@
...
+        read = reader.readNextEvent(READ_TIMEOUT);
+        assertFalse(read.isCheckpoint());
+        assertEquals("data of size 30", read.getEvent());
+
+        // Update the retention stream-cut.
+        log.info("{} generating stream-cuts for {}/{}", READER_GROUP, SCOPE, STREAM);
+        CompletableFuture<Map<Stream, StreamCut>> futureCuts2 = readerGroup.generateStreamCuts(streamCutExecutor);
+        // Wait for 5 seconds to force reader group state update. This will allow for the silent
+        // checkpoint event generated as part of generateStreamCuts to be picked and processed.
+        Exceptions.handleInterrupted(() -> TimeUnit.SECONDS.sleep(5));
```
RaulGracia, line 194: Same here about sleeps.

RaulGracia (changes requested):

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
anirudhkovuru: One executor handles the controller, the other executor is for the client to process the silent checkpoint generated as part of the `generateStreamCut` process. The same has been done in the `StreamCutsTest` system test.

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
anirudhkovuru: I added the shutdown

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
anirudhkovuru: This is being used in other tests too so we can try dealing with this as a separate issue.

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
anirudhkovuru: Similarly for this timeout.

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @After
+    public void tearDown() {
+        streamManager.close();
+        ExecutorServiceHelpers.shutdown(executor);
+        ExecutorServiceHelpers.shutdown(streamCutExecutor);
+    }
+
+    @Test
+    public void basicCBRTest() throws Exception {
```
pbelgundi, line 128: Can we call this singleSubscriberCBR ?

anirudhkovuru: I've changed this name.

test/system/src/test/java/io/pravega/test/system/SingleSubscriberUpdateRetentionStreamCutTest.java
anirudhkovuru: I've changed the check interval to 1 min.

RaulGracia: Well, if this is going to take 1 minute, what can happen is that you evaluate one second before the condition evaluates to true, and then the test waits for 1 minute doing nothing. Evaluating every 1 or 5 seconds should be fine.

anirudhkovuru: I've changed it to 1 sec

RaulGracia (approved):

andreipaduroiu (approved):

anirudhkovuru:
The system test run is green for this PR. The build is 1561 on pravega-branch-builder and 4083 on test-pravega-continuous.

RaulGracia:
@pbelgundi can you please have another look? This seems close to get merged.

pbelgundi (approved):
LGTM

Issue #955, opened 2017-04-06
Support for persistent read-write locks in Task module


bhargav-gulavani:
**Problem description**
1. Currently locks in task module are exclusive locks. There is a need for read-write locks, especially for creating transactions, we just need read lock on the stream, so as to support concurrent creation of multiple transactions. However, we do not want the stream metadata, especially current segments, to be updated while a transaction is being created. Other update operations on the stream like scale, delete, need need write lock on the stream. 
2. If a lock cannot be acquired, currently task module fails the operation and returns `LockFailedException` back to the caller. In response to `LockFailedException`, the caller is expected to retry the operation. It would be better, if, as with classical locks, our lock request waits until other conflicting requests are complete and return once a lock is actually acquired.

**Problem location**
Controller Task module.

**Suggestions for an improvement**
1. Provide support for read write lock using ZK persistent nodes
2. If a lock cannot be acquired immediately, lock operation should wait until other conflicting locks are released to acquire the lock and then return.

PR #5623, opened 2021-02-03
Issue 5622: Bug fix - Enforce max limit for retention correctly and handle overlaps with min bound optimally

### LINKED ISSUE #5622: Enforce max limit on retention policy correctly
**Problem description**
The retention policy enforcement has a bug where we take the latest streamcut above max limit for choosing the max bound. Instead we should choose the oldest streamcut below max limit from the retention set. 
Secondly, the scenario where subscriber lowerbound streamcut overlaps with min bound, we should choose the streamcut just before the subscriber lowerbound instead of max to get the best possible retention outcome. 
Thirdly, for time based retention, with the changed definition of max streamcut bound, we cannot consider any subscriber streamcut before max bound to be breaching the time limits. so we need to still check against a streamcut that was outside of the max bound to determine if subscriber streamcut is indeed breaching the max time limit conclusively or not. 

**Problem location**
StreamMetadataTasks

Fix as suggested

shiveshr:

**Change log description**  
- Fix the max bound stream cut identification logic to choose a streamcut which satisfies the max limit. 
- Handle max bound streamcut comparison with subscriber lowerbound to determine if the subscriber lowerbound is conclusively breaching the time bounds from the policy
- Fix the issue with handling of subscriber lowerbound overlapping with min bound streamcut and choose most recent streamcut before lowerbound for truncation

**Purpose of the change**  
Fixes #5622 

**What the code does**  
In streammetadatatasks, we modify the size and time based retention logic to take the correct max bounds.
since the semantics of max bound changes, the relative position of subscriber lowerbound vis a vis the policy limits, particularly for time limits also changes and has been handled in following fashion:
```
Streamcuts collected in retention set:
T+0: sc0   
T+1: sc1  
T+2: sc2
lets says policy.max = 2

say subscriber lowerbound is slb. 
```
Since subscriber lowerbound doesnt have a logical time associated with it, unless slb is before sc0 we cannot conclusively say that it is breaching the policy.max. 

Optimization and fix in overlap case with Min bound:
```
lets say we have a size based retention policy where min = 2, max = 10. 
say stream has 2 segments: seg0 and seg1. 
tail streamcut: seg0/1, seg1/10
sc0: seg0/1, seg1/8 ==> size retained if truncated at = 2
sc1: seg0/1, seg1/7 ==> size retained if truncated at = 3
sc2: seg0/0, seg1/7 ==> size retained if truncated at = 4
sc3: seg0/0, seg1/6 ==> size retained if truncated at = 5
sc4: seg0/0, seg1/0 ==> size retained if truncated at = 11 
subscriber streamcut: slb: seg0/0, seg1/10 ==> size retained if truncated at = 1
```
Here slb overlaps with min (and max). Also, we cannot truncate at SLB because it will break the min policy. We could trunacte at sc3 (max bound) which is safe but not optimal. optimal is sc2 which is behind SLB. So i have modified the logic to pick the correct and most optimal streamcut. 

Modified the greaterThan method for comparing streamcuts with a `compare` method in StreamMetadataStore which returns 1 0 or -1 for when greaterthanequal, overlapping, and lessthan respectively. 

**How to verify it**  
Unit tests added to cover each of the scenarios.


controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -249,14 +249,22 @@ private boolean greaterThan(Map<Long, Long> cut1, Map<StreamSegmentRecord, Integ
...
         CompletableFuture<ImmutableMap<StreamSegmentRecord, Integer>> span1Future = computeStreamCutSpan(streamcut1);
         CompletableFuture<ImmutableMap<StreamSegmentRecord, Integer>> span2Future = computeStreamCutSpan(streamcut2);
         return CompletableFuture.allOf(span1Future, span2Future)
                     .thenApply(v -> {
                         ImmutableMap<StreamSegmentRecord, Integer> span1 = span1Future.join();
                         ImmutableMap<StreamSegmentRecord, Integer> span2 = span2Future.join();
-                        return greaterThan(streamcut1, span1, streamcut2, span2);
+                        boolean gt = greaterThan(streamcut1, span1, streamcut2, span2);
+                        boolean lt = greaterThan(streamcut2, span2, streamcut1, span1);
+                        if (gt) { // if its greater then return 1
```
pbelgundi, line 261: Expressing "Greater", "Overlapping" & "Lesser" values as enums could make the code more readable compared to using integers.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -249,14 +249,22 @@ private boolean greaterThan(Map<Long, Long> cut1, Map<StreamSegmentRecord, Integ
...
-    public CompletableFuture<Boolean> isStreamCutStrictlyGreaterThan(Map<Long, Long> streamcut1, Map<Long, Long> streamcut2) {
+    public CompletableFuture<Integer> compareStreamCuts(Map<Long, Long> streamcut1, Map<Long, Long> streamcut2) {
         CompletableFuture<ImmutableMap<StreamSegmentRecord, Integer>> span1Future = computeStreamCutSpan(streamcut1);
         CompletableFuture<ImmutableMap<StreamSegmentRecord, Integer>> span2Future = computeStreamCutSpan(streamcut2);
         return CompletableFuture.allOf(span1Future, span2Future)
                     .thenApply(v -> {
                         ImmutableMap<StreamSegmentRecord, Integer> span1 = span1Future.join();
                         ImmutableMap<StreamSegmentRecord, Integer> span2 = span2Future.join();
-                        return greaterThan(streamcut1, span1, streamcut2, span2);
+                        boolean gt = greaterThan(streamcut1, span1, streamcut2, span2);
```
pbelgundi, line 259: Instead of 2 method calls to `greaterThan` here with reverse parameters, we could perhaps have another method like compareCuts() that makes the same 2 calls as here and returns the appropriate enum.

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
```diff
@@ -685,15 +685,19 @@
     CompletableFuture<Integer> getSegmentSealedEpoch(long segmentId);
 
     /**
-     * Method to compare streamcuts to check if streamcut1 is strictly ahead of streamcut2. 
-     * Strictly means if the two streamcuts are overlapping for any range, then this method will reply in negative. 
+     * Compares two Stream cuts and returns :
+     *  1 if streamcut1 is ahead of or identical to streamcut2.
```
pbelgundi, line 689: > 1 if streamcut1 is ahead of or identical to streamcut2.
So `1` means S1 is strictly ahead of S2? or orverlaps with S2?... but overlap should be `0`...a bit confused about this...

controller/src/main/java/io/pravega/controller/store/stream/StreamMetadataStore.java
```diff
@@ -1441,12 +1443,12 @@
...
      *                   
-     * @return A completable future which when completed will hold a boolean which will indicate if streamcut1 is strictly
-     * ahead of streamcut2. 
+     * @return A completable future which when completed will hold an integer which will determine the order between 
+     * streamcut1 and streamcut2. 
      */
-    CompletableFuture<Boolean> streamCutStrictlyGreaterThan(final String scope, final String streamName,
-                                                            Map<Long, Long> streamCut1, Map<Long, Long> streamCut2,
-                                                            final OperationContext context, final Executor executor);
+    CompletableFuture<Integer> compareStreamCut(final String scope, final String streamName,
```
pbelgundi, line 1449: we could consider having a comparator for streamcuts.

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
shiveshr: 0: overlap means one streamcut has some segments where it is greater-than and on some it is less-than another streamcut. 
1: means streamcut is greater than or eq to another streamcut on all segments

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: the compare method is the one thats calling these two methods back to back and making use of their result.
this is not a costly method, its a simple in memory comparison of two streamcuts.

we could create a common method which will need to do the same comparisons that these two methods do and there wouldnt be much room for optimizing that.

controller/src/main/java/io/pravega/controller/store/stream/StreamMetadataStore.java
shiveshr: comparator's definition is 0 = equal. there is no value for overlap there. 
by using comparator interface we would be using it incorrectly for the definitions.

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
pbelgundi: This is precisely why I think it may be better to use enums instead of 0,1 values, so we can better explain what each value implies.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: Yes a common method would be better though internally it would make the same calls as is being done here, it keeps the code more readable. It may be good to add a javadoc comment to this method and that new method explaining what each one achieves.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
 
     private CompletableFuture<Map<Long, Long>> getTruncationStreamCutBySizeLimit(String scope, String stream, OperationContext context, RetentionPolicy policy,
                                                                                  RetentionSet retentionSet, Map<Long, Long> lowerBound) {
-        // if the lowerbound on subscribers streamcuts satisfies the policy size bound, then return it. 
-        // else return the stream cut that satisfies maximum bound on size. 
         // 1. if lowerbound.size < max and lowerbound.size > min truncate at lowerbound
-        // 2. if lowerbound.size < min, truncate at max irrespective of if lowerbound overlaps with max or not. 
+        // 2. if lowerbound.size < min, truncate at streamcut behind lowerbound that satisfies the policy. (we look at max and below).
```
pbelgundi, line 948: Can't seem to wrap my head easily around terms "behind" and "forward" here... 
Could we instead reference everything w.r.t the tail of the Stream?
If `lowerbound.size < min` I thought we should be truncating at a stream-cut closest-to the `min` but with size > min. (further from the tail than min)

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
         // 3. if lowerbound.size > max, truncate at max
         long currentSize = retentionSet.getLatest().getRecordingSize();
+        Map.Entry<StreamCutReferenceRecord, StreamCutReferenceRecord> limits = getBoundStreamCuts(policy, retentionSet,
```
pbelgundi, line 951: It would be good to explain what this line of code does or have atleast have some  method level comments on `getBoundStreamCuts` explaining what that method takes, computes and returns.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
+                x -> currentSize - x.getRecordingSize());
 
         // if lowerbound is empty simply return min
         if (lowerBound == null || lowerBound.isEmpty()) {
-            return retentionSet
-                    .getRetentionRecords().stream()
-                    .filter(x -> currentSize - x.getRecordingSize() > policy.getRetentionParam())
-                    .max(Comparator.comparingLong(StreamCutReferenceRecord::getRecordingTime))
+            return Optional.ofNullable(limits.getValue())
```
pbelgundi, line 956: This perhaps means if there is no subscriber.lowerbound, we truncate at a Stream-Cut that satisfies the `min`. What does `limits` have?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
         // from the retention set then we can safely truncate at lowerbound. Else we will truncate at the max time bound if it
         // exists
-        // 1. if LB > min => truncate at min
-        // 2. if LB < max => truncate at max
-        // 3. if LB < min && LB > max => truncate at LB
-        // 4. if LB < min && overlaps max => truncate at max
+        // 1. if LB is ahead of min => truncate at min
+        // 2. if LB is behind max => truncate at max
+        // 3. if LB is behind min && LB is ahead of max => truncate at LB
+        // 4. if LB is behind min && overlaps max => truncate at max
```
pbelgundi, line 1016: `4. if LB is behind min && overlaps max => truncate at max`
Here, why not truncate at `LB` instead of truncate at max?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
+        // 5. if LB overlaps with min and max ==> so its got both recent data and older data.. what do we do..
```
pbelgundi, line 1017: Here we should truncate at a stream-cut that would make the stream size closest to but **lesser than** `max` post truncation. where are the 2 options?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
+                .max(Comparator.comparingLong(StreamCutReferenceRecord::getRecordingTime));
 
         CompletableFuture<StreamCutRecord> limitMaxFuture = limits.getKey() == null ? CompletableFuture.completedFuture(null) :
                 streamMetadataStore.getStreamCutRecord(scope, stream, limits.getKey(), context, executor);
-        return CompletableFuture.allOf(limitMaxFuture, limitMinFuture)
+        CompletableFuture<StreamCutRecord> maxBoundFuture = maxBoundRef.map(x -> streamMetadataStore.getStreamCutRecord(scope, stream, x, context, executor))
+                                                                    .orElse(CompletableFuture.completedFuture(null));
+        return CompletableFuture.allOf(limitMaxFuture, limitMinFuture, maxBoundFuture)
                 .thenCompose(v -> {
                     StreamCutRecord limitMax = limitMaxFuture.join();
```
pbelgundi, line 1037: My understanding is that `limitMax` and `limitMin` are StreamCut Records that satisfy the min/max criteria for the 
 Stream. It may be good to explain that with a comment.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
+                                           .map(x -> streamMetadataStore.getStreamCutRecord(scope, stream, x, context, executor)
+                                                                        .thenApply(StreamCutRecord::getStreamCut))
+                                    .orElse(CompletableFuture.completedFuture(lowerBound));
                         }
-                      }
+                    }
                 });
     }
 
     private CompletableFuture<Map<Long, Long>> getTruncationStreamCutByTimeLimit(String scope, String stream, OperationContext context,
```
pbelgundi, line 1003: Inside this method it may be good to explain the flow of code at significant points. The logic involved here is very complex and very difficult to understand/recollect without any comments in the code.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
+        // 3. if LB is behind min && LB is ahead of max => truncate at LB
+        // 4. if LB is behind min && overlaps max => truncate at max
+        // 5. if LB overlaps with min and max ==> so its got both recent data and older data.. what do we do.. 
+        //      there are two options: either truncate beyond max or not truncate at all.  
         CompletableFuture<StreamCutRecord> limitMinFuture = limits.getValue() == null ? CompletableFuture.completedFuture(null) :
                 streamMetadataStore.getStreamCutRecord(scope, stream, limits.getValue(), context, executor);
 
         // if lowerbound is empty simply return min
         if (lowerBound == null || lowerBound.isEmpty()) {
             return limitMinFuture.thenApply(min -> Optional.ofNullable(min).map(StreamCutRecord::getStreamCut).orElse(null));
```
pbelgundi, line 1024: This is same as `// 1. if LB is ahead of min => truncate at min` so it would be good to state that here.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -944,118 +944,158 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
-                    if (retainedSize < policy.getRetentionParam()) {
-                        // if retainedSize is less than min size then truncate the stream at max bound.
-                        return maxBound.get().map(x -> streamMetadataStore.getStreamCutRecord(scope, stream, x, context, executor)
-                        .thenApply(StreamCutRecord::getStreamCut))
-                        .orElse(CompletableFuture.completedFuture(null));
-                        } else {
+                .thenCompose(sizeTillLB -> {
+                    long retainedSizeLB = currentSize - sizeTillLB;
+                    // if retainedSize is less than min size then we need to truncate at min or the most recent streamcut
+                    // strictly behind lowerbound. we actually choose the max
```
pbelgundi, line 965: Didn't get this part of the comment: 
`the most recent streamcut strictly behind lowerbound. we actually choose the max`

I thought it should be truncated at a stream-cut that satisfies min

pbelgundi:
As discussed earlier, since we're changing the interpretation of `Max` with this PR we should also change the default value for MAX, so we can tell from looking at the RetentionPolicy if the User has set a max value or not.
We should not default it to same value as `min` when the user does not provide a max value.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: we truncate at min if it doesnt overlap with SLB.. if it does, we choose the first streamcut from retentionset strictly behind SLB.
the comment about max is not correct.. removed it.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: this is checking if SLB is available or not. if there is no SLB we always truncate at min. 
in presence of SLB all computations come into equation. SLB ahead of min is handled later and there is comment around it describing the case.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: i looked at the comments.. and i see comments describing each step in the code along the way.. i understand that some of it may not be clear.. but is there any part specifically that do you want described with more clarity.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: the next line in the comment gives out two options:
        //      there are two options: either truncate beyond max or not truncate at all.  

and we we truncate beyond max. i have updated the comment to reflect that.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: if it overlaps with max, it means it has some data which is older than max while some data which is newer than max.
since our limit on time is satisfied by `max` we wouldnt want to retain older data than that.. hence truncating at max which satisfies the criteria more accurately than truncating at LB

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: limits has min and max.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: replaced behind with less than and forward with greater than.. the terminology we use elsewhere

shiveshr: >  Could we instead reference everything w.r.t the tail of the Stream?
how do you mean? 

basically we compare two streamcuts to determine if one streamcut is greater than or less than the other.. 
greater than means it points to more recent data. less than means older data.. overlapping means its pointing to some older data and some newer data.

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
shiveshr: 1) that enum will only be used for this method which is an internal method. 2) the javadoc of enum or the javadoc for the method signature will capture this exact same comment describing what each of the responses mean. so i do not see how enum may be better suited here. but i can make it an enum regardless.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: so a common method is just a wrapper over these two methods, right?.. isnt this method exactly that in a way?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: I now understand that `getBoundStreamCuts` returns stream-cuts that satisfy the **min** and **max** bounds for the Stream, but it is necessary to explain in the javadoc **how** we arrive at those. It is important to clarify that we calculate all sizes from the current tail of the Stream. While this may be mentioned in the design & user documentation having it in the code keeps the code easier to understand.
For example, If the min limit is 10 GB, there may be stream-cuts, truncating at which will give us sizes of 9.8 GB or 12GB... but here we choose "atleast 10GB" so we select the stream-cut that retains 12GB, though 9.8 GB is _closer_ to the min limit.
However for max, if the max limit is say 50 GB and we have stream-cuts that would keep 45 GB and 51 GB, we choose the one with 45GB because max is "atmost", regardless of the fact that the Stream-cut retaining 51G is closer to the expected max.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -946,118 +946,169 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
     private CompletableFuture<Map<Long, Long>> getTruncationStreamCutBySizeLimit(String scope, String stream, OperationContext context, RetentionPolicy policy,
                                                                                  RetentionSet retentionSet, Map<Long, Long> lowerBound) {
-        // if the lowerbound on subscribers streamcuts satisfies the policy size bound, then return it. 
-        // else return the stream cut that satisfies maximum bound on size. 
         // 1. if lowerbound.size < max and lowerbound.size > min truncate at lowerbound
-        // 2. if lowerbound.size < min, truncate at max irrespective of if lowerbound overlaps with max or not. 
+        // 2. if lowerbound.size < min, truncate at streamcut less than lowerbound that satisfies the policy. 
         // 3. if lowerbound.size > max, truncate at max
         long currentSize = retentionSet.getLatest().getRecordingSize();
+        // we get the streamcuts from retentionset that satisfy the min and max bounds with min pointing to most recent
```
pbelgundi, line 953: "with min pointing to most recent" what does that mean?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -946,118 +946,169 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
 
     private CompletableFuture<Map<Long, Long>> getTruncationStreamCutByTimeLimit(String scope, String stream, OperationContext context,
                                                                                  RetentionPolicy policy, RetentionSet retentionSet,
                                                                                  Map<Long, Long> lowerBound) {
-        Map.Entry<StreamCutReferenceRecord, StreamCutReferenceRecord> limits =
-                getBoundStreamCuts(policy, retentionSet);
-        // if subscriber lowerbound is ahead of streamcut corresponding to the max time and is behind stream cut for min time
+        long currentTime = retentionClock.get().get();
+
+        // we get the streamcuts from retentionset that satisfy the min and max bounds with min pointing to most recent
```
pbelgundi, line 1016: This line essentially is the same as line #955 in getRetentionStreamCutsBySizeLimit().
But the  comments don't seem to match....it may be good to just explain at the method definition of getBoundStreamCuts() what this method does and what values it returns.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -946,118 +946,169 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
+                                                 case LessThan: // min less than (behind) lowerbound. truncate at min
+                                                     return CompletableFuture.completedFuture(limitMin.getStreamCut());
+                                                 default:
+                                                     // we cannot have min greater (ahead) than lowerbound as retainedSizeLB < min.
+                                                     // so this is the overlapping case. we need to find streamcut before lowerbound
+                                                     // and truncate at it.
+                                                     // since this is a caught up bound (meaning truncating at subscribers LB will break min policy), 
+                                                     // we cannot force truncate at the max if max overlapped with LB.
+                                                     // so we will find the streamcut from retention set before/lessThan LB (and since it overlaps with min 
+                                                     // so it will definitely be before/lessThan min).
```
pbelgundi, line 988: We'd never want to truncate at a stream-cut before/lesser-than than min.
The concepts of before/after, behind/ahead are very confusing to grasp here.
It may be good to explain the _terms to describe relative positions of stream-cuts_  ( ex:  `behind/ahead or before/after or greaterThan/lessThan` ) in some method definition and then use just 1 set of terms -- using multiple of these interchangably makes things very confusing.
My guess is :
behind/before/lesserThan == further from the Stream tail
ahead/after/greaterThan == closer to the Stream tail.
So `SC1` is greaterThan `SC2` means `SC1` is closer to the Stream tail than `SC2`

controller/src/main/java/io/pravega/controller/store/stream/StreamCutComparison.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Enumeration that encapsulates the result of comparison of two streamcuts.
+ * Two streamcuts are compared by comparing each segment and offset in each of the streamcuts with the corresponding 
+ * overlapping segments in the second streamcut. 
+ * A segment X from Streamcut SC1 is considered being ahead of segment Y in streamcut SC2 if either X is a successor of Y OR
+ * both X and Y are same segments and the offset in SC1 for X is greater than offset for X in SC2. 
+ */
+public enum StreamCutComparison {
+    // Indicates that for each segment S in streamcut SC1 there exists either the same segment with same or lesser offset
```
pbelgundi, line 20: Is `S` a Segment? or a StreamCut? Can we explain this in a more simple way?

controller/src/main/java/io/pravega/controller/store/stream/StreamCutComparison.java
```diff
@@ -0,0 +1,31 @@
...
+ * Enumeration that encapsulates the result of comparison of two streamcuts.
+ * Two streamcuts are compared by comparing each segment and offset in each of the streamcuts with the corresponding 
+ * overlapping segments in the second streamcut. 
+ * A segment X from Streamcut SC1 is considered being ahead of segment Y in streamcut SC2 if either X is a successor of Y OR
+ * both X and Y are same segments and the offset in SC1 for X is greater than offset for X in SC2. 
+ */
+public enum StreamCutComparison {
+    // Indicates that for each segment S in streamcut SC1 there exists either the same segment with same or lesser offset 
+    // in streamcut SC2 OR all overlapping segments of S in streamcut SC2 are all predecessors of S. 
+    GreaterThanEquals,
```
pbelgundi, line 22: How is `GreaterThanEquals` different from `overlap` ?

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: It is important to go back to definition of a stream cut. Then it becomes obvious that greater than after and ahead mean same thing. 
a stream cut is an artifact that splits the stream into two halves. So logically each event position will either be in first half or second half or on the stream cut.
Now before/behind etc are events that were written before this stream cut.. ahead/after/greater than are events that were written after this stream cut.

controller/src/main/java/io/pravega/controller/store/stream/StreamCutComparison.java
shiveshr: Segment s in stream cut sc1.
Stream cut is described as offsets on different segments. It’s a map where keys are segments.

controller/src/main/java/io/pravega/controller/store/stream/StreamCutComparison.java
pbelgundi: Ok.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: the comment is same as  line 955 because they are both calling the same method with different delta functions for size and time based retained value computation respectively.
hence identical comments.  
As suggested i will add more comments inside the getBoundStreamCuts method to make it clear what it is computing.

pbelgundi (approved):
LGTM

PR #5651, opened 2021-02-05
Issue 5650: Fix for Controller CreateReaderGroup API race condition

### LINKED ISSUE #5650: CreateReaderGroup errors out when trying to re-create a partially created ReaderGroup
**Problem description**
Idempotent creation of RG is not handled correctly on Controller. 
If a ReaderGroup is partially created and we attempt to invoke createReaderGroup again for this ReaderGroup with either the same or a difference ID, the create fails with following exception:
```
Exception in thread "main" io.grpc.StatusRuntimeException: ALREADY_EXISTS: addNewEntry: key: readerGroupTAIL-1651491015 table: _system/_tables/longevity/readerGroupsInScope.#.ce03cea7-f55d-4c30-8ce5-f03d7d061f57
```

**Problem location**
Controller

1. In `addReaderGroupToScope` handle StoreException.DataExistsException if entry for ReaderGroup already exists and return the id as part of response for CreateReaderGroup API.
2. Also make sure other steps for createReaderGroup are executed in an idempotent way for every invocation of createReaderGroup()

pbelgundi:

**Change log description**  
createReaderGroup API in Controller is not idempotent when trying to recreate a partially created ReaderGroup

**Purpose of the change**  
Fixes #5650

**What the code does**  
Makes createReaderGroup() API Idempotent and it would now return the `ReaderGroupId` and `Generation` for the existing/new Reader Group so that State-Synchronizer can persist these values on client.

**How to verify it**  
All Unit and Integration Tests should pass.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
+++ b/controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
@@ -222,7 +222,7 @@ public class PravegaTablesScope implements Scope {
                         storeHelper.addNewEntryIfAbsent(tableName, readerGroupName, getIdInBytes(readerGroupId)),
                         e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException,
                         () -> storeHelper.createTable(tableName)
-                                .thenCompose(v -> storeHelper.addNewEntry(tableName, readerGroupName,
+                                .thenCompose(v -> storeHelper.addNewEntryIfAbsent(tableName, readerGroupName,
                                         getIdInBytes(readerGroupId))))));
 
     }
--- a/controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
+++ b/controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
@@ -219,10 +219,12 @@ public class PravegaTablesScope implements Scope {
     public CompletableFuture<Void> addReaderGroupToScope(String readerGroupName, UUID readerGroupId) {
         return getReaderGroupsInScopeTableName()
                 .thenCompose(tableName -> Futures.toVoid(Futures.exceptionallyComposeExpecting(
-                        storeHelper.addNewEntry(tableName, readerGroupName, getIdInBytes(readerGroupId)),
+                        storeHelper.addNewEntryIfAbsent(tableName, readerGroupName, getIdInBytes(readerGroupId)),
                         e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException,
                         () -> storeHelper.createTable(tableName)
-                                .thenCompose(v -> storeHelper.addNewEntry(tableName, readerGroupName, getIdInBytes(readerGroupId))))));
+                                .thenCompose(v -> storeHelper.addNewEntry(tableName, readerGroupName,
+                                        getIdInBytes(readerGroupId))))));
+
     }
 
     public CompletableFuture<Void> removeReaderGroupFromScope(String readerGroup) {
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
@@ -355,16 +355,15 @@ public class StreamMetadataTasks extends TaskBase {
                                          .thenCompose(done -> streamMetadataStore.getReaderGroupId(scope, rgName)
                                          .thenCompose(rgId -> streamMetadataStore.getReaderGroupConfigRecord(scope, rgName, null, executor)
                                                  .thenApply(cfgRecord -> CreateReaderGroupResponse.newBuilder()
-                                                         .setReaderGroupId(rgId.toString())
-                                                         .setGeneration(cfgRecord.getObject().getGeneration())
+                                                         .setConfig(encodeReaderGroupConfigRecord(scope, rgName, cfgRecord.getObject(), rgId))
                                                          .setStatus(CreateReaderGroupResponse.Status.SUCCESS).build()))));
                      }
                      return streamMetadataStore.getReaderGroupId(scope, rgName)
                              .thenCompose(rgId -> streamMetadataStore.getReaderGroupConfigRecord(scope, rgName, null, executor)
                                      .thenApply(cfgRecord -> Controller.CreateReaderGroupResponse.newBuilder()
-                                     .setReaderGroupId(rgId.toString())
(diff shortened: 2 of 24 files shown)


controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -219,10 +219,12 @@ public void refresh() {
     public CompletableFuture<Void> addReaderGroupToScope(String readerGroupName, UUID readerGroupId) {
         return getReaderGroupsInScopeTableName()
                 .thenCompose(tableName -> Futures.toVoid(Futures.exceptionallyComposeExpecting(
-                        storeHelper.addNewEntry(tableName, readerGroupName, getIdInBytes(readerGroupId)),
+                        storeHelper.addNewEntryIfAbsent(tableName, readerGroupName, getIdInBytes(readerGroupId)),
                         e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException,
                         () -> storeHelper.createTable(tableName)
-                                .thenCompose(v -> storeHelper.addNewEntry(tableName, readerGroupName, getIdInBytes(readerGroupId))))));
+                                .thenCompose(v -> storeHelper.addNewEntry(tableName, readerGroupName,
```
shiveshr, line 225: shouldnt this also be addNewEntryIfAbsent?

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -351,9 +352,18 @@ private ReaderGroupConfiguration getRGConfigurationFromRecord(final String scope
                          return eventHelper.addIndexAndSubmitTask(event,
```
shiveshr, line 352: the event being passed here needs to be updated with correct rg id too.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -417,24 +428,35 @@ private ReaderGroupConfiguration getRGConfigurationFromRecord(final String scope
...
+                                                        .thenApply(cfgRecord ->
+                                                        CreateReaderGroupResponse.newBuilder()
+                                                                .setConfig(encodeReaderGroupConfigRecord(scope, rgName, cfgRecord.getObject(), rgId))
+                                                                .setStatus(CreateReaderGroupResponse.Status.SUCCESS).build())));
                                     }
-                                    return CompletableFuture.completedFuture(CreateReaderGroupStatus.Status.SUCCESS);
+                                    return streamMetadataStore.getReaderGroupId(scope, rgName)
+                                            .thenCompose(rgId -> streamMetadataStore.getReaderGroupConfigRecord(scope, rgName, null, executor)
+                                                    .thenApply(cfgRecord -> CreateReaderGroupResponse.newBuilder()
+                                                            .setConfig(encodeReaderGroupConfigRecord(scope, rgName, cfgRecord.getObject(), rgId))
```
shiveshr, line 458: we are calling this encode method twice. perhaps we can compute once and reuse.

shiveshr (changes requested):
Overall looks good but i have few comments that should be addressed before this is ready to be merged.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: True.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
pbelgundi: Good catch. Yes, will change.

pbelgundi: fixed

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: As discussed, the event should have rgId from the config passed by user to avoid overwriting the existing config with a new one passed in createRG API.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -449,6 +472,41 @@ private ReaderGroupConfiguration getRGConfigurationFromRecord(final String scope
...
 
+    /**
+     * The method translates the internal object ReaderGroupConfigRecord object into REST response object.
+     *
+     * @param scope the scope of the Reader Group.
+     * @param rgName the name of the Reader Group.
+     * @param rgConfig The configuration of Reader Group.
+     * @return Stream properties wrapped in StreamResponse object
+     */
+    private static final ReaderGroupConfiguration encodeReaderGroupConfigRecord(String scope, String rgName,
```
shiveshr, line 483: should this method be moved to model helper?

shiveshr (approved):
LGTM, one minor comment

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: done.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -364,7 +364,7 @@ private ReaderGroupConfiguration getRGConfigurationFromRecord(final String scope
         return streamMetadataStore.getReaderGroupId(scope, rgName)
                 .thenCompose(rgId -> streamMetadataStore.getReaderGroupConfigRecord(scope, rgName, null, executor)
                         .thenApply(cfgRecord -> Controller.CreateReaderGroupResponse.newBuilder()
-                                .setConfig(encodeReaderGroupConfigRecord(scope, rgName, cfgRecord.getObject(), rgId))
+                                .setConfig(io.pravega.controller.server.rest.ModelHelper.encodeReaderGroupConfigRecord(scope, rgName, cfgRecord.getObject(), rgId))
```
shiveshr, line 367: I know at some of the places we have brought in our wire objects over to this class, esp the response status codes. But all the encoding and decoding of complex objects should be handled at the model helper and our business logic should deal with the internal model classes exclusively and we should not have to encode it here. 

can we open a new github issue for this and i am ok with not addressing it here. 

thanks

shiveshr (approved):

PR #5654, opened 2021-02-05
Issue 5647: (SegmentStore) (BugFix) Table Segment Conditional Updates may fail if Compaction occurrs simultaneously

### LINKED ISSUE #5647: Table Segment Conditional Update failure during Compaction.
**Problem description**
If a Table compaction occurs while, or just before a conditional update, the conditional update will never succeed. It will error out with `StreamSegmentTruncatedException`.

Scenario:
1. Key A is at offset O1. The Segment Key Cache has this as {A->O1}
2. We move Key A to Offset O2 as part of compaction. Note that compaction does not update the SegmentKeyCache.
3. We reindex a to map to O2. Since this was not part of the Tail Cache, the SegmentKeyCache is not touched.
4. We truncate the segment at offset T (O1 < T < O2) post compaction.
5. A `get` or `iterator` issued right now would initially try to read A from O1, figure out it was truncated, and refresh the cache with the right value.
6. A conditional update (the validator part) does not have this logic, hence it will not succeed.

**Problem location**
ContainerKeyIndex.

Either:
1. Patch ContainerKeyIndex.validateConditionalUpdate to handle this situation (and clear the cache)
2. Clear the SegmentKeyCache after truncation (if possible) for all pointers to below the truncation offset.
3. Do #5621 (remove the SegmentKeyCache and replace with IndexCache, at a lower level).

andreipaduroiu:
**Change log description**  
Fixed a bug in ContainerKeyIndex where conditional update validation would fail if the update key has recently been relocated to a new position and the old key has been truncated away.

**Purpose of the change**  
Fixes #5647.

**What the code does**  
The steps for a table segment compaction are as follows:
1. A group of Table Entries is copied from the head of the segment to the tail.
2. (Later) Those entries are re-indexed and the index now points to the new locations.
3. The old entries are truncated away.

Sometime after 2), the SegmentKeyCache should be updated with the new location. If not, we might be trying to access the old location. The ContainerTableExtensionImpl already handles this for reads and iterators, however this functionality has not been added for conditional updates. Conditional updates may need to read the table segment in case it needs to do a proper validation of the provided key version.

This code unifies the two paths to use the same piece of code. 

**How to verify it**  
New unit test added. Build must pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerKeyIndex.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/ContainerKeyIndex.java
@@ -19,6 +19,7 @@ import io.pravega.common.concurrent.MultiKeySequentialProcessor;
 import io.pravega.common.util.BufferView;
 import io.pravega.segmentstore.contracts.ReadResult;
 import io.pravega.segmentstore.contracts.SegmentProperties;
+import io.pravega.segmentstore.contracts.StreamSegmentTruncatedException;
 import io.pravega.segmentstore.contracts.tables.BadKeyVersionException;
 import io.pravega.segmentstore.contracts.tables.ConditionalTableUpdateException;
 import io.pravega.segmentstore.contracts.tables.KeyNotExistsException;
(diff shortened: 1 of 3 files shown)


sachin-j-joshi (commented):
LGTM

sachin-j-joshi (approved):

PR #5662, opened 2021-02-06
Issue 5475: (SLTS) Eliminate unnecessary pessimistic checks.

### LINKED ISSUE #5475: Seemingly unnecessary doGetStreamSegmentInfo call during read path in ExtendedS3Storage
**Problem description**
Pravega cluster with S3 compatible tier 2 storage (ECS).  Noticed when reading stream data from tier 2 that there was roughly 1 get object metadata api call +1 get acl call per data read.  

Our test query ends up reading approx 75GB from tier 2.  tier 2 is test VM ECS deployment.
test run results in:
1. 138879 read requests to ECS, total processing time on ECS 1492337ms
2. 138083 requests for get metadata + get acl (each), total processing time on ECS 538872ms

So it seems to add non-trivial amount of overhead.

Traced it to ExtendedS3Storage#doGetStreamSegmentInfo.  Noticed that the return value of doGetStreamSegmentInfo was ignored.

**Problem location**
ExtendedS3Storage#doOpenRead

It appears like the call to doGetStreamSegmentInfo could simply be removed.
I commented it out for a test and total run time of our test app decreased by about 30% or so.

sachin-j-joshi:

**Change log description**  
Eliminate pessimistic check and instead optimistically proceed to read/write directly and handle failures only when we encounter them.

**Purpose of the change**  
Fixes #5475  only for SLTS. (please do not close issue).

**What the code does**  

Eliminate pessimistic check and instead optimistically proceed to read/write directly and handle failures only when we encounter them.

**How to verify it**  
No regression

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
@@ -177,23 +177,21 @@ class ReadOperation implements Callable<CompletableFuture<Integer>> {
         val chunkBufferOffset = new AtomicInteger(bufferOffset);
         // Note that technically it is possible that read actually request reads less than requested bytes, requiring additional reads on the same chunk.
         // Hence the for loop below.
-        return chunkedSegmentStorage.getChunkStorage().openRead(chunkName)
-                .thenComposeAsync(chunkHandle ->
-                    Futures.loop(
-                            () -> chunkBytesRemaining.get() > 0,
-                            () -> chunkedSegmentStorage.getChunkStorage().read(chunkHandle,
-                                    chunkFromOffset.get(),
-                                    chunkBytesRemaining.get(),
-                                    buffer,
-                                    chunkBufferOffset.get())
-                                    .thenAccept(n -> {
-                                        Preconditions.checkState(n != 0, "Zero bytes read chunk=%s, fromOffset=%d", chunkName, fromOffset);
-                                        chunkBytesRemaining.addAndGet(-n);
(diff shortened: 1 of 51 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
```diff
@@ -177,23 +177,21 @@ private void logEnd() {
...
-                                    .thenAccept(n -> {
-                                        Preconditions.checkState(n != 0, "Zero bytes read chunk=%s, fromOffset=%d", chunkName, fromOffset);
-                                        chunkBytesRemaining.addAndGet(-n);
-                                        chunkFromOffset.addAndGet(n);
-                                        chunkBufferOffset.addAndGet(n);
-                                    }),
-                            chunkedSegmentStorage.getExecutor()
-                    ), chunkedSegmentStorage.getExecutor());
+        val chunkHandle = ChunkHandle.readHandle(chunkName);
+        return Futures.loop(
```
andreipaduroiu, line 181: TIP: If you are passed 0 bytes, then avoid the `Futures.loop` altogether. Setting it up is rather expensive in terms of thread pool tasks, so it might be worth it to exit early in that case.

andreipaduroiu (approved):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ReadOperation.java
sachin-j-joshi: ok. we already have that check on line 85.

PR #5665, opened 2021-02-06
Issue 5664: (SLTS)  fix possible thread visibility issues..

### LINKED ISSUE #5664: SLTS - fix possible thread visibility issues.
**Problem description**
There are couple of places in SLTS code where arrays and other data structures are protected from concurrent modifications yet could still potentially show thread visibility issue.

**Problem location**
SLTS operations.

Use Collections.synchronized /synchronized to force memory barrier
Use volatile for mutable variables

sachin-j-joshi:

**Change log description**  
Use Collections.synchronized /synchronized to force memory barrier
Use volatile for mutable variables

**Purpose of the change**  
Fixes #5664 

**What the code does**  
There are couple of places in SLTS code where arrays and other data structures are protected from concurrent modifications yet could still potentially show thread visibility issue.

**How to verify it**  
Use Collections.synchronized /synchronized to force memory barrier
Use volatile for mutable variables

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
@@ -42,6 +42,7 @@ import java.util.Arrays;
 import java.util.ConcurrentModificationException;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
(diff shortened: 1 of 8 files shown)


andreipaduroiu (commented):
This looks good. Thanks for doing it.

One comment before I approve it. Even though a Collection may be synchronized, if you iterate through it (using `iterator`), that is not thread safe and it will still throw a `ConcurrentModificationException` if you modify it while some other thread iterates through it.

I can't tell from this PR whether this is the case here, but would you mind checking the code you modified for places where these are accessed to ensure they are not iterated over.

If you need iteration, your best bet is to encapsulate them and make a copy every time you need an iterator (so iterate over the copy).

sachin-j-joshi:
The code does not modify any lists concurrently. The code is just series of lambdas that are executed one after another via `CompletableFuture.then...` chaining.  So I think its is safe in that sense.

This change fixes possible thread visibility issues as execution may jump from one core to another.

andreipaduroiu (approved):

PR #5671, opened 2021-02-08
Issue 5666: Watermarking system test failure

### LINKED ISSUE #5666: Sporadic system test failure in WatermarkingTest
**Problem description**
We have seen sporadic failures in our system tests related to `WatermarkingTest`. The test failure is as follows:

1) The test starts correctly creating a couple of writers that write some events with time-related values.

2) Then several scale events are done to test watermarks with more complex segment compositions.

3) After that, a watermark reader is instantiated and the test fetches 4 watermarks. It passes the assertions defined to validate that watermarks are correct.

4) Then, based on these watermarks, the test creates the start and end streamcuts for a event stream reader. 

5) Just after reading the first event with that reader, the test executes `reader.getCurrentTimeWindow(streamObj)` and expects that : ` assertNotNull(currentTimeWindow.getLowerTimeBound());`. Is that precise assertion, the one that checks that `TimeWindow` content is not null the one that fails the test.

One reason for that `TimeWindow`  to have null contents is that the àdvanceTo` method was not executed when the check is done, as it is the one setting the `passedTimestamp` (i.e., time window lower bound) value:
https://github.com/pravega/pravega/blob/2e70f876c5eaeaf0ed51504174c5e6b13f0645a1/client/src/main/java/io/pravega/client/stream/impl/WatermarkReaderImpl.java#L134

But looking at the logs, I see that segments are acquired by the reader:
```
2021-02-04 12:12:02,353 71056 [Time-limited test] INFO  i.p.c.s.impl.EventStreamReaderImpl - EventStreamReaderImpl( id=myreader) acquiring segments {SegmentWithRange(segment=testWatermarkingScope780264939/testWatermarkingStream/2.#epoch.0, range=null)=360, SegmentWithRange(segment=testWatermarkingScope780264939/testWatermarkingStream/3.#epoch.0, range=null)=360, SegmentWithRange(segment=testWatermarkingScope780264939/testWatermarkingStream/0.#epoch.0, range=null)=-1, SegmentWithRange(segment=testWatermarkingScope780264939/testWatermarkingStream/1.#epoch.0, range=null)=540, SegmentWithRange(segment=testWatermarkingScope780264939/testWatermarkingStream/4.#epoch.0, range=null)=540}
```
This means that `acquireSegmentsIfNeeded()` has been executed in `EventStreamReaderImpl` has been executed, which also involved executing: `waterMarkReaders.forEach((stream, reader) -> reader.advanceTo(groupState.getLastReadpositions(stream)));`
Maybe we need more logs to confirm this.

Assuming that the `advanceTo` method has actually been executed, then the reason for the failure may be that `processFunction` in `WatermarkingReaderImpl` is not setting the `passedTimestamp` value for some reason. For instance, one of the important points in that method is the `compare` method used in it, which compares sets of segment ranges. If we compare the previous set of segments with what the reader sees on a successful test execution, we see that:
```
2021-02-04 14:47:10,047 69431 [Time-limited test] INFO  i.p.c.s.impl.EventStreamReaderImpl - EventStreamReaderImpl( id=myreader) acquiring segments {SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/10.#epoch.3, range=null)=0, SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/9.#epoch.3, range=null)=90, SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/8.#epoch.2, range=null)=0, SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/3.#epoch.0, range=null)=900, SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/7.#epoch.2, range=null)=0, SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/5.#epoch.1, range=null)=360, SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/4.#epoch.0, range=null)=630, SegmentWithRange(segment=testWatermarkingScope553039405/testWatermarkingStream/6.#epoch.1, range=null)=270}
```
That is, it looks like in a successful test execution, more scale have actually been executed (i.e., there are more segments) and none of the segments is at position `-1`.  This may be related to the behavior in the `WatermarkingReaderImpl:compare`  method behavior, which could cause the `TimeWindow` contents to be `null` and hence fail the test.

**Problem location**
Perhaps in `WatermarkingReaderImpl`, assuming that the scale of the Stream has been correct.

Understand the issue and fix the test.

RaulGracia:
**Change log description**  
Fixes WatermarkingTest to make sure that the time window converges before executing assertions.

**Purpose of the change**  
Fixes #5666.

**What the code does**  
The `TimeWindow` of the reader being evaluated at the end of the `WatermarkingTest` system test is not immediately updated when writers writer the data. Instead, the update of `TimeWindow` is asynchronous, and this induced sporadic failures in this test, as was not waiting for the state of the `TimeWindow` to be the expected one. This PR just uses `assertEventallyEquals` to execute the same assertions that were defined before, but when the state of `TimeWindow` is the right one.

**How to verify it**  
Got 5/5 successful executions of `WatermarkingTest` in system tests.


shiveshr (approved):
LGTM

PR #5674, opened 2021-02-08
Issue 5673: (SLTS) Fix wrong GC config

### LINKED ISSUE #5673: SLTS - Wrong config for garbage collector sleep.
**Problem description**
Currently garbage collector sleep is configured to be in minutes , it should be in milliseconds.

**Problem location**
GarbageCollector

Change unit from seconds to milliseconds.
Also change log.info to log.debug to reduce noise.

sachin-j-joshi:

**Change log description**  
Change unit from seconds to milliseconds.
Also change log.info to log.debug to reduce noise.

**Purpose of the change**  
Fixes #5673

**What the code does**  

- Change default settings from 60 seconds to 10 milliseconds. (Note that - Every 10 millisecond, new iteration starts and waits for items to be available for GC on its dedicated thread, and deletes 10 chunks max per iteration. We need small wait time between iterations.)
- Fix spelling mistake.
- Also change log.info to log.debug to reduce noise.

**How to verify it**  
No regression

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
@@ -41,8 +41,8 @@ public class ChunkedSegmentStorageConfig {
     public static final Property<Integer> GARBAGE_COLLECTION_DELAY = Property.named("garbage.collection.delay.seconds", 60);
     public static final Property<Integer> GARBAGE_COLLECTION_MAX_CONCURRENCY = Property.named("garbage.collection.concurrency.max", 10);
     public static final Property<Integer> GARBAGE_COLLECTION_MAX_QUEUE_SIZE = Property.named("garbage.collection.queue.size.max", 16 * 1024);
-    public static final Property<Integer> GARBAGE_COLLECTION_SLEEP = Property.named("garbage.collection.sleep.seconds", 60);
-    public static final Property<Integer> GARBAGE_COLLECTION_MAX_ATTEMPS = Property.named("garbage.collection.attempts.max", 3);
+    public static final Property<Integer> GARBAGE_COLLECTION_SLEEP = Property.named("garbage.collection.sleep.millis", 10);
+    public static final Property<Integer> GARBAGE_COLLECTION_MAX_ATTEMPTS = Property.named("garbage.collection.attempts.max", 3);
 
 
     /**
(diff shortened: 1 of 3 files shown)


andreipaduroiu (approved):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
```diff
@@ -41,8 +41,8 @@
     public static final Property<Integer> GARBAGE_COLLECTION_DELAY = Property.named("garbage.collection.delay.seconds", 60);
     public static final Property<Integer> GARBAGE_COLLECTION_MAX_CONCURRENCY = Property.named("garbage.collection.concurrency.max", 10);
     public static final Property<Integer> GARBAGE_COLLECTION_MAX_QUEUE_SIZE = Property.named("garbage.collection.queue.size.max", 16 * 1024);
-    public static final Property<Integer> GARBAGE_COLLECTION_SLEEP = Property.named("garbage.collection.sleep.seconds", 60);
```
RaulGracia, line 44: This is going to flood our logs if we ever want to use DEBUG level for a period of time to have more details about the activity of the system. In our system tests, this will for instance make the size of logs much larger without need. Also, it looks a super low interval for something that is supposed to periodically delete garbage, which is going to induce a significant amount of CPU usage without a clear justification. I could agree on lowering this value, but anything below 1-second I think that deserves a justification that I don't see in the PR.

RaulGracia (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageConfig.java
sachin-j-joshi: 1. GC polling runs on its own dedicated thread.
2. During each iteration the code blocks until at least one metadata item is eligible for deletion. (So iteration don't finish until there is something to delete. This way there are no empty iterations creating lots of useless logs.)
3. Each iteration it only deletes GARBAGE_COLLECTION_MAX_CONCURRENCY number of items.
4. However when there are lots of chunks to be deleted the queue becomes full quickly if we don't drain the queue fast enough. 
5. By reducing interval between iterations we are responding to increased delete queue faster, at the same time when queue is empty we are blocked thus adjusting delete speed to queue length.

RaulGracia: Ok, so my next question is: if the behavior of the GC thread is blocking as you mention (this could be mentioned in the parameter javadoc, by the way), what is the point of having a sleep at all? That is, it will delete chucks as long as there are chunks to delete, or it will remain blocked otherwise. This parameter seems a candidate for removal, so we could simplify the configuration of SLTS.

sachin-j-joshi: This parameter allow us to throttle the rate at which background delete tasks are run. However I don't know what is a good value. In such cases it's better to have config value that you can tweak and experiment with.

RaulGracia (approved):

PR #5684, opened 2021-02-09
Issue 5683: Update the year and include Pravega Authors in NOTICE file

### LINKED ISSUE #5683: Update the year in all legal related files
**Problem description**
NOTICE file needs to be update to reflect the current year

**Problem location**
NOTICE

Change the  range to `2017-2021`

medvedevigorek:
**Change log description**  
* Updates NOTICE file the new year range `2017-2021`
* Adds Pravega Authors at the top to comply with CNCF guidelines 

**Purpose of the change**  
Fixes #5683 

**What the code does**  
No code change is done, just an update in NOTICE file 

**How to verify it**  
No special verification is required, build should pass though

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,6 +1,6 @@
-Copyright (c) 2017-2020 Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright (c) 2021 Pravega Authors.
+Copyright (c) 2017-2021 Dell Inc., or its subsidiaries. All Rights Reserved.
 
 This software contains source code from Apache BookKeeper, distributed under
-the Apache License Version 2.0, and copyrighted to the
-Apache Software Foundation.
+the Apache License Version 2.0, and copyrighted to the Apache Software Foundation.
 http://bookkeeper.apache.org


NOTICE
```diff
@@ -1,6 +1,6 @@
-Copyright (c) 2017-2020 Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright (c) 2021 Pravega Authors.
```
fpj, line 1: @derekm do we have to make a similar change to the `HEADER` file so that we reflect the copyright change onto the license header of new files?

derekm (approved):
This looks good...

Should we save HEADER file change for a PR where we also sweep through all files and add "Copyright (c) Pravega Authors" to all files?

(In the HEADER file itself we can replace the Dell copyright statement with "Pravega Authors", such that new files will lack a Dell copyright statement. But in all files where the old HEADER was placed, we should add the Pravega Authors copyright statement without removing the Dell copyright statement.)

NOTICE
derekm: @fpj -- Yeah I just asked the same question in my review response. Whether HEADER change goes in this PR or not, maybe we have another PR that updates all files with the old header to add "Pravega Authors" without removing Dell copyrights.

derekm: (Updating all files with the old headers seems preferable to updating each file as it is edited.)

medvedevigorek: Let's not mix multiple things into a single PR. opened #5697 to properly handle the copyright section.

fpj (approved):

PR #5692, opened 2021-02-09
Issue 5690: Adding new stream or rg or kvt to scope should lazily create the metadata table if it doesnt exist

### LINKED ISSUE #5690: Post create scope, lazily create streamsinscopeTable, rgsInScopeTable and kvtsInScopeTable and delete them with delete scope
**Problem description**
CreateScope is not implemented as a failover protected workflow. Which means we could have failed after adding the entry to scopes table but without creating corresponding streamsInScope table, rgsInScope table and kvtsInScope table.

These should be created lazily if they do not exist if an attempt to add streams/rgs/kvts to scope is attempted. 

Similarly, at the time of delete scope, these tables should be removed. Presently we only remove streamsInScopeTable. 

**Problem location**
PravegaTablesScopes

Fix the bugs.

shiveshr:

**Change log description**  
- Post create scope, lazily create streamsinscopeTable, rgsInScopeTable and kvtsInScopeTable if they dont exist
- delete scope should not be allowed if rg table and kvt table are non empty

**Purpose of the change**  
Fixes #5690 

**What the code does**  
In PravegaTablesScopes we create tables for stream, kvt, and rg but that is not done as a workflow. So after adding the entry we could have failed. This could make the scope unusable unless createScope is called again.
The way we fix it is, any access to the scope (for add`resource` or list`resource`) should not result in datanotfound if the entry in scopes table was successfully created. 
So we lazily create table as part of addStream, addReaderGroup and addKvt if they dont exist.
and we return empty list if tables dont exist. 

Secondly, deletescope should be predicated on whether all resources under the scope have been cleaned up or not. 

**How to verify it**  
Unit test added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
+++ b/controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.tuple.Pair;
 
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
@@ -28,6 +29,7 @@ import java.util.concurrent.CompletionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 
 import static io.pravega.controller.store.stream.PravegaTablesStreamMetadataStore.DATA_NOT_FOUND_PREDICATE;
 import static io.pravega.controller.store.stream.PravegaTablesStreamMetadataStore.SCOPES_TABLE;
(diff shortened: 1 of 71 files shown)


controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -69,14 +71,8 @@ public String getName() {
...
         // If scope entry exists in Scopes table, create the streamsInScope table before throwing DataExists exception
-        return Futures.handleCompose(Futures.exceptionallyComposeExpecting(storeHelper.addNewEntry(
-                SCOPES_TABLE, scopeName, newId()),
-                DATA_NOT_FOUND_PREDICATE,
-                () -> storeHelper.createTable(SCOPES_TABLE)
-                                 .thenCompose(v -> {
-                                     log.debug("table for scopes created {}", SCOPES_TABLE);
-                                     return storeHelper.addNewEntryIfAbsent(SCOPES_TABLE, scopeName, newId());
-                                 })), (r, e) -> {
+        return Futures.handleCompose(withCreateTableIfAbsent(() -> storeHelper.addNewEntry(
```
pbelgundi, line 74: withCreateTableIfAbsent() can also be used on Streams, ReaderGroups and KVTs

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -140,37 +136,35 @@ public String getName() {
...
+                            storeHelper.deleteTable(kvtsInScopeTableName, true), 
+                            storeHelper.deleteTable(rgsInScopeTableName, true))
+                                     .thenAccept(v -> log.debug("tables deleted {} {} {}", streamsInScopeTableName,
+                                             kvtsInScopeTableName, rgsInScopeTableName));
+                })
                 .thenCompose(deleted -> storeHelper.removeEntry(SCOPES_TABLE, scopeName));
     }
 
     @Override
     public CompletableFuture<Pair<List<String>, String>> listStreams(int limit, String continuationToken, Executor executor) {
```
pbelgundi, line 157: We also need a listReaderGroups() API for listing all ReaderGroups inside a scope.

pbelgundi (approved):
LGTM with some minor comments.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shiveshr: we presently have a rest api for listing readergroups in a scope.. it does so by listing all streams in a scope and then filtering those with prefix _RG. 
Perhaps that should be changed to a new api in streammetadatastore which lists them from the rg metadata.. just that it wont list any older rgs.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shiveshr: we use it dont we?

pbelgundi: yes we do.

PR #5706, opened 2021-02-10
Issue 5695: Sporadic failure in OffsetTruncationTest system test

### LINKED ISSUE #5695: Sporadic failure in OffsetTruncationTest system test
**Problem description**
Found a sporadic failure of `OffsetTruncationTest` system test. The current debug of the problem is as follows:
1) The test first writes some data (i.e., 200 events) on a 2-segment Stream.
2) After that, the test creates a ReaderGroup and 2 readers to read the first 50 events. After the reads complete, the test initiates a checkpoint:
https://github.com/pravega/pravega/blob/a1f01d7d2064d309be37c52bd1354ae756800ed9/test/system/src/test/java/io/pravega/test/system/OffsetTruncationTest.java#L133
 and gets the `StreamCut` that represents the point in the Stream after reading these 50 events:
https://github.com/pravega/pravega/blob/a1f01d7d2064d309be37c52bd1354ae756800ed9/test/system/src/test/java/io/pravega/test/system/OffsetTruncationTest.java#L134
3) The previous `StreamCut` is used to truncate the Stream at that specific point. That is, the StreamCut represents a point in the Stream that, when starting reading from it, there should be 150 events left.
4) Then, we create a ReaderGroup starting from that StreamCut, but the test fails as it reads more events than expected:
```
2021-02-04 04:00:20,932 20627 [main] ERROR i.p.t.s.f.TestFrameworkException - TestFramework Exception. Type: InternalError, Details: Exception while running test method: offsetTruncationTest
java.lang.AssertionError: Expected read events:  expected:<150> but was:<175>
	at io.pravega.test.system.OffsetTruncationTest.offsetTruncationTest(OffsetTruncationTest.java:142)
```

**Problem location**
One problem we have is that we do not have the Controller logs for this test run. In any case, with the test driver log, we do see the following:

The StreamCut used to truncate the Stream in the failed build looks like this:
```
2021-02-04 04:00:11,123 10818 [Time-limited test] INFO  i.p.c.admin.impl.StreamManagerImpl - Truncating scope/stream: testOffsetTruncationScope1954283559/testOffsetTruncationStream with stream cut: testOffsetTruncationScope1954283559/testOffsetTruncationStream:0=0, 1=450
```
As you can see, only one of the segments has an offset > 0 in that case. Conversely, on a successful execution, the StreamCut actually represents the situation in which readers have read data from both Stream segments (25 events from each segment):
```
2021-02-09 07:33:21,842 7409 [Time-limited test] INFO  i.p.c.admin.impl.StreamManagerImpl - Truncating scope/stream: testOffsetTruncationScope24583830/testOffsetTruncationStream with stream cut: testOffsetTruncationScope24583830/testOffsetTruncationStream:0=450, 1=450
```
This leads the reader group to only encounter 1 `SegmentTruncatedException` in the failed case instead of 2 of such exceptions, and therefore reading more data than expected.

With this, we can infer that the truncation process was correct, but the StreamCut provided as input for truncation was wrong, as it was missing the read events from one of the Segments.

Inspect if there could be a collision of the Checkpoint initiated in the test with an automatic Checkpoint within the ReaderGroup that led the test to use the automated one that represents an a point in which not all the events are read.

RaulGracia:
**Change log description**  
Increased reader timeout in system tests to prevent sporadic failures due to not being able to read all the events. Added logs and checks in OffsetTruncationTest to make this kind of sporadic failures evident.

**Purpose of the change**  
Fixes #5695.

**What the code does**  
The original sporadic failure for this test (as well as others that use the same read/write methods, like `BatchClientSystemTests`) was that, probably due to environment slowness, it took too long for the readers to receive any data from the test stream and they exited the read loop thinking that they read all the events needed. This lead the test logic to build incorrect `StreamCut`s (created upon the current position of readers in a `ReaderGroup`), thus failing on event count assertions. To mitigate this problem, this PR does the following:
- It increases the reader timeout from 1 to 10 seconds. With 1 second, we rarely observed this issue, so with 10 seconds it may be enough to reduce it almost completely even in slow environments. 
- This PR introduces a couple log messages to know the final write/read situation of the test:
`2021-02-15 14:50:50,231 10992 [Time-limited test] INFO  i.p.t.system.AbstractReadWriteTest - Writer EventStreamWriterImpl(stream=StreamImpl(scope=testOffsetTruncationScope99087015, 
streamName=testOffsetTruncationStream), closed=false) finished writing 200 events.`
`2021-02-15 14:50:50,551 11312 [Thread-3] INFO  i.p.t.system.AbstractReadWriteTest - Reader EventStreamReaderImpl( id=testOffsetTruncationRG753252179-1) finished reading 25 events.`
- This PR also makes the test to fail fast in the case that the number of events read is not the expected one, thus making debugging easier and more explicit.
- This PR adds the test of the `generateStreamCuts()` call, which was not exercised before in system tests.
- Finally, some minor style improvements are added with no impact on code logic.

**How to verify it**  
Executed multiple +10 times `OffsetTruncationTest` without reproduction of the original problem.


shrids (approved):
LGTM.

PR #5713, opened 2021-02-10
Issue 5712: Internally idempotent conditional append 

### LINKED ISSUE #5712: Prevent case of a successful append that throws in StateSynchronizer.
**Problem description**
A state synchronizer can end up calling an update function more than needed if the conditional append it performs succeeds but it believes that it has failed.

**Problem location**
ConditionalOutputStreamImpl and SynchronizerUpdate functions

Making the update functions idempotent should prevent this from causing problems. 
To fix the underlying issue the ConditionalOutputStreamImpl needs to treat InvalidEventNumber as a retryable exception so that it will re-connect and resolve if the event was sent or not using the event number from the AppendSetup in the event that upon the first retry the AppendSetup received stale data.

tkaitchuck:
**Change log description**  
* Allows ConditionalOutputStream to internally retry InvalidEventNumber exceptions
* Changes reader group updates to double check the assumptions they make before updating the reader group.

**Purpose of the change**  
Fixes #5712

**What the code does**  
Makes ConditionalOutputStream retry InvalidEventNumber replies from the server as introduced in https://github.com/pravega/pravega/pull/5710 which will allow the retry to resolve the ambiguity of whether or not a conditional append was successful if the connection drops before an ack can be received.

### DIFF
--- a/client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
@@ -15,22 +15,23 @@ import io.pravega.auth.AuthenticationException;
 import io.pravega.auth.TokenExpiredException;
 import io.pravega.client.connection.impl.ConnectionPool;
 import io.pravega.client.connection.impl.RawClient;
-import io.pravega.client.security.auth.DelegationTokenProvider;
 import io.pravega.client.control.impl.Controller;
+import io.pravega.client.security.auth.DelegationTokenProvider;
 import io.pravega.common.Exceptions;
 import io.pravega.common.util.Retry.RetryWithBackoff;
 import io.pravega.shared.protocol.netty.ConnectionFailedException;
 import io.pravega.shared.protocol.netty.Reply;
 import io.pravega.shared.protocol.netty.WireCommands;
 import io.pravega.shared.protocol.netty.WireCommands.AppendSetup;
+import io.pravega.shared.protocol.netty.WireCommands.AuthTokenCheckFailed;
 import io.pravega.shared.protocol.netty.WireCommands.ConditionalAppend;
 import io.pravega.shared.protocol.netty.WireCommands.ConditionalCheckFailed;
 import io.pravega.shared.protocol.netty.WireCommands.DataAppended;
 import io.pravega.shared.protocol.netty.WireCommands.Event;
+import io.pravega.shared.protocol.netty.WireCommands.InvalidEventNumber;
 import io.pravega.shared.protocol.netty.WireCommands.SegmentIsSealed;
 import io.pravega.shared.protocol.netty.WireCommands.SetupAppend;
 import io.pravega.shared.protocol.netty.WireCommands.WrongHost;
-import io.pravega.shared.protocol.netty.WireCommands.AuthTokenCheckFailed;
 import java.nio.ByteBuffer;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -116,7 +117,7 @@ class ConditionalOutputStreamImpl implements ConditionalOutputStream {
         if (reply instanceof AppendSetup) {
             return (AppendSetup) reply;
         } else {
-            throw handleUnexpectedReply(reply);
+            throw handleUnexpectedReply(reply, "AppendSetup");
         }
     }
 
@@ -126,12 +127,12 @@ class ConditionalOutputStreamImpl implements ConditionalOutputStream {
         } else if (reply instanceof ConditionalCheckFailed) {
             return false;
         } else {
-            throw handleUnexpectedReply(reply);
+            throw handleUnexpectedReply(reply, "DataAppended");
         }
     }
 
     @VisibleForTesting
-    RuntimeException handleUnexpectedReply(Reply reply) {
+    RuntimeException handleUnexpectedReply(Reply reply, String expectation) {
         closeConnection(reply.toString());
         if (reply instanceof WireCommands.NoSuchSegment) {
             throw new NoSuchSegmentException(reply.toString());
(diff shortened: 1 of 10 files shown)


andreipaduroiu (changes requested):
Preemptively blocking until ready to merge. (I have not yet reviewed)

client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
```diff
@@ -126,19 +127,22 @@ private boolean transformDataAppended(Reply reply) {
...
         closeConnection(reply.toString());
         if (reply instanceof WireCommands.NoSuchSegment) {
             throw new NoSuchSegmentException(reply.toString());
         } else if (reply instanceof SegmentIsSealed) {
             throw Exceptions.sneakyThrow(new SegmentSealedException(reply.toString()));
         } else if (reply instanceof WrongHost) {
             throw Exceptions.sneakyThrow(new ConnectionFailedException(reply.toString()));
+        } else if (reply instanceof InvalidEventNumber) {
+            throw Exceptions.sneakyThrow(new ConnectionFailedException(
+                    "Got stale data from setupAppend on segment " + segmentId + " for ConditionalOutputStream."));
```
shrids, line 145: Including the writer-id will help us to easily co-relate logs from the Segment store

common/src/main/java/io/pravega/common/util/Retry.java
```diff
@@ -187,7 +187,11 @@ private RetryAndThrowBase(RetryWithBackoff params) {
                     Exceptions.handleInterrupted(() -> Thread.sleep(sleepFor));
 
                     delay = Math.min(params.maxDelay, params.multiplier * delay);
-                    log.debug("Retrying command {} Retry #{}, timestamp={}", r.toString(), attemptNumber, Instant.now());
+                    log.debug("Retrying command {} due to \"{}\" Retry #{}, timestamp={}",
```
shrids, line 190: +1

client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
andreipaduroiu: Writer Id + event number would be even better.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -187,16 +187,23 @@ boolean handleEndOfSegment(SegmentWithRange segmentCompleted) throws ReaderNotIn
         AtomicBoolean reinitRequired = new AtomicBoolean(false);
         boolean result = sync.updateState((state, updates) -> {
             if (!state.isReaderOnline(readerId)) {
+                log.error("Reader " + readerId + " is offline according to the state but is attempting to update it.");
```
andreipaduroiu, line 190: I believe this should be a `warn`. No actual error got caught, but it is an invalid/unexpected condition.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -187,16 +187,23 @@ boolean handleEndOfSegment(SegmentWithRange segmentCompleted) throws ReaderNotIn
...
-                reinitRequired.set(false);
-                //This check guards against another checkpoint having started.
-                if (state.getCheckpointForReader(readerId) == null) {
-                    updates.add(new SegmentCompleted(readerId, segmentCompleted, segmentToPredecessor));
-                    return true;
-                }
+                return false;
+            } 
+            if (!state.getSegments(readerId).contains(segmentCompleted.getSegment())) {
+                log.error("Reader " + readerId + " is does not own the segment " + segmentCompleted + "but is attempting to release it.");
```
andreipaduroiu, line 195: Same here.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
```diff
@@ -490,7 +497,12 @@ void checkpoint(String checkpointName, PositionInternal lastPosition) throws Rea
                 reinitRequired.set(true);
             } else {
                 reinitRequired.set(false);
-                updates.add(new CheckpointReader(checkpointName, readerId, lastPosition.getOwnedSegmentsWithOffsets()));
+                String cpName = state.getCheckpointForReader(readerId);
+                if (cpName.equals(checkpointName)) {
+                    updates.add(new CheckpointReader(checkpointName, readerId, lastPosition.getOwnedSegmentsWithOffsets()));
+                } else {
+                    log.error("{} was asked to checkpoint for {} but the state says its next checkpoint should be {}", readerId, checkpointName, cpName);
```
andreipaduroiu, line 504: and here

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
tkaitchuck: Our definition of error is "Something bad happened that can't be automatically be handled"
It meets this criteria. If this line is reached a user will have to reinitialize the reader because we cannot get it back into a good state.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
tkaitchuck: See above.

client/src/main/java/io/pravega/client/segment/impl/ConditionalOutputStreamImpl.java
tkaitchuck: Added

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
tkaitchuck: This one I agree. Changed.

client/src/main/java/io/pravega/client/stream/impl/ReaderGroupStateManager.java
andreipaduroiu: OK. Got it.

andreipaduroiu (approved):

PR #5721, opened 2021-02-11
Issue 5767: Ensure the WriterState object contains the updated last event number for the given writer.

### LINKED ISSUE #5767: Reduce BadAttributeUpdate exceptions during Conditional Appends.
**Describe the bug**
Reduce retries by the client due to BadAttributeException due to possible stale entries in the `WriterState` object in AppendProcessor.

**To Reproduce**
This is observed frequently by the ConditionalOutputStreamImpl.

**Expected behavior**
Improve the code to ensure the latest stored event number is used. This will avoid redundant retries by the client.

**Screenshots**
NA

**Additional context**
NA

shrids:
**Change log description**  
Ensure the WriterState object contains the updated last event number for the given writer.

**Purpose of the change**  
Fixes #5767 

**What the code does** 
The segment writers reconnect to the SegmentStore on a connection drop. Post a connection drop the writers perform a setup append operation. As part of the setup append operation the data from the segment attributes are used to return the last store event number. In the current code, since this is a reconnect, the `WriterState `for this writer already exists;  this WriterState is used to fetch the `lasteventnumber` for the attribute update operation via `io.pravega.segmentstore.contracts.StreamSegmentStore#append`. Since this event number need not be the latest event number the SSS returns an exception due to bad attributes. This is retried again by the client.

This PR ensures the SSS attributes are used as the source of truth for the last event number and uses it for the WriterState updation too. This ensures we avoid redundant retries from the client.

**How to verify it**  

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -173,7 +173,16 @@ public class AppendProcessor extends DelegatingRequestProcessor {
                                 } else {
                                     // Last event number stored according to Segment store.
                                     long eventNumber = attributes.getOrDefault(writer, Attributes.NULL_ATTRIBUTE_VALUE);
-                                    this.writerStates.putIfAbsent(Pair.of(newSegment, writer), new WriterState(eventNumber));
+
+                                    // Create a new WriterState object based on the attribute value for the last event number for the writer.
+                                    // It should be noted that only one connection for a given segment writer is created by the client.
+                                    // The event number sent by the AppendSetup command is an implicit ack, the writer acks all events
+                                    // below the specified event number.
+                                    WriterState current = this.writerStates.put(Pair.of(newSegment, writer), new WriterState(eventNumber));
+                                    if (current != null) {
+                                        log.info("SetupAppend invoked again for writer {}. Last event number from store is {}. Prev writer state {}",
+                                                writer, eventNumber, current);
+                                    }
                                     connection.send(new AppendSetup(setupAppend.getRequestId(), newSegment, writer, eventNumber));
                                 }
                             } catch (Throwable e) {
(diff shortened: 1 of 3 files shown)


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -171,8 +171,18 @@ public void setupAppend(SetupAppend setupAppend) {
...
                                 } else {
+                                    // Last event number stored according to Segment store.
                                     long eventNumber = attributes.getOrDefault(writer, Attributes.NULL_ATTRIBUTE_VALUE);
-                                    this.writerStates.putIfAbsent(Pair.of(newSegment, writer), new WriterState(eventNumber));
+
+                                    // Create a new WriterState object based on the attribute value for the last event number for the writer.
+                                    // It should be noted that only one connection for a given segment writer is created by the client.
+                                    // The event number sent by the AppendSetup command is an implicit ack, the writer acks all events
+                                    // below the specified event number.
+                                    WriterState current = this.writerStates.put(Pair.of(newSegment, writer), new WriterState(eventNumber));
```
tkaitchuck, line 181: We need to worry about a race here. There could be concurrent modification of the writer state. So we want to have an `update` function to check that we aren't replacing a newer value.

shrids: Observed that all the append requests which are in pipeline, with requests to `store.Append(..)` sent before disconnect,  would have already fetched the reference to `WriterState` object. Hence adding a new `WriterState` object in response to a SetupAppend request by the writer on reconnect would imply that any updates to the previous writerState object would not affect the newly added writer state object.

Do let me know if I have missed any aspect here...

tkaitchuck: Sounds good.

tkaitchuck (approved):

PR #5738, opened 2021-02-12
Issue 5569: PravegaRequestProcessor warnings should include stack trace.

### LINKED ISSUE #5569: PravegaRequestProcessor warnings should include stack trace
**Problem description**
PravegaRequestProcessor warnings do not include stack trace making it harder to debug.

**Problem location**
PravegaRequestProcessor::handleException

PravegaRequestProcessor warnings should include stack trace. Do not do this for all warnings. Only for those of unknown origins. For example, `BadAttributeException` (others too) do not need a stack trace because they only originate from one point and indicate user error vs something wrong on our end.

sachin-j-joshi:

**Change log description**  
Issue 5569: PravegaRequestProcessor warnings should include stack trace.

**Purpose of the change**  
Fixes #5569 

**What the code does**  
PravegaRequestProcessor warnings now include stack trace.

**How to verify it**  
Build should pass


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -923,58 +923,58 @@ private Void handleException(long requestId, String segment, long offset, String
         };
 
         if (u instanceof StreamSegmentExistsException) {
-            log.info(requestId, "Segment '{}' already exists and cannot perform operation '{}'.",
-                     segment, operation);
+            log.info(requestId, "Segment '{}' already exists and cannot perform operation '{}'. Exception : {}",
```
andreipaduroiu, line 926: These will double-log the exception stack trace. No need to add an explicit `{}` to the message. Just put the exception as the last parameter.

andreipaduroiu: Also, some of these are well known exceptions; they can only come from one place and they are the result of user error. Adding a 20-40 line stack trace is going to significantly pollute the logs - that's the reason we removed them in the first place.

co-jo (approved):
LGTM.

RaulGracia:
@sachin-j-joshi @andreipaduroiu is this PR still valid?

sachin-j-joshi:
Closing as this is no more required.

Issue #5634, opened 2021-02-04
System Test Logging Enhancements


co-jo:
**Problem description**

In the current state of Pravega, retrieving logs from a systems test deployment requires manually querying each process/service. Running the test suite on a set of non-local machines (or cloud environments) introduces several complications with respect to logging.

* Services that perform restart operations can lose logs from previous 'lives'. 
* Long running tests can overwrite previous logs due to storage limitations.
* Querying each process can be tedious when many components are involved. 

**Problem location**

`test/system/...`

**Suggestions for an improvement**

Create a service that exists outsides of the lifecycle of core Pravega components whose job is to aggregate logs into a centralize location. 
* A basic log sink will be able to preserve the logs across multiple restarts.
* Centralizes the location in which logs can be found.

RaulGracia:
Closed via #5744.

PR #5744, opened 2021-02-13
Issue 5634: System Test Log Retention

co-jo:
**Change log description**  

* Adds a new step in the gradle system test deployment: `runFluentBitInstall`, which calls to a `fluentBitSetup.sh` script.
* On a call to the `startK8SystemTests` task, this script will setup fluent-bit and the PVC (using the default storageclass: `standard`) used to persist all logs generated by _all_ pods.

**Purpose of the change**  
Completes #5634.

**What the code does**  

* A `fluent-bit` DaemonSet is installed which will watch the logs of all pods and aggregate them into a single directory on the defined PVC. As configured, it avoids storing any extra information provided by fluent-bit or Kubernetes as apart of the logs. Meaning that fluent-bit stores _only_ the output generated by the application, and no extra metadata (as it does by default).

* An ad-hoc Kubernetes deployment is also deployed, which checks the current size of these logs at regular intervals, and rotates them once they exceed the defined threshold. When a rotation happens, it will will convert the compressed log's epoch suffix to a more human readable format (e.g ` 2021-02-16T00:17:10Z`).
 
The defaults are set to check at 10 second intervals and rotate when a log exceeds 10 MB.

**How to verify it**  

Run system tests, and verify that a pod can mount the created PVC and read the logs.


build.gradle
```diff
@@ -1004,7 +1004,7 @@ project('test:system') {
         environment "zookeeperOperatorVersion",System.getProperty("zookeeperOperatorVersion", "latest")
         // The below properties are used to to specify the pravega published chart and repository to deploy pravega,bookkeeper & zookkeeper operators using helm docker image properties
         environment "publishedChartName",System.getProperty("publishedChartName", "pravega")
-        environment "helmRepository",System.getProperty("helmRepository", "pravega")
+        environment "helmRepository",System.getProperty("helmRepository", "https://charts.pravega.io")
```
co-jo, line 1007: `pravega` is not a valid default..

build.gradle
```diff
@@ -1013,8 +1013,19 @@ project('test:system') {
         commandLine './kubernetes/setupTestPod.sh'
     }
 
+    task runFluentBitUninstall(type: Exec) {
+        environment 'KEEP_PVC', System.getProperty("keepSystemTestLogs", "")
+        commandLine './kubernetes/fluentBitSetup.sh', 'uninstall'
+    }
+
+    task runFluentBitInstall(type: Exec) {
+        dependsOn 'runFluentBitUninstall'
```
co-jo, line 1022: This actually diverges from the current bevhaviour for other components, i.e. the components deployed for a system test must be explicitly cleaned up before rerunning the `startK8SystemTests` task. Should I remove this to keep with that behaviour?

test/system/kubernetes/fluentBitSetup.sh
```diff
@@ -0,0 +1,360 @@
...
+        persistentVolumeClaim:
+          claimName: $PVC_NAME
+EOF
+}
+
+####################################
+#### Main Process
+####################################
+
+# Kuberentes produces a log file for each pod (/var/log/containers/*). Each line
```
ManishKumarKeshri, line 224: spelling.

ManishKumarKeshri (approved):

test/system/kubernetes/fluentBitSetup.sh
```diff
@@ -0,0 +1,360 @@
+#!/usr/bin/env bash
```
andreipaduroiu, line 1: This file needs a copyright header. Find one in another `.sh` file and paste it here please.

andreipaduroiu (changes requested):

build.gradle
RaulGracia: Nit: looks like there is missing a space after comma in that new line (and in previous ones).

build.gradle
```diff
@@ -1013,8 +1013,19 @@ project('test:system') {
         commandLine './kubernetes/setupTestPod.sh'
     }
 
+    task runFluentBitUninstall(type: Exec) {
+        environment 'KEEP_PVC', System.getProperty("keepSystemTestLogs", "")
+        commandLine './kubernetes/fluentBitSetup.sh', 'uninstall'
+    }
+
+    task runFluentBitInstall(type: Exec) {
```
RaulGracia, line 1021: What about if the cluster already has a fluentbit service running? Is this going to install/uninstall it on the default namespace anyway?

co-jo: By default it will only uninstall a helm installation with the name `pravega-fluent-bit` on the default namespace yes, so there should be a low probability of conflicts with other fluent-bit deployments unrelated to pravega. However I'll remove it anyways as other components do not follow this pattern.

build.gradle
```diff
@@ -999,22 +999,27 @@ project('test:system') {
...
-        environment "desiredPravegaCMVersion",System.getProperty("desiredPravegaCMVersion", "0.9.0")
-        environment "desiredBookkeeperCMVersion",System.getProperty("desiredBookkeeperCMVersion", "0.9.0")
+        environment "desiredPravegaCMVersion", System.getProperty("desiredPravegaCMVersion", "0.9.0")
+        environment "desiredBookkeeperCMVersion", System.getProperty("desiredBookkeeperCMVersion", "0.9.0")
         environment "tlsEnabled", System.getProperty("tlsEnabled", "false")
 
         commandLine './kubernetes/setupTestPod.sh'
     }
 
+    task runFluentBitInstall(type: Exec) {
```
RaulGracia, line 1016: What about adding a task to download the logs assuming they are stored in the PVC?

RaulGracia (approved):
LGTM. I assume that this has been tested for a while on Jenkins (if not, do so before we mege this).

RaulGracia (changes requested):
Looks good, but do we have real executions of system tests with this PR so we can evaluate how it works?

ManishKumarKeshri:
Looks good. I saw that log files ":" in their name, maybe replace it with another character if possible.

ManishKumarKeshri (approved):

test/system/kubernetes/fluentBitSetup.sh
```diff
@@ -0,0 +1,665 @@
...
+
+# Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
```
RaulGracia, line 3: One last thing: this header is obsolete, please change it to the new one updated few days ago.

RaulGracia (changes requested):

RaulGracia (approved):

PR #5751, opened 2021-02-15
Issue 5750: EventProcessor's readergroup failover fails if there are readers without a position in the zookeeper

### LINKED ISSUE #5750: EventProcessor's readergroup failover fails if there are readers without a position in the zookeeper
**Problem description**
We see following error in controller, which eventually is retried indefinitely and eventually causes oom as gc doesnt happen fast enough. 
```
ERROR i.p.c.s.e.ControllerEventProcessors - Error notifying failure of process=172.28.17.7-d900cd95-0347-4109-8087-a490924e9d58 in event processor group EventProcessorGroup[commitStreamReaders]
io.pravega.controller.store.checkpoint.CheckpointStoreException: ReaderGroup is not empty.
        at io.pravega.controller.store.checkpoint.ZKCheckpointStore.removeReaderGroup(ZKCheckpointStore.java:141)
        at io.pravega.controller.eventProcessor.impl.EventProcessorGroupImpl.notifyProcessFailure(EventProcessorGroupImpl.java:268)
        at io.pravega.controller.server.eventProcessor.ControllerEventProcessors.lambda$handleOrphanedReaders$33(ControllerEventProcessors.java:412)
```

The reason for this is that we create the RG-Data znode for a readergroup per controller instance where we record the list of readers in it. but these readers may not have a checkpointed child znode under this rg-data.
This could lead to a situation where rg-data has the reference to reader but there is no child node. 
in this case failover of such a controller does not happen due to the bug which only looks at children nodes and removes those readers. 

**Problem location**
ZkCheckpointStore.java

Fix the bug

shiveshr:

**Change log description**  
Fixes a bug where if RG-Data znode has the information about readers but the child znode for reader was not created then attempting to delete readergroup fails.

**Purpose of the change**  
Fixes #5750 

**What the code does**  
The readergroup's readers positions should return readers if they are part of readergroup data even if their child znode with position was not present. 

**How to verify it**  
Unit test added.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/checkpoint/ZKCheckpointStore.java
+++ b/controller/src/main/java/io/pravega/controller/store/checkpoint/ZKCheckpointStore.java
@@ -80,6 +80,8 @@ class ZKCheckpointStore implements CheckpointStore {
     public Map<String, Position> getPositions(String process, String readerGroup) throws CheckpointStoreException {
         Map<String, Position> map = new HashMap<>();
         String path = getReaderGroupPath(process, readerGroup);
+        ReaderGroupData rgData = groupDataSerializer.deserialize(ByteBuffer.wrap(getData(path)));
+        rgData.getReaderIds().forEach(x -> map.put(x, null));
         for (String child : getChildren(path)) {
             Position position = null;
             byte[] data = getData(path + "/" + child);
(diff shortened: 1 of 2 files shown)


shrids (approved):
LGTM

controller/src/main/java/io/pravega/controller/store/checkpoint/ZKCheckpointStore.java
```diff
@@ -80,6 +80,8 @@ public void setPosition(String process, String readerGroup, String readerId, Pos
     public Map<String, Position> getPositions(String process, String readerGroup) throws CheckpointStoreException {
         Map<String, Position> map = new HashMap<>();
         String path = getReaderGroupPath(process, readerGroup);
+        ReaderGroupData rgData = groupDataSerializer.deserialize(ByteBuffer.wrap(getData(path)));
+        rgData.getReaderIds().forEach(x -> map.put(x, null));
         for (String child : getChildren(path)) {
             Position position = null;
             byte[] data = getData(path + "/" + child);
```
shrids, line 87: I think we should add a precondition check for the return value of `map.put(` at line 91 (below). If the put operation returns a null the check should fail ...

PR #5757, opened 2021-02-17
Issue 5700: Run Subset of System Tests

### LINKED ISSUE #5700: Run Subset of System Tests
**Problem description**

Currently when the system tests are deployed to a cluster _all_ system tests are run and there exists no programmatic way to isolate individual system tests.  This is particularly useful because system tests can take extended periods of time to complete.

**Problem location**

`test/system`

* Allow the user to supply via some configuration/parameter the subset of system tests to run.

co-jo:
**Change log description**  

* The `startK8SystemTests` and `startSystemTests` gradle tasks now accept two new flags: `excludeSystemTests` and `includeSystemTests`, which accept a `,` delimited list of patterns to include or exclude a test. These patterns are forwarded to gradle so the same pattern processing applies. 

* Enable one to skip cluster installation during a systems test deployment and provide details for an existing cluster.
* Add option to skip log collection via `stdout` stream of the K8 pod(s). This will be helpful to avoid redundant work if the cluster has other log collection facilities, such as fluent-bit.

**Purpose of the change**  
Closes #5700 

**What the code does**  

* Adds a means to configure the set of tests to run for the system test based gradle tasks. As mentioned above, the flags are mostly forwarded to the underlying `Test.excludes/includes` field, unless it detects the input string does not contain a file path delimiter or ends in '.class'. In this case it modifies a string such as `-DexcludeTests=ZooKeeperTest` to properly match against the expected test.

* Adds necessary flags to allow use of the system test code on an existing cluster. The appropriate flags are forwarded to both the system test deployment code, and to the test pods that are deployed on the Kubernetes cluster.

**How to verify it**  

* Run system tests on an existing cluster by providing the appropriate flags.


RaulGracia:
@co-jo please, let us know when this PR is ready for review.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/ZookeeperK8sService.java
```diff
@@ -73,34 +73,34 @@ public void clean() {
...
-                        .exceptionally(t -> {
-                           log.warn("Exception observed while checking status of pod: {}. Details: {} ", getID(), t.getMessage());
-                           return false;
-                       }).join();
+                .thenApply(statuses -> statuses.stream()
+                        .filter(podStatus -> podStatus.getContainerStatuses()
+                                .stream()
+                                .allMatch(st -> st.getState().getRunning() != null))
+                        .count() * 2 - statuses.size())
+                .thenApply(netCount -> netCount > 0)
```
co-jo, line 81: Previously, it was expected that an exact number of instances are available, which places heavy restrictions on the compatible clusters that can be used (if not deployed via system test scripts). Change it so that only a majority of nodes are required to be in a running state.

build.gradle
```diff
@@ -1038,6 +1048,16 @@ project('test:system') {
     task startK8SystemTests(type: Test) {
         dependsOn 'copyTestsToK8sCluster', 'runFluentBitInstall' // ensure test artifacts are copied to the cluster.
         description 'Used to invoke K8s based System Tests, example usage: gradle startK8SystemTests'
+
+        def excludedTests = System.getProperty('excludeSystemTests')
```
RaulGracia, line 1052: Wondering if we could have this code block in a function rather than repeat it twice in the same file.

build.gradle
```diff
@@ -1058,25 +1080,35 @@ project('test:system') {
...
-        systemProperty "tier2Config", System.getProperty("tier2Config")
-        systemProperty "pravegaOperatorVersion", System.getProperty("pravegaOperatorVersion", "latest")
-        systemProperty "bookkeeperOperatorVersion", System.getProperty("bookkeeperOperatorVersion", "latest")
-        systemProperty "zookeeperOperatorVersion", System.getProperty("zookeeperOperatorVersion", "latest")
-        systemProperty "desiredPravegaCMVersion", System.getProperty("desiredPravegaCMVersion", "0.9.0")
-        systemProperty "desiredBookkeeperCMVersion", System.getProperty("desiredBookkeeperCMVersion", "0.9.0")
-        systemProperty "publishedChartName", System.getProperty("publishedChartName", "pravega")
-        systemProperty "helmRepository", System.getProperty("helmRepository", "pravega")
+        // tier2 configuration, specified as comma separated key values k1=v1,k2=v2
+        systemProperty "tier2config", System.getProperty("tier2config")
```
RaulGracia, line 1084: This change seems gratuitous and it breaks the camelCase style of the rest of parameters. Also, this could be breaking other Jenkins jobs that rely on the former property name.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
```diff
@@ -119,7 +122,7 @@ private void verifyPravegaPodRestart(Map<String, V1ContainerStatus> podStatusBef
     private V1Pod getTestPod(String className, String methodName, String podName) {
         log.info("Running test pod with security enabled :{}, transport enabled: {}", Utils.AUTH_ENABLED, Utils.TLS_AND_AUTH_ENABLED);
         V1Pod pod =  new V1PodBuilder()
-                .withNewMetadata().withName(podName).withNamespace(NAMESPACE).withLabels(ImmutableMap.of("POD_NAME", podName, "app", APP)).endMetadata()
+                .withNewMetadata().withName(podName).withNamespace(NAMESPACE).withLabels(ImmutableMap.of("POD_NAME", podName, "app", "pravega-system-tests")).endMetadata()
```
RaulGracia, line 125: Maybe you could leave the `APP` constant as it was, so we can avoid having this hardcoded string in the middle there.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
```diff
@@ -73,8 +73,11 @@
                 .thenCompose(v -> client.createClusterRoleBinding(getClusterRoleBinding())) // ensure test pod has cluster admin rights.
                 .thenCompose(v -> client.deployPod(NAMESPACE, pod)) // deploy test pod.
                 .thenCompose(v -> {
+                    CompletableFuture<Void> logDownload = CompletableFuture.completedFuture(null);
                     // start download of logs.
-                    CompletableFuture<Void> logDownload = client.downloadLogs(pod, "./build/test-results/" + podName);
+                    if (!Utils.isSkipLogDownloadEnabled()) {
```
RaulGracia, line 78: Ok, good, this should be speeding up things.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/ZookeeperK8sService.java
```diff
@@ -77,8 +77,8 @@ public boolean isRunning() {
                                                       .filter(podStatus -> podStatus.getContainerStatuses()
                                                                                     .stream()
                                                                                     .allMatch(st -> st.getState().getRunning() != null))
-                                                      .count())
-                        .thenApply(runCount -> runCount == DEFAULT_INSTANCE_COUNT)
+                                                      .count() * 2 - statuses.size())
```
RaulGracia, line 80: Why do we need this change?

RaulGracia (changes requested):
Looks good, jut few minor comments.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
co-jo: Yes, that was an unintended change.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/ZookeeperK8sService.java
co-jo: See: https://github.com/pravega/pravega/pull/5757#discussion_r608818218

build.gradle
co-jo: Right.. this change was not intentional.

build.gradle
```diff
@@ -970,6 +970,15 @@ project('test:system') {
...
         ext.repoUrl = project.hasProperty("repoUrl") ? project.repoUrl : ""
 
+        def excludedTests = System.getProperty('excludeSystemTests')
+        def includedTests = System.getProperty('includeSystemTests')
+        // Let included tests take precedence if both are specified.
+        if (includedTests != null) {
+            includes = includedTests.split(',').collect { getSystemTestClass(it) }
+        } else if (excludedTests != null ) {
+            excludes = excludedTests.split(',').collect { getSystemTestClass(it) }
+        }
```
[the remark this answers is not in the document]
co-jo: There are two reasons. After some experimentation I've concluded that due to how the test framework is setup, we are unable to leverage the `--tests` flag / `test.filter` functionality. It works at the class level, but breaks down once you specify the method to use. 

The nice thing about the `TestFilter` class is that it provides much of the work I've manually implemented with `getSystemTestClass`. The downside is that there is no 'exclude' filter, it is seemingly only inclusive. Having an option to exclude tests makes certain use cases much easier, so we still want that. Because of this, I see no point of using `TestFilter` functionality if we cannot isolate specific methods.

ManishKumarKeshri (approved):

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
RaulGracia: Ok, so why haven't you reverted it?

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
```diff
@@ -231,4 +231,9 @@ private static boolean isTLSEnabled() {
         String tlsEnabled = Utils.getConfig("tlsEnabled", "false");
         return Boolean.valueOf(tlsEnabled);
     }
+
+    public static boolean isSkipLogDownloadEnabled() {
+        String config = getConfig("skipLogDownload", "false");
+        return config.trim().equalsIgnoreCase("true") ? true : false;
```
RaulGracia, line 237: Wouldn't just `return config.trim().equalsIgnoreCase("true")` be equivalent?

RaulGracia (changes requested):
Couple of minor comments.

test/system/src/main/java/io/pravega/test/system/framework/Utils.java
co-jo: Yes. I just followed the above calls for consistency purposes.

RaulGracia (approved):

PR #5773, opened 2021-02-21
Issue 5772: (SLTS) - BaseMetadataStore does not evict entries from buffer. 

### LINKED ISSUE #5772: SLTS - BaseMetadataStore does not evict entries from buffer.
**Describe the bug**
Because of bug in logic where the entries in committed transaction are cleared pre-maturely, they are not marked inactive as expected. As such, this results in  BaseMetadataStore not evicting these entries from buffer.

Instead of clearing the keys here
https://github.com/pravega/pravega/blob/master/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java#L300

They should be cleared after this line
https://github.com/pravega/pravega/blob/master/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java#L308

**To Reproduce**
This can be easily seen in any run of SLTS where metrics are available.

**Expected behavior**
BaseMetadataStore should evict non-active entries from buffer when size exceeds set limits.

**Screenshots**

image

**Additional context**
None

sachin-j-joshi:
**Change log description**  
Fix bug that prevented BaseMetadataStore from evicting entries from buffer. 

**Purpose of the change**  
Fix #5772 

**What the code does**  
Fix bug that prevented BaseMetadataStore from evicting entries from buffer. 

**How to verify it**  
Test should pass.
Metrics should show entries being evicted regularly.

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
@@ -188,6 +188,7 @@ abstract public class BaseMetadataStore implements ChunkMetadataStore {
     /**
      * Keep count of records in buffer. ConcurrentHashMap.size() is an expensive operation.
      */
+    @Getter
     private final AtomicInteger bufferCount = new AtomicInteger(0);
 
     /**
@@ -297,7 +298,6 @@ abstract public class BaseMetadataStore implements ChunkMetadataStore {
                 .thenRunAsync(() -> {
                     //  Step 5 : Mark transaction as commited.
                     txn.setCommitted();
-                    txnData.clear();
                 }, executor)
                 .whenCompleteAsync((v, ex) -> {
                     if (shouldReleaseKeys.get()) {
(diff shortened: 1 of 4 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
```diff
@@ -188,6 +189,7 @@
     /**
      * Keep count of records in buffer. ConcurrentHashMap.size() is an expensive operation.
      */
+    @Getter
```
andreipaduroiu, line 192: I know this is used just for testing, but you are essentially exposing your inner components to the outside. Someone could take this and mess around with the values inside.

I suggest you convert this to a **packgage-private** hand-written getter and add the `@VisibleForTesting` tag. That way, if you just return the value, nobody could modify the contents.

andreipaduroiu (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/BaseMetadataStore.java
sachin-j-joshi: ok

sachin-j-joshi: good catch. Fixed

andreipaduroiu (approved):

PR #5778, opened 2021-02-23
Issue 5777: Remove generation parameter from Controller deleteReaderGroup API

### LINKED ISSUE #5777: Controller deleteReaderGroup API should not take `generation` as a parameter.
**Describe the bug**
Currently the deleteReaderGroup API takes ReaderGroup generation as parameter, which is unnecessary and causes problems with deleting the ReaderGroup when ReaderGroupConfig on Controller and StateSynchronizer are not in sync.

**To Reproduce**
When ReaderGroupConfig on Controller is ahead of RGConfig on StateSynchronizer, invoking ReaderGroupManager.deleteReaderGroup() always causes it to fail.

**Expected behavior**
Controller API deleteReaderGroup() should only take ReaderGroup scope, name and Id as parameters for deletion.
The `Id` can guard against an old delete request accidentally deleting an existing Reader Group re-created with same name.

pbelgundi:

**Change log description**  
This PR removes the generation param from deleteReaderGroup API on Controller.

**Purpose of the change**  
Fixes #5777

**What the code does**  
Remove generation parameter from Controller deleteReaderGroup API
Changed deleteReaderGroup() API on ReaderGroupManager (Client), such that if StateSynchronizer Stream for a ReaderGroup is not found, ReaderGroupId from Controller is fetched and deleteReaderGroup is issued on Controller.

**How to verify it**  
All unit, intergration and system tests should pass.

### DIFF
--- a/client/src/main/java/io/pravega/client/admin/impl/ReaderGroupManagerImpl.java
+++ b/client/src/main/java/io/pravega/client/admin/impl/ReaderGroupManagerImpl.java
@@ -21,12 +21,7 @@ import io.pravega.client.state.InitialUpdate;
 import io.pravega.client.state.StateSynchronizer;
 import io.pravega.client.state.SynchronizerConfig;
 import io.pravega.client.state.Update;
-import io.pravega.client.stream.ReaderGroup;
-import io.pravega.client.stream.ReaderGroupConfig;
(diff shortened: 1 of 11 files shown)


shiveshr (approved):
LGTM

PR #5779, opened 2021-02-23
Issue 5776: resetReaderGroup fails if there are pending updates

### LINKED ISSUE #5776: resetReaderGroup fails if there are pending updates
**Describe the bug**
When resetting a reader group, it is possible for the controller's reader group to be ahead of the client's config generation due to a prior update. Currently this is handled on `ReaderGroupStateManager` by the readers who periodically update the client's config generation to the controller's. 

However, in the absence of readers actively reading events, the client will never catch up and the controller will never allow the reset to happen. The client will keep retrying with the response from the controller being `INVALID_CONFIG`.

**To Reproduce**
1. Create a reader group on the client.
2. Call the controller method, `controller.updateReaderGroup` to update the reader group outside the client.
3. Try to reset the reader group on the client.
4. See error.

**Expected behavior**
The reset should go through with no errors.

anirudhkovuru:
Signed-off-by: anirudhkovuru <anirudhkovuru@gmail.com>

**Change log description**  
`resetReaderGroup` now makes sure to update the `ReaderGroupState` when necessary in order to keep consistent with the controller.   

**Purpose of the change**  
Fixes #5776 

**What the code does**  
- Added new exception `ReaderGroupConfigRejectedException` to indicate when the controller returns an `INVALD_CONFIG`.
- When this new exception is encountered, the code updates its config to the one on the controller and then tries the update again.

**How to verify it**  
All tests pass and build should succeed.

### DIFF
--- a/client/src/main/java/io/pravega/client/control/impl/Controller.java
+++ b/client/src/main/java/io/pravega/client/control/impl/Controller.java
@@ -138,6 +138,7 @@ public interface Controller extends AutoCloseable {
      * @param rgName Stream name.
      * @param config ReaderGroup configuration.
      * @throws IllegalArgumentException if Stream does not exist.
+     * @throws ReaderGroupConfigRejectedException if the provided ReaderGroupConfig is invalid
      * @return A future which will throw if the operation fails, otherwise
      *         the subscriber was updated in Stream Metadata and a long indicating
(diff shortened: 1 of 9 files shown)


shiveshr (approved):
LGTM

PR #5792, opened 2021-03-02
Issue 5211: Adds EventHelperFuture to fix NPE in StreamMetadataTasks

### LINKED ISSUE #5211: Flaky test StreamMetadataTest#testMetadataOperations on 0.8.0rc1 code
**Problem description**
While testing 0.8.0rc1 I saw this flaky test.
I am using JDK11 on Mac

i```
o.pravega.test.integration.controller.server.StreamMetadataTest > testMetadataOperations FAILED
        at io.pravega.controller.task.Stream.StreamMetadataTasks.sealStream(StreamMetadataTasks.java:494)
        at io.pravega.controller.server.ControllerService.sealStream(ControllerService.java:224)
        at io.pravega.controller.server.eventProcessor.LocalController.sealStream(LocalController.java:212)
        at io.pravega.test.integration.controller.server.StreamMetadataTest.testMetadataOperations(StreamMetadataTest.java:93)
```

**Problem location**
The test

Make the test consistently pass

### PR DESCRIPTION
**Pull request by kotlasaicharanreddy**
**Change log description**  

- Introduction of eventhelper in `StreamMetadataTasks.java` is not initialized in constructor but after the object after an explicit `initialize()` method.
- This causes `Null Pointer Exception` if request is received before initialize can happen.
- Added `EventHelperFuture` to fix NullPointerException.

**Purpose of the change**  
Fixes #5211

**What the code does**  

- Replaces `eventhelper` with `EventHelperFuture` accordingly. 

**How to verify it**  

- `testeventHelperNPE()` test should always pass.
- Build should succeed and all test must pass.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
+++ b/controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
@@ -338,15 +338,23 @@ public class StreamMetadataTasks extends TaskBase {
                  .thenCompose(complete -> {
                      if (!complete) {
                          //3. Create Reader Group Metadata inside Scope and submit the event
-                         return eventHelper.addIndexAndSubmitTask(buildCreateRGEvent(scope, rgName, config, requestId, createTimestamp),
-                                 () -> streamMetadataStore.addReaderGroupToScope(scope, rgName, config.getReaderGroupId()))
+                         return validateReaderGroupId(config)
+                                 .thenCompose(conf -> eventHelper.addIndexAndSubmitTask(buildCreateRGEvent(scope, rgName, conf, requestId, createTimestamp),
+                                 () -> streamMetadataStore.addReaderGroupToScope(scope, rgName, conf.getReaderGroupId()))
                                          .thenCompose(x -> eventHelper.checkDone(() -> isRGCreated(scope, rgName, executor))
-                                         .thenCompose(done -> buildCreateSuccessResponse(scope, rgName)));
+                                         .thenCompose(done -> buildCreateSuccessResponse(scope, rgName))));
                      }
                      return buildCreateSuccessResponse(scope, rgName);
                  });
          });
         }, e -> Exceptions.unwrap(e) instanceof RetryableException, READER_GROUP_OPERATION_MAX_RETRIES, executor);
+    }
+
+    private CompletableFuture<ReaderGroupConfig> validateReaderGroupId(ReaderGroupConfig config) {
+        if (ReaderGroupConfig.DEFAULT_UUID.equals(config.getReaderGroupId())) {
+            return CompletableFuture.completedFuture(ReaderGroupConfig.cloneConfig(config, UUID.randomUUID(), 0L));
+        }
+        return CompletableFuture.completedFuture(config);
     }
 
     public CompletableFuture<Boolean> isRGCreationComplete(String scope, String rgName) {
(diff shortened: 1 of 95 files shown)

### CODE REVIEW

File: controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java [hunk 1]
```diff
@@ -135,15 +135,15 @@
     private static final int READER_GROUP_OPERATION_MAX_RETRIES = 10;
 
     private final AtomicLong retentionFrequencyMillis;
-    
+
```
shiveshr, line 138: nit: revert please

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -1853,8 +1854,14 @@ public TaskBase copyWithContext(Context context) {
...
-        if (eventHelper != null) {
-            eventHelper.close();
+        if (eventHelperFuture.isDone()) {
+            eventHelperFuture.thenAccept(eventHelper -> {
+                if (eventHelper != null) {
+                    eventHelper.close();
+                }
+            }).join();
+        } else {
+            eventHelperFuture.cancel(true);
```
andreipaduroiu, line 1864: This isn't going to have any effect. 

Furthermore, above (in `initialize`) when you set the value, that call will have no effect if this future is already completed (whether successfully or with exception). That means that the `EventHelper` object you create will be a source of resource leaks as you never close it.

Please rewrite your code to handle that situation.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -190,17 +190,18 @@ private StreamMetadataTasks(final StreamMetadataStore streamMetadataStore,
...
         this.retentionClock = new AtomicReference<>(System::currentTimeMillis);
+        this.eventHelperFuture = new CompletableFuture<>();
         this.setReady();
     }
 
     @Synchronized
     public void initializeStreamWriters(final EventStreamClientFactory clientFactory,
                                         final String streamName) {
-        this.eventHelper = new EventHelper(clientFactory.createEventWriter(streamName,
+        this.eventHelperFuture.complete(new EventHelper(clientFactory.createEventWriter(streamName,
```
andreipaduroiu, line 200: This is a bad pattern. Never complete a future while holding a lock. You are holding the instance-level lock here with `@Synchronized`.

I recommend you rewrite this code to either remove the `@Synchronized` or complete this future outside of the scope of the lock.

andreipaduroiu (changes requested):

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shiveshr: we should do the following:
create eventHelper and set it in a lock protected variable and use a boolean to indicate if it needs to be set or not.
init method:
```
synchronize(lock) {
if(toSetEventHelper) {
eventhelper = new EventHelper();
toSetEventHelper = false;
}
}
// after we release the lock, if close was invoked then eventhelper may have been closed. but completing the future with that is safe (and its ok even if eventhelperfuture was cancelled in close method).
if (eventHelper != null) {
eventHelperFuture.complete(eventHelper);
}

```
close method:
```
synchronize(lock) {
toSetEventHelper = false;
if (eventhelper != null) {
eventHelper.close();
}
}
eventhelperfuture.cancel(true);
```

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
kotlasaicharanreddy: Changes are done as per your recommendation.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
kotlasaicharanreddy: Changes are done as per suggested.

andreipaduroiu (approved):
My changes have been approved. @shiveshr it's in your hands now.

shiveshr (approved):
LGTM

PR #5794, opened 2021-03-02
Issue 5793: Add root cause to bug report template.

### LINKED ISSUE #5793: Add observation about root cause to bug report
Clarify that the root cause should be added when known to the bug report template.

fpj:

**Change log description**  
* Adds a comment about describing the root cause in the bug report issue template.

**Purpose of the change**  
Fixes #5793 

**What the code does**  
There is no code change, only a change to issue template.

**How to verify it**  
The change does not affect the build, it should be reflected in the bug report template.


sachin-j-joshi (commented):
Should we have a separate "Actual Behavior" section ?
We have a description section and Expected behavior already.

fpj:
@sachin-j-joshi what about a section called `Expected and actual incorrect behavior` ? Describe the expected behavior and the observed one.

Note that this is different from the "root cause", which is what is causing the incorrect behavior.

sachin-j-joshi:
Yes. That works.

RaulGracia (approved):

PR #5795, opened 2021-03-02
Issue 5791: Fix LocalPravegaEmulator dependencies for external projects

### LINKED ISSUE #5791: LocalPravegaEmulator fails with missing protobuf and jjwt classes
**Describe the bug**
In 0.9.0 snapshots, using `LocalPravegaEmulator` from `pravega-standalone` in external tests fails due to missing protobuf and jjwt classes.

**To Reproduce**
1. Run test in external project here: https://github.com/derekm/local-pravega-emulator/commit/d18622bb502d0cb4fdfb80343c117bcd80b018b7

**Expected behavior**
`LocalPravegaEmulator` should work out-of-the-box with minimal dependencies in an external project.

**Root cause**
* `withoutJaxbAndJjwt` Dependency Excludes in `pravega-client`'s use of `pravega-shared-security`.
* `pravega-client`'s & `pravega-shared-controller-api`'s use of `grpc-protobuf` 1.24.2 brings in `protobuf-java` 3.9.0, but `protobuf-java` is elsewhere inconsistently force-upgraded to 3.13.0.

derekm:
**Change log description**  
Overrides gRPC's older transitive protobuf-java version with our force-upgraded protobuf-java version, otherwise external projects pick up an older protobuf-java version that Pravega was not compiled against.

Adds jjwt dependency to pravega-standalone for users of LocalPravegaEmulator. For some reason, pravega-standalone requires jjwt, and jjwt is otherwise excluded as a transitive dependency by a rule in pravega-client.

**Purpose of the change**  
Fixes #5791

**What the code does**  
Pins protobuf-java to our force-upgraded version for subprojects that use `apply plugin: 'com.google.protobuf'`. Includes jjwt in pravega-standalone to solve a class not found exception suffered by users of LocalPravegaEmulator.

**How to verify it**  
Using build artifacts from this branch, run the test in the external project here: derekm/local-pravega-emulator@d18622b


build.gradle
```diff
@@ -648,6 +648,8 @@ project('shared:controller-api') {
         compile "io.grpc:grpc-protobuf:" + grpcVersion
         compile "io.grpc:grpc-stub:" + grpcVersion
         compile group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: nettyBoringSSLVersion
+        // override grpc's transitive protobuf version with allprojects' force-upgraded version
+        compile group: 'com.google.protobuf', name: 'protobuf-java', version: protobufProtocVersion
```
shrids, line 652: From the dependency list for Pravega-client we can observe that the client (via `shared:controller-api` ) already depends on `protobuf-java` with the correct version i.e. (protobufProtocVersion=3.13.0)

Dependency Snip :
```
|    +--- io.grpc:grpc-protobuf:1.24.2
|    |    +--- io.grpc:grpc-api:1.24.2 (*)
|    |    +--- com.google.protobuf:protobuf-java:3.9.0 -> 3.13.0
```
The above snip indicates the project already depends on 3.13.0 version for `protobuf-java` dependency.  This dependency is added as part of the `protobuf.gradle` which auto added when apply plugin : `com.google.protobuf` is used .
Hence I think this change here is not required...

Complete depencdency output : 
[Client-deps.txt](https://github.com/pravega/pravega/files/6080971/Client-deps.txt)

derekm: @shrids -- The problem is that that override does not become a part of the jar's Maven pom xml generated by Gradle, so when the jar is used external to the Gradle build, the user will get the wrong version of the dependency. Adding this line ensures that the jar will also require the version of protobuf-java that it was build with. You can check this by bringing the jar into an external project.

shrids: Thanks for the clarification. 
Could you add this dependency inside the protobuf.gradle file so that any project that does apply plugin : com.google.protobuf overrides the version of `protobuf-java` ?

derekm: @shrids Interesting idea! I'll check if that works.

andreipaduroiu:
@derekm What is the status on this PR? Are we still pursuing it?

derekm:
@andreipaduroiu Yes, this is still a valid PR... But @shrids had a good suggestion for relocating where to cement the protobuf override version, which I need to test. I've been trying to find time to pick this back up again recently. I definitely want to get this in as soon as possible, as it impacts users with tests.

FWIW, I've also started using testcontainers.org for Java integration tests, which overlaps with LocalPravegaEmulator, but it adds more dependencies for the system running the tests (that is, requires Docker).

derekm:
@andreipaduroiu @shrids -- This PR is ready for consideration. Relocating the protobuf-java version override to protobuf.gradle, as @shrids suggested, works great.

I have tested that this fixes external uses of LocalPravegaEmulator in JUnit testing via https://github.com/derekm/local-pravega-emulator and https://github.com/pravega/snmp-trap-handler.

Integration tests can use LocalPravegaEmulator without any missing dependencies or missing version overrides.

andreipaduroiu (approved):
I don't see any problems with this.

shrids (approved):
+1

PR #5800, opened 2021-03-03
Issue 5798: (SLTS) Fix close in GarbageCollector, ExtendedS3ChunkStorage and HDFSChunkStorage.

### LINKED ISSUE #5798: SLTS garbage collector does not close ScheduledExecutorService instance it creates in constructor.
**Describe the bug**
SLTS garbage collector does not close ScheduledExecutorService instance it creates in constructor.
The base class AbstractThreadPoolService::close does not actually close the ScheduledExecutorService passed in as the constructor parameter. 

**To Reproduce**
1. cause several restarts of segment store containers.
2. Take a memory dump of segment store.
3. Look at the GarbageCollection instances and ScheduledExecutorService  pointed by executor field.

**Expected behavior**
All resources created should be released in close().

**Actual behavior**
In several recent heap dumps we found that ScheduledExecutorService created by GarbageCollector is not closed properly.

**Screenshots**
None

**Additional context**
This can possibly lead to a leak , where running executor is holding on to the reference of GarbageCollector which holds reference to TableBasedMetadataStore which holds reference to StreamSegmentContainer

sachin-j-joshi:
Issue 5798: (SLTS) Fix close in GarbageCollector, ExtendedS3ChunkStorage and HDFSChunkStorage.

**Change log description**  
Fix close in GarbageCollector, ExtendedS3ChunkStorage and HDFSChunkStorage.
Make serializer static in TableBasedMetadataStore.

**Purpose of the change**  
Fixes #5798 

**What the code does**  
Fix close in GarbageCollector, ExtendedS3ChunkStorage and HDFSChunkStorage.

**How to verify it**  
Test should pass.

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
@@ -248,24 +248,21 @@ public class GarbageCollector extends AbstractThreadPoolService implements AutoC
         // Find chunks to delete.
         val chunksToDelete = new ArrayList<GarbageChunkInfo>();
         int count = 0;
-        try {
-            // Wait until you have at least one item or timeout expires.
-            GarbageChunkInfo info = garbageChunks.poll(config.getGarbageCollectionDelay().toMillis(), TimeUnit.MILLISECONDS);
-            log.trace("{}: deleteGarbage - retrieved {}", traceObjectId, info);
-            while (null != info ) {
-                queueSize.decrementAndGet();
-                chunksToDelete.add(info);
 
-                count++;
-                if (count >= maxItems) {
-                    break;
-                }
-                // Do not block
-                info = garbageChunks.poll();
-                log.trace("{}: deleteGarbage - retrieved {}", traceObjectId, info);
+        // Wait until you have at least one item or timeout expires.
+        GarbageChunkInfo info = Exceptions.handleInterruptedCall(() -> garbageChunks.poll(config.getGarbageCollectionDelay().toMillis(), TimeUnit.MILLISECONDS));
+        log.trace("{}: deleteGarbage - retrieved {}", traceObjectId, info);
+        while (null != info ) {
+            queueSize.decrementAndGet();
+            chunksToDelete.add(info);
+
+            count++;
+            if (count >= maxItems) {
+                break;
             }
-        } catch (InterruptedException e) {
-            throw new CompletionException(e);
+            // Do not block
+            info = garbageChunks.poll();
+            log.trace("{}: deleteGarbage - retrieved {}", traceObjectId, info);
         }
 
         // Sleep if no chunks to delete.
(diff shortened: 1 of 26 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -249,8 +249,8 @@ void addToGarbage(String chunkToDelete, long startTime, int attempts) {
         val chunksToDelete = new ArrayList<GarbageChunkInfo>();
         int count = 0;
         try {
-            // Block until you have at least one item.
-            GarbageChunkInfo info = garbageChunks.take();
+            // Wait until you have at least one item or timeout expires.
+            GarbageChunkInfo info = garbageChunks.poll(config.getGarbageCollectionDelay().toMillis(), TimeUnit.MILLISECONDS);
```
andreipaduroiu, line 253: I don't suppose this will throw a TimeoutException, will it? I don't see how, as the code compiles and you don't rethrow it or otherwise catch it.

andreipaduroiu: And line 264 below is intentional, yes? We do not pass a timeout because we just want to know if something is immediately available or not. Is my understanding correct?

andreipaduroiu: (sorry but I cannot seem to be able to comment on line 267.) 

That is the wrong way to handle `InterruptedException` in Java. There are special semantics with that exception in that, if you catch it, the thread will become "un-interrupted", which may defeat the purpose of interrupting it. 

I strongly advise you use `Exceptions.handleInterruptible` for whatever call it is that may throw this. I suppose now it's the `poll(timeout)`.

andreipaduroiu (commented):
Looks good. Just a few questions.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: Double checked the documentation, it does not throw TimeoutException. 
It returns null.
> the head of this queue, or null if the specified waiting time elapses before an element with an expired delay becomes available

sachin-j-joshi: yes line 264 is intentional. We do not want to wait for additional elements to be available.

sachin-j-joshi: what happens if I don't catch InterruptedException ?

sachin-j-joshi: done

andreipaduroiu (approved):

PR #5804, opened 2021-03-05
Issue 5789: (SegmentStore) Improving stability during Segment Container Recoveries

### LINKED ISSUE #5789: Segment Container Recovery Improvements
Segment Container recovery may fail due to cache full exception under various conditions. The full range of scenarios under which this could happen is diverse, however most of them can be auto-resolved by temporarily easing off the load on the cluster. There is a subset, however, which is harder to recover from, and that would require some more work on the segment store to prevent getting into that situation.

1. Add new throttling source. Limit maximum number of operations in Tier 1 to some preconfigured value (i.e., 1M).
2. During recovery, trim cache after processing any `StorageMetadataCheckpointOperation` or `MetadataCheckpointOperation`. This includes removing any ReadIndex for segments that are `Deleted` or (`SealedInStorage` and ! `Merged`) and evicting any cache entries that are fully written to LTS.
3. Update `MetadataCheckpointOperation` and `StorageMetadataCheckpointOperation` to have highest priority - bypass throttling.
4. Consider increasing frequency of `StorageMetadataCheckpointOperaiton.`

andreipaduroiu:
**Change log description**  
- Added a new throttler (`OperationLog`) that limits the maximum number of outstanding operations in the Tier 1 Operation Log to 1 million. The throttler will kick in gradually at 95% of that value and apply full backpressure when it hits 1 million. This is to prevent lengthy recoveries.
- Cleaning up inactive or deleted Segments during recovery - we may end up accumulating much more than we need since we may be loading up multiple checkpoints. This helps situations when we could exceed the max allowed segment count even though a previous instance of that container was never close to that limit.
- Cleaning up the Read Index cache during recovery after processing `MetadataCheckpointOperations` or `StorageCheckpointOperations`. Only essential data is preserved in the cache. This helps situations we could exceed the cache limit for a container/process even though a previous instance of that container was never close to that limit.

**Purpose of the change**  
Fixes #5789.

**What the code does**  
- Operation Log Size Throttler
    - Using existing throttler mechanism to define a new throttler type which feeds in from the `OperationLog::size` when it needs to run. 
 - Cleanup during recovery
     - Previously a `MetadataCheckpointOperation` would only be processed (deserialized) when it was the first operation in the log during recovery. All subsequent ones would be ignored since they were redundant.
     - However these checkpoints are not totally redundant. They provide information about which segments had been evicted since the last checkpoint and also snapshot the Storage (LTS) state of each segment at the time of the checkpoint.
     - Subsequent checkpoints are now incrementally applied. 
         - If an existing segment is no longer present in this checkpoint, then it is unregistered (disconsidered) from the recovery. That means it had previously been evicted and is thus of no further interest to us.
         - A segment's Storage state (length, sealed, deletion status) is now applied from the checkpoint.
         - No other information is applied.
     -  The `RecoveryProcessor` will trigger a `ReadIndex` cleanup after it processes any kind of checkpoint operation (Metadata or Storage). This cleanup involves completely unregistering indices that are pointing to inactive segments (previously evicted) and trimming the cache to only keep those entries that had not previously been fully flushed to Storage.

**How to verify it**  
All tests (old and new) must pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
@@ -422,7 +422,7 @@ class ContainerMetadataUpdateTransaction implements ContainerMetadata {
                 // and rebuild the metadata from the information we have so far.
                 if (this.processedCheckpoint) {
                     // A MetadataCheckpoint should be processed fully only if is the first checkpoint encountered.
-                    // Any subsequent MetadataCheckpoints should be partially processed and applied
+                    // Any subsequent MetadataCheckpoints should be partially processed and applied.
                     // But we can (should) only process at most one MetadataCheckpoint per recovery. Any additional
                     // ones are redundant (used just for Truncation purposes) and contain the same information as
                     // if we processed every operation in order, up to them.
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
@@ -399,12 +399,12 @@ class ContainerMetadataUpdateTransaction implements ContainerMetadata {
                     // But we can (should) only process at most one MetadataCheckpoint per recovery. Any additional
                     // ones are redundant (used just for Truncation purposes) and contain the same information as
                     // if we processed every operation in order, up to them.
-                    log.debug("MetadataUpdate[{}-{}}]: Skipping MetadataCheckpointOperation with SequenceNumber {} because we already have metadata changes.",
+                    log.debug("MetadataUpdate[{}-{}}]: Skipping MetadataCheckpointOperation({}) because we already have metadata changes.",
                             this.containerId, this.transactionId, operation.getSequenceNumber());
                     return;
                 }
 
-                log.info("MetadataUpdate[{}-{}}]: Recovering MetadataCheckpointOperation with SequenceNumber {}.",
+                log.info("MetadataUpdate[{}-{}}]: Recovering MetadataCheckpointOperation({}).",
                         this.containerId, this.transactionId, operation.getSequenceNumber());
                 clear();
 
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
@@ -18,6 +18,8 @@ import io.pravega.common.io.serialization.VersionedSerializer;
 import io.pravega.common.util.ImmutableDate;
 import io.pravega.segmentstore.contracts.Attributes;
 import io.pravega.segmentstore.contracts.ContainerException;
+import io.pravega.segmentstore.contracts.SegmentProperties;
+import io.pravega.segmentstore.contracts.SegmentType;
 import io.pravega.segmentstore.contracts.StreamSegmentException;
 import io.pravega.segmentstore.contracts.StreamSegmentNotExistsException;
 import io.pravega.segmentstore.contracts.TooManyActiveSegmentsException;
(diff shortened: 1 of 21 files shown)


andreipaduroiu:
System tests passed on this change.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
```diff
@@ -396,22 +421,25 @@ private void processMetadataOperation(MetadataCheckpointOperation operation) thr
                 // metadata is serialized in this operation. We need to discard whatever we have accumulated so far
                 // and rebuild the metadata from the information we have so far.
                 if (this.processedCheckpoint) {
+                    // A MetadataCheckpoint should be processed fully only if is the first checkpoint encountered.
+                    // Any subsequent MetadataCheckpoints should be partially processed and applied
```
RaulGracia, line 425: Nit: missing final dot.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/DurableLog.java
```diff
@@ -329,9 +329,9 @@ public boolean isOffline() {
         log.debug("{}: Queuing MetadataCheckpointOperation.", this.traceObjectId);
         MetadataCheckpointOperation op = new MetadataCheckpointOperation();
         return this.operationProcessor
-                .process(op, OperationPriority.Normal)
+                .process(op, OperationPriority.SystemCritical)
                 .thenApply(v -> {
-                    log.info("{}: MetadataCheckpointOperation durably stored.", this.traceObjectId);
+                    log.info("{}: MetadataCheckpointOperation({}} stored.", this.traceObjectId, op.getSequenceNumber());
```
RaulGracia, line 334: Something seems wrong with the log parameter: instead of `({}}` you may want to use `({})`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ThrottlerCalculator.java
```diff
@@ -65,6 +65,16 @@
      */
     @VisibleForTesting
     static final double DURABLE_DATALOG_THROTTLE_THRESHOLD_FRACTION = 0.1;
+    /**
+     * Maximum size (in number of operations) of the OperationLog, above which maximum throttling will be applied.
+     */
+    @VisibleForTesting
+    static final int OPERATION_LOG_MAX_SIZE = 1_000_000;
```
RaulGracia, line 72: It would be good to experiment whether this throttler comes into play at the right time (e.g., not too early for very small appends or too late for large ones). Actually, as the objective of this is not to be left without cache space when a recovery occurs, it would be better to throttle based on the number of bytes of the operation log in relation to the cache size configured (I don't know if this is possible).

andreipaduroiu: Well, there's always a balance between one thing and another. The objective of this throttler is to prevent the Tier 1 log from getting too big, which would cause unnecessarily long delays during recovery. 1 Million operations to recover is reasonably large buffer. The recovery speed is a function of both the read throughput from BK and the number of operations processed. Each operation has to be deserialized, written to the cache, its effects applied to metadata, etc.

If we do some quick math:
- If all we had were 100 byte appends, 1M x 100 byte is about 100MB. That should have been no problem for the StorageWriter to dump to LTS. If it had trouble doing that prior to the recovery, then we likely had another issue which had to be resolved.
    - For a 10GB cache with 4 containers on a pod, that is about 2.5GB cache per container, so this means a backlog of 25 million operations to recover. We would prefer not to get into this situation to begin with.
- For 10KB appends (more likely, due to client-side batching), 1M x 10KB is about 10GB. At this point we should get parity with whatever cache size we set (assuming only one container on the box - if multiple containers, then parity would happen much earlier).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/ContainerMetadataUpdateTransaction.java
andreipaduroiu: fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/DurableLog.java
andreipaduroiu: fixed.

RaulGracia (approved):

Issue #5697, opened 2021-02-09
Update Copyright section to include "Pravega Authors" 


medvedevigorek:
**Problem description**
The copyright section of source files should include "Pravega Authors" as the project is moving towards CNCF umbrella 

**Problem location**
HEADER and all source files 

**Suggestions for an improvement**
Either update just a HEADER file or HEADER and all source files to reflect the new change.

fpj:
We need to at least change the `HEADER`  file so that new files from now own add the correct copyright statement. 

@derekm do we need to update the license header in all files or is it only from the point we update the `HEADER` onward?

We need to also change the checkstyle configuration to make sure that new files have the correct statement. (I'm actually not entirely sure whether we can check only new files.)

derekm:
@fpj -- I think the choice is ours whether to update all files at once or not. Can we say we don't maintain the in-place headers, we just add the current header to new files?

derekm:
Fixed by #5807 and #5684

PR #5807, opened 2021-03-05
Issue 5806: Update HEADER

### LINKED ISSUE #5806: Update HEADER file
**Describe the bug**
The license HEADER needs to be updated to reflect the copyright change.

**Additional context**
With the move to CNCF, we need to update the HEADER file.

fpj:
**Change log description**  
* Updates HEADER file to reflect copyright changes due to the move to CNCF.
* Updates checkstyle to validate the correct copyright.

**Purpose of the change**  
Fixes #5806 

**What the code does**  
There is no code change.

**How to verify it**  
New pull requests with the old or a different copyright statement in the license header should fail on checkstyle.


HEADER
```diff
@@ -1,7 +1,14 @@
-Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright 2021 Pravega Authors.
```
derekm, line 1: As far as I know, we are still keeping year relegated to NOTICE file. NOTICE file was updated by Igor previously.

derekm (changes requested):

derekm:
We also have Issue #5697

HEADER
fpj: That's correct, I'll remove it.

derekm (approved):

checkstyle/checkstyle.xml
```diff
@@ -9,7 +9,7 @@
 
   <!-- header -->
   <module name="RegexpHeader">
-    <property name="header" value="/\*\*\n * Copyright \(c\) Dell Inc\.\, or its subsidiaries\. All Rights Reserved\."/>
+    <property name="header" value="/\*\*\n * Copyright \(c\) Pravega Authors. All Rights Reserved\."/>
```
derekm, line 12: Updating the rule with this regex requires we proactively update all file headers throughout the repo. Do we do this sweep change in this PR? Or do we allow the regex to accept both headers?

derekm (changes requested):
We either need to update Checkstyle `RegexpHeader` to accept both headers, or we need to make a sweeping change to update all file headers in the repo.

HEADER
```diff
@@ -1,7 +1,14 @@
-Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright 2021 The Pravega Authors.
```
derekm, line 1: remove year

NOTICE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2021 Pravega Authors.
+Copyright (c) 2021 The Pravega Authors.
```
derekm, line 1: remove "The "

derekm (changes requested):
stray year and stray "The"

derekm (changes requested):
Also -- As in the NOTICE file, we are advised by CNCF policy to not remove old copyright statements, but only add to them. In other words, existing in-place headers should be:
```
/**
 * Copyright Pravega Authors.
 * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
```
or the like.

derekm (approved):
Expressed permission to change Dell copyright statements given by Sean Evans of Legal:
>"Confirming both that we can change the Dell copyright statements and we cannot change the copyright statements for non-Dell authors."

PR #5811, opened 2021-03-05
Issue 5810: (SegmentStore) Fixed a StorageWriter bug that could lead to data loss

### LINKED ISSUE #5810: StorageWriter may lose in-memory operations if unable to initialize a SegmentAggregator.
**Describe the bug**
Each StorageWriter iteration works along these lines
1. Fetch a sequence of operations from `DurableLog`
2. Process each operation in sequence.
3. When processing them, determine which Segment they belong to and initialize an appropriate `SegmentAggregator` for that segment if necessary.
4. Flush anything if needed
5. Ack.

The problem is that if, in step 3, the `SegmentAggregator.initialize` errors out (it needs to do a `Storage.getStreamSegmentInfo` on LTS), then the entire iteration is aborted and all non-processed operations are lost. They have already been fetched out of the `DurableLog` and now there's nowhere to get them from again. 

Fortunately there are plenty of safeguards and sanity checks in the `StorageWriter` and `SegmentAggregator` classes to detect situations like these before any actual data loss or corruption happens (the data itself will still be in Tier 1 since it cannot be truncated out).

**To Reproduce**
See above.

**Expected behavior**
StorageWriter should be resilient in the face of `SegmentAggregator.initialize` errors and not "lose" in-flight operations.

andreipaduroiu:
**Change log description**  
Fixed a bug in StorageWriter where it could lose already read Operations if an operation that had to be processed required the initialization of SegmentAggregator, but that initialization ended up throwing an error (i.e., from getting segment info from Storage).

**Purpose of the change**  
Fixes #5810.

**What the code does**  
See #5810 for context on this.

The fix for this was to store the last read sequence of operations inside the `WriterState` object (as a reference) and clean it up upon a successful completion of `processReadResult`. If any sort of exception occurred during `processReadResult`, the operation on which it failed would still remain the first one in that read queue (we now `peek` it vs `poll`-ing it), and it will be attempted next time. For example, if we read 10 operations, process 5, and the 6th one fails, then the next writer iteration will resume from 6.

To support this, a few more changes have been made:
- the `StorageWriter.readData` will check the `WriterState` for any previously read operations. If any are still outstanding, it will return those vs. attempting to read new ones.
- `processReadResult` peeks the operation, processes it, then removes it from the queue
- Any exception coming from `SegmentAggregator.add` is now interpreted as `DataCorruptionException`. Even if the exception itself does not imply a data corruption, chances are high that it will be thrown again upon a retry, so we might as well stop everything and force a recovery. 
    - Furthermore, the real reason this was changed was because an operation may have to be sent to multiple `WriterProcessor` instances (i.e., `SegmentAggregator`, `AttributeAggregator`, `WriterTableProcessor`). If it has already been sent to one of them but the next one failed (for whatever reason), resending it to the first one will cause another exception (due to the multitude of sanity checks we perform). So, for simplicity, we stop the writer and start fresh.

**How to verify it**  
New unit tests added to verify scenario.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
@@ -187,10 +187,22 @@ class StorageWriter extends AbstractThreadPoolService implements Writer {
      * @return A CompletableFuture that, when complete, will indicate that the read has been performed in its entirety.
      */
     private CompletableFuture<Queue<Operation>> readData(Void ignored) {
+        val lastRead = this.state.getLastRead();
+        if (lastRead != null && !lastRead.isEmpty()) {
+            // This happens if, during a previous iteration, we had an unexpected error preventing processReadResult from
+            // execution to completion (i.e., SegmentAggregator.initialize() failed). In that case, it is imperative that
+            // we do not miss out on operations, otherwise that can translate to either a blocking error (no progress can
+            // be made) or a data loss.
+            log.info("{}: Iteration[{}] Not performing a new read because there are still {} items unprocessed from the previous iteration.",
+                    this.traceObjectId, this.state.getIterationId(), lastRead.size());
+            return CompletableFuture.completedFuture(lastRead);
+        }
+
         try {
             Duration readTimeout = getReadTimeout();
             return this.dataSource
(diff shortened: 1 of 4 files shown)


RaulGracia (approved):
LGTM.

sachin-j-joshi (approved):
LGTM

PR #5813, opened 2021-03-07
Issue 5812: Convenience APIs for working with SegmentRanges and StreamCuts

### LINKED ISSUE #5812: Convenience APIs for working with SegmentRanges and StreamCuts

The purpose of the changes would be a couple of APIs in order to help creating SegmentRanges.  The main api needed is:
- given a Segment, start stream cut, and end stream cut, return a SegmentRangeImpl for the given Segment start->end offset.

Also if a segment did not exist at head or tail of a stream (e.g. there was scale up/down), there did not seem to be an easy way to get beginning/end stream cut info for them.  So additional apis are:
- get first stream cut for a Segment
- get last stream cut for a Segment

Alternatives have involved doing a lot of extra work listing segments in order to produce similar results.

**Additional context**
My use case is consuming a pravega stream in parallel (in context of presto db).
The number of workers available for this processing can vary, but in a typical scenario there will be a few hundred (400 in my test system).
Working on a flexible system where each worker can process part of a segment of a stream.  As such would like to be able to create a SegmentRange of configurable desired byte size that can be handed off to a worker.  For example so far in our testing, approx ~32MB / worker seems to work well.  But this will need to change depending on workers available, overall size of stream, etc.

At a high level, to do my application uses a streams associated Watermark stream.
The application will read watermarks from the Watermark stream and will keep track of the distance in bytes between 2 stream cuts in a Segment.  When the distance is > configurable size a SegmentRangeImpl will be created for it (SegmentRange cannot be exactly the desired size, but will be close).

The suggested APIs help in this translation from Watermarks to a usable SegmentRange that can be processed.

I will link a draft PR of the changes to pravega, and would welcome any comments, suggestions, etc

Thanks a lot!

adrdc:
Convenience APIs for working with SegmentRanges and StreamCuts

Signed-off-by: Andrew Robertson <andrew.robertson@dell.com>

**Change log description**  
Introduces following APIs:
- given Segment, start + end StreamCuts, return SegmentRange for the segment start->end offset
- fetch first stream cut for a given segment
- fetch last stream cut for a given segment

**Purpose of the change**  
Fixes #5812 

**What the code does**  
Implements said APIs

**How to verify it**  

### DIFF
--- a/client/src/main/java/io/pravega/client/BatchClientFactory.java
+++ b/client/src/main/java/io/pravega/client/BatchClientFactory.java
@@ -16,6 +16,7 @@ import io.pravega.client.batch.StreamSegmentsIterator;
 import io.pravega.client.batch.impl.BatchClientFactoryImpl;
 import io.pravega.client.connection.impl.SocketConnectionFactoryImpl;
 import io.pravega.client.segment.impl.NoSuchSegmentException;
+import io.pravega.client.segment.impl.Segment;
 import io.pravega.client.stream.EventStreamReader;
 import io.pravega.client.stream.Serializer;
 import io.pravega.client.stream.Stream;
@@ -79,6 +80,22 @@ public interface BatchClientFactory extends AutoCloseable {
      * @return A SegmentIterator over the requested segment
      */
     <T> SegmentIterator<T> readSegment(SegmentRange segment, Serializer<T> deserializer);
+
+    /**
+     * Return first offset for a given {@link Segment}
+     *
+     * @param segment The {@link Segment}
+     * @return {@link StreamCut} pointing to the starting offset of the {@link Segment}.
+     */
+    StreamCut firstStreamCut(Segment segment);
+
+    /**
+     * Return last offset for a given {@link Segment}
+     *
+     * @param segment The {@link Segment}
+     * @return {@link StreamCut} pointing to the last offset of the {@link Segment}.
+     */
+    StreamCut lastStreamCut(Segment segment);
 
     /**
      * Closes the client factory. This will close any connections created through it.
--- a/client/src/main/java/io/pravega/client/BatchClientFactory.java
+++ b/client/src/main/java/io/pravega/client/BatchClientFactory.java
@@ -82,7 +82,7 @@ public interface BatchClientFactory extends AutoCloseable {
     <T> SegmentIterator<T> readSegment(SegmentRange segment, Serializer<T> deserializer);
 
     /**
-     * Return first offset for a given {@link Segment}
+     * Return first offset for a given {@link Segment}.
      *
      * @param segment The {@link Segment}
      * @return {@link StreamCut} pointing to the starting offset of the {@link Segment}.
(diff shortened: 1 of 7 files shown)


client/src/main/java/io/pravega/client/BatchClientFactory.java
```diff
@@ -80,6 +81,22 @@ static BatchClientFactory withScope(String scope, ClientConfig config) {
...
+    StreamCut firstStreamCut(Segment segment);
+
+    /**
+     * Return last offset for a given {@link Segment}.
+     *
+     * @param segment The {@link Segment}
+     * @return {@link StreamCut} pointing to the last offset of the {@link Segment}.
+     */
+    StreamCut lastStreamCut(Segment segment);
+
```
tkaitchuck, line 99: These are returning a StreamCut object. But because the data within the StreamCut only pertains to one segment, it isn't a well formed StreamCut.
I think what you want to do is have a method:
```
SegmentRange currentSegmentRange(Segment segment);
```
The implementation of which would grab the lowest and highest offsets from the segment is the implementation of the two methods here do, and return it as a SegmentRange.

tkaitchuck (changes requested):

client/src/main/java/io/pravega/client/BatchClientFactory.java
adrdc: thanks a lot!  so, i made the changes.  i still need to verify the changes on my app side (changing repos, build systems, so a bit delayed on that there), and i'll change from draft after that

client/src/main/java/io/pravega/client/batch/impl/SegmentRangeImpl.java
```diff
@@ -75,4 +77,20 @@ public SegmentRangeImpl build() {
...
+     * @return {@link SegmentRange} covering start-end
+     */
+    public static SegmentRange fromStreamCuts(final Segment segment, final StreamCut start, final StreamCut end) {
+        Preconditions.checkState(!start.equals(StreamCut.UNBOUNDED), "Start StreamCut may not be UNBOUNDED");
+        Preconditions.checkState(!end.equals(StreamCut.UNBOUNDED), "End StreamCut may not be UNBOUNDED");
+        long startOffset = start.asImpl().getPositions().getOrDefault(segment, -1L);
+        long endOffset = end.asImpl().getPositions().getOrDefault(segment, -1L);
+        Preconditions.checkState(startOffset >= 0 && endOffset > startOffset, "Start offset should be less than end offset.");
+        return new SegmentRangeImpl(segment, startOffset, endOffset);
+    }
```
tkaitchuck, line 95: I think this is no longer needed

client/src/main/java/io/pravega/client/stream/impl/StreamCutImpl.java
```diff
@@ -63,9 +63,13 @@ public StreamCutImpl(Stream stream, Map<Segment, Long> positions) {
         this.positions = ImmutableMap.copyOf(positions);
     }
 
+    public StreamCutImpl(Stream stream, Segment segment, long position) {
+        this(stream, Collections.singletonMap(segment, position));
+    }
+
```
tkaitchuck, line 69: Also no longer needed.

andreipaduroiu:
@adrdc Are we still pursuing this? If I don't hear back by Sept 30, I will close this PR (and we can reopen it if still needed).

PR #5827, opened 2021-03-10
Issue 5802: Controller fails with OOM on aggressive scaling and transactions.

### LINKED ISSUE #5802: Aggressive scaling and rolling transactions cause Controller to go OOM
**Describe the bug**
Controller fails with OOM on aggressive scaling and rolling transactions.

**To Reproduce**
Start a Pravega cluster and run 2 longevity medium scale workloads one that triggers scaling and the other with rolling transactions both at the same time on this cluster.

**Expected behavior**
Controller should not go OOM

pbelgundi:
**Change log description**  
Controller fails with OOM on aggressive scaling and rolling transactions.
This is because on every scale/rolling-txn operation we compute the number of _splits_ and _merges_ on the set of all epochs for this Stream. This is done so we can emit a metric on the number of splits & merges on the Stream at that point.
This computation is an expensive operation and having a large number of scales and rolling transactions causes us to go OOM. 

**Purpose of the change**  
Fixes #5802

**What the code does**  
Instead of computing the total number of splits/merges on **every scale** we could pre-compute and store the cumulative split/merge count till this Epoch inside the `Epoch Record` when it is created and then just fetch the number from this record.

**How to verify it**  


controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
```diff
@@ -37,25 +38,52 @@
...
     private final int referenceEpoch;
     private final ImmutableList<StreamSegmentRecord> segments;
     private final long creationTime;
     @Getter(AccessLevel.PRIVATE)
     private final Map<Long, StreamSegmentRecord> segmentMap;
-    
+    /**
+     * Number of splits performed as part of scale operation.
+     */
+    private final long splits;
```
shiveshr, line 53: is it number of splits in this scale operation or overall so far?

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
```diff
@@ -37,25 +38,52 @@
...
+     *
+     * Principle to calculate the number of splits and merges:
+     * 1- An event has occurred if a reference range is present (overlaps) in at least two consecutive target ranges.
+     * 2- If the direction of the check in 1 is forward, then it is a split, otherwise it is a merge.
+     *
+     * @param referenceSegmentsList Reference segment list.
+     * @param targetSegmentsList Target segment list.
+     * @return Number of splits/merges.
+     */
+    public static long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegmentsList, List<StreamSegmentRecord> targetSegmentsList) {
```
shiveshr, line 82: this method doesnt seem fit here as neither of the input params are EpochRecord. we should keep this in its original place.

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
```diff
@@ -37,25 +38,52 @@
...
+        return referenceSegmentsList.stream().filter(
+                segment -> targetSegmentsList.stream().filter(target -> target.overlaps(segment)).count() > 1 ).count();
     }
-    
+
```
shiveshr, line 86: nit: please revert this and all other lines where only indentation is changed

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: It is overall so far, since the Epoch Record will maintain only cumulative counts.

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: Ok

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: Its a static method, so input params don't need to be members of EpochRecord class.
Its a utility method to find the splits/merges given the segments of earlier and this epoch, I think it fits better here.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -533,33 +533,17 @@ private boolean streamCutEqualOrAfter(Map<Long, Long> cut1, Map<StreamSegmentRec
...
-     * 2- If the direction of the check in 1 is forward, then it is a split, otherwise it is a merge.
-     *
-     * @param referenceSegmentsList Reference segment list.
-     * @param targetSegmentsList Target segment list.
-     * @return Number of splits/merges.
-     */
-    private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegmentsList, List<StreamSegmentRecord> targetSegmentsList) {
-        return referenceSegmentsList.stream().filter(
-                segment -> targetSegmentsList.stream().filter(target -> target.overlaps(segment)).count() > 1 ).count();
+                .map(record -> {
```
shiveshr, line 536: i dont see any changes here other than indentation, please revert this.

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
shiveshr: please make that explicit in the comment above that its cumulative count

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
shiveshr: then it should be in a utility class and not in an unrelated data class (epoch record).
there is a utility class called record helper that can be used for this. 

secondly, moving this method out of persistent stream base should be done if we see this method being used in multiple places.. otherwise it can be a private method there the way it is.

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: Sure.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
- *
+ * <p>
```
shiveshr: nit: please revert changes to the license.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -130,12 +132,12 @@ public String getScopeName() {
 
         long creationTime = createStreamResponse.getTimestamp();
         final ImmutableList.Builder<StreamSegmentRecord> builder = ImmutableList.builder();
-        
+
         IntStream.range(0, numSegments).boxed()
-                 .forEach(x -> builder.add(newSegmentRecord(0, startingSegmentNumber + x, creationTime,
-                                                                    x * keyRangeChunk, (x + 1) * keyRangeChunk)));
+                .forEach(x -> builder.add(newSegmentRecord(0, startingSegmentNumber + x, creationTime,
```
shiveshr: nit: revert

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -173,19 +175,19 @@ public String getScopeName() {
...
-                            StreamTruncationRecord previous = existing.getObject();
-                            // check greater than
-                            Exceptions.checkArgument(streamCutEqualOrAfter(streamCut, span, previous.getStreamCut(), previous.getSpan()),
-                                    "StreamCut", "Supplied streamcut is behind previous truncation point");
-
-                            return computeTruncationRecord(previous, streamCut, span)
-                                    .thenCompose(prop ->
-                                            Futures.toVoid(setTruncationData(new VersionedMetadata<>(prop, existing.getVersion()))));
-                        });
+                            .thenCompose(epochs -> {
```
shiveshr: nit: revert

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -199,8 +201,8 @@ private boolean streamCutEqualOrAfter(Map<Long, Long> cut1, Map<StreamSegmentRec
                                 cut1.get(e1.getKey().segmentId()) < cut2.get(e2.getKey().segmentId()))
                                 || (e2.getKey().overlaps(e1.getKey()) && e1.getValue() < e2.getValue())));
     }
-    
-    private CompletableFuture<StreamTruncationRecord> computeTruncationRecord(StreamTruncationRecord previous, 
+
```
shiveshr, line 202: nit: revert

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -212,19 +214,19 @@ private boolean streamCutEqualOrAfter(Map<Long, Long> cut1, Map<StreamSegmentRec
...
-                                             ImmutableSet.Builder<Long> builder = ImmutableSet.builder();
-                                             segmentsBetween.stream().map(StreamSegmentRecord::segmentId)
-                                                            .filter(x -> !streamCut.containsKey(x))
-                                                            .forEach(builder::add);
-
-                                             return new StreamTruncationRecord(ImmutableMap.copyOf(streamCut), span, previous.getDeletedSegments(), 
-                                                     builder.build(),
-                                                     previous.getSizeTill() + sizeBetween, true);
-                                         }));
+                .thenCompose(segmentsBetween -> sizeBetweenStreamCuts(previous.getStreamCut(), streamCut, segmentsBetween)
```
shiveshr, line 215: and here

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -259,53 +261,53 @@ private boolean streamCutEqualOrAfter(Map<Long, Long> cut1, Map<StreamSegmentRec
...
-                                } else if (e2.getKey().overlaps(e1.getKey())) {
-                                    // overlapping segment. compare segment id.
-                                    comparison = Long.compare(e1.getKey().segmentId(), e2.getKey().segmentId());
-                                } else {
-                                    continue;
-                                }
-                                foundGt = !foundGt ? comparison > 0 : foundGt;
-                                foundLt = !foundLt ? comparison < 0 : foundLt;
+                .thenApply(v -> {
+                    ImmutableMap<StreamSegmentRecord, Integer> span1 = span1Future.join();
```
shiveshr, line 271: revert please

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
```diff
@@ -70,23 +92,22 @@ public static EpochRecord fromBytes(final byte[] record) {
...
     public Set<Long> getSegmentIds() {
         return segmentMap.keySet();
     }
 
     public StreamSegmentRecord getSegment(long segmentId) {
         return segmentMap.get(segmentId);
     }
-    
+
```
shiveshr, line 103: nit: revert

controller/src/main/java/io/pravega/controller/store/stream/ScaleMetadata.java
```diff
@@ -10,7 +10,6 @@
 package io.pravega.controller.store.stream;
 
 import lombok.Data;
-
```
shiveshr, line 13: revert

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
```diff
@@ -704,4 +704,6 @@
      * is strictly before the supplied streamcut. 
      */
     CompletableFuture<StreamCutReferenceRecord> findStreamCutReferenceRecordBefore(Map<Long, Long> streamCut, RetentionSet retentionSet);
+
+    CompletableFuture<SimpleEntry<Long, Long>> getSplitMergeCountsTillEpoch(EpochRecord epochRecord);
```
shiveshr, line 708: please add javadoc

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
```diff
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
- *
+ * <p>
```
shiveshr, line 3: revert

controller/src/test/java/io/pravega/controller/server/eventProcessor/ScaleRequestHandlerTest.java
```diff
@@ -861,7 +861,7 @@ private void concurrentDistinctScaleRun(String stream, String funcToWaitOn, bool
         streamStore2.close();
     }
 
-    @Test(timeout = 30000)
+    @Test(timeout = 50000)
```
shiveshr, line 864: why are we increasing this timeout?

shiveshr (changes requested):
some nitpicks for indentation changes

controller/src/main/java/io/pravega/controller/store/stream/Stream.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -1036,16 +1035,29 @@ protected ReaderGroup getReaderGroup(String scope, final String name, RGOperatio
...
-            AtomicLong totalNumMerges = new AtomicLong(0L);
-            scaleMetadataList.forEach(x -> {
-                totalNumMerges.addAndGet(x.getMerges());
-                totalNumSplits.addAndGet(x.getSplits());
-            });
-
-            return new SimpleEntry<>(totalNumSplits.get(), totalNumMerges.get());
+        Stream stream = getStream(scopeName, streamName, context);
+        return stream.getActiveEpoch(true).thenCompose(lastEpoch -> stream.getSplitMergeCountsTillEpoch(lastEpoch));
+        /*
```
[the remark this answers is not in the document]
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -533,33 +533,17 @@ private boolean streamCutEqualOrAfter(Map<Long, Long> cut1, Map<StreamSegmentRec
...
-     *
-     * Principle to calculate the number of splits and merges:
-     * 1- An event has occurred if a reference range is present (overlaps) in at least two consecutive target ranges.
-     * 2- If the direction of the check in 1 is forward, then it is a split, otherwise it is a merge.
-     *
-     * @param referenceSegmentsList Reference segment list.
-     * @param targetSegmentsList Target segment list.
-     * @return Number of splits/merges.
-     */
-    private long findSegmentSplitsMerges(List<StreamSegmentRecord> referenceSegmentsList, List<StreamSegmentRecord> targetSegmentsList) {
```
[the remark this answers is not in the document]

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1100,7 +1084,7 @@ private boolean isStreamCutValidInternal(Map<Long, Long> streamCut, int epochLow
                         // sealed segments
                         ImmutableList.Builder<StreamSegmentRecord> sealedSegmentsBuilder = ImmutableList.builder();
                         epochTransition.getSegmentsToSeal().forEach(x -> sealedSegmentsBuilder.add(currentEpoch.getSegment(x)));
-                        
+
```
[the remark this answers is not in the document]

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/ScaleMetadata.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/records/EpochRecord.java
pbelgundi: fixed

controller/src/test/java/io/pravega/controller/server/eventProcessor/ScaleRequestHandlerTest.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1118,10 +1118,13 @@ private boolean isStreamCutValidInternal(Map<Long, Long> streamCut, int epochLow
                         ImmutableList<StreamSegmentRecord> newSegments = newSegmentsBuilder.build();
                         builder.addAll(newSegments);
                         // epoch record
-                        EpochRecord epochRecord = new EpochRecord(epochTransition.getNewEpoch(), epochTransition.getNewEpoch(), 
-                                builder.build(), time);
-
-                        HistoryTimeSeriesRecord timeSeriesRecord = 
+                        SimpleEntry<Long, Long> splitMergeCountLastEpoch = getSplitMergeCountsTillEpoch(currentEpoch).join();
```
shiveshr, line 1121: we should not be blocking here.

shiveshr: we should get the future and add callbacks to perform any subsequent processing.. 
we use join in the code only in following pattern:
create future1 and future2..
```
future1 = getFuture();
future2 = getFuture();
CompletableFuture.allOf(future1, future2) 
    .thenApply(v -> {// chain subsequent processing to completion of "both" futures.
           result1 = future1.join(); // this join is going to return the result immediately without blocking as the result is available with future1 which is completed for sure.
           result2 = future2.join(); // same here
           // use my results to do some computation.
     });
```

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1118,10 +1118,13 @@ private boolean isStreamCutValidInternal(Map<Long, Long> streamCut, int epochLow
...
+                        ImmutableList<StreamSegmentRecord> epochSegments = builder.build();
+                        List<StreamSegmentRecord> newEpochSegments = epochSegments.stream().collect(Collectors.toList());
```
shiveshr, line 1123: dont epoch segments capture new epoch segments? because we are building the new segment list in that builder.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1133,6 +1136,39 @@ private boolean isStreamCutValidInternal(Map<Long, Long> streamCut, int epochLow
...
                 });
     }
+
+    @Override
+    public CompletableFuture<SimpleEntry<Long, Long>> getSplitMergeCountsTillEpoch(EpochRecord epochRecord) {
+        if (epochRecord.hasSplitMergeCounts()) {
+            return CompletableFuture.completedFuture(new SimpleEntry<>(epochRecord.getSplits(), epochRecord.getMerges()));
+        } else {
+            //migration case: build the cumulative count from all previous epochs
+            return fetchEpochs(0, epochRecord.getEpoch(), true).thenApply(this::mapToScaleMetadata)
```
shiveshr, line 1146: even in this case, i was thinking we can actually do better -- we dont need to create each of the epochs. fetchEpochs actually computes epochs from history timeseries. 
what we need is the history timeseries only. and doing one pass on history timeseries will enable us to get the splits and merges count. so we dont actually have to do the fetchEpochs and then maptoscalemetadata here. which will even avoid lots of objects being created in memory. It will be O(number of segments) memory that we will use if we just travese history time series as opposed to working with epoch records.

shiveshr: we dont need to do this now.. i have created #5885 that we can work on for 0.10.0 and it will also help with streamcut related operations too.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1257,11 +1293,21 @@ private boolean isStreamCutValidInternal(Map<Long, Long> streamCut, int epochLow
                                             committingTxnRecord.getNewActiveEpoch()),
                                             timeStamp + 1, x.getKeyStart(), x.getKeyEnd())));
 
-                            EpochRecord duplicateTxnEpoch = new EpochRecord(committingTxnRecord.getNewTxnEpoch(), 
-                                    transactionEpochRecord.getReferenceEpoch(), duplicateTxnSegmentsBuilder.build(), timeStamp);
-
+                            SimpleEntry<Long, Long> splitMergeCountTillLastEpoch = getSplitMergeCountsTillEpoch(transactionEpochRecord).join();
```
shiveshr, line 1296: same comment about `join`

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: epochSegments is an ImmutableList and `findSegmentSplitsMerges` needs a List as input, hence the conversion.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: Ok. got it.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: Fixed

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: fixed as suggested.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1254,34 +1290,47 @@ private boolean isStreamCutValidInternal(Map<Long, Long> streamCut, int epochLow
...
-                                    .thenCompose(x -> updateHistoryTimeSeries(timeSeriesRecordActiveEpoch))
-                                    .thenCompose(x -> createSegmentSealedEpochRecords(
-                                            activeEpochRecord.getSegments().stream().map(StreamSegmentRecord::segmentId)
-                                                             .collect(Collectors.toList()), duplicateTxnEpoch.getEpoch()))
-                                    .thenCompose(x -> createSegmentSealedEpochRecords(
-                                            duplicateTxnEpoch.getSegments().stream().map(StreamSegmentRecord::segmentId)
-                                                             .collect(Collectors.toList()), duplicateActiveEpoch.getEpoch()));
-                        })
-                        .thenCompose(r -> updateSealedSegmentSizes(sealedTxnEpochSegments)));
+                                                    committingTxnRecord.getNewActiveEpoch()),
```
shiveshr, line 1293: i dont see any code change in this block. could you please revert this. thanks

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
```diff
@@ -96,6 +96,21 @@
         this.executor = executor;
     }
 
+    @VisibleForTesting
+    PravegaTablesStreamMetadataStore(CuratorFramework curatorClient, ScheduledExecutorService executor,
```
shiveshr, line 100: do we need this new constructor? can we reuse one of the existing constructors that takes segment helper for which we have a test mock that mocks the underlying store's behaviour.

shiveshr (approved):
LGTM, some minor comments

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
pbelgundi: Yes, needed to create this constructor, because I had to mock getCachedData() on PravegaTablesStoreHelper. I tried mocking getEntry() on SegmentHelper mock, but it required metadataTableName which is not available in the test case and without the table name, just with the 'key' for getEntry set to "epoch-0" the mock method did not return the expected value.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: fixed.

PR #5834, opened 2021-03-11
Issue #5833: Have controller client-connections perform callbacks directly.

### LINKED ISSUE #5833: Run controller client threads via client-internal threadpool (Reduce threads in client)
Today inside of `ControllerImpl` on the client there is a thread pool created internal to `gRPC` which creates threads with the ids:
`grpc-default-executor`. These threads are not needed as all they do is complete futures. Instead it should be configured to simply perform this work inline in the calling thread.

tkaitchuck:

**Change log description**  
* Have controller client-connections perform callbacks directly.
* Make SegmentMetadataClient async to avoid the need to block client internal threads.
* Remove auxiliary thread pools.

**Purpose of the change**  
Fixes #5833 

**What the code does**  
Removes several thread pools and runs everything on the client from the `clientInternal` threadpool. 
For the controller client this is done by enabling the use of the "directExecutor" controller client.
This requires some changes to SegmentMetadataClient and to unit tests which depend on this behavior.


client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -179,6 +179,7 @@ public ControllerImpl(final ControllerImplConfig config,
         this(NettyChannelBuilder.forTarget(config.getClientConfig().getControllerURI().toString())
                                 .nameResolverFactory(new ControllerResolverFactory(executor))
                                 .defaultLoadBalancingPolicy("round_robin")
+                                .directExecutor()
```
shrids, line 182: Observing that `grpc-default-worker-ELG-*` threads created even with this changeset. The number of threads is 2 * available processors .

tkaitchuck: Sorry, I wrote the wrong thing in the issue. This removes the `grpc-default-executor` threads from the client.

shrids (approved):
LGTM.
With this change 
* `grpc-default-executor-*` threads are not created.
* `RPCAsyncCallback` is now invoked on `grpc-default-worker-ELG-*` threads.

shrids (commented):
Integration tests are failing to complete with this changeset. This needs to be investigated...

shrids (changes requested):
Test **io.pravega.test.integration.StreamRecreationTest#testStreamRecreation** never completes with this change set.
(Based onthe inital thread dump we can see chanined completion stages of other futures being executed on grpc event loop threads grpc-default-worker-ELG-* )

shrids (approved):
LGTM...

shrids:
System test run `test-pravega-continuous/3685` has been triggered.

shrids (changes requested):
Fetching `StreamInfo `using `StreamManagerImpl ` is blocked during the system test run.

```
2021-03-17 13:37:16,643 8614 [Time-limited test] INFO  i.p.c.admin.impl.StreamManagerImpl - Fetching StreamInfo for scope/stream: testBatchClientScope161028058/testBatchClientStream
2021-03-17 13:37:16,644 8615 [StreamManager-Controller-1] INFO  i.p.c.control.impl.ControllerImpl - [requestId=-7669086914625901044] Tagging client request (getCurrentSegments-testBatchClientScope161028058-testBatchClientStream).
// post this the system test times out.
```

client/src/test/java/io/pravega/client/connection/impl/RawClientTest.java
```diff
@@ -151,7 +150,8 @@ public void testExceptionHandling() throws ConnectionFailedException {
         assertTrue(future.isCompletedExceptionally());
         assertFutureThrows("The future should be completed exceptionally", future,
                 t -> t instanceof ConnectionFailedException);
-
+        rawClient.close();
```
shrids, line 153: Do we need to create a new `RawClient`?

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -230,12 +230,12 @@ private void failConnection(Throwable throwable) {
             } else if (failSetupConnection) {
                 setupConnection.releaseExceptionallyAndReset(throwable);
             }
-            if (oldConnectionSetupCompleted != null) {
-                oldConnectionSetupCompleted.completeExceptionally(throwable);
-            }
             if (oldConnection != null) {
                 oldConnection.close();
             }
+            if (oldConnectionSetupCompleted != null) {
```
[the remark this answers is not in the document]
tkaitchuck: Yes. While it has no effect in production this new order makes the test deterministic. Before there was a race between `close` being called call and the futures triggered from the completion. This forces the close to always happen first.

client/src/test/java/io/pravega/client/connection/impl/RawClientTest.java
tkaitchuck: Yes. In real use the raw client can't be used once the exception is thrown. The mock was previously allowing that and this test was taking advantage of it. Now it mimics how it would actually get used.

shrids (approved):
LGTM

shrids:
System tests with this changeset is triggered via `pravega-branch-builder/1088/`
(In-progress)

shrids (changes requested):
System test failed with this changeset. Investigation in progress. Blocking the PR until this is resolved.

shrids (approved):
`test-pravega-continuous/3750/` run was successful with this changeset.

PR #5857, opened 2021-03-17
Issue 5856: Sporadic failure in Sporadic failure of BookieFailoverTest

### LINKED ISSUE #5856: Sporadic failure in Sporadic failure of BookieFailoverTest
**Describe the bug**
In our system tests, we sporadically see `BookieFailoverTest` failing due to:
```
2021-03-08 23:52:44,232 289495 [main] ERROR i.p.t.s.f.TestFrameworkException - TestFramework Exception. Type: InternalError, Details: Exception while running test method: bookieFailoverTest
java.lang.AssertionError: Unable to read events. Test failure
	at io.pravega.test.system.AbstractReadWriteTest.stopReaders(AbstractReadWriteTest.java:391)
	at io.pravega.test.system.BookieFailoverTest.bookieFailoverTest(BookieFailoverTest.java:192)
```
The root cause seems to be that sometimes the recovery of Segment Containers plus the sleeps throughout the test take longer than the retry configuration of readers.

**To Reproduce**
Run system tests via gradle.

**Expected behavior**
The test should pass consistently.

**Screenshots**
n/a

**Additional context**
n/a

RaulGracia:
**Change log description**  
Reduced the sleeps in BookkeeperFailoverTest and set the Bookkeeper client write timeout/retries to lower values.

**Purpose of the change**  
Fixes #5856.

**What the code does**  
Added some minor changes to the `BookkeeperFailoverTest` to make it to pass more reliably:
- Removed 2 out of 3 sleeps (`BOOKIE_FAILOVER_WAIT_MILLIS`) and replaced them by check for progress of readers and writers.
- Set `bookkeeper.write.timeout.milliseconds=10000` and `bookkeeper.write.attempts.count.max=3` to trigger container recoveries sooner.
- Use more Segments in the test Stream, so we make sure that all the containers have some data to recover.

Looking at the new executions, the test takes 1-3 minutes less to execute compared to `master`.

**How to verify it**  
At the moment, we had 7/7 successful runs of this test.


test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
```diff
@@ -144,40 +145,50 @@ public void tearDown() {
...
     @Test
-    public void bookieFailoverTest() throws ExecutionException, InterruptedException {
+    public void bookieFailoverTest() throws Exception {
         createWriters(clientFactory, NUM_WRITERS, SCOPE, STREAM);
         createReaders(clientFactory, readerGroupName, SCOPE, readerGroupManager, STREAM, NUM_READERS);
 
         // Give some time to create readers before forcing a bookie failover.
-        Exceptions.handleInterrupted(() -> Thread.sleep(BOOKIE_FAILOVER_WAIT_MILLIS));
+        AssertExtensions.assertEventuallyEquals("Writers and/or readers not progressing.", true,
+            () -> testState.getEventWrittenCount() > 100 && testState.getEventReadCount() > 100, 1000, IO_PROGRESS_CHECK_TIMEOUT);
```
ManishKumarKeshri, line 154: Parameter value = 100, better if have it configurable. I think so, but may not be required.

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
```diff
@@ -144,40 +145,50 @@ public void tearDown() {
...
+        log.info("Waiting until writers and readers cannot make further progress.");
+        long writeCountDiff, readCountDiff;
+        do {
+            writeCountDiff = testState.getEventWrittenCount();
+            readCountDiff = testState.getEventReadCount();
+            Exceptions.handleInterrupted(() -> Thread.sleep(2000));
+            writeCountDiff = testState.getEventWrittenCount() - writeCountDiff;
+            readCountDiff = testState.getEventReadCount() - readCountDiff;
+            log.info("Written ({}) and read ({}) events after downscale within 2 seconds.", writeCountDiff, readCountDiff);
+        } while (writeCountDiff + readCountDiff > 0);
```
ManishKumarKeshri, line 168: It will only ensure either read or write is happening, not both. Is that enough?

ManishKumarKeshri (changes requested):

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
RaulGracia: Done.

test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java
RaulGracia: No, this is going to leave the loop only if neither writes or reads are happening. Here we want to ensure that readers and writers cannot make further progress because of the downscale of Bookkeeper. Once this happens, we can assume that the system cannot execute IO and the effect of the Bookie failover is ongoing. The next assertion after the sleep checks that this is actually true for `BOOKIE_FAILOVER_WAIT_MILLIS` milliseconds.

ManishKumarKeshri (approved):

shrids (approved):

PR #5862, opened 2021-03-18
Issue 5861: (SLTS) Make metadata buffer size and metadata cache size configurable.

### LINKED ISSUE #5861: SLTS:  metadata buffer size and metadata should be configurable.
**Describe the bug**
Currently SLTS metadata buffer size and metadata cache size are hardcoded, it should be configurable.

**To Reproduce**
They should be configurable via config file.

**Expected behavior**
Currently SLTS buffer size and cache is hardcoded, it should be configurable

**Screenshots**
Not applicable.

**Additional context**
None.

sachin-j-joshi:

**Change log description**  
Make metadata buffer size and metadata cache size configurable.

**Purpose of the change**  
Fixes #5861

**What the code does**  
Make metadata buffer size and metadata configurable.
- Add config settings to ChunkedSegmentStorageConfig 
- Pass ChunkedSegmentStorageConfig instance to BaseMetadataStore via factory.
- Rest is just plumbing

**How to verify it**  
Tests should pass


RaulGracia (changes requested):
Looks good, but you have piggy-backed in this PR a default value change of `MAX_METADATA_ENTRIES_IN_CACHE` from 5K to 16K. Will approve once you confirm that this has no negative impact in the long running operation of the Segment Store (e.g., OOM).

sachin-j-joshi:
Most overhead of metadata is associated with the strings for segment and chunk names used in metadata. 
The size of metadata is expected to be less than 1K even in worst case. So this change potentially adds 11MB to the working set. This should not cause any OOM issues per se.

Original 5K value was chosen arbitrarily, the new value 16K is also equally arbitrary except it is bigger :) 
So good idea to have a some perf runs to come up with that value.
However more metadata is cached the better, so we are here choosing slightly bigger value.

Although I agree its the last straw that breaks camel's back.
This change is more about being able to change the cache size so that we can test different sizes.
I'll revert back the value change and we'll plug in better value after some experimentation.

RaulGracia (approved):

PR #5871, opened 2021-03-22
Issue 5880: Reduce threads spawned while creating a StateSynchronizer and ReaderGroupManager.

### LINKED ISSUE #5880: Reduce the resources used by ReaderGroupManager and StateSyncrhonizer.
**Describe the bug**
Reduce threads spawned while creating a StateSynchronizer and ReaderGroupManager...

**To Reproduce**
Ensure the defaults used by StateSyncrhonizer and RederGroupManager are toned down.

**Screenshots**
NA

**Additional information**
NA

shrids:
**Change log description**  
Reduce threads spawned while creating a StateSynchronizer and ReaderGroupManager.

**Purpose of the change**  
Fixes #5880 .

**What the code does**  
Reduce threads spawned while creating a StateSynchronizer and ReaderGroupManager.
- Configure connection pooling to use a max of 1 connection per segment store.
- Reduce the number of internal thread pools to 1.

**How to verify it**  

Note: Verification is in progress.

### DIFF
--- a/client/src/main/java/io/pravega/client/ClientConfig.java
+++ b/client/src/main/java/io/pravega/client/ClientConfig.java
@@ -95,7 +95,7 @@ public class ClientConfig implements Serializable {
      * An internal property that determines if the client config.
      */
     @Getter(AccessLevel.PACKAGE)
-    private final boolean canOverrideMaxConnectionsConfiguration;
+    private final boolean isDefaultMaxConnections;
 
     /**
      * An internal property that determines whether TLS enabled is derived from Controller URI. It cannot be set
@@ -193,7 +193,7 @@ public class ClientConfig implements Serializable {
         private boolean validateHostName = true;
 
         private boolean deriveTlsEnabledFromControllerURI = true;
-        private boolean canOverrideMaxConnectionsConfiguration = true;
+        private boolean isDefaultMaxConnections = true;
 
         /**
          * Note: by making this method private, we intend to hide the corresponding property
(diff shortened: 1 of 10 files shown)


shrids:
System test run pravega-branch-builder/1038 is in progress.

client/src/main/java/io/pravega/client/SynchronizerClientFactory.java
```diff
@@ -36,9 +36,15 @@
      * @return Instance of ClientFactory implementation.
      */
     static SynchronizerClientFactory withScope(String scope, ClientConfig config) {
-        val connectionFactory = new SocketConnectionFactoryImpl(config);
-        return new ClientFactoryImpl(scope, new ControllerImpl(ControllerImplConfig.builder().clientConfig(config).build(),
-                connectionFactory.getInternalExecutor()), connectionFactory);
+        // Change the max number of number of allowed connections to the segment store to 1.
+        val updatedConfig = config.toBuilder()
```
tkaitchuck, line 40: I don't think this change is appropriate. Here a user has a setting in the config that they can set, and we are just overriding it to be 1. Given that a factory can be used to create many objects there ought to be some way to configure the number of connections. This removes that.

I think a better solution would be to change the ClientConfig object to track if the value set is the default. Then if it has not been set then we can make the default different for different paths, such as by setting 1 here.

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
```diff
@@ -42,7 +44,13 @@ public static ReaderGroupManager withScope(String scope, URI controllerUri) {
      * @return Instance of Stream Manager implementation.
      */
     public static ReaderGroupManager withScope(String scope, ClientConfig clientConfig) {
-        return new ReaderGroupManagerImpl(scope, clientConfig, new SocketConnectionFactoryImpl(clientConfig));
+        // Change the max number of number of allowed connections to the segment store to 1.
+        val updatedClientConfig = clientConfig.toBuilder()
+                .maxConnectionsPerSegmentStore(1)
```
tkaitchuck, line 49: Same here.

tkaitchuck (changes requested):

client/src/main/java/io/pravega/client/SynchronizerClientFactory.java
shrids: Thanks, modified the code to ensure `ClientConfig `object is modified only if the default value is used.

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
shrids: fixed it.

tkaitchuck (approved):

client/src/main/java/io/pravega/client/ClientConfig.java
```diff
@@ -187,6 +193,7 @@ public boolean isEnableTlsToSegmentStore() {
         private boolean validateHostName = true;
 
         private boolean deriveTlsEnabledFromControllerURI = true;
+        private boolean canOverrideMaxConnectionsConfiguration = true;
```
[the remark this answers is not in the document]
shrids: Thanks I have improved the code...

shrids:
test-pravega-continuous-1/1395/ is in progress.

PR #5883, opened 2021-03-24
Issue 5875: Add request id tracing to all parts of controller's processing of request

### LINKED ISSUE #5875: Request Id tracing across controller process, esp in metadata store calls to segment store

Request id is received on the grpc layer and traced on the background worker frameworks. However, it is not used to trace all the metadata store calls. And for any background job that controller brings up, the request id is not generated and propagated making it hard to trace the origin and flow of requests. 

We already have the request id included with wire commands. Its about ensuring that it is used consistently. 

**Additional context**
This will improve the request tracking across controller.

shiveshr:
**Change log description**  
At the start of each operation (either a user request, or originating from a background work) we generate a new request id and trace it across all logs and use it while interacting with segment store.
- Adds request id tracing to metadata store in controller. 
- Request id tracing to all calls into segment store to use the request id generation at the start of an operation. 
Optimizes caching in metadata store by caching "mutable" records and used within an operation if it was fetched after the operation was started.  

**Purpose of the change**  
Fixes #5875, #5358  

**What the code does**  
Tracks using request id received in grpc request. 
A new class called RPCTracingTags added where tag names are defined as constants and used in both controllerServiceImpl and ControllerImpl to tag correctly. 
These ids are then passed all the way down to metadata store and segment helper for tagging these requests. However, beyond segment helper, a new request id (flow id) is generated and passed to segment stores as part of connection pooling. 
The new ids are also generated for any background work that controller initiates - like auto scaling, retention and commit transactions. These are also used in all metadata and segment store calls and logged. 
To enable metadata store to include requestId, we augmented it to Operation context 
Any other source of stream operation generation relies on generates request ids. This is passed to metadata store via OperationContext. 
`Operation context` is now an interface with multiple implementations - scopeContext, StreamContext, KvtContext and RGContext where it holds the object of interest so that its not refreshed. It also includes request id and start time. 
the start time is used to optimize on fetches from segment store. We basically fetch from cache if it was loaded in cache after the start time.. unless the processing logic explicitly requests to ignore the cached value. this is important for consistency (and we already have those apis available and nothing is modified). 
So all requests now include caller's and its duly logged too.. 
Caching: the time is used in the caching in PravegaTableStoreHelper and cached entries are returned predicated on time when it was loaded in the cache. Also, after writing/updating the entry in the store, we again load them in the cache, predicated on if its a cache miss or cache has staler version. 
Cache earlier stored the byte array. Instead now we cache the deserialized object so that we dont spend cycles deserializing the same byte array each time someone accesses a value from the cache. 

Another optimization is to also cache the scope id and use the cached value and only when we get DataContainerNotFound exception do we attempt to fetch the scope id again. This reduces one store call per new operation context - to read the scopes table to get the scope id. We still continue to read the stream id(rg id/kvt id) as the first thing before accessing its metadata though. So even if the actual metadata is cached, we still first get the resource id to make sure we are not accessing an older iteration of the resource. 

So all store code path is modified to use operation context. and if the supplied operation context is null, a new on the fly operation context is created and used. 

Also fixed one code `bug` that was found because of a failing test due to new changes. The bug was in deleteKvt where the kvt entry was first removed from kvtInScopes table and then its corresponding metadata table is removed. This should have been done in the reversed order - remove the table.. then remove the id. 

Formatting:
I also took the liberty to fix the formatting wherever necessary (which has bulked up the change list, sorry). But there were several places where we did not follow the 120 char width which made the code unreadable. This is the primary formatting change and there are few other places where the indentation is also fixed. 

**How to verify it**  
Unit tests added.


controller/src/main/java/io/pravega/controller/store/stream/StreamOperationContext.java
```diff
@@ -0,0 +1,39 @@
...
+    @Getter
+    private final Scope scope;
+    @Getter
+    private final Stream stream;
+    @Getter
+    private final long requestId;
+    @Getter
+    private final long operationStartTime = System.currentTimeMillis();
+
+    StreamOperationContext(Scope scope, Stream stream, long requestId) {
```
shrids, line 32: We can use @AllArgsConstructor @NonNull here..

controller/src/main/java/io/pravega/controller/store/stream/VersionedTransactionData.java
```diff
@@ -28,7 +28,8 @@
 @Data
 @AllArgsConstructor
 public class VersionedTransactionData {
-    public static final VersionedTransactionData EMPTY = new VersionedTransactionData(Integer.MIN_VALUE, new UUID(0, 0), null,
+    public static final VersionedTransactionData EMPTY = new VersionedTransactionData(Integer.MIN_VALUE,
```
shrids, line 31: this seems like a formatting change...

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
```diff
@@ -68,6 +69,8 @@
     static final String COMPLETED_TX_BATCH_ROOT_PATH = COMPLETED_TX_ROOT_PATH + "/batches";
     static final String COMPLETED_TX_BATCH_PATH = COMPLETED_TX_BATCH_ROOT_PATH + "/%d";
 
+    private static final TagLogger log = new TagLogger(LoggerFactory.getLogger(SegmentHelper.class));
```
shrids, line 72: remove SegmentHelper.class reference here.

shrids (changes requested):
Thanks for the changeset Shivesh. I did not get a chance to go through all the files, will update the PR as when I make progress. Do fix the conflicts with your branch...

controller/src/main/java/io/pravega/controller/store/stream/VersionedTransactionData.java
shiveshr: yes, i have made several formatting changes, particularly of the type where the line exceeded 120 chars.. it makes it very difficult to read the code so took care of that debt. have mentioned it in the description

controller/src/main/java/io/pravega/controller/util/RetryHelper.java
```diff
@@ -36,7 +37,8 @@
     public static final Predicate<Throwable> RETRYABLE_PREDICATE = e -> {
         Throwable t = Exceptions.unwrap(e);
         return RetryableException.isRetryable(t) || (t instanceof CheckpointStoreException &&
-                ((CheckpointStoreException) t).getType().equals(CheckpointStoreException.Type.Connectivity));
+                ((CheckpointStoreException) t).getType().equals(CheckpointStoreException.Type.Connectivity)) || 
+                t instanceof IOException;
```
shrids, line 41: when is IOExcpetion thrown?

controller/src/test/java/io/pravega/controller/server/eventProcessor/requesthandlers/StreamRequestProcessorWithPravegaTablesStore.java
```diff
@@ -0,0 +1,57 @@
...
+import io.pravega.controller.store.stream.StreamStoreFactory;
+import io.pravega.test.common.TestingServerStarter;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.apache.curator.test.TestingServer;
+import org.junit.After;
+import org.junit.Before;
+
+public class StreamRequestProcessorWithPravegaTablesStore extends StreamRequestProcessorTest {
```
shrids, line 30: nit: Can we add a Test suffix to the class?

shrids (commented):
Completed one more pass over the changeset.

pbelgundi:
@shiveshr Could you please remove this log or change its log level to "trace" as part of this PR?
"
2021-04-29 08:37:04,560 185332 [controllerpool-13] INFO  i.p.c.c.impl.ConnectionPoolImpl - Reusing connection: ConnectionPoolImpl.Connection(uri=10.244.1.206:12345, flowHandler=java.util.concurrent.CompletableFuture@6473f9ea[Completed normally])
"

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -266,29 +275,32 @@ private String getTransactionName(String scope, String stream, long segmentId, U
...
-                    handleReply(requestId, r, connection, transactionName, WireCommands.MergeSegments.class, type);
+        return sendRequest(connection, clientRequestId, request)
+                .thenCompose(r -> {
+                    handleReply(clientRequestId, r, connection, transactionName, WireCommands.MergeSegments.class, type);
                     if (r instanceof WireCommands.NoSuchSegment) {
                         WireCommands.NoSuchSegment reply = (WireCommands.NoSuchSegment) r;
                         if (reply.getSegment().equals(transactionName)) {
-                            return TxnStatus.newBuilder().setStatus(TxnStatus.Status.SUCCESS).build();
+                            // idempotent case when segment is already merged, we get the write offset for the parent segment.
+                            return getSegmentInfo(scope, stream, targetSegmentId, delegationToken, clientRequestId)
```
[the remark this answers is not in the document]
shiveshr: this is used for watermarking. we get the offset of parent segment where the txn was merged and use that to compute the watermark.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesReaderGroup.java
```diff
@@ -52,132 +53,131 @@
...
             return CompletableFuture.completedFuture(id);
         } else {
-            return readerGroupsInScopeTableNameSupplier.get()
-                    .thenCompose(readerGroupsInScopeTable ->
-                            storeHelper.getEntry(readerGroupsInScopeTable, getName(),
-                                    x -> BitConverter.readUUID(x, 0)))
+            // first get the scope id from the cache.
+            // if the cache does not contain scope id then we load it from the supplier. 
+            // if cache contains the scope id then we load the streamid. if not found, we load the whole shit
+            return Futures.exceptionallyComposeExpecting(
```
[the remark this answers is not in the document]
shiveshr: you are right and its triplicated actually. 
but in three different classes that are otherwise unrelated. 
let me see if i can take this into some common helper.

controller/src/main/java/io/pravega/controller/util/RetryHelper.java
shiveshr: our serializers and deserializers for metadata records can throw IOExceptions. Similarly curator can throw IO exception while reading a znode. Since IO exceptions are generally retriable, so i have included it here.

controller/src/main/java/io/pravega/controller/store/stream/Cache.java
```diff
@@ -27,30 +28,63 @@
...
     }
 
     public void invalidateCache(final CacheKey key) {
         cache.invalidate(key);
     }
 
-    public void put(CacheKey cacheKey, VersionedMetadata<?> record) {
-        cache.put(cacheKey, record);
+    public void put(CacheKey cacheKey, VersionedMetadata<?> record, long time) {
+        synchronized (getLock(cacheKey)) {
```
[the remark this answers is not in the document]
shiveshr: updated the javadoc and comments in the code to explain the rationale.
our cache class is a wrapper over guava's cache class. it provides a method for put which has been changed to be a conditional put. 
however, guava cache does not have any api for conditional puts. 
so what we do here is we take a hash of the key (key hash modulo n) where n is the number of locks we have created on this cache. so for conditional update for a key, we first take the corresponding lock and then we do a get and set conditionally for that key. 
by creating n locks we increase the concurrency for this class in its ability to do multiple concurrent puts and keys contest for the same lock only if their hash puts them to use the same lock.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesReaderGroup.java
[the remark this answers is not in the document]
shiveshr: done

controller/src/main/java/io/pravega/controller/store/stream/Cache.java
```diff
@@ -17,40 +17,94 @@
...
+    }
+
+    
+    public void put(CacheKey cacheKey, VersionedMetadata<?> record, long time) {
+        // acquire the lock for key partition. Then perform a conditional update - get, compare and swap.
+        // condition for update => if either the entry doesnt exist in cache. OR the entry in cache is older (lower key version)
+        // than new value to be updated. 
+        synchronized (getLockObject(cacheKey)) {
+            Cache.CacheValue existing = cache.getIfPresent(cacheKey);
+            if (existing == null || record.getVersion().compareTo(existing.getValue().getVersion()) > 0) {
```
shrids, line 88: nit: Ensure record is not null.

shrids (approved):
Completed one more pass, no blocking comments from my end.
Since this is a huge changeset could we run the system tests too ?

shrids:
`test-pravega-continuous-1/1512` system run is successful with this changeset...

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -90,6 +89,7 @@
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.Random;
```
pbelgundi, line 92: Why the change from RandomFactory to Random?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -167,41 +168,46 @@ void setTimeout(Duration duration) {
         final long requestId = connection.getFlow().asLong();
         Pair<Byte, Integer> extracted = extractFromPolicy(policy);
 
-        return sendRequest(connection, requestId, new WireCommands.CreateSegment(requestId, qualifiedStreamSegmentName,
+        return sendRequest(connection, clientRequestId, new WireCommands.CreateSegment(requestId, qualifiedStreamSegmentName,
```
pbelgundi, line 171: We seem to have replaced "FlowId" with clientRequestId. Does this also work for Segment Store?

shared/controller-api/src/main/java/io/pravega/shared/controller/tracing/RPCTracingTags.java
```diff
@@ -0,0 +1,63 @@
+/**
```
pbelgundi, line 1: Could these be added to the existing RequestTracker class instead of creating a new class ?

pbelgundi (commented):
Mostly LGTM, except for few minor comments and questions

shared/controller-api/src/main/java/io/pravega/shared/controller/tracing/RPCTracingTags.java
shiveshr: i think its better organized if its kept in a dedicated class for these constants..

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: we se the flowId inside the wirecommand.
the client requestid here is something that this api received and captures the request context.. 
its simply used for logging purpose

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
shiveshr: no reason, we just want a random number for the request id, in case its not provided for by the caller.
i could change it back to random factory and use it to create a Random object.

pbelgundi:
> > 2021-04-29 08:37:04,560 185332 [controllerpool-13] INFO i.p.c.c.impl.ConnectionPoolImpl - Reusing connection: ConnectionPoolImpl.Connection(uri=10.244.1.206:12345, flowHandler=java.util.concurrent.CompletableFuture@6473f9ea[Completed normally])
> 
> @pbelgundi ideally since this PR is covering pravega controller, and about request ids, we should handle it in a separate issue. secondly, @shrids do you see any drawbacks in changing the log level as suggested?

I have addressed this in PR: https://github.com/pravega/pravega/pull/6014

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
pbelgundi: Ok.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
pbelgundi: I see the _requestId_ being logged by processAndRethrowException(). It would be good to also add it to this log inside sentRequest():
```
log.debug("Sending request: {}", request);
```

controller/src/main/java/io/pravega/controller/server/eventProcessor/LocalController.java
```diff
@@ -86,28 +87,28 @@ public LocalController(ControllerService controller, boolean authorizationEnable
 
     @Override
     public CompletableFuture<Boolean> checkScopeExists(String scopeName) {
-        return Futures.exceptionallyExpecting(this.controller.getScope(scopeName).thenApply(v -> true),
+        return Futures.exceptionallyExpecting(this.controller.getScope(scopeName, 0L).thenApply(v -> true),
```
RaulGracia, line 90: Are these `0L` the request IDs? Because the `RequestTracker` identifies requests based on the ID + request type. So, 2 successive requests of any of these methods would be undistinguishable from one another as the ID is the same.

controller/src/main/java/io/pravega/controller/store/stream/ZkInt96Counter.java
```diff
@@ -119,26 +119,28 @@ public ZkInt96Counter(ZKStoreHelper storeHelper) {
...
-                                                                                     // next counter by calling counter.incrementAndGet otherwise there will
-                                                                                     // be a collision with counter used by someone else.
-                                                                                     counter.set(previous.getMsb(), previous.getLsb());
-                                                                                     limit.set(nextLimit.getMsb(), nextLimit.getLsb());
-                                                                                     refreshFutureRef = null;
-                                                                                     log.info("Refreshed counter range. Current counter is {}. Current limit is {}", counter.get(), limit.get());
-                                                                                 }
-                                                                             });
-                                                       }));
+        return storeHelper
```
pbelgundi, line 122: What is the change here? Perhaps a formatting change that needs to be reverse...

pbelgundi (approved):
LGTM with a few minor comments.

controller/src/main/java/io/pravega/controller/store/stream/Cache.java
```diff
@@ -15,42 +15,98 @@
...
+        }
     }
 
     public void invalidateCache(final CacheKey key) {
-        cache.invalidate(key);
+        synchronized (getLockObject(key)) {
+            cache.invalidate(key);
+        }
+    }
+
```
RaulGracia, line 81: Nit: extra space.

RaulGracia (commented):
Looks good, just few minor comments. One of them is the post important regarding some places (e.g., LocalController) in which it looks like the `requestId` is initialized always to `0L`. If this is the case, this could confuse `RequestTracker`, as it fetches requests identifying them with the ID + request type, right?

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -1191,14 +1273,19 @@ private TxnSegments convert(CreateTxnResponse response) {
...
+    public CompletableFuture<Void> commitTransaction(final Stream stream, final String writerId, final Long timestamp, 
+                                                     final UUID txId) {
         Exceptions.checkNotClosed(closed.get(), this);
         Preconditions.checkNotNull(stream, "stream");
         Preconditions.checkNotNull(txId, "txId");
         long traceId = LoggerHelpers.traceEnter(log, "commitTransaction", stream, txId);
-        log.info("Commit transaction {} invoked on Stream {} for writerId {} with timestamp {}", txId, stream, writerId, timestamp);
+        final long requestId = requestIdGenerator.get();
+
+        log.info(requestId, "Commit transaction {} invoked on Stream {} for writerId {} with timestamp {}",
```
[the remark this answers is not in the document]
shiveshr: The transactions also result in some metadata operations (like table reads or table updates) where the txn id is not logged directly. iIth this change we are using the request id in all our metadata operations now universally, which will include it for txn related operations too like create, abort, commit etc where the metadata operations (at debug levels) can be traced from request id and we will know which read table log is for which request.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shiveshr: we are logging it inside the sendRequest method too. its just that its at trace and not debug level. 

https://github.com/pravega/pravega/blob/3ea93f2994e42b3f90b1922eb055e8ba80f03408/controller/src/main/java/io/pravega/controller/server/SegmentHelper.java#L741

controller/src/main/java/io/pravega/controller/store/stream/ZkInt96Counter.java
shiveshr: yes, its only formatting. have kept such formatting changes here too because there were many places in our store code where the 120 char width was not adhered to (esp here) making the code difficult to read thru.

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -702,47 +724,49 @@ private TableSegmentEntry convertFromWireCommand(Map.Entry<WireCommands.TableKey
...
+                log.warn(callerRequestId, "Exception tearing down connection: ", e);
             }
         }
     }
 
-    private <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(RawClient connection, long requestId, T request) {
-        CompletableFuture<Reply> future = Futures.futureWithTimeout(() -> connection.sendRequest(requestId, request), timeout.get(), "request", executorService);
+    private <T extends Request & WireCommand> CompletableFuture<Reply> sendRequest(RawClient connection, long clientRequestId, 
+                                                                                   T request) {
+        log.trace(clientRequestId, "Sending request to segment store with: flowId: {}: reply: {}",
```
pbelgundi, line 741: the message says "reply" but we're printing the "request".

controller/src/main/java/io/pravega/controller/server/eventProcessor/LocalController.java
shiveshr: Yes, these 0L are request ids. 
these apis bypass grpc, hence request tracker doesnt track them. 
The local controller in controller service is used to basically bypass the call to controller from pravega clients used in controller. 
These clients are used to write events into internal streams or read them back.

so we dont really care to track these requests per se hence i had used 0L. but yes, using `0L` is quite pointless. I am changing it to generate a new request id in such situations.

RaulGracia:
thanks @shiveshr, will approve once the build passes (now there are few tests failing, perhaps due to the last commit).

RaulGracia (approved):

PR #5892, opened 2021-03-25
Issue 5493: Unit test failure TimeoutServiceZkStoreTest.testPingSuccess

### LINKED ISSUE #5493: Unit test failure: TimeoutServiceZkStoreTest.testPingSuccess
**Problem description**
Stack trace:
```
io.pravega.controller.timeout.TimeoutServiceZkStoreTest > testPingSuccess FAILED
    java.lang.AssertionError
```

**Problem location**
`TimeoutServiceZkStoreTest.testPingSuccess`

Test should pass reliably.

RaulGracia:
**Change log description**  
Adjusted polling times in testPingSuccess to prevent it from failing sporadically in slow instances.

**Purpose of the change**  
Fixes #5493.

**What the code does**  
This test fails sporadically when it expects `getTaskCompletionQueue()` to be empty when polling elements and it is not. The most plausible reason for that sporadic failure is that the call that induces the failure has a poll interval excessively large that brings the polling timeout too close (1 second) from the actual Transaction timeout. If for some reason there test gets stuck in between the `poll()` call and the `assertNull(result)` for around 1 second, then the assertion can fail as the Transaction timeout would lead  `getTaskCompletionQueue()` to potentially contain one element.

This PR just makes the polling interval of that assertion shorter, while providing more time to the next assertion that actually expects the Transaction to time out and `getTaskCompletionQueue()` to contain one element. This should make these sporadic failures far less likely.

**How to verify it**  
Build should pass consistently on this test. While it is not a sufficient validation, executing multiple time this test locally works well.


controller/src/test/java/io/pravega/controller/timeout/TimeoutServiceTest.java
```diff
@@ -225,13 +225,13 @@ public void testPingSuccess() throws InterruptedException {
         PingTxnStatus pingStatus = timeoutService.pingTxn(SCOPE, STREAM, txData.getId(), txData.getVersion(), LEASE);
         Assert.assertEquals(PingTxnStatus.Status.OK, pingStatus.getStatus());
 
-        result = timeoutService.getTaskCompletionQueue().poll((long) (0.5 * LEASE), TimeUnit.MILLISECONDS);
+        result = timeoutService.getTaskCompletionQueue().poll((long) (0.1 * LEASE), TimeUnit.MILLISECONDS);
```
shiveshr, line 228: we should infact do the following: 
have larger lease to guarantee that the lease has not expired and hence task queue is empty. 

and later in the code where we want the lease to take expired, we should poll for at least greater than lease period. 

but this second aspect can make this test run for long if the lease period is large.

So ideal fix would be to inject a clock in the timeoutService.. and progress the time manually from the test so that we can guarantee when the clock goes past the lease. 
otherwise we always have to choose between a large enough lease to not run into these timing issues vs choosing a small time to ensure test does not take long to complete.

RaulGracia:
Closing now, will address this one from another angle in the future.

PR #5896, opened 2021-03-26
Issue 5893: Sporadic test failure AppendTest.miniBenchmark

### LINKED ISSUE #5893: Sporadic test failure AppendTest.miniBenchmark
**Describe the bug**
Found test `AppendTest.miniBenchmark` failing sporadically on Jenkins:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at io.pravega.test.integration.AppendTest.miniBenchmark(AppendTest.java:418)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

**To Reproduce**
Occurs sporadically when executing `gradlew build` on Jenkins.

**Screenshots**
n/a

**Additional information**
While the test worked well, it looks like is the slowness of the instance the one that caused the assertion error. The test is expecting both sync and async writes to take less than 10 seconds, but in a slow instance they can take more:
```
Max latency: 3039.230622
Blocking took: 10969ms.
Non blocking took: 5426ms.
```
Adding a larger timeout or a larger assertion time would be enough to mitigate this problem.

RaulGracia:
**Change log description**  
Reduced load in benchmark-like integration tests to avoid sporadic failures in slow environments.

**Purpose of the change**  
Fixes #5893.

**What the code does**  
Increased timeout assertions in `AppendTest` as they are dependent on the environment performance. Also, reduced load in `WriteBatchTest` as it was consistently failing on a low power VM. Also, replaced `System.out.println` calls by `log` calls.

**How to verify it**  
Modified tests should pass more reliably.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -180,11 +180,11 @@ project('common_server') {
 }
 
 def withoutLogger = { exclude group: 'org.slf4j', module: 'slf4j-log4j12'
-                      exclude group: 'org.slf4j', module: 'slf4j-simple' }
+    exclude group: 'org.slf4j', module: 'slf4j-simple' }
 
 def withoutJaxbAndJjwt = { exclude group: 'javax.xml.bind', module: 'jaxb-api'
-                    exclude group: 'org.glassfish.jaxb', module: 'jaxb-runtime'
-                    exclude group: 'io.jsonwebtoken', module: 'jjwt'}
+    exclude group: 'org.glassfish.jaxb', module: 'jaxb-runtime'
+    exclude group: 'io.jsonwebtoken', module: 'jjwt'}
 
 project('shared:authplugin') {
     dependencies {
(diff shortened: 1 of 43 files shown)


tkaitchuck (changes requested):
Please increase the number of events to make these timeouts reasonable. The point of this test is to catch performance regressions, and I fear this undermines that.

RaulGracia:
@tkaitchuck testing performance in a build seems tricky to me, given that we are not in control of the resources available in instances executing the build. Thus, writing a number of events within some time interval may be fine for some instances and may be too much for others. This is the case with this test: while it passes most of the time, there are situations in which the instance is so slow that the time-based assertions fail (e.g., the assertion failed because the sync writes took 10.9 seconds instead of 10 seconds). If I add more events to write, I'm concerned that this PR will not be fixing anything, given that we increase both the time and the load, so we end up in the same situation.

tkaitchuck:
If your PR caused it to take 10.9 seconds that is a major performance regression because it normally takes < 2 sec to complete.
So I would say it has done it's job. The point of the test was not to be super tight, but close enough to catch an order of magnitude decrease in performance. This change would undo that, in which case there is no point in having the test at all.

RaulGracia:
@tkaitchuck the point is that no PR caused this test to take 10.9 seconds; instead, if we execute the exact same code multiple times on Jenkins, it could be that one of the runs, perhaps due to environmental issues or resource scarcity, make this test to be slow enough to fail the assertion. This basically means that this test induces sporadic false positives on slow or overloaded environments. But here we have a trade-off between the rate of false positives and the accuracy detecting performance regressions, which looks hard to do in an environment that normally has low resources. In any case, will close the PR as the fix consist just to make the test less demanding in favor of reducing false positives.

PR #5910, opened 2021-04-01

### LINKED ISSUE #5846: Create exception to force Segment Container shutdown

**Solutions**

ManishKumarKeshri:
**Change log description**  

**Purpose of the change**  
Fixes #5846 

**What the code does**  

**How to verify it**  
Build shall pass.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/ServiceHaltException.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/ServiceHaltException.java
@@ -15,38 +15,49 @@
  */
 package io.pravega.segmentstore.server;
 
+import io.pravega.segmentstore.contracts.StreamingException;
+import lombok.Getter;
+
 /**
- * Exception that is thrown whenever we detect an unrecoverable data corruption.
- * Usually, after this is thrown, our only resolution may be to suspend processing in the container or completely bring it offline.
+ * Exception that is thrown whenever we detect an unrecoverable state.
+ * Usually, after this is thrown, our only resolution will be to shut down the Segment Container, DurableLog and StorageWriter.
  */
-public class DataCorruptionException extends ServiceHaltException {
+public class ServiceHaltException extends StreamingException {
     /**
      *
      */
     private static final long serialVersionUID = 1L;
 
     /**
-     * Creates a new instance of the DataCorruptionException class.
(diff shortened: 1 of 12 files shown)


andreipaduroiu (approved):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/ServiceHaltException.java
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.pravega.segmentstore.server;
+
+import io.pravega.segmentstore.contracts.StreamingException;
+import lombok.Getter;
+
+public class ServiceHaltException extends StreamingException {
```
RaulGracia, line 21: Add comments to this Exception.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/ServiceHaltException.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     *
```
RaulGracia, line 23: Do we need a blank comment here?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/MemoryStateUpdater.java
```diff
@@ -142,13 +143,13 @@ void cleanupReadIndex() {
      * @param operations An Iterator iterating over the operations to process (in sequence).
      * @param callback   A Consumer that will be invoked on EVERY {@link Operation} in the operations iterator, in the
      *                   order returned from the iterator, regardless of whether the operation was processed or not.
-     * @throws DataCorruptionException If a serious, non-recoverable, data corruption was detected, such as trying to
-     *                                 append operations out of order.
+     * @throws ServiceHaltException    If a serious, non-recoverable state was detected, such as unable to create a
+     *      *                          CachedStreamSegmentAppendOperation.
```
RaulGracia, line 147: Remove this extra `*`.

RaulGracia (changes requested):

ManishKumarKeshri:
Moved the changes to https://github.com/pravega/pravega/pull/5952.

PR #5914, opened 2021-04-06
Issue 5634: Follow Up Logging Patches

### LINKED ISSUE #5634: System Test Logging Enhancements

co-jo:
**Change log description**  
* Adds the fluent-bit helm repository to the setup scripts.
* Removes invalid function call in description.
* Log timestamp now based on time of last modification instead of time of rotation.

**Purpose of the change**  
Fixes #5634 

**What the code does**  
* Fixes above issues.

**How to verify it**  

Run systems tests.


RaulGracia (changes requested):
Not working on Jenkins, it fails during the log collection phase.

co-jo:
> Not working on Jenkins, it fails during the log collection phase.

Fixed.

RaulGracia (approved):

PR #5926, opened 2021-04-12
Issue 5925: Properly set atTail flag.

### LINKED ISSUE #5925: SegmentRead.atTail not always properly set.
**Describe the bug**
The atTail flag which is sent to the client on the SegmentRead message is not always correct. This doesn't have any direct impact on the correctness of the client because it is assumed that additional data can be appended at any time. However for diagnostic / debugging purposes it would be preferable if this information was correct.

tkaitchuck:
**Change log description**
* Properly set `atTail` flag on SegmentRead message
* Increase code coverage

**Purpose of the change**  
Fixes #5925. This also significantly increases code coverage. (5925 was discovered by inspecting code coverage reports to look for uncovered code paths)

**What the code does**  
Ensures that read reply data returned to client is correct. (The client does not currently take advantage of this information but it may in the future.)


controller/src/test/java/io/pravega/controller/server/v1/ControllerServiceTest.java
```diff
@@ -144,11 +144,11 @@ public void setup() throws Exception {
         startTs = System.currentTimeMillis();
         OperationContext context = streamStore.createContext(SCOPE, stream1);
         streamStore.createStream(SCOPE, stream1, configuration1, startTs, context, executor).get();
-        streamStore.setState(SCOPE, stream1, State.ACTIVE, context, executor);
+        streamStore.setState(SCOPE, stream1, State.ACTIVE, context, executor).get();
```
tkaitchuck, line 147: This is unrelated to the rest of the change but fixes a flakey test.

shrids (approved):
LGTM.

controller/src/test/java/io/pravega/controller/server/v1/ControllerServiceTest.java
shrids: ok

PR #5929, opened 2021-04-13
Issue 5928: Always upload logs from github actions builds.

### LINKED ISSUE #5928: Upload report even if build fails
It is useful to have a build report if a github actions build fails. As such we should upload it even in the event of a test failure.

tkaitchuck:

**Change log description**  
Uploads artifacts from github actions even in the event of build failure.

**Purpose of the change**  
Fixes #5928

**What the code does**  
No code changes

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -124,6 +124,7 @@ jobs:
       - name: Tar Reports
         run: tar --use-compress-program zstd -cf reports-${{github.job}}.tzst `echo ${{env.REPORTS_LOCATIONS}}`
       - name: Upload Reports
+        if: always()
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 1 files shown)


fpj:
The change looks good, but there are a couple of points I wanted to understand: where are we uploading it to and is there a limit on the amount storage space we use for such logs?

tkaitchuck:
It is being uploaded to the Github's UI. If you click on the build and look in the upper right there is now a pull down that says "artifacts" where they are located. We are limited to 5GB in total. Currently each build uses ~6mb and is configured to be retained for 4 days after the build. At this rate we can support ~200 builds per day with no issue. If we find we are exceeding this limit we can either lower the retention time or pay for additional storage.

fpj (approved):

PR #5948, opened 2021-04-16

### LINKED ISSUE #5846: Create exception to force Segment Container shutdown

ManishKumarKeshri:
**Change log description**  

**Purpose of the change**  
Fixes #5846 

**What the code does**  

**How to verify it**  
Build shall pass.


andreipaduroiu:
I believe this is a duplicate of #5952. Please reopen if this is incorrect.

PR #5952, opened 2021-04-20

### LINKED ISSUE #5846: Create exception to force Segment Container shutdown

ManishKumarKeshri:
**Change log description**  

**Purpose of the change**  
Fixes #5846 

**What the code does**  

**How to verify it**  
Build shall pass.


RaulGracia:
@ManishKumarKeshri codecov seems to be reporting correctly, but your change does not meet the coverage check. Please, add the necessary test coverage to have this check passing.

ManishKumarKeshri:
HI @RaulGracia , Based on code coverage report, can you suggest on improving the coverage of DebugRecoveryProcessor -> recoverOperation or any other files. https://app.codecov.io/gh/pravega/pravega/compare/5952/diff

andreipaduroiu (approved):

andreipaduroiu:
I have reviewed this PR and approved the changes. However due to their nature, these lines of code are very hard to reach in a deterministic manner (using unit tests) - they are usually the last line of defense in case we really could not figure out why we ended up in a bad situation and hence prefer to shut down vs continuing. 

I do not know how to add more unit tests to cover these - if it were easy, they would likely have already been covered. 

@fpj would you be willing to make an exception for this PR and merge it in with low code coverage? Feel free to inspect the report (linked in a CodeCov comment above) and see the comments around the lines of code touched. We literally changed `DataCorruptionException` to `ServiceHaltException` and did not alter anything in the functioning of the system.

RaulGracia:
@fpj can you advise whether we can merge this PR or not (due to the issue with codecov)?

PR #5954, opened 2021-04-20
Issue 5953: Missing "styler" property in java.gradle

### LINKED ISSUE #5953: Missing "styler" property in java.gradle
**Describe the bug**
In #4999, text output for a failed build has been set to red color. This is relying on a property called `styler` that is supposed to contain the definition of the red color. However, the definition of that property is not present, which makes the builds to fail without reporting what test has actually failed:
https://github.com/pravega/pravega/blob/4dec5b2027c87d518f126a6b9fa8c4ed37f037cc/gradle/java.gradle#L119

The error is as follows:
```
8:20:28  FAILURE: Build failed with an exception.
18:20:28  
18:20:28  * Where:
18:20:28  Script '/var/jenkins_home/workspace/pravega/gradle/java.gradle' line: 119
18:20:28  
18:20:28  * What went wrong:
18:20:28  Execution failed for task ':segmentstore:server:host:test'.
18:20:28  > Could not get unknown property 'styler' for task ':segmentstore:server:host:test' of type org.gradle.api.tasks.testing.Test.
```

**To Reproduce**
Just see a failed build report executing `gradlew build`

**Screenshots**
n/a

**Additional information**
n/a

RaulGracia:
**Change log description**  
Explicitly use text colors in test build failure output.

**Purpose of the change**  
Fixes #5953.

**What the code does**  
Replaces the undefined variable `styler` by the actual text color expected in a failure output. Also, the text color is reset after the output for a failed build has been printed.

**How to verify it**  
Observed a failed build and the original error does not appear in Jenkins.

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -113,11 +113,11 @@ plugins.withId('java') {
             while (outputCache.size() > 300) outputCache.remove()
         }
 
-	afterTest { descriptor, result ->
+	    afterTest { descriptor, result ->
             if (result.resultType == TestResult.ResultType.FAILURE && outputCache.size() > 0) {
                 println()
-		def red = new String((char) 27) + "[31m"
-		def black = new String((char) 27) + "[0m"
+		        def red = new String((char) 27) + "[31m"
+		        def black = new String((char) 27) + "[0m"
                 println("${red}${descriptor.getClassName()}.${descriptor.getDisplayName()} FAILED. Output (Last 300 lines):${black}")
                 outputCache.each { print(" > $it") }
             }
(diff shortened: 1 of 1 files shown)


gradle/java.gradle
```diff
@@ -116,7 +116,9 @@ plugins.withId('java') {
         afterTest { descriptor, result ->
             if (result.resultType == TestResult.ResultType.FAILURE && outputCache.size() > 0) {
                 println()
-                println("${styler['red'](${descriptor.getClassName()}.${descriptor.getDisplayName()} + " FAILED" )} Output (Last 300 lines):")
```
shrids, line 119: good catch, `styler `is not defined...
``` 
// found this snip in gradle docs... 
def styler = 'black red green yellow blue magenta cyan white'
            .split().toList().withIndex(30)
            .collectEntries { key, val -> [(key) : { "\033[${val}m${it}\033[0m" }] }
```

shrids (approved):

Issue #5939, opened 2021-04-14
KVT Beta 2: Update KVT Config in Controller and add API to retrieve it


andreipaduroiu:
**Describe the solution you'd like**
See [PDP-48/Controller](https://github.com/pravega/pravega/wiki/PDP-48-Key-Value-Tables-(Beta-2)#controller) for details.

We need to:
- Add `PrimaryKeyLength` (int) and `SecondaryKeyLength` (int) to the KeyValueTable Configuration.
- Make sure this is properly passed in from the Client.
- Make sure that `SegmentHelper` takes the sum of these two values and  encodes it into the `CreateTableSegment` request when needing to create a KVT-related Table Segment
    - IMPORTANT: Segments created by the Controller for its own metadata MUST pass 0 for this argument.
- Create a `GetKVTConfig` (name TBD) API that is accessible from the client which, when given a valid Scope+KVT Name, will return the 1) Partition Count, 2) Primary Key Length and 3) Secondary Key Length.
    - This should be accessible from the Client (so it needs to be fully implemented in all implementations of the `Controller` interface.
-  Find all `TODO`s in the codebase (in the feature branch) that reference this issue (search for `5939`) and address them. For example: `ModelHelper.java` method `encode(final KeyValueTableConfig config)`.

Issue #5931, opened 2021-04-14
Key-Value Tables Beta 2


andreipaduroiu:
Master issue for KVT Beta 2 work tracking. 
- PDP: [PDP-48](https://github.com/pravega/pravega/wiki/PDP-48-Key-Value-Tables-(Beta-2))
- Feature branch `feature-kvt-beta2` (collect all PRs into this one, then merge it into `master` when done).

Tasks:
- [x] Create feature branch.
- [x] #5934 
- [x] #5938 
- [x] #5935 
- [x] #5936 
- [x] #6042: 
- [x] #5937 
- [x] #5939 
- [x] #5940 
- [x] #5941 
- [x] #5942 

Final Task:
- [ ] Merge feature branch into `master` (which should close this issue).

PR #5960, opened 2021-04-22
Issue 5938: (KVT Beta 2) New Client API and Base Implementation.

### LINKED ISSUE #5938: KVT Beta 2: Public KeyValueTable API changes
See [PDP-48/Client](https://github.com/pravega/pravega/wiki/PDP-48-Key-Value-Tables-(Beta-2)#client-api) for details, including a link to the proposed API (under review).

We need to:
1. Make these changes to `KeyValueTable.java`
2. Delete the `KeyValueTableImpl.java` class (or empty it out - it will be reimplemented).
2. Ignore/comment-out any unit/integration tests that may be broken due to this. This will be restored in #5942 or other issues (but before #5931 is merged to master)

NOTE: this will effectively cripple the existing KVT functionality (yes, #5934 will do it too). The other changes in #5931 will restore them.

andreipaduroiu:
**Change log description**  
Updated `KeyValueTable.java` to reflect the new API proposed in PDP-48.
Updated `KeyValueTableImpl.java` to implement new contracts (except Iterators)
Updated `TableKey.java` and `TableEntry.java` to support new API contract.

NOTE: this is not a complete change. This PR will break KVT functionality, which is why several tests have been commented out and/or ignored (each has a TODO pointing to a GitHub issue to resolve it and the `feature-kvt-beta2` branch should not be merged to master before those are done). Those items will be resolved in future PRs against this branch.

**Purpose of the change**  
Fixes #5938. 

**What the code does**  
See description above.

Other changes: 
- KeyValueTableFactoryImpl.java. Hard-coding a KeyValueTableConfiguration with predefined Key and Value lengths. This will be fixed in a future PR>
- KVT Tests: Updated to use new API.
    - These tests now pass on the new `KeyValueTableImpl.java` code
    - These tests do not pass on the KVT Integration tests, which is why those tests have been ignored (for now). They will be re-enabled when the Controller and Segment Store changes will be done.
    - KVT Iterator Tests: ignored for now. These will be re-enabled and fixed in a future PR.
- User CLI: temporarily commented out KVT commands. These will be re-enabled and fixed in a future PR. They will also require some syntax updates to support new new API features.

**How to verify it**  
Build must pass.


client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -98,197 +102,147 @@
...
-     * @return A CompletableFuture that, when completed, will contain the {@link Version} associated with the newly
-     * inserted or updated entry. Notable exceptions:
-     * <ul>
-     * <li>{@link ConditionalTableUpdateException} If the Key is already present in the {@link KeyValueTable} for the
-     * provided Key Family.
-     * See the {@link KeyValueTable} doc for more details on Conditional Update Responses.
-     * </ul>
-     */
-    CompletableFuture<Version> putIfAbsent(@Nullable String keyFamily, @NonNull KeyT key, @NonNull ValueT value);
+    CompletableFuture<Version> put(@NonNull TableEntry entry);
```
tkaitchuck, line 136: Previously the structure and verbage of these APIs was explicitly modeled on those of [ConcurrentMap](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/concurrent/ConcurrentMap.html)

Is there some reason that you want to overload the `put` method instead?

client/src/main/java/io/pravega/client/tables/KeyValueTableConfiguration.java
```diff
@@ -36,4 +36,25 @@
...
+     */
+    private final int primaryKeyLength;
+
+    /**
+     * The number of bytes for the Secondary Key. This value cannot be changed after the Key-Value Table has been created.
+     *
+     * @param secondaryKeyLength The number of bytes for the Secondary Key.
+     * @return The number of bytes for the Primary Key.
+     */
+    private final int secondaryKeyLength;
```
tkaitchuck, line 54: Because these are set by a builder they are implicitly optional and as no default value is provided they will all default to `-1`. I don't think this is desireable.

client/src/main/java/io/pravega/client/tables/TableEntry.java
```diff
@@ -23,72 +24,110 @@
...
+
+    /**
+     * Creates a new {@link TableEntry} with a Key made of only a Primary Key, with a version that indicates the key
+     * must not exist.
+     *
+     * @param primaryKey The Entry Key (Primary Key only).
+     * @param value      The Entry Value.
+     * @return A {@link TableEntry} with a version set to {@link Version#NOT_EXISTS}.
+     */
+    public static TableEntry notExists(ByteBuffer primaryKey, ByteBuffer value) {
```
tkaitchuck, line 92: I find this function name confusing. IE: "How can an entry that does not exist have a value?"

tkaitchuck (changes requested):

client/src/main/java/io/pravega/client/KeyValueTableFactory.java
```diff
@@ -52,19 +51,12 @@ static KeyValueTableFactory withScope(String scope, ClientConfig config) {
...
      * @param clientConfiguration A {@link KeyValueTableClientConfiguration} to use for configuring the
      *                            {@link KeyValueTable} client.
-     * @param <KeyT>              Key Type.
-     * @param <ValueT>            Value Type.
      * @return A {@link KeyValueTable} that provides access to the requested Key-Value Table.
      */
-    <KeyT, ValueT> KeyValueTable<KeyT, ValueT> forKeyValueTable(
-            @NonNull String keyValueTableName, @NonNull Serializer<KeyT> keySerializer,
-            @NonNull Serializer<ValueT> valueSerializer, @NonNull KeyValueTableClientConfiguration clientConfiguration);
+    KeyValueTable forKeyValueTable(
```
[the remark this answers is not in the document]
andreipaduroiu: If I remember when we came up with that name last year, we had to choose between `createKeyValueTable` or something else. The problem I had with `create` was that it may be confusing to some folks who may believe we are actually creating the table at that point. This method simply creates a client-side object that provides access to a KVT.

I looked at our EventStreamWriter/Reader and for that case, using `create` makes a lot of sense - we are creating a "writer" or a "reader". In the KVT case, however we don't have such a denomination.

If you have a better suggestion than `for` (which is not `create`), then we can definitely address this.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
- * Key-Value Table are associated with a single Key Family; in this case the entire Key-Value Table load will be placed
- * on a single backing Table Segment instead of spreading it across many Table Segments, leading to eventual performance
- * degradation.
+ * <li> The Primary Key is mandatory and is used to group all {@link TableKey}s with the same Primary Key together in the
+ * same Table Partition; this allows multiple keys/entries sharing the same Primary Key to be updated/removed atomically.
+ * <li> The Secondary Key is optional.
+ * <li> Two {@link TableKey} instances are considered equal if they have the same Primary Key and same Secondary Key.
+ * The same Secondary Key may be "associated" with different Primary Keys, in which case the resulting {@link TableKey}s
+ * will be distinct. For example, consider Primary Keys PK1 != PK2 and Secondary Key SK1; key PK1.SK1 is different from
+ * PK2.SK1 and both are different from PK1 and PK2 (no secondary key associated).
```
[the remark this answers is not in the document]
andreipaduroiu: rephrased to `with no secondary key`

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
+ * <li> Multiple {@link TableKey}/{@link TableEntry} instances with the same Primary Key can be updated or removed
+ * atomically (either all changes will be applied or none will).
+ * <li> {@link TableKey}s that do not share the same Primary Key may be hashed to different Key-Value Table Partitions
+ * and cannot be used for multi-key/entry atomic updates or removals.
+ * <li> {@link TableKey}s sharing the same Primary Key are grouped into the same Table Segment; as such, the choice of
+ * Primary Keys can have performance implications. An ideally balanced Key-Value Table is one where there are no
+ * Secondary Keys or the number of Secondary Keys associated with each Primary Key is approximately the same. To enable
```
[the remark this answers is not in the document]
andreipaduroiu: Rephrased this whole paragraph.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
+ * a uniform distribution of Keys over the Key-Value Table, consider not using Secondary Keys at all. If this situation
+ * cannot be avoided (i.e., multi-entry atomic updates are required), then it is recommended that Primary Keys themselves
+ * be diversified. Such an approaches will ensure that the Key-Value Table load will be spread across all its Table Segments.
```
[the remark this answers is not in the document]
andreipaduroiu: Rephrased.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
+ * An undesirable situation is an extreme case where all the Keys in the  Key-Value Table are associated with a single
+ * Primary Key; in this case the entire Key-Value Table load will be placed on a single backing Table Segment instead of
+ * spreading it across many Table Segments, leading to eventual performance degradation.
```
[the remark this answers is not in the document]
andreipaduroiu: Rephrased.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
- * <li> Unconditional Removals will remove a Key regardless of what that Key's version is. The operation will also
- * succeed (albeit with no effect) if the Key does not exist. Such removals can be performed using
- * {@link #remove(String, Object)}.
- * <li> Conditional Removals will remove a Key only if the specified {@link Version} matches the one that is currently
- * present on the server. Such removals can be performed using {@link #remove(String, Object, Version)}.
+ * is currently present on the server. Conditional inserts can be performed using {@link #put} or {@link #putAll} by passing
+ * a {@link TableEntry} that was created using {@link TableEntry#notExists} and will only succeed if the associated
+ * {@link TableKey} does not already exist. Conditional updates can performed using the same methods, by passing
+ * {@link TableEntry} instances that were either created using {@link TableEntry#versioned} or were retrieved from a
+ * {@link #get}, {@link #getAll} or {@link #entryIterator} call.
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
+ * is. The operation will also succeed (albeit with no effect) if the {@link TableKey} does not exist. Such removals can
+ * be performed using {@link #remove} or {@link #removeAll} by passing a {@link TableKey} created using
+ * {@link TableKey#unversioned}.
+ * <li> Conditional Removals will remove a {@link TableKey} only if the specified {@link Version} matches the one that is
+ * currently present on the server. Such removals can be performed using {@link #remove} or {@link #removeAll} by passing
+ * a {@link TableKey} created using {@link TableKey#versioned} or retrieved from a {@link #get}, {@link #getAll},
+ * {@link #keyIterator} or {@link #entryIterator} call.
  * <li> Multi-key updates allow mixing different types of updates in the same update batch. Some entries may be conditioned
- * on their Keys not existing at all ({@link TableEntry#getKey()}{@link TableKey#getVersion()} equals
+ * on their {@link TableKey}s not existing at all ({@link TableEntry#getKey()}{@link TableKey#getVersion()} equals
```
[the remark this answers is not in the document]
andreipaduroiu: Reorganized this paragraph.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
  * {@link Version#NOT_EXISTS}), some may be conditioned on specific versions and some may not have condition attached at
  * all ({@link TableEntry#getKey()}{@link TableKey#getVersion()} equals {@link Version#NO_VERSION}). All the conditions
  * that are present in the update batch must be satisfied in order for the update batch to be accepted - the condition
- * checks and updates are performed atomically. Use {@link #replaceAll(String, Iterable)} for such an update.
+ * checks and updates are performed atomically. Use {@link #putAll(Iterable)} for such an update.
  * <li> Multi-key removals allow mixing different types of removals in the same removal batch. Some removals may be
```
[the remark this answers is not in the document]
andreipaduroiu: Reorganized this paragraph.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,64 @@
...
- * Key-Value Table, it is highly recommended not to use Key Families at all. If this situation cannot be avoided (i.e.,
- * multi-entry atomic updates or iterators are required), then it is recommended that Key Families themselves be
- * diversified and Keys be equally distributed across them. Such approaches will ensure that the Key-Value Table load
- * will be spread across all its Table Segments. An undesirable situation is an extreme case where all the Keys in the
- * Key-Value Table are associated with a single Key Family; in this case the entire Key-Value Table load will be placed
- * on a single backing Table Segment instead of spreading it across many Table Segments, leading to eventual performance
- * degradation.
+ * <li> The Primary Key is mandatory and is used to group all {@link TableKey}s with the same Primary Key together in the
+ * same Table Partition; this allows multiple keys/entries sharing the same Primary Key to be updated/removed atomically.
+ * <li> The Secondary Key is optional.
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -98,197 +102,147 @@
...
-    /**
-     * The maximum serialization length of a Table Segment Key.
-     */
-    int MAXIMUM_SERIALIZED_KEY_LENGTH = TableSegment.MAXIMUM_KEY_LENGTH / 2;
-
+public interface KeyValueTable extends AutoCloseable {
     /**
      * The maximum serialized length of a Table Segment Value.
      */
     int MAXIMUM_SERIALIZED_VALUE_LENGTH = TableSegment.MAXIMUM_VALUE_LENGTH;
```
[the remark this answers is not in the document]
andreipaduroiu: Good catch. Removed "serialized" from name and doc.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: From user's feedback, that was a rather complicated API that caused more confusion than it helped. 

So a few things were done to this API to simplify it:
- Remove the parametrized types and exposing `ByteBuffer` for keys and values.
- Coalesced redundant APIs (like the ones you mentioned - inspired from Java's Map).

This reduces the API to a bare-bones one which allows for greatest flexibility (from a user's POV). In a subsequent PR/release (not yet planned), it should be trivial to add a thin layer on top of this API for those users that do want a more familiar API and/or parametrized types. Translating from a Map-style API to this one is trivial and can be done at a later time without needing any changes underneath.

client/src/main/java/io/pravega/client/tables/KeyValueTableConfiguration.java
andreipaduroiu: We want them to be explicitly set by the user. When the Controller portion is done, it will have validation for these values, both on the client side and Controller side.

client/src/main/java/io/pravega/client/tables/TableEntry.java
andreipaduroiu: This is used for conditional inserts: condition the update on the predicate that the key must not exist. So: insert <key, value> if key does not already exist. 

If the Javadoc for this one isn't clear, I can change that - let me know.

client/src/main/java/io/pravega/client/KeyValueTableFactory.java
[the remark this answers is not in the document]
tkaitchuck: I think the problem is implied by the name of this interface. 
All of the others are: "Batch**Client**Factory", "ByteStream**Client**Factory", "Synchronizer**Client**Factory"
So when they have methods with "create" such as "createRevisionedStreamClient" the meaning and implication is clear. 

If this were a "KeyValueTableClientFactory" and it had a method "createKeyValueTableClient" there would be no confusion. It is rather long. We could try to shorten it by either dropping "table" or abbreviating "KeyValue" as "KV". I don't have a strong opinion either way.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
tkaitchuck: I don't see the change.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
tkaitchuck: Same here. Do you need to push?

client/src/main/java/io/pravega/client/KeyValueTableFactory.java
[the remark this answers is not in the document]
andreipaduroiu: I don't have a strong preference either. I created #5978 to track this and get it done later - I don't want to delay this PR over this pre-existing condition.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
andreipaduroiu: It is there - did you look at the file directly? This comment is marked 'outdated' by GitHub, so it must have picked up my updates from last week.

client/src/main/java/io/pravega/client/tables/KeyValueTableConfiguration.java
andreipaduroiu: Fixed

client/src/main/java/io/pravega/client/tables/TableEntry.java
fpj: Would either `exclusive` or `absent` be better?

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -98,197 +122,147 @@
...
-     * </ul>
-     */
-    CompletableFuture<Version> putIfAbsent(@Nullable String keyFamily, @NonNull KeyT key, @NonNull ValueT value);
-
-    /**
-     * Unconditionally inserts new or updates existing {@link TableEntry} instances that belong to the same Key Family
+     * Inserts new or updates existing {@link TableEntry} instances that have the same {@link TableKey#getPrimaryKey()}
      * into this {@link KeyValueTable}. All changes are performed atomically (either all or none will be accepted).
+     * <p>
+     * See {@link #put(TableEntry)} for details on Conditional/Unconditional Updates or Insertions. Conditional and
```
fpj, line 162: Ok, so to confirm, even if there is only a single conditional update in a set of many updates and the condition fails, the whole batch fails, including the unconditional ones, yes?

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -98,197 +122,147 @@
...
-     * Unconditionally removes a Key from this Table Segment. If the Key does not exist, no action will be taken.
+     * Removes a {@link TableKey} from the {@link KeyValueTable}.
+     * <p>
+     * If {@link TableKey#isVersioned()} returns true, this will be a Conditional Removal. The {@link TableKey} will only
+     * be removed if there exists a {@link TableKey} in the {@link KeyValueTable} matching the given {@code key} and {@link Version}.
+     * <p>
+     * Otherwise, it will be executed as an Unconditional Removal ({@link TableKey#getVersion()} returns
+     * {@link Version#NO_VERSION} - such as when the key was created using {@link TableKey#unversioned}).
+     * <p>
+     * NOTE: it is possible to invoke this method with a {@link TableKey} created using {@link TableKey#notExists}. This
```
fpj, line 183: We have an `exists` call below, so detecting whether a key exists should really be done via that call. I'm wondering if we should simply not allow `notExists` with remove.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -98,197 +122,147 @@
...
      *
-     * @param keyFamily     The Key Family for which to iterate over keys.
-     * @param maxKeysAtOnce The maximum number of {@link TableKey}s to return with each call to
-     *                      {@link AsyncIterator#getNext()}.
-     * @param state         (Optional) An {@link IteratorState} that represents a continuation token that can be used to
-     *                      resume a previously interrupted iteration. This can be obtained by invoking
-     *                      {@link IteratorItem#getState()}. A null value will create an iterator that lists all keys.
-     * @return An {@link AsyncIterator} that can be used to iterate over all the Keys in this {@link KeyValueTable} that
-     * belong to a specific Key Family.
+     * @param maxKeysAtOnce The maximum number of {@link TableKey}s to return with each call to {@link AsyncIterator#getNext()}.
```
fpj, line 249: I know this as pagination, and I'm wondering if we should call it that way.

client/src/main/java/io/pravega/client/tables/TableKey.java
```diff
@@ -23,71 +24,116 @@
...
 @ToString
-public class TableKey<KeyT> {
+public class TableKey {
     /**
-     * The Key.
+     * The Primary Key.
      *
-     * @param key The Key.
-     * @return The Key.
+     * @param primaryKey Primary Key..
```
fpj, line 35: Typo.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
- * will have no effect on that Key if it doesn't exist but it will  prevent the rest of the removals from the same batch
- * from being applied - this can be used in scenarios where a set of Keys must be removed only if a particular Key is not
- * present. All the conditions that are present in the removal batch must be satisfied in order for the removal batch to
- * be accepted - the condition checks and updates are performed atomically. Use {@link #removeAll(String, Iterable)} for
- * such a removal.
+ * is currently present on the server. Conditional updates can performed using {@link #put} or {@link #putAll} by passing
+ * {@link TableEntry} instances that were either created using {@link TableEntry#versioned} or were retrieved from a
+ * {@link #get}, {@link #getAll} or {@link #entryIterator} call. Conditional inserts can be performed using the same
+ * methods by passing a {@link TableEntry} that was created using {@link TableEntry#notExists} and will only succeed if
+ * the associated {@link TableKey} does not already exist.
```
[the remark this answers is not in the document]
fpj: This comment seems interesting, and I'd like to understand more concretely the concern. About this specifically:

> IE: We can't write in the javadoc for getValue that it returns the value associated with the key at the version

At least in the javadoc of `get` below, it says that it returns the latest and version is ignored, so it is not really to provide the ability of reading different versions of the data, but only to provide the conditional update functionality.

fpj (changes requested):

client/src/main/java/io/pravega/client/tables/TableEntry.java
andreipaduroiu: I like `absent`. I'll take this one step further and rename `unversioned` to `anyVersion`. We'll have:
- `TableEntry.anyVersion`
- `TableEntry.versioned`
- `TableEntry.absent`

client/src/main/java/io/pravega/client/tables/IteratorArgs.java
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.pravega.client.tables;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.nio.ByteBuffer;
+import lombok.NonNull;
+
+public interface IteratorArgs {
```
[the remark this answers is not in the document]
andreipaduroiu: Added Javadocs throughout this file.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
+ * same Table Partition; this allows multiple keys/entries sharing the same Primary Key to be updated/removed atomically.
+ * The Primary Keys must all have the same length throughout the Key-Value Table. This needs to be configured when the
+ * Key-Value Table is created (See {@link KeyValueTableConfiguration#getPrimaryKeyLength()} and must be greater than 0.
+ *
+ * <li> The Secondary Key is optional. The Secondary Keys must all have the same length throughout the Key-Value Table.
+ * This needs to be configured when the Key-Value Table is created (See {@link KeyValueTableConfiguration#getSecondaryKeyLength()}
+ * and must be at least 0. For Key-Value Tables with no Secondary Keys, set this value to 0.
+ *
+ * <li> The Primary Keys must all have the same length throughout the Key-Value Table. This needs to be configured when
+ * the Key-Value Table is created (See {@link KeyValueTableConfiguration#getPrimaryKeyLength()}.
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
+ *
+ * <li> Two {@link TableKey} instances are considered equal if they have the same Primary Key and same Secondary Key.
+ * The same Secondary Key may be "associated" with different Primary Keys, in which case the resulting {@link TableKey}s
+ * will be distinct. For example, consider Primary Keys PK1 != PK2 and Secondary Key SK1; key PK1.SK1 is different from
+ * PK2.SK1 and both are different from PK1 and PK2 (with no secondary key).
```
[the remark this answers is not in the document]
andreipaduroiu: That's a good point. With this scheme, we won't be able to mix-and-match. I've updated the doc to indicate that.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
+ * <li> {@link TableKey}s that do not share the same Primary Key may be hashed to different Key-Value Table Partitions
+ * and cannot be used for multi-key/entry atomic updates or removals.
+ *
+ * <li> {@link TableKey}s sharing the same Primary Key are grouped into the same Table Segment; as such, the choice of
+ * Primary Keys can have performance implications. An ideally balanced Key-Value Table is one where there the load on each
+ * Table Segment is approximately the same; this can be achieved by not having any Secondary Keys at all or where each
+ * Primary Key has approximately the same number of Secondary Keys. To enable a uniform distribution of Keys over the
+ * Key-Value Table, consider not using Secondary Keys at all. If this situation cannot be avoided (i.e., multi-entry
+ * atomic updates are required), then it is recommended to use a sufficiently large number of Primary Keys compared to
+ * the number of Partitions configured on the Key-Value Table (the higher the ratio of Primary Keys to Partitions, the
```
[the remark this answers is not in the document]
andreipaduroiu: Both are valid points. @fpj or @tkaitchuck  do you have any suggestions on how I could improve the wording here, if at all?

client/src/main/java/io/pravega/client/tables/TableEntry.java
```diff
@@ -23,72 +24,110 @@
...
+    /**
+     * Creates a new {@link TableEntry} with a Key made of only a Primary Key, with no specific version.
+     *
+     * @param primaryKey The Entry Key (Primary Key only).
+     * @param value      The Entry Value.
+     * @return An unversioned {@link TableEntry} (version set to {@link Version#NO_VERSION}).
+     */
+    public static TableEntry unversioned(ByteBuffer primaryKey, ByteBuffer value) {
+        return unversioned(primaryKey, null, value);
+    }
```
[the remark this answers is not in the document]
andreipaduroiu: Yes, I have considered that. However I can't think of an elegant way of dealing with it. The behavior of the KVT is dependent on the configuration that the user specified at runtime (when creating it). We can't force the user to choose different APIs based on whether they chose a secondary key or not, just like we can't fix key-length problems at compile time. 

This will be a runtime error and the request will be rejected with the appropriate error message in this case. I believe this is acceptable.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
andreipaduroiu: @tkaitchuck I prefer API simplicity over having different classes that have different purposes which make interoperability hard. 

An example of one way that this API makes things easier is multi-key conditional updates. 

Suppose we do this:
- Get Entry E1 {Key1, Value1, Version1}
- Insert/Update Entry E2 {Key2, Value2) but **only if** E1 hasn't changed.

We'd do this like (pseudocode, not actual API):
```
   TableEntry e1 = kvt.get(k1).join();
   TableEntry e2 = ...;
   kvt.putAll(e1, e2).join();
```

The `putAll` will condition the update on `e1` being unchanged. If we had different classes to represent different things, the user would have to write additional code to convert from one to another.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
- * present. All the conditions that are present in the removal batch must be satisfied in order for the removal batch to
- * be accepted - the condition checks and updates are performed atomically. Use {@link #removeAll(String, Iterable)} for
- * such a removal.
+ * is currently present on the server. Conditional updates can performed using {@link #put} or {@link #putAll} by passing
+ * {@link TableEntry} instances that were either created using {@link TableEntry#versioned} or were retrieved from a
+ * {@link #get}, {@link #getAll} or {@link #entryIterator} call. Conditional inserts can be performed using the same
+ * methods by passing a {@link TableEntry} that was created using {@link TableEntry#notExists} and will only succeed if
+ * the associated {@link TableKey} does not already exist.
+ *
+ * <li> Unconditional Removals will remove a {@link TableKey} regardless of what that {@link TableKey}'s {@link Version}
```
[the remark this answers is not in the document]
andreipaduroiu: Rephrased.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
+ * is. Such removals can be performed using {@link #remove} or {@link #removeAll} by passing a {@link TableKey} created
+ * using {@link TableKey#unversioned}. The operation will also succeed (albeit with no effect) if the {@link TableKey}
+ * does not exist.
+ *
+ * <li> Conditional Removals will remove a {@link TableKey} only if the specified {@link Version} matches the one that is
+ * currently present on the server. Such removals can be performed using {@link #remove} or {@link #removeAll} by passing
+ * a {@link TableKey} created using {@link TableKey#versioned} or retrieved from a {@link #get}, {@link #getAll},
+ * {@link #keyIterator} or {@link #entryIterator} call.
+ *
+ * <li> Multi-key updates allow mixing different types of updates in the same update batch. All the conditions that are
```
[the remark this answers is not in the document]
andreipaduroiu: You can't. I added another bullet to indicate this.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
+ * present in the update batch must be satisfied in order for the update batch to be accepted - the condition checks and
+ * updates are performed atomically. Some entries may be conditioned on their {@link TableKey}s not existing at all
+ * ({@link TableEntry#getKey()}{@link TableKey#getVersion()} equals {@link Version#NOT_EXISTS}), some may be conditioned
+ * on specific versions and some may not have condition attached at all ({@link TableEntry#getKey()}{@link TableKey#getVersion()}
+ * equals {@link Version#NO_VERSION}). Use {@link #putAll(Iterable)} for such an update.
+ *
+ * <li> Multi-key removals allow mixing different types of removals in the same removal batch. All the conditions that
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -98,197 +122,147 @@
...
-     * Unconditionally inserts a new or updates an existing Entry in the {@link KeyValueTable}.
+     * Inserts a new or updates an existing {@link TableEntry} in the {@link KeyValueTable}.
+     * <p>
+     * If {@link TableEntry#isVersioned()} returns true, this will be a:
+     * <ul>
+     * <li> Conditional Insert if the {@code entry}'s {@link TableEntry#getKey()} {@link TableKey#getVersion()} equals
+     * {@link Version#NOT_EXISTS} (i.e., if the {@link TableEntry} was created using {@link TableEntry#notExists}).
+     * The {@link TableEntry} will only be inserted if there doesn't exist a {@link TableKey} in the {@link KeyValueTable}
+     * matching the given {@code entry}'s {@link TableEntry#getKey()}).
+     * <li> Conditional Update if the {@code entry}'s {@link TableEntry#getKey()} {@link TableKey#getVersion()} does not
```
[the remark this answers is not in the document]
andreipaduroiu: I rephrased the Javadoc for both these methods.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
```diff
@@ -34,58 +32,84 @@
...
+ * Key-Value Table is created (See {@link KeyValueTableConfiguration#getPrimaryKeyLength()} and must be greater than 0.
+ *
+ * <li> The Secondary Key is optional. The Secondary Keys must all have the same length throughout the Key-Value Table.
+ * This needs to be configured when the Key-Value Table is created (See {@link KeyValueTableConfiguration#getSecondaryKeyLength()}
+ * and must be at least 0. For Key-Value Tables with no Secondary Keys, set this value to 0.
+ *
+ * <li> The Primary Keys must all have the same length throughout the Key-Value Table. This needs to be configured when
+ * the Key-Value Table is created (See {@link KeyValueTableConfiguration#getPrimaryKeyLength()}.
+ *
+ * <li> Two {@link TableKey} instances are considered equal if they have the same Primary Key and same Secondary Key.
```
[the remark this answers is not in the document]
andreipaduroiu: I updated this paragraph (including some bullets above). I clarified that Secondary Keys are optional at the table level, and not at the key level. If configured at the table level, then all keys must have Secondary Keys; if the KVT is configured with 0-length Secondary keys (i.e., no such secondary keys), then all keys are made just of their Primary Keys.

This gives flexibility to the user to choose how their Table will be used. Do they want to use it as a plain KV store where there isn't any relation between the keys? If so, then 0-length SKs make sense. Do they want to "relate" some keys to each other and enable multi-key atomic updates? Then configure Secondary Keys. Do they want both? Then the user can provide a "null" Secondary Key - a special value that the user knows is null.

This API is really low-level and I intended it to be this way. We can build fancier APIs on top of it that cater to specific needs. Those APIs can be added later without having to change the base implementation. Some examples may include parametrized types, Map-like APIs (we can even implement `java.collections.Map` on top of this), etc.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: Yes. the whole thing is rejected with no side effects.

If this text is not clear, I'm open to suggestions how to rephrase it:
```
and will prevent the committal of the entire batch of updates if it fails to meet the condition.
```

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: We could. It will have no effect though. In my opinion, if the operation is harmless, then documenting it as such (so users don't get confused about it) should be sufficient.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: What would sound better as an arg? I thought about a better name, but adding the word "page" implies UI concerns to me. Really, I want the user to be in control of how many items to return at once, for whatever reasons the user may have.

client/src/main/java/io/pravega/client/tables/TableKey.java
andreipaduroiu: Fixed

client/src/main/java/io/pravega/client/tables/TableKey.java
```diff
@@ -15,79 +15,35 @@
...
- *
- * @param <KeyT> Type of the Key.
  */
-@Getter
-@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
-@ToString
-public class TableKey<KeyT> {
+@Data
+@RequiredArgsConstructor
+public class TableKey {
```
tkaitchuck, line 28: Final?

client/src/main/java/io/pravega/client/tables/IteratorArgs.java
```diff
@@ -38,21 +42,76 @@ static IteratorArgs fromBytes(ByteBuffer buffer) {
...
+     * @param fromPrimaryKey A {@link ByteBuffer} that indicates the Primary Key (see {@link TableKey#getPrimaryKey()})
+     *                       to begin the iteration at. If not provided (null), the iterator will begin with the first
+     *                       {@link TableKey}/{@link TableEntry} in the {@link KeyValueTable}. This argument is inclusive.
+     * @param toPrimaryKey   A {@link ByteBuffer} that indicates the Primary Key (see {@link TableKey#getPrimaryKey()})
+     *                       to end the iteration at. If not provided (null), the iterator will end with the last
+     *                       {@link TableKey}/{@link TableEntry} in the {@link KeyValueTable}. This argument is inclusive.
+     * @return A new {@link IteratorArgs} that can be passed to {@link KeyValueTable#keyIterator} or to
+     * {@link KeyValueTable#entryIterator}.
+     */
     static IteratorArgs forRange(@Nullable ByteBuffer fromPrimaryKey, @Nullable ByteBuffer toPrimaryKey) {
```
tkaitchuck, line 99: I am surprised we would support this because we hash the primary keys to obtain the segment, so this can't operate against a subset of the segments even when the range is small. Also even though it's a iterator, this can't possibly do this in order right?

client/src/main/java/io/pravega/client/tables/IteratorArgs.java
```diff
@@ -38,21 +42,76 @@ static IteratorArgs fromBytes(ByteBuffer buffer) {
...
+     * the arguments provided, this may iterate through all {@link TableKey}/{@link TableEntry} instances within the
+     * {@link KeyValueTable} or a sub-range.
+     *
+     * @param primaryKeyPrefix A {@link ByteBuffer} that indicates the Primary Key prefix for all {@link TableKey}/
+     *                         {@link TableEntry} instances to return. If not provided (null), the iterator will list
+     *                         all {@link TableKey}/{@link TableEntry} instances in the {@link KeyValueTable}.
+     * @return A new {@link IteratorArgs} that can be passed to {@link KeyValueTable#keyIterator} or to
+     * {@link KeyValueTable#entryIterator}.
+     */
     static IteratorArgs forPrefix(@Nullable ByteBuffer primaryKeyPrefix) {
```
tkaitchuck, line 115: Same comment as above.

client/src/main/java/io/pravega/client/tables/Put.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Creates a new instance of the {@link Put} class for a conditional update.
+     *
+     * @param key     The {@link TableKey} to update.
+     * @param value   The Value to associate with {@link TableKey}.
+     * @param version (Optional) The {@link Version} to condition the update on. This can be retrieved from
+     *                {@link TableEntry#getVersion()} or it may be {@link Version#NOT_EXISTS} or {@link Version#NO_VERSION}.
+     */
+    public Put(TableKey key, ByteBuffer value, @Nullable Version version) {
```
tkaitchuck, line 49: Version can be both `null` and `NO_Version`. We should probably pick one that we want users to use.

client/src/main/java/io/pravega/client/tables/Remove.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Creates a new instance of the {@link Remove} class for a conditional removal.
+     *
+     * @param key     The {@link TableKey} to remove.
+     * @param version Optional) The {@link Version} to condition the removal on. This can be retrieved from
+     *                {@link TableEntry#getVersion()} or it may be {@link Version#NOT_EXISTS} or {@link Version#NO_VERSION}.
+     */
+    public Remove(TableKey key, @Nullable Version version) {
```
tkaitchuck, line 46: Same

client/src/main/java/io/pravega/client/tables/TableEntry.java
```diff
@@ -16,18 +16,13 @@
...
-import lombok.ToString;
 
 /**
  * A {@link KeyValueTable} Entry.
  */
-@Getter
-@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
-@ToString
+@Data
 public class TableEntry {
```
tkaitchuck, line 26: Final?

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
fpj: From what you are saying, we get very different behavior for the same primitive (i.e., KVT) based on the configuration in `KeyValueTableConfiguration`. `TableKey` is used to express keys independent of choice of using SKs for the the table or not. If I choose to use SKs with a given size `s`, then all `TableKey` instances for that table must have an SK of length `s`, yes? What's the way to determine what a KVT configuration is?

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
fpj: One aspect that stands out to me is that the application is not necessarily free to choose keys as they are part of the applications semantics and the key distribution is driven by the workload. What we can tell the application developer is how to configure a KVT depending on the type of use case, but the only knob I actually see with respect to the distribution of keys is the number of partitions. Choosing a larger number of partitions won't necessarily help with distributing the load more evenly. In fact, the degree of parallelism is upper bounded by the number of segment stores for reads and number of segment containers for writes, but we make no guarantee about how the load is distributed.

The above refers to accesses to individual pairs. The other point I touched upon is how to efficiently process a whole table and in particular what we offer that helps the application split the workload if needed.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: I think this is covered by the commit I pushed on April 30.

client/src/main/java/io/pravega/client/tables/TableEntry.java
andreipaduroiu: I have removed all these methods in the API simplification I pushed on April 30.

client/src/main/java/io/pravega/client/tables/TableEntry.java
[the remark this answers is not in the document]
andreipaduroiu: This method has been removed.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
andreipaduroiu: I removed this method.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
andreipaduroiu: That configuration is defined when the KVT is created and the Controller stores it.

There is an issue (#5939) to make that information accessible from the Controller; the Client will fetch it every time it creates a `KeyValueTableImpl` object and may even choose to expose it as a read-only property (i.e., getter) on it.

client/src/main/java/io/pravega/client/tables/TableKey.java
andreipaduroiu: Fixed.

client/src/main/java/io/pravega/client/tables/IteratorArgs.java
andreipaduroiu: When implemented (not necessarily as part of this initial revamp (release 0.10), this will be a distributed iterator across all partitions. It will issue the request to all table partitions and merge the result on the client-side. Since all Keys are stored in order in each Table Segment, this will be a "simple" merge iterator. The difficulty will be from issuing all those parallel requests and managing them - that's why I am willing to push this to a subsequent update (it's defined in the API, but may be implemented after 0.10). 

This was described in the PDP and #5931 (see last task at the bottom about global iterators).

client/src/main/java/io/pravega/client/tables/IteratorArgs.java
andreipaduroiu: Same answer.

client/src/main/java/io/pravega/client/tables/TableEntry.java
andreipaduroiu: Fixed

client/src/main/java/io/pravega/client/tables/Put.java
andreipaduroiu: I clarified the Javadoc and defaulted this field to `Version.NO_VERSION`.  This makes it consistent with `Insert` which defaults it (internally) to `Version.NOT_EXISTS` as that is a conditional insert.

client/src/main/java/io/pravega/client/tables/Remove.java
andreipaduroiu: Fixed.

tkaitchuck (approved):

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
fpj: I'm not necessarily suggesting to replace it verbatim, but I thought we could say something like:

```
<li> {@link TableKey}s sharing the same Primary Key are grouped into the same Table Segment; as such, the choice of
 * table key can have performance implications. As Primary Keys are hashed to Table Segments, a large number of Primary
 * Keys (larger than the number of Table Segments) and a uniform distribution of operations across keys leads to an even
 * distribution of load across Table Segments. Many realistic workloads are skewed, however, and can lead to an uneven
 * load distribution across Table Segments.
 *
 * The use of Secondary Keys has also implications to load balance. All keys containing a given Primary Key map to the
 * same Table Segment independent of Secondary Key. In the presence of Secondary Keys, the total load for a given
 * Primary Key depends on the load aggregate across all of its Secondary Keys. When the load distribution is highly
 * skewed, one option to consider it eliminating Secondary Keys and only using Primary Keys. The application must be
 * able to encode the key into a single one rather than split into two parts. To illustrate, if we have two Table Keys
 * "A.B" and "A.C", "." representing the separation between Primary and Secondary Keys, then they map to the same Table
 * Segment because they have the same Primary Key "A". If instead, we use "AB" and "AC" as keys, eliminating the Secondary
 * Keys but retaining the necessary information in the Primary Key, then we should be able to map those keys to different
 * Table Segments depending on the hashing.
 
 ```

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
fpj: I'd say `... prevent the entire batch from committing`. I think. `committal` means something else.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
fpj: What about: `iterationSize`, `maxIterationSize`, `iterationPage`, `maxIterationPage`, `iterationLimit`.

andreipaduroiu: I like `maxIterationSize`. Fixed.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
andreipaduroiu: Rephrased.

client/src/main/java/io/pravega/client/tables/KeyValueTable.java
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

PR #5980, opened 2021-04-27
Issue 5935: (KVT Beta2) Changing Segment Attributes from UUIDs to AttributeIds

### LINKED ISSUE #5935: KVT Beta 2: Change Segment Attribute Representation from UUID to a generic AttributeId
Currently all Segment Attributes are 16 bytes so in the code they are represented as UUIDs. With PDP-48, we will have:
- Stream Segments and Hash Table Segments (as they are today - both core and extended attributes are 16 bytes)
- Fixed-Key-Length Table Segments: core attributes are 16 bytes, extended attributes are an arbitrary value set at the creation of the segment.

As such, we need to:
- Create a new `AttributeId` abstraction for the Segment Attributes
- Create an `AttributeId.UUID` implementation that is fully compatible with the current `UUID` API (what we use from it)
- Update all usages of `UUID` in the segment store codebase (where appropriate) to use `AttributeId` (preferred to `AttributeId.UUID` if possible)
- Define (not necessarily fully implement) an `AttributeId.Variable` for arbitrary-sized attributes (these will be used later).

andreipaduroiu:
**Change log description**  
- Changed the in-memory representation of all Segment Attributes from UUID to a more extensible `AttributeId`.
- Added support for variable-length Attribute Ids (no longer 16-byte UUIDs; but must be the same for the whole segment).

**Purpose of the change**  
Fixes #5935.

**What the code does**  
For Fixed-Key-Length Table Segments, we will want to put all Table Entry Keys into the underlying Table Segment's Attributes. For that, we need to make those attributes more flexible, in that they should allow for more than just 16-byte UUID keys. This PR aims to do that. 

Notes:
- Core Attributes are unchanged - they are still 16-byte UUIDs
- Extended Attributes can be either UUIDs (for all stream segments and legacy table segments) or "Variable" (wrapping a byte array).
- All Extended Attribute Ids for a Segment must be of the same type (either UUID or Variable) and must have the same length. This is now enforced in various places in the API to prevent accidental data corruption.
- This change should be fully backward compatible and nothing should change in terms of stored data size, memory data size or performance for stream segments or legacy table segments. 

Following is a summary of the changes in various classes/components:
- `SegmentMetadata`
    - Added a new property named `getAttributeIdLength` which will return the length of the Attribute Ids for all extended attributes within that segment
    - This is fetched from the core segment attribute `ATTRIBUTE_ID_LENGTH` which is set once upon segment creation (and never modified later).
- `AttributeId`
    - Abstraction for a general Attribute Id
    - `AttributeId.UUID` identifies those legacy UUID attributes that are used for stream segments and legacy table segments.
    - `AttributeId.Variable` identifies/wraps those attributes that are will be used for Fixed-Key-Length Table Segments (or any segments that require more flexibility).
- `SegmentAttributeBTreeIndex.java`
    - This is the adapter class that writes to and reads from the segment attribute index file. While it could already handle an arbitrary key size, this has been modified to select that key size based on the `SegmentMetadata.getAttributeIdLength`.
- `AttributeUpdateCollection.java`
    - This is a new class which is meant to replace `Collection<AttributeUpdate>` everywhere those were used.
    - It segregates attributes based on their type and performs some basic validation. This class is meant to simplify and/or make some serialization operations more efficient.
- `UpdateAttributesOperation.java`
    -   Changed the `Serializer` by adding Revision 1 (to Version 0). In Revision 0, we only serialize those UUID attributes, while in Revision 1 we serialize only the Variable Attributes. This change should be fully compatible with older code.
- `StreamSegmentAppendOperation.java`
    - Changed the `Serializer` by adding a whole new Version 1 for serialization and slightly altering Version 0.
    - This was done on purpose - we 1) want to make this backwards compatible and 2) absolutely do not want to add any sort of overhead for stream segment appends.
    - Version 0 serializer has been "frozen" - it is kept there solely for testing backwards compatibility purposes.
    - Version 1 serializes everything in a format compatible with Version 0 - so whatever it writes can be read back with Version 0. 
    - For attributes, Version 1 first serializes UUID attributes and then Variable attributes. A Version 0 deserializer (old code) will only read the UUID attributes and ignore the rest.
    - This class' unit tests have been enhanced with thorough validation of compatibility between these serializers.
    - In the future (i.e., 2 releases from whenever this is merged in), we should drop support for Version 0 and begin writing with version 1 - Version 1 also writes 8 bytes less per attribute, which should shave off 16 bytes at least for each append).
- All other changes (the majority of modified LOCs) are:
    - Unit tests for the above
    - Replacing `UUID` with `AttributeId` where appropriate.
    - Fallout from having to modify the `StreamSegmentStore` API to use `AttributeUpdateCollection` in lieu of the previous `Collection<AttributeUpdate>`.

**How to verify it**  
New tests added. Build must pass.
These classes require extra care during the review:
- `StreamSegmentAppendOperation.java` and `UpdateAttributesOperation.java` - the serialization for these operations has changed (see above).


RaulGracia (approved):
LGTM, but to be on the safe side, it would be great to test cluster upgrades from 0.9 to the branch with KVT Beta2 (once all PRs are in) to validate that there are no issues related with this change in attributes.

PR #5323, opened 2020-11-13
Issue 5098: Health Check (Initial Implementation)

co-jo:
**Change log description**  

Add new shared project to enable health checking functionalities.

See: https://github.com/co-jo/pravega/wiki/PDP-45-(Pravega-Health-Check) for revised PDP.

**Purpose of the change**  
Makes progress towards #5098 . 

**What the code does**  

...

**How to verify it**  
(Optional: steps to verify that the changes are effective)

PR #5991, opened 2021-05-03
Issue 5098: Health Check Framework (1/4) 

co-jo:
**Change log description**  

* Creates a new shared project `shared:health` which implements (most) functionality outlined in this PDP: https://github.com/co-jo/pravega/wiki/PDP-45-(Pravega-Health-Check)

**Purpose of the change**  

* Implement the base framework which other clients will use to perform various health related checks.

**What the code does**  

Implements the Health Check framework. See PDP for further details.

**Extra**

This PR is a subset of a previous PR (#5323) that covers just the health check project. The purpose of this was to make the contents of the aforementioned PR more cohesive. This PR should also mark the first in a serious of PRs that will cover the following:

1. Refactoring out common REST server classes from the `controller` project into some shared project `shared:rest` [(2/4)](https://github.com/pravega/pravega/pull/5429).
2. Integrating the `controller` with the health check framework [(3/4)](https://github.com/co-jo/pravega/pull/2/files).
    * This will also cover security concerns. Should be able to hook into existing security classes.
    * As I do not have permissions to create a branch within this repo, I created a [PR](https://github.com/co-jo/pravega/pull/2/files) on my fork (into this branch) to isolate these changes.
3. Integration the `segmentstore` with the health check framework [(4/4)](https://github.com/co-jo/pravega/pull/3).
    * This will also cover security concerns.

**How to verify it**  


shared/health/src/main/java/io/pravega/shared/health/impl/CompositeHealthContributor.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * The {@link StatusAggregator} used to perform the aggregation of all the {@link HealthContributor} dependencies.
+     */
+    @Getter
+    private final StatusAggregator aggregator;
+
+    private final ContributorRegistry registry;
+
+    private final Collection<HealthContributor> contributors = ImmutableSet.of();
```
andreipaduroiu, line 54: I don't get how this is useful. You are assigning an empty immutable list to this field - how is it used afterwards?

shared/health/src/main/java/io/pravega/shared/health/impl/CompositeHealthContributor.java
```diff
@@ -0,0 +1,121 @@
...
+        Health.HealthBuilder builder = Health.builder().name(getName());
+        Collection<Status> statuses = new ArrayList<>();
+        // Fetch the Health Status of all dependencies.
+        val children =  contributors().stream()
+                .filter(Objects::nonNull)
+                .map(contributor -> {
+                    Health health = contributor.getHealthSnapshot(includeDetails);
+                    statuses.add(health.getStatus());
+                    if (health.getStatus() == Status.UNKNOWN) {
+                        log.warn("{} has a Status of 'UNKNOWN'. This indicates `doHealthCheck` does not set a status" +
```
andreipaduroiu, line 89: Would this be logged every time we call `getHealthSnapshot`? I really do not like dumping log warnings every time we invoke a method like this. Is there another way to handle this?

shared/health/src/main/java/io/pravega/shared/health/impl/CompositeHealthContributor.java
```diff
@@ -0,0 +1,121 @@
...
+     * while {@link HealthComponent} will provide the {@link Health} results for all it's descendants.
+     *
+     * @param includeDetails Whether or not to request details from each of the {@link HealthContributor}.
+     * @return The {@link Health} result of the {@link CompositeHealthContributor}.
+     */
+    public Health getHealthSnapshot(boolean includeDetails) {
+        Health.HealthBuilder builder = Health.builder().name(getName());
+        Collection<Status> statuses = new ArrayList<>();
+        // Fetch the Health Status of all dependencies.
+        val children =  contributors().stream()
```
andreipaduroiu, line 83: Why do you generate `children` if in the last line of this method you may not include it if `includeDetails` is false?

shared/health/src/main/java/io/pravega/shared/health/impl/CompositeHealthContributor.java
```diff
@@ -0,0 +1,121 @@
...
+     * {@link HealthContributor} objects to perform the aggregate health check on. This is helpful because it gives us
+     * flexibility in defining where the contributors may be but also avoids the requirement of being bound to a
+     * {@link ContributorRegistry} instance.
+     *
+     * Currently only registrations take place via the {@link ContributorRegistry}, so the in the case none is defined,
+     * an empty {@link ImmutableSet} will *always* be returned as an alternative.
+     *
+     * @return The {@link Collection} of {@link HealthContributor}.
+     */
+    public Collection<HealthContributor> contributors() {
```
andreipaduroiu, line 112: Please add `@Override` to all overridden/implemented methods.

shared/health/src/main/java/io/pravega/shared/health/impl/ContributorRegistryImpl.java
```diff
@@ -0,0 +1,321 @@
...
+     * @param parent The {@link HealthComponent} to add 'contributor' as a child/dependency.
+     * @return The {@link HealthContributor} registered under the {@link ContributorRegistry}.
+     */
+    @NonNull
+    @Override
+    synchronized public HealthContributor register(HealthComponent component, HealthComponent parent) {
+        parent = Objects.isNull(parent) ? root : parent;
+        // A HealthComponent should only exist if defined during construction, instead of adding it dynamically.
+        if (!components.contains(parent.getName()) || !contributors.containsKey(parent.getName())) {
+            log.warn("Attempting to register {} under unrecognized {} -- aborting.", component, parent);
```
andreipaduroiu, line 129: Is this log warn necessary? I think throwing an exception would be better.

Would the caller expect a `null` as a return value?

shared/health/src/main/java/io/pravega/shared/health/impl/ContributorRegistryImpl.java
```diff
@@ -0,0 +1,321 @@
...
+    /**
+     * Registers a {@link HealthContributor} under the specified parent {@link HealthComponent}.
+     *
+     * @param contributor The {@link HealthContributor} object to add to the registry as a child of 'parent'.
+     * @param parent The {@link HealthComponent} to add 'contributor' as a child/dependency.
+     * @return The {@link HealthContributor} registered under the {@link ContributorRegistry}.
+     */
+    @NonNull
+    synchronized public HealthContributor register(HealthContributor contributor, HealthComponent parent) {
+        parent = Objects.isNull(parent) ? root : parent;
```
andreipaduroiu, line 145: == null

shared/health/src/main/java/io/pravega/shared/health/impl/ContributorRegistryImpl.java
```diff
@@ -0,0 +1,321 @@
...
+            log.warn("{} has already been registered.", contributors.get(contributorName));
+            return contributors.get(contributorName);
+        }
+        log.debug("Registering {} to {}.", contributor, contributors.get(componentName));
+        contributors.put(contributorName, contributor);
+        // A new contributor should have no child relations to overwrite.
+        // Realistically, this should be restricted because it implies non HealthComponent contributors can have children.
+        children.put(contributorName, new HashSet<>());
+        // Add the child relation.
+        children.get(componentName).add(contributor);
```
andreipaduroiu, line 175: Combine this and the above line into 1. Create a map, add something to it, then put the map in the children collection.

shared/health/src/main/java/io/pravega/shared/health/impl/ContributorRegistryImpl.java
```diff
@@ -0,0 +1,321 @@
...
+        // 'contributor' should not have any existing parents when registered.
+        parents.put(contributorName, new HashSet<>());
+        // Add the parent relation.
+        parents.get(contributorName).add(contributors.get(componentName));
```
andreipaduroiu, line 179: Same here

shared/health/src/main/java/io/pravega/shared/health/impl/ContributorRegistryImpl.java
```diff
@@ -0,0 +1,321 @@
...
+        for (HealthContributor parent : parents.get(name)) {
+            children.get(parent.getName()).remove(contributor);
+        }
+        // Remove all child -> {name} relations.
+        for (HealthContributor child : children.get(name)) {
+            Collection<HealthContributor> from = parents.get(child.getName());
+            from.remove(contributor);
+            // Validate that this contributor is still reachable.
+            if (from.isEmpty()) {
+                log.debug("> {} removal caused {} to become unreachable.", contributor, child);
```
andreipaduroiu, line 232: Why is this debug log needed?

shared/health/src/main/java/io/pravega/shared/health/impl/HealthComponent.java
```diff
@@ -0,0 +1,65 @@
...
+    private final String name;
+
+    /**
+     * Creates a new instance of the {@link HealthComponent} class.
+     * @param name The name which this object is addressable by.
+     * @param aggregator The {@link StatusAggregator} which defines the aggregation logic.
+     * @param registry The {@link ContributorRegistry} which to register this component under.
+     */
+    @NonNull
+    protected HealthComponent(String name, StatusAggregator aggregator, ContributorRegistry registry) {
```
andreipaduroiu, line 55: validate args

shared/health/src/main/java/io/pravega/shared/health/impl/HealthConfigImpl.java
```diff
@@ -0,0 +1,188 @@
...
+@Slf4j
+public class HealthConfigImpl implements HealthConfig {
+
+    private final Map<String, HealthComponent> components;
+
+    private final Set<HealthComponent> roots;
+
+    private final Map<String, Set<String>> relations;
+
+    private HealthConfigImpl(Map<String, HealthComponent> components, Set<HealthComponent> roots, Map<String, Set<String>> relations) {
```
andreipaduroiu, line 43: validate args?

shared/health/src/main/java/io/pravega/shared/health/impl/HealthConfigImpl.java
```diff
@@ -0,0 +1,188 @@
...
+        /**
+         * Defines a new {@link HealthComponent} to be created.
+         * @param name The name of the to-be {@link HealthComponent}.
+         * @param aggregator The aggregation rule to use.
+         *
+         * @return A reference to the *same* object, with updates to its internal data-structures made.
+         */
+        public Builder define(String name, StatusAggregator aggregator) {
+            if (components.containsKey(name)) {
+                log.warn("Overwriting a pre-existing component definition -- aborting.");
```
andreipaduroiu, line 70: Please reconsider this warn message.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthConfigImpl.java
```diff
@@ -0,0 +1,188 @@
...
+         * Defines that there should be a parent-child relation between a {@link HealthComponent} called 'child'
+         * and a {@link HealthComponent} called 'parent'.
+         *
+         * @param child The name of the child {@link HealthComponent}.
+         * @param parent The name of the parent {@link HealthComponent}.
+         * @return A reference to the *same* object, with updates to its internal data-structures made.
+         */
+        public Builder relation(String child, String parent) {
+            if (!components.containsKey(child) || !components.containsKey(parent)) {
+                log.warn("At least one of the components in this relation has not been defined -- aborting.");
```
andreipaduroiu, line 90: and here. Best if you throw an exception. Nobody will really look at logs except when debugging some production issue, at which point log noise will be really, really distracting.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthConfigImpl.java
```diff
@@ -0,0 +1,188 @@
...
+         * @return Whether or not the specification is cycle-free.
+         * @throws Exception
+         */
+        private boolean isValid() {
+            boolean cycle = false;
+            HashMap<String, Boolean> visited = new HashMap<>();
+            for (val component : relations.entrySet()) {
+                cycle |= validate(component.getKey(), visited);
+                if (cycle) {
+                    break;
```
andreipaduroiu, line 110: `return false`

then `return true` at the end, and so you can get rid of the `cycle` variable.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthConfigImpl.java
```diff
@@ -0,0 +1,188 @@
...
+         * @param name The name of the {@link HealthComponent} currently being checked.
+         * @param visited The list of {@link HealthComponent} visited so far.
+         *
+         * @return Whether or not this component has been searched previously and is cycle-free.
+         */
+        private boolean validate(String name, Map<String, Boolean> visited) {
+            if (visited.containsKey(name) && visited.get(name)) {
+                return true;
+            }
+            boolean cycle = false;
```
andreipaduroiu, line 129: See if you can do the same here too

shared/health/src/main/java/io/pravega/shared/health/impl/HealthConfigImpl.java
```diff
@@ -0,0 +1,188 @@
...
+            }
+            return cycle;
+        }
+
+        public HealthConfig empty() {
+            HealthConfig config = null;
+            try {
+                config = build();
+            } catch (Exception e) {
+                log.error("Error building empty HealthConfig.", e);
```
andreipaduroiu, line 145: so ... will you just return `null` in this case? I don't understand how this is useful.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
```diff
@@ -0,0 +1,129 @@
...
+@Slf4j
+public class HealthEndpointImpl implements HealthEndpoint {
+
+    private final ContributorRegistry registry;
+
+    /**
+     * Creates a new instance of the {@link HealthEndpointImpl} class.
+     * @param registry The {@link ContributorRegistry} to query.
+     */
+    HealthEndpointImpl(ContributorRegistry registry) {
```
andreipaduroiu, line 41: Tip: 1 use `@RequiredArgsConstructor` on the class, which will generate this for you.
TIp 2: Use `@NotNull` on the field in conjunction with Tip 1 above.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param id  The id/name of the {@link HealthComponent} to check the {@link Health} of.
+     * @param includeDetails Whether or not to include detailed information provided by said {@link HealthComponent}.
+     * @return The {@link Health} result of the {@link HealthContributor} with name 'id'.
+     */
+    @NonNull
+    @Override
+    public Health getHealth(String id, boolean includeDetails) {
+        HealthContributor result = registry.get(id);
+        if (Objects.isNull(result)) {
```
andreipaduroiu, line 65: == null

But why not just return null??? You can define in the contract that "if not registered, it will return null".

shared/health/src/main/java/io/pravega/shared/health/impl/HealthServiceUpdaterImpl.java
```diff
@@ -0,0 +1,126 @@
...
+    public void close() {
+        if (isRunning()) {
+            shutDown();
+        }
+        if (state() == State.RUNNING) {
+            Futures.await(Services.stopAsync(this, this.executorService));
+        }
+        log.info("Stopping ScheduledExecutorService.");
+        try {
+            executorService.shutdown();
```
andreipaduroiu, line 120: `ExecutorServiceHelpers.shutdown` properly shuts down the executor service

shared/health/src/test/java/io/pravega/shared/health/HealthServiceTests.java
```diff
@@ -0,0 +1,240 @@
...
+                service.components().size());
+        // Except its root.
+        Assert.assertEquals("The ContributorRegistry should not maintain any references to HealthContributors",
+                1,
+                service.registry().contributors().size());
+    }
+
+    @Before
+    @SneakyThrows
+    public void before() {
```
andreipaduroiu, line 72: Why not put the contents of `start` here? and that of `stop` in `after`?

shared/health/src/test/java/io/pravega/shared/health/HealthServiceUpdaterTests.java
```diff
@@ -0,0 +1,81 @@
...
+        service = factory.createHealthService("health-service", true);
+        healthServiceUpdater = service.getHealthServiceUpdater();
+        healthServiceUpdater.awaitRunning();
+    }
+
+    @After
+    public void after() {
+        factory.close();
+        service.close();
+        healthServiceUpdater.stopAsync();
```
andreipaduroiu, line 53: does this need a `.close()` too?

shared/health/src/test/java/io/pravega/shared/health/HealthServiceUpdaterTests.java
```diff
@@ -0,0 +1,81 @@
...
+        assertHealthServiceStatus(Status.DOWN);
+    }
+
+    private void assertHealthServiceStatus(Status expected) {
+        try {
+            AssertExtensions.assertEventuallyEquals(expected,
+                    () -> healthServiceUpdater.getLatestHealth().getStatus(),
+                    (healthServiceUpdater.getInterval() + 1) * 1000);
+        } catch (Exception e) {
+            Assert.fail(e.getMessage());
```
andreipaduroiu, line 78: ??? 

Why not bubble the exception up?

shared/health/src/test/java/io/pravega/shared/health/StatusAggregatorTests.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class StatusAggregatorTests {
+
+    List<Status> statuses = Arrays.asList(Status.UP, Status.UP, Status.DOWN);
```
andreipaduroiu, line 28: private static final

shared/health/src/test/java/io/pravega/shared/health/impl/ContributorRegistryTests.java
```diff
@@ -0,0 +1,386 @@
...
+     *        ├─ TWO - HealthIndicator
+     *        ├─ THREE - HealthIndicator
+     *
+     * NOTE: This was constructed to ensure validity, but this should not be done in practice. This loop hole allows one to
+     * introduce cycles into this graph, which should be *strictly* forbidden.
+     */
+    @Test
+    public void testUnregisterInternalContributor() {
+        // Create the contributors to register.
+        ArrayList<HealthContributor> contributors = new ArrayList<>(Arrays.asList(
```
andreipaduroiu, line 139: Why do you make an array list of an array list?

Below too

shared/health/src/test/java/io/pravega/shared/health/impl/ContributorRegistryTests.java
```diff
@@ -0,0 +1,386 @@
...
+                    }
+                }
+                // Do the same for the parent set.
+                for (val collection : parents.entrySet()) {
+                    for (val contributor : collection.getValue()) {
+                        assert contributor.equals(contributors.get(contributor.getName()));
+                    }
+                }
+            } catch (AssertionError e) {
+                toString();
```
andreipaduroiu, line 356: What is the point of this? It will have no effect. 

Please let the exception bubble up. In general, please do not catch any `*Error`, especially `AssertionError` - those are indications that the tests found something bad.

andreipaduroiu (changes requested):
In addition to the above/below comments, can you please verify that in all your unit tests you properly close all instances that require closing? This is very hard for a reviewer to verify as we can't figure out from looking at plain-text code.

Example: if class X implements AutoCloseable, then any instance of X must invoke `.close()` in a `finally` block or `tearDown`, or just decorate it with the `@Cleanup` tag (search for this elsewhere in the code).

shared/health/src/main/java/io/pravega/shared/health/DetailsProvider.java
```diff
@@ -0,0 +1,70 @@
...
+ * A details object encapsulates any health related, non-{@link Status} state that an operator may be interested in. A details
+ * object should provide information that helps gauge the well-being of the associated component.
+ */
+@ThreadSafe
+public class DetailsProvider {
+    /**
+     * The underlying {@link java.util.Collection} used to hold the detailed information.
+     */
+    @Getter
+    private final Map<String, Supplier<Object>> suppliers = Collections.synchronizedMap(new HashMap<>());
```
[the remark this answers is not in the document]
co-jo: No.

shared/health/src/main/java/io/pravega/shared/health/impl/CompositeHealthContributor.java
co-jo: This iteration over all children must still be done to build the aggregate `Status`, so I figured  it would not be that wasteful to simply return the children `Health` results in case the details are requested.

Is something like this better?:
```
        Collection<Status> statuses = new ArrayList<>();
        List<Health> children = new ArrayList<>();
        getContributors().stream()
                .filter(Objects::nonNull)
                .forEach(contributor -> {
                    Health health = contributor.getHealthSnapshot(includeDetails);
                    statuses.add(health.getStatus());
                    if (includeDetails) {
                        children.add(health);
                    }
                });
```

shared/health/src/main/java/io/pravega/shared/health/impl/CompositeHealthContributor.java
co-jo: That's correct. If you think this is too verbose I think we can rely on this effect being noticed anyways simply through how this status may affect any parent statuses (e.g. it is apart of an aggregate).

shared/health/src/main/java/io/pravega/shared/health/impl/ContributorRegistryImpl.java
co-jo: No, an exception would be better. No calls actually use the `HealthContributor` return value at the moment, so I'm considering just making this method `void`.

shared/health/src/main/java/io/pravega/shared/health/impl/CompositeHealthContributor.java
co-jo: The initial idea was to allow for the flexibility to have `HealthContributor` objects to be either registered under a `HealthRegistry` or within the `CompositeHealthContributor` itself. In retrospect this doesn't add much utility and unnecessarily increases complexity. Right now it serves the purpose for allowing a `CompositeHealthContributor` to be instantiated without having to be registered under a registry for the purposes of the 'health hierarchy' build/validation. See https://github.com/pravega/pravega/pull/5991/files#diff-6caef6c95eeb4891e53cf79c4d0811dcb8b4d2efe72cef05ca814711c7bd247bR180. The `HealthConfig` is being phased out to a different approach so this shouldn't be an issue moving forward.

shared/health/src/main/java/io/pravega/shared/health/impl/ContributorRegistryImpl.java
co-jo: I'm not sure I understand. First on 169 the children set for the soon to be registered `HealthContributor` is initialized, then on 171 the contributor is added as a child to the parent (`componentName`).

shared/health/src/main/java/io/pravega/shared/health/HealthContributor.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * A human-readable identifier used for logging.
+     * @return The name provided.
+     */
+    String getName();
+
+    /**
+     * Closes the {@link HealthContributor} and forwards the closure to all its children.
+     */
+    void close();
```
andreipaduroiu, line 47: Make this interface extend `AutoCloseable`. 

And even though `AutoCloseable` defines a `close() throws Exception`, it is OK for your override to not declare `throws Exception`.

If the interface/class implements `AutoCloseable`, then whomever is using it knows that `close` must be invoked; it will also enable you to write `try-with-resources` code in Java.

shared/health/src/main/java/io/pravega/shared/health/HealthServiceUpdater.java
```diff
@@ -0,0 +1,49 @@
...
+ * This is useful in cases where health information is not regularly queried by some client. In the event of a crash or failure,
+ * (should we want to provide this information) it allows us to place an upper bound on how stale this {@link Health}
+ * information may be.
+ */
+public interface HealthServiceUpdater extends Service, AutoCloseable {
+
+    /**
+     * The interval at which to perform the service level health check.
+     */
+    Duration DEFAULT_INTERVAL_SECONDS = Duration.ofSeconds(10);
```
andreipaduroiu, line 33: Is hardcoding this a good idea? 

Are we considering to make this configurable so that we may tweak it for various deployments? 

When you write your tests, are you going to rely on it (so force the test to take 10+ seconds)?

shared/health/src/main/java/io/pravega/shared/health/StatusAggregationRule.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return A {@link Status} object describing the aggregate status.
+     */
+    public static Status majority(Collection<Status> statuses) {
+        if (statuses.isEmpty()) {
+            return Status.UNKNOWN;
+        }
+        int isAliveStatus = 0;
+        // Should be a strict majority.
+        for (Status status : statuses) {
+            isAliveStatus = status.isAlive() ? isAliveStatus + 1 : isAliveStatus - 1;
```
andreipaduroiu, line 46: `isAliveStatus += status.isAlive ? 1 : -1`

shared/health/src/test/java/io/pravega/shared/health/HealthServiceUpdaterTests.java
```diff
@@ -0,0 +1,82 @@
...
+        healthServiceUpdater = service.getHealthServiceUpdater();
+        healthServiceUpdater.awaitRunning();
+    }
+
+    @After
+    public void after() {
+        factory.close();
+        service.close();
+        healthServiceUpdater.stopAsync();
+        healthServiceUpdater.awaitTerminated();
```
andreipaduroiu, line 52: You need to invoke `.close()` on healthServiceUpdater. If you look in your implementation, this is the only place where you shut down the executor it creates; otherwise it will leave threads hanging around for the remainder of your process.

Please verify your other unit tests as well - it is not exactly easy to trace this on Github code review.

andreipaduroiu (changes requested):
Looking much better. I left a few more comments, but they should be easy to resolve.

shared/health/src/test/java/io/pravega/shared/health/HealthServiceUpdaterTests.java
co-jo: `close` is invoked for the healthServiceUpdater from within the `close` call of the service.

shared/health/src/main/java/io/pravega/shared/health/Health.java
```diff
@@ -0,0 +1,76 @@
...
+    @Getter
+    private final String name;
+
+    @Getter
+    @Builder.Default
+    private final Status status = Status.UNKNOWN;
+
+    @Getter
+    @Builder.Default
+    private final Map<String, Object> details = ImmutableMap.of();
```
pbelgundi, line 48: What is this map with _Object_ values for? How does it help provide health information?

pbelgundi:
I went over the PDP & code and my understanding from the same is this:
1. A Health Service created using this framework can be used to _query_ the health of all components that have "registered" with this service. 
2. The service stores a tree of **Health Contributors** with a single root and this is perhaps how a new Contributor can be added to the Service:
`service.getRoot().register(contributor);`
3. Based on the HealthEndpoint query, the "health tree" for the appropriate contributor (component) is returned.
4. The service stores all health information in memory and so on restart it need to perform the Health check again to get Health information about Contributors.
5. A HealthServiceUpdater thread periodically performs health check on the root contributor. This in-turn should trigger health checks for its children, grandchildren etc...?
The HealthServiceUpdater invokes `root.getHealthSnapshot()` on every health check run, which inturn invokes doHealthCheck(Health.HealthBuilder builder) on the root Contributor.
6. Each component ( HealthContributor) is expected the implement the `doHealthCheck` method that would return the Health for the component.

Since this is a "framework" a lot of implementation details are not clear from the documentation/code.
1. How does a component register itself with this service?  or who registers all pravega components with this service? 
2. Who brings up the Health Check service? Is the Pravega Operator expected to bring it up as part of a Pravega deployment?
3. Can the interval at which `HealthServiceUpdater` runs be changed?
4. I see that doHealthCheck() returns a `Status` object instead of returning a Health object. The what the is "Health" object for?
5. Since the service can have only 1 root object I assume it can store health data for only 1 Pravega cluster. Why not enable the service to have multiple roots so it can store data for multiple pravega clusters?

pbelgundi:
> Refactoring out common REST server classes from the controller project into some shared project

Why is this needed?

co-jo:
> I went over the PDP & code and my understanding from the same is this:
> 
> 1. A Health Service created using this framework can be used to _query_ the health of all components that have "registered" with this service.
> 2. The service stores a tree of **Health Contributors** with a single root and this is perhaps how a new Contributor can be added to the Service:
>    `service.getRoot().register(contributor);`
> 3. Based on the HealthEndpoint query, the "health tree" for the appropriate contributor (component) is returned.
> 4. The service stores all health information in memory and so on restart it need to perform the Health check again to get Health information about Contributors.
> 5. A HealthServiceUpdater thread periodically performs health check on the root contributor. This in-turn should trigger health checks for its children, grandchildren etc...?
>    The HealthServiceUpdater invokes `root.getHealthSnapshot()` on every health check run, which inturn invokes doHealthCheck(Health.HealthBuilder builder) on the root Contributor.
> 6. Each component ( HealthContributor) is expected the implement the `doHealthCheck` method that would return the Health for the component.
> 
> Since this is a "framework" a lot of implementation details are not clear from the documentation/code.
> 
> 1. How does a component register itself with this service?  or who registers all pravega components with this service?
> 2. Who brings up the Health Check service? Is the Pravega Operator expected to bring it up as part of a Pravega deployment?
> 3. Can the interval at which `HealthServiceUpdater` runs be changed?
> 4. I see that doHealthCheck() returns a `Status` object instead of returning a Health object. What the is "Health" object for?
> 5. Since the service can have only 1 root object I assume it can store health data for only 1 Pravega cluster. Why not enable the service to have multiple roots so it can store data for multiple pravega clusters?

Thanks for taking the time to provide your feedback. The points you outlined with respect to your understanding are correct. Addressing your questions:

1/2. I'll be setting up the SegmentStore/Controller with this health service in upcoming PR(s), but only requires creating an instance of `HealthService`, passing down the root `HealthContributor` or whatever `HealthContributor` you want to be the parent of another `HealthContributor` and exposing the details via a REST server. There is no singular `HealthService` that watches and keeps track of all cluster components, so there is no built in concept of having a cluster wide health check. Each SegmentStore/Controller instance will have its own `HealthService`.

Having said that, a standalone cluster 'aware' HealthService can easily be built with this framework to address that requirement, but no such thing is included in this PR.

3. Yes.
4. The `doHealthCheck` call is for defining the health related logic for _just_ the `HealthContributor` itself, so that means defining any details it wants to export and seeing (without consideration from its child contributors) if it should consider itself healthy or not. That information is persisted in the passed in `Health.HeathlBuilder` object.

The reason why `Status` is returned is so we can properly reconcile the `Status` as defined from `doHealthCheck` for that individual`HealthContributor` and the `Status` objects of all its registered children. The `getHealthSnapshot` call does this reconciliation between the parent and child to build the `Health` object for the parent.
5. By definition there can only be one root in a tree. Not sure I'm understanding correctly, but how would having two 'roots' be any different then just registering each cluster under the root (as it is now) as 'cluster-1', 'cluster-2', etc?

shared/health/src/main/java/io/pravega/shared/health/Health.java
co-jo: This is for any information that a `HealthContributor` wants to expose, apart from its health `Status`. The reason why its of type `Object` is to allow the consuming service whether it be the controller or segementstore to define how to how these objects should be serialized.

co-jo:
> > Refactoring out common REST server classes from the controller project into some shared project
> 
> Why is this needed?

As mentioned in the other PR, the SegmentStore will be using a REST server to expose these health details.

pbelgundi:
> > > Refactoring out common REST server classes from the controller project into some shared project
> > 
> > 
> > Why is this needed?
> 
> As mentioned in the other PR, the SegmentStore will be using a REST server to expose these health details.

I'm not sure I follow this, I thought the _HealthService_ was expected to expose health details for a component so why does the Controller REST API need to do it? I presume these APIs are embodiments of actual "health checks" to be performed on Segment Store and so would have implementation details of how the check is to be performed. This is something internal to Segment Store and so not sure why Controller REST API needs to change for that.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthContributorImpl.java
```diff
@@ -0,0 +1,162 @@
...
+ * The {@link HealthContributorImpl} class defines the base logic required to build {@link HealthContributor} objects
+ * that depend on (or are composed of) other {@link HealthContributor} objects. To do this we need to due two things:
+ * reducing/reconciling the (potentially) many different {@link Status} states and a way to persist these relationships.
+ *
+ * {@link HealthContributorImpl} objects are used *strictly* for grouping, meaning that they do not define their own
+ * health checking logic and therefore does not export any details ({@link Health#getDetails}).
+ */
+@Slf4j
+@ThreadSafe
+public abstract class HealthContributorImpl implements HealthContributor {
```
pbelgundi, line 47: I would not call this class an "Impl" as it is an abstract class and not a concrete one that would have the actual health check implementation. Please consider changing the name of this one to indicate that.

pbelgundi:
> > I went over the PDP & code and my understanding from the same is this:
> > 
> > 1. A Health Service created using this framework can be used to _query_ the health of all components that have "registered" with this service.
> > 2. The service stores a tree of **Health Contributors** with a single root and this is perhaps how a new Contributor can be added to the Service:
> >    `service.getRoot().register(contributor);`
> > 3. Based on the HealthEndpoint query, the "health tree" for the appropriate contributor (component) is returned.
> > 4. The service stores all health information in memory and so on restart it need to perform the Health check again to get Health information about Contributors.
> > 5. A HealthServiceUpdater thread periodically performs health check on the root contributor. This in-turn should trigger health checks for its children, grandchildren etc...?
> >    The HealthServiceUpdater invokes `root.getHealthSnapshot()` on every health check run, which inturn invokes doHealthCheck(Health.HealthBuilder builder) on the root Contributor.
> > 6. Each component ( HealthContributor) is expected the implement the `doHealthCheck` method that would return the Health for the component.
> > 
> > Since this is a "framework" a lot of implementation details are not clear from the documentation/code.
> > 
> > 1. How does a component register itself with this service?  or who registers all pravega components with this service?
> > 2. Who brings up the Health Check service? Is the Pravega Operator expected to bring it up as part of a Pravega deployment?
> > 3. Can the interval at which `HealthServiceUpdater` runs be changed?
> > 4. I see that doHealthCheck() returns a `Status` object instead of returning a Health object. What the is "Health" object for?
> > 5. Since the service can have only 1 root object I assume it can store health data for only 1 Pravega cluster. Why not enable the service to have multiple roots so it can store data for multiple pravega clusters?
> 
> Thanks for taking the time to provide your feedback. The points you outlined with respect to your understanding are correct. Addressing your questions:
> 
> 1/2. I'll be setting up the SegmentStore/Controller with this health service in upcoming PR(s), but only requires creating an instance of `HealthService`, passing down the root `HealthContributor` or whatever `HealthContributor` you want to be the parent of another `HealthContributor` and exposing the details via a REST server. There is no singular `HealthService` that watches and keeps track of all cluster components, so there is no built in concept of having a cluster wide health check. Each SegmentStore/Controller instance will have its own `HealthService`.
> 
> Having said that, a standalone cluster 'aware' HealthService can easily be built with this framework to address that requirement, but no such thing is included in this PR.
> 

I don't see why we would want to run multiple health services for a single product, that too when a single service is able to capture the status of a component and all its children. This is essentially an observability/diagnostic service and so we'd want to keep the footprint as low as possible.

> 1. Yes.
> 2. The `doHealthCheck` call is for defining the health related logic for _just_ the `HealthContributor` itself, so that means defining any details it wants to export and seeing (without consideration from its child contributors) if it should consider itself healthy or not. That information is persisted in the passed in `Health.HeathlBuilder` object.
> 

That perhaps means "something" needs to implement the  "doHealthCheck()" for Controller and yet another thing for  SegmentStore. Who does that? Something like _ControllerHealthContributorImpl_ and _SegmentStoreHealthContributorImpl_?
How will these make the actual health check invocations on these respective components?


> The reason why `Status` is returned is so we can properly reconcile the `Status` as defined from `doHealthCheck` for that individual`HealthContributor` and the `Status` objects of all its registered children. The `getHealthSnapshot` call does this reconciliation between the parent and child to build the `Health` object for the parent.

I see that the "Health" object has the Status object inside it, so I would find it more intuitive to just return the Health object as return value of `doHealthCheck`

> 5. By definition there can only be one root in a tree. Not sure I'm understanding correctly, but how would having two 'roots' be any different then just registering each cluster under the root (as it is now) as 'cluster-1', 'cluster-2', etc?

Looks like you didn't read it right. I'm not asking for a tree with multiple roots, but multiple trees, that can be maintained by a single service. I don't see why we need to limit the design to allow only 1 tree.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthContributorImpl.java
```diff
@@ -0,0 +1,162 @@
...
+                statuses.add(health.getStatus());
+            } else {
+                contributors.remove(name);
+            }
+        }
+
+        Status status = Status.DOWN;
+        // Perform own health check logic.
+        try {
+            status = doHealthCheck(builder);
```
pbelgundi, line 116: `getHealthSnapshot()` returns a `Health` object for a component and all its subcomponents. Since doHealthCheck() has actual health check implementation for 1 component that could be subcomponent of another one or a top level component I would expect this to return a Health object going by the composite design pattern.

shared/health/src/main/java/io/pravega/shared/health/Health.java
pbelgundi: Ok.

co-jo:
> > > > Refactoring out common REST server classes from the controller project into some shared project
> > > 
> > > 
> > > Why is this needed?
> > 
> > 
> > As mentioned in the other PR, the SegmentStore will be using a REST server to expose these health details.
> 
> I'm not sure I follow this, I thought the _HealthService_ was expected to expose health details for a component so why does the Controller REST API need to do it? I presume these APIs are embodiments of actual "health checks" to be performed on Segment Store and so would have implementation details of how the check is to be performed. This is something internal to Segment Store and so not sure why Controller REST API needs to change for that.

Maybe a better description is to say that the `HealthService` is the entry point or top-level container for all health related concerns. It 'exposes' health information in the sense that given a reference to the `HealthService`, any health information of any registered `HealthContributor` can be retrieved. The `HealthService` does not provide a direct means of exposing that information outside of the controller/segmentstore. 

Having said that, I'm still not sure I understand your concern here. There are minimal changes to the 'controller REST API' and is largely a simple relocation of the REST services into a shared package so we can re-use said work in the segmentstore. Are you suggesting we ought to have a REST server directly in the `health` project?

co-jo:
> I don't see why we would want to run multiple health services for a single product, that too when a single service is able to capture the status of a component and all its children. This is essentially an observability/diagnostic service and so we'd want to keep the footprint as low as possible.

I'm not sure what your concern is pointing at exactly. Yes there are multiple `HealthService` instances (one per segmentstore/controller instance), but how else will information be exposed from each instance?

> That perhaps means "something" needs to implement the "doHealthCheck()" for Controller and yet another thing for SegmentStore. Who does that? Something like _ControllerHealthContributorImpl_ and _SegmentStoreHealthContributorImpl_?
> How will these make the actual health check invocations on these respective components?

There is no single `HealthContributor` for each segmentstore/controller. Each respective team will have to consider what they want to track and create an extension of  `HealthContributor` implementing the `doHealthCheck` which defines the logic..

> I see that the "Health" object has the Status object inside it, so I would find it more intuitive to just return the Health object as return value of `doHealthCheck`

The `doHealthCheck` is not responsible for building the `Health` object of the component. It only should define the logic for that `HealthContributor` without having to consider other child objects, so it would not make sense to return a `Health` object.

> Looks like you didn't read it right. I'm not asking for a tree with multiple roots, but multiple trees, that can be maintained by a single service. I don't see why we need to limit the design to allow only 1 tree.

What value does that add? Separating out multiple trees breaks the recursive nature of the service. Why not just have each separate 'tree' as a child node to the root?

shared/health/src/main/java/io/pravega/shared/health/impl/HealthContributorImpl.java
co-jo: `doHealthCheck` is responsible for a) Determining the status of some arbitrary component/class and b) declaring the details you wish to export for said component/class. The important part is this is done _without_ consideration of its children components, so it cannot build a complete `Health` object. `getHealthSnapshot` reconciles the children and the parent to build the `Health` object. I don't see any violation of the 'composite design pattern'.

andreipaduroiu (approved):

shared/health/src/main/java/io/pravega/shared/health/HealthEndpoint.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Fetches the results from the list of {@link java.util.function.Supplier} provided during {@link HealthContributor}
+     * construction.
+     *
+     * @param id The id of some {@link HealthContributor} to search for.
+     * @return The {@link Map} of details results.
+     */
+    Map<String, Object> getDetails(String id);
```
pbelgundi, line 101: What is the ID to be provided here? How does the operator or any client get to know this id?

pbelgundi (changes requested):

shared/health/src/main/java/io/pravega/shared/health/HealthEndpoint.java
co-jo: That is the name/identifier of the `HealthContributor` to request. Correct me if I'm wrong but I'd imagine the operator would only care about the top/service level status, so one would just make a call without the `id` specified. If a client wants to understand what are the available components they just have to make this top level request (`getHealth()`), returning all the health information for the service.

pbelgundi: The operator for doing the _readiness_ and _liveness_ probes needs to know the **overall** readiness/liveness of the whole component it is probing, so yes you are right it needs to know these values only for the top level component.
I see that HealthEndpoint has both methods `boolean isAlive()` and `boolean isAlive(String id)`, so perhaps the expectation is to invoke `isAlive()` _without args_ to understand the liveness for a top-level component (eg Controller) and provide the `id` only in case of sub-components of Controller. Is my understanding correct?

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Provides a simpler interface for retrieving {@link Health} information from the {@link io.pravega.shared.health.HealthService}.
+ * The {@link HealthEndpoint} should be used to serve information about the {@link Health} of a component for external classes.
+ */
+@Slf4j
+@RequiredArgsConstructor
+public class HealthEndpointImpl implements HealthEndpoint {
+
+    @NonNull
+    private final HealthContributor root;
```
pbelgundi, line 41: `HealthEndpointImpl` and `HealthServiceImpl` look identical to me. Both hold the root HealthContributor object and the updater object. Both are also exposed to clients.
How are these 2 APIs different: `healthService.getHealth()` and  `healthService.getEndpoint().getHealth()` ? You may consider refactoring these into a single class.

pbelgundi (changes requested):

shared/health/src/main/java/io/pravega/shared/health/HealthService.java
```diff
@@ -0,0 +1,49 @@
...
+     * @return The {@link HealthEndpoint} instance.
+     */
+    HealthEndpoint getEndpoint();
+
+    /**
+     * Provides the {@link HealthServiceUpdater} that currently performs the background {@link Health} requests (of the root).
+     *
+     * @return The  {@link HealthServiceUpdater}.
+     */
+    HealthServiceUpdater getHealthServiceUpdater();
```
pbelgundi, line 38: Is it necessary to expose the `HealthServiceUpdator` to Clients? What would clients do with it? I see only 2 public APIs on this class: `getLatestHealth()` and `getDuration()`. 
**Do we want to allow clients to trigger a health probe synchronously?** If not, we should not be exposing this class.
Classes exposed to clients become part of the public API and are difficult to modify later. I would not recommend doing this unless absolutely necessary.

pbelgundi (changes requested):

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
co-jo: I see your point -- there definitely is a little overlap but I wouldn't consider them __identical__ per se. There is no `healthService.getHealth()` call and the `HealthServiceUpdater` and root `HealthContriubtor` are not directly exposed in the `HealthEndpoint` as they are in the `HealthService`. I think there is value in separating the current `HealthEndpoint` logic in it's own class but I don't feel strongly one way or the other. @andreipaduroiu Do you have any thoughts on this?

shared/health/src/main/java/io/pravega/shared/health/HealthService.java
co-jo: The current reason why the `HealthServiceUpdater` is exposed is to allow clients to properly start and stop the underlying executor services. I'm not quite sure how we would handle those cases if made private.

shared/health/src/main/java/io/pravega/shared/health/HealthEndpoint.java
co-jo: > 
> 
> The operator for doing the _readiness_ and _liveness_ probes needs to know the **overall** readiness/liveness of the whole component it is probing, so yes you are right it needs to know these values only for the top level component.
> I see that HealthEndpoint has both methods `boolean isAlive()` and `boolean isAlive(String id)`, so perhaps the expectation is to invoke `isAlive()` _without args_ to understand the liveness for a top-level component (eg Controller) and provide the `id` only in case of sub-components of Controller. Is my understanding correct?

Thats correct. The parameterized calls are expected to be used mostly by some human operator or some higher level script  or something given that the root is the only `HealthContributor` that can reliably be known to exist ahead of time.

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
andreipaduroiu: I looked at both these classes and I see both your points. The way these will be used, from what I understand, is that the application bootstrapper (`ServiceStarter` for segment store and `ControllerServiceStarter` for the controller) will create a `HealthServiceImpl` object and manage it, while the soon-to-be-added REST health endpoint will call `HealthService.getEndpoint()` and work with that.

So in essence, the `HealthServiceImpl` is nothing more than a manager for `HealthEndpoint`. Another term we could use to describe it is a "factory", however, it does not create new `HealthEndpoint` objects, it only has one and it always returns that one.

From another POV, the `HealthServiceImpl` adds some value on top of `HealthEndpointImpl`, but not much. I can see why we'd want to merge the two together. However, from a testing POV, it will be easier to keep the separation since we can mock a `HealthEndpoint` without having to worry about whatever else the `HealthService` might do.

My recommendation, to simplify the design would be:
1. Remove the `HealthService` interface (whatever `HealthServiceImpl` implements). We only expect one of these objects per process and there's no point in maintaining an interface as well.
2. Rename `HealthServiceImpl` to `HealthEndpointManager`. Keep all logic within it the same.
3. Consider what exactly the newly named `HealthEndpointManager` should expose. It does need to expose the endpoint, but does it need to expose the root or updater? If the answer is "only for testing", then add package-private getters and mark them as `@VisibleForTesting`; if the answer is "no", then don't expose them at all.

pbelgundi: Yes this sounds good to me.

pbelgundi (approved):
Approving, assuming the review comments mentioned are addressed as discussed.

shared/health/src/main/java/io/pravega/shared/health/HealthServiceManager.java
```diff
@@ -0,0 +1,93 @@
...
+     * the various types of requests the {@link HealthServiceManager} will be able to fulfill.
+     *
+     * @return The {@link HealthEndpoint} instance.
+     */
+    public HealthEndpoint getEndpoint() {
+        return this.endpoint;
+    }
+
+    @VisibleForTesting
+    public HealthServiceUpdater getHealthServiceUpdater() {
```
pbelgundi, line 69: Please remove the public modifier, so this can be used by tests but is not exposed to clients.

shared/health/src/main/java/io/pravega/shared/health/HealthServiceFactory.java
```diff
@@ -0,0 +1,61 @@
...
+import io.pravega.common.Exceptions;
+import lombok.NonNull;
+
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Provides instances of a {@link HealthServiceManager} and optionally starts the {@link HealthServiceUpdater}.
+ */
+public class HealthServiceFactory implements AutoCloseable {
```
andreipaduroiu, line 27: I think we can further simplify this by removing this factory class. There is no added value in having a factory that creates a single type, which can be achieved by invoking that type's constructor.

shared/health/src/main/java/io/pravega/shared/health/HealthServiceManager.java
```diff
@@ -0,0 +1,93 @@
...
+    public HealthEndpoint getEndpoint() {
+        return this.endpoint;
+    }
+
+    @VisibleForTesting
+    HealthServiceUpdater getHealthServiceUpdater() {
+        return updater;
+    }
+
+    public void close() {
```
andreipaduroiu, line 73: `@Override` and make this class implement `AutoCloseable`

shared/health/src/test/java/io/pravega/shared/health/HealthServiceTests.java
```diff
@@ -0,0 +1,183 @@
...
+import org.junit.rules.Timeout;
+
+import io.pravega.shared.health.TestHealthContributors.HealthyContributor;
+
+/**
+ * The {@link HealthServiceTests} encapsulates much of the same processes that {@link HealthEndpoint} performs, so
+ * an explicit test class for the former is skipped.
+ */
+@Slf4j
+public class HealthServiceTests {
```
andreipaduroiu, line 40: Rename to `HealthManagerTests`

andreipaduroiu (changes requested):

andreipaduroiu (approved):

PR #6014, opened 2021-05-12
issue 6004: Change Controller API retry mechanism to retry on all exceptions except for specific non-retryable ones

### LINKED ISSUE #6004: Controller does not retry API operations failing due to OutofMemory Errors
**Describe the bug**
Recently we hit the "OOM: unable to create native threads" issue on Controller, on some environments and Controller APIs impacted by this were not **retried** as this exception is not under the set of retry able exceptions for Controller.
This needs to change as this error is intermittent and when it occurs because of lack of memory, it goes away after we're able to provide the necessary memory to the Controller pod.

**To Reproduce**
Deploy a Pravega cluster and run longevity with 1 mediumScale plan , 1 newScaleTx plan, and mediumScaleTxBytes plan all together. One of the 2 Controllers hits this issue.

pbelgundi:

**Change log description**  
Currently Controller APIs, retry a failed operation only when the exception thrown by the sub-operation is a _Retryable Exception_. In every other case, the exception propagates to the client and the API fails. Due to this, we skip retries on a lot of intermittent, transient exceptions like OOM, TimeoutException etc...
In this PR the retry logic on Controller has been changed to retry on all exceptions except for specific "non-retryable" ones where it does not make sense to retry the operation.

**Purpose of the change**  
Fixes #6004

**What the code does**  
Argument to withRetriesAsync() has been changed from RETRYABLE_PREDICATE to NON_RETRYABLBE_PREDICATE, to indicate that we retry on every exception that is not a NON-RETRYABLE Exception.
This PR also has some log improvement changes.

**How to verify it**  
All unit system and integration tests should pass


controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -371,7 +378,7 @@ private ReaderGroupConfiguration getRGConfigurationFromRecord(final String scope
                      return buildCreateSuccessResponse(scope, rgName);
                  });
          });
-        }, e -> Exceptions.unwrap(e) instanceof RetryableException, READER_GROUP_OPERATION_MAX_RETRIES, executor);
+        }, retryPredicate, READER_GROUP_OPERATION_MAX_RETRIES, executor);
```
shiveshr, line 381: lets not retry on UNCONDITIONAL predicate. we should fail for exceptions which we know are non retryable.

shiveshr (changes requested):
Please add specific failing exceptions for each request so that we dont retry unconditionally on each and every exception.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
pbelgundi: Did not hard code UNCONDITIONAL_PREDICATE here, as at this point I could not see any specific exceptions for which we would not want to retry. When these are found we can change the retryPredicate to include those.
Specific scenarios like - SCOPE_NOT_FOUND are already dealt with earlier in this code.

shiveshr: i did not follow your point. by using unconditional predicate here, arent we saying this should be retried on all exceptions?

pbelgundi: Yes we are. Because at this point we do not know of any exceptions that need to be "non-retryable".
If and when we do, we change the retryPredicate accrodingly for that specific method.

shiveshr: but we do know that. for each case we should know which are certainly failing exceptions. that set can be expanded if we realize we havent covered some exception but for most cases we know. 
Things like StoreExceptions.DataNotFound (or DataContainerNotFound) are non retryable for almost all workflows, saving some exceptions.
similarly when we do deletes with conditional flag (delete if empty) then StoreExceptions.DATA_CONTAINS_ELEMENTS is a failing exception. So on and so forth.
based on context we know which are failing cases which should not be retried upon.

pbelgundi:
Closing this PR as it is not needed any more

PR #6018, opened 2021-05-13
Issue 2887: AutoScale stream if configuration changes the minNumberOfSegments

### LINKED ISSUE #2887: StreamManager.updateStream should scale stream to uphold minumum number of segments (or fixed number of segments)
**Problem description**
Currently updating minimum number of segments for a stream results in scaling policy update in stream configuration. 
The policy enforcement happens during subsequent scale requests. This means if a stream had lesser number of segments than the minimum number of segments specified in the policy, then it will continue to remain below the newly configured value. 

We should evaluate if we should automatically trigger a scale up in response to updateStream request to change the number of segments in the stream to at least meet minimum number of segments requirement.

Similarly, for a fixed sized stream, if the minNumberOfSegments are changed, the stream should automatically change the partitioning to handle the traffic.

**Problem location**
UpdateStreamTask

shiveshr:

**Change log description**  
Applies configuration change and enforces minimum number of segments. Scales stream if the stream has lesser number of segments (for auto scaling) and different number of segments (for fixed policy). 

**Purpose of the change**  
Fixes #2887 

**What the code does**  
Abstracts out scaling logic from ScaleOperationTask into StreamMetadataTask and uses that in UpdateStreamTask.
A stream can be scaled as part of update workflow if the active segments in the stream are different.

**How to verify it**  
Unit tests added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateStreamTask.java
+++ b/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateStreamTask.java
@@ -102,17 +102,17 @@ public class UpdateStreamTask implements StreamTask<UpdateStreamEvent> {
                 .thenCompose(etr -> streamMetadataStore.updateVersionedState(scope, stream, State.UPDATING, state, context, executor)
                 .thenCompose(updated -> updateStreamForAutoStreamCut(scope, stream, configProperty, updated)
                         .thenCompose(x -> notifyPolicyUpdate(context, scope, stream, configProperty.getStreamConfiguration(), requestId))
-                        .thenCompose(x -> handleSegmentCount(scope, stream, configProperty, etr, context, executor, requestId))
+                        .thenCompose(x -> handleSegmentCounUpdates(scope, stream, configProperty, etr, context, executor, requestId))
                         .thenCompose(x -> streamMetadataStore.completeUpdateConfiguration(scope, stream, record, context, executor))
                         .thenCompose(x -> streamMetadataStore.updateVersionedState(scope, stream, State.ACTIVE, updated, context, executor)))));
     }
 
-    private CompletableFuture<Void> handleSegmentCount(final String scope, final String stream, 
-                                                       final StreamConfigurationRecord config, 
-                                                       final VersionedMetadata<EpochTransitionRecord> etr, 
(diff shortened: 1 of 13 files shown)


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateStreamTask.java
```diff
@@ -84,11 +97,55 @@ public UpdateStreamTask(final StreamMetadataTasks streamMetadataTasks,
...
+        final double keyRangeChunk = 1.0 / numSegments;
+        
+        List<Map.Entry<Double, Double>> newRange = IntStream.range(0, numSegments).boxed()
+                                                            .map(x -> new AbstractMap.SimpleEntry<>(x * keyRangeChunk, (x + 1) * keyRangeChunk))
+                                                            .collect(Collectors.toList());
+        log.debug("{} Scaling stream to update minimum number of segments to {}", requestId, numSegments);
+        return streamMetadataStore.resetEpochTransition(scope, stream, etr, context, executor)
+              .thenCompose(reset -> streamMetadataStore.submitScale(scope, stream, new ArrayList<>(activeEpoch.getSegmentIds()), newRange,
+                      System.currentTimeMillis(), reset, context, executor))
+              .thenCompose(updated -> streamMetadataTasks.processScale(scope, stream, updated, context, requestId, streamMetadataStore)
```
shrids, line 146: Is this being treated as a manual scale here? Also, why is `io.pravega.controller.store.stream.StreamMetadataStore#startScale` not required here?

shrids (approved):
No blocking comments apart from one query..

shrids:
There are multiple integration tests that are fail, please take a look at it ...

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateStreamTask.java
shiveshr: startScale simply does some state sanity check for state being SCALING. And migrates epoch transition record if required. 
Here we are actually resetting the epoch transition explicitly to EMPTY whereby discarding any outstanding scale requests because we are going to replace all the segments with new set of segments matching the new policy so any outstanding scale requests are irrelevant and discardable.

PR #6019, opened 2021-05-13
Issue 5766: Recursive deleteScope does not handle delete of KVTs and ReaderGroups

### LINKED ISSUE #5766: Recursive deleteScope does not handle delete of KVTs and ReaderGroups
**Describe the bug**
The new `deleteScope` API which deletes the scope while also providing the option to remove the primitives present in the scope such as streams, does not remove other primitives such as kvts and readergroups. These can interfere with the deletion of the scope even after all the streams are removed.

**To Reproduce**
1. Create a scope.
2. Create a kvt under the scope.
3. Call the `deleteScope` API with the `deleteStreams` set to true.
4. The following error shows up:
```
Caused by: java.lang.IllegalStateException: Scope testScope is not empty.
```

**Expected behavior**
When using this API with the option to delete all the primitives in the scope then all the readergroups and kvts present in the scope should also be deleted to make the deletion of the scope does not fail due to it.

anirudhkovuru:
**Change log description**  
Made changes to `deleteScope` in the `StreamManagerImpl` class. Setting the `forceDelete` option will delete all the streams, key-value tables and readergroups present in the scope and then delete it.

**Purpose of the change**  
Fixes #5766 

**What the code does**  
Added code to list existing key-value tables and then delete them. For readergroups, the list of readergroups is obtained from the list of streams by identifying the ReaderGroup streams (`_RG` prefix). The UUID is obtained from `getReaderGroupConfig` and then deleted.

**How to verify it**  
All tests should pass and build should succeed.


client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
```diff
@@ -184,6 +196,31 @@ public boolean deleteScope(String scopeName, boolean deleteStreams) throws Delet
...
+                } catch (Exception e) {
+                    String message = String.format("Failed to delete key-value table %s", kvt.getKeyValueTableName());
+                    throw new DeleteScopeFailedException(message, e);
+                }
+            }
+
+            for (String groupName: readerGroupList) {
+                try {
+                    Futures.getThrowingException(controller.getReaderGroupConfig(scopeName, groupName)
+                            .thenCompose(conf -> controller.deleteReaderGroup(scopeName, groupName,
```
shiveshr, line 214: remind me again, this will delete the RG stream as well, right?

shiveshr (commented):
LGTM, apart from one comment

client/src/main/java/io/pravega/client/admin/impl/StreamManagerImpl.java
anirudhkovuru: Yes. However, the RG stream is already being deleted early on while deleting all the streams. So the `deleteStream` call during `deleteReaderGroup` is idempotent.

shiveshr:
@anirudhkovuru LGTM, please fix the build failure

shiveshr (approved):

PR #6031, opened 2021-05-17
Issue 6030: (SegmentStore) Ignoring missing ledgers that were previously marked as Empty.

### LINKED ISSUE #6030: BookKeeperLog should not open-fence Ledgers which are marked as LedgerStatus.Empty
**Describe the bug**
If we try to open-fence ledgers that have previously been marked as Empty (only possible from a previous fence-out operation) then we might hit a missing ledger (deleted by us; but we crashed before updating metadata). This currently prevents the recovery from proceeding (or even starting).

We should modify the code in one of two ways:
1. Do not even attempt to open-fence empty ledgers (we are 100% certain they are empty since we only assign that status after a previous recovery that verified they are empty).
2. Continue as we do now, but if we get a `LedgerNotFoundException`, simply log a WARN and move on. 

Option 2 is preferable since we always attempt to re-verify a previous evaluation vs blindly trusting it.

andreipaduroiu:
**Change log description**  
During the BookKeeperLog initialization phase, we are now ignoring ledgers that we cannot open-fence due to them missing but have been previously identified as being Empty.

**Purpose of the change**  
Fixes #6030.

**What the code does**  
If a ledger has been marked as Empty, it means that we absolutely know it has no data. As such, during recovery we will auto-delete it. However, if we crash at one point in this process, it may be possible for the ledger to actually be deleted but we didn't have a chance to update the metadata accordingly. A subsequent recovery would attempt to re-open-fence this ledger and fail because it doesn't exist.

This change handles missing ledger exceptions during open-fencing and, if that ledger has been marked as empty previously, it is safe to skip it and move on. If the ledger, however, is not marked as Empty, then we will re-throw the exception.

**How to verify it**  
New unit test added.

### DIFF
--- a/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
+++ b/segmentstore/storage/impl/src/main/java/io/pravega/segmentstore/storage/impl/bookkeeper/Ledgers.java
@@ -240,9 +240,8 @@ public final class Ledgers {
             try {
                 handle = openFence(ledgerMetadata.getLedgerId(), bookKeeper, config);
             } catch (DurableDataLogException ex) {
-                boolean notExists = ex.getCause() instanceof BKException.BKNoSuchLedgerExistsException
-                        || ex.getCause() instanceof BKException.BKNoSuchLedgerExistsOnMetadataServerException;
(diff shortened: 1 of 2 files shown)


RaulGracia (approved):
LGTM (system tests are being executed now).

RaulGracia:
System tests have passed (build 3835 ).

Issue #6021, opened 2021-05-13
OperationProcessor queue may be left closed while the container is in RUNNING state


RaulGracia:
**Describe the bug**
Under a high load scenario, we sporadically face an issue in which:
1. A Bookie restarts/disconnects due to high load, which lead the Bookkeeper client to throw `BKNotEnoughBookiesException` and then some Segment Containers restart.

2. After that, a Segment Store seems that have all the containers recovered:
```
2021-05-12 15:02:12,342 6791252 [core-6] INFO  i.p.s.s.h.ZKSegmentContainerMonitor - Container Changes: Desired = [81, 18, 4, 85, 86, 39, 9, 61], Current = [81, 18, 4, 85, 86, 39, 9, 61], PendingTasks = [], ToStart = [], ToStop = [].
```
3. However, some operations against a Container (the most active one) in that Segment Store are failing:
```
2021-05-12 14:59:10,680 6609590 [storage-io-102] WARN  i.p.s.s.chunklayer.ReadOperation - ChunkedSegmentStorage[61] read - late op=1793443990, segment=_system/containers/storage_metadata_61$attributes.index, offset=306027643632, bytesRead=15734, latency=114.
io.pravega.common.ObjectClosedException: Object 'PriorityBlockingDrainingQueue' has been closed and cannot be accessed anymore.
...
2021-05-12 15:18:37,872 7776782 [epollEventLoopGroup-10-12] ERROR i.p.s.s.h.h.PravegaRequestProcessor - [requestId=75900249738051584] Error (Segment = 'flink-longevity-0/flink-longevity-0/39.#epoch.0', Operation = 'truncateSegment')
io.pravega.common.ObjectClosedException: Object 'PriorityBlockingDrainingQueue' has been closed and cannot be accessed anymore.
        at io.pravega.common.Exceptions.checkNotClosed(Exceptions.java:255)
...
2021-05-12 15:18:37,872 7776782 [epollEventLoopGroup-10-12] ERROR i.p.s.s.h.h.PravegaRequestProcessor - [requestId=75900249738051584] Error (Segment = 'flink-longevity-0/flink-longevity-0/39.#epoch.0', Operation = 'truncateSegment')
io.pravega.common.ObjectClosedException: Object 'PriorityBlockingDrainingQueue' has been closed and cannot be accessed anymore.
        at io.pravega.common.Exceptions.checkNotClosed(Exceptions.java:255)
```
[...]

PR #6033, opened 2021-05-18
Refine Bookkeeper client re-creation

RaulGracia:

**Change log description**  
Allow to parameterize the behavior of the Bookkeeper Client reset mechanism in BookkeeperLogFactory. Only enforce the Bookkeeper Client reset when the problem is known to be solved via a Bookkeeper client reset.

**Purpose of the change**  
TBD

**What the code does**  
In #5317, we introduced a mechanism to reset the Bookkeeper Client used by the Segment Store if a container was being persistently restarted. The goal was to fix an issue in the Bookkeeper Client that could lead to cache old DNS entries for a restarted Bookie until the underlying timeout for the TCP stack gets triggered (see full description here https://github.com/apache/bookkeeper/issues/2482).

However, that first approach has 3 main drawbacks:
- _Coarse grained_: The existing mechanism resets the Bookkeeper Client just considering Container restarts, irrespective of the actual root cause that originated the restart. Clearly, not all the problems that can induce a container restart can be solved via re-creating the Bookkeeper Client.
- _Impact on all other containers_: If a Container is malfunctioning and restarting continuously, the Bookkeeper Client will be continuously re-created as well. As the Bookkeeper Client is shared across Containers, the re-creation of the Bookkeeper Client will also induce the restart of all other Containers in the Segment Store as well. Thus, the current mechanism is amplifying the problem on a single container to all others within the same instance.
- _Not configurable_: The current mechanism uses 2 hardcoded constants to determine when to re-create the Bookkeeper Client that cannot be configured.

The current PR addresses these 3 drawbacks:
- Now, the `BookkeeperLogFactory` passes a callback to both the `BookkeeperLog` and `LogReader` classes. Such a callback is expected to be invoked when an exception occurs and will consume the exception thrown that is going to be computed by the Bookkeeper client reset logic in `BookkeeperLogFactory`.
- Only certain exceptions will be considered to reset the Bookkeeper client upon log re-creation. At the moment, as considered in the original issue (https://github.com/apache/bookkeeper/issues/2482), the only known symptom of the DNS problem is having continuous `BKException.BKReadException` upon recovery. Therefore, we only reset the Bookkeeper client in this case (despite we can add more cases in the future if necessary).
- We have added 2 new parameters to the Pravega configuration (`bookkeeper.client.reset.after.exception.num` and `bookkeeper.client.reset.inspection.time.seconds`) that replace the previously hardcoded constants (while keeping the same default value).

In turn, this change will prevent most of the unnecessary Container restarts that led to the error messages described in #6021.

**How to verify it**  
Existing tests have been adapted to the changes in the mechanism.
(DO NOT MERGE UNTIL WE VERIFY THAT THIS CHANGE SOLVES THE ORIGINAL PROBLEM)


andreipaduroiu (changes requested):
Blocking to make sure the deleted system tests aren't accidentally merged into the master branch

RaulGracia:
Apart from parameterization, there is little we can do here. The initial intention of this PR was to narrow down the cases for which we want to re-create the Bookkeeper client, specially when the old bookie IP cached issue pops up. However, after sever experiments, it turns out that it is hard to distinguish when the problem comes from caching and old bookie IP or any other reason. In my view, we can leave the current mechanism as it is until we fix the actual problem in the Bookkeeper client itself (at this point, we could remove the Bookkeeper client re-creation mechanism if we think it is worth it).

PR #6037, opened 2021-05-18
Issue 5957: Various log (fluent-bit) fixes.

### LINKED ISSUE #5957: Add option to skip fluent-bit installation.
**Describe the bug**

Fluent-Bit is a fairly common log aggregation tool and forcing the user to install it during a `system` test deployment may result in the collection of the same log multiple times. Depending on the configuration provided, it may also cause conflicts between Kubernetes resources (if things in each deployment map to the same thing).

**To Reproduce**

1. Use the `fluent-bit` chart to install some fluent-bit instance.
2. Run the system test suite.
3. Notice that  the logs are being processed by both fluent-bit deployments.

**Additional information**
Add any other information about the problem here, including the root cause if known.

co-jo:
**Change log description**  

* Pulls out the inline 'fluent-bit rotation' script into its own file.
* Fixes the issue where multiple processes were concurrently writing to the export archive, thus losing logs.
* Add options to pull logs from only the active/running logs, or all logs ever recorded in a given namespace.
* Adds java option to skip the fluent-bit installation.
* Adds java option to specify the `alpine` container repository to use for the fluent-bit watch pod.

**Purpose of the change**  
* Fixes #5957 

**What the code does**  
* See change log description.

**How to verify it**  
* Run the system test suite, and attempt to collect the logs.


test/system/kubernetes/fluentBitSetup.sh
```diff
@@ -27,9 +29,12 @@ KEEP_PVC=false
 NAMESPACE=${NAMESPACE:-"default"}
 NAME=${NAME:-"pravega-fluent-bit"}
 TAR_NAME="pravega-logs-export.tar"
+SKIP_FORCE_ROTATE=${SKIP_FORCE_ROTATE:-"false"}
+ALPINE_IMAGE=${ALPINE_IMAGE:-"alpine"}
```
Tristan1900, line 33: Should we specify the image tag here such as `apline:latest` or just 'alpine' is fine?

anishakj:
@co-jo Has these changes are tested and verified that log rotation is happening?

test/system/kubernetes/fluentBitSetup.sh
co-jo: The fallback tag is `latest` so its implicitly included in a sense. I can add it to be a little more clear.

anishakj (approved):
LGTM

test/system/kubernetes/fluentBitRotater.sh
```diff
@@ -0,0 +1,261 @@
...
+    mv "$name" "$dest"
+  fi
+}
+
+# Catch any 'orphans' -- those .log files that are from an old/restarted container and no longer
+# will receive any new appends. This does not work for restarted pods in a deployment because
+# instead of just the container tag changing, a five character string is also appended to the pod name.
+# So even if the container id is different and the log file is older, we can't differentiate between
+# a restarted pod and simply a longer lived pod.
+orphans() {
```
RaulGracia, line 117: I think it does, but just to confirm, this method compresses the `.log` files left without being rolled on a compressed file, right? If so, good point.

test/system/kubernetes/fluentBitRotater.sh
```diff
@@ -0,0 +1,261 @@
...
+        reclaim
+        end_time=$(date +%s%3N)
+        if [ -n "$rotated" ]; then
+            log "Rotation cycle completed in $((end_time-start_time)) milliseconds."
+        fi
+        echo ""
+    ) 200> /var/lock/rotate.lock 
+}
+
+watch() {
```
RaulGracia, line 227: Who is taking care of starting and monitoring this process? If it crashes, I'm afraid that the PVC with logs will get filled, right?

RaulGracia (approved):
Looks good, just a couple of comments. I also assume that this has been tested with system tests, right?

test/system/kubernetes/fluentBitRotater.sh
co-jo: Thats correct. If a pod restarts it is inevitable that there will be some remaining logs left over (and will never receive any future appends) so this tries to catch those and compress them.

test/system/kubernetes/fluentBitRotater.sh
co-jo: This process is started by a [deployment](https://github.com/pravega/pravega/blob/041d2ba9582aaa65dd9fb8046554163053290000/test/system/kubernetes/fluentBitSetup.sh#L472) so if it crashes it will at least be restarted. If there's a fundamental issue with the logic and it keeps restarting before finishing any reclamation the PVC would overflow yeah. 

I haven't tested it really in a 'hostile' environment so I would want to follow this PR up and make sure that a crash does not prevent successive restarts from failing.

co-jo:
> Looks good, just a couple of comments. I also assume that this has been tested with system tests, right?

I'll deploy another run to be sure.

co-jo:
> > Looks good, just a couple of comments. I also assume that this has been tested with system tests, right?
> 
> I'll deploy another run to be sure.

Issue with downloading logs via the `K8sClient`. Do not merge yet.

andreipaduroiu (changes requested):
Blocking until @co-jo says it's good to merge.

test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/K8SequentialExecutor.java
```diff
@@ -163,6 +153,59 @@ private V1Pod getTestPod(String className, String methodName, String podName) {
         return pod;
     }
 
+    private static String getArgs() {
```
co-jo, line 156: I'm not sure why this hasn't been a problem for other people, or it  has just gone unnoticed, but this makes sure that all `JAVA_OPTS` are properly forwarded to the individual system test job pods. The problem I was running into was due to this, as the `bookkeeperImageVersion` (among others) were not being forwarded to the test pod and caused downstream effects. 

All the options are likely not strictly used, but it would be nice not to worry about this issue going forward.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
```diff
@@ -745,8 +745,10 @@ public void close() {
         public void onFailure(ApiException e, int responseCode, Map<String, List<String>> responseHeaders) {
             if (CONFLICT.getStatusCode() == responseCode || NOT_FOUND.getStatusCode() == responseCode) {
                 log.warn("Exception observed for method {} with response code {}", method, responseCode);
+                log.warn("\tResponseBody: {}", e.getResponseBody());
```
co-jo, line 748: These two statements provide the _actual_ error message generated by the call, instead of just a basic `ApiException` thrown and a response code, which I found particularly unhelpful,

co-jo:
> Blocking until @co-jo says it's good to merge.

Turns out to just be a configuration issue. I detailed the exact problem (and pushed a fix) listed above but it's good to be merged.

test/system/src/main/java/io/pravega/test/system/framework/kubernetes/K8sClient.java
co-jo: For example here a call that returns a 404 response:
`
2021-05-26 20:17:47,892 1225 [OkHttp https://10.100.200.1/...] WARN  i.p.t.s.f.kubernetes.K8sClient - ResponseBody: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"zookeeperclusters.zookeeper.pravega.io \"zookeeper\" not found","reason":"NotFound","details":{"name":"zookeeper","group":"zookeeper.pravega.io","kind":"zookeeperclusters"},"code":404}
`

Instead of:

`
2021-05-26 04:18:21,676 1084 [OkHttp https://10.100.200.1/...] ERROR i.p.t.s.f.kubernetes.K8sClient - Exception observed for method createCustomObject with response code 400                                                                                                                                                                       io.kubernetes.client.openapi.ApiException:
`

RaulGracia:
@co-jo please, complete the "How to verify" section of this PR with the details of tests using real system tests.

andreipaduroiu (approved):

andreipaduroiu (changes requested):
@co-jo Please add copyright headers to all new files you added.

andreipaduroiu (approved):

PR #6039, opened 2021-05-19
Issue 6036: Potential ConcurrentModificationException working with Segment attributes

### LINKED ISSUE #6036: Potential ConcurrentModificationException working with Segment attributes
**Describe the bug**
In a recent longevity experiment, we got this exception from the Segment Store:
```
io.pravega.segmentstore.storage.metadata.StorageMetadataException: Transaction failed
       at io.pravega.segmentstore.storage.metadata.TableBasedMetadataStore.handleException(TableBasedMetadataStore.java:201)
       at io.pravega.segmentstore.storage.metadata.TableBasedMetadataStore.lambda$writeAll$8(TableBasedMetadataStore.java:189)
       at java.base/java.util.concurrent.CompletableFuture.uniExceptionally(Unknown Source)
       at java.base/java.util.concurrent.CompletableFuture$UniExceptionally.tryFire(Unknown Source)
       at java.base/java.util.concurrent.CompletableFuture.postComplete(Unknown Source)
       at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(Unknown Source)
       at io.pravega.common.function.Callbacks.invokeSafely(Callbacks.java:54)
       at io.pravega.common.concurrent.Futures.lambda$exceptionListener$5(Futures.java:277)
       at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(Unknown Source)
       at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(Unknown Source)
       at java.base/java.util.concurrent.CompletableFuture.postComplete(Unknown Source)
       at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(Unknown Source)
       at io.pravega.common.function.Callbacks.invokeSafely(Callbacks.java:54)
       at io.pravega.segmentstore.server.logs.operations.CompletableOperation.fail(CompletableOperation.java:111)
       at io.pravega.segmentstore.server.logs.OperationProcessor$QueueProcessingState.failOperation(OperationProcessor.java:672)
       at io.pravega.segmentstore.server.logs.OperationProcessor.processOperations(OperationProcessor.java:318)
       at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(Unknown Source)
       at java.base/java.util.concurrent.CompletableFuture$Completion.run(Unknown Source)
       at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
       at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
       at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
       at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
       at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
       at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.util.ConcurrentModificationException: null
       at java.base/java.util.HashMap$HashIterator.nextNode(Unknown Source)
       at java.base/java.util.HashMap$EntryIterator.next(Unknown Source)
       at java.base/java.util.HashMap$EntryIterator.next(Unknown Source)
       at io.pravega.common.util.CollectionHelpers$ConvertedIterator.next(CollectionHelpers.java:271)
       at com.google.common.collect.Iterators$ConcatenatedIterator.next(Iterators.java:1364)
       at com.google.common.collect.Iterators$1.next(Iterators.java:141)
       at java.base/java.util.HashMap.putMapEntries(Unknown Source)
       at java.base/java.util.HashMap.<init>(Unknown Source)
       at io.pravega.segmentstore.server.logs.SegmentMetadataUpdateTransaction.getAttributes(SegmentMetadataUpdateTransaction.java:148)
```

The problem is detected `SegmentMetadataUpdateTransaction`, where in its `getAttributes` it does:
```
public Map<UUID, Long> getAttributes() {
        HashMap<UUID, Long> result = new HashMap<>(this.baseAttributeValues); // Problem happens here
        result.putAll(this.attributeUpdates);
        return result;
}
```
The previous `HashMap` constructor iterates over the input `Map` entries to populate the new one. This means that, somewhere else and at the same time, the `baseAttributeValues` map is being modified. While this class does not modify that `Map`, we see that it is actually initialized in the constructor as a reference for `baseMetadata.getAttributes()`.

It turns out that the actual object is of a `StreamSegmentMetadata.AttributesView` object. While that class is meant to be synchronized and non-modifiable, we also see that the implementation of `entrySet()`, used in the `HashMap` constructor, is like this:
```
        public Set<Map.Entry<UUID, Long>> entrySet() {
            synchronized (StreamSegmentMetadata.this) {
                return CollectionHelpers.joinSets(
                        coreAttributes.entrySet(), Callbacks::identity,
                        extendedAttributes.entrySet(), e -> new AbstractMap.SimpleImmutableEntry<>(e.getKey(), e.getValue().value));
            }
        }
```
So, what it is used to any class using a reference to this object is whather is returned from `CollectionHelpers.joinSets`,  which in turn says that:

> Returns an unmodifiable Set View made up of the given Sets while translating the items into a common type. The returned
> Set View does not copy any of the data from any of the given Sets, therefore any changes in the two Sets will be
> reflected in the View.

Therefore, what other classes will use when calling to `entrySet()` is not a copy, but a reference to the `coreAttributes` and `extendedAttributes` maps in `StreamSegmentMetadata`. As these maps can be modified along a Segment's lifetime, if any class is iterating the `StreamSegmentMetadata.AttributesView` at the same time that `coreAttributes` or `extendedAttributes` are modified, could get a `ConcurrentModificationException`.

**To Reproduce**
Hard to reproduce. It would require a concurrent iteration and modification of a Segment's attributes.

**Screenshots**
n/a

**Additional information**
Using `ConcurrentHashMap` for Segment attributes may prevent this problem from happening.

RaulGracia:
**Change log description**  
Use concurrent maps in StreamSegmentMetadata to store attributes.

**Purpose of the change**  
Fixes #6036.

**What the code does**  
We use `ConcurrentHashMap` in `StreamSegmentMetadata` to store `coreAttributes` and `extendedAttributes`, as they could be modified and iterated concurrently.

**How to verify it**  
Added a new test to verify that concurrent updates and accesses to `StreamSegmentMetadata` attributes are safe.


andreipaduroiu (commented):
You need to look carefully where this exception is being

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
```diff
@@ -116,7 +117,7 @@
         this.deletedInStorage = baseMetadata.isDeletedInStorage();
         this.deleted = baseMetadata.isDeleted();
         this.baseAttributeValues = baseMetadata.getAttributes();
-        this.attributeUpdates = new HashMap<>();
+        this.attributeUpdates = new ConcurrentHashMap<>(); // This one could be iterated and modified concurrently.
```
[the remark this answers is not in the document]
RaulGracia: I know that according to the stack trace this class was the one that got the exception, but was not the root cause of it. However, I fount that it also contained a `HashMap` that was being iterated and modified at different public methods, which could lead to a similar problem than in `StreamSegmentMetadata`. But you are right, this class is marked as `@NonThreadSafe`, so users of this class are aware that this kind of issues can happen without external synchronization. Will revert it.

andreipaduroiu (approved):

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadataTests.java
```diff
@@ -212,6 +218,47 @@ private void checkAttributesEqual(Map<UUID, Long> expected, Map<UUID, Long> actu
...
+        // Step 1: initial set of attributes.
+        Random rnd = new Random(0);
+        val expectedAttributes = generateAttributes(rnd);
+
+        metadata.updateAttributes(expectedAttributes);
+        ExecutorService executorService = null;
+        CompletableFuture<Void> iterateOverView;
+        CompletableFuture<Void> modifyBackingAttributes;
+        try {
+            executorService = ExecutorServiceHelpers.newScheduledThreadPool(4, "testAttributes");
```
andreipaduroiu, line 238: You can just do `@Cleanup("shutdownNow")` to avoid the `try-finally`. We do that all over the place.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadataTests.java
```diff
@@ -212,6 +218,47 @@ private void checkAttributesEqual(Map<UUID, Long> expected, Map<UUID, Long> actu
...
+            Map<UUID, Long> attributesView = metadata.getAttributes();
+            iterateOverView = CompletableFuture.runAsync(() -> {
+                for (int i = 0; i < 100; i++) {
+                    new HashMap<>(attributesView); // Internally, this iterates over the input map.
+                }
+            }, executorService);
+            modifyBackingAttributes = CompletableFuture.runAsync(() -> {
+                for (int i = 0; i < 100; i++) {
+                    metadata.updateAttributes(generateAttributes(new Random(i)));
```
andreipaduroiu, line 247: I see what you're trying to do there, but the thing is you are not really guaranteed to test your change at all. You may or may not have a concurrent iteration with a modification. 

This is a non-deterministic way of verifying things. May I suggest a simpler approach:
1. Create your metadata and put some attributes in it.
2. Get an iterator of those attributes (and maybe fetch 1 item)
3. Modify the attributes
4. Attempt to get the remainder of those attributes from the iterator.

You'll see that the "old code" would throw the `ConcurrentModificationException` while the new one should not. The advantage of this test is that it's deterministic and it explicitly verifies what you are changing.

andreipaduroiu (changes requested):

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadataTests.java
RaulGracia: This has been removed in the new test.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/StreamSegmentMetadataTests.java
RaulGracia: Right, I changed the test according to your suggestion.

andreipaduroiu (approved):

PR #6043, opened 2021-05-19
Issue 6042: (KVT Beta2) Keeping track of the number of extended attributes per Segment

### LINKED ISSUE #6042: KVT Beta 2: Keep Track of number of Attributes per Segment
The B+TreeIndex does not have an easy way to get the number of entries stored in its index. The only way we can do it is by iterating through all entries. 

We need a way to maintain basic statistics (including at least number of entries) and persist these durably alongside the rest of the data in the index. This change must be backwards compatible (do not break existing segments), but it is OK if statistics are not enabled for existing segments. We only need that for the new Fixed-Key-Length Table Segments.

andreipaduroiu:
**Change log description**  
Keeping track of the number of Extended Attributes per Segment (if enabled)

**Purpose of the change**  
Fixes #6042.

**What the code does**  
BTreeIndex has a new option named `maintainStatistics`. If enabled, it will keep track of the number of Entries and Pages (nodes) within it. This feature only works if the BTreeIndex has been originally created with this option (so upgrading existing indices will not work - nor is it required). 

The Statistics are stored immediately after the Root Page (and implicitly below the Index Footer/Root Pointer) every time the index persists an update. This is loaded in the `initialize` method (if it exists) and maintained in memory for quick reference.

This is not yet wired up into upstream code. When implemented in a subsequent PR, the `SegmentAttributeBTreeIndex` will expose the `BTreeIndex.getStatistics().getEntryCount()` as its own method, then in `StreamSegmentContainer.DirectSegmentWrapper`, we'll expose it for whomever needs it.

As of right now, this feature is disabled for all segments. When Fixed-Key-Length Table Segments are introduced (next PR), we'll enable it for those ones only.

**How to verify it**  
Existing unit tests updated to verify new functionality. New unit tests added to validate backward compatibility.

### DIFF
--- a/common_server/src/main/java/io/pravega/common/util/btree/BTreeIndex.java
+++ b/common_server/src/main/java/io/pravega/common/util/btree/BTreeIndex.java
@@ -876,7 +876,6 @@ public class BTreeIndex {
     private void setState(long length, long rootPageOffset, int rootPageLength) {
         this.state = new IndexState(length, rootPageOffset, rootPageLength);
         log.debug("{}: IndexState: {}, Stats: {}.", this.traceObjectId, this.state, this.statistics);
-        //System.out.println(String.format("%s: IndexState: %s, Stats: %s.", this.traceObjectId, this.state, this.statistics));
     }
 
     private long getFooterOffset(long indexLength) {
--- a/common_server/src/main/java/io/pravega/common/util/btree/BTreeIndex.java
+++ b/common_server/src/main/java/io/pravega/common/util/btree/BTreeIndex.java
@@ -22,8 +22,8 @@ import io.pravega.common.util.AsyncIterator;
 import io.pravega.common.util.BufferViewComparator;
 import io.pravega.common.util.ByteArraySegment;
 import io.pravega.common.util.IllegalDataFormatException;
+import java.io.IOException;
 import java.time.Duration;
-import java.util.AbstractMap;
 import java.util.ArrayList;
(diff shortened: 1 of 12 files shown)


common_server/src/main/java/io/pravega/common/util/btree/BTreeIndex.java
```diff
@@ -803,16 +866,17 @@ private ByteArraySegment generateMinKey() {
...
                 });
     }
 
     private void setState(long length, long rootPageOffset, int rootPageLength) {
-        IndexState s = new IndexState(length, rootPageOffset, rootPageLength);
-        this.state.set(s);
-        log.debug("{}: IndexState: {}.", this.traceObjectId, s);
+        this.state = new IndexState(length, rootPageOffset, rootPageLength);
+        log.debug("{}: IndexState: {}, Stats: {}.", this.traceObjectId, this.state, this.statistics);
+        //System.out.println(String.format("%s: IndexState: %s, Stats: %s.", this.traceObjectId, this.state, this.statistics));
```
RaulGracia, line 879: Please, remove this line.

RaulGracia (approved):
LGTM (just one minor comment).

common_server/src/main/java/io/pravega/common/util/btree/BTreeIndex.java
andreipaduroiu: Oops. Fixed.

PR #6048, opened 2021-05-20
Issue 6045: Storage - StorageReadManager does not correctly handle failures in Storage::openRead

### LINKED ISSUE #6045: Storage - StorageReadManager does not correctly handle failures in Storage::openRead
**Describe the bug**
Currently the `StorageReadManager `saves the return value of `Storage::openRead` in a member field.
This instance of `CompletableFuture `is used for all the subsequent calls.

_Expected behavior_

- In case failure in `Storage::openRead` the member variable is not initialized and hence next time attempt will be made to re-initialize it
- The code should retry only the transient errors and not permeant errors.
- The initialization should be thread/safe and/or serialized so that there is no race between two concurrent calls
_Actual behavior_
- In case failure the failed future is saved and then `openRead `is never retried again.  This leads to situation where once failed the `StorageReadManager` is stuck in failed state forever.
- Also, there is a possibility of race condition during initialization which can results in concurrent calls to `openRead`

**To Reproduce**
This bug is hard to reproduce in end to end testing. However it should be very easy to reproduce in a unit test.

Inspecting code here shows the problem
- https://github.com/pravega/pravega/blob/r0.9/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java#L195
- https://github.com/pravega/pravega/blob/r0.9/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java#L136

**Screenshots**
N/A

**Additional information**
N/A

sachin-j-joshi:

**Change log description**  
Correctly handle failures in Storage::openRead when initializing StorageReadManager by caching handle only when openRead call is successful.
**Purpose of the change**  
Fixes #6045 

**What the code does**  
Cache handle only when openRead call is successful.

Failures in openRead are rare and may include mostly transient but also non-transient errors.  The best course of action here is to not cache exception state.

**How to verify it**  
Existing unit test has been modified for given scenario.

### DIFF
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java
@@ -201,11 +201,14 @@ public class StorageReadManager implements AutoCloseable {
     private CompletableFuture<SegmentHandle> getHandle() {
         synchronized (this.lock) {
             if (this.handle == null) {
-                return storage.openRead(this.segmentName)
-                        .thenApply(h -> {
-                            // only cache when successful.
-                            this.handle =  CompletableFuture.completedFuture(h);
-                            return h;
+                this.handle = storage.openRead(this.segmentName)
+                        .whenComplete((h, ex) -> {
+                            if (ex != null) {
+                                synchronized (this.lock) {
+                                    log.debug("{}: storage.openRead failed for {}. Resetting handle. {}", this.traceObjectId, this.segmentName, ex);
+                                    this.handle = null;
+                                }
+                            }
                         });
             }
 
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java
@@ -201,7 +201,12 @@ public class StorageReadManager implements AutoCloseable {
     private CompletableFuture<SegmentHandle> getHandle() {
         synchronized (this.lock) {
             if (this.handle == null) {
-                this.handle = storage.openRead(this.segmentName);
+                return storage.openRead(this.segmentName)
+                        .thenApply(h -> {
+                            // only cache when successful.
+                            this.handle =  CompletableFuture.completedFuture(h);
+                            return h;
+                        });
             }
 
             return this.handle;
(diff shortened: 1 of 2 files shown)


segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java
```diff
@@ -201,7 +201,12 @@ private Request findOverlappingRequest(Request request) {
     private CompletableFuture<SegmentHandle> getHandle() {
         synchronized (this.lock) {
             if (this.handle == null) {
-                this.handle = storage.openRead(this.segmentName);
+                return storage.openRead(this.segmentName)
+                        .thenApply(h -> {
+                            // only cache when successful.
+                            this.handle =  CompletableFuture.completedFuture(h);
```
RaulGracia, line 207: Nit: extra space after `=`.

RaulGracia (approved):
LGTM, one minor comment.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java
```diff
@@ -201,7 +201,12 @@ private Request findOverlappingRequest(Request request) {
     private CompletableFuture<SegmentHandle> getHandle() {
         synchronized (this.lock) {
             if (this.handle == null) {
-                this.handle = storage.openRead(this.segmentName);
+                return storage.openRead(this.segmentName)
```
andreipaduroiu, line 204: If you get multiple concurrent requests for reading from this segment, this will `openRead` for each such request until one of them succeeds and assigns the handle. 

The previous approach was storing the future in the `this.handle` for the first request coming in, and any requests that come in while `openRead` is executing will piggyback onto that one.

Please fix this code to what it was before, but add an exception handler to reset `this.handle` to `null` if an exception happens during `openRead`.

andreipaduroiu (changes requested):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/reading/StorageReadManager.java
sachin-j-joshi: Thanks. Good catch. Totally forgot about that.

andreipaduroiu (approved):

PR #6057, opened 2021-05-21
Issue 6046: SLTS - Add metrics for count of read and write requests.

### LINKED ISSUE #6046: SLTS - Add metrics for count of read and write requests

Currently we have total bytes read and latency related metrics. But don't have the metric for number of read and write requests.

We also need following metrics in SLTS
- count of read requests
- count of write requests

This is a primary number and it is hard and error-prone to derive it indirectly

sachin-j-joshi:

**Change log description**  
Add metrics for count of read and write requests.
Do not count system segments metrics with user segment metrics.
Add missing metric for concat.

**Purpose of the change**  
Fixes #6046

**What the code does**  
Add metrics for count of read and write requests.
Do not count system segments metrics with user segment metrics.
Add missing metric for concat.

**How to verify it**  
Build should pass

### DIFF
--- a/shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
+++ b/shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
@@ -113,6 +113,9 @@ public final class MetricsNames {
     public static final String STORAGE_READ_BYTES = PREFIX + "segmentstore.storage.read_bytes";          // Counter
     public static final String STORAGE_WRITE_BYTES = PREFIX + "segmentstore.storage.write_bytes";        // Counter
     public static final String STORAGE_CONCAT_BYTES = PREFIX + "segmentstore.storage.concat_bytes";      // Counter
+
+    public static final String STORAGE_READ_COUNT = PREFIX + "segmentstore.storage.read_count";      // Counter
+    public static final String STORAGE_WRITE_COUNT = PREFIX + "segmentstore.storage.write_count";      // Counter
     public static final String STORAGE_CREATE_COUNT = PREFIX + "segmentstore.storage.create_count";      // Counter
     public static final String STORAGE_DELETE_COUNT = PREFIX + "segmentstore.storage.delete_count";      // Counter
     public static final String STORAGE_CONCAT_COUNT = PREFIX + "segmentstore.storage.concat_count";      // Counter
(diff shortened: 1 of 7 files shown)


shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -152,12 +155,18 @@
     public static final String SLTS_SYSTEM_READ_BYTES = PREFIX + "segmentstore.storage.slts.system_read_bytes";     // Counter
     public static final String SLTS_SYSTEM_WRITE_BYTES = PREFIX + "segmentstore.storage.slts.system_write_bytes";   // Counter
     public static final String SLTS_CONCAT_BYTES = PREFIX + "segmentstore.storage.slts.concat_bytes";      // Counter
+
+    public static final String SLTS_READ_COUNT = PREFIX + "segmentstore.storage.slts.read_count";      // Counter
```
andreipaduroiu, line 159: What is the difference between these ant he ones 40 lines above? These ones have `.slts` in their name. 

If these are for the system segments/journal, let's name them as such. If they are for the metadata tables, let's name them as such. 

IMO, since we already have a lot of SLTS/Storage metrics, it is very important to clearly differentiate what they do by their name, otherwise someone unfamiliar with the inner workings of this will easily get confused.

andreipaduroiu (changes requested):

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
sachin-j-joshi: `segmentstore.storage.`  metrics record only for the individual single chunk storage call

`segmentstore.storage.slts.`  metrics record for entire end to end call to SLTS.
This is required because single SLTS call will often have multiple steps - access table segment metadata and potentially results in multiple calls to ChunkStorage. 
Rolling storage did not have these end to end metrics but we already have these second set for SLTS.

For some reason we did not have any metrics for count of read and write calls

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -152,12 +155,18 @@
...
+    public static final String SLTS_WRITE_COUNT = PREFIX + "segmentstore.storage.slts.write_count";      // Counter
     public static final String SLTS_CREATE_COUNT = PREFIX + "segmentstore.storage.slts.create_count";      // Counter
     public static final String SLTS_DELETE_COUNT = PREFIX + "segmentstore.storage.slts.delete_count";      // Counter
     public static final String SLTS_CONCAT_COUNT = PREFIX + "segmentstore.storage.slts.concat_count";      // Counter
     public static final String SLTS_TRUNCATE_COUNT = PREFIX + "segmentstore.storage.slts.truncate_count";  // Counter
     public static final String SLTS_SYSTEM_TRUNCATE_COUNT = PREFIX + "segmentstore.storage.slts.system_truncate_count";  // Counter
 
+    public static final String SLTS_SYSTEM_READ_COUNT = PREFIX + "segmentstore.storage.slts.system_read_count";      // Counter
```
sachin-j-joshi, line 167: segmentstore.storage.slts.system_ metrics are used to track writes to storage metadata segments.

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
andreipaduroiu: OK, I understand now. Would it make sense to rename the first one to include `chunk` in it?

sachin-j-joshi: Agree. But may be later. 
I want to keep naming convention matching other metrics (Eg. concat/truncate/delete counts).
We should change name for all of those and also update existing dashboards etc. That would be a bigger and separate change in itself.

RaulGracia (changes requested):
Looks good, but please add a test to verify that the new metrics are accounting what you would expect.

andreipaduroiu:
@sachin-j-joshi Are we still pursuing this PR? I don't see any activity in the last 2 months.

sachin-j-joshi:
This may not be required at this point.

Issue #6010, opened 2021-05-12
Implement Stream tag APIs on the Pravega Client


shrids:
**Describe the solution you'd like**
Implement Stream tags APIs on the Pravega Client

We need to:
- Enable APIs on the Controller Client.
- Enable APIs on the StreamManager.

RaulGracia:
Closed via #6059.

PR #6059, opened 2021-05-24
Issue 6009, 6010: Enable Stream Tags.

### LINKED ISSUE #6009: Implement StreamTag APIs on the Controller.
Implement StreamTag APIs on the Controller based on the PDP.

We need to:
- Ensure the StreamConfigurationRecord stores the Tags for a given stream.
- Enable a mechanism to store the inverted index for the tags on the Controller using Table Segments
- Enable GRPC APIs for the tag-related APIs.

shrids:
**Change log description**  
Enable StreamTags on Pravega. Enable tag related GRPC APIs on the controller and ensure these APIs can be invoked by the
PravegaClient.

**Purpose of the change**  
Fixes #6009 #6010 

**What the code does**  
The code enables StreamTag feature for Pravega. This PR includes the following changes.

1.  Client-side changes:
StreamManager APIs:

- The user can now pass stream tags during the stream creation using `StreamConfiguration`.  
```java
   /**
     * Creates a new stream
 
     * @param scopeName  The name of the scope to create this stream in.
     * @param streamName The name of the stream to be created.
     * @param config The configuration the stream should use.
     * @return True if stream is created
     */
    boolean createStream(String scopeName, String streamName, StreamConfiguration config);

    /**
     * Change the configuration for an existing stream.
     *
     * @param scopeName  The name of the scope to create this stream in.
     * @param streamName The name of the stream who's config is to be changed.
     * @param config     The new configuration.
     * @return True if stream configuration is updated
     */
    boolean updateStream(String scopeName, String streamName, StreamConfiguration config);

```

- A new API has been added to fetch the tags for the given Stream.
```java
 /**
     * Gets the Tags associated with a stream.
     *
     * @param scopeName Scope name.
     * @param streamName Stream name.
     * @return Tags associated with the stream.
     */
    Set<String> getStreamTags(String scopeName, String streamName);
```
-  The existing getStreamInfo() API now returns the current stream configuration along with other data.
```java 
  /**
     * Get information about a given Stream, {@link StreamInfo}.
     * This includes {@link StreamCut}s pointing to the current HEAD and TAIL of the Stream and the current 
     * {@link StreamConfiguration}
     *
     * @param scopeName The scope of the stream.
     * @param streamName The stream name.
     * @return stream information.
     */
    @Beta
    StreamInfo getStreamInfo(String scopeName, String streamName);
```
- New API to list streams with the given tag in a scope.
```java 
/**
     * Gets an iterator to list all streams with the provided tag.
     *
     * @param scopeName The name of the scope for which to list streams in.
     * @param tagName The name of the tag.
     *
     * @return Iterator of Stream to iterator over all streams in scope with the provided tag.
     */
    Iterator<Stream> listStreams(String scopeName, String tagName);
```

2. Controller side changes:

The controller GRPC server has now enabled two new APIs.
- API to fetch the current configuration of the provided Stream. This includes the tag information also.
```java
// Fetch the current Stream configuration.
public void getStreamConfiguration(StreamInfo request, StreamObserver<StreamConfig> responseObserver) ;
```
- API to list the streams associated with the given tag in a scope. The user can pass a continuationtoken which helps the client to iterate over the results.
```java
    public void listStreamsInScopeForTag(Controller.StreamsInScopeWithTagRequest request,
                                   StreamObserver<Controller.StreamsInScopeResponse> responseObserver);
```

Compatibility:
- The changes are backward compatible. i.e the older version of the client will work fine with the newer server with tag-related changes.

**How to verify it**  


controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
...
+        // Get unversioned keys.
+        final List<TableSegmentKey> keys = tableKeys.stream()
+                                                    .map(k -> TableSegmentKey.unversioned(k.getBytes(StandardCharsets.UTF_8)))
+                                                    .collect(Collectors.toList());
+        return segmentHelper.readTable(tableName, keys, authToken.get(), requestId) // read the latest version of keys.
+                            .whenComplete((v, ex) -> releaseKeys(keys))
+                            .thenCompose(currentEntries -> {
+
+                                List<TableSegmentEntry> updatedList = currentEntries.stream().map(entry -> {
+                                    String k = entry.getKey().getKey().toString(Charset.defaultCharset());
```
andreipaduroiu, line 316: You use `defaultCharset` here yet you use `UTF_8` above. Default Charset may differ from machine to machine so the cluster data may not be portable in case we want to failover the cluster to another location. Please use a hardcoded encoding (UTF_8 sounds good) instead of relying on system-wide defaults.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
```diff
@@ -285,6 +286,31 @@ PravegaTablesScope newScope(final String scopeName) {
                 executor);
     }
 
+    @Override
+    public CompletableFuture<Void> addStreamTagsToIndex(String scope, String streamName, StreamConfiguration config,
+                                                        OperationContext ctx, Executor executor) {
+        OperationContext context = getOperationContext(ctx);
+        if (streamName.startsWith(NameUtils.INTERNAL_NAME_PREFIX)) {
+            // Tags are not allowed on internal streams.
+            return CompletableFuture.completedFuture(null);
```
andreipaduroiu, line 295: Should we throw an exception or at least log a warning that this was attempted?

controller/src/main/java/io/pravega/controller/store/stream/records/TagRecord.java
```diff
@@ -0,0 +1,135 @@
...
+    private static byte[] compressArrayOption(final Set<String> set) throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        OutputStream out = new DeflaterOutputStream(baos);
+        for (String s : set) {
+            out.write(s.getBytes(StandardCharsets.UTF_8));
+            out.write(',');
+        }
+        out.flush();
+        out.close();
+        return baos.toByteArray();
```
andreipaduroiu, line 121: Tip:
Use `ByteBufferOutputStream`, then get the contents as a `ByteArraySegment`, then pass that to `revisionDataOutput.writeBuffer`. 

That will save you the trouble of all the extra array copies that happen when you invoke `toByteArray`. 

`writeBuffer/writeArray` are equivalent and both can be read back with `readArray` (like you already have)

controller/src/main/java/io/pravega/controller/store/stream/records/TagRecord.java
```diff
@@ -0,0 +1,135 @@
...
+
+        @Override
+        protected TagRecord.TagRecordBuilder newBuilder() {
+            return TagRecord.builder();
+        }
+    }
+
+    private static byte[] compressArrayOption(final Set<String> set) throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        OutputStream out = new DeflaterOutputStream(baos);
```
andreipaduroiu, line 114: I believe we can do better here. Wrap `out` in a `DataOutputStream` and use the `.writeUTF`. That will save you from having to generate an array with `s.getBytes`.

controller/src/main/java/io/pravega/controller/store/stream/records/TagRecord.java
```diff
@@ -0,0 +1,135 @@
...
+        out.flush();
+        out.close();
+        return baos.toByteArray();
+    }
+
+    private static Set<String> decompressArrayOption(final byte[] compressed) throws IOException {
+        @Cleanup
+        InputStream in = new InflaterInputStream(new ByteArrayInputStream(compressed));
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        IOUtils.copy(in, baos);
```
andreipaduroiu, line 128: This has a number of unnecessary array copies Wrap `in` in a `DataInputStream` and use `readUTF` (the opposite of what I suggested you do above when you generate this). This will save you the trouble of having to call `IOUtils.copy` and `baos.toString` and `streams.split(...)`. 

Actually you don't even need to encode the comma (separator) in this case as `writeUTF` and `readUTF` take care of that for you.

controller/src/test/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasksTest.java
```diff
@@ -135,7 +135,7 @@
     private static final String SCOPE = "scope";
     private static final String STREAM = "stream1";
     @Rule
-    public Timeout globalTimeout = new Timeout(30, TimeUnit.HOURS);
+    public Timeout globalTimeout = new Timeout(1, TimeUnit.MINUTES);
```
andreipaduroiu, line 138: LOL!

andreipaduroiu (changes requested):

client/src/main/java/io/pravega/client/stream/StreamConfiguration.java
```diff
@@ -59,7 +68,52 @@
...
+            }
+            switch (tags == null ? 0 : tags.size()) {
+                case 0:
+                    tagsSet = Collections.emptySet();
+                    break;
+                case 1:
+                    tagsSet = Collections.singleton(this.tags.get(0));
+                    break;
+                default:
+                    tagsSet = new HashSet<>(this.tags.size());
```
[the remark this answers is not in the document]
shrids: done, simplified the code

client/src/main/java/io/pravega/client/admin/StreamManager.java
```diff
@@ -137,6 +138,25 @@ public static StreamManager create(ClientConfig clientConfig) {
...
+    Iterator<Stream> listStreams(String scopeName, String tagName);
+
+    /**
+     * Gets the Tags associated with a stream.
+     *
+     * @param scopeName Scope name.
+     * @param streamName Stream name.
+     * @return Tags associated with the stream.
+     */
+    Set<String> getStreamTags(String scopeName, String streamName);
```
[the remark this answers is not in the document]
shrids: yes, updated the method signature.

client/src/main/java/io/pravega/client/stream/StreamConfiguration.java
```diff
@@ -59,7 +68,52 @@
...
+        public StreamConfiguration build() {
+            Set<String> tagSet = validateTags(this.tags);
+            return new StreamConfiguration(this.scalingPolicy, this.retentionPolicy, this.timestampAggregationTimeout, tagSet);
+        }
+
+        private Set<String> validateTags(List<String> tags) {
+            Set<String> tagsSet;
+            if (tags != null) {
+                Preconditions.checkArgument(tags.size() < MAX_TAG_COUNT, "Maximum number of tags allowed is 128");
+                tags.forEach(tag -> Preconditions.checkArgument(tag.length() < MAX_TAG_LENGTH, "Maximum length of a tag allowed is 256"));
```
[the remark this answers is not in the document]
shrids: Any UTF8 string is allowed as a tag.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: fixed it.

controller/src/main/java/io/pravega/controller/store/stream/records/TagRecord.java
shrids: Thanks 👍  
Improved the code.

controller/src/main/java/io/pravega/controller/store/stream/records/TagRecord.java
shrids: Yes, using writeUTF makes it more better. Improved the code here.

controller/src/main/java/io/pravega/controller/store/stream/records/TagRecord.java
shrids: ok, fixed it.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
shrids: The methods `addStreamTagsToIndex` and `removeTagsFromIndex` are invoked as part of the Controller workflows during stream creation, deletion, or update for all streams.. This check ensures this step is skipped for internal streams.

andreipaduroiu (approved):
My comments have been addressed - thank you.

@shiveshr This is in your hands now.

RaulGracia:
@shiveshr can you have a look to this one please?

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -121,42 +143,53 @@ public String getName() {
...
-                getQualifiedTableName(INTERNAL_SCOPE_NAME, scopeName, String.format(KVTABLES_IN_SCOPE_TABLE_FORMAT, id.toString())));
+                                                getQualifiedTableName(INTERNAL_SCOPE_NAME, scopeName, String.format(KVTABLES_IN_SCOPE_TABLE_FORMAT, id.toString())));
+    }
+
+    public CompletableFuture<String> getStreamTagsInScopeTableName(String stream, OperationContext context) {
+        return getId(context).thenApply(id ->
+                                                getQualifiedTableName(INTERNAL_SCOPE_NAME, scopeName, String.format(STREAM_TAGS_IN_SCOPE, id.toString(), HASH.hashToBucket(stream, 25))));
+    }
+
+    public CompletableFuture<List<String>> getStreamTagsInScopeTableName(OperationContext context) {
```
shiveshr, line 175: nit:
i see that one is returning all the chunk names while the other hashes the stream name and returns the chunk specific table name.. 
so these are two different functions and the overloading them is perhaps not a good idea and should be given different names.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -180,37 +213,73 @@ public String getName() {
...
+        return getNextTableSegment(tag, continuationToken, context).thenCompose(pair -> {
+            if (pair.getLeft().isEmpty() && !pair.getRight().contains(LAST_TAG_CHUNK)) {
+                return listStreamsForTag(tag, pair.getRight(), executor, context);
+            } else {
+                return CompletableFuture.completedFuture(pair);
+            }
+        });
+    }
+
+    CompletableFuture<Pair<List<String>, String>> getNextTableSegment(String tag, String ct, OperationContext context) {
```
shiveshr, line 259: the method name says get next table segment. and it returns a pair - list of strings and a continuation token. 
could you please add javadoc explaining what this is doing here. i believe it is returning the list of stream names from the continuation token, right?

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -180,37 +213,73 @@ public String getName() {
...
+        return getStreamTagsInScopeTableName(context).thenApply(
+                strings -> {
+                    if (ct.isEmpty()) {
+                        return strings.get(0);
+                    } else {
+                        // return next index
+                        return strings.get(strings.indexOf(ct) + 1);
```
shiveshr, line 266: i dont quite follow the logic here? what does strings contain -- all the chunk ids?
and we are saying we want to read from the next chunk? right?
perhaps adding some comments in the code explaining the logic here would be good.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
                 .whenComplete((r, ex) -> releaseEntries(entries));
     }
 
+    public CompletableFuture<List<Version>> removeValues(final String tableName, Set<String> tableKeys, final String appendValue, long requestId) {
+        Supplier<String> errorMessage = () -> String.format("remove values: on table: %s", tableName);
+
+        // read values and add values
```
shiveshr, line 255: nit: this comment says read and add while the call below is read and remove.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
                 .whenComplete((r, ex) -> releaseEntries(entries));
     }
 
+    public CompletableFuture<List<Version>> removeValues(final String tableName, Set<String> tableKeys, final String appendValue, long requestId) {
```
shiveshr, line 252: please add java doc explaining what this method is about. 
i believe its taking some keys in a table segment and doing a get and put on the value against those keys, right?
when this is a removeValues method, why does it have an append value as a param?

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
...
+        Supplier<String> errorMessage = () -> String.format("remove values: on table: %s", tableName);
+
+        // read values and add values
+        return withRetries(() -> removeValuesTables(tableName, appendValue, tableKeys, requestId)
+                .thenApplyAsync(x -> {
+                    log.trace("appendValues {} updated to table {} with version {}", appendValue, tableName, x);
```
shiveshr, line 258: nit: remove values

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
                 .whenComplete((r, ex) -> releaseEntries(entries));
     }
 
+    public CompletableFuture<List<Version>> removeValues(final String tableName, Set<String> tableKeys, final String appendValue, long requestId) {
+        Supplier<String> errorMessage = () -> String.format("remove values: on table: %s", tableName);
+
+        // read values and add values
+        return withRetries(() -> removeValuesTables(tableName, appendValue, tableKeys, requestId)
```
shiveshr, line 256: be careful with this withRetries method here.. with conditional updates the retries should be used judiciously or you can get a write conflict on a retry.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
...
+                .thenApplyAsync(x -> {
+                    log.trace("appendValues {} updated to table {} with version {}", appendValue, tableName, x);
+                    return x.stream().map(v -> (Version) new Version.LongVersion(v.getSegmentVersion())).collect(Collectors.toList());
+                }), errorMessage, true, requestId);
+    }
+
+
+    public CompletableFuture<List<Version>> appendValues(final String tableName, Set<String> tableKeys, final String appendValue, long requestId) {
```
shiveshr, line 264: same here, please add the java doc.. 
do we need this method in this class or is this business logic very specific to the tags and should be moved there?

i believe you are appending a new value to the end of existing values.. where this method would do a get and put. right?
this is because this method is implicitly assuming that the value is a List<Strings> to which it needs to append. 
if we want to create a helper method of this sort, it should use generics for the value and clearly call out that it works when the value is actually serialized as a list. 

Ideally i would want to move this logic to either the caller or some helper class for that business logic if it has to be reused. it wouldnt be a natural fit for the storehelper which is generic for all types of values. only thing it fixes is that keys be strings.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
...
+
+        CompletableFuture<List<TableSegmentEntry>> f1 = segmentHelper.readTable(tableName, keys, authToken.get(), requestId) // read the latest version of keys.
+                                                                     .whenComplete((v, ex) -> releaseKeys(keys));
+        f1.thenCompose(currentEntries -> {
+            List<TableSegmentEntry> updatedList = currentEntries.stream().map(entry -> {
+                String k = entry.getKey().getKey().toString(Charset.defaultCharset());
+                byte[] array = getArray(entry.getValue());
+                byte[] updatedBytes;
+                if (array.length == 0) {
+                    updatedBytes = TagRecord.builder().tagName(k).build().toBytes();
```
shiveshr, line 289: again, it is assuming that the value is of type tagRecord. this is very business logic specific and should go into tag specific caller class.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +249,89 @@ public void invalidateCache(String table, String key) {
...
+                                    } else {
+                                        TagRecord record = TagRecord.fromBytes(array);
+                                        updatedBytes = record.toBuilder().stream(appendValue).build().toBytes();
+                                    }
+                                    return TableSegmentEntry.versioned(k.getBytes(StandardCharsets.UTF_8),
+                                                                       updatedBytes,
+                                                                       entry.getKey().getVersion().getSegmentVersion());
+
+                                }).collect(Collectors.toList());
+                                return segmentHelper.updateTableEntries(tableName, updatedList, authToken.get(), requestId)
```
shiveshr, line 330: i think the reason you have put this in here is because we dont have updateTableEntries as a store helper method. 
I suggest you add that helper method which can take a batch of keys and values and update them. 
and then you can write your higher level business logic to use it.
also this updateEntries should be a generalized form of udateEntry method where we do it for more than one key value pair. maybe we can restrict its implementation to expect a single value type.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStreamMetadataStore.java
```diff
@@ -223,7 +224,17 @@ Scope newScope(final String scopeName) {
     }
 
     @Override
-    public CompletableFuture<Void> addReaderGroupToScope(String scopeName, String rgName, UUID readerGroupId, 
+    public CompletableFuture<Void> addStreamTagsToIndex(String scope, String name, StreamConfiguration config, OperationContext context, Executor executor) {
+        return CompletableFuture.completedFuture(null);
```
shiveshr, line 228: throw unsupported instead of returning null

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStreamMetadataStore.java
```diff
@@ -223,7 +224,17 @@ Scope newScope(final String scopeName) {
...
+    }
+
+    @Override
+    public CompletableFuture<Void> removeTagsFromIndex(String scope, String name, Set<String> tagsRemoved, OperationContext context, Executor executor) {
+        return CompletableFuture.completedFuture(null);
```
shiveshr, line 233: same here.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
```diff
@@ -285,6 +286,31 @@ PravegaTablesScope newScope(final String scopeName) {
...
+
+    @Override
+    public CompletableFuture<Void> removeTagsFromIndex(String scope, String streamName, Set<String> tagsRemoved,
+                                                       OperationContext ctx, Executor executor) {
+        OperationContext context = getOperationContext(ctx);
+        if (streamName.startsWith(NameUtils.INTERNAL_NAME_PREFIX)) {
+            // Tags are not allowed on internal streams.
+            return CompletableFuture.completedFuture(null);
+        } else {
+            return Futures.completeOn(((PravegaTablesScope) getScope(scope, context))
```
shiveshr, line 310: what happens when the last stream with the tag is removed. do we also remove the tag entry?

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
```diff
@@ -174,20 +175,30 @@ ZKScope newScope(final String scopeName) {
...
+    public CompletableFuture<Pair<List<String>, String>> listScopes(String continuationToken, int limit,
                                                                     Executor executor, long requestId) {
         // Pagination not supported for zk based store. 
         throw new UnsupportedOperationException();
     }
 
     @Override
-    public CompletableFuture<Void> addReaderGroupToScope(String scopeName, String rgName, UUID readerGroupId, 
+    public CompletableFuture<Void> addStreamTagsToIndex(String scope, String name, StreamConfiguration config, OperationContext context, Executor executor) {
+        return CompletableFuture.completedFuture(null);
```
shiveshr, line 186: unspported

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
```diff
@@ -174,20 +175,30 @@ ZKScope newScope(final String scopeName) {
...
+    }
+
+    @Override
+    public CompletableFuture<Void> removeTagsFromIndex(String scope, String name, Set<String> tagsRemoved, OperationContext context, Executor executor) {
+        return CompletableFuture.completedFuture(null);
```
shiveshr, line 191: unspported

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
```diff
@@ -760,15 +761,18 @@ private boolean isSubscriberToNonSubscriberTransition(final ReaderGroupConfigRec
...
-                                // 3. update new configuration in the store with updating flag = true
-                                // if attempt to update fails, we bail out with no harm done
-                                () -> streamMetadataStore.startUpdateConfiguration(scope, stream, newConfig,
-                                        context, executor))
-                                // 4. wait for update to complete
-                                .thenCompose(x -> eventHelper.checkDone(() -> isUpdated(scope, stream, newConfig, context))
-                                        .thenApply(y -> UpdateStreamStatus.Status.SUCCESS)));
+                        Set<String> currentTags = new HashSet<>(configProperty.getObject().getStreamConfiguration().getTags());
+                        currentTags.removeAll(newConfig.getTags());
+                        return removeTagIndexEntries(scope, stream, currentTags, context ).thenCompose(v ->
```
shiveshr, line 766: this should be done as part of the workflow so that we dont fail in between the removal and have only partial removal.

shiveshr (changes requested):
Largerly looks good but for few comments that should be addressed.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -452,6 +456,53 @@ private RetryAndThrowConditionally createRetryConfig(final ControllerImplConfig
...
+                        RPCAsyncCallback<StreamsInScopeResponse> callback = new RPCAsyncCallback<>(requestId,
+                                                                                                   LIST_STREAMS_IN_SCOPE_FOR_TAG,
+                                                                                                   scopeName);
+                        ScopeInfo scopeInfo = ScopeInfo.newBuilder().setScope(scopeName).build();
+                        StreamsInScopeWithTagRequest request = StreamsInScopeWithTagRequest.newBuilder()
+                                                                                           .setScope(scopeInfo)
+                                                                                           .setContinuationToken(token)
+                                                                                           .setTag(tag).build();
+                        new ControllerClientTagger(client, timeoutMillis).withTag(requestId,
+                                                                                  LIST_STREAMS_IN_SCOPE, scopeName)
```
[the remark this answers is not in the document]
shrids: thanks, fixed it.

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
```diff
@@ -367,16 +368,17 @@ public static final SegmentId decode(final Segment segment) {
     public static final StreamConfig decode(String scope, String streamName, final StreamConfiguration configModel) {
         Preconditions.checkNotNull(configModel, "configModel");
         final StreamConfig.Builder builder = StreamConfig.newBuilder()
-                .setStreamInfo(createStreamInfo(scope, streamName))
-                .setScalingPolicy(decode(configModel.getScalingPolicy()));
+                                                         .setStreamInfo(createStreamInfo(scope, streamName))
```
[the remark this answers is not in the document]
shrids: done.

controller/src/main/java/io/pravega/controller/server/eventProcessor/LocalController.java
```diff
@@ -142,6 +147,19 @@ public LocalController(ControllerService controller, boolean authorizationEnable
         return new ContinuationTokenAsyncIterator<>(function, "");
     }
 
+    @Override
+    public AsyncIterator<Stream> listStreamsForTag(String scopeName, String tag) {
+        final Function<String, CompletableFuture<Map.Entry<String, Collection<Stream>>>> function = token ->
+                controller.listStreams(scopeName, tag, token, requestIdGenerator.nextLong())
```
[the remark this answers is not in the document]
shrids: removed method overloading.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -1029,6 +1067,52 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request,
...
+        String scopeName = request.getScope().getScope();
+        String tag = request.getTag();
+        RequestTag requestTag = requestTracker.initializeAndTrackRequestTag(requestIdGenerator.nextLong(),
+                                                                            LIST_STREAMS_IN_SCOPE_FOR_TAG, scopeName);
+        log.info(requestTag.getRequestId(), "{} called for scope {} and tags {}", LIST_STREAMS_IN_SCOPE_FOR_TAG, scopeName, tag);
+
+        final AuthContext ctx = this.grpcAuthHelper.isAuthEnabled() ? AuthContext.current() : null;
+        Function<String, CompletableFuture<Controller.StreamsInScopeResponse>> streamsFn = delegationToken ->
+                listWithFilter(request.getContinuationToken().getToken(), pageLimit,
+                               (x, y) -> controllerService.listStreams(scopeName, tag, x, requestTag.getRequestId()),
```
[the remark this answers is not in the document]
shrids: ok, fixed it.

controller/src/main/java/io/pravega/controller/task/Stream/StreamMetadataTasks.java
shrids: fixed it.

client/src/main/java/io/pravega/client/stream/StreamConfiguration.java
```diff
@@ -16,18 +16,27 @@
...
 import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.Singular;
 
 /**
  * The configuration of a Stream.
  */
 @Data
-@Builder
+@Builder(toBuilder = true)
```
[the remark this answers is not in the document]
shrids: This enables us to convert an object to a builder. This is is helpful when we need to modify a subset of attributes while keeping the rest of the object the same.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -1029,6 +1067,52 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request,
...
+                               x -> grpcAuthHelper.isAuthorized(authorizationResource.ofStreamInScope(scopeName, x),
+                                                                AuthHandler.Permissions.READ, ctx),
+                               x -> StreamInfo.newBuilder().setScope(scopeName).setStream(x).build(), requestTag.getRequestId())
+                        .handle((response, ex) -> {
+                            if (ex != null) {
+                                if (Exceptions.unwrap(ex) instanceof StoreException.DataNotFoundException) {
```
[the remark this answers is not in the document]
shrids: yes, that is my understanding too.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shrids: improved the naming here.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: fixed it.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shrids: Ok, I have improved the code and added comments here.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: fixed it.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStreamMetadataStore.java
shrids: Good point, we are not removing the tag entry from the Table if the TagRecord has zero streams. Will improve this logic.

shrids: done.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: fixed it.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: done.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: done.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: improved the code here. Let me know if it can be improved further.

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
```diff
@@ -129,9 +129,10 @@ public static final RetentionPolicy encode(final Controller.RetentionPolicy poli
     public static final StreamConfiguration encode(final StreamConfig config) {
         Preconditions.checkNotNull(config, "config");
         return StreamConfiguration.builder()
-                .scalingPolicy(encode(config.getScalingPolicy()))
-                .retentionPolicy(encode(config.getRetentionPolicy()))
-                .build();
+                                  .scalingPolicy(encode(config.getScalingPolicy()))
+                                  .retentionPolicy(encode(config.getRetentionPolicy()))
+                                  .tags(config.getTags().getTagList())
```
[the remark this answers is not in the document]
shrids: Verified that tags are always initialized by lombok and is never null.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shrids: fixed it.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStreamMetadataStore.java
shrids: throwing unsupported causes in memory tests to fail, hence returning a completed future. Let me know if this should be modified.

controller/src/main/java/io/pravega/controller/store/stream/InMemoryStreamMetadataStore.java
shrids: throwing unsupported causes in memory tests to fail, hence returning a completed future. Let me know if this should be modified.

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
shrids: throwing unsupported causes in memory tests to fail, hence returning a completed future. Let me know if this should be modified.

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
shrids: throwing unsupported causes in memory tests to fail, hence returning a completed future. Let me know if this should be modified.

client/src/main/java/io/pravega/client/stream/StreamConfiguration.java
```diff
@@ -16,18 +16,27 @@
...
 public class StreamConfiguration implements Serializable {
-    
+
     private static final long serialVersionUID = 1L;
+    private static final int MAX_TAG_COUNT = 128;
+    private static final int MAX_TAG_LENGTH = 256;
```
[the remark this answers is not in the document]
shrids: Yes, we can increase it. As I understand the factors that we need to consider before increasing it are 
- StreamConfigurationRecord serialization size should be less than MaxValue of the TableSegment. (Max allowed is 1MB)
- Increasing the StreamConfigurationRecord affects the usage pattern of the Cache on the controller.

At the moment we do not have any requirement to store huge amount of tags or an increased size for each tags. 

Also, the reverse index of Tags to Streams is stored in a compressed format on the TableSegments. This is upperbounded by the Stream name length. So a Tag can be associated with around 137K Streams with a length of 255 characters or 1million Streams with a length of 32 characters.

controller/src/main/java/io/pravega/controller/store/InMemoryScope.java
```diff
@@ -141,6 +141,11 @@ public String getName() {
         return CompletableFuture.completedFuture(new ImmutablePair<>(result, newContinuationToken));
     }
 
+    @Override
+    public CompletableFuture<Pair<List<String>, String>> listStreamsForTag(String tag, String continuationToken, Executor executor, OperationContext context) {
+        return null;
```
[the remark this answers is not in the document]
shrids: fixed it.

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
shiveshr: shouldnt this be called only when configuration includes tags? so the existing tests should be fine, right?

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -180,37 +217,77 @@ public String getName() {
...
+     */
+    CompletableFuture<Pair<List<String>, String>> getStreamsFromNextTagChunk(String tag, String token, OperationContext context) {
+        return getAllStreamTagsInScopeTableNames(context).thenApply(
+                strings -> {
+                    if (token.isEmpty()) {
+                        // token is empty, try reading from the first tag table.
+                        return strings.get(0);
+                    } else {
+                        // return next index
+                        return strings.get(strings.indexOf(token) + 1);
```
shiveshr, line 274: how are we using the continuation token here?
strings refers to all the table chunk names, right?
token is a particular chunk number, right?
why do we need to do strings.indexOf(token)?

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -180,37 +217,77 @@ public String getName() {
...
+            if (pair.getLeft().isEmpty() && !pair.getRight().contains(LAST_TAG_CHUNK)) {
+                return listStreamsForTag(tag, pair.getRight(), executor, context);
+            } else {
+                return CompletableFuture.completedFuture(pair);
+            }
+        });
+    }
+
+    /*
+        Fetch the Streams from the next available Tag Chunk table.
```
shiveshr, line 264: i thought all chunks of entry for the same tag are in the same table segment.. 
remind me of the model and nomenclature again please:

1. streams corresponding to a tag are all written in one table segment based on the hash of the tag, right?
2. streams for a tag are chunked across multiple rows, right?

so if this method takes a "tag" as param, i would presume its looking for all streams corresponding to that tag. 
and that all can be found in a single chunk.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +248,101 @@ public void invalidateCache(String table, String key) {
...
+                                                        // Delete TagRecords with empty streams.
+                                                        .thenCompose(keyVersions -> deleteEmptyTagRecords(tableName, requestId, result.getRight(), result.getLeft(), keyVersions))
+                                                        .whenComplete((v, ex) -> {
+                                                            releaseEntries(result.getLeft());
+                                                        });
+                                });
+        }, errorMessage, true, requestId);
+    }
+
+    private CompletableFuture<Void> deleteEmptyTagRecords(String tableName, long requestId, List<Integer> deleteIndex,
```
shiveshr, line 290: i thought we decided to remove this from here.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +248,101 @@ public void invalidateCache(String table, String key) {
                 .whenComplete((r, ex) -> releaseEntries(entries));
     }
 
+    /**
+     * Remove value of every entry for the provided keys.
```
shiveshr, line 252: this logic should be moved out.
the caller should use an updateValues method which you can add here. 
where they get and update all the lists and issue a conditional update on the helper. 

alternatively, you can pass an update lambda.

getAndUpdateValues(final String tableName, Set<String> tableKeys, 
                                                BiFunction<String, V, V> updateFunction,
                                                Function<byte[], V> fromBytes,
                                                long requestId)
logically following can be done:
1. Map<K+keyVersion, V> vbytes = getValues(tableKeys);
2. updates = foreach vbyte: vBytes -> update = updateFunction(fromBytes(vByte))
3. updateValues(k+keyVersion, updates);

shiveshr: presently here the helper is assuming that the values are a list of string and one entry from the list has to be removed.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +248,101 @@ public void invalidateCache(String table, String key) {
...
+     * Append value to every entry corresponding to the provided keys.
+     *
+     * @param tableName      Table name.
+     * @param tableKeys      List of keys whose value needs to be updated.
+     * @param appendValue    Value to be appended to all the keys.
+     * @param updateFunction Update function which is used to compute the updated entries.
+     * @param requestId      Request id.
+     * @return Future which completes when the operation is complete.
+     */
+    public CompletableFuture<List<Version>> appendValues(final String tableName, Set<String> tableKeys, final String appendValue,
```
shiveshr, line 322: same here. 
add an updateValues method in helper. which your code can then make use of.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: Thanks, I have improved the abstraction now. Let me know.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: Done, Improved the abstraction.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: Fixed it.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shrids: I have improved the java docs. Could you check it now ?

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shrids: updated the java docs.

controller/src/main/java/io/pravega/controller/store/stream/ZKStreamMetadataStore.java
shrids: This function is invoked in `UpdateStreamTask` workflow. The check if tags are present is performed in the MetadataStore. I did not want to expose this logic in the UpdateStreamTask. Let me know if this can be improved further.

shrids:
System test run is successful with this changeset.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateStreamTask.java
```diff
@@ -99,25 +99,27 @@ public UpdateStreamTask(final StreamMetadataTasks streamMetadataTasks,
...
                 .thenCompose(updated -> updateStreamForAutoStreamCut(scope, stream, configProperty, updated)
-                        .thenCompose(x -> notifyPolicyUpdate(context, scope, stream, configProperty.getStreamConfiguration(), 
+                        .thenCompose(x -> notifyPolicyUpdate(context, scope, stream, configProperty.getStreamConfiguration(),
                                 requestId))
-                        .thenCompose(x -> handleSegmentCounUpdates(scope, stream, configProperty, etr, context, executor, 
-                                requestId))
-                        .thenCompose(x -> streamMetadataStore.completeUpdateConfiguration(scope, stream, record, context, 
+                        .thenCompose(x -> handleSegmentCountUpdates(scope, stream, configProperty, etr, context, executor,
+                                                                    requestId))
+                        .thenCompose(x -> streamMetadataStore.removeTagsFromIndex(scope, stream, configProperty.getRemoveTags(), context, executor))
```
shiveshr, line 109: remove tags from index and update tags to index can get some concurrency conflicts because multiple streams could be concurrently adding or removing tags. 
should we open an issue to think deeper about how we would want to solve it?

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
```diff
@@ -996,7 +1034,7 @@ public void listStreamsInScope(Controller.StreamsInScopeRequest request,
         final AuthContext ctx = this.grpcAuthHelper.isAuthEnabled() ? AuthContext.current() : null;
         Function<String, CompletableFuture<Controller.StreamsInScopeResponse>> streamsFn = delegationToken ->
                 listWithFilter(request.getContinuationToken().getToken(), pageLimit,
-                        (x, y) -> controllerService.listStreams(scopeName, x, y, requestTag.getRequestId()),
+                        (x, y) -> controllerService.listStreamsForTag(scopeName, x, y, requestTag.getRequestId()),
```
shiveshr, line 1037: do we intend to change it here?

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -121,42 +145,55 @@ public String getName() {
...
         return getId(context).thenApply(id ->
-                getQualifiedTableName(INTERNAL_SCOPE_NAME, scopeName, String.format(KVTABLES_IN_SCOPE_TABLE_FORMAT, id.toString())));
+                                                getQualifiedTableName(INTERNAL_SCOPE_NAME, scopeName, String.format(KVTABLES_IN_SCOPE_TABLE_FORMAT, id.toString())));
+    }
+
+    public CompletableFuture<String> getAllStreamTagsInScopeTableNames(String stream, OperationContext context) {
+        return getId(context)
+                .thenApply(id -> getQualifiedTableName(INTERNAL_SCOPE_NAME, scopeName,
+                                                       String.format(STREAM_TAGS_IN_SCOPE, id.toString(),
+                                                                     HASH.hashToBucket(stream, 25))));
```
shiveshr, line 176: whats "25" here? should it be a constant?

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
```diff
@@ -223,8 +305,80 @@ public void refresh() {
...
+                                                                                                   context.getRequestId()))
+                                                        .collect(Collectors.toList())));
+    }
+
+    private TableSegmentEntry removeStreamFromEntry(String removeValue, TableSegmentEntry currentEntry) {
+        String k = currentEntry.getKey().getKey().toString(StandardCharsets.UTF_8);
+        byte[] array = storeHelper.getArray(currentEntry.getValue());
+        byte[] updatedBytes;
+        if (array.length == 0) {
+            updatedBytes = TagRecord.builder().tagName(k).build().toBytes();
```
shiveshr, line 367: nit: we can simply return current entry from here without having to create a new object.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +247,62 @@ public void invalidateCache(String table, String key) {
                 .whenComplete((r, ex) -> releaseEntries(entries));
     }
 
+    /*
+        Conditional delete of the specified key with the specified version.
+        If the conditional delete fails it is ignored.
+     */
+    private CompletableFuture<Void> conditionalDeleteOfKey(String tableName, long requestId, TableSegmentKey key,
```
shiveshr, line 254: we have a method for condition delete of key called - removeEntry where if you provide a version then it is used to delete the key.
is there a need for this new method?

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
```diff
@@ -244,6 +247,62 @@ public void invalidateCache(String table, String key) {
...
+     *
+     * @param tableName Table Name.
+     * @param tableKey Key to update.
+     * @param updateFunction A function which updates the values.
+     * @param attemptCleanup A Predicate which decides if a cleanup should be invoked.
+     * @param requestId Request id.
+     * @return A future which completes when the update operation has completed. A conditional update failure will cause
+     * the future to complete exceptionally.
+     */
+    public CompletableFuture<Void> getAndUpdateValues(final String tableName, final String tableKey, final Function<TableSegmentEntry, TableSegmentEntry> updateFunction,
```
shiveshr, line 281: nit: its updating a single entry - should we rename this method to getAndUpdateEntry instead?

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
```diff
@@ -131,20 +131,20 @@
     private final ScheduledExecutorService executor;
 
     @VisibleForTesting
-    PravegaTablesStream(final String scopeName, final String streamName, PravegaTablesStoreHelper storeHelper, 
+    PravegaTablesStream(final String scopeName, final String streamName, PravegaTablesStoreHelper storeHelper,
```
shiveshr, line 134: there doesnt seem to be any change in this class. please revert it.
also if possible, please revert the formatting changes in other classes, particularly PersistentStreamBase and PravegaTablesScope.

shiveshr (approved):
LGTM except for few minor comments.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/v1/ControllerServiceImpl.java
shrids: fixed it.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: Yes, The main differences with removeEntry method are 
- if the conditional delete fails due to a concurrent update this error is ignored. 
- cache is not used for the tag values while the removeEntry method uses that.

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: Yes, corrected it.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shrids: yes, thanks for pointing it. Fixed it.

controller/src/main/java/io/pravega/controller/store/stream/PravegaTablesStream.java
shrids: done.

controller/src/main/java/io/pravega/controller/store/PravegaTablesScope.java
shrids: yes, improved the code.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateStreamTask.java
shiveshr: Could you please create a github issue and put the issue id here. thanks

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shiveshr: remove entry simply removes from the cache as well if it exists. so its safe.

remove entry will throw the concurrency update error. if supplied key version is null, only in that case it does unconditional delete. otherwise it does conditional delete. 

https://github.com/pravega/pravega/blob/9b4e6156390d2e2c9e9897efd4687773e9c9ed6e/controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java#L562

it does handle the idempotent delete when it gets dataNotFound exception and converts it to a happy case.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/UpdateStreamTask.java
shrids: https://github.com/pravega/pravega/issues/6096

controller/src/main/java/io/pravega/controller/store/PravegaTablesStoreHelper.java
shrids: Thanks, I have improved the code to remove this code duplication.

PR #6064, opened 2021-05-27
Issue 6004: Change Retry predicate for Controller Events so they are always retried unless on specific exceptions

### LINKED ISSUE #6004: Controller does not retry API operations failing due to OutofMemory Errors

pbelgundi:

**Change log description**  
The Event Processor framework on Controller processes Controller Events. If any exception is thrown during event processing, the Event is retried, unless the Exception is _StoreException.OperationNotAllowedException_, in which case the event processing is not retried. 
This PR changes the Event Retry logic from _inclusive_ to _exclusive_, saying do not retry on **these** set of exceptions and retry on everything else.

**Purpose of the change**  
Fixes #6004 

**What the code does**  
Changed Retry predicate for Controller Event Processing to include exceptions on which we do not want the event to be Retried. On any other Exception or Error, the event will be retried.

**How to verify it**  
All Unit, Intergation and System Tests should pass.

### DIFF
--- a/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
+++ b/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
@@ -21,6 +21,7 @@ import io.pravega.common.concurrent.Futures;
 import io.pravega.common.tracing.RequestTag;
 import io.pravega.common.util.Retry;
 import io.pravega.controller.eventProcessor.impl.SerializedRequestHandler;
+import io.pravega.controller.store.stream.EpochTransitionOperationExceptions;
 import io.pravega.controller.store.stream.OperationContext;
 import io.pravega.controller.store.stream.StoreException;
 import io.pravega.controller.store.stream.StreamMetadataStore;
@@ -66,8 +67,15 @@ import static io.pravega.controller.eventProcessor.impl.EventProcessorHelper.wit
 @Slf4j
 public abstract class AbstractRequestProcessor<T extends ControllerEvent> extends SerializedRequestHandler<T> 
         implements StreamRequestProcessor {
-    protected static final Predicate<Throwable> OPERATION_NOT_ALLOWED_PREDICATE = e -> Exceptions.unwrap(e) 
-            instanceof StoreException.OperationNotAllowedException;
+    protected static final Predicate<Throwable> OPERATION_NOT_ALLOWED_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.OperationNotAllowedException;
+    protected static final Predicate<Throwable> ILLEGAL_STATE_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.IllegalStateException;
+    protected static final Predicate<Throwable> DATA_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException;
+    protected static final Predicate<Throwable> VM_ERROR_PREDICATE = e -> Exceptions.unwrap(e) instanceof java.lang.VirtualMachineError;
+    protected static final Predicate<Throwable> SEGMENT_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataContainerNotFoundException;
+    protected static final Predicate<Throwable> EVENT_RETRY_PREDICATE = OPERATION_NOT_ALLOWED_PREDICATE.or(ILLEGAL_STATE_PREDICATE).or(DATA_NOT_FOUND_PREDICATE)
+                                                                            .or(SEGMENT_NOT_FOUND_PREDICATE).or(VM_ERROR_PREDICATE).negate();
+    protected static final Predicate<Throwable> SCALE_EVENT_RETRY_PREDICATE = EVENT_RETRY_PREDICATE
+                                                                                .or(e -> e instanceof EpochTransitionOperationExceptions.ConflictException).negate();
 
     protected final StreamMetadataStore streamMetadataStore;
 
--- a/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
+++ b/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
@@ -87,7 +87,7 @@ public class CommitRequestHandler extends AbstractRequestProcessor<CommitEvent> 
 
     @Override
     public CompletableFuture<Void> processCommitTxnRequest(CommitEvent event) {
-        return withCompletion(this, event, event.getScope(), event.getStream(), OPERATION_NOT_ALLOWED_PREDICATE);
+        return withCompletion(this, event, event.getScope(), event.getStream(), EVENT_RETRY_PREDICATE);
     }
 
     /**
--- a/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/StreamRequestHandler.java
+++ b/controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/StreamRequestHandler.java
@@ -16,7 +16,6 @@
 package io.pravega.controller.server.eventProcessor.requesthandlers;
 
 import io.pravega.common.tracing.TagLogger;
-import io.pravega.controller.store.stream.EpochTransitionOperationExceptions;
 import io.pravega.controller.store.stream.StreamMetadataStore;
 import io.pravega.shared.controller.event.AutoScaleEvent;
 import io.pravega.shared.controller.event.ControllerEvent;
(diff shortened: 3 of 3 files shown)


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -66,8 +67,15 @@
...
         implements StreamRequestProcessor {
-    protected static final Predicate<Throwable> OPERATION_NOT_ALLOWED_PREDICATE = e -> Exceptions.unwrap(e) 
-            instanceof StoreException.OperationNotAllowedException;
+    protected static final Predicate<Throwable> OPERATION_NOT_ALLOWED_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.OperationNotAllowedException;
+    protected static final Predicate<Throwable> ILLEGAL_STATE_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.IllegalStateException;
+    protected static final Predicate<Throwable> DATA_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException;
+    protected static final Predicate<Throwable> VM_ERROR_PREDICATE = e -> Exceptions.unwrap(e) instanceof java.lang.VirtualMachineError;
+    protected static final Predicate<Throwable> SEGMENT_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataContainerNotFoundException;
+    protected static final Predicate<Throwable> EVENT_RETRY_PREDICATE = OPERATION_NOT_ALLOWED_PREDICATE.or(ILLEGAL_STATE_PREDICATE).or(DATA_NOT_FOUND_PREDICATE)
```
shiveshr, line 75: retry predicate should be:
OPERATION_NOT_ALLOWED_PREDICATE || not(ILLEGAL_STATE_PREDICATE) || not(DATA_NOT_FOUND_PREDICATE)
we dont want to retry on data_not_found predicate or illegal state predicate for instance.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -66,8 +67,15 @@
...
+                                                                            .or(SEGMENT_NOT_FOUND_PREDICATE).or(VM_ERROR_PREDICATE).negate();
+    protected static final Predicate<Throwable> SCALE_EVENT_RETRY_PREDICATE = EVENT_RETRY_PREDICATE
```
shiveshr, line 77: there are few failing exceptions for scale which should not be retried upon.. like EpochTransitionException.InputInvalidException or PreConditionFailureException or ConditionInvalidException

i believe the idea is to include all the known exceptions on which we definitely want to fail. So we only fail on failing exceptions. For everything else we retry. Which includes any other exception.

shiveshr (changes requested):
The idea is to invert the retryable predicate here to retry on anything other than known failing exceptions. 
There would be some common known failing exceptions like DataNotFound and some workflow specific which needs to be described in the workflow specific predicates

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
pbelgundi: Perhaps you missed noticing that I have added a `negate()` at the end to indicate this predicate returns "false" if any of these conditions match.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
pbelgundi: Yes the idea is just that. I have added the other exceptions you mentioned.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
shiveshr: if its negate, then operationnotallowed should be removed from here.. we want to retry on operationnotallowed as that is thrown if the state transition is not allowed (like if its in scaling and a workflow is trying to set it to committing_txn).

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
shiveshr: then please check for all other workflows to see which all exceptions can be thrown that are definitely failing exceptions.

pbelgundi: added more exceptions, pls check

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
pbelgundi: Ok. Removed OPERATION_NOT_FOUND from Predicate.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -66,8 +67,20 @@
 @Slf4j
 public abstract class AbstractRequestProcessor<T extends ControllerEvent> extends SerializedRequestHandler<T> 
         implements StreamRequestProcessor {
-    protected static final Predicate<Throwable> OPERATION_NOT_ALLOWED_PREDICATE = e -> Exceptions.unwrap(e) 
-            instanceof StoreException.OperationNotAllowedException;
+    protected static final Predicate<Throwable> ILLEGAL_STATE_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.IllegalStateException;
+    protected static final Predicate<Throwable> DATA_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataNotFoundException;
+    protected static final Predicate<Throwable> SEGMENT_NOT_FOUND_PREDICATE = e -> Exceptions.unwrap(e) instanceof StoreException.DataContainerNotFoundException;
+    protected static final Predicate<Throwable> EVENT_RETRY_PREDICATE = ILLEGAL_STATE_PREDICATE.or(DATA_NOT_FOUND_PREDICATE)
```
shiveshr, line 73: nit: lets rename it event_writeback_predicate.

shiveshr, line 73: nit: lets rename it event_writeback_predicate.
also, if we write back the event, then presently we still fail the future with the cause for failure which goes to concurrenteventprocessor which also has a writeback logic on retryable exceptions. 
we should actually complete the future in `withCompletionMethod` (line 182)

pbelgundi: fixed.

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
```diff
@@ -66,8 +67,20 @@
...
+                                                                            .or(SEGMENT_NOT_FOUND_PREDICATE)
+                                                                            .or(e -> Exceptions.unwrap(e) instanceof IllegalArgumentException)
+                                                                            .or(e -> Exceptions.unwrap(e) instanceof NullPointerException)
+                                                                            .negate();
+    protected static final Predicate<Throwable> SCALE_EVENT_RETRY_PREDICATE = EVENT_RETRY_PREDICATE
+                                                                                .or(e -> e instanceof EpochTransitionOperationExceptions.ConflictException)
+                                                                                .or(e -> e instanceof EpochTransitionOperationExceptions.ConditionInvalidException)
```
shiveshr, line 80: this seems incorrect.. we want to retry on conflict exception. 

if you look at existing condition it is using conflict as a retry predicate condition. 
so we should remove it from here because now we are negating.

shiveshr: also does the negate at the end also negate the event retry predicate? that would be incorrect too

shiveshr (changes requested):

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/AbstractRequestProcessor.java
pbelgundi: Fixed this.

pbelgundi:
System tests completed successfully on the PR branch.

shiveshr (approved):

PR #6066, opened 2021-05-27
Issue 6063: Ensure getTxn API starts the Transaction Pinger.

### LINKED ISSUE #6063: Improvement of Transaction lifetime control
Sometimes it is hard to promise a transaction will always be available to be committed in extreme cases, due to the fact that server will abort the transaction after 90 seconds without client heartbeat that comes from the same writer who create the transaction. 
If we experiement the following test:
1. start a transaction in process A, this will return transaction Id 00000000-0000-0000-0000-000000000001:
   TransactionalEventStreamWriter::beginTxn()
2. Exit process A
3. Wait 80 seconds and start another process B
4. In process B, try to gain existing transaction created in 1. :
   TransactionalEventStreamWriter::getTxn(00000000-0000-0000-0000-000000000001) 
5. did nothing (no write/flush/commit) and leave the transaction for another 10 sec and then get the status:
   Transaction::checkStatus()
6. the transaction status returned is ABORTED instead of OPEN

The second writer will not try to keep the transaction alive, nor would it extend the lease for another 90 seconds.

1. The second writer will try to heartbeat the Transaction implictly just like what the first writer did.
2. As soon as the second writer calls TransactionalEventStreamWriter::getTxn(), the server would extend the lease of the transaction for another 90 seconds.

N/A

**Additional context**
This issue is found during the testi of Exactly-Once semantics of PSearch. 
For exactly once processing PSearch now uses two phase commit leveraging checkpoint from reader side and transactions from writer side. When a checkpoint is triggered, the two phase commit went through these steps:
1. Coordinator triggered manual checkpoint in Pravega
2. Readers reads the checkpoint event and calls Txn::flush()
3. If the flush succeeded, the reader will read the next event; if not, the corresponding worker will retry until retry exhausted, which then cause the worker to explictly abort the Txn and the checkpoint will be rollback.
4. The coordinator will see the checkpoint completed, then it will mark and persist the checkpoint state as completed.

All these transactions will be persisted with TxnID in case of failure, and be committed asynchronously. If one of the workers crashed, I would like one of the other workers to simply pick the Txn up and try commit it. However, such transition might take unknown period, especially in some extreme cases where we don't know if such a transition would succeed within 90 seconds, because a new worker always need some time to prepare all the stuffs like reset all those transactions. In worst case there could be an avalanche effect. This is why I would like to have a lightweight approach to simply switch the owner of these transactions when I know a worker died, and try keeping these transactions alive before they are ready to be committed.

shrids:
**Change log description**  
Ensure getTxn API starts the Transaction Pinger.

**Purpose of the change**  
Fixes #6063 

**What the code does**  
Ensure getTxn API starts the Transaction Pinger.

**How to verify it**  
All the existing and modified tests should pass.

### DIFF
--- a/client/src/test/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterTest.java
+++ b/client/src/test/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterTest.java
@@ -47,7 +47,9 @@ import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
 
 public class TransactionalEventStreamWriterTest extends ThreadPooledTestSuite {
     @Rule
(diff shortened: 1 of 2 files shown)


client/src/test/java/io/pravega/client/stream/impl/TransactionalEventStreamWriterTest.java
```diff
@@ -125,19 +127,26 @@ public void testGetTxn() throws TxnFailedException {
...
         TransactionalEventStreamWriter<String> writer = new TransactionalEventStreamWriterImpl<>(stream, "id", controller, streamFactory, serializer,
                 config, executorService());
         writer.beginTxn();
+        writer.close();
 
-        Transaction<String> txn = writer.getTxn(txnId);
+        //Create a new transactional eventWriter
+        @Cleanup
+        TransactionalEventStreamWriter<String> writer1 = new TransactionalEventStreamWriterImpl<>(stream, "id", controller, streamFactory, serializer,
+                                                                                                  EventWriterConfig.builder().transactionTimeoutTime(10000).build(),
```
shrids, line 138: The assumption here is that the `transactionTimeout ` passed here is the same value that was passed during the transaction creation.

tkaitchuck (approved):

PR #6077, opened 2021-06-09
Changes for backpressure on high transaction commit backlog

pbelgundi:
**Change log description**  
This PR has changes for implementing backpressure in Controller when committing transactions count is beyond a certain threshold for each Stream. The threshold can be set using value of Controller property "transaction.open.limit" at startup.

**Purpose of the change**  
Implement throttle  that applies backpressure when committing transactions count is beyond a certain threshold for each Stream.

**How to verify it**  


controller/src/main/java/io/pravega/controller/store/stream/records/CommittingTxnsCountRecord.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+
+/**
+ * A serializable class that encapsulates a Stream's State.
+ */
+@Data
+@Builder
+@Slf4j
+@AllArgsConstructor
+public class CommittingTxnsCountRecord {
```
shiveshr, line 38: this could just be an integer. we do not need a wrapping serializable class here.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -2213,11 +2215,11 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
...
-                            // if entry matches record's position then include it
-                            transactionsMap.add(txnRecord);
-                            if (transactionsMap.size() >= limit) {
-                                break;
-                            }
+                             if (transactionsMap.size() < limit) {
+                                 // if entry matches record's position then include it
+                                 transactionsMap.add(txnRecord);
+                             }
+                             committingTxnCount.addAndGet(1);
```
shiveshr, line 2222: this is counting txns only in the lowest epoch. 
so either we fetch all txns which are found in orderer and check which of those are already marked for committing, which can be optimized by doing these fetches in parallel (present logic is trying to fetch in batches of upto `limit` as it is trying to create a batch of `limit` size only. but you are doing a count of all such txns so doing gets in parallel may be faster.

shiveshr: Although i would suggest we dont do the exact count anyway (which can change while we are counting).
instead we can go with the approximate value which is simply zkorderer.children_count

in happy case if a txn is in orderer, then will also be marked as committing. so we dont need the accurate count but an approximate count that can tell whether there is a growing backlog or not.

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
```diff
@@ -369,44 +375,58 @@ public void initializeStreamWriters(final EventStreamWriter<CommitEvent> commitW
...
-                        return new ImmutablePair<>(txnFuture.join(), segments);
-                    }, executor);
-                }), e -> {
-            Throwable unwrap = Exceptions.unwrap(e);
-            return unwrap instanceof StoreException.WriteConflictException || unwrap instanceof StoreException.DataNotFoundException;
-        }, 5, executor));
+        return validate.thenCompose(validated -> checkTxnRateLimit(scope, stream, ctx))
+                .thenCompose(limitCheckPassed -> {
+                if (!limitCheckPassed) {
+                    throw new RuntimeException("Create transaction failed because Open Transactions limit exceeded. Try again later.");
```
shiveshr, line 381: should we add delays rather than throw this back to the caller?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -2180,8 +2180,8 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
                       });
     }
 
-    protected CompletableFuture<Map<Long, UUID>> getAllOrderedCommittingTxnsHelper(ZkOrderedStore txnCommitOrderer,
-                                                                                   OperationContext context) {
+    protected CompletableFuture<Map<Long, UUID>> getAllOrderedTxnsHelper(ZkOrderedStore txnCommitOrderer,
```
shiveshr, line 2183: nit: no need to change the name

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -2656,9 +2656,6 @@ private StreamSegmentRecord newSegmentRecord(int epoch, int segmentNumber, long
      */
     abstract CompletableFuture<List<VersionedTransactionData>> getOrderedCommittingTxnInLowestEpoch(
             int limit, OperationContext context);
-    
-    @VisibleForTesting
```
shiveshr, line 2660: why is this method removed?

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -976,9 +977,13 @@ public OperationContext createStreamContext(String scopeName, String streamName,
         Stream streamObj = getStream(scope, stream, context);
         return Futures.completeOn(streamObj.completeCommittingTransactions(record, context, writerTimes, writerIdToTxnOffsets), executor)
                 .thenAcceptAsync(result -> {
-                    streamObj.getNumberOfOngoingTransactions(context).thenAccept(count ->
-                            TransactionMetrics.reportOpenTransactions(scope, stream, count));
-                    TransactionMetrics.getInstance().commitTransactionComplete(timer.getElapsed());
+                    streamObj.getAllOrderedTxns(context)
```
shiveshr, line 980: instead of doing it again here, whereby adding to the commit-txn latency, we could do it when this was called first -- trystartcommittingTxn and update the count at that point.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: I had to make this method public to call it from CommitRequestHandler hence remove from here and add to Stream class, but I can reverse all these changes now.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: done

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -2213,11 +2215,12 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
                 switch (txnRecord.getStatus()) {
                     case COMMITTING:
                         if (txnRecord.getCommitOrder() == order) {
-                            // if entry matches record's position then include it
-                            transactionsMap.add(txnRecord);
-                            if (transactionsMap.size() >= limit) {
-                                break;
-                            }
+                             if (transactionsMap.size() < limit) {
```
shiveshr: werent we going to revert this?

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -2172,16 +2172,18 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
                               TransactionMetrics.getInstance().reportTransactionBatchCreate(timer2.getElapsed());
                               Timer timer3 = new Timer();
                               return txnCommitOrderer.removeEntities(getScope(), getName(), toPurge)
-                                      .thenAccept(k -> TransactionMetrics.getInstance().reportTransactionZkOrdererPurgeStale(timer3.getElapsed()));
-                          })
-                                        .thenApply(v -> transactionsMap.stream().sorted(
+                                      .thenAccept(k -> {
```
shiveshr, line 2175: this should be thenCompose otherwise that future is dangling

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -2172,16 +2172,18 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
...
+                                          TransactionMetrics.getInstance().reportTransactionZkOrdererPurgeStale(timer3.getElapsed());
+                                          updateCommittingTxnsCount(allTxns.size(), context);
```
shiveshr, line 2177: this is the dangling future

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/task/Stream/StreamTransactionMetadataTasks.java
pbelgundi: For a POC I think this should be good

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
pbelgundi: fixed

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

PR #6080, opened 2021-06-10
Issue 5939: (KVT Beta 2) Update KVT Config in Controller and add API to retrieve it

### LINKED ISSUE #5939: KVT Beta 2: Update KVT Config in Controller and add API to retrieve it

anirudhkovuru:
**Change log description**  
Updated the KVT config on the controller and implemented an API in the client to retrieve the config.

**Purpose of the change**  
Fixes #5939 

**What the code does**  
For persisting KVT config on the controller:
- `Controller.proto` is updated to accommodate the new fields in the `KeyValueTableConfiguration` to generate the data structure needed to send it over the wire. The functions `encode` and `decode` in the `ModelHelper` also account for this change.
- The `CreateTableEvent` also passes this new information ahead.
- The `CreateTableTask` obtains this data and persists it in the controller in the form of a `KVTConfigurationRecord` which has also been updated to contain the new data.

Using the key lengths:
- The key lengths are added into `keyLength` in the `CreateTableTask` and passed onto `TableMetadataTasks.createNewSegments` where they can be used for the segment-store call, `segmentHelper.createTableSegment`. 

Added a new API on the client to retrieve this new config:
- A new API `getKeyValueTableConfiguration` has been added to `Controller.java` interface on the client.
- The controller GRPC server enables this API through changes to the `ControllerServiceImpl` and `ControllerService` classes.
- The GRPC server returns a status, if successful we get a `SUCCESS` status and the required config else a `FAILURE` status which is thrown back to the client as a `ControllerFailureException`. If the key-value table in question doesn't exist a `TABLE_NOT_FOUND` status is sent which is thrown to the client as an `IllegalArgumentException`. 

Resolved all TODOs referencing the issue.

**How to verify it**  
All tests should pass and build should succeed.


anirudhkovuru:
Need to add integration tests for these changes.

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
```diff
@@ -1580,6 +1581,30 @@ private void closeChannel() {
...
+
+        final CompletableFuture<KeyValueTableConfig> result = this.retryConfig.runAsync(() -> {
+            RPCAsyncCallback<KeyValueTableConfig> callback = new RPCAsyncCallback<>(requestId,
+                    GET_KEY_VALUE_TABLE_CONFIGURATION, scope, kvtName);
+            new ControllerClientTagger(client, timeoutMillis).withTag(requestId, GET_KEY_VALUE_TABLE_CONFIGURATION,
+                    scope, kvtName)
+                    .getKeyValueTableConfiguration(ModelHelper.createKeyValueTableInfo(scope, kvtName), callback);
+            return callback.getFuture();
+        }, this.executor);
+        return result.thenApplyAsync(ModelHelper::encode, this.executor)
```
andreipaduroiu, line 1599: Can we confirm what the behavior will be if the KVT does not exist? Will it:
1. Throw an exception (i.e., return a failed future with an exception)? If so, which one?
2. Return null?

Let's update this method's contract in `Controller.java` to clearly define the behavior in this case.

client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
```diff
@@ -2301,6 +2319,22 @@ public void testKVTablesInScope() {
                 e -> Exceptions.unwrap(e) instanceof RuntimeException);
     }
 
+    @Test
+    public void testGetKeyValueTableConfiguration() {
+        CompletableFuture<KeyValueTableConfiguration> kvtConfig = controllerClient.getKeyValueTableConfiguration("scope1", "kvtable1");
+        assertEquals(Integer.BYTES, kvtConfig.join().getPartitionCount());
```
andreipaduroiu, line 2325: May I suggest setting different values for these fields? That way you can verify you aren't cross-assigning them by accident.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
```diff
@@ -127,6 +127,8 @@
         Preconditions.checkNotNull(kvtConfig, "kvTableConfig");
         Preconditions.checkArgument(createTimestamp >= 0);
         Preconditions.checkArgument(kvtConfig.getPartitionCount() > 0);
+        Preconditions.checkArgument(kvtConfig.getPrimaryKeyLength() > 0);
+        Preconditions.checkArgument(kvtConfig.getSecondaryKeyLength() > 0);
```
andreipaduroiu, line 131: Secondary Key length may be 0. That's a valid value (see `KeyValueTableConfiguration.java` in the client. Primary Key must be greater than 0 (so you have it correct).

controller/src/main/java/io/pravega/controller/store/kvtable/records/KVTConfigurationRecord.java
```diff
@@ -81,8 +81,8 @@ private void read00(RevisionDataInput revisionDataInput,
                                       .kvtName(revisionDataInput.readUTF());
             KeyValueTableConfiguration config = KeyValueTableConfiguration.builder()
                     .partitionCount(revisionDataInput.readInt())
-                    .primaryKeyLength(4) // TODO fix this with https://github.com/pravega/pravega/issues/5939
-                    .secondaryKeyLength(4) // TODO fix this with https://github.com/pravega/pravega/issues/5939
+                    .primaryKeyLength(revisionDataInput.readInt())
```
andreipaduroiu, line 84: Just FYI. This change is NOT backward compatible. In this particular case, it is OK, as the whole feature is not backward compatible so we do not support "upgrades" from the previous one.

But in another case (or in the future, if we decide to extend this), don't forget to declare a new revision (in `declareRevisions`) and add your data in there.

shared/controller-api/src/main/java/io/pravega/shared/controller/event/kvtable/CreateTableEvent.java
```diff
@@ -68,7 +70,8 @@ protected byte getWriteVersion() {
 
         @Override
         protected void declareVersions() {
-            version(0).revision(0, this::write00, this::read00);
+            version(0).revision(0, this::write00, this::read00)
```
andreipaduroiu, line 73: Do we need to do this? Is `CreateTableEvent` used for anything other than KVTs? If not, then we may not need to worry about backwards compatibility (in this particular scenario). Please double check before taking my suggestion for granted :)

controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
```diff
@@ -264,15 +265,15 @@ public String retrieveDelegationToken() {
     }
 
     public CompletableFuture<Void> createNewSegments(String scope, String kvt,
-                                                      List<Long> segmentIds, long requestId) {
+                                                      List<Long> segmentIds, int keyLength, long requestId) {
```
pbelgundi, line 268: It may be good to create another overloaded `createNewSegments()` method that creates all segments with keyLength=0 (_does not take `keylength` as a parameter and invokes this method_), so existing non-KVT createNewSegments() calls don't need to change.

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
```diff
@@ -264,15 +265,15 @@ public String retrieveDelegationToken() {
...
         return Futures.toVoid(Futures.allOfWithResults(segmentIds
                 .stream()
                 .parallel()
-                .map(segment -> createNewSegment(scope, kvt, segment, retrieveDelegationToken(), requestId))
+                .map(segment -> createNewSegment(scope, kvt, segment, keyLength, retrieveDelegationToken(), requestId))
                 .collect(Collectors.toList())));
     }
 
-    private CompletableFuture<Void> createNewSegment(String scope, String kvt, long segmentId, String controllerToken,
+    private CompletableFuture<Void> createNewSegment(String scope, String kvt, long segmentId, int keyLength, String controllerToken,
```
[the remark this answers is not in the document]
andreipaduroiu: +1. We should pass it down all the way to `SegmentHelper` and write a `TODO` in there to wire it up to the wire command. The other PR that implements the wire command is not merged into the branch yet. Whichever of these 2 PRs gets merged second will have to connect the two there. But we do have to pass down that value all the way to `SegmentHelper`.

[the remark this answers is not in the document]
andreipaduroiu: UPDATE: The other PR has been merged into the feature branch. Let's merge the feature branch into this one and let's plumb it all the way through (have SegmentHelper pass the value down to the wire).

client/src/main/java/io/pravega/client/control/impl/ControllerImpl.java
anirudhkovuru: I've added status to the GRPC server interaction. The server now returns a status indicating if the getting the config for the KVT was successful or not. Including a status for in case the KVT in question is not found.

client/src/test/java/io/pravega/client/control/impl/ControllerImplTest.java
anirudhkovuru: Set different values as suggested.

controller/src/main/java/io/pravega/controller/server/ControllerService.java
anirudhkovuru: Updated this check to reflect the same.

shared/controller-api/src/main/java/io/pravega/shared/controller/event/kvtable/CreateTableEvent.java
anirudhkovuru: As with the previous case, since this change is not backwards compatible anyways I think there is little reason to have an new revision. The new serialization is being done in the original revision.

controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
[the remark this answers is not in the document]
anirudhkovuru: The SegmentHelper now uses the new keyLength value.

controller/src/main/java/io/pravega/controller/task/KeyValueTable/TableMetadataTasks.java
pbelgundi: Oh yes, didn't realize this is a method in TableMetadataTasks.

pbelgundi (approved):
LGTM

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
```diff
@@ -146,8 +146,8 @@ public static final KeyValueTableConfiguration encode(final KeyValueTableConfig
         Preconditions.checkNotNull(config.getKvtName(), "kvtName");
```
pbelgundi, line 146: You may consider adding a Preconditions check for PrimaryKey>0 & SecondaryKey>=0

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
```diff
@@ -435,7 +435,10 @@ public static final KeyValueTableConfig decode(String scopeName, String kvtName,
         Preconditions.checkNotNull(kvtName, "kvtName");
         Preconditions.checkArgument(config.getPartitionCount() > 0, "Number of partitions should be > 0.");
```
pbelgundi, line 436: Preconditions check for secondary and primary key could be added.

pbelgundi:
@anirudhkovuru Once this PR is approved, please ensure a successful a system test run on the PR branch.

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
anirudhkovuru: Added checks

client/src/main/java/io/pravega/client/control/impl/ModelHelper.java
anirudhkovuru: Added checks

andreipaduroiu (approved):

PR #6085, opened 2021-06-14
Issue 6021: OperationProcessor queue may be left closed while the container is in RUNNING state

### LINKED ISSUE #6021: OperationProcessor queue may be left closed while the container is in RUNNING state

RaulGracia:
**Change log description**  
Fixes a bug in which the BookkeeperLog could get closed throwing a non-fatal exception from the OperationProcessor viewpoint, leaving the latter one in an inconsistent state.

**Purpose of the change**  
Fixes #6021. 

**What the code does**  
Use `ObjectClosedException` as the exception for writes failed in the `BookkeeperLog` when it gets closed by any type of error. With this, the `OperationProcessor` will shut down completely, as it considers `ObjectClosedException` as a fatal exception to bubble up. Currently, `BookkeeperLog` is setting `CancellationException` as the exception when it gets closed, which is not considered fatal by the `OperationProcessor` to shut down (leaving it in an inconsistent state in which `operationQueue` is closed but the `OperationProcessor` is running).

**How to verify it**  
Added one unit test to check the shutdown behavior of the `OperationProcessor` both `CancellationException` and `ObjectClosedException`.


andreipaduroiu (approved):
LGTM. Please let me know when it's OK to merge (pending any other verifications you may be doing).

RaulGracia:
We have been testing this change for various days on a couple of clusters and we have not seen the original problem. While it is hard to say that the exact same situation has been reproduced during the testing, my recollection is that this issue appeared more easily and we have not been able to do so with various days of testing. Note that in all the logs analyzed we have seen the `BookkeeperLog` being closed in various ways and plenty of network glitches and container recoveries, which previously were able to trigger the original problem. Given that, and considering that the change itself is small and quite harmless, I'm inclined to merge this PR.

Issue #4647, opened 2020-03-26
SegmentHelper should use TableSegment Client


andreipaduroiu:
**Problem description**
When #4569 and #4333 are done, we should refactor `SegmentHelper.java` to use `io.pravega.client.tables.impl.TableSegmentImpl` since that class provides a nice API to the server-side table segments.

**Problem location**
`SegmentHelper.java`.

**Suggestions for an improvement**
Make use of the APIs in `TableSegmentImpl.` (when #4569 and #4333 are done)

PR #6091, opened 2021-06-17
Issue 5941: (KVT Beta2) Client Iterators

### LINKED ISSUE #5941: KVT Beta 2: Implement KVT PrimaryKey and Global iterators in Client
Implement the `keyIterator`/`entryIterator` for `KeyValueTable` for the following cases:
1. All Keys/Entries for a Primary Key
2. All Keys/Entries for a Primary Key that have Secondary Keys between two specified (may be optional) values (i.e., range iterators).
3. All Keys/Entries for a Primary key that have Secondary Keys beginning with a specified prefix (may be empty).
4. Primary Key Ranges and Prefixes (global - all segment iterators).
5. Some Client-based unit tests that have been disabled with #5934 or #5938) may be re-enabled (and updated) to verify parts of this. New tests will be required for point 3) above.

At this point, #5938 and #5940 should already be implemented. 

Until then, point 2) above refers to [this API](https://gist.github.com/andreipaduroiu/8fff0ba8953342972cf82fd4be5ab33f#file-kvt2_raw-java-L123) and point 3) refers to [this API](https://gist.github.com/andreipaduroiu/8fff0ba8953342972cf82fd4be5ab33f#file-kvt2_raw-java-L127).

andreipaduroiu:
**Change log description**  
Implemented client-side Key-Value Table Key/Entry Iterators for 
- Whole Primary Keys (with Secondary Key range or prefix)
- Primary Key Range (with all Secondary Keys for those PKs returned)
- Primary Key Prefix (with all Secondary Keys for those PKs returned)

Integrated these with the segment store side (un-ignoring integration tests).

**Purpose of the change**  
Fixes #5941.

**What the code does**  
- Unified `KeyValueTable` `keyIterator` and `entryiterator` into a single API, to help reduce duplication.
    - The new `iterator()` will return a `KeyValueTableIterator.Builder` which can be used to specify arguments, including iteration size and ranges, etc.
    - The Builder will have two APIs available: `keys` which will return a Key Iterator for the given arguments and `entries` which will return an Entry iterator for the given arguments.
- Retired the `IteratorState` concept from KVT
    - This applied only to internal, hash-based Table Segments.
-  Simplified `IteratorItem`
    - Removed the `state` concept from it. 
    - For Controller APIs, I created a `HashTableIteratorItem` which is in the `impl` package. At the moment this is used only by `SegmentHelper` (inside the Controller; there is no expectation that it will ever be used in public KVTs).
    - I considered moving this into the Controller project, however the end goal is that the Controller's `SegmentHelper` will eventually use the APIs provided here in the Client vs invoking the wire commands directly (see #4647).
- Updated `SegmentIteratorArgs` to support `fromKey` and `toKey`
    -  All Iterator arguments can be translated to key range iterators, hence these are the only two args needed to invoke segment iterators.
- `WireCommands.ReadTableKeys` and `WireCommands.ReadTableEntries`
    - Added `fromKey` and `toKey` as arguments.
    - ContinuationToken has been replaced with `fromKey`. As such, `fromKey` is overloaded to mean either continuation token for hash table segments or the beginning of a range for sorted, fixed-key-length table segments. The determination of when it means what is made on the server when it figures out what kind of table segment it deals with.
    - This is OK, as the continuation token itself is the next key to resume iteration from, so `fromKey` can be interpreted as that.

Client-side Iterator Implementation
- Contract defined in `KeyValueTableIterator` and implemented in `KeyValueTableIteratorImpl`
- The Builder for the iterator accepts arguments and converts them into a range iterator. The range is made of a [fromPrimaryKey, fromSecondaryKey] and a [toPrimarykey, toSecondaryKey] combination. When constructing it, it makes sure that the range is valid and satisfiable with the iterators.
- Both Primary Key iterators and Global Iterators can be constructed this way. The difference is that the latter may require fetching multiple Primary Keys, and is thus handled differently.
- Single-segment iterators iterate over a single Primary Keys. These iterators delegate to the underlying segment iterator and return that one.
- Multi-segment iterators issue parallel iterators to each Segment involved. Each such iterator is sorted, and the result is a "Merge Iterator" based on those ones (using a heap to figure out the iterator to fetch items from).

**How to verify it** 
Existing unit tests must pass. New unit tests added to validate functionality.


client/src/main/java/io/pravega/client/tables/impl/SegmentIteratorArgs.java
```diff
@@ -29,17 +30,58 @@
 @Builder
 class SegmentIteratorArgs {
     /**
-     * Optional. If specified, all items returned by {@link AsyncIterator#getNext()} will have {@link TableSegmentKey}s
-     * that begin with the specified prefix.
+     * The Table Segment Key to begin iteration at (inclusive).
      */
-    private final ByteBuf keyPrefixFilter;
+    @NonNull
+    private final ByteBuf fromKey;
```
tkaitchuck, line 36: Please use BufferView

client/src/main/java/io/pravega/client/tables/impl/SegmentIteratorArgs.java
```diff
@@ -29,17 +30,58 @@
...
+            // End of iteration.
+            return null;
+        }
+
+        final byte[] result = lastKey.copy().array();
+        int index = result.length - 1;
+        ByteBuf resultBuf = null;
+        while (index >= 0) {
+            int v = result[index] & 0xFF;
+            if (v >= 0xFF) {
```
tkaitchuck, line 70: Because it can be > 0xFF this is the same as an equality check, in which case if it is equal to 0xFF then adding 1 would result in 0 anyway, so this part can be combined with the other branch. The if is only needed for the break, which can be done by checking if the value after increment is not 0.

client/src/main/java/io/pravega/client/tables/impl/TableEntryHelper.java
```diff
@@ -0,0 +1,147 @@
...
+                    "Not expecting a Secondary Key.");
+            return Unpooled.wrappedBuffer(primaryKey);
+        } else {
+            Preconditions.checkArgument(secondaryKey.remaining() == this.config.getSecondaryKeyLength(),
+                    "Invalid Secondary Key Length. Expected %s, actual %s.", this.config.getSecondaryKeyLength(), secondaryKey.remaining());
+            return Unpooled.wrappedBuffer(primaryKey, secondaryKey);
+        }
+    }
+
+    private DeserializedKey deserializeKey(ByteBuf keySerialization) {
```
tkaitchuck, line 108: Can this be ByteBuffer?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1975,8 +1975,11 @@ void releaseInternal() {
         @ToString.Exclude
         final String delegationToken;
         final int suggestedKeyCount;
-        final ByteBuf continuationToken; // this is used to indicate the point from which the next keys should be fetched.
-        final ByteBuf prefixFilter;      // this is used to indicate any prefix filters to apply to keys.
+        /**
+         * See {@link ReadTableEntries#fromKey}.
+         */
+        final ByteBuf fromKey;
+        final ByteBuf toKey;
```
tkaitchuck, line 1982: This blatantly violates backward compatibility of the wireprotocol. Any old client contacting a new server would have their wirecommand completely misinterpreted.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2082,8 +2085,16 @@ void releaseInternal() {
...
+        /**
+         * Depending on Table Segment Type, {@link #fromKey} and {@link #toKey} have different meanings.
+         * - For Hash Table Segments, {@link #fromKey} is a continuation token generated on the server, while {@link #toKey} is not used.
+         * - For Fixed-Key-Length Table Segments, {@link #fromKey} and {@link #toKey} are used to indicate the bounds of the iteration.
+         *
+         * It is highly recommended to use the Client's TableSegmentImpl API which correctly encodes this in the requests
+         * it makes vs creating ad-hoc requests.
+         */
+        final ByteBuf fromKey;
+        final ByteBuf toKey;
```
tkaitchuck, line 2097: Same here

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1975,8 +1975,11 @@ void releaseInternal() {
         @ToString.Exclude
         final String delegationToken;
         final int suggestedKeyCount;
-        final ByteBuf continuationToken; // this is used to indicate the point from which the next keys should be fetched.
-        final ByteBuf prefixFilter;      // this is used to indicate any prefix filters to apply to keys.
+        /**
+         * See {@link ReadTableEntries#fromKey}.
```
tkaitchuck, line 1979: All changes require bumping the wire protocol version.

tkaitchuck (changes requested):

client/src/main/java/io/pravega/client/tables/KeyValueTableIterator.java
```diff
@@ -0,0 +1,140 @@
...
+import java.nio.ByteBuffer;
+import lombok.NonNull;
+
+/**
+ * Defines a preconfigured Iterator for a {@link KeyValueTable}, which can be used to iterate over {@link TableKey} or
+ * {@link TableEntry} instances.
+ *
+ * See {@link Builder} for configuring, and {@link #keys()} and {@link #entries()} for running.
+ */
+public interface KeyValueTableIterator {
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed.

client/src/main/java/io/pravega/client/tables/IteratorItem.java
```diff
@@ -20,24 +20,12 @@
...
 
 /**
- * An iteration result item returned by {@link AsyncIterator} when invoking {@link KeyValueTable#entryIterator}
- * or {@link KeyValueTable#keyIterator}.
+ * An iteration result item returned by {@link AsyncIterator} when invoking {@link KeyValueTable#iterator()}.
  *
  * @param <T> Iterator Item type.
  */
 @Data
 public class IteratorItem<T> {
```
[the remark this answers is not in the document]
andreipaduroiu: Fixed

client/src/main/java/io/pravega/client/tables/impl/HashTableIteratorItem.java
```diff
@@ -0,0 +1,128 @@
...
+            return this.token.readableBytes() == 0;
+        }
+
+        /**
+         * Deserializes the {@link State} from its serialized form obtained from calling {@link #getToken()}.
+         *
+         * @param serializedState A serialized {@link State}.
+         * @return The IteratorState object.
+         */
+        public static State fromBytes(ByteBuf serializedState) {
```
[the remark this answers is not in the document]
andreipaduroiu: These are used in the Controller. I don't want to move this class into the Controller package/project, as it deals with Table Segments. So I need to have these be public.

client/src/main/java/io/pravega/client/tables/impl/HashTableIteratorItem.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        /**
+         * Deserializes the IteratorState from its serialized form obtained from calling {@link #toBytes()} .
+         *
+         * @param serializedState A serialized {@link State}.
+         * @return The IteratorState object.
+         */
+        public static State fromBytes(ByteBuffer serializedState) {
```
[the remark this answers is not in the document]
andreipaduroiu: See above

client/src/main/java/io/pravega/client/tables/impl/TableEntryHelper.java
andreipaduroiu: The `TableSegment` API uses ByteBuf in all its APIs. This is preexisting code and I didn't add it in this feature branch.

This is the same reason why `SegmentIteratorArgs` has `ByteBuf` instead of `ByteBuffer`. All Table Segment APIs use that, while externally we expose `ByteBuffer`. 

I get it that you're trying to get rid of ByteBuf in the Client and I believe that is a worthwhile effort. Would you be OK if I create a follow-up issue to be done after this PR (or after we merge the feature branch in)? Changing ByteBuf to ByteBuffer requires delicate code changes to ensure we don't flip the read/write state accidentally or do other things that can corrupt the data. I prefer to have a dedicated PR that will cleanse it. The good part about it is that it won't touch the external API - it's all internals so from the POV of the user, nothing should change.

client/src/main/java/io/pravega/client/tables/impl/SegmentIteratorArgs.java
andreipaduroiu: See below 
https://github.com/pravega/pravega/pull/6091#discussion_r655550752

client/src/main/java/io/pravega/client/tables/impl/HashTableIteratorItem.java
```diff
@@ -0,0 +1,128 @@
...
+    @RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+    @EqualsAndHashCode
+    @Getter
+    public static class State {
+        /**
+         * No state. Providing this value will result in an iterator that starts from the beginning (i.e., not resuming an
+         * existing iteration).
+         */
+        public static final State EMPTY = new State(Unpooled.EMPTY_BUFFER);
+        private final ByteBuf token;
```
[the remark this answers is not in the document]
andreipaduroiu: See below https://github.com/pravega/pravega/pull/6091#discussion_r655550752

client/src/main/java/io/pravega/client/tables/impl/SegmentIteratorArgs.java
andreipaduroiu: Thanks for the tip. I simplified it as per your suggestion.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
andreipaduroiu: Done

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
andreipaduroiu: Thanks for the pointer. I agree with you. I brought back the old fields and reworked these 2 commands to share more code as opposed from copy-pasting it.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
andreipaduroiu: See above.

client/src/main/java/io/pravega/client/tables/impl/TableEntryHelper.java
tkaitchuck: That's fine

client/src/main/java/io/pravega/client/tables/impl/KeyValueTableIteratorImpl.java
```diff
@@ -0,0 +1,447 @@
...
+    @VisibleForTesting
+    static class PeekingIterator<T> {
+        private final AsyncIterator<IteratorItem<T>> innerIterator;
+        private final Function<T, TableKey> getKey;
+        private volatile Iterator<T> currentBatchIterator = null;
+        @Getter
+        private volatile T current = null;
+        @Getter
+        private volatile TableKey currentKey = null;
+
```
tkaitchuck, line 250: These three are always being set together and are all volatile this at the very least severely complicates the thread safety. Perhaps a new static inner class can be made which holds these three items immutably and then this class has a single `current` variable which is volatile and set when the iterator advances.

client/src/main/java/io/pravega/client/tables/impl/KeyValueTableIteratorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+                                    .collect(Collectors.toList()));
+                        }
+                        this.currentBatchIndex.set(0);
+                    });
+        }
+    }
+
+    @Data
+    @VisibleForTesting
+    static class PeekingIteratorItem<T> {
```
tkaitchuck, line 285: Strange name

tkaitchuck (approved):

client/src/main/java/io/pravega/client/tables/impl/KeyValueTableIteratorImpl.java
andreipaduroiu: I resolved this by 1) removing the iterator altogether (and having a dedicated collection with a moving index) and 2) creating a `PeekingIteratorItem` class that contains the `contents` and `contentsKey`.

PR #6102, opened 2021-06-23
Issue 5098: Controller/SegmentStore HealthCheck Integration

co-jo:
**Change log description**  

* Incorporates both the refactored `shared:rest` project as well the `shared:health` project.
* Sets up the rest endpoint and generated required models to server `Health` data externally.
* Adds the `HealthServiceManager` (the top level component) to the `ControllerServiceStarter`.

**Purpose of the change**  
Progress towards completion of #5323

**What the code does**  

* See above.

**How to verify it**  

* Run unit tests.


controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
```diff
@@ -391,6 +399,15 @@ protected void shutDown() throws Exception {
         log.info("Initiating controller service shutDown");
 
         try {
+            if (healthServiceManager != null) {
+                log.info("Stopping the HealthService.");
+                healthServiceManager.close();
+            }
+
+            if (healthServiceManager != null) {
```
andreipaduroiu, line 407: This is duplicated.

controller/src/main/java/io/pravega/controller/server/impl/ControllerServiceConfigImpl.java
```diff
@@ -80,15 +83,6 @@
                             storeClientConfig.getStoreType() == StoreType.PravegaTable,
                     "If controllerCluster is enabled, store type should be Zookeeper");
         }
-        if (eventProcessorConfig.isPresent()) {
```
andreipaduroiu, line 83: Why did you remove these?

controller/src/main/java/io/pravega/controller/server/impl/ControllerServiceConfigImpl.java
```diff
@@ -101,6 +95,7 @@
         this.restServerConfig = restServerConfig;
         this.retentionFrequency = retentionFrequency == null ? Duration.ofMinutes(Config.MINIMUM_RETENTION_FREQUENCY_IN_MINUTES)
                 : retentionFrequency;
+        this.healthCheckFrequency = healthCheckFrequency == null ? Duration.ofSeconds(10) : healthCheckFrequency;
```
andreipaduroiu, line 98: Is there a way we can default this the same way it is done above (see `this.retentionFrequency`). We'd have to add a field to `Config` for that.

controller/src/main/java/io/pravega/controller/server/rest/resources/HealthImpl.java
```diff
@@ -0,0 +1,292 @@
...
+
+    private void getReadiness(String id, SecurityContext securityContext, AsyncResponse asyncResponse, String method) {
+        long traceId = LoggerHelpers.traceEnter(log, method);
+        try {
+            restAuthHelper.authenticateAuthorize(getAuthorizationHeader(), authorizationResource.ofScopes(), READ_UPDATE);
+            boolean ready = endpoint.isReady(id);
+            asyncResponse.resume(Response.status(Response.Status.OK)
+                    .entity(ready)
+                    .build());
+        } catch (AuthException e) {
```
andreipaduroiu, line 190: It would be nice if we could factor out this repeated try-catch logic into its own method. That should simplify the code in this class.

For example, 

```
private void handleException(Runnable r) {
  try {
    r.run();
  } /// put ugly catch code here
}
```

then you'd invoke it like:
```
handleException(() -> { your code here });
```

shared/health/src/main/java/io/pravega/shared/health/ContributorNotFoundException.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.pravega.shared.health;
+
+public class ContributorNotFoundException extends RuntimeException {
```
andreipaduroiu, line 18: Javadoc explaining what the exception is.

shared/controller-api/src/main/swagger/Controller.yaml
```diff
@@ -26,28 +26,33 @@ info:
...
-  description: "Stream related APIs"
-  externalDocs:
-    description: "Find out more about pravega"
-    url: "https://pravega.io"
-- name: "ReaderGroups"
-  description: "Reader group related APIs"
-  externalDocs:
-    description: "Find out more about pravega"
-    url: "https://pravega.io"
+  - name: "Scopes"
```
andreipaduroiu, line 29: Is it possible to revert these spacing changes?

andreipaduroiu (changes requested):

controller/src/main/java/io/pravega/controller/server/impl/ControllerServiceConfigImpl.java
co-jo: Accidental.

shared/controller-api/src/main/swagger/Controller.yaml
co-jo: I didn't actually intend to make these changes, was some automatic linting from IntelliJ. However, it does look like these changes makes the indentation consistent across the entire file. For example in certain places it indents starting on the first non '-' and others it indents starting at the '-'. Before the PR, it seems majority followed the latter convention.

controller/src/main/java/io/pravega/controller/server/ControllerServiceStarter.java
```diff
@@ -391,6 +399,11 @@ protected void shutDown() throws Exception {
         log.info("Initiating controller service shutDown");
 
         try {
+            if (healthServiceManager != null) {
+                log.info("Stopping the HealthService.");
+                healthServiceManager.close();
```
andreipaduroiu, line 404: For some reason, these method are called once in `shutDown` and once in `close. I see that the other components are closed again in `close`. I think it would be worth it to add it there too.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
```diff
@@ -99,6 +107,17 @@ public void start() throws Exception {
             statsProvider.start();
         }
 
+        log.info("Initializing HealthService ...");
+        healthServiceManager = new HealthServiceManager(serviceConfig.getHealthCheckInterval());
+        healthServiceManager.start();
+
+        log.info("Initializing RESTServer ...");
+        restServer = new RESTServer(serviceConfig.getRestServerConfig(), Set.of(new HealthImpl(
```
andreipaduroiu, line 115: Just wondering - why do we need a collection here? Isn't it just one element? Is it because that's how the `RESTService` requires a collection? 

FYI `Collections.singleton` is equivalent here.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
```diff
@@ -51,6 +53,9 @@
     public static final Property<Boolean> SECURE_ZK = Property.named("zk.connect.security.enable", false, "secureZK");
     public static final Property<String> ZK_TRUSTSTORE_LOCATION = Property.named("zk.connect.security.tls.trustStore.location", "", "zkTrustStore");
     public static final Property<String> ZK_TRUST_STORE_PASSWORD_PATH = Property.named("zk.connect.security.tls.trustStore.pwd.location", "", "zkTrustStorePasswordPath");
+    public static final Property<String> REST_LISTENING_HOST = Property.named("rest.listener.host", "localhost");
+    public static final Property<Integer> REST_LISTENING_PORT = Property.named("rest.listener.port", 6061);
+    public static final Property<Integer> HEALTH_CHECK_INTERVAL = Property.named("health.interval", 10);
```
andreipaduroiu, line 58: Always include the time unit in your property names. For example `health.interval.seconds` (I assume it's seconds, right?)

shared/health-bindings/src/test/java/io/pravega/shared/health/bindings/HealthTests.java
```diff
@@ -0,0 +1,366 @@
...
+        Assert.assertEquals("Unexpected number of child contributors.",
+            0,
+            healthResult.getChildren().size());
+
+        Assert.assertEquals("Child HealthContributor failed to report its detail entry.",
+                1,
+                healthResult.getDetails().size());
+    }
+
+    // Be mindful that the results will be dependent on the StatusAggregatorRule used.
```
andreipaduroiu, line 158: I don't understand this comment. Is it a TODO? If not, then I don't think it belongs here - should it go next to some actual check you're making?

shared/health-bindings/src/test/java/io/pravega/shared/health/bindings/HealthTests.java
```diff
@@ -0,0 +1,366 @@
...
+        assertStatus(streamResourceURI, HealthStatus.DOWN);
+
+        // Make sure that even though we have a majority of healthy reports, we still are considered failing.
+        healthServiceManager.getRoot().register(new StaticHealthyContributor("sample-healthy-indicator-two"));
+        streamResourceURI = UriBuilder.fromUri(getURI("/v1/health/status"))
+                .scheme(getURLScheme()).build();
+        assertStatus(streamResourceURI, HealthStatus.DOWN);
+    }
+
+    // Service Readiness is in essence a proxy for the Status as a HealthComponent does not provide its own
```
andreipaduroiu, line 189: Same here.

shared/health-bindings/src/test/java/io/pravega/shared/health/bindings/HealthTests.java
co-jo: Many of these comments are outdated and no longer valid, will make sure to check of references to outdated concepts.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
co-jo: That's correct. `RestApplication` accepts `Set<Object>` in its constructor.

pbelgundi (changes requested):

shared/controller-api/src/main/swagger/Controller.yaml
```diff
@@ -388,6 +393,196 @@ paths:
...
+      produces:
+        - application/json
+      responses:
+        200:
+          description: The alive status.
+          schema:
+            type: boolean
+        500:
+          description: Internal server error while fetching the liveness state of the Controller.
+  /health/details:
```
pbelgundi, line 441: What is the difference between "/health/details" API and the "/health" API? Both seem to return the _Health_ object for Controller.

pbelgundi (changes requested):

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
```diff
@@ -62,12 +63,17 @@ public String getDefaultContributorName() {
     @NonNull
     @Override
     public Health getHealth(String id) {
+        Health health;
         if (id == null || id.equals(root.getName())) {
-            return updater.getLatestHealth();
+            health = updater.getLatestHealth();
         } else {
             List<String> path = Arrays.asList(id.split("/"));
```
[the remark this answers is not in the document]
co-jo: That is correct. Good suggestion.

shared/controller-api/src/main/swagger/Controller.yaml
co-jo: `/health/details` returns just the `Map<String, Object>` associated with the `Health` object, instead of the `Health` object itself.

shared/health/src/main/java/io/pravega/shared/health/HealthEndpoint.java
```diff
@@ -112,4 +112,5 @@ default boolean isAlive() {
      * @return The {@link Health} of the whole health service.
      */
     Health getHealth();
+
```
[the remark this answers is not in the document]
co-jo: I'm not terribly familiar with how the operators work or what information is has available to them. From my perspective the only thing to mention is the use of the liveness/readiness routes.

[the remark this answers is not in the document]
pbelgundi: Operators are first hand users of this Health Framework, so it it may help to understand their needs. Today, they only invoke Readiness and Liveness checks on a component (top level component - controller/segment store & Bookkeeper) and expect to get a boolean result indicating weather the check passed or failed. It may help to update the PDP with details of how these can be invoked for each component.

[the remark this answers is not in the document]
pbelgundi: How would the Bookkeeper operator check on BK Health? Is it expected to be a sub-component inside Segment Store, since there is no separate Health Service for Bookkeeper...

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
```diff
@@ -62,12 +63,17 @@ public String getDefaultContributorName() {
     @NonNull
     @Override
     public Health getHealth(String id) {
+        Health health;
         if (id == null || id.equals(root.getName())) {
-            return updater.getLatestHealth();
+            health = updater.getLatestHealth();
```
pbelgundi, line 68: I don't see the actual health checks implemented in this PR. (_A class that implements AbstractHealthContributor_ in Controller and Segment Store) . Is that going to be part of another PR?

pbelgundi (changes requested):

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
co-jo: That's correct. Those are intended to be added gradually.

shared/health/src/main/java/io/pravega/shared/health/HealthEndpoint.java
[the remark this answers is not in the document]
co-jo: > 
> 
> How would the Bookkeeper operator check on BK Health? Is it expected to be a sub-component inside Segment Store, since there is no separate Health Service for Bookkeeper...

A `HealthContributor` would need to be implemented for this purpose, which would then be called via the `/v1/health/readiness/bookkeeper` route.

shared/health/src/test/java/io/pravega/shared/health/HealthManagerTests.java
```diff
@@ -80,14 +79,14 @@ public void testHealth() throws Exception {
      */
     @Test
     public void testHealthInvalidName() {
-        Assert.assertNull("An exception should be thrown given an unregistered contributor.",
-                service.getEndpoint().getHealth("unknown-contributor-name"));
+        AssertExtensions.assertThrows("An exception should be thrown given an unregistered contributor.",
+                () -> TestHealthContributors.awaitHealthContributor(service, "unknown-contributor-name"),
+                e -> e instanceof TimeoutException);
```
andreipaduroiu, line 84: What is the timeout here? I hope it is really small, otherwise I would prefer if this test didn't take too long.

Is this test taking less than 1 second to finish?

co-jo: It times out after 3 times the interval supplied for the `HealthServiceManager` which is 1 second in this case. I will change all instances to instead take a timeout of 100 milliseconds.

andreipaduroiu (approved):

shared/controller-api/src/main/swagger/Controller.yaml
pbelgundi: Okay.

shared/health/src/main/java/io/pravega/shared/health/HealthEndpoint.java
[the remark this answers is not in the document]
pbelgundi: Ofcourse a HealthContributor needs to be implemented... my understanding is this would be a child under Segment Store HealthContributor..... Since there is no separate Health REST service for Bookkeeper?

shared/health/src/main/java/io/pravega/shared/health/impl/HealthEndpointImpl.java
pbelgundi: Okay

shared/health/src/main/java/io/pravega/shared/health/HealthEndpoint.java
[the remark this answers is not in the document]
co-jo: That's right. Are you suggesting it be handled differently? I'm not sure I understand your concern.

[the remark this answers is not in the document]
pbelgundi: No. I'm not suggesting it to be handled differently. I'm just trying to get clarity on how this is expected to work. Cool. I think this is great

pbelgundi (approved):
LGTM

PR #6069, opened 2021-05-31
Issue 6067: Auto scaledown is triggered for segments with steady traffic between target/2 and target rates. 

ldr7:
Signed-off-by: Atharva <Atharva.Joshi@dell.com>

**Change log description**  
Adds a get cache call in the report method for AutoScaleProcessor and adds a test covering the scenario.  

**Purpose of the change**  
Fixes #6067 

**What the code does**  

- Adds a get cache call in every report so that the last access time is updated. 
- Adds a test covering the scenario. Report is called on a segment before cache expiry and its ensured that cache cleanup does not evict the segment from the cache.
- Implements constructors in AutoScaleProcessor to enable the test.
- Creates classes in SimpleCacheTests to enable the test.
- Adds a test time dependency for segmentstore:host on common:test. 

**How to verify it**  
Run the unit test - AutoScaleProcessorTest.testCacheGetOnReport

PR #6103, opened 2021-06-24
Issue 6069: Max batch size not being honoured sometimes

shiveshr:

**Change log description**  
Fixes the issue that was causing Max batch count for txn to exceed the max limit.

**Purpose of the change**  
Fixes #6069 

**What the code does**  
  the issue was that whenever we had collected 100 txns we thought we were breaking from the loop (but the loop had a switch case and the [break](https://github.com/pravega/pravega/blob/fe8a2a3dbfcef518b2749ab46cdcbf75a16bdae0/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java#L2260) actually applied to the switch case and not to the for loop. which meant that we could end up going up to 2 x limit -1 the batch count in case where we did not get limit items from first iteration and end up picking up to limit elements from second iteration. 

**How to verify it**  
Unit test added

### DIFF
--- a/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
+++ b/controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
@@ -2244,6 +2244,9 @@ public abstract class PersistentStreamBase implements Stream {
         return Futures.loop(() -> from.get() < txnIds.size() && transactionsMap.size() < limit, 
                 () -> getTransactionRecords(epoch, txnIds.subList(from.get(), till.get()), context).thenAccept(txns -> {
             for (int i = 0; i < txns.size(); i++) {
+                if (transactionsMap.size() >= limit) {
+                    break;
+                }
                 ActiveTxnRecord txnRecord = txns.get(i);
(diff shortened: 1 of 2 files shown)


controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -2244,6 +2244,9 @@ boolean compareWriterPositions(Map<Long, Long> position1, Map<Long, Long> positi
         return Futures.loop(() -> from.get() < txnIds.size() && transactionsMap.size() < limit, 
                 () -> getTransactionRecords(epoch, txnIds.subList(from.get(), till.get()), context).thenAccept(txns -> {
             for (int i = 0; i < txns.size(); i++) {
```
pbelgundi, line 2246: Instead of the 2 break statements, we could add an additional check to the for loop entry condition `transactionsMap.size<limit`

pbelgundi (changes requested):

pbelgundi (approved):
LGTM

PR #6108, opened 2021-06-28
Issues 6086: SLTS - Robust garbage collection (PDP-53).

### LINKED ISSUE #6086: SLTS: Robust garbage collection (PDP-53)
This is a tracking task for implementation of [PDP-53](https://github.com/pravega/pravega/wiki/PDP-53:-Robust-garbage-collection-for-SLTS)

- Currently there are several  cases where garbage chunks won't be properly deleted.
- Please refer to [Shortcomings of the current implementation](https://github.com/pravega/pravega/wiki/PDP-53:-Robust-garbage-collection-for-SLTS#shortcomings-of-the-current-implementation)  for details.

[PDP-53](https://github.com/pravega/pravega/wiki/PDP-53:-Robust-garbage-collection-for-SLTS) discusses a change to our SLTS garbage collection implementation that would make it more robust and comprehensively cover additional failure modes without adding complexity.

Please refer to [Discarded approaches](https://github.com/pravega/pravega/wiki/PDP-53:-Robust-garbage-collection-for-SLTS#discarded-approaches) in PDP-53

**Additional context**

-  [PDP 51: Segment Container Event Processor](https://github.com/pravega/pravega/wiki/PDP-51:-Segment-Container-Event-Processor)
-  [SLTS Design](https://github.com/pravega/pravega/wiki/PDP-34:-Simplified-Tier-2)

sachin-j-joshi:
Initial implementation of PDP-53

**Change log description**  
Initial implementation of [PDP-53](https://github.com/pravega/pravega/wiki/PDP-53:-Robust-garbage-collection-for-SLTS).
Make Garbage collection in SLTS more robust and comprehensively cover additional failure modes without adding complexity.

**Purpose of the change**  
Fix #6086 

**What the code does**  
_Design_ 

- Please refer to [PDP-53](https://github.com/pravega/pravega/wiki/PDP-53:-Robust-garbage-collection-for-SLTS)

The main crux of the change is 2 fold 
1. Pre insert all GC tasks  to GC queue before actual commit of transaction. 
2. In the GC code process those GC tasks only if the   
This way we eliminate whole class of GC leaks that can happen because  of failure after transaction is committed.

_What is included in this PR_ 
- Includes bulk of changes say (95%) required for implementing PDP-53. This part does not yet use Event processor.
- Hooking up with actual Event processor

_What is NOT included in this PR_ 
- GC of the journal files

_What code does_ 

- Rewrite GarbageCollector class to deal with new logic. Corresponding changes to all GC tests and new tests for new logic.
- Change various operations to move GC task enqueue before transaction commit. 
- Segment delete operation now works in three steps 
        - in first step it marks the segment deleted and enqueues a task, 
        - in second step it enumerates all the chunks for deleted segments and enqueues GC tasks for each of them . This step also cleans up other non-chunk metadata
        - lastly each chunk is deleted indecently as each GC task for chunks is executed.
- GC does not directly depend on EventWriter api . Instead it uses `AbstractTaskQueue<TaskType>`  to abstract out implementation details of persistent queue.
- Most `ChunkedSegmentStorageTests `now also check for presence or absence of appropriate GC tasks for each test scenario.
- other small changes 
  - Prevent deletion of chunks from segments which have active transactions associated with them.
  - Change throttling logic.  
  - Get rid of separate suspended state.

**How to verify it**  
All tests should pass


segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/FileSystemIntegrationTest.java
```diff
@@ -60,6 +60,8 @@ protected ServiceBuilder createBuilder(ServiceBuilderConfig.Builder configBuilde
                 .withStorageFactory(setup -> useChunkedSegmentStorage ?
                         new FileSystemSimpleStorageFactory(ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder()
                                 .journalSnapshotInfoUpdateFrequency(Duration.ofMillis(10))
+                                .garbageCollectionDelay(Duration.ofMillis(10))
```
andreipaduroiu, line 63: I think we should do this for every LTS integration test, not just for File System.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -229,7 +229,12 @@ private MetadataStore createMetadataStore() {
             ChunkedSegmentStorage chunkedStorage = (ChunkedSegmentStorage) this.storage;
             val snapshotInfoStore = getStorageSnapshotInfoStore();
             // Bootstrap
-            return chunkedStorage.bootstrap(snapshotInfoStore);
+            // Note: Temp change to be removed after next PR (part 2 of 3)
```
andreipaduroiu, line 232: Please put a TODO here and link to a GitHub issue.

Or, alternatively, create a feature branch to collect your PRs in it.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -306,7 +311,14 @@ protected void doStart() {
         }
 
         // Delayed start. Secondary services need not be started in order for us to accept requests.
-        delayedStart.thenComposeAsync(v -> startSecondaryServicesAsync(), this.executor)
+        delayedStart
+                .thenComposeAsync( v -> {
+                    if (this.storage instanceof ChunkedSegmentStorage) {
```
andreipaduroiu, line 316: Please roll this into `startSecondaryServicesAsync`

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/AbstractTaskQueue.java
```diff
@@ -0,0 +1,41 @@
...
+package io.pravega.segmentstore.storage.chunklayer;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Represents background task queue.
+ *
+ * @param <TaskType>
+ */
+public interface AbstractTaskQueue<TaskType> extends AutoCloseable {
```
andreipaduroiu, line 26: This is not a queue, but a collection of queues. 

I am trying to understand why do we need yet another abstraction for this. Could we not just have a collection of queues (index them by name), and get them by name; then add the task to those objects directly?

This interface, as it is, doesn't explain what happens if we try to add a task to an inexistent queue, for example. To be more concise, this interface (if you decide to keep it), needs a bit more documentation on what each method is doing and what is the caller's expectation on each of those futures returned by it.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkStorageMetrics.java
```diff
@@ -65,6 +65,19 @@
     static final Counter SLTS_SYSTEM_WRITE_BYTES = STATS_LOGGER.createCounter(MetricsNames.SLTS_SYSTEM_WRITE_BYTES);
     static final Counter SLTS_CONCAT_BYTES = STATS_LOGGER.createCounter(MetricsNames.SLTS_CONCAT_BYTES);
 
+    static final Counter SLTS_GC_TASK_PROCESSED = STATS_LOGGER.createCounter(MetricsNames.SLTS_GC_TASK_PROCESSED);
```
andreipaduroiu, line 68: I think by now we have more metrics for SLTS than we have for everything else :)

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -176,7 +178,10 @@ public ChunkedSegmentStorage(int containerId, ChunkStorage chunkStorage, ChunkMe
                 chunkStorage,
                 metadataStore,
                 config,
-                executor);
+                executor,
+                System::currentTimeMillis,
```
andreipaduroiu, line 182: FYI `currentTimeMillis` is orders of magnitude slower than `nanoTime`. If all you need is a timer, consider using `AbstractTimer`; there's a `Timer` implementation using `System.nanoTime` and you can also have a `ManualTimer` (it's somewhere in the unit test package) which you can control at will.

andreipaduroiu: Also why do you expose both this and the "delayed future` below? Would it not have been simpler to create 2 constructors for this?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -191,14 +196,18 @@ public ChunkedSegmentStorage(int containerId, ChunkStorage chunkStorage, ChunkMe
...
         this.logPrefix = String.format("ChunkedSegmentStorage[%d]", containerId);
-
+        this.taskQueue = taskQueue;
         // Now bootstrap
-        return this.systemJournal.bootstrap(epoch, snapshotInfoStore)
-                .thenRun(() -> garbageCollector.initialize());
+        return this.systemJournal.bootstrap(epoch, snapshotInfoStore);
+    }
+
+    public CompletableFuture<Void> finishBootstrap() {
```
andreipaduroiu, line 209: Javadoc

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/mocks/InMemoryTaskQueue.java
```diff
@@ -0,0 +1,54 @@
...
+import io.pravega.segmentstore.storage.chunklayer.GarbageCollector;
+import lombok.Getter;
+import lombok.val;
+
+import java.util.ArrayList;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
+
+public class InMemoryTaskQueue implements AbstractTaskQueue<GarbageCollector.TaskInfo> {
```
andreipaduroiu, line 28: Can this be moved to the test package?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/mocks/InMemoryTaskQueue.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public CompletableFuture<Void> addQueue(String queueName, Boolean ignoreProcessing) {
+        taskQueueMap.put(queueName, new LinkedBlockingQueue<GarbageCollector.TaskInfo>());
+        return CompletableFuture.completedFuture(null);
+    }
+
+    @Override
+    public synchronized CompletableFuture<Void> addTask(String queueName, GarbageCollector.TaskInfo task) {
+        taskQueueMap.get(queueName).add(task);
```
andreipaduroiu, line 40: NullPointerException

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/mocks/InMemoryTaskQueue.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
+
+public class InMemoryTaskQueue implements AbstractTaskQueue<GarbageCollector.TaskInfo> {
+    @Getter
+    private final ConcurrentHashMap<String, LinkedBlockingQueue<GarbageCollector.TaskInfo>> taskQueueMap = new ConcurrentHashMap<>();
+
+    @Override
+    public CompletableFuture<Void> addQueue(String queueName, Boolean ignoreProcessing) {
+        taskQueueMap.put(queueName, new LinkedBlockingQueue<GarbageCollector.TaskInfo>());
```
andreipaduroiu, line 34: What if it's already there?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/mocks/InMemoryTaskQueue.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public synchronized CompletableFuture<Void> addTask(String queueName, GarbageCollector.TaskInfo task) {
+        taskQueueMap.get(queueName).add(task);
+        return CompletableFuture.completedFuture(null);
+    }
+
+    public ArrayList<GarbageCollector.TaskInfo> drain(String queueName, int maxElements) {
+        val list = new ArrayList<GarbageCollector.TaskInfo>();
+        taskQueueMap.get(queueName).drainTo(list, maxElements);
```
andreipaduroiu, line 46: same here

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
```diff
@@ -95,9 +95,11 @@ public void testCreate() throws Exception {
                     ex -> ex instanceof StreamSegmentExistsException);
 
             // Delete and make sure it can be recreated.
+            /*
             s.openWrite(segmentName).thenCompose(handle -> s.delete(handle, null)).join();
```
andreipaduroiu, line 99: Put a TODO here with an associated GitHub issue so you don't forget about it.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageMockTests.java
```diff
@@ -339,73 +338,6 @@ public void testIOExceptionDuringWrite() throws Exception {
         //verify(spyChunkStorage, times(1)).doDelete(any());
     }
 
-    @Test
-    public void testFileNotFoundExceptionDuringGarbageCollection() throws Exception {
```
andreipaduroiu, line 343: Was this deleted on purpose or is it temporary too?

shared/metrics/src/main/java/io/pravega/shared/MetricsNames.java
```diff
@@ -160,6 +160,20 @@
 
     public static final String SLTS_GC_QUEUE_SIZE = PREFIX + "segmentstore.storage.slts.GC_queue_record_count";  // Counter
 
+    public static final String SLTS_GC_TASK_PROCESSED = PREFIX + "segmentstore.storage.slts.GC.task_processed_count";          // Counter
+
+    public static final String SLTS_GC_CHUNK_NEW = PREFIX + "segmentstore.storage.slts.GC.chunk_new_count";                     // Counter
+    public static final String SLTS_GC_CHUNK_QUEUED = PREFIX + "segmentstore.storage.slts.GC.chunk_queued_count";           // Counter
```
andreipaduroiu, line 166: Please fix the comment indentation (on the right) for all of these.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollectorTests.java
```diff
@@ -36,21 +41,24 @@
...
+import static org.mockito.Mockito.*;
+
 /**
  * Tests for {@link GarbageCollector}.
  */
 @Slf4j
 public class GarbageCollectorTests extends ThreadPooledTestSuite {
     public static final int CONTAINER_ID = 42;
+    public static final long TXN_ID = 123;
     protected static final Duration TIMEOUT = Duration.ofSeconds(3000);
```
andreipaduroiu, line 62: This is pretty large. Can we set it to 30 seconds please.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -90,14 +104,28 @@
...
+    @Getter
+    private final String taskQueueName;
+
+    @Getter
+    private final String failedQueueName;
+
+    /**
+     * Instance of {@link MultiKeySequentialProcessor}.
+     */
+    private final MultiKeySequentialProcessor<String> taskSchedular;
```
andreipaduroiu, line 127: typo in name

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
     }
 
     /**
      * Initializes this instance.
+     * @param taskQueue Task queue to use.
      */
-    public void initialize() {
-        Services.startAsync(this, this.executor);
+    public CompletableFuture<Void> initialize(AbstractTaskQueue<TaskInfo> taskQueue) {
+        // Temp change to be removed after next PR (part 2 of 3)
```
andreipaduroiu, line 181: TODO with Github Issue

Please write TODOs with Github issue(s) below as well - I see you used similar patterns to disable features.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
-                                log.error("{}: Error during doRun.", traceObjectId, ex);
-                            }
-                            return null;
-                        }, executor),
-                executor);
-        return loopFuture;
+    CompletableFuture<Void>  addChunkToGarbage(long transactionId, String chunkToDelete, long startTime, int attempts) {
+        if (null != taskQueue) {
+            return taskQueue.addTask(taskQueueName, new TaskInfo(chunkToDelete, startTime, attempts, TaskInfo.DELETE_CHUNK, transactionId))
+                    .thenRunAsync(() -> {
```
andreipaduroiu, line 216: Does this need to be `Async`? Looks like a very small method.

Same below.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                        log.debug("{}: deleteGarbage - Segment is not marked as deleted. segment={}.", traceObjectId, streamSegmentName);
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        return new ChunkIterator(storageExecutor, txn, segmentMetadata)
+                                .forEach((metadata, name) -> {
+                                    metadata.setActive(false);
+                                    txn.update(metadata);
+                                    chunksToDelete.add(name);
+                                })
+                                .thenRunAsync(() -> this.addChunksToGarbage(txn.getVersion(), chunksToDelete), storageExecutor)
```
andreipaduroiu, line 288: Combine these two thenRun + the next thenCompose into a single method.

andreipaduroiu (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                        queueSize.incrementAndGet();
+                        SLTS_GC_CHUNK_QUEUED.inc();
+                    }, this.storageExecutor);
+        }
+        return CompletableFuture.completedFuture(null);
     }
 
-    private boolean canRun() {
-        return isRunning() && getStopException() == null && !closed.get();
+    CompletableFuture<Void>  addSegmentToGarbage(long transactionId, String segmentToDelete) {
```
RaulGracia, line 224: Nit: extra space before method name.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+        if (null != taskQueue) {
```
RaulGracia, line 225: If `taskQueue`, this method just does nothing right? Is that ok or we should check that `taskQueue` is not null and throw otherwise?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+            val startTime = currentTimeSupplier.get() + config.getGarbageCollectionDelay().toMillis();
```
RaulGracia, line 226: I think I have seen the behavior of this reflected in the metrics of our experiments. This value, which is 60 seconds if I'm not mistaken, is just a fixed delay for work that needs to be done anyway, right? My question is: why is this the right behavior? Why do we need to buffer in the queue 60 seconds worth of tasks before start processing them? Can this lead to overloading the queue under a spike of delete tasks?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
-    /**
-     * Adds given chunks to list of garbage chunks.
-     *
-     * @param chunksToDelete List of chunks to delete.
-     */
-    void addToGarbage(Collection<String> chunksToDelete) {
-        val currentTime = currentTimeSupplier.get();
-
-        chunksToDelete.forEach(chunkToDelete -> addToGarbage(chunkToDelete, currentTime + config.getGarbageCollectionDelay().toMillis(), 0));
+    CompletableFuture<Void>  trackNewChunk(long transactionId, String chunktoTrack) {
```
RaulGracia, line 247: Nit: extra spaces before method name.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+        if (null != taskQueue) {
```
RaulGracia, line 248: Same comment about `taskQueue` being null.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+            val startTime = currentTimeSupplier.get() + config.getGarbageCollectionDelay().toMillis();
+            return taskQueue.addTask(taskQueueName, new TaskInfo(chunktoTrack, startTime, 0, TaskInfo.DELETE_CHUNK, transactionId))
```
RaulGracia, line 250: The method is about a `trackNewChunk` but the `TaskInfo` is `TaskInfo.DELETE_CHUNK`. Is that right?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                        SLTS_GC_CHUNK_NEW.inc();
+                    }, this.storageExecutor);
+        }
+        return CompletableFuture.completedFuture(null);
+    }
 
-        if (queueSize.get() >= config.getGarbageCollectionMaxQueueSize()) {
-            log.warn("{}: deleteGarbage - Queue full. Could not delete garbage. Chunks skipped", traceObjectId);
+    private CompletableFuture<Void> failTask(TaskInfo infoToRetire) {
+        if (null != taskQueue) {
```
RaulGracia, line 260: Same comment about taskQueue being null.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                                .thenRunAsync(() -> deleteBlockIndexEntriesForChunk(txn, streamSegmentName, segmentMetadata.getStartOffset(), segmentMetadata.getLength()), storageExecutor)
+                                .thenComposeAsync(v -> {
+                                    txn.delete(segmentMetadata.getName());
+                                    return txn.commit();
+                                }, storageExecutor)
+                                .handleAsync((v, e) -> {
+                                    if (null != e) {
+                                        log.error(String.format("%s deleteGarbage - Could not delete metadata for garbage segment=%s.",
+                                                traceObjectId, streamSegmentName), e);
+                                        failed.set(e);
```
RaulGracia, line 298: Why do you need this variable? Could the `thenComposeAsync` just get as input whether the computation has completed successfully or not (e.g., boolean) and use it instead of having this `AtomicReference<Throwable>`? You are not event using the throwable object inside that variable in `thenComposeAsync`, right?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
-            log.info("{}: deleteGarbage - suspended - sleeping for {}.", traceObjectId, config.getGarbageCollectionDelay());
-            return CompletableFuture.completedFuture(false);
-        }
+     * @param operation    The Callable to execute.
+     * @param <R>       Return type of the operation.
+     * @param keyNames The names of the keys involved in this operation (for sequencing purposes).
+     * @return A CompletableFuture that, when completed, will contain the result of the operation.
+     * If the operation failed, it will contain the cause of the failure.
+     * */
+    private <R> CompletableFuture<R> executeSerialized(Callable<CompletableFuture<R>> operation, String... keyNames) {
```
RaulGracia, line 368: If you think that this could be a more general utility in other parts of the code, maybe you can think of moving it to `common` package. If not, I don't know if making these methods so generic is worth it, as I see that they are private and only called from one place.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                                .thenComposeAsync(v -> f, storageExecutor));
+                    } else {
+                        futures.add(f);
+                    }
+                }
+            }
         }
+        return Futures.allOf(futures)
+                .thenRunAsync(() -> {
+                    queueSize.addAndGet(-1 * batch.size());
```
RaulGracia, line 353: This looks odd. You can use just `-batch.size()` or `Math.negateExact()`.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -374,22 +501,87 @@ public void report() {
         ChunkStorageMetrics.DYNAMIC_LOGGER.reportGaugeValue(SLTS_GC_QUEUE_SIZE, queueSize.get());
     }
 
-    @RequiredArgsConstructor
+    /**
+     * Represents a Task info.
+     */
+    public static abstract class AbstractTaskInfo {
+        public static final int DELETE_CHUNK = 1;
```
RaulGracia, line 508: Why do you use these hardcoded values for determining the types of tasks? What about if we add or delete operations in the future? I would suggest to use a subclass per task type and then discriminate tasks based on their instance types. Alternatively, enums could be also another approach here.

RaulGracia (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/AbstractTaskQueue.java
sachin-j-joshi: The main reason is that storage can not have direct dependency on code in container.
In addition this abstraction is at the GC concept level without need to also worry about details of serialization/deserialization etc. etc.
The third and also equally important reason is that it makes GC implementation very loosely coupled to actual underlying event processor implementation.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
sachin-j-joshi: ok makes sense

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
sachin-j-joshi: The second PR depended on Raul's change.
Now that that PR is in, I can combine part 1 and 2 into a single PR.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -306,7 +311,14 @@ protected void doStart() {
...
+                        return ((ChunkedSegmentStorage) this.storage).finishBootstrap();
```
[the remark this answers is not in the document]
sachin-j-joshi: This is because the storage unit tests can not depend on container code. 
All unit tests work at higher level of abstraction of `AbstractTaskQueue<TaskType>` and do not need to depend on actual underlying implementation.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -87,7 +87,7 @@
 
     /**
      * Metadata store containing all storage data.
-     * Initialized by segment container via {@link ChunkedSegmentStorage#bootstrap(SnapshotInfoStore)} ()}.
+     * Initialized by segment container via {@link ChunkedSegmentStorage#bootstrap(SnapshotInfoStore, AbstractTaskQueue)} ()}.
```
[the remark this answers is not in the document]
sachin-j-joshi: AbstractTaskQueue is a way to untangle dependencies.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
sachin-j-joshi: System::currentTimeMillis is a method to get current time.
Delayed future is supplier of delayed execution.
These are two separate unrelated things.

Having these as suppliers allows me to precise control the sequence of steps/scenarios in unit tests.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: I'll be combining PR part 1 and 2 , now that Raul's changes are in.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: yes it needs to Async as the underlying implementation of taskQueue is using EventProcessor async API

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: It is checked during initialization of container and as such it is not possible to have.
However during many of unit tests GC is not initialized and the taskQueue will be null and in fact it is not required to be initialized in that particular scenario. 
In that sense it is okay for it to be null in unit tests and yet it is correctly enforced when GC tests initialize GC and integration tests initialize container etc.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: This is correct. 

I'll add a comment

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,339 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                    val failed = new AtomicReference<Throwable>();
+                    if (null == segmentMetadata) {
+                        log.debug("{}: deleteGarbage - Segment metadata does not exist. segment={}.", traceObjectId, streamSegmentName);
+                        return CompletableFuture.completedFuture(null);
+                    } else if (segmentMetadata.isActive()) {
+                        log.debug("{}: deleteGarbage - Segment is not marked as deleted. segment={}.", traceObjectId, streamSegmentName);
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        return new ChunkIterator(storageExecutor, txn, segmentMetadata)
+                                .forEach((metadata, name) -> {
```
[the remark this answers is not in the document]
sachin-j-joshi: This is being processed as background task and as such throttled inherently by fact that we always process a batch of GC tasks.
This method is just updating metadata and not doing actual LTS calls. 
So this should be fast ( compared to LTS calls)

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: Enums are clumsy and have lots of boiler plate and require me to do exactly same thing - define consts.

I feel subclass is overkill right now.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: Original intention was to treat transient errors differently. However it will be a bigger change so postponing that change and reverting back to using Boolean as before.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/FileSystemIntegrationTest.java
sachin-j-joshi: done

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
sachin-j-joshi: on second thought, finishing storage bootstrap is a separate and pre-requisite step for any other secondary service to work.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: fixed.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollectorTests.java
sachin-j-joshi: oops.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: merged 2 PRs

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: These are two separate steps.
Also this is background processing task. Releasing thread pool thread intermittently like this would be actually benefetial

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -374,22 +501,87 @@ public void report() {
...
     @Data
-    class GarbageChunkInfo implements Delayed {
-        @Getter
+    @RequiredArgsConstructor
+    @Builder(toBuilder = true)
+    @EqualsAndHashCode(callSuper = true)
+    public static class TaskInfo extends AbstractTaskInfo {
+        @NonNull
         private final String name;
         private final long scheduledDeleteTime;
```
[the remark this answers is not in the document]
sachin-j-joshi: good catch. Fixed.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageMockTests.java
sachin-j-joshi: redundant test, covered elsewhere

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: so far this very specific to GC/ChunkedSegmentStorage code.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: We want to delay process deletion logic (currently after 1 minute) that is the whole point of having GC and not do this work inline. Each SLTS operation must be atomic - it can not have partial results (more specifically effect that are non idempotent)

This particular behavior is desired new behavior - to break down segment deletion into multiple steps.

First step will just mark segment inactive and enqueue GC task - this way delete is fast for storage writer and doesn't fail, when this GC task is processed it will clean up metadata and will enqueue further tasks to cleanup etc in background.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -306,7 +311,14 @@ protected void doStart() {
         }
 
         // Delayed start. Secondary services need not be started in order for us to accept requests.
-        delayedStart.thenComposeAsync(v -> startSecondaryServicesAsync(), this.executor)
+        delayedStart
+                .thenComposeAsync( v -> {
```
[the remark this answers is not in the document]
andreipaduroiu: I see these extra spaces everywhere. Are your IDE formatting rules different from the recommended ones?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -191,14 +196,22 @@ public ChunkedSegmentStorage(int containerId, ChunkStorage chunkStorage, ChunkMe
...
      *
      * @param snapshotInfoStore Store that saves {@link SnapshotInfo}.
+     * @param taskQueue  Task queue to use for garbage collection.
      */
-    public CompletableFuture<Void> bootstrap(SnapshotInfoStore snapshotInfoStore) {
+    public CompletableFuture<Void> bootstrap(SnapshotInfoStore snapshotInfoStore, AbstractTaskQueueManager<GarbageCollector.TaskInfo> taskQueue) {
 
         this.logPrefix = String.format("ChunkedSegmentStorage[%d]", containerId);
-
+        this.taskQueue = taskQueue;
```
andreipaduroiu, line 204: You have a possible leak here. `taskQueue` is created in `StreamSegmentContainer` and is passed here via this method. However neither `StreamSegmentContainer` nor this class ever close it.

Normally it is up to whomever creates the object to manage its lifecycle, so I would have argued that the container should have closed it. However in this case, I think it's better if this class does it; if you choose to go this way, please clearly indicate in this argument's Javadoc what's going to happen to it once the `ChunkedSegmentStorage` class is closed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -229,7 +229,8 @@ private MetadataStore createMetadataStore() {
             ChunkedSegmentStorage chunkedStorage = (ChunkedSegmentStorage) this.storage;
             val snapshotInfoStore = getStorageSnapshotInfoStore();
             // Bootstrap
-            return chunkedStorage.bootstrap(snapshotInfoStore);
+            StorageEventProcessor eventProcessor = new StorageEventProcessor(this.metadata.getContainerId(), this.containerEventProcessor, chunkedStorage);
```
andreipaduroiu, line 232: See my comment in `ChunkedSegmentStorage` below regarding this instance.

Furthermore you need to ensure that if `.bootstrap` fails, whether sync or async, that you close `eventProcessor` here; otherwise this is another source of leaks.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,336 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+     * @return A CompletableFuture that, when completed, will contain the result of the operation.
+     * If the operation failed, it will contain the cause of the failure.
+     * */
+    private <R> CompletableFuture<R> executeExclusive(Callable<CompletableFuture<R>> operation, String... keyNames) {
+        return CompletableFuture.completedFuture(null).thenComposeAsync(v -> {
+            Exceptions.checkNotClosed(this.closed.get(), this);
+            try {
+                return operation.call();
+            } catch (CompletionException e) {
+                throw new CompletionException(Exceptions.unwrap(e));
```
andreipaduroiu, line 387: You can merge both of these into one. `Exceptions.unwrap` will return its argument if it's not a `CompletionException`.

andreipaduroiu (commented):
I will do one more pass later and focus on `GarbageCollector`. Looking better now; I left a few more comments.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
sachin-j-joshi: Fixed. Closing it with ChunkedSegmentStorage::close

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
sachin-j-joshi: Fixed. Closing it with ChunkedSegmentStorage::close

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    CompletableFuture<Void> processEvents(List<BufferView> events) {
+        log.debug("{}: processEvents called with {} events", traceObjectId, events.size());
+        ArrayList<GarbageCollector.TaskInfo> batch = new ArrayList<>();
+        for (val event : events) {
+            try {
+                batch.add(serializer.deserialize(event));
+            } catch (IOException e) {
+                log.error("{}: processEvents failed.", traceObjectId, e);
```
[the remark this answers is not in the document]
sachin-j-joshi: Not much to add here. The deserialize failed. No events will be processed as whole batch is failed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public CompletableFuture<Void> addTask(String queueName, GarbageCollector.TaskInfo task) {
+        try {
+            val processor = eventProcessorMap.get(queueName);
+            if (null != processor) {
+                return processor.add(serializer.serialize(task), Duration.ofMillis(1000)).thenAccept( l -> {
+
+                });
+            }
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
sachin-j-joshi: fixed

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param queueName Name of the queue.
+     * @param task      Task to add.
+     */
+    @Override
+    public CompletableFuture<Void> addTask(String queueName, GarbageCollector.TaskInfo task) {
+        try {
+            val processor = eventProcessorMap.get(queueName);
+            if (null != processor) {
+                return processor.add(serializer.serialize(task), Duration.ofMillis(1000)).thenAccept( l -> {
```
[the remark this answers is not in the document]
sachin-j-joshi: using Futures.toVoid

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,336 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        return new ChunkIterator(storageExecutor, txn, segmentMetadata)
+                                .forEach((metadata, name) -> {
+                                    metadata.setActive(false);
+                                    txn.update(metadata);
+                                    chunksToDelete.add(name);
+                                })
+                                .thenComposeAsync(v -> this.addChunksToGarbage(txn.getVersion(), chunksToDelete), storageExecutor)
+                                .thenRunAsync(() -> deleteBlockIndexEntriesForChunk(txn, streamSegmentName, segmentMetadata.getStartOffset(), segmentMetadata.getLength()), storageExecutor)
```
andreipaduroiu, line 286: roll this into the next one (`thenComposeAsync`)

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,336 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                                .thenComposeAsync(v -> {
+                                    txn.delete(segmentMetadata.getName());
+                                    return txn.commit();
+                                }, storageExecutor)
+                                .handleAsync((v, e) -> {
+                                    if (null != e) {
+                                        log.error(String.format("%s deleteGarbage - Could not delete metadata for garbage segment=%s.",
+                                                traceObjectId, streamSegmentName), e);
+                                        failed.set(true);
```
andreipaduroiu, line 295: I believe you can have `handleAsync` return `true/false` and then use that value downstream; no need for a `failed` shared object.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,336 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+
+    public CompletableFuture<Void> processBatch(List<TaskInfo> batch) {
+        ArrayList<CompletableFuture<Void>> futures = new ArrayList<>();
+        for (val infoToDelete : batch) {
+            if (metadataStore.isTransactionActive(infoToDelete.transactionId)) {
+                log.debug("{}: deleteGarbage - transaction is still active - re-queuing {}.", traceObjectId, infoToDelete.transactionId);
+                taskQueue.addTask(taskQueueName, infoToDelete);
+            } else {
+                val f = executeSerialized(() -> processTask(infoToDelete), infoToDelete.name);
+                if (null != f) {
```
andreipaduroiu, line 337: Under what conditions would `f == null`?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
andreipaduroiu: Why do you catch `CompletionException` just to rethrow another `CompletionException` with its inner exception?

This whole try-catch can be rewritten as:

```
 try {
                return operation.call();
} catch(Exception ex){
   throw new CompletionException(ex);
}
```

Everywhere we log or handle exceptions, we invoke `Exceptions.unwrap` - if you look in that method it recursively unwraps `CompletionException` until it finds the real exception inside.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,336 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
-        if (count == 0) {
-            log.debug("{}: deleteGarbage - no work - sleeping for {}.", traceObjectId, config.getGarbageCollectionDelay());
-            return CompletableFuture.completedFuture(false);
+        if (infoToDelete.taskType == TaskInfo.DELETE_SEGMENT) {
+            return  deleteSegment(infoToDelete);
+        }
+        if (infoToDelete.taskType == TaskInfo.DELETE_JOURNAL) {
+            return  CompletableFuture.completedFuture(null);
         }
+        return null;
```
andreipaduroiu, line 405: I would prefer to log a WARN if we get here. Throwing might be too extreme as we may have problems when we do upgrades and add new types of tasks. But we should at least know about them.

sachin-j-joshi: Fixed.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param queueName Name of the queue.
+     * @param task      Task to add.
+     */
+    @Override
+    public CompletableFuture<Void> addTask(String queueName, GarbageCollector.TaskInfo task) {
+        Preconditions.checkNotNull(queueName, "queueName");
+        Preconditions.checkNotNull(task, "task");
+        try {
+            val processor = eventProcessorMap.get(queueName);
+            if (null != processor) {
```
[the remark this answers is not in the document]
sachin-j-joshi: Fixed this one

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,336 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+     * Adds given chunks to list of garbage chunks.
      *
-     * @return The Duration.
+     * @param chunksToDelete List of chunks to delete.
      */
-    @Override
-    protected Duration getShutdownTimeout() {
-        return SHUTDOWN_TIMEOUT;
+    CompletableFuture<Void> addChunksToGarbage(long transactionId, Collection<String> chunksToDelete) {
+        if (null != taskQueue) {
```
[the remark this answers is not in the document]
sachin-j-joshi: This is mainly to aid unit testing without need to initialize everything every time everywhere.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkStorageMetrics.java
```diff
@@ -65,6 +65,19 @@
...
+
+    static final Counter SLTS_GC_CHUNK_NEW = STATS_LOGGER.createCounter(MetricsNames.SLTS_GC_CHUNK_NEW);
+    static final Counter SLTS_GC_CHUNK_QUEUED = STATS_LOGGER.createCounter(MetricsNames.SLTS_GC_CHUNK_QUEUED);
+    static final Counter SLTS_GC_CHUNK_DELETED = STATS_LOGGER.createCounter(MetricsNames.SLTS_GC_CHUNK_DELETED);
+    static final Counter SLTS_GC_CHUNK_RETRY = STATS_LOGGER.createCounter(MetricsNames.SLTS_GC_CHUNK_RETRY);
+    static final Counter SLTS_GC_CHUNK_FAILED = STATS_LOGGER.createCounter(MetricsNames.SLTS_GC_CHUNK_FAILED);
+
+    static final Counter SLTS_GC_SEGMENT_QUEUED = STATS_LOGGER.createCounter(MetricsNames.SLTS_GC_SEGMENT_QUEUED);
```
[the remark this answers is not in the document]
sachin-j-joshi: There are several unit tests that check queue size after each operation.
I think we are covered here.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/StorageTestBase.java
andreipaduroiu: @sachin-j-joshi I think this comment got lost among the other ones.

andreipaduroiu (approved):

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private final static GarbageCollector.TaskInfo.Serializer SERIALIZER = new GarbageCollector.TaskInfo.Serializer();
+
+    private final int containerID;
+    private final ContainerEventProcessor eventProcessor;
+    private final Function<List<GarbageCollector.TaskInfo>, CompletableFuture<Void>> callBack;
+    private final int maxItemsAtOnce;
+    private final String traceObjectId;
+    @Getter
+    private final ConcurrentHashMap<String, ContainerEventProcessor.EventProcessor> eventProcessorMap = new ConcurrentHashMap<>();
```
RaulGracia, line 53: `ContainerEventProcessorImpl` already has this map, for which this is creating a redundant map for the same. You can have here just a `Set<String>` with the names of the processors you need. Then, you could create a new method in the `ContainerEventProcessor` interface called `get(String processorName)` that returns the `CompletableFuture<EventProcessor>` in the case it exists or `null` otherwise. This will help this class and any other class with a similar use case to avoid keeping references to the `EventProcessor`s in maps like this.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Constructor.
+     * @param containerID Container id.
```
RaulGracia, line 57: Nit: keep a blank line between the comment and the subsequent lines with tags (like `@param`). Normally you keep that blank line, but there are cases like this one and others in which there is no blank line.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/AbstractTaskQueueManager.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Manages group of background task queues.
+ *
+ * @param <TaskType> Type of tasks.
+ */
+public interface AbstractTaskQueueManager<TaskType> extends AutoCloseable {
+    /**
+     * Adds a queue by the given name.
+     * @param queueName Name of the queue.
```
RaulGracia, line 29: Same here about blank line after comment.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param queueName Name of the queue.
+     * @param ignoreProcessing Whether the processing should be ignored.
+     */
+    @Override
+    public CompletableFuture<Void> addQueue(String queueName, Boolean ignoreProcessing) {
+        Preconditions.checkNotNull(queueName, "queueName");
+        val config = new ContainerEventProcessor.EventProcessorConfig(maxItemsAtOnce, Long.MAX_VALUE);
+        val f = ignoreProcessing ?
+                eventProcessor.forDurableQueue(queueName) :
```
RaulGracia, line 85: These calls can fail, for instance if the internal segment is not yet ready or any other temporary issue. What is the expected behavior if it does fail? Is this going to retry, it will shut down the container or it will just carry on? Because in the latter case, we may face a situation in which the Segment Store works but the `EventProcessor`s are not initialized, so garbage would not be collected, right? I say that not as a theoretical scenario, but I remember seeing this problem in one of the initial longevity runs in which I induced restarts in purpose (the code may have changed since then, but we need tests to validate this case).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,352 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
-                        .handleAsync((v, ex) -> {
-                            if (null != ex) {
-                                log.error("{}: Error during doRun.", traceObjectId, ex);
-                            }
-                            return null;
-                        }, executor),
-                executor);
-        return loopFuture;
+    CompletableFuture<Void> addChunkToGarbage(long transactionId, String chunkToDelete, long startTime, int attempts) {
+        Preconditions.checkState(null != taskQueue, "taskQueue must not be null.");
```
RaulGracia, line 208: `addChunkToGarbage` and `addSegmentToGarbage` seem to share almost the same logic. It looks simple to create a common method for both, and just use as input the right metric and `TaskInfo` type depending on whether it runs for a chunk or a segment.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,352 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+                    queueSize.incrementAndGet();
+                    SLTS_GC_SEGMENT_QUEUED.inc();
+                }, this.storageExecutor);
     }
 
     /**
-     * Adds given chunks to list of garbage chunks.
-     *
-     * @param chunksToDelete List of chunks to delete.
+     * Adds new chunk to track
```
RaulGracia, line 249: Nit: Needed blank line in many of these javadoc comments.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,352 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+        @Cleanup
+        val txn = metadataStore.beginTransaction(false, streamSegmentName);
+        return txn.get(streamSegmentName)
+                .thenComposeAsync(storageMetadata -> {
+                    val segmentMetadata = (SegmentMetadata) storageMetadata;
+                    if (null == segmentMetadata) {
+                        log.debug("{}: deleteGarbage - Segment metadata does not exist. segment={}.", traceObjectId, streamSegmentName);
+                        return CompletableFuture.completedFuture(null);
+                    } else if (segmentMetadata.isActive()) {
+                        log.debug("{}: deleteGarbage - Segment is not marked as deleted. segment={}.", traceObjectId, streamSegmentName);
```
RaulGracia, line 289: Is that just `debug` that GC has been instructed to delete an active Segment? Looks like a wrong logic that may require us to notice, right? For this reason I don't know if `warn` may be needed here (unless this case is common and harmless for some reason that I ignore).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
```diff
@@ -132,240 +160,352 @@ public GarbageCollector(int containerId, ChunkStorage chunkStorage,
...
+     * Delete block index entries for given chunk.
      */
-    void addToGarbage(String chunkToDelete, long startTime, int attempts) {
-        if (queueSize.get() < config.getGarbageCollectionMaxQueueSize()) {
-            garbageChunks.add(new GarbageChunkInfo(chunkToDelete, startTime, attempts));
-            queueSize.incrementAndGet();
-        } else {
-            log.debug("{}: deleteGarbage - Queue full. Could not delete garbage. chunk {}.", traceObjectId, chunkToDelete);
+    void deleteBlockIndexEntriesForChunk(MetadataTransaction txn, String segmentName, long startOffset, long endOffset) {
+        val firstBlock = startOffset / config.getIndexBlockSize();
```
RaulGracia, line 336: Would it be advisable to add some preconditions on this calculation (e.g., `firstBlock>0`, `config.getIndexBlockSize()>0`)?

RaulGracia (changes requested):
Some minor comments (and one relevant comment about retrying on queue creation).

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: There are subtle differences like extra parameters like attempts, start time etc. These methods are not exactly same.
These methods are small enough and extracting method will probably will be more complicated than necessary so this apparent duplication is okay.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: I think its not needed here - these are checked independently in other places.
Eg. https://github.com/pravega/pravega/blob/master/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/metadata/SegmentMetadata.java#L212

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/GarbageCollector.java
sachin-j-joshi: This is okay.
In the new design we optimistically enqueue GC tasks **before**  SLTS transaction is committed. In case of transaction failure , the segment/chunks will be still marked active and that means transaction failed and GC task need to be fixed.
https://github.com/pravega/pravega/wiki/PDP-53-(Robust-Garbage-Collection-for-SLTS)#proposal

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
sachin-j-joshi: We need this lookup to translate StorageEventProcessor::addTask to processor.add on every call.
Looking up by calling get() will be too much overhead and will eat up time on core thread pool.
Also realistically there are very few entries in that map.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StorageEventProcessor.java
sachin-j-joshi: This method is called during container bootstrap/startup.
As of today we do not retry individual steps in container startup.
In case of failure the container will fail to start and go through re-initialization (similar to when other components fail during startup).

RaulGracia (approved):

andreipaduroiu (changes requested):
Unit tests failing due to new changes.

andreipaduroiu (approved):

PR #6114, opened 2021-06-30
Issue: 6070 Configuration of TLSv1.3 in Pravega Components

### LINKED ISSUE #6070: Implementation of TLS1.3 protocol in Pravega's Controller and Segment Store
Currently Pravega supports TLS 1.2 which was released 10 years back. TLS 1.3 protocol has been released since August 2018. This can be a requirement or feature for security conscious clients. TLS 1.3 protocol provides performance improvements and security enhancements over TLS 1.2.

Enable TLS 1.3 protocol in Pravega components(Controller and Segment Store) to make communications secure.

kotlasaicharanreddy:
Signed-off-by: kotlasaicharanreddy <kotlasaicharan@yahoo.com>

**Change log description**  

- This PR implements `TLSv1.3` support in Pravega components: `Segment store` and `Controller`.
- Segment store and Controller's GRPC endpoint uses `Netty` library to support secure communication.
- Controller's REST endpoint uses `Grizzly` library for TLS communication.
- This PR also includes changes for pravega standalone to support `TLSv1.3` protocol.
- Implements Strict mode (either TLSv1.2 or TLSv1.3) or mixed mode (TLSv1.2 and TLSv1.3) communication.

Prominent files with changes:

- [GRPCServerConfigImpl](https://github.com/pravega/pravega/blob/8c94d005009381fdecd0736327877ac5ca67de41/controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java)
- [RESTServerConfigImpl](https://github.com/pravega/pravega/blob/8c94d005009381fdecd0736327877ac5ca67de41/shared/rest/src/main/java/io/pravega/shared/rest/impl/RESTServerConfigImpl.java)
- [TLSHelper](https://github.com/pravega/pravega/blob/8c94d005009381fdecd0736327877ac5ca67de41/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/security/TLSHelper.java)
- [ServiceConfig](https://github.com/pravega/pravega/blob/8c94d005009381fdecd0736327877ac5ca67de41/segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java)
- [Config](https://github.com/pravega/pravega/blob/8c94d005009381fdecd0736327877ac5ca67de41/controller/src/main/java/io/pravega/controller/util/Config.java)
- [TLSProtocolVersion](https://github.com/pravega/pravega/blob/8c94d005009381fdecd0736327877ac5ca67de41/common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java)

**Purpose of the change**  
Fixes #6070 

**What the code does**  

- A new property `security.tls.protocolVersion` is added to provide TLS property version in Controller and Segment Store properties.
- Valid values are `TLSv1.2` (strict) or `TLSv1.3` (strict) or `TLSv1.2,TLSv1.3` (mixed). 
- The default value is the mixed mode to support the clients which can communicate with server only in TLSv1.2

Segment Store changes

- In [TLSHelper](https://github.com/pravega/pravega/blob/2882fdabaa24227385f54caa3d74e936a29eef70/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/security/TLSHelper.java) class following change is made to facilitate TLS protocol version in segment store server.
```
SslContext result = SslContextBuilder.forServer(certificateFile, serverKeyFile).protocols(tlsProtocolVersion).build();
```
Controller Changes

- Controller changes include both GRPC and REST support for TLSv1.3. 

1. GRPC Changes
    In [GRPCServer](https://github.com/pravega/pravega/blob/2882fdabaa24227385f54caa3d74e936a29eef70/controller/src/main/java/io/pravega/controller/server/rpc/grpc/GRPCServer.java) class `SSLContext `is utilized to support TLS protocol version in Controller's GRPC Server.
```
SslContext ctx = GrpcSslContexts.forServer(new File(serverConfig.getTlsCertFile()), new File(serverConfig.getTlsKeyFile())).protocols(serverConfig.getTlsProtocolVersion()).build();
```

2. REST Changes
    In [RESTServer](https://github.com/pravega/pravega/blob/2882fdabaa24227385f54caa3d74e936a29eef70/shared/rest/src/main/java/io/pravega/shared/rest/RESTServer.java) class `httpServer `is modified to support TLS protocol version in Controller's REST server side.
```
httpServer = GrizzlyHttpServerFactory.createHttpServer(baseUri, resourceConfig, true,
                        new SSLEngineConfigurator(contextConfigurator, false, false, false).setEnabledProtocols(restServerConfig.tlsProtocolVersion()));
```

 

**How to verify it**  

- All tests must pass and build should succeed.
- Run [TlsEnabledInProcPravegaClusterTest](https://github.com/pravega/pravega/blob/master/standalone/src/test/java/io/pravega/local/TlsEnabledInProcPravegaClusterTest.java) and inspecting logs will show the TLS protocol version supported by Pravega.


sarlaccpit (commented):
A couple general comments:

I like the new setting; I think it'll keep things open and flexible.


The amount of files touched in this PR is mostly due to the fact that AbstractConnectionListener was not designed with a builder or config object in mind, so each new field has an enormous ripple effect.

1. Since we are taking on this change now, we might as well introduce some builder and/or config object to incapsulate at least the TLS related parameters.  So that later on let's say, one can pass a list of ciphers to restrict and other settings.

2. Besides the expected side effects in the unit and integration tests, there should be at least 1 new test: that sets up the ctrl/ss in a TLS1.3 only and make sure the client can still talk to it.

sarlaccpit:
I know it's still in draft, but whenever you get the chance, can you update the PR title and the description to full reflect all that's happening (it's not really TLS 1.3 support anymore, but more related to the configuration of protocols allowed etc.  Also it's not just the segment store anymore). Thanks!

standalone/src/test/java/io/pravega/local/PravegaEmulatorResource.java
```diff
@@ -71,6 +71,7 @@ public PravegaEmulatorResource(boolean authEnabled, boolean tlsEnabled, boolean
                 .enableRestServer(restEnabled)
                 .enableAuth(authEnabled)
                 .enableTls(tlsEnabled)
+                .tlsProtocolVersion(TestUtils.getTlsProtocolVersion())
```
shrids, line 74: Please add tests to verify strict mode for standalone. This would be an automated way to verify different TLS protocol versions.

test/integration/src/main/java/io/pravega/test/integration/selftest/adapters/InProcessMockClientAdapter.java
```diff
@@ -79,6 +79,7 @@
     //region Members
 
     private static final String LISTENING_ADDRESS = "localhost";
+    private static final String TLS_PROTOCOL_VERSION = "TLSv1.2,TLSv1.3";
```
andreipaduroiu, line 82: Why is this repeated in every single test? Don't we have a `TestUtils` common class which we can reference here?

test/testcommon/src/main/java/io/pravega/test/common/TestUtils.java
```diff
@@ -37,11 +37,19 @@
     private static final int BASE_PORT = 32768;
     private static final int MAX_PORT_COUNT = 28233;
     private static final Random RAND = new Random();
+    private static final String TLS_PROTOCOL_VERSION = "TLSv1.2,TLSv1.3";
```
andreipaduroiu, line 40: What is the difference between this and `SecurityConfigDefaults`?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -69,13 +69,14 @@
      * @param streamSegmentStore  The SegmentStore to delegate all requests to.
      * @param tableStore          The SegmentStore to delegate all requests to.
      * @param tokenExpiryExecutor The executor to be used for running token expiration handling tasks.
+     * @param tlsProtocolVersion the version of the TLS protocol
      */
     @VisibleForTesting
     public PravegaConnectionListener(boolean enableTls, int port, StreamSegmentStore streamSegmentStore,
-                                     TableStore tableStore, ScheduledExecutorService tokenExpiryExecutor) {
+                                     TableStore tableStore, ScheduledExecutorService tokenExpiryExecutor, String tlsProtocolVersion) {
```
andreipaduroiu, line 76: Is there anything we can do to prevent this constructor from growing even more?

andreipaduroiu (changes requested):

pbelgundi:
@kotlasaicharanreddy When the PR has a lot of changed files, it helps to state the names of files with prominent changes in the PR description and makes it easier to review the PR. Could you please add those file names in the PR description?

controller/src/conf/controller.config.properties
```diff
@@ -49,6 +49,7 @@ controller.security.auth.enable=${AUTHORIZATION_ENABLED}
 controller.security.auth.delegationToken.signingKey.basis=${TOKEN_SIGNING_KEY}
 controller.security.auth.delegationToken.ttl.seconds=${ACCESS_TOKEN_TTL_IN_SECONDS}
 controller.security.tls.enable=${TLS_ENABLED}
+controller.security.tls.protocolVersion=${TLS_PROTOCOL_VERSION}
```
pbelgundi, line 52: Given that we're adding a new property to both segment store and controller that indicates the TLS version to be used by each component, this should require a change on pravega-operator to get this value from users. Please open an issue on pravega-operator for the same.

pbelgundi (changes requested):

pbelgundi:
Please fix failing tests.

controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -125,6 +125,9 @@
     public static final Property<Boolean> PROPERTY_TLS_ENABLED = Property.named(
             "security.tls.enable", false, "auth.tlsEnabled");
 
+    public static final Property<String> PROPERTY_TLS_PROTOCOL_VERSION = Property.named(
+            "security.tls.protocolVersion", "");
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Changed it to mixed mode. Thanks.

test/integration/src/main/java/io/pravega/test/integration/selftest/adapters/InProcessMockClientAdapter.java
kotlasaicharanreddy: Removed this and referred to  `SecurityConfigDefaults` common class in all tests. Thanks.

test/testcommon/src/main/java/io/pravega/test/common/SecurityConfigDefaults.java
```diff
@@ -19,6 +19,8 @@
  * Holds default security configuration values.
  */
 public class SecurityConfigDefaults {
+    public static final String TLS_PROTOCOL_VERSION = "TLSv1.2,TLSv1.3";
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Referred to this in all tests.

controller/src/conf/controller.config.properties
sarlaccpit: @pbelgundi , is it the case?     Aren't these settings fed (blindly) under the PravegaCluster.pravega.options section of the PravegaCluster CRD?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
sarlaccpit: @andreipaduroiu ; agreed, I had mentioned that in point 1 of https://github.com/pravega/pravega/pull/6114#pullrequestreview-697598121

One thing I had suggested to Sai is to make call to either address this additional refactoring in this PR or create a separate issue and handle it there.   It IS kind of a separate issue, BUT we happen to be stepping right on it here by making these changes.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -73,6 +74,7 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
         this.authorizationEnabled = authorizationEnabled;
         this.userPasswordFile = userPasswordFile;
         this.tlsEnabled = tlsEnabled;
+        this.tlsProtocolVersion = tlsProtocolVersion;
```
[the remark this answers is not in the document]
sarlaccpit: I see what you're saying, but empty is implicitly the same as mixed though.   Meaning without setting this property at all, the default behavior from the libraries would be to open to TLSv1.2 and 1.3.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
sarlaccpit: My personal preference would be to create a separate issue for the refactoring of all the settings that are getting passed to the PravegaConnectionListener and leave this PR focused on what it's currently doing.

This will allow us to proceed with what this PR has to give and take the time to discuss the perfect options of the constructor (multiple constructore, factory, builders etc.) as a separate effort.  Thoughts?

controller/src/conf/controller.config.properties
sarlaccpit: @pbelgundi I took the branch for a spin to see if my understanding correct.  It worked as I expected:

I passed these to the PravegaCluster.pravega.options section of the CR spec:
```
controller.security.tls.protocolVersion: TLSv1.3
pravegaservice.security.tls.protocolVersion: TLSv1.3
```

The result was the controllers and segmentstore got these in their respective configmaps:
```
-Dpravegaservice.security.tls.protocolVersion=TLSv1.3
-Dcontroller.security.tls.protocolVersion=TLSv1.3
```

And ultimately responded correctly to my inquieries using openssl s_client -connect, meaning in this case it accepted TLS1.3 and refused 1.2.

So I don't think there are any pravega operator changes required.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
gaddamas: IntelliJ refactor can help in extracting Builder which can be used to construct PravegaConnectionListener that has some mandatory and some optional parameters.

controller/src/conf/controller.config.properties
pbelgundi: Thanks for checking this @sarlaccpit . I wasn't sure if this needed changes on operators.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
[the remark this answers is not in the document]
pbelgundi: It would be good to make it explicit in the code instead of making implicit assumptions. That keeps code more readable and avoids future bugs.

controller/src/test/java/io/pravega/controller/server/ZKBackedControllerServiceStarterTest.java
```diff
@@ -262,6 +262,7 @@ protected ControllerServiceConfig createControllerServiceConfigWithEventProcesso
                                                                                             .port(grpcPort)
                                                                                             .authorizationEnabled(false)
                                                                                             .tlsEnabled(false)
+                                                                                            .tlsProtocolVersion("TLSv1.2,TLSv1.3")
```
shrids, line 265: Use the constant defined in SecurityConfigDefaults  here an all other places.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/GRPCServer.java
```diff
@@ -72,8 +76,14 @@ public GRPCServer(ControllerService controllerService, GRPCServerConfig serverCo
         }
 
         if (serverConfig.isTlsEnabled() && !Strings.isNullOrEmpty(serverConfig.getTlsCertFile())) {
-            builder = builder.useTransportSecurity(new File(serverConfig.getTlsCertFile()),
-                    new File(serverConfig.getTlsKeyFile()));
+            try {
+                builder = builder.useTransportSecurity(new File(serverConfig.getTlsCertFile()),
+                        new File(serverConfig.getTlsKeyFile()));
+                SslContext ctx = GrpcSslContexts.forServer(new File(serverConfig.getTlsCertFile()), new File(serverConfig.getTlsKeyFile())).protocols(serverConfig.getTlsProtocolVersion().split(",")).build();
```
shrids, line 82: Please ensure the logic related to config like getTlsProtocolVersion().split(",") is perfomed in GRPCServerConfig class.

shared/rest/src/main/java/io/pravega/shared/rest/RESTServer.java
```diff
@@ -71,7 +71,7 @@ protected void startUp() {
                 contextConfigurator.setKeyStoreFile(restServerConfig.getKeyFilePath());
                 contextConfigurator.setKeyStorePass(JKSHelper.loadPasswordFrom(restServerConfig.getKeyFilePasswordPath()));
                 httpServer = GrizzlyHttpServerFactory.createHttpServer(baseUri, resourceConfig, true,
-                        new SSLEngineConfigurator(contextConfigurator, false, false, false));
+                        new SSLEngineConfigurator(contextConfigurator, false, false, false).setEnabledProtocols(restServerConfig.tlsProtocolVersion().split(",")));
```
shrids, line 74: Same comment as the GRPCServiceConfig.

controller/src/conf/controller.config.properties
sarlaccpit: Cool; hey btw I had this question too (I'll try it for myself but wondering what you think will happen based on your understanding of the operator):

Say I deploy a PravegaCluster with the options shown above.    I believe those are "opaque" to the operator in the sense that it's not understanding them specifically, it's just passing them along mechanically.

BUT, would the operator react if I patched the PravegaCluster and changed options after the fact?  For example, changing these values from TLSv1.2 to TLSv1.3, on a running cluster let's say.

Even though the operator is not _understanding_ these options, is it _watching_ them (and reacting by respining pods?)?

This is more of a pravega operator question, a tiny bit outside the scope of this PR, but it's related too (meaning if the answer is no, the operator doesn't react, I'll need to create some issue/enhancement)

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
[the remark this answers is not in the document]
sarlaccpit: Yes I agree.

controller/src/main/java/io/pravega/controller/server/Main.java
```diff
@@ -91,6 +91,7 @@ public static void main(String[] args) {
                     .host(Config.REST_SERVER_IP)
                     .port(Config.REST_SERVER_PORT)
                     .tlsEnabled(Config.TLS_ENABLED)
+                    .tlsProtocolVersion(Config.TLS_PROTOCOL_VERSION)
```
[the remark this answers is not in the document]
sarlaccpit: I don't dispute that validating property values is absolutely a great idea; but isn't it a bigger problem than just this property?  Each of the properties could in theory have a bogus/invalid/incompatible value. Is there a facility already where property values are validated?   

If not, while I fully support the sentiment, I'm concerned about extending the scope of the PR.   Just a thought.

controller/src/test/java/io/pravega/controller/server/ZKBackedControllerServiceStarterTest.java
kotlasaicharanreddy: Done. Thanks.

cli/admin/src/test/java/io/pravega/cli/admin/dataRecovery/DataRecoveryTest.java
```diff
@@ -353,7 +355,7 @@ public void close() throws Exception {
             this.streamSegmentStore = this.serviceBuilder.createStreamSegmentService();
             this.tableStore = this.serviceBuilder.createTableStoreService();
             this.server = new PravegaConnectionListener(false, servicePort, this.streamSegmentStore, this.tableStore,
-                    this.serviceBuilder.getLowPriorityExecutor());
+                    this.serviceBuilder.getLowPriorityExecutor(), TLS_PROTOCOL_VERSION);
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Created a new constructor in `PravegaConnectionListener` class and it eliminates the propagation of the changes in most of the test files.

test/testcommon/src/main/java/io/pravega/test/common/TestUtils.java
kotlasaicharanreddy: Initially used this because some of tests already imported `TestUtils`. Referred to `SecurityConfigDefaults` in all tests and removed this.

test/integration/src/test/java/io/pravega/test/integration/WriteBatchTest.java
```diff
@@ -101,7 +102,7 @@ public void setup() throws Exception {
         StreamSegmentStore store = serviceBuilder.createStreamSegmentService();
         TableStore tableStore = serviceBuilder.createTableStoreService();
 
-        this.server = new PravegaConnectionListener(false, servicePort, store, tableStore, serviceBuilder.getLowPriorityExecutor());
+        this.server = new PravegaConnectionListener(false, servicePort, store, tableStore, serviceBuilder.getLowPriorityExecutor(), tlsProtocolVersion);
```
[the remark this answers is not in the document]
kotlasaicharanreddy: No, I've created a new constructor and this change no longer exists.

standalone/src/test/java/io/pravega/local/PravegaEmulatorResource.java
```diff
@@ -86,6 +86,7 @@ public PravegaEmulatorResource(boolean authEnabled, boolean tlsEnabled, boolean
         }
         if (tlsEnabled) {
             emulatorBuilder.certFile(SecurityConfigDefaults.TLS_SERVER_CERT_PATH)
+                    .tlsProtocolVersion(SecurityConfigDefaults.TLS_PROTOCOL_VERSION)
```
sarlaccpit, line 89: One thing we had spoken about is to add a new constructor to PravegaEmulatorResource, such that a non default value can be passed for tlsProtocolVersion.

This would easily allow to extend TlsEnabledInProcPravegaClusterTest to exercise a TLS1.3 only test for example.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -69,13 +69,20 @@
...
         this(enableTls, false, "localhost", port, streamSegmentStore, tableStore,
                 SegmentStatsRecorder.noOp(), TableSegmentStatsRecorder.noOp(), new PassingTokenVerifier(), null,
-                null, true, tokenExpiryExecutor);
+                null, true, tokenExpiryExecutor, tlsProtocolVersion);
+    }
+
+    @VisibleForTesting
+    public PravegaConnectionListener(boolean enableTls, int port, StreamSegmentStore streamSegmentStore,
+                                     TableStore tableStore, ScheduledExecutorService tokenExpiryExecutor) {
+        this(enableTls, port, streamSegmentStore, tableStore, tokenExpiryExecutor, "TLSv1.2,TLSv1.3");
```
sarlaccpit, line 85:
```suggestion
        this(enableTls, port, streamSegmentStore, tableStore, tokenExpiryExecutor, SecurityConfigDefaults.TLS_PROTOCOL_VERSION);
```

standalone/src/test/java/io/pravega/local/TlsProtocolVersion13Test.java
```diff
@@ -0,0 +1,33 @@
+/**
```
sarlaccpit, line 1: Very nice addition to the test suite.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/GRPCServer.java
```diff
@@ -72,8 +76,14 @@ public GRPCServer(ControllerService controllerService, GRPCServerConfig serverCo
...
+                ((NettyServerBuilder) builder).sslContext(ctx);
+            } catch (SSLException ex) {
+                throw new RuntimeException(ex);
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Moved the SSLContext logic to a new method and used `sneakythrows` tag for exceptions.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
kotlasaicharanreddy: Done, removed the hardcoded value.

standalone/src/test/java/io/pravega/local/PravegaEmulatorResource.java
kotlasaicharanreddy: Added tests to verify strict mode for both TLSv1.2 and TLSv1.3 protocol versions.

standalone/src/test/java/io/pravega/local/PravegaEmulatorResource.java
kotlasaicharanreddy: Added tests to verify strict mode for standalone.

controller/src/main/java/io/pravega/controller/server/Main.java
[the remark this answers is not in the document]
andreipaduroiu: We do validate our properties where appropriate. Most of our property values are primitive types (boolean, int, long) so besides number parsing, the only other validation we do is in case some properties must be positive, negative (or greater/smaller than other properties). A new one is also `Duration` where we express time.

We do this in the `*Config` object when we create it. The whole point is that when we pass that `Config` object to the application components, it already contains pre-parsed & validated configuration. There are a few benefits to this, of which the most important ones that come to my mind are:
1. Failing the service early. Do not start the service if config is bad vs detecting the bad config value after the server has started (in which case what do you do? restart? begin throwing and logging errors? ignore?)
2. Keeping parsing and validation logic in one place, which aids with testing and encapsulation. Future changes will mostly require touching one place, not everywhere in the code.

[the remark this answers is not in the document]
sarlaccpit: Nothing to add here. Makes a lot of sense.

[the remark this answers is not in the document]
kotlasaicharanreddy: I've created enum (TLSProtocolVersion) to parse and validate `tlsProtocolVersion`  variable in both Controller and Segment store sides.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
kotlasaicharanreddy: I've created a constructor for `pravegaconnectionlistener` (constructor chaining) which reduces the no of test files.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -73,6 +78,7 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
...
```
andreipaduroiu, line 81: The reason I asked you to use an enum is not just for validation. It's also so that you may convert your config string from whatever format you have in the config file into something that's easy for you to use in the code. I see you only invoke `parse` for validation purposes, but you still carry your full string everywhere and then use `.split(',')` where needed. 

I suggest you either parse the config value here and hold it into an array (which you then pass anyway to the `GrpcSslContexts`, or you create a class that takes in your string, parses it (and validates it in the meantime), and holds it in some form internally. It can then either return it as a string (i.e., in `toString`) or as an array, depending on how you use it.

controller/src/main/java/io/pravega/controller/util/TLSProtocolVersion.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Arrays;
+
+public class TLSProtocolVersion {
+
+    public enum TlsProtocolVersion {
+
+        TLSv1_2("TLSv1.2"),
+        TLSv1_3("TLSv1.3"),
+        TLS1_2ANDTLS1_3("TLSv1.2,TLSv1.3");
```
sarlaccpit, line 26: technically, one could say "TLSv1.3,TLSv1.2" and it would be valid too.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
sarlaccpit: @kotlasaicharanreddy , I think what Andrei is talking about is something like this:

```
import java.util.Arrays;

public class TLSProtocolVersion {
    private String[] tlsProtocols;

    public TLSProtocolVersion(String s) {
        tlsProtocols = TlsProtocolVersion.parse(s);
    }

    public String[] tlsProtocols() {
        return tlsProtocols;
    }

    public enum TlsProtocolVersion {
        TLSv1_2("TLSv1.2"),
        TLSv1_3("TLSv1.3"),
        TLS1_2ANDTLS1_3("TLSv1.2,TLSv1.3"),
        TLS1_3ANDTLS1_2("TLSv1.3,TLSv1.2");

        String protocol;

        TlsProtocolVersion(String tlsprotocol) {
            protocol = tlsprotocol;
        }

        public static String[] parse(String s) {
            if (Arrays.stream(TlsProtocolVersion.values()).anyMatch(e -> e.protocol.equals(s))) {
                return s.split(",");
            }
            return null;
        }
    }
}
```

With that (really just a slightly re-arranged version of your code where all we're doing is only parsing once and holding on the result which can be looked at for validation and used as is when needed.

We validate it and anywhere we need this property, it's available in its parsed -ready-to-use form. (String array)

```
TLSProtocolVersion tlsp = new TLSProtocolVersion(tlsProtocolVersion);  
Preconditions.checkArgument( tlsp.tlsProtocols() != null, "message...");

tlsp.tlsProtocols(); holds the String[] that can be passed directly, already parsed, into the TLSHelpers. etc.
       
```

controller/src/main/java/io/pravega/controller/util/TLSProtocolVersion.java
kotlasaicharanreddy: Added this.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/GRPCServer.java
kotlasaicharanreddy: Moved away the split or regex logic from here.

shared/rest/src/main/java/io/pravega/shared/rest/RESTServer.java
kotlasaicharanreddy: Moved away the split or regex logic from here.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
kotlasaicharanreddy: Modified `TLSProtocolVersion` class as suggested. Thanks.

shared/rest/src/main/java/io/pravega/shared/rest/RESTServer.java
```diff
@@ -73,7 +73,7 @@ protected void startUp() {
                 contextConfigurator.setKeyStorePass(JKSHelper.loadPasswordFrom(restServerConfig.getKeyFilePasswordPath()));
                 httpServer = GrizzlyHttpServerFactory.createHttpServer(baseUri, resourceConfig, true,
                         new SSLEngineConfigurator(contextConfigurator, false, false, false)
-                                .setEnabledProtocols(TLSProtocolVersion.getTlsProtocolVersionList(restServerConfig.tlsProtocolVersion())));
+                                .setEnabledProtocols((new TLSProtocolVersion(restServerConfig.tlsProtocolVersion())).tlsProtocols().toArray(new String[0])));
```
sarlaccpit, line 76: There is a series of problems still.  

1. the fact tlsProtocols() returns a list has now made  the line above veyr strange with additional conversations etc.  Why can't  tlsProcotols() simply return a String[] which can be natively consumed here ?
2. We have 3 copies of TlsProtocolsVersion class.   Can it not be in one shared placed, used by all 3 components that need it?
3. If you look at this line and the others in Segmentstore and grpc,  you are still taking the raw String of tlsprotocols and parsing it here (even though it's buried inside the constructor/function calls).   I am wondering why TLSProtocolsVersion cannot be instantiated (and therefore parsed) once (once per config type; grpc, rest, and segment store), and passed around as is.   Meaning in this function let's say, an object of class TLSProtocolVersion is passed into here, rather than instantiated here based on restServerConfig.tlsProtocolVersion.   Have restServerConfig.tlsProtocol BE of type TLSProtocolsVersion.   Is that possible ?  (same comment in the other 2 configs)

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/security/TLSHelper.java
```diff
@@ -47,7 +46,6 @@
     public static SslContext newServerSslContext(String pathToCertificateFile, String pathToServerKeyFile, String tlsProtocolVersion) {
         Exceptions.checkNotNullOrEmpty(pathToCertificateFile, "pathToCertificateFile");
         Exceptions.checkNotNullOrEmpty(pathToServerKeyFile, "pathToServerKeyFile");
```
sarlaccpit, line 48: We will still need to check the argument.  But the checking can be done by checking `.tlsProtocols() != null` on the already parsed object.

controller/src/conf/controller.config.properties
pbelgundi: > BUT, would the operator react if I patched the PravegaCluster and changed options after the fact? For example, changing these values from TLSv1.2 to TLSv1.3, on a running cluster let's say.
If I recollect it right, these options are passed as java system properties to the Controller and Segment Store server processes. So after these processes start these properties can be changed by the server. However, incase the process dies and has to be restarted, it would start the processes with the old values (passed to operator) as those are the ones the operator is "aware" of. Hope that explains it :)

shared/rest/src/main/java/io/pravega/shared/rest/RESTServer.java
kotlasaicharanreddy: 1. It was modified to `List` to make it immutable because of `spotbug` error, but removed this entire List and now a simple String[] is being returned.
2. Removed duplication and moved it to a shared place in `io/pravega/common/security` which is being used by all 3 components.
3. Used only ENUM as Java supports it without a Class so this avoid instantiation.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -66,13 +68,18 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
                     "accessTokenTtlInSeconds should be -1 (token never expires), 0 (token immediately expires) "
                             + "or a positive integer representing the number of seconds after which the token expires.");
         }
+        if (tlsEnabled && (tlsProtocolVersion != null)) {
```
sarlaccpit, line 71: Having a _class_:
```
public class TLSProtocolVersion {
    private String[] tlsProtocols;

    public TLSProtocolVersion(String s) {
        tlsProtocols = TlsProtocolVersion.parse(s);
    }

    public String[] tlsProtocols() {
        return tlsProtocols;
    }

    public boolean isValid() {
        return tlsProtocols == null;
    }

    public enum TlsProtocolVersion {
        TLSv1_2("TLSv1.2"),
        TLSv1_3("TLSv1.3"),
        TLS1_2ANDTLS1_3("TLSv1.2,TLSv1.3"),
        TLS1_3ANDTLS1_2("TLSv1.3,TLSv1.2");

        String protocol;

        TlsProtocolVersion(String tlsprotocol) {
            protocol = tlsprotocol;
        }

        public static String[] parse(String s) {
            if (Arrays.stream(TlsProtocolVersion.values()).anyMatch(e -> e.protocol.equals(s))) {
                return s.split(",");
            }
            return null;
        }
    }
}
```
Allows an instance to be created and kept around and access to the parsed value.  This way, you can parse and validate exactly once:
```
  if (tlsEnabled && (tlsProtocolVersion != null)) {
            this.tlsProtocolVersion = new TLSProtocolVersion(tlsProtocolVersion);
            Preconditions.checkArgument(tlsProtocolVersion.tlsProtocols() != null,
                    "TLSProtocolVersion should be TLSv1.2 (strict 1.2), TLSv1.3 (strict 1.3), mixed mode");
  }
```

And at this point you have:
`this.tlsProtocolVersion.tlsProtocols()` --> returns `String[]` ready to go into the consumers.

Just taking GRPCServerConfigImpl as an example, the raw String from the config file comes in the constructor, but TLSProtocolVersion instance is what's kept around and referenced to by the lower level api calls that needs the String[]. 

GRPCServerConfigImpl.tlsProtocolVersion is of type `TLSProtocolVersion` essentially, not String.

standalone/src/test/java/io/pravega/local/TlsProtocolVersion13Test.java
kotlasaicharanreddy: Thanks.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/security/TLSHelper.java
kotlasaicharanreddy: Added validation for tlsprotocols

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -57,9 +56,6 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
                                 boolean replyWithStackTraceOnError, boolean requestTracingEnabled) {
 
         Preconditions.checkArgument(port > 0, "Invalid port.");
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Added validation for tlsprotocols

shared/rest/src/main/java/io/pravega/shared/rest/impl/RESTServerConfigImpl.java
```diff
@@ -46,9 +45,6 @@
         Exceptions.checkArgument(port > 0, "port", "Should be positive integer");
         Exceptions.checkArgument(!tlsEnabled || !Strings.isNullOrEmpty(keyFilePath),
                 "TLS", "KeyFilePath should not be empty when TLS is enabled. ");
-        if (tlsEnabled) {
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Added validation for tlsprotocols

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/security/TLSHelper.java
```diff
@@ -37,34 +38,40 @@
...
      * @return a {@link SslContext} built from the specified {@code pathToCertificateFile} and {@code pathToServerKeyFile}
      * @throws NullPointerException if either {@code certificateFile} or {@code serverKeyFile} is null
      * @throws IllegalStateException if either {@code certificateFile} or {@code serverKeyFile} doesn't exist or is unreadable.
      * @throws RuntimeException if there is a failure in building the {@link SslContext}
      */
-    public static SslContext newServerSslContext(File certificateFile, File serverKeyFile) {
+    public static SslContext newServerSslContext(File certificateFile, File serverKeyFile, String tlsProtocolVersion) {
         Preconditions.checkNotNull(certificateFile);
         Preconditions.checkNotNull(serverKeyFile);
+        Preconditions.checkNotNull(tlsProtocolVersion);
```
andreipaduroiu, line 68: Why don't you do the validation here? You parse this string above just to verify it's valid, then below you parse it yet again so you can get the protocols. 

I suggest you get rid of the extra validation above and just do it here - only once. You can have it throw the `IllegalArgEx` here if it can't be parsed.

shared/rest/src/main/java/io/pravega/shared/rest/impl/RESTServerConfigImpl.java
```diff
@@ -34,20 +35,26 @@
...
     @Builder
     RESTServerConfigImpl(final String host, final int port, boolean authorizationEnabled, String userPasswordFile,
-                         boolean tlsEnabled, String keyFilePath, String keyFilePasswordPath) {
+                         boolean tlsEnabled, String tlsProtocolVersion, String keyFilePath, String keyFilePasswordPath) {
         Exceptions.checkNotNullOrEmpty(host, "host");
         Exceptions.checkArgument(port > 0, "port", "Should be positive integer");
         Exceptions.checkArgument(!tlsEnabled || !Strings.isNullOrEmpty(keyFilePath),
                 "TLS", "KeyFilePath should not be empty when TLS is enabled. ");
+        if (tlsEnabled && tlsProtocolVersion != null) {
+            Exceptions.checkArgument(TLSProtocolVersion.parse(tlsProtocolVersion) != null,
```
andreipaduroiu, line 50: Just parse it once and hold onto the parsed class. Do not just parse it just for the sake of validation.

standalone/src/test/java/io/pravega/local/PravegaEmulatorResource.java
```diff
@@ -95,6 +94,32 @@ public PravegaEmulatorResource(boolean authEnabled, boolean tlsEnabled, boolean
         pravega = emulatorBuilder.build();
     }
 
+    /*
+    Implemented Builder pattern in java to make way for optional parameters such as "tlsProtocolVersion"
+     */
+    public static class PravegaEmulatorResourceBuilder {
```
andreipaduroiu, line 100: You can achieve exactly the same thing by decorating the constructor you have just removed above with `@Builder`. Lombok will auto-generate this builder for you. 

Please check out other cases where we used the builder pattern (search for `@Builder` in the codebase). The builder is the one that creates the class, and the class does not have a constructor that takes in a builder!

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.pravega.local;
+
+        import org.junit.ClassRule;
```
andreipaduroiu, line 18: Please fix this indentation.

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
```diff
@@ -0,0 +1,33 @@
...
+        import org.junit.Test;
+
+        import static io.pravega.local.PravegaSanityTests.testWriteAndReadAnEvent;
+
+public class TlsProtocolVersion12Test {
+
+    @ClassRule
+    public static final PravegaEmulatorResource EMULATOR = new PravegaEmulatorResource.PravegaEmulatorResourceBuilder(false, true, false).tlsProtocolVersion("TLSv1.2").build();
```
andreipaduroiu, line 26: Why is this static? It will essentially be shared with all other tests which may prevent you from running tests in parallel. Is this intentional?

common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package io.pravega.common.security;
+
+import java.util.Arrays;
+
+public enum TLSProtocolVersion {
+    TLSv1_2("TLSv1.2"),
+    TLSv1_3("TLSv1.3"),
+    TLS1_2ANDTLS1_3("TLSv1.2,TLSv1.3"),
```
pbelgundi, line 23: As discussed yesterday, please remove values like `TLS1_2ANDTLS1_3("TLSv1.2,TLSv1.3"),` . As we add more and more TLS versions to this enum the number of combinations can grow exponentially. Instead it would be better for the enum to hold just absolute values `TLSv1_2,` `TLSv1_3 etc.... and we could use an enum array to bag the list of accepted protocol versions

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
pbelgundi: > Allows an instance to be created and kept around and access to the parsed value. This way, you can parse and validate exactly once:
I don;t see the need for a class here. I would recommend creating an array of enum instances from the String value passed to the Server early in the code (as early as possible) and storing the same in the Server Config and using this array of enums everywhere else, removing the need for further validation lower down.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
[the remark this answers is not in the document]
pbelgundi: I still don;t see this fixed.

pbelgundi (changes requested):

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
kotlasaicharanreddy: Yes, this is static because Fields and Methods with @ClassRule annotation must be public and static.

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
kotlasaicharanreddy: Fixed.

controller/src/conf/controller.config.properties
sarlaccpit: @pbelgundi yes it explains it.  And what I meant is definitely going through the operator (meaning patching the PravegaCluster CR).   I patched the CR to modify these 2 options and I observed that operator reacted immediately to that, and restarted the SS and CTRL accordingly, with the new settings.

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static io.pravega.local.PravegaSanityTests.testWriteAndReadAnEvent;
+
+public class TlsProtocolVersion12Test {
+
+    @ClassRule
+    public static final PravegaEmulatorResource EMULATOR = PravegaEmulatorResource.builder().tlsEnabled(true).tlsProtocolVersion("TLSv1.3").build();
```
sarlaccpit, line 26: I think this is meant to be tlsv1.2

sarlaccpit (commented):
The whole PR looks good to me a this point. Just one small comment on one test.

controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -319,6 +325,12 @@
         WRITES_TO_RGSTREAMS_WITH_READ_PERMISSIONS = p.getBoolean(PROPERTY_WRITES_TO_RGSTREAMS_WITH_READ_PERMISSIONS);
 
         TLS_ENABLED = p.getBoolean(PROPERTY_TLS_ENABLED);
+        String protocol = p.get(PROPERTY_TLS_PROTOCOL_VERSION);
+        if (Strings.isNullOrEmpty(protocol)) {
```
shrids, line 329: this check is redundant.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -73,6 +77,11 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
         this.authorizationEnabled = authorizationEnabled;
         this.userPasswordFile = userPasswordFile;
         this.tlsEnabled = tlsEnabled;
+        if (tlsProtocolVersion == null) {
+            this.tlsProtocolVersion = Config.TLS_PROTOCOL_VERSION;
```
shrids, line 81: `Config.TLS_PROTOCOL_VERSION `already has a default value. This default value is passed while building the GRPCServiceConfig and hence this is never passed as null right? 
If yes, can you simply add a `Preconditions.checkNull(...)` for the tlsProtocolVersion?

kotlasaicharanreddy: Fixed.Implemented this RESTServerConfigImpl with builder pattern.

controller/src/main/java/io/pravega/controller/util/Config.java
kotlasaicharanreddy: Fixed the redundancy.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
shrids: please remove redundant changes in test file due to this change.

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
kotlasaicharanreddy: Fixed. Changed it to TLSv1.2

common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java
kotlasaicharanreddy: Fixed. Removed unnecessary values.

standalone/src/test/java/io/pravega/local/PravegaEmulatorResource.java
kotlasaicharanreddy: Fixed. Used @Builder annotation for implementing builder pattern.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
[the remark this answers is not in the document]
kotlasaicharanreddy: Fixed. If `tlsprotocolversion` is empty, defaulted it to mixed mode.

shrids (approved):
Apart from the ones listed below I do not have any further comments.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/security/TLSHelper.java
kotlasaicharanreddy: Fixed.Removed the extra validation.

shared/rest/src/main/java/io/pravega/shared/rest/impl/RESTServerConfigImpl.java
kotlasaicharanreddy: Fixed. Parsing is done only in `serviceConfig `and `Config`. Removed the parsing from the components.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
kotlasaicharanreddy: With the new changes, tlsprotocols are parsed early in the code (`Serviceconfig, Config`) classes and passed on to the respective components.

test/integration/src/main/java/io/pravega/test/integration/selftest/adapters/OutOfProcessAdapter.java
```diff
@@ -205,6 +208,7 @@ private Process startController(int controllerId) throws IOException {
                 .sysProp(configProperty(Config.COMPONENT_CODE, Config.PROPERTY_PWD_AUTH_HANDLER_ACCOUNTS_STORE),
                         pathOfConfigItem(SecurityConfigDefaults.AUTH_HANDLER_INPUT_FILE_NAME))
                 .sysProp(configProperty(Config.COMPONENT_CODE, Config.PROPERTY_TLS_ENABLED), this.testConfig.isEnableSecurity())
+                .sysProp(configProperty(Config.COMPONENT_CODE, Config.PROPERTY_TLS_PROTOCOL_VERSION), Arrays.stream(SecurityConfigDefaults.TLS_PROTOCOL_VERSION).map(Object::toString).collect(Collectors.joining(",")))
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Fixed. Used `PROPERTY_TLS_PROTOCOL_VERSION.getDefaultValue()`

test/integration/src/main/java/io/pravega/test/integration/selftest/adapters/OutOfProcessAdapter.java
```diff
@@ -238,6 +242,7 @@ private Process startSegmentStore(int segmentStoreId) throws IOException {
                 .sysProp(ServiceBuilderConfig.CONFIG_FILE_PROPERTY_NAME, getSegmentStoreConfigFilePath())
                 .sysProp(configProperty(ServiceConfig.COMPONENT_CODE, ServiceConfig.ZK_URL), getZkUrl())
                 .sysProp(configProperty(ServiceConfig.COMPONENT_CODE, ServiceConfig.ENABLE_TLS), this.testConfig.isEnableSecurity())
+                .sysProp(configProperty(ServiceConfig.COMPONENT_CODE, ServiceConfig.TLS_PROTOCOL_VERSION), Arrays.stream(SecurityConfigDefaults.TLS_PROTOCOL_VERSION).map(Object::toString).collect(Collectors.joining(",")))
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Fixed.

shared/rest/src/main/java/io/pravega/shared/rest/impl/RESTServerConfigImpl.java
```diff
@@ -63,6 +75,7 @@ public String toString() {
                 .append(String.format("host: %s, ", host))
                 .append(String.format("port: %d, ", port))
                 .append(String.format("tlsEnabled: %b, ", tlsEnabled))
+                .append(String.format("tlsProtocolVersion: %s, ", String.format("tlsProtocolVersion: %s, ", Arrays.stream(tlsProtocolVersion).map(Object::toString).collect(Collectors.joining(",")))))
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Fixed.

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
```diff
@@ -296,6 +298,7 @@ private void startLocalSegmentStore(int segmentStoreId) throws Exception {
                         .with(ServiceConfig.LISTENING_PORT, this.segmentStorePorts[segmentStoreId])
                         .with(ServiceConfig.CLUSTER_NAME, this.clusterName)
                         .with(ServiceConfig.ENABLE_TLS, this.enableTls)
+                        .with(ServiceConfig.TLS_PROTOCOL_VERSION, Arrays.stream(this.tlsProtocolVersion).map(Object::toString).collect(Collectors.joining(",")))
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Fixed. Removed mapping string to an object.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -69,13 +70,20 @@
...
-                                     TableStore tableStore, ScheduledExecutorService tokenExpiryExecutor) {
+                                     TableStore tableStore, ScheduledExecutorService tokenExpiryExecutor, String[] tlsProtocolVersion) {
         this(enableTls, false, "localhost", port, streamSegmentStore, tableStore,
                 SegmentStatsRecorder.noOp(), TableSegmentStatsRecorder.noOp(), new PassingTokenVerifier(), null,
-                null, true, tokenExpiryExecutor);
+                null, true, tokenExpiryExecutor, tlsProtocolVersion);
+    }
+
+    @VisibleForTesting
+    public PravegaConnectionListener(boolean enableTls, int port, StreamSegmentStore streamSegmentStore,
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Fixed. Added javadoc.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
```diff
@@ -73,6 +77,11 @@ public GRPCServerConfigImpl(final int port, final String publishedRPCHost, final
         this.authorizationEnabled = authorizationEnabled;
         this.userPasswordFile = userPasswordFile;
         this.tlsEnabled = tlsEnabled;
+        if (tlsProtocolVersion == null) {
+            this.tlsProtocolVersion = Config.TLS_PROTOCOL_VERSION.toArray(new String[Config.TLS_PROTOCOL_VERSION.size()]);
```
pbelgundi, line 81: Like I mentioned earlier, it may be better to just store an immutable array of String and use that instead of creating copies of this array everytime.

pbelgundi (changes requested):

common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.pravega.common.security;
+
+import java.util.Arrays;
+
+public class TLSProtocolVersion {
```
pbelgundi, line 20: I'd would have preferred this to be an Enum instead of a class that just encapsulates a String array. That keeps code more readable and maintanable.

pbelgundi:
Looks fine to me and I can approve once a few minor comments I have provided are addressed and we have all tests passing

common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private String[] protocols;
+
+    public TLSProtocolVersion(String s) {
+        protocols = parse(s);
+    }
+
+    public String[] getProtocols() {
+        return Arrays.copyOf(protocols, protocols.length);
```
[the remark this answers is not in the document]
andreipaduroiu: Actually you would. Even if it's final, it's theoretically possible for whomever invokes `.getProtocols()` to modify the contents of your array. 

I suppose the fact that it's an array is required by various other APIs not within our control that expect an array. 

If we could get away with a `List`, then we could have `Collections.unmodifiableList(Arrays.asList(protocols))` which returns an immutable wrapper.

controller/src/main/java/io/pravega/controller/server/Main.java
```diff
@@ -91,6 +91,7 @@ public static void main(String[] args) {
                     .host(Config.REST_SERVER_IP)
                     .port(Config.REST_SERVER_PORT)
                     .tlsEnabled(Config.TLS_ENABLED)
+                    .tlsProtocolVersion(Config.TLS_PROTOCOL_VERSION.toArray(new String[Config.TLS_PROTOCOL_VERSION.size()]))
```
andreipaduroiu, line 94: This is an example where you convert a List to an array. We could have just passed in the List itself here and hold it as a list inside the class.

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
andreipaduroiu: Why don't you use your `TlsProtocolVersion` class here?

controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -291,6 +298,7 @@
     public static final int HEALTH_CHECK_FREQUENCY;
 
     private static final String METRICS_PATH = "controller.metrics.";
+    private static final Logger log = org.slf4j.LoggerFactory.getLogger(Config.class);
```
andreipaduroiu, line 301: Where is this being used?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -73,17 +74,19 @@
...
     public AbstractConnectionListener(boolean enableTls, boolean enableTlsReload, String host, int port,
-                                      String certFile, String keyFile) {
+                                      String certFile, String keyFile, String[] tlsProtocolVersion) {
         this.enableTls = enableTls;
         this.enableTlsReload = this.enableTls && enableTlsReload;
         this.host = Exceptions.checkNotNullOrEmpty(host, "host");
         this.port = port;
         this.pathToTlsCertFile = certFile;
         this.pathToTlsKeyFile = keyFile;
+        this.tlsProtocolVersion = Arrays.copyOf(tlsProtocolVersion, tlsProtocolVersion.length);
```
andreipaduroiu, line 89: Here too.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
```diff
@@ -272,6 +275,12 @@
     @Getter
     private final boolean enableTls;
 
+    /**
+     * Tls Protocol Version
+     */
+    @Getter
+    private final String[] tlsProtocolVersion;
```
andreipaduroiu, line 282: Why not just keep a `TlsProtocolVersion` instance here?

segmentstore/server/src/test/java/io/pravega/segmentstore/server/store/ServiceBuilderConfigTests.java
```diff
@@ -131,8 +131,13 @@ public void testGetConfig() throws Exception {
                     if (p.getDefaultValue() != null && p.getDefaultValue() instanceof Boolean) {
                         configBuilder.with(p, nextValue.incrementAndGet() % 2 == 0);
                     } else {
-                        // Any number can be interpreted as a string or number.
-                        configBuilder.with(p, Integer.toString(nextValue.incrementAndGet()));
+                        //Property security.tls.protocolVersion cannot be an Integer.
+                        if (p.getName().equals("security.tls.protocolVersion")) {
```
andreipaduroiu, line 135: Do not hardcode property values in the code. If we ever change this, this code will break. Please reference `ServiceConfig.TLS_PROTOCOL_VERSION`. You can just use `p == ServiceConfig.TLS_PROTOCOL_VERSION` since there's only once instance of this property.

shared/rest/src/main/java/io/pravega/shared/rest/impl/RESTServerConfigImpl.java
```diff
@@ -22,24 +22,34 @@
...
     private final boolean authorizationEnabled;
     private final String userPasswordFile;
     private final boolean tlsEnabled;
+    private final String[] tlsProtocolVersion;
     private final String keyFilePath;
     private final String keyFilePasswordPath;
 
-    @Builder
+    public static final class RESTServerConfigImplBuilder {
+        private String[] tlsProtocolVersion = new String[] {"TLSv1.2", "TLSv1.3"};
```
andreipaduroiu, line 44: 1. `final`
2. Why is this hardcoded?
3. Why not use `TlsProtocolVersion` class here?

standalone/src/main/java/io/pravega/local/SingleNodeConfig.java
```diff
@@ -198,6 +206,7 @@ private SingleNodeConfig(TypedProperties properties) {
         this.passwd = properties.get(PASSWD);
         this.enableRestServer = properties.getBoolean(ENABLE_REST_SERVER);
         this.enableTls = properties.getBoolean(ENABLE_TLS);
+        this.tlsProtocolVersion = new TLSProtocolVersion(properties.get(TLS_PROTOCOL_VERSION)).getProtocols();
```
andreipaduroiu, line 209: Just keep the `TlsProtocolVersion` instance here.

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static io.pravega.local.PravegaSanityTests.testWriteAndReadAnEvent;
+
+public class TlsProtocolVersion12Test {
+
+    @ClassRule
+    public static final PravegaEmulatorResource EMULATOR = PravegaEmulatorResource.builder().tlsEnabled(true).tlsProtocolVersion(new String[] {"TLSv1.2"}).build();
+
+    @Test(timeout = 30000)
+    public void testTlsProtocolVersiontls1_2() throws Exception {
```
andreipaduroiu, line 29: Do we actually need this test? How long does it take to run? If it takes 10+ seconds, is there a way we can live without having these tests? (you essentially do it twice, here and below).

test/testcommon/src/main/java/io/pravega/test/common/SecurityConfigDefaults.java
```diff
@@ -19,6 +19,8 @@
  * Holds default security configuration values.
  */
 public class SecurityConfigDefaults {
+    public static final String[] TLS_PROTOCOL_VERSION = "TLSv1.2,TLSv1.3".split(",");
```
andreipaduroiu, line 22: new String[]{ "TLSv1.2", "TLSv1.3"};

andreipaduroiu (changes requested):
Looking better. Please address the remaining comments from me and @pbelgundi and we should be good to go afterwards.

common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java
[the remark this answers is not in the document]
pbelgundi: Looks like Spotbugs also has a problem with storing stuff an array (when stored in a static reference) So we could store and use an immutable list everywhere...I'm fine with that too....

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
kotlasaicharanreddy: These were added to verify clients communication with Pravega in tls1.2 and tls1.3

controller/src/main/java/io/pravega/controller/server/rpc/grpc/impl/GRPCServerConfigImpl.java
kotlasaicharanreddy: Reason not to use TlsProtocolVersion here is to avoid calling `parse` method multiple times. `Tlsprotocolversion` is used only  in Serviceconfig and config classes and from these classes parsed value is passed to `Segmentstore and Controller's(GRPC,REST)` components.

common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.pravega.common.security;
+
+import java.util.Arrays;
+
+public class TLSProtocolVersion {
+
+    private String[] protocols;
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Fixed.

controller/src/main/java/io/pravega/controller/util/Config.java
```diff
@@ -319,6 +327,8 @@
         WRITES_TO_RGSTREAMS_WITH_READ_PERMISSIONS = p.getBoolean(PROPERTY_WRITES_TO_RGSTREAMS_WITH_READ_PERMISSIONS);
 
         TLS_ENABLED = p.getBoolean(PROPERTY_TLS_ENABLED);
+        String[] protocols = new TLSProtocolVersion(p.get(PROPERTY_TLS_PROTOCOL_VERSION)).getProtocols();
+        TLS_PROTOCOL_VERSION = Collections.unmodifiableList(Arrays.asList(protocols));
```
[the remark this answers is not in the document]
kotlasaicharanreddy: Variables in this class are initiated in `Static` block. This makes the String array mutable. So used List in config file.

controller/src/main/java/io/pravega/controller/util/Config.java
kotlasaicharanreddy: Fixed.

common/src/main/java/io/pravega/common/security/TLSProtocolVersion.java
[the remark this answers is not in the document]
kotlasaicharanreddy: We need copyOf() to make it immutable. String[] works as expected in serviceConfig class but not in Config class.
 Variables in Config are initialized in a static block which makes them immutatble (spotbugs error).

sarlaccpit (approved):

pbelgundi (approved):
Looks fine

andreipaduroiu (approved):

andreipaduroiu (changes requested):
@kotlasaicharanreddy There are still some unaddressed comments.

controller/src/main/java/io/pravega/controller/server/Main.java
kotlasaicharanreddy: Passing List to end components  again requires converting it to String[], for example Grizzly's library used in RESTServer takes only String[] as parameter.

shared/rest/src/main/java/io/pravega/shared/rest/impl/RESTServerConfigImpl.java
kotlasaicharanreddy: 1. If it's final makes the @builder annotation fail as a value cannot be assigned to a final variable.
2.  Because import Config.TLS_PROTOCOL_VERSION causes a dependency of Controller in Shared package.
3. TlsProtocolVersion is just a utility class to parse and validate the values.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/ServiceConfig.java
kotlasaicharanreddy: TlsProtocolVersion is just a utility class created for parsing and validating the string passed to it.

standalone/src/test/java/io/pravega/local/TlsProtocolVersion12Test.java
andreipaduroiu: You still haven't answered my question. These tests are not trivial and usually take a while to run. How long do they take? We also need to factor in whether our build time is going to grow as a result of this change.

kotlasaicharanreddy: Ran the test multiple times in isolation, it took ~13-15 seconds for completion. It should take less time when it's run in parallel with other tests.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/store/ServiceBuilderConfigTests.java
kotlasaicharanreddy: Fixed.

andreipaduroiu (approved):

PR #6123, opened 2021-07-02
Issue 6099: Table Segment Key Count

### LINKED ISSUE #6099: Table Segment Entry Count API
We need an API (exposed via the Wire Protocol) that returns the current number of entries in a Table Segment. The number of entries may be approximate if there are ongoing updates to the table, but should be consistent if those updates are all conditional updates.

andreipaduroiu:
**Change log description**  
Created a new API to retrieve information about a Table Segment:
- Length and Start Offset
- Key Length
- Entry Count

**Purpose of the change**  
 Fixes #6099.

**What the code does**  
- Created new Wire Commands `GetTableSegmentInfo` and `TableSegmentInfo` to request and get info about a Table Segment.
- Wired these up from `TableSegmentImpl` (client-side) via the `PravegaRequestProcessor` and into the `TableStore` endpoint of the Segment Store.
- Implemented this inside the Segment Store as an eventually consistent property (value will converge to accurate result as elements are indexed in the background). 
- Removed never-used Wire Commands and support for merging and sealing Table Segments. This functionality has not been implemented in the past 3 years and is not on our roadmap.

**How to verify it**  
Several unit tests have been added and some have been updated.


segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/BadSegmentTypeException.java
```diff
@@ -19,21 +19,21 @@
  * Exception that is thrown whenever a segment of the wrong type is accessed (i.e., we want a StreamSegment but were given
  * the name of a Table Segment).
  */
-public class BadSegmentTypeException extends StreamSegmentException {
+public class BadSegmentTypeException extends IllegalArgumentException {
```
RaulGracia, line 22: What are the side effects of this to other components?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/stat/TableSegmentStatsRecorder.java
```diff
@@ -122,6 +130,11 @@ public void iterateKeys(String tableSegmentName, int resultCount, Duration elaps
             public void iterateEntries(String tableSegmentName, int resultCount, Duration elapsed) {
             }
 
+            @Override
+            public void getInfo(String tableSegmentName, Duration elapsed) {
+
```
RaulGracia, line 135: Nit: extra space.

RaulGracia (approved):
LGTM, just a couple of minor comments.

RaulGracia:
Although apparently not related, this test has failed in this build (as well in others):
```
io.pravega.segmentstore.server.containers.StreamSegmentContainerTests > testTransactionOperations FAILED
    java.lang.AssertionError: getContent() did not return a completed future for segmentSegment_73
        at io.pravega.segmentstore.server.containers.StreamSegmentContainerTests.checkReadIndex(StreamSegmentContainerTests.java:2392)
        at io.pravega.segmentstore.server.containers.StreamSegmentContainerTests.checkReadIndex(StreamSegmentContainerTests.java:2368)
        at io.pravega.segmentstore.server.containers.StreamSegmentContainerTests.checkReadIndex(StreamSegmentContainerTests.java:2363)
        at io.pravega.segmentstore.server.containers.StreamSegmentContainerTests.testTransactionOperations(StreamSegmentContainerTests.java:1333)
````
Looks like this test is becoming increasingly unreliable.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/BadSegmentTypeException.java
andreipaduroiu: This exception has been there for 3 years but never used anywhere. I have made it extend `IllegalArgumentException` and replaced a single place in the code throwing `IllegalArgumentException` with it. So any existing code will still work as it did, but now we have a specific exception to tell us when we try to access the wrong segment type.

PR #6125, opened 2021-07-05
Issue 6061: Reduce update time for recording of commit offsets post transaction segment merge.

### LINKED ISSUE #6061: Simplify processing for recording of commit offsets, post transaction segment merge during transaction Commit processing
Commit Request processing takes the maximum amount of time in transaction processing. Hence optimizing some aspects of this would help improve transaction performance under load.

Optimizations recommended:
recordCommitOffset() consolidation - During Commit Event processing we call `recordCommitOffsets`  for every transaction and update the ActiveTransactionRecord for the said transaction with the commit offsets. This leads to 1 call per transaction to Segment Store. This can instead be done as a bulk update operation reducing the number of update calls to 1 call per commit event instead of 1 per transaction.

pbelgundi:

**Change log description**  
During transaction Commit Event processing we recordCommitOffsets for every transaction, after all transaction segments have been merged. This causes one network call to be made to Segment Store per transaction, for updating the ActiveTxnRecord. This can instead be done as a bulk update operation after segments for all transactions have been merged.

**Purpose of the change**  
Fixes #6061

**What the code does**  
Instead of adding transaction offsets, to ActiveTxnRecord and then computing writer marks from here, we directly provide offsets along with timestamp and writerID for each writer to the generateMarksForTransactions() method thus avoiding the need to update ActiveTxnRecord for each transaction.

**How to verify it**  


controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -387,4 +407,15 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
         return streamMetadataStore.getState(event.getScope(), event.getStream(), true, null, executor)
                                   .thenApply(state -> state.equals(State.COMMITTING_TXN) || state.equals(State.SEALING));
     }
+
+    /* This is a data class that represents a writer mark. Writers send mark information
+     * for watermarking purposes, containing time and position.
+     */
+    @Data
+    @AllArgsConstructor
+    public static class TxnWriterMark {
```
andreipaduroiu, line 416: FYI this class is not thread safe. You should mark it as such (`@NotThreadSafe`) or make it thread safe if you pass it between threads or work concurrently on it. This is even more important since you marked it as `public` which means you intend on giving it to outside components.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1964,26 +1940,32 @@ private TxnStatus handleDataNotFoundException(Throwable ex) {
...
-                                        return CompletableFuture.completedFuture(versioned);
+                                        List<VersionedTransactionData> emptyTransactionData = new LinkedList<>();
+                                        return CompletableFuture.completedFuture(new SimpleEntry<>(versioned, emptyTransactionData));
                                     } else {
-                                        Map.Entry<UUID, ActiveTxnRecord> firstEntry = list.get(0);
-                                        ImmutableList.Builder<UUID> txIdList = ImmutableList.builder(); 
-                                        list.forEach(x -> txIdList.add(x.getKey()));
-                                        List<Long> positions = list.stream().map(x -> x.getValue().getCommitOrder())
+                                        ImmutableList.Builder<UUID> txIdList = ImmutableList.builder();
+                                        list.forEach(x -> {
```
andreipaduroiu, line 1947: Isn't there an API on ImmutableList to either wrap or make a copy of `list`? 

Actually where do you use this anyway?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
shiveshr: this should be a data class, immutable.. all the fields in this class should be final.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
```diff
@@ -1732,49 +1732,25 @@ private long getNewEpochMergeCount(final long mergeCountPrevEpoch, ImmutableList
...
-     * 
-     * @param committingTransactionsRecord Committing transaction record
+     *
      * @return A completableFuture, which when completed will have marks reported for all transactions in the committing 
      * transaction record for which a writer with time and position information is available. 
      */
-    CompletableFuture<Void> generateMarksForTransactions(CommittingTransactionsRecord committingTransactionsRecord,
-                                                         OperationContext context) {
+    CompletableFuture<Void> generateMarksForTransactions(OperationContext context,
+                                                         Map<String, CommitRequestHandler.TxnWriterMark> writerMarks) {
```
shiveshr, line 1740: should we move txn writer mark class out of commit request handler, given that it is to be used in the store's api. 
it should be in the store and used by commit request handler

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java
shiveshr: we could have done list.stream.map.collect and then made an immutable copy. that would have used 2x the memory.

shiveshr (changes requested):
LGTM, few minor comments that once addressed

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
pbelgundi: fixed. Made the class immuatble and moved to AbstractStreamMetadataStore.

controller/src/main/java/io/pravega/controller/store/stream/PersistentStreamBase.java

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
pbelgundi: > FYI this class is not thread safe. You should mark it as such (`@NotThreadSafe`) or make it thread safe if you pass it between threads or work concurrently on it. This is even more important since you marked it as `public` which means you intend on giving it to outside components.

I needed to mark it as public as it is used outside its parent package by CommitRequestHandler.

controller/src/main/java/io/pravega/controller/store/stream/AbstractStreamMetadataStore.java
```diff
@@ -1297,6 +1292,17 @@ String getScopedStreamName(String scope, String stream) {
     // region reader group
     abstract ReaderGroup newReaderGroup(final String scope, final String name);
 
+    /* This is a data class that represents a writer mark. Writers send mark information
+     * for watermarking purposes, containing time and position.
+     */
+    @Data
+    @AllArgsConstructor
+    public static class TxnWriterMark {
```
shiveshr, line 1300: we should make this a top level class instead of a nested class. 
its used in the Store interface as well. So keeping this in the implementation  of store class does not look good.

pbelgundi: This is just a "helper" class used for passing Transactional WriterMarks & used only by the CommitRequestHandler to pass data to the Store, as such I think it should be an inner class. We do have another class `io.pravega.controller.store.stream.records.WriterMark` to represent the actual WriterMark.

shiveshr: but this is used in the store interface in one of the methods, right? if so, this should not be defined in the abstract implementation of the store interface where this is used.

pbelgundi: Made TxnWriterMark a top level class, as suggested.

shiveshr (approved):

PR #6127, opened 2021-07-06
Issue 6128: Ensure RevisionedStreamClient logs the fact that no new updates are present to read post a revision.

### LINKED ISSUE #6128: Improve RevisionedStreamClient logs.
**Describe the bug**
Ensure we log the following:

- RevisionedStreamClient should log the fact that no new updates are present to read post a revision. 
- On a Segment truncation the offset of the truncation should be logged.
- If `getStreamInfo` returns a stale value we should log the fact in EventStreamReaderImpl.

This will simplify debugging scenarios where the reader never made progress.

**To Reproduce**
NA

**Screenshots**
NA

**Additional information**
This will simplify debugging scenarios where the reader never made progress.

shrids:
**Change log description**  
Ensure RevisionedStreamClient logs the fact that no new updates are present to read post a revision.

**Purpose of the change**  
Fixes #6128 

**What the code does**  
Ensure RevisionedStreamClient logs the fact that no new updates are present to read post a revision. Improve logging for scnearios where the client observes Segment truncation. These logs will simplify debugging of scenarios where readers do not progress due to segment truncation.

**How to verify it**  
No Changes in functionality. All the existing tests should continue to pass.

### DIFF
--- a/client/src/main/java/io/pravega/client/state/impl/RevisionedStreamClientImpl.java
+++ b/client/src/main/java/io/pravega/client/state/impl/RevisionedStreamClientImpl.java
@@ -133,17 +133,17 @@ public class RevisionedStreamClientImpl<T> implements RevisionedStreamClient<T> 
     }
 
     @Override
-    public Iterator<Entry<Revision, T>> readFrom(Revision start) {
-        log.trace("Read segment {} from revision {}", segment, start);
+    public Iterator<Entry<Revision, T>> readFrom(Revision revision) {
+        log.trace("Read segment {} from revision {}", segment, revision);
         synchronized (lock) {
-            long startOffset = start.asImpl().getOffsetInSegment();
+            long startOffset = revision.asImpl().getOffsetInSegment();
             SegmentInfo segmentInfo = Futures.getThrowingException(meta.getSegmentInfo());
             long endOffset = segmentInfo.getWriteOffset();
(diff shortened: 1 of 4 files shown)


client/src/main/java/io/pravega/client/state/impl/RevisionedStreamClientImpl.java
```diff
@@ -142,6 +142,9 @@ public void writeUnconditionally(T value) {
             if (startOffset < segmentInfo.getStartingOffset()) {
                 throw new TruncatedDataException(format("Data at the supplied revision {%s} has been truncated. The current segment info is {%s}", start, segmentInfo));
             }
+            if (startOffset == endOffset) {
+                log.info("No new updates to be read from revision {}", start);
```
andreipaduroiu, line 146: Please include more identifying information (segment name, offsets, etc - anything that may be useful in debugging).

Also, do these need to be INFOs - how often are we going to log them?

tkaitchuck (approved):

client/src/main/java/io/pravega/client/state/impl/RevisionedStreamClientImpl.java
shrids: The start object of type Revision logs all the details like segment name and offsets. I have changed the log level to DEBUG since INFO is quite noisy is a normal run.

PR #6134, opened 2021-07-07
Issue 6145: Custom thread pool

tkaitchuck:
**Changelog**
Provide a custom implementation of ScheduledThreadPoolExecutor.

**Purpose of the change**  
Fixes: #6145

**What the code does**  
This provides a custom impementation of the same interfaces as `ScheduledThreadPoolExecutor` in the standard library. However instead of using a heap data structure under lock to store the queued messages, it uses a lock-free queue. The queue works by splitting the delayed and non-delayed tasks. The non-delayed tasks are in a dequeue so adding and removing them can be done in O(1) without holding a lock. The delayed tasks still take O(log(n)) but do so with a lock-free data structure so that we can have parallelism when adding and removing them. 

**ATTENTION REVIEWERS**
The only major downside to this approach is that invoking `cancel` on a future to prevent a not-yet-run task from running takes O(n) in this implementation rather than the O(log(n)) it takes in `ScheduledThreadPoolExecutor`. 
There is one additional difference; When cancel is called, the boolean `shouldInterrupt` flag is ignored and assumed to be false.
`cancel` is not commonly used, but as this will impact all of the code. Someone from every area to make sure this is OK for all areas of the code they are aware of.


andreipaduroiu:
@tkaitchuck Let's get this PR ready for review and merging to master. Let's create an issue, format the PR description and title and make it build.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceFactory.java
```diff
@@ -155,17 +149,18 @@ ScheduledExecutorService newScheduledThreadPool(int size, String poolName, int t
...
-
-        // Do not execute any delayed tasks after shutdown.
-        result.setExecuteExistingDelayedTasksAfterShutdownPolicy(false);
-
-        // Remove tasks from the executor once they are done executing. By default, even when canceled, these tasks are
-        // not removed; if this setting is not enabled we could end up with leaked (and obsolete) tasks.
-        result.setRemoveOnCancelPolicy(true);
+        ThreadPoolScheduledExecutorService result = this.createScheduledExecutor.apply(size, threadFactory);
+
+        //These aren't needed anymore as ThreadPoolScheduledExecutorService mandates all of them
```
andreipaduroiu, line 154: Let's remove the commented out code, but do leave a comment explaining that these options are mandated by default - that way, if we end up moving away from this executor, we know what we need to bring back without having surprises.

common/src/main/java/io/pravega/common/concurrent/Scheduled.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.pravega.common.concurrent;
+
+public interface Scheduled {
```
andreipaduroiu, line 19: Javadoc explaining what this is and what it does.

common/src/main/java/io/pravega/common/concurrent/ThreadPoolScheduledExecutorService.java
```diff
@@ -0,0 +1,376 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+@Slf4j
+@ToString(of = "runner")
+public class ThreadPoolScheduledExecutorService extends AbstractExecutorService implements ScheduledExecutorService  {
```
andreipaduroiu, line 48: Let's add some Javadoc explaining what this executor does and how it may differ from the JDK implementation. We should also include those options that are forced on by default (the ones you commented out in the factory)

common/src/main/java/io/pravega/common/concurrent/ThreadPoolScheduledExecutorService.java
```diff
@@ -0,0 +1,376 @@
...
+
+    private static final AtomicLong COUNTER = new AtomicLong(0);
+    private final ThreadPoolExecutor runner;
+    private final ScheduledQueue<ScheduledRunnable<?>> queue;
+
+    public ThreadPoolScheduledExecutorService(int corePoolSize, int maxPoolSize, long keepAliveMillis,
+            ThreadFactory threadFactory, RejectedExecutionHandler handler) {
+        this.queue = new ScheduledQueue<ScheduledRunnable<?>>();
+        // While this cast looks sketchy as hell, its ok because runner is private and will only
```
andreipaduroiu, line 57: Let's soften the language a bit please.

common/src/main/java/io/pravega/common/concurrent/ThreadPoolScheduledExecutorService.java
```diff
@@ -0,0 +1,376 @@
...
+            this.isDelayed = false;
+            this.scheduledTimeNanos = 0;
+            this.task = task;
+            this.future = new CompletableFuture<R>();
+        }
+
+        private ScheduledRunnable(Callable<R> task, long delay, TimeUnit unit) {
+            this.id = COUNTER.incrementAndGet();
+            this.isDelayed = true;
+            this.scheduledTimeNanos = unit.toNanos(delay) + System.nanoTime();
```
andreipaduroiu, line 151: I'm thinking that in the interest of testing, you might want to have a way to pass a time supplier (and default it to `System.nanoTime`). Since this will be core to all of our components, I want this properly tested and not rely on the actual passage of time to verify it (which will likely slow down the tests).

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,345 @@
...
+    private final Semaphore blocker = new Semaphore(1);
+    
+    @Data
+    private static final class FireTime implements Comparable<FireTime> {
+        private final long timeNanos;
+        private final long sequenceNumber;
+
+        @Override
+        public int compareTo(FireTime other) {
+            if (this.timeNanos < other.timeNanos)
```
andreipaduroiu, line 55: Please use proper formatting rules. Enclose all code blocks with braces. How did Checkstyle not catch this?

andreipaduroiu (changes requested):
Looks good. I will have another pass soon to see if I missed anything.

In the meantime, it would be very nice if we could add some metrics to this one (since now, we can!).

There already is a class in the Segment Store called `SegmentStoreMetrics/ThreadPool`. Can we verify that that class still works with this new type of executor?

Additionally, it would be great if we could expose other metrics, such as:
- Avg time in queue
- Avg procesisng time

We should already be able to extract size of the queue and number of threads - see the metric class I mentioned above.

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,345 @@
...
+     */
+    @Override
+    public E take() throws InterruptedException {
+        return poll(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
+    }
+
+
+    @Override
+    public E poll(long timeout, TimeUnit unit) throws InterruptedException {
+        blocker.tryAcquire();
```
shiveshr, line 80: tryacquire will not block, right? which means even if it doesnt get the permit here the subsequent code will be executed. is that intended?

common/src/main/java/io/pravega/common/concurrent/ThreadPoolScheduledExecutorService.java
tkaitchuck: While it technically relies on time, I did the unit tess carefully. The all of the tests together add up to 0.6 sec and they are robust against delays or other timing issues.

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
tkaitchuck: Yes. That is intended.

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,347 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import javax.annotation.Nonnull;
+import lombok.Data;
+import lombok.val;
+
+/**
+ * Provides a unbounded blocking queue which {@link Scheduled} items can be added to.
+ * Items which are scheduled will not be returned from {@link #poll()} or {@link #take()} until their scheduled time.
+ */
+public class ScheduledQueue<E extends Scheduled> extends AbstractQueue<E> implements BlockingQueue<E> {
```
Tristan1900, line 40: Just curious, what's the difference between ScheduledQueue and [DelayQueue](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/DelayQueue.html), they look pretty similar to me. Thanks!

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,347 @@
...
+     * This call is O(1).
+     */
+    @Override
+    public int size() {
+        return (int) Long.min(itemsAdded.get() - itemsRemoved.get(), Integer.MAX_VALUE);
+    }
+
+    /**
+     * Always returns {@code Integer.MAX_VALUE} because
+     * a {@code OptionalDealyQueue} is not capacity constrained.
```
andreipaduroiu, line 243: Where is this class? Are you referring to `ScheduledQueue`?

common/src/main/java/io/pravega/common/concurrent/ThreadPoolScheduledExecutorService.java
```diff
@@ -0,0 +1,388 @@
...
+
+/**
+ * An implementation of {@link ScheduledExecutorService} which uses a thread pool. 
+ * 
+ * This class is similar to ScheduledThreadPoolExecutor but differs in the following ways:
+ * 
+ * 1. The thread pool supports growing. IE: {@code maxPoolSize} and {@code corePoolSize} don't have to be the same.
+ * 2. Queued tasks are stored in a lock-free queue so that no bottlenecks can occur on submit.
+ * 3. Scheduling a task without a delay is O(1) as opposed to O(log(n))
+ * 4. Canceling a task actually removes it from the queue and is O(n) as opposed to a no-op which leaves it in the queue or O(log(n)) when {@code setRemoveOnCancelPolicy(true)}.
```
andreipaduroiu, line 55: I wonder, could we avoid the `O(n)` if we mark the task as "cancelled" and just drop it when its time is up? I understand we may have a (temporary) leak of pointers, but it should avoid the `O(n)` task, wouldn't it?

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
tkaitchuck: https://github.com/pravega/pravega/issues/6145#issuecomment-879453547

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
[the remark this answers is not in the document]
tkaitchuck: If the thread is interrupted it will throw as soon as it attempts to block below. We have no need to check for iterupted early. It's very unlikely as it should only occur upon shutdown, so it doesn't provide an advantage to avoid one map lookup in that case.

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
tkaitchuck: Yes

common/src/main/java/io/pravega/common/concurrent/ThreadPoolScheduledExecutorService.java
tkaitchuck: It is also holding onto anything in the stack of the runnable. Which may be non-trivial amount of memory. 
If it is considered desirable I could do this for non-delayed tasks, as these may be assumed to run almost immediately, but keep the removal logic for delayed ones.

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
tkaitchuck: I also added a comment.

common/src/test/java/io/pravega/common/concurrent/ThreadPoolScheduledExecutorServiceTest.java
```diff
@@ -0,0 +1,304 @@
...
+    
+    @Test(timeout = 10000)
+    public void testRunsDelayTask() throws Exception {
+        ThreadPoolScheduledExecutorService pool = createPool(1, 1);
+        CompletableFuture<Long> result = new CompletableFuture<Long>();
+        long startTime = System.nanoTime();
+        Future<Boolean> future = pool.schedule(() -> result.complete(System.nanoTime()), 100, MILLISECONDS);
+        long runTime = result.get(5, SECONDS);
+        assertTrue(runTime > startTime + 50 * 1000 * 1000);
+        assertTrue(future.get(0, MILLISECONDS));
```
tkaitchuck, line 71: TODO: This can fail.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceFactory.java
```diff
@@ -71,20 +68,18 @@
 
         // In all of the below, the ThreadFactory is created in this class, and its toString() returns the pool name.
         if (this.detectionLevel == ThreadLeakDetectionLevel.None) {
-            this.createScheduledExecutor = (size, factory) -> new ScheduledThreadPoolExecutor(size, factory, new CallerRuns(factory.toString()));
+            this.createScheduledExecutor = (size, factory) -> new ThreadPoolScheduledExecutorService(size, size, 100, factory, new AbortPolicy());
```
shiveshr, line 71: nit: can we make this `100` constant?
should this be passed as in input param?

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,350 @@
...
+     * Retrieves, but does not remove, the head of this queue, or
+     * returns {@code null} if this queue is empty. 
+     * 
+     * Unlike {@code poll}, this method returns the elements who's scheduled time has not yet arrived.
+     *
+     * @return the head of this queue, or {@code null} if this queue is empty
+     */
+    @Override
+    public E peek() {
+        val ready = readyTasks.peek();
```
shiveshr, line 220: though it doesnt matter, but the behaviour of this method, because of its lockless nature is that an entry could be added to ready tasks either before this line after this method's invocation or after it has peeked here. 
so it would not give a consistent answer as compared to delayed queue. but i think its ok as long as we explain that in the javadocs.

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,350 @@
...
+/**
+ * Provides a unbounded blocking queue which {@link Scheduled} items can be added to.
+ * Items which are scheduled will not be returned from {@link #poll()} or {@link #take()} until their scheduled time.
+ * 
+ * This class is similar to DelayQueue but it allows a delay to be optional. This allows adding and polling non-delayed tasks in O(1).
+ * It also it lock-free, and may offer higher throughput under contention. 
+ */
+public class ScheduledQueue<E extends Scheduled> extends AbstractQueue<E> implements BlockingQueue<E> {
+   
+    private final AtomicLong itemsAdded = new AtomicLong(0);
```
shiveshr, line 45: why are we counting all the items ever added and removed. 
i know that Long.Max is a very large value and we will not exhaust this practically. 
but can we do better?

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,350 @@
...
+        blocker.drainPermits();
+        while (itemCount < maxElements) {
+            E item = readyTasks.poll();
+            if (item == null) {
+                break;
+            }
+            c.add(item);
+            itemCount++;
+        }
+        while (itemCount < maxElements) {
```
shiveshr, line 323: same issue with drainTo here. its behaviour is that if readyTasks are added to after it had drained all elements in it and moved to delayed tasks then it could cause inconsistency.

take following example:
r`i` represents read tasks and d`i` represents delayed tasks.

lets say readyTasks has 1 entry r1.
delayedTasks has 0 entries.


some app could call following:
scheduledQueue.drainTo(2); 
while that is running and has read r1, concurrently following could have been called in another thread:
scheduledQueue.offer(r2);
scheduledQueue.offer(d1);

drainTo method will end up returning r1, d1.
when the order of entry was r1, r2, d1 and d1 was added after r2 so if d1 is drained r2 should have been drained as well.

shiveshr (changes requested):

shiveshr (approved):
LGTM, except for one comment. removing the block.

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceFactory.java
```diff
@@ -246,13 +237,13 @@ private void logNewThreadPoolCreated(String poolName) {
...
     @VisibleForTesting
-    void checkThreadPoolLeak(ThreadPoolExecutor e, Exception stackTraceEx) {
+    void checkThreadPoolLeak(ThreadPoolScheduledExecutorService e, Exception stackTraceEx) {
         if (this.detectionLevel == ThreadLeakDetectionLevel.None) {
             // Not doing anything in this case.
             return;
         }
 
-        if (!e.isShutdown() || !e.isTerminated()) {
+        if (!e.isShutdown()) {
```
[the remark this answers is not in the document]
tkaitchuck: Both are passed to the impl threadpool. But they have different meaning. Shutdown means shutdown was called. Terminated means all the treads have exited their run methods and stopped. It is possible for the running threads to not hold a reference to the pool of which they are a member. As such it may be the case that the pool is GCed after being properly closed before the thread has actually exited. This was causing some spurious test failures.

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
```diff
@@ -0,0 +1,350 @@
...
+import javax.annotation.Nonnull;
+import lombok.Data;
+import lombok.val;
+
+/**
+ * Provides a unbounded blocking queue which {@link Scheduled} items can be added to.
+ * Items which are scheduled will not be returned from {@link #poll()} or {@link #take()} until their scheduled time.
+ * 
+ * This class is similar to DelayQueue but it allows a delay to be optional. This allows adding and polling non-delayed tasks in O(1).
+ * It also it lock-free, and may offer higher throughput under contention.
```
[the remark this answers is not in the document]
tkaitchuck: Neither of those data structures use locks (per their javadocs). I believe they do everything internally with compareAndSets.

andreipaduroiu (approved):

common/src/main/java/io/pravega/common/concurrent/ExecutorServiceFactory.java
tkaitchuck: Removed

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
tkaitchuck: As the inherited javadoc says:
"the behavior of this operation is undefined if the specified collection is modified while the operation is in progress."

common/src/main/java/io/pravega/common/concurrent/ScheduledQueue.java
tkaitchuck: Added a line in the javadoc.

shrids (approved):
LGTM.

PR #5930, opened 2021-04-14
Issue 5944:  Improve createReaderGroup and getReaderGroup APIs

shrids:
**Change log description**  
- Ensure `getReaderGroup` API throws a `ReaderGroupNotFoundException `if the ReaderGroup does not exist.
- Ensure `createReaderGroup` API indicates if the reader group already exists.

**Purpose of the change**  
Fixes #5944 

**What the code does**  
- Ensure `getReaderGroup` API throws a `ReaderGroupNotFoundException `if the ReaderGroup(RG) does not exist.
- Ensure `createReaderGroup` API indicates if the RG already exists. 

```
// Updated API signature to create a new RG.
  /**
     * Creates a new ReaderGroup.
     *
     * Readers will be able to join the group by calling
     * {@link io.pravega.client.EventStreamClientFactory#createReader(String, String, Serializer, ReaderConfig)}
     * . Once this is done they will start receiving events from the point defined in the config
     * passed here.
     * <p>
     * Note: This method is idempotent assuming called with the same name and config. This method
     * may block.
     * @param groupName The name of the group to be created.
     * @param config The configuration for the new ReaderGroup.
     * @return True if ReaderGroup is created and false if it already exists.
     * @throws ConfigMismatchException If the reader group already exists with a different configuration. Use 
     * {@link ReaderGroup#resetReaderGroup} to change the reader group configuration.
     * 
     */
    boolean createReaderGroup(String groupName, ReaderGroupConfig config) throws ConfigMismatchException;
```

**How to verify it**  
All the existing and the newly added tests should pass.

PR #6136, opened 2021-07-07
Issue 6075: Remove ConfigMismatchException from createReaderGroup API.

### LINKED ISSUE #6075: Extra warning log when call getReaderGroup that not yet created
**Describe the bug**
Pravega used to persist the idempotence of the call `createReaderGroup`, so flink connector used to have those part of code like this in the checkpoint hook.
```
readerGroupManager.createReaderGroup(readerGroupName, readerGroupConfig);
readerGroup = readerGroupManager.getReaderGroup(readerGroupName);
```

In a recovery case, the connector would reset the reader group with adding a start streamcut in the config, and when the Flink job recovers, the `createReaderGroup` call would not fail before, but after #5930 fix, it would fail now as the reader group config is not the same as stored.

So in https://github.com/pravega/flink-connectors/pull/516, we changed this into a "get first and then create" way as https://github.com/pravega/flink-connectors/pull/516/files#diff-077ca27a878ae584603740b5a82c4e4d826748be01c56649493eede64e3fb29dR87

When running Flink applications with the new connector, it would always print a warn log like
```
09:38:11,985 WARN  io.pravega.client.control.impl.ControllerImpl                 - gRPC call for getCurrentSegments with trace id 8096979106827907034 and parameters [examples, _RGflinkmbjns4awk3ajmcrvzqeb] failed with server error.
io.pravega.shaded.io.grpc.StatusRuntimeException: NOT_FOUND: get entry: key: _RGflinkmbjns4awk3ajmcrvzqeb table: _system/_tables/examples/streamsInScope.#.5903df2a-4697-432b-a91c-dfa96bbf6623
        at io.pravega.shaded.io.grpc.Status.asRuntimeException(Status.java:533)
        at io.pravega.shaded.io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:442)
...
```

This comes from the first `getReaderGroup` call with a non-existing RG. 

Is there a way that we can get rid of such warn logs?

**Additional information**

@shrids has some discussion with me about this problem, and we have some thoughts on this as listed, but still controversial.

1. Add a `boolean checkReaderGroupExists(String rgName)` in the `ReaderGroupManager`
2. Turn down the log level from `log.warn` into `log.debug`
3. Change the GRPC return for this as a normal NOT_FOUND instead of an exception 

We would like to involve more people in the discussion to see how we can solve this.

shrids:
**Change log description**  
- Remove ConfigMismatchException from createReaderGroup API as it complicates the usage pattern.

**Purpose of the change**  
Fixes #6075 

**What the code does**  
Remove ConfigMismatchException from createReaderGroup API as it complicates the usage pattern.

**How to verify it**  
Existing and newly added tests should pass.


client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
```diff
@@ -73,11 +72,10 @@ public static ReaderGroupManager withScope(String scope, ClientConfig clientConf
...
      * @param groupName The name of the group to be created.
      * @param config The configuration for the new ReaderGroup.
-     * @return True if ReaderGroup was created.
-     * @throws ConfigMismatchException If the reader group already exists with a different configuration. Use {@link ReaderGroup#resetReaderGroup} to change
-     * the reader group configuration.
+     * @return The {@link ReaderGroupConfig} of the ReaderGroup. If a ReaderGroup already exists then the older configuration
+     * is returned.
      */
-    boolean createReaderGroup(String groupName, ReaderGroupConfig config) throws ConfigMismatchException;
+    ReaderGroupConfig createReaderGroup(String groupName, ReaderGroupConfig config);
```
pbelgundi, line 78: It would help to change the `createReaderGroup` API to a `getOrCreate()` API, so that it would create the ReaderGroup if one does not exist or else just return a reference to a Reader Group object. 
With a `getOrCreateReaderGroup()` API users would not need to issue 2 separate calls `createReaderGroup` and then again `getReaderGroup`.

However, I also see is that the Reader Group class does not hold a reference to the Reader Group config, which also needs to be fixed..

pbelgundi (changes requested):

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
shrids: A `ReaderGroup` object uses a StateSynchronizer. This implies the user would have to close this ReaderGroup object which would be returned as part of a `getOrCreate()`, which I think would be cumbersome for users. 
The ReaderGroupManager already has an API for fetching the ReaderGroup object.
```java
    /**
     * Returns the requested reader group.
     * 
     * @param groupName The name of the group
     * @return Reader group with the given name
     * @throws ReaderGroupNotFoundException If the reader group does not exist.
     */
    ReaderGroup getReaderGroup(String groupName) throws ReaderGroupNotFoundException;
```

crazyzhou (approved):
LGTM. I would prefer this approach that exposing the controller config to the user can bring more flexibility.
In fact, it returns back to the behavior of idempotent `createReaderGroup` call before this change #5930 .
Flink connector can take this benefit and change back to the original "create-and-then-get" style in 0.9 without breaking compatibility(ref: https://github.com/pravega/flink-connectors/blob/r0.9/src/main/java/io/pravega/connectors/flink/ReaderCheckpointHook.java#L85)

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
fpj, line 78: This is replacing the configuration unconditionally, right? As we spoke offline, the concern is that the application accidentally overwrites an existing configuration, so is the idea that we pass the responsibility of checking the configuration to the application? By the time the application is able to check, it might be late.

If you can elaborate on how you see the application using it, I'd appreciate it.

fpj (changes requested):

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
shrids: The API does not overwrite the ReaderGroup configuration. If the ReaderGroup(RG) exists then the current configuration of the RG is returned.

pbelgundi: It would help to merge this API and the getReaderGroup() API into a single - getOrCreateReaderGroup() API that returns a ReaderGroup. Also ReaderGroupConfiguration should be part of the ReaderGroup object.
I don't see the point in having 2 separate APIs here, if create is going to return the same existing Reader Group configuration.... its actually doing a getOrCreate() on the ReaderGroup configuration without explicitly saying so....

From API design perspective, it looks strange that a **createReaderGroup()** API returns **ReaderGroupConfiguration**. I would expect it to either return nothing (_if no exception is thrown the reader group is created_) or return a reference to the created reader group (which essentially is doing a getOrCreateReaderGroup())

crazyzhou: For the proposal for `getOrCreateReaderGroup`, it's a better and more natural solution to me. It can make the current Flink connector usage simpler into one line.
For the responsibility to close the reader group, I think it should leave to the user and it's always in this way.
Deleting the old `createReaderGroup` call may break the compatibility, we need to check and test if we choose this, and I would propose deprecate it now and then delete in the future releases.

andreipaduroiu: @shrids Doesn't this break backward compatibility? How would we handle existing users that have already written code against this API?

shrids: The change to return a boolean was done in the 0.10.x/master branch. The API in 0.9.x is 
```
void createReaderGroup(String groupName, ReaderGroupConfig config);
```
Hence this does not break backward compatibility.

The drawback of the API in 0.9.x is that if the reader group exists it is not initmated to the user. It also does not indicate to the user if reader group config passed is different from that of the existing reader group configuration.

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
```diff
@@ -73,11 +72,10 @@ public static ReaderGroupManager withScope(String scope, ClientConfig clientConf
      * may block.
      * @param groupName The name of the group to be created.
      * @param config The configuration for the new ReaderGroup.
-     * @return True if ReaderGroup was created.
-     * @throws ConfigMismatchException If the reader group already exists with a different configuration. Use {@link ReaderGroup#resetReaderGroup} to change
-     * the reader group configuration.
+     * @return The {@link ReaderGroupConfig} of the ReaderGroup. If a ReaderGroup already exists then the older configuration
```
tkaitchuck, line 75:
```suggestion
     * @return The {@link ReaderGroupConfig} of the ReaderGroup. If a ReaderGroup already exists then the existing configuration
```

tkaitchuck (approved):

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
shrids: I have added an helper API ` ReaderGroup getOrCreate(String groupName, ReaderGroupConfig config)` in the ReaderGroupManager.

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
shrids: fixed it.

client/src/main/java/io/pravega/client/admin/ReaderGroupManager.java
eolivelli: in my opinion it is better to save users from breaking changes in the API

can't we keep `boolean` as return type ?

client/src/main/java/io/pravega/client/admin/impl/ReaderGroupManagerImpl.java
```diff
@@ -98,18 +97,16 @@ public boolean createReaderGroup(String groupName, ReaderGroupConfig config) thr
         ReaderGroupConfig controllerConfig = getThrowingException(controller.createReaderGroup(scope, groupName, config));
         if (!controllerConfig.equals(config)) {
             log.warn("ReaderGroup {} already exists with pre-existing configuration {}", groupName, controllerConfig);
-            throw new ConfigMismatchException(groupName, controllerConfig);
```
eolivelli, line 101: with this change we are returning a new configuration.
but this is dangerous because it is very hard to understand from the returned value if the request configuration has been applied or not.
the user must compare the configuration (returned vs given) in order to understand if the configuration is the expected one


IIUC this is not much user friendly and error prone

eolivelli (changes requested):

RaulGracia:
Closing due to inactivity. @tkaitchuck, please let us know if this is required, so we can create a new PR.

PR #6138, opened 2021-07-07
Issue 6093: Conditional Segment Merge

### LINKED ISSUE #6093: Conditional Segment Merger
We need the ability to condition segment mergers on attribute values. 

Currently a segment merger is performed unconditionally. A request is received on the Segment Store to merge Segment A into Segment B, and A is concatenated at the end of B as a result. The only preconditions for this to work are that both A and B exist, and B is not sealed.

We need to expand this ability to support the following:
- Unconditional segment merge (the functionality we have now).
- Merge Segment A into B and atomically check and update a set of attributes on B. (new)

**Proposed Solution**
1. Update the `StreamSegmentStore.mergeSegment` contract interface to accept a nullable `Collection<AttributeUpdate>`. See `append()` for example. For simplicity (as this is used in a lot of tests), consider adding an overload of the method and have a default implementation of the old API to simply invoke the new one with no attribute updates.
2. Update the `MergeSegmentOperation` to support attributes. See `StreamSegmentAppendOperation` and `UpdateAttributesOperation` for examples. Essentially they need to implement the `AttributeUpdaterOperation` interface. A new serialization revision may be required to support this.
3. Update the `SegmentMetadataUpdateTransaction` to properly handle the new feature of the `MergeSegmentOperation`. Both the `preProcessOperation` and `acceptOperation` methods (for that operation) need some adjustment. There already are helpers to deal with attributes (see how appends and update attributes are handled). This should be as simple as invoking two methods - they deal with validation and acceptance so not much else should be needed here.
4. Update PravegaRequestProcessor and WireCommands to add support for attribute updates. We'll want to see if we want to support multiple attribute updates at the same time or just one.

RaulGracia:
**Change log description**  
Adds a new `mergeStreamSegment()` call that enables to conditionally update a set of attributes on the target Segment and merge the source Segment into the target Segment if the attributes have been successfully updated. 

**Purpose of the change**  
Fixes #6093.

**What the code does**  
According to the description of #6093, this PR introduces the following changes:
- Adds a new `mergeStreamSegment()` method in `StreamSegmentStore` that accepts a `AttributeUpdateCollection` as argument. These arguments are expected to be conditionally updated on the `target` segment before doing the actual merge. Note that the existing `mergeStreamSegment()` is still there for compatibility.
- Modified `PravegaRequestProcessor` and `WireCommands.MergeSegments` to handle attribute updates. Note that we enable to update a set of attributes on a single request.
- Added `attribtueUpdates` to `MergeSegmentOperation`, which now implements `AttributeUpdaterOperation`.
- `SegmentMetadataUpdateTransaction` now invokes `preProcessAttributes` upon `preProcessAsTargetSegment` and `acceptAttributes` upon `acceptAsTargetSegment`, so attribute updates take place when executing the mergeSegments operation.

There are two relevant points when using this new conditional call:
- If the attribute update update fails when merging segments, the client will get a `SegmentAttributeUpdated` reply message for the `requestId` of the merge operation and `success=false`. This is similar to what the client get when trying to execute an `UpdateAttribute` command.
- If the source segment of a transaction is empty, the server will perform the attribute update operation anyway against the target segment (if there are attributes to update).

**How to verify it**  
Tests added to verify the new call.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/StreamSegmentContainer.java
```diff
@@ -593,7 +602,7 @@ public boolean isOffline() {
                                     sourceMetadata.getLength(), sourceMetadata.getAttributes()));
                 } else {
                     // Source now has some data - we must merge the two.
-                    MergeSegmentOperation operation = new MergeSegmentOperation(targetSegmentId, sourceSegmentId);
+                    MergeSegmentOperation operation = new MergeSegmentOperation(targetSegmentId, sourceSegmentId, attributeUpdates);
```
andreipaduroiu, line 605: This one is a bit tricky. Check how we process `StreamSegmentAppendOperation` and  `UpdateAttributesOperation`.

You need to change this method to invoke `processAttributeUpdaterOperation` as there are some scenarios which will not work out of the box - that method takes care of it for you.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
```diff
@@ -397,7 +397,7 @@ void preProcessOperation(DeleteSegmentOperation operation) {
      * @throws IllegalArgumentException        If the operation is for a different Segment.
      */
     void preProcessAsTargetSegment(MergeSegmentOperation operation, SegmentMetadataUpdateTransaction sourceMetadata)
-            throws StreamSegmentSealedException, StreamSegmentNotSealedException, MetadataUpdateException {
+            throws StreamSegmentSealedException, StreamSegmentNotSealedException, MetadataUpdateException, BadAttributeUpdateException {
```
andreipaduroiu, line 400: Update Javadoc too to reflect the new exception that might be thrown.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -113,17 +127,38 @@ public long getLength() {
...
+        return AttributeUpdateCollection.from(attributeUpdates);
+    }
+
+    @Override
+    public boolean hasOnlyCoreAttributes() {
+        return AttributeUpdaterOperation.super.hasOnlyCoreAttributes();
+    }
+
+    @Override
+    public boolean isInternal() {
```
andreipaduroiu, line 153: You may omit this method as it is defaulted in `AttributeUpdaterOperation`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -113,17 +127,38 @@ public long getLength() {
...
+
+    //region AttributeUpdaterOperation implementation
+
+    @Override
+    public AttributeUpdateCollection getAttributeUpdates() {
+        return AttributeUpdateCollection.from(attributeUpdates);
+    }
+
+    @Override
+    public boolean hasOnlyCoreAttributes() {
```
andreipaduroiu, line 148: You may omit this method. It's already declared in the interface.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -18,17 +18,25 @@
...
  * Log Operation that indicates a Segment is to be merged into another Segment.
  */
-public class MergeSegmentOperation extends StorageOperation {
+public class MergeSegmentOperation extends StorageOperation implements AttributeUpdaterOperation {
     //region Members
 
     private long streamSegmentOffset;
     private long length;
     private long sourceSegmentId;
+    private Collection<AttributeUpdate> attributeUpdates;
```
andreipaduroiu, line 39: Please use `AttributeUpdateCollection`

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/StreamSegmentStore.java
```diff
@@ -188,6 +188,22 @@
...
+     * @param targetSegmentName The name of the StreamSegment to merge into.
+     * @param sourceSegmentName The name of the StreamSegment to merge.
+     * @param attributeUpdates  A Collection of Attribute-Values to set on the target StreamSegment. May be null.
+     *                          See Notes about AttributeUpdates in the interface Javadoc.
+     * @param timeout           Timeout for the operation.
+     * @return A CompletableFuture that, when completed normally, will contain a MergeStreamSegmentResult instance with information about the
+     * source and target Segments. If the operation failed, the future will be failed with the causing exception.
+     * @throws IllegalArgumentException If any of the arguments are invalid.
+     */
+    CompletableFuture<MergeStreamSegmentResult> mergeStreamSegment(String targetSegmentName, String sourceSegmentName,
```
andreipaduroiu, line 204: Please use `AttributeUpdateCollection` instead of `Collection<AttributeUpdates>` - be consistent with the rest of the API.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -148,12 +183,23 @@ protected void beforeSerialization(MergeSegmentOperation o) {
         }
 
         private void write00(MergeSegmentOperation o, RevisionDataOutput target) throws IOException {
-            target.length(SERIALIZATION_LENGTH);
+            int attributesLength = o.attributeUpdates == null ? target.getCompactIntLength(0) :
```
andreipaduroiu, line 186: This will break backwards compatibility. Please declare a new revision and ONLY write the attributes in that new revision (and read them from there). Leave revision 0 as it was before.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/StreamSegmentService.java
```diff
@@ -119,6 +119,15 @@ public StreamSegmentService(SegmentContainerRegistry segmentContainerRegistry, S
                 "mergeTransaction", targetStreamSegment, sourceStreamSegment);
     }
 
+    @Override
+    public CompletableFuture<MergeStreamSegmentResult> mergeStreamSegment(String targetStreamSegment, String sourceStreamSegment,
+                                                                          Collection<AttributeUpdate> attributes, Duration timeout) {
+        return invoke(
+                sourceStreamSegment,
+                container -> container.mergeStreamSegment(targetStreamSegment, sourceStreamSegment, attributes, timeout),
+                "mergeTransaction", targetStreamSegment, sourceStreamSegment);
```
andreipaduroiu, line 128: `mergeStreamSegment`

Please fix above too - I believe this has been inconsistent for a very long time.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1238,13 +1240,24 @@ public static WireCommand readFrom(DataInput in, int length) throws IOException
     }
 
     @Data
+    @AllArgsConstructor
```
andreipaduroiu, line 1243: This is included in `@Data`. Do you need it for something specific?

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1238,13 +1240,24 @@ public static WireCommand readFrom(DataInput in, int length) throws IOException
...
     public static final class MergeSegments implements Request, WireCommand {
         final WireCommandType type = WireCommandType.MERGE_SEGMENTS;
         final long requestId;
         final String target;
         final String source;
         @ToString.Exclude
         final String delegationToken;
+        final List<ConditionalAttributeUpdate> attributeUpdates;
```
andreipaduroiu, line 1251: FYI this would have required bumping the Wire Protocol version. However we have already bumped it for release 0.10, so no need to do it again. But in the future any such changes to the wire format would require that.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2534,4 +2553,35 @@ public void release() {
...
+
+    /**
+     * Convenience class to encapsulate the contents of an attribute update when several should be serialized in the same
+     * WireCommand.
+     */
+    @Data
+    public static final class ConditionalAttributeUpdate {
+        public static final int LENGTH = 4 * Long.BYTES; // UUID (2 longs) + oldValue + newValue
+
+        private final UUID attributeId;
```
andreipaduroiu, line 2565: I think we should likely take this one step further. We should include support for `AttributeUpdateType` as well. Perhaps in the future, we'd want to do Replace/ReplaceIfEquals/Accumulate/etc. 

If we're doing this now, we might as well add another byte to indicate the type.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/WireCommandsTest.java
```diff
@@ -894,6 +894,16 @@ public void testConditionalBlockEnd() throws IOException {
                 ce -> ce.getData().refCnt());
     }
 
+    @Test
+    public void testMergeSegmentsWithAttributes() throws IOException {
+        List<WireCommands.ConditionalAttributeUpdate> attributeUpdates = Arrays.asList(
```
andreipaduroiu, line 899: Please add a unit test to verify backwards compatibility (like when you receive a serialization that does not include attributes).

andreipaduroiu (changes requested):

andreipaduroiu:
@RaulGracia Looks like the segment container unit tests and segment store integration tests are failing because of segments not sealed while attempting the merge. Are we skipping this step with the new approach?

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
RaulGracia: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Done.

segmentstore/contracts/src/main/java/io/pravega/segmentstore/contracts/StreamSegmentStore.java
RaulGracia: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: True, my fault. I added new write (`write01`) and read (`read00`) methods for the new operation version, as well as the new revision declaration.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/store/StreamSegmentService.java
RaulGracia: Done.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
RaulGracia: According to the documentation, `@Data` includes `@RequiredArgsConstructor` instead of `@AllArgsConstructor`, right? (I don't think it matters much in this case, though). In any case, this is needed because the behavior of `@Data` seems to create the constructor if no constructor is yet defined. As you can see, I have added a custom constructor (without `attributeUpdates`) for comparability. In this case, the "all args" constructor does not get generated (I tested that by removing the custom constructor, and in that case the "all args" constructor gets generated). Apparently, by explicitely adding the annotation `@AllArgsConstructor`, Lombok generates the constructor with all the arguments, irrespective of other custom constructors defined.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
RaulGracia: Ok, got it.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
RaulGracia: Done, now `ConditionalAttributeUpdate` can take any attribute update type.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/WireCommandsTest.java
RaulGracia: Added a test called `testCompatibilityMergeSegmentsV5()` to cover that case.

RaulGracia:
@pbelgundi I have done another pass on this PR and added tests to cover corner cases. Please, have a look whenever you have some time.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -470,7 +470,17 @@ public void mergeSegments(MergeSegments mergeSegments) {
...
 
         log.info(mergeSegments.getRequestId(), "Merging Segments {} ", mergeSegments);
-        segmentStore.mergeStreamSegment(mergeSegments.getTarget(), mergeSegments.getSource(), TIMEOUT)
+
+        // Populate the AttributeUpdates for this mergeSegments operation, if any.
+        AttributeUpdateCollection attributeUpdates = new AttributeUpdateCollection();
+        if (mergeSegments.getAttributeUpdates() != null) {
+            for (WireCommands.ConditionalAttributeUpdate update : mergeSegments.getAttributeUpdates()) {
+                attributeUpdates.add(new AttributeUpdate(AttributeId.fromUUID(update.getAttributeId()),
+                    AttributeUpdateType.get(update.getAttributeUpdateType()), update.getNewValue(), update.getOldValue()));
```
andreipaduroiu, line 479: Thanks for adding `AttributeUpdateType`. One thing to do is we need to make sure that the values within `WireCommands.ConditionalAttributeUpdate` match those from `AttributeUpdateType`. See my comment in `WireCommands`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -490,6 +500,15 @@ public void mergeSegments(MergeSegments mergeSegments) {
...
                                         });
                             return null;
+                        } else if (Exceptions.unwrap(e) instanceof BadAttributeUpdateException) {
+                            log.debug(mergeSegments.getRequestId(), "Conditional merge failed (Source segment={}, Target segment={})",
+                                    mergeSegments.getSource(), mergeSegments.getTarget(), e);
+                            final Consumer<Throwable> failureHandler = t -> {
+                                log.error("Error (Segment = '{}', Operation = '{}')", mergeSegments.getTarget(), "handling result of {}" + operation, t);
+                                connection.close();
+                            };
+                            invokeSafely(connection::send, new SegmentAttributeUpdated(mergeSegments.getRequestId(), false), failureHandler);
```
andreipaduroiu, line 510: I understand why you are using `invokeSafely`, but that is not used anywhere else in this class. If you observe how each request is handled, there's a catch-all `handleException` at the end of each future chain which should also catch problems with `connection.send`. I recommend we be consistent with regards to this behavior.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -47,6 +54,12 @@ public MergeSegmentOperation(long targetSegmentId, long sourceSegmentId) {
         this.streamSegmentOffset = -1;
     }
 
+    public MergeSegmentOperation(long targetSegmentId, long sourceSegmentId, AttributeUpdateCollection attributeUpdates) {
+        this(targetSegmentId, sourceSegmentId);
+        // this.attributeUpdates is already set as an empty collection in the above constructor. Overwrite it only if needed.
+        this.attributeUpdates = attributeUpdates == null ? this.attributeUpdates : attributeUpdates;
```
andreipaduroiu, line 60: I don't understand what exactly is in `this.attributes` (the first part of the `?` operator)? It's null at this point. 
I think just setting `this.attributeUpdates = attributeUpdates` should suffice.

EDIT: I see you initialize it to an empty list. Then it makes sense. However I would prefer if the behavior is consistent with `UpdateAttributesOperation` and `StreamSegmentAppendOperation`.  Those do allow `null` for this.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -47,6 +54,12 @@ public MergeSegmentOperation(long targetSegmentId, long sourceSegmentId) {
         this.streamSegmentOffset = -1;
     }
 
+    public MergeSegmentOperation(long targetSegmentId, long sourceSegmentId, AttributeUpdateCollection attributeUpdates) {
+        this(targetSegmentId, sourceSegmentId);
```
andreipaduroiu, line 58: I think it's simpler if you make the other constructor invoke this one (with `null` for the last arg), and move the body of the other constructor in here.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -113,17 +126,28 @@ public long getLength() {
...
     public String toString() {
         return String.format(
-                "%s, SourceSegmentId = %d, Length = %s, MergeOffset = %s",
+                "%s, SourceSegmentId = %d, Length = %s, MergeOffset = %s, Attributes = %d",
                 super.toString(),
                 getSourceSegmentId(),
                 toString(getLength(), -1),
-                toString(getStreamSegmentOffset(), -1));
+                toString(getStreamSegmentOffset(), -1),
+                this.attributeUpdates.size());
```
andreipaduroiu, line 134: FYI if you decide to initialize this to null, this may require some changes.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -113,17 +126,28 @@ public long getLength() {
...
+    }
+
+    //endregion
+
+    //region AttributeUpdaterOperation implementation
+
+    @Override
+    public AttributeUpdateCollection getAttributeUpdates() {
+        return AttributeUpdateCollection.from(attributeUpdates);
```
andreipaduroiu, line 143: Why do you make a collection out of the collection? It's supposed to be immutable in the first place.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -113,17 +126,28 @@ public long getLength() {
...
     }
 
     //endregion
 
     static class Serializer extends OperationSerializer<MergeSegmentOperation> {
         private static final int SERIALIZATION_LENGTH = 5 * Long.BYTES;
+        private static final int ATTRIBUTE_UUID_UPDATE_LENGTH = RevisionDataOutput.UUID_BYTES + Byte.BYTES + 2 * Long.BYTES;
```
andreipaduroiu, line 150: Tip: add a small comment explaining what all of these are for (it will help you remember later on when you revisit this code)

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -156,12 +181,42 @@ private void write00(MergeSegmentOperation o, RevisionDataOutput target) throws
...
 
+        private void write01(MergeSegmentOperation o, RevisionDataOutput target) throws IOException {
+            if (o.attributeUpdates == null || o.attributeUpdates.isEmpty()) {
+                return;
+            }
+            target.length(target.getCollectionLength(o.attributeUpdates.size(), ATTRIBUTE_UUID_UPDATE_LENGTH));
+            target.writeCollection(o.attributeUpdates, this::writeAttributeUpdateUUID00);
+        }
+
+        private void writeAttributeUpdateUUID00(RevisionDataOutput target, AttributeUpdate au) throws IOException {
```
andreipaduroiu, line 192: I know there's no written convention about this, but the suffix of these methods represent the Version & Revision for which they apply. If you began using these in Version 0, Revision 1, then let's set the suffix to 01 please. Same with read.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
```diff
@@ -156,12 +181,42 @@ private void write00(MergeSegmentOperation o, RevisionDataOutput target) throws
...
+            target.writeLong(au.getAttributeId().getBitGroup(0));
+            target.writeLong(au.getAttributeId().getBitGroup(1));
+            target.writeByte(au.getUpdateType().getTypeId());
```
andreipaduroiu, line 195: FYI To save space, you don't need this field. It is completely ignored upon recovery (as the validation happens once, before it's actually serialized). Since it's only 1 byte (and since appends already have it - I forgot why), then it's really up to you to decide whether to keep it or not.

Same with `getComparisonValue`. This would be a total of 9 bytes you can reduce this by.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1321,14 +1334,21 @@ public void writeFields(DataOutput out) throws IOException {
             out.writeUTF(target);
             out.writeUTF(source);
             out.writeUTF(delegationToken == null ? "" : delegationToken);
+            for (ConditionalAttributeUpdate entry : attributeUpdates) {
```
andreipaduroiu, line 1337: You need to write the number of entries. I see you use a while loop to read them, but you are essentially tying your hands now as you won't be able to add any new fields in the future.

Encoding the number of attribute updates enables you to do a for loop in the read method, and then if (when) we add new fields at the end, we do not have to go out of our way to make it happen.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -2539,4 +2558,38 @@ public void release() {
...
+    /**
+     * Convenience class to encapsulate the contents of an attribute update when several should be serialized in the same
+     * WireCommand.
+     */
+    @Data
+    public static final class ConditionalAttributeUpdate {
+        public static final int LENGTH = 4 * Long.BYTES + 1; // UUID (2 longs) + oldValue + newValue + updateType (1 byte)
+
+        private final UUID attributeId;
+        private final byte attributeUpdateType;
```
andreipaduroiu, line 2571: We need to add at least some static final constants with what values to put here. The Controller won't know the semantics of this, nor does it have access to the Segment store's `AttributeUpdateType.` I suggest we add 2 constants whose values match exactly the ones in `AttributeUpdateType`. For now, we can have `Replace` and `ReplaceIfEquals`. We  can add the other ones later if needed.

andreipaduroiu: And we'd have to somehow validate that the values encoded are correct.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
RaulGracia: Done, added `REPLACE` and `REPLACE_IF_EQUALS` in `ConditionalAttributeUpdate`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
RaulGracia: Ok, sent response as in other parts of `PravegaRequestProcessor` (just replying back with `connection.send()`).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Done, now it can accept `null` values. But note that `UpdateAttributesOperation` actually does not accept `null` values in `attribtueUpdates`, as it has a `Preconditions.checkNotNull` in its constructor.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: My fault, fixed (actually, I fixed that at another point as well).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Done.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Yeah, I was wondering about this as well, but I will take that non-written rule as if it was written from now on.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/operations/MergeSegmentOperation.java
RaulGracia: Ok, good to know, but as it is such a tiny piece of data, I think that the clarity in the code worth it to keep.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
RaulGracia: Good catch, thanks. I have done that writing and reading first the number of entries.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
RaulGracia: Done, and added tests in `PravegaRequestProcessorTests` to validate the values.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -490,6 +500,11 @@ public void mergeSegments(MergeSegments mergeSegments) {
                                                                                             properties.getLength()));
                                         });
                             return null;
+                        } else if (Exceptions.unwrap(e) instanceof BadAttributeUpdateException) {
+                            log.debug(mergeSegments.getRequestId(), "Conditional merge failed (Source segment={}, Target segment={})",
+                                    mergeSegments.getSource(), mergeSegments.getTarget(), e);
```
andreipaduroiu, line 505: This is going to print the full stack trace which will pollute the logs. We don't need it for `BadAttributeUpdateException`; just print the message (`e.toString()` - remember to add a `{}` to the message).

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/WireCommandsTest.java
```diff
@@ -894,14 +896,63 @@ public void testConditionalBlockEnd() throws IOException {
...
+        public void writeFields(DataOutput out) throws IOException {
+            out.writeLong(requestId);
+            out.writeUTF(target);
+            out.writeUTF(source);
+            out.writeUTF(delegationToken == null ? "" : delegationToken);
+        }
+    }
+
+    @Test
+    public void testCompatibilityMergeSegmentsV5() throws IOException {
```
andreipaduroiu, line 940: This is good. FYI you could have achieved the same thing by taking the serialization of a "current" version and then trimming out the trailing bytes (i.e., trim 4 bytes from the end if you serialize 0 attribute updates). That's how we verify some of the other wire commands.

andreipaduroiu (approved):

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
RaulGracia: Done.

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/WireCommandsTest.java
RaulGracia: Ok, good to know.

tkaitchuck (commented):
Bump the wire command version

RaulGracia:
Thanks @tkaitchuck, in [this comment](https://github.com/pravega/pravega/pull/6138#discussion_r665712253) @andreipaduroiu   mentioned that bumping up the version of the wire commands is not necessary in this particular PR, given that it has been already bumped up in `master` recently (for 0.10 release). Given that, I assume that is not necessary to do it again, right?

pbelgundi (approved):
LGTM

PR #6140, opened 2021-07-07
Issue 6139: (SegmentStore) Improving StorageWriter workflow when SLTS metadata is not available

### LINKED ISSUE #6139: Unable to recover SLTS Metadata Table if severely backed up
**Describe the bug**
Upon every iteration of the Storage Writer, it picks new operations to process. If any of those operations references a Segment it wasn't previously working with, it will create a new `SegmentAggregator` for it and fetch information from the `Storage` adapter. If the `Storage` adapter is implemented by SLTS, that would result in a SLTS Metadata Table request. If that segment isn't available for reading (freshly after a recovery - not fully preindexed), we would not be able to initialize those segments. This would lead to the Storage writer to not even attempt to flush any segments, hence the SLTS metadata table stands no chance of being indexed -> deadlock.

**To Reproduce**
1. Need to have at least 1 non-system segment. Call this segment A.
2. Cause the SLTS Metadata Table (SLTS_META) to exceed 128MB in unindexed data. 
3. Cause the owning container to failover.
4. Upon recovery, the Storage Writer will process the recovered operations.
4.1. It will successfully recover the SLTS metadata table (initialize it)
4.2. When trying to initialize segment A, it will call SLTS_META.getInfo(A). This will be rejected/timed out because the `ContainerTableExtensionImpl` detected that SLTS_META is not fully indexed but it refuses to preindex due to it having too much unindexed data (> 128MB).
4.3. The Storage Writer logs and error and starts a new iteration.

There is no way out of this situation. Step 4.2 above would always fail due to SLTS_META not being ready. However the only way at this point for SLTS_META to be ready is if the Storage Writer makes indexing progress on it. However the StorageWriter cannot, because it trips over step 4.2. It is deadlocked.

**Discarded solutions**
These solutions were either attempted or thought of but would not resolve the issue in all cases:
1. Change StorageWriter to still try to flush even if it's unable to initialize Segment Aggregators. This does not cover the case when the SLTS metadata table segment is not already active in the Storage Writer.
2. Increase the pre-index recovery limits. This cannot go over 2GB due to internal constraints (`BufferView` can hold at most 2GB of data, even if composite). If we are backed up even more than that, this won't work.

**Sugested Solution**
- Make the preindex recovery thresholds configurable. This enables an operator to temporarily raise them to as much as needed.
- Change the `WriterTableProcessor` indexer and the `ContainerKeyIndex` pre-index task to process the unindexed tail in small batches vs all at once.
    - For example, if the backup is 3GB, it can do that in about 30 x 100MB batches instead of trying to gulp all 3GB at once.

andreipaduroiu:
**Change log description**  
 Improved the StorageWriter workflow by allowing it to flush (if possible) even if it's unable to initialize new segments.

**Purpose of the change**  
Fixes #6139.

**What the code does**  
The `StorageWriter` can now ignore (after logging) non-critical errors while it's processing its read result (`processReadResult`) and moving on to the `flush` stage. This can help alleviate pressure and exit from a possible deadlock (see #6139).

Unrelated change (but reverting a previous unwanted change):
- `java.gradle`: Removed redundant "PASSED" message for every test. Since we are logging "START", if we don't see anything else, then we know the test succeeded. "FAILED" and "SKIPPED" are the only important ones and represent exceptional cases. This change helps reduce build output clutter.

**How to verify it**  
All existing unit and integration tests must pass.
A new unit test was added to verify this functionality.

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -94,7 +94,7 @@ plugins.withId('java') {
         testLogging.showCauses = true
         testLogging.showExceptions = true
         testLogging.showStackTraces = true
-        testLogging.events = ["STARTED", "PASSED", "SKIPPED", "FAILED"]
+        testLogging.events = ["STARTED", "SKIPPED", "FAILED"]
         maxParallelForks = System.properties['maxParallelForks'] ? System.properties['maxParallelForks'].toInteger() : Math.min(4, Runtime.runtime.availableProcessors())
         minHeapSize = "128m"
         maxHeapSize = "512m"
--- a/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
+++ b/segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/StorageWriter.java
@@ -266,10 +266,17 @@ class StorageWriter extends AbstractThreadPoolService implements Writer {
                 this.executor)
                 .thenRun(() -> {
                     // Clear the last read result from the state before exiting.
-                    Preconditions.checkState(readResult.isEmpty(), "processReadResult exited normally but there are still {} items to process.", readResult.size());
+                    Preconditions.checkState(readResult.isEmpty(), "processReadResult exited normally but there are still %s items to process.", readResult.size());
                     this.state.setLastRead(null);
                     logStageEvent("InputRead", result);
-                });
+                })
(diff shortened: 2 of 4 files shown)


RaulGracia (approved):

andreipaduroiu:
This doesn't solve the problem. We'll try another approach.

PR #6149, opened 2021-07-13
Issue 6139: (SegmentStore) Improving Table Segment Pre-Caching and Background Indexing

### LINKED ISSUE #6139: Unable to recover SLTS Metadata Table if severely backed up

andreipaduroiu:
**Change log description**  
- Made `TableExtensionConfig` externally-configurable. This should allow us to properly tweak this if we have to (for debugging/repair purposes).
- Changed `ContainerKeyIndex` pre-caching to execute in batches, no bigger than 128MB. This will enable precaching of larger segments without the risk of running out of memory.
- Changed `WriterTableProcessor` to index a maximum of 128MB at once, even if the backlog is bigger. Remaining unindexed data will be processed in later iterations.

**Purpose of the change**  
Fixes #6139.

**What the code does**  
- `TableExtensionConfig` 
    - This defines the major config points for the `ContainerTableExtensionConfig` component. The original purpose was to make unit testing easier.
    - This change converts it to an externally-configurable component (just like any other config components) which should enable us to tweak these values if we need to get the Pravega cluster out of a bad state.
- `ContainerKeyIndex` pre-caching
    - Using an async loop to break down the pre-cache segment range into smaller ranges and process them in sequence. This way, no more than 128MB (or whatever is configured) of data will be read in memory at any given time.
 - `WriterTableProcessor`
     - Changed the `flush` behavior to cap the amount of data to process at 128MB. The actual exact amount of data will vary, as it will try to stop on an append boundary (otherwise it may try to stop in the middle of a table entry).

**How to verify it**  
All existing unit tests must pass - this was an "idempotent" change.
Some unit tests have been updated to check for new cases and new unit tests have been added to verify the same.

### DIFF
--- a/common/src/main/java/io/pravega/common/util/TypedProperties.java
+++ b/common/src/main/java/io/pravega/common/util/TypedProperties.java
@@ -154,28 +154,27 @@ public class TypedProperties {
      *                                does not have a default value set, or when the property cannot be parsed as a positive Integer.
      */
     public int getPositiveInt(Property<Integer> property) {
-        return getPositiveInt(property, true);
+        int value = getInt(property);
+        if (value <= 0) {
+            throw new ConfigurationException(String.format("Property '%s' must be a positive integer.", property));
+        }
+        return value;
     }
 
     /**
-     * Gets the value of an Integer property only if it is greater than 0.
+     * Gets the value of an Integer property only if it is non-negative (greater than or equal to 0).
      *
      * @param property The Property to get.
-     * @param strict   If true, then the value must be strictly greater than 0; otherwise 0 is accepted.
      * @return The property value or default value, if no such is defined in the base Properties.
      * @throws ConfigurationException When the given property name does not exist within the current component and the property
-     *                                does not have a default value set, or when the property cannot be parsed as a positive Integer.
+     *                                does not have a default value set, or when the property cannot be parsed as a
+     *                                non-negative Integer.
      */
-    public int getPositiveInt(Property<Integer> property, boolean strict) {
+    public int getNonNegativeInt(Property<Integer> property) {
         int value = getInt(property);
-        if (strict) {
-            if (value <= 0) {
-                throw new ConfigurationException(String.format("Property '%s' must be a positive integer.", property));
-            }
-        } else if (value < 0) {
+        if (value < 0) {
             throw new ConfigurationException(String.format("Property '%s' must be a non-negative integer.", property));
         }
-
         return value;
     }
 
--- a/common/src/main/java/io/pravega/common/util/TypedProperties.java
+++ b/common/src/main/java/io/pravega/common/util/TypedProperties.java
@@ -17,12 +17,11 @@ package io.pravega.common.util;
 
 import com.google.common.base.Preconditions;
 import io.pravega.common.Exceptions;
-import lombok.extern.slf4j.Slf4j;
-
 import java.time.Duration;
 import java.time.temporal.TemporalUnit;
 import java.util.Properties;
 import java.util.function.Function;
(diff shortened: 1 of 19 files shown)


common/src/main/java/io/pravega/common/util/TypedProperties.java
```diff
@@ -155,9 +154,43 @@ public boolean getBoolean(Property<Boolean> property) throws ConfigurationExcept
...
+    /**
+     * Gets the value of an Integer property only if it is greater than 0.
+     *
+     * @param property The Property to get.
+     * @param strict   If true, then the value must be strictly greater than 0; otherwise 0 is accepted.
+     * @return The property value or default value, if no such is defined in the base Properties.
+     * @throws ConfigurationException When the given property name does not exist within the current component and the property
+     *                                does not have a default value set, or when the property cannot be parsed as a positive Integer.
+     */
+    public int getPositiveInt(Property<Integer> property, boolean strict) {
```
sachin-j-joshi, line 169: I think it is better to have a separate overload
`public int getNonNegativeInt(Property<Integer> property)`
instead of passing strict.

andreipaduroiu: Done.

RaulGracia (approved):
LGTM, but it would be great to test this change on a saturated cluster to see how it behaves in a situation similar to the original issue.

sachin-j-joshi (approved):
Looks okay to me.

PR #6154, opened 2021-07-17
Issue 5699: SLTS - Old system journal files should be garbage collected.

### LINKED ISSUE #5699: SLTS - Old system journal files should be garbage collected.
**Problem description**
Currently old system journal files are not garbage collected.

**Problem location**
SystemJournal

journal files older that last N snapshots should be garbage collected.

sachin-j-joshi:

**Change log description**  
Old system journal files are now added garbage collector queue after new snapshot is saved.

**Purpose of the change**  
Fixes #5699

**What the code does**  
Keeps track of all journal files added since last snapshot and then after new snapshot is created these file are added to GC queue.

Note - This code will not clean up previously created garbage. It just prevents GC leaks in future. 

**How to verify it**  
All tests should pass.
System tests should be green.

### DIFF
--- a/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
+++ b/segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
@@ -897,9 +897,7 @@ public class SystemJournal {
                 },
                 v -> epochToRecover.incrementAndGet(),
                 executor)
-                .thenAcceptAsync(vv -> {
-                    pendingGarbageChunks.addAll(journalsProcessed);
-                }, executor);
+                .thenRunAsync(() -> pendingGarbageChunks.addAll(journalsProcessed), executor);
     }
 
     private CompletableFuture<Void> processJournalContents(MetadataTransaction txn, BootstrapState state, String systemLogName, ByteArrayInputStream input) {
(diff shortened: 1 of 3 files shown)


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -880,7 +896,10 @@ private SystemSnapshotRecord readSnapshotRecord(SnapshotInfo snapshotInfo, byte[
                             executor);
                 },
                 v -> epochToRecover.incrementAndGet(),
-                executor);
+                executor)
+                .thenAcceptAsync(vv -> {
```
andreipaduroiu, line 900: thenRunAsync

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -865,6 +880,7 @@ private SystemSnapshotRecord readSnapshotRecord(SnapshotInfo snapshotInfo, byte[
                                                         containerId, epochToRecover.get(), fileIndexToRecover.get());
                                                 return CompletableFuture.completedFuture(null);
                                             } else {
+                                                journalsProcessed.add(systemLogName);
```
andreipaduroiu, line 883: Any reason why we don't add it directly to `pendingGarbageChunks`?

andreipaduroiu (approved):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
sachin-j-joshi: I think the intention is to add them to garbage list only after successful replay. (In reality not able to replay journals is a critical failure and will fail the container, so most likely there won't be any GC running to delete any chunks prematuarely. But thought that this way its just guaranteed to be no GC if there is no successful replay)

Issue #6137, opened 2021-07-07
Large Events: Server (SegmentStore) Implementation


co-jo:
**Describe the solution you'd like**

#6116 Sets up the contract to be facilitate client-server communication. This will allow us to proceed with the required server changes in parallel. This feature request addresses the server side changes required to fulfill the handling of [large events](https://github.com/pravega/pravega/wiki/PDP-43A-Large-Events-(Simplified)) (#6107).

The following outlines the expectations of the implementation:

1. Create a new segment type: `Transient`  in `SegmentType.java`.
  - No need to update `StreamSegmentStore` or `StreamSegmentContainer`. Current API should allow pass-through.

**`MetadataStore.java`:**

2. Update `createSegment` in `MetadataStore.java` to handle `Transient` segments.
  - Ensure that if a segment is `Transient` it may not be a `TableSegment`. This can be done in `SegmentType.java` or `MetadataStore.java`.
  - `Transient` segments are _not_ to be written to the metadata table.
  - `Transient` segments are to be mapped directly into in-memory metadata using `StreamSegmentMapOperation`.
  - If a `Transient` segment is evicted from memory, the eviction process should not auto-write it to the `MetadataSegment` (Verify using a unit test).
    * Eviction of `Transient` segments may lead to garbage if the client forgets about it an it's evicted. In this case, nobody cleans it up. Perhaps we can pin it to memory?

**`MetadataCleanupService`**:

3. Separate out `Transient` segments from the existing eviction logic.
- Do not flush `Transient` segments to Metadata Segment.  
- Consider `Transient` segment candidates for eviction. Any segment with type `Transient` that is not merge is a candidate.
- For each candidate identify those that are 'actively' (exact semantics TBD) used and remove them from the candidate set.
- Delete all remaining candidates using `MetadataStore::delete`.

**`AppendProcessor`** Integration:

- When a connection is closed, if the connection is for a transient segment which has not been merged, issue a delete for said segment. 
- How to identify `Transient` segments that are still needed?
  1. Heartbeat from `AppendProcessor` into the SegmentStore, which will require a lot of plumbing.
  2. Have the `MetadataStore` check (via some new API that is wired into the `AppendProcessor`) if a transient segment is still worked upon.

PR #6171, opened 2021-07-20
Issue 6137: Large Appends SegmentStore

co-jo:
**Change log description**  

* Adds new `TransientSegment` type/role. 
* Adds `createTransientSegment` call in the `AppendProcessor`.
* Overload the `close` call on a `TrackedConnection` to allow for a call back.
* Adds a `close` call on the `AppendProcessor`, which will be called on the closure of its `TrackedConnection`.
* Create according `NameUtils` methods for new `TransientSegment` type.
* Setup boilerplate code for deleting unused transient segements.

**Purpose of the change**  
Partially completes #6137.

**What the code does**  

* See above.

**How to verify it**  

* Run unit tests (TBD).

### DIFF
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -81,8 +81,6 @@ import org.slf4j.LoggerFactory;
 import static io.pravega.segmentstore.contracts.Attributes.ATTRIBUTE_SEGMENT_TYPE;
 import static io.pravega.segmentstore.contracts.Attributes.CREATION_TIME;
 import static io.pravega.segmentstore.contracts.Attributes.EVENT_COUNT;
-import static io.pravega.segmentstore.contracts.Attributes.SCALE_POLICY_RATE;
-import static io.pravega.segmentstore.contracts.Attributes.SCALE_POLICY_TYPE;
 
 /**
  * Process incoming Append requests and write them to the SegmentStore.
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -60,6 +60,8 @@ import io.pravega.shared.protocol.netty.WireCommands.SegmentIsSealed;
 import io.pravega.shared.protocol.netty.WireCommands.SetupAppend;
 import io.pravega.shared.protocol.netty.WireCommands.WrongHost;
 import io.pravega.shared.protocol.netty.WireCommands.CreateTransientSegment;
+import io.pravega.shared.protocol.netty.WireCommands.TransientSegmentCreated;
+import io.pravega.shared.NameUtils;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collection;
@@ -260,14 +262,24 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
 
     @Override
     public void createTransientSegment(CreateTransientSegment createTransientSegment) {
-        long traceId = LoggerHelpers.traceEnter(log, "createTransientSegment", createTransientSegment);
+        String operation = "createTransientSegment";
+        long traceId = LoggerHelpers.traceEnter(log, operation, createTransientSegment);
         Collection<AttributeUpdate> attributes = Arrays.asList(
                 new AttributeUpdate(CREATION_TIME, AttributeUpdateType.None, System.currentTimeMillis()),
                 new AttributeUpdate(ATTRIBUTE_SEGMENT_TYPE, AttributeUpdateType.None, SegmentType.TRANSIENT_SEGMENT.getValue())
         );
-        String parentSegment = createTransientSegment.getParentSegment();
-        store.createStreamSegment(parentSegment, SegmentType.TRANSIENT_SEGMENT, attributes, TIMEOUT);
-        LoggerHelpers.traceLeave(log, "createTransientSegment", traceId);
+        String transientSegmentName = NameUtils.getTransientNameFromId(createTransientSegment.getParentSegment(), UUID.randomUUID());
+        store.createStreamSegment(transientSegmentName, SegmentType.TRANSIENT_SEGMENT, attributes, TIMEOUT)
+                .thenAccept(v -> {
+                    connection.send(new TransientSegmentCreated(createTransientSegment.getRequestId(), transientSegmentName));
+                })
+                .exceptionally(e -> handleException(createTransientSegment.getWriterId(),
+                        createTransientSegment.getRequestId(),
+                        transientSegmentName,
+                        operation,
+                        e));
+
+        LoggerHelpers.traceLeave(log, operation, traceId);
     }
 
 
@@ -367,11 +379,11 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
         }
     }
 
-    private void handleException(UUID writerId, long requestId, String segment, String doingWhat, Throwable u) {
-        handleException(writerId, requestId, segment, -1L, doingWhat, u);
+    private Void handleException(UUID writerId, long requestId, String segment, String doingWhat, Throwable u) {
+        return handleException(writerId, requestId, segment, -1L, doingWhat, u);
     }
 
-    private void handleException(UUID writerId, long requestId, String segment, long eventNumber, String doingWhat, Throwable u) {
+    private Void handleException(UUID writerId, long requestId, String segment, long eventNumber, String doingWhat, Throwable u) {
         if (u == null) {
             IllegalStateException exception = new IllegalStateException("No exception to handle.");
             log.error(requestId, "Append processor: Error {} on segment = '{}'", doingWhat, segment, exception);
@@ -417,6 +429,8 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
             logError(segment, u);
             connection.close(this::close); // Closing connection should reinitialize things, and hopefully fix the problem
         }
+
+        return null;
     }
 
     private void logError(String segment, Throwable u) {
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -102,6 +102,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
     private final boolean replyWithStackTraceOnError;
     private final ConcurrentHashMap<Pair<String, UUID>, WriterState> writerStates = new ConcurrentHashMap<>();
     private final ScheduledExecutorService tokenExpiryHandlerExecutor;
+    private final Collection<String> transientSegmentNames;
 
     //endregion
 
@@ -118,6 +119,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
         this.tokenVerifier = tokenVerifier;
         this.replyWithStackTraceOnError = replyWithStackTraceOnError;
         this.tokenExpiryHandlerExecutor = tokenExpiryHandlerExecutor;
+        this.transientSegmentNames = Collections.emptySet();
     }
 
     /**
@@ -442,6 +444,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
     @Override
     public void close() {
         connection.close();
+        transientSegmentNames.forEach(name -> store.deleteStreamSegment(name, TIMEOUT));
     }
 
     //endregion
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -66,6 +66,7 @@ import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.UUID;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
@@ -119,7 +120,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
         this.tokenVerifier = tokenVerifier;
         this.replyWithStackTraceOnError = replyWithStackTraceOnError;
         this.tokenExpiryHandlerExecutor = tokenExpiryHandlerExecutor;
-        this.transientSegmentNames = Collections.emptySet();
+        this.transientSegmentNames = new HashSet<>();
     }
 
     /**
@@ -272,7 +273,10 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
         );
         String transientSegmentName = NameUtils.getTransientNameFromId(createTransientSegment.getParentSegment(), createTransientSegment.getWriterId());
         store.createStreamSegment(transientSegmentName, SegmentType.TRANSIENT_SEGMENT, attributes, TIMEOUT)
-                .thenAccept(v -> connection.send(new TransientSegmentCreated(createTransientSegment.getRequestId(), transientSegmentName)))
+                .thenAccept(v -> {
+                    transientSegmentNames.add(transientSegmentName);
+                    connection.send(new TransientSegmentCreated(createTransientSegment.getRequestId(), transientSegmentName));
+                })
                 .exceptionally(e -> handleException(createTransientSegment.getWriterId(),
                         createTransientSegment.getRequestId(),
                         transientSegmentName,
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -448,7 +448,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
     @Override
     public void close() {
         connection.close();
-        // The AppendProccessor marks the tracked Transient Segments for deletion -- but does not synchronously wait for
+        // The AppendProcessor marks the tracked Transient Segments for deletion -- but does not synchronously wait for
         // the deletion to complete. The Transient Segemnts will be cleaned up at the SegmentStore's discretion.
         transientSegmentNames.forEach(name -> store.deleteStreamSegment(name, TIMEOUT));
     }
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -72,6 +72,7 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.pravega.shared.security.token.JsonWebToken;
 import lombok.Builder;
@@ -104,6 +105,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
     private final ConcurrentHashMap<Pair<String, UUID>, WriterState> writerStates = new ConcurrentHashMap<>();
     private final ScheduledExecutorService tokenExpiryHandlerExecutor;
     private final Collection<String> transientSegmentNames;
+    private final AtomicBoolean closed = new AtomicBoolean(false);
 
     //endregion
 
@@ -447,6 +449,17 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
 
     @Override
     public void close() {
+        if (closed.compareAndSet(false, true)) {
+            closeConnection();
+        }
+    }
+
+    @Override
+    public void connectionDropped() {
+        closeConnection();
+    }
+
+    private void closeConnection() {
         connection.close();
         // The AppendProcessor marks the tracked Transient Segments for deletion -- but does not synchronously wait for
         // the deletion to complete. The Transient Segemnts will be cleaned up at the SegmentStore's discretion.
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -120,7 +120,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
         this.tokenVerifier = tokenVerifier;
         this.replyWithStackTraceOnError = replyWithStackTraceOnError;
         this.tokenExpiryHandlerExecutor = tokenExpiryHandlerExecutor;
-        this.transientSegmentNames = new HashSet<>();
+        this.transientSegmentNames = Collections.synchronizedSet(new HashSet<>());
     }
 
     /**
@@ -448,6 +448,8 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
     @Override
     public void close() {
         connection.close();
+        // The AppendProccessor marks the tracked Transient Segments for deletion -- but does not synchronously wait for
+        // the deletion to complete. The Transient Segemnts will be cleaned up at the SegmentStore's discretion.
         transientSegmentNames.forEach(name -> store.deleteStreamSegment(name, TIMEOUT));
     }
 
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -450,20 +450,16 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
     @Override
     public void close() {
         if (closed.compareAndSet(false, true)) {
-            closeConnection();
+            connection.close();
+            // The AppendProcessor marks the tracked Transient Segments for deletion -- but does not synchronously wait for
+            // the deletion to complete. The Transient Segments will be cleaned up at the SegmentStore's discretion.
+            transientSegmentNames.forEach(name -> store.deleteStreamSegment(name, TIMEOUT));
         }
     }
 
     @Override
     public void connectionDropped() {
-        closeConnection();
-    }
-
-    private void closeConnection() {
-        connection.close();
-        // The AppendProcessor marks the tracked Transient Segments for deletion -- but does not synchronously wait for
-        // the deletion to complete. The Transient Segemnts will be cleaned up at the SegmentStore's discretion.
-        transientSegmentNames.forEach(name -> store.deleteStreamSegment(name, TIMEOUT));
+        close();
     }
 
     //endregion
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -269,6 +269,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
         );
         String parentSegment = createTransientSegment.getParentSegment();
         store.createStreamSegment(parentSegment, SegmentType.TRANSIENT_SEGMENT, attributes, TIMEOUT);
+        LoggerHelpers.traceLeave(log, "createTransientSegment", traceId);
     }
 
 
--- a/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
+++ b/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
@@ -143,7 +143,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
         connection.send(new Hello(WireCommands.WIRE_VERSION, WireCommands.OLDEST_COMPATIBLE_VERSION));
         if (hello.getLowVersion() > WireCommands.WIRE_VERSION || hello.getHighVersion() < WireCommands.OLDEST_COMPATIBLE_VERSION) {
             log.warn(hello.getRequestId(), "Incompatible wire protocol versions {} from connection {}", hello, connection);
-            connection.close(this::close);
+            close();
         }
     }
 
@@ -224,7 +224,7 @@ public class AppendProcessor extends DelegatingRequestProcessor implements AutoC
                     // and retrying the request with a new token.
                     log.debug("Closing client connection for writer {} due to token expiry, when processing " +
                                 "request {} for segment {}", writerId, requestId, segment);
-                    connection.close(this::close);
+                    this.close();
                 }
                 return null;
             }, token.durationToExpiry(), this.tokenExpiryHandlerExecutor);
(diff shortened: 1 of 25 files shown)


segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -408,5 +441,10 @@ private void logError(String segment, Throwable u) {
         }
     }
 
+    @Override
+    public void close() {
+        // Todo: Delete Transient Segments.
+    }
```
tkaitchuck, line 447: (This is a question, not a request for a change)
This ties the lifetime to the physical connection, rather than that of the writer / session. So on the client side if an

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -252,6 +260,29 @@ public void append(Append append) {
...
+    @Override
+    public void createTransientSegment(CreateTransientSegment createTransientSegment) {
+        String operation = "createTransientSegment";
+        long traceId = LoggerHelpers.traceEnter(log, operation, createTransientSegment);
+        Collection<AttributeUpdate> attributes = Arrays.asList(
+                new AttributeUpdate(CREATION_TIME, AttributeUpdateType.None, System.currentTimeMillis()),
+                new AttributeUpdate(ATTRIBUTE_SEGMENT_TYPE, AttributeUpdateType.None, SegmentType.TRANSIENT_SEGMENT.getValue())
+        );
+        String transientSegmentName = NameUtils.getTransientNameFromId(createTransientSegment.getParentSegment(), UUID.randomUUID());
+        store.createStreamSegment(transientSegmentName, SegmentType.TRANSIENT_SEGMENT, attributes, TIMEOUT)
```
tkaitchuck, line 272: This needs to be registered to be removed upon drop before the operations is invoked on the store.

tkaitchuck (changes requested):
I believe we will also need some recovery logic which special cases them.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -252,6 +260,29 @@ public void append(Append append) {
...
 
+    @Override
+    public void createTransientSegment(CreateTransientSegment createTransientSegment) {
+        String operation = "createTransientSegment";
+        long traceId = LoggerHelpers.traceEnter(log, operation, createTransientSegment);
+        Collection<AttributeUpdate> attributes = Arrays.asList(
+                new AttributeUpdate(CREATION_TIME, AttributeUpdateType.None, System.currentTimeMillis()),
+                new AttributeUpdate(ATTRIBUTE_SEGMENT_TYPE, AttributeUpdateType.None, SegmentType.TRANSIENT_SEGMENT.getValue())
+        );
+        String transientSegmentName = NameUtils.getTransientNameFromId(createTransientSegment.getParentSegment(), UUID.randomUUID());
```
andreipaduroiu, line 271: Let's pass in the `Writer Id` here and then use that and a random UUID (inside the NameUtils) to generate the name. This will help us in the future to identify which transient segments were created from which writers.

andreipaduroiu: @tkaitchuck Thoughts?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -408,5 +441,10 @@ private void logError(String segment, Throwable u) {
         }
     }
 
+    @Override
+    public void close() {
+        // Todo: Delete Transient Segments.
```
andreipaduroiu, line 446: This is a good step however it may not cover all cases. If the connection is correctly closed (for whatever reason), this may send a signal to delete the transient segment. However:
1. There is no guarantee that the "delete" will actually make it to the segment container and be processed.
2. The process may crash and this may not even be invoked at all.

It is for this reason that I wanted to leave this mechanism for a subsequent PR; we'd have to think about a "clean" way to check if a transient segment is still used from within the segment store.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/TrackedConnection.java
```diff
@@ -73,6 +78,11 @@ public void close() {
         this.connection.close();
     }
 
+    public void close(RunnableWithException callback) {
```
andreipaduroiu, line 81: 1. Please add a Javadoc explaining what this method does and what it does with the argument.
2. What if your argument is null?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/TrackedConnection.java
```diff
@@ -73,6 +78,11 @@ public void close() {
...
+        this.close();
+        Callbacks.invokeSafely(callback, ex -> log.error("Unable to run callback on TrackedConnection close.", ex));
```
andreipaduroiu, line 83: Tip: when logging errors, please include AS MUCH information as you can. The exception is mandatory, but is there anything else from this class that you may include? For example, we have a `toString` method that may return some useful info.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -74,12 +80,14 @@
...
 
+import static io.pravega.segmentstore.contracts.Attributes.ATTRIBUTE_SEGMENT_TYPE;
+import static io.pravega.segmentstore.contracts.Attributes.CREATION_TIME;
 import static io.pravega.segmentstore.contracts.Attributes.EVENT_COUNT;
 
 /**
  * Process incoming Append requests and write them to the SegmentStore.
  */
-public class AppendProcessor extends DelegatingRequestProcessor {
+public class AppendProcessor extends DelegatingRequestProcessor implements AutoCloseable {
```
andreipaduroiu, line 90: Make sure you update **every single unit test** that uses `AppendProcessor` to close it when done. See my comment below in `AppendProcessorTest`

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/AppendProcessorTest.java
```diff
@@ -161,6 +168,33 @@ public void testAppend() {
...
+        UUID writerId = UUID.randomUUID();
+        @Cleanup
+        ServiceBuilder serviceBuilder = newInlineExecutionInMemoryBuilder(getBuilderConfig());
+        serviceBuilder.initialize();
+        StreamSegmentStore store = serviceBuilder.createStreamSegmentService();
+
+        ServerConnection connection = mock(ServerConnection.class);
+        ConnectionTracker tracker = mock(ConnectionTracker.class);
+        val mockedRecorder = Mockito.mock(SegmentStatsRecorder.class);
+        AppendProcessor processor = AppendProcessor.defaultBuilder()
```
andreipaduroiu, line 184: Add `@Cleanup` to this instance (which is essentially a `try-with-resources+finally`) and to every other place in this file (even the ones you haven't touched) where `AppendProcessor` is instantiated.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/AppendProcessorTest.java
```diff
@@ -1014,6 +1048,23 @@ public void testAppendAfterSealThrows() throws Exception {
...
+        return ServiceBuilderConfig
+                .builder()
+                .include(ServiceConfig.builder()
+                        .with(ServiceConfig.CONTAINER_COUNT, 1)
+                        .with(ServiceConfig.THREAD_POOL_SIZE, 3)
+                        .with(ServiceConfig.LISTENING_PORT, TestUtils.getAvailableListenPort()))
+                .build();
+    }
+
+    private static ServiceBuilder newInlineExecutionInMemoryBuilder(ServiceBuilderConfig config) {
```
andreipaduroiu, line 1062: Every other test in this class mocks the segment store. I strongly recommend we continue following that pattern. Here, you're just testing the `AppendProcessor` and need not start the whole segment store for that.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -179,6 +180,26 @@ protected void doStop() {
...
+                .forEach(sm -> {
+                    // Todo: AppendProcessor - isTransientSegmentActive()?
+                });
+
+        LoggerHelpers.traceLeave(log, this.traceObjectId, "deleteUnusedTransientSegments", traceId);
+        return CompletableFuture.completedFuture(null);
+    }
+
+    private CompletableFuture<Void> runOnceInternal() {
+        return evictUnusedSegments().thenAccept(unused -> deleteUnusedTransientSegments());
```
andreipaduroiu, line 200: `thenCompose`

(`thenAccept` will just initiate, but not await `deleteUnusedTransientSemgents` to complete - please refer to the `CompletableFuture` JavaDoc for details).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -179,6 +180,26 @@ protected void doStop() {
                 }, this.executor);
     }
 
+    private CompletableFuture<Void> deleteUnusedTransientSegments() {
+        long lastSeqNo = this.lastIterationSequenceNumber.getAndSet(this.metadata.getOperationSequenceNumber());
```
andreipaduroiu, line 184: This has side effects. It's also invoked in `evictUnusedSegments` which means that likely the `MetadataCleaner` will not function correctly anymore. 

I think we may have to factor some of this code out into `runInternal` and then pass whatever we need as arguments to both these methods.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -179,6 +180,26 @@ protected void doStop() {
...
+        long traceId = LoggerHelpers.traceEnterWithContext(log, this.traceObjectId, "deleteUnusedTransientSegments", lastSeqNo);
+
+        // Serialize only those segments that are still alive (not deleted or merged - those will get removed anyway).
+        this.metadata.getEvictionCandidates(lastSeqNo, this.config.getMaxConcurrentSegmentEvictionCount())
+                .stream()
+                .filter(sm -> sm.getType() == SegmentType.TRANSIENT_SEGMENT && sm.isDeleted())
```
andreipaduroiu, line 190: You're getting all the deleted transient segments. Your code is not complete yet, but I assume you'll want to delete those, right? 

In that case, your filter should be `!sm.isDeleted()`.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -179,6 +180,26 @@ protected void doStop() {
...
+                .forEach(sm -> {
```
andreipaduroiu, line 191: This would be another filter.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -179,6 +180,26 @@ protected void doStop() {
...
+                    // Todo: AppendProcessor - isTransientSegmentActive()?
```
andreipaduroiu, line 192: You need to write some code to actually do something - like delete them using the `metadataStore`.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
```diff
@@ -141,6 +143,10 @@ public void close() {
      * </ul>
      */
     CompletableFuture<Void> createSegment(String segmentName, SegmentType segmentType, Collection<AttributeUpdate> attributes, Duration timeout) {
+        if (segmentType.isTransient() && !NameUtils.isTransientSegment(segmentName)) {
+            return Futures.failedFuture(new IllegalStateException("Invalid name '" + segmentName + "' for SegmentType: Transient"));
```
andreipaduroiu, line 147: `IllegalArgumentException` is more appropriate since it's an argument that's wrong.

`IllegalStateException` should be thrown when a method invocation is inconsistent with the current state of the object.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
```diff
@@ -149,7 +155,13 @@ public void close() {
         }
 
         ArrayView segmentInfo = SegmentInfo.serialize(SegmentInfo.newSegment(segmentName, segmentType, attributes));
-        CompletableFuture<Void> result = createSegment(segmentName, segmentInfo, new TimeoutTimer(timeout));
+        CompletableFuture<Void> result;
+        if (segmentType != SegmentType.TRANSIENT_SEGMENT) {
+            result = createSegment(segmentName, segmentInfo, new TimeoutTimer(timeout));
+        } else {
+            result = CompletableFuture.runAsync(() -> submitAssignment(newSegment(segmentName, segmentType, attributes), false, timeout));
```
andreipaduroiu, line 162: 1. Please factor this out into its own method.
2. Why `.runAsync`? Isn't `submitAssignment` returning a future? The way you wrote it, you'll return a CompletableFuture that will almost immediately complete - since all it does is initiate `submitAssignment` but not await its completion. Please check the Javadoc for `CompletableFuture` to understand how the methods you're using work.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
```diff
@@ -478,6 +480,9 @@ private void preProcessAttributes(AttributeUpdateCollection attributeUpdates)
             if (Attributes.isUnmodifiable(u.getAttributeId())) {
                 throw new MetadataUpdateException(this.containerId,
                         String.format("Attribute Id '%s' on Segment Id %s ('%s') may not be modified.", u.getAttributeId(), this.id, this.name));
+            } else if (NameUtils.isTransientSegment(this.name) && !Attributes.isCoreAttribute(u.getAttributeId())) {
```
andreipaduroiu, line 483: How come? We should allow extended attributes on transient segments. We should just make sure their total count doesn't exceed some small threshold we set.

shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
```diff
@@ -184,6 +199,22 @@ public static String getTransactionNameFromId(String parentStreamSegmentName, UU
...
     }
 
+    /**
+     * Returns the transient name for a TransientSegment based on the name of the current Parent StreamSegment, and the transientId.
+     *
+     * @param parentStreamSegmentName The name of the Parent StreamSegment for this transient segment.
+     * @param transientId             The unique Id for the transient segment.
+     * @return The name of the Transient StreamSegmentId.
+     */
+    public static String getTransientNameFromId(String parentStreamSegmentName, UUID transientId) {
```
andreipaduroiu, line 209: Did you add a unit test to `SegmentToContainerMapper` to validate that the transient segment name will map to the exact same container id as its parent name? There is code in it to do that for transaction segments; we need to make sure the same holds for transient segments too.

andreipaduroiu (changes requested):
We're heading in the right direction, however we need a lot more unit tests to verify all the functionality you have added and all the corner cases that arise from there.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
co-jo: 1. Okay
2. It is, but is returning a `Long` and not `Void`. I didn't believe you could directly cast the types.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
co-jo: Was working on the changes after I submitted the PR.. fixing.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/AppendProcessorTest.java
co-jo: I was trying to mock it out first but the connection would always fail to verify the `TransientSegmentCreated` was sent.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
co-jo: Sorry yes, thats right.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -44,6 +43,8 @@
 class MetadataCleaner extends AbstractThreadPoolService {
     //region Private
 
+    static final Duration TIMEOUT = Duration.ofMinutes(1);
```
andreipaduroiu, line 46: What is this timeout for? If you need it for some operation below, please see if there's something available in `ContainerConfig` (and if not, let's add it in there).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -180,24 +178,24 @@ protected void doStop() {
...
-
         // Serialize only those segments that are still alive (not deleted or merged - those will get removed anyway).
         this.metadata.getEvictionCandidates(lastSeqNo, this.config.getMaxConcurrentSegmentEvictionCount())
                 .stream()
-                .filter(sm -> sm.getType() == SegmentType.TRANSIENT_SEGMENT && sm.isDeleted())
+                .filter(sm -> sm.getType().isTransientSegment() && !sm.isDeleted())
+                //.filter(sm -> appendProcessor.isTransientSegmentActive(sm.getName()))
                 .forEach(sm -> {
-                    // Todo: AppendProcessor - isTransientSegmentActive()?
+                    metadataStore.deleteSegment(sm.getName(), TIMEOUT);
```
andreipaduroiu, line 189: This is not awaiting the completion. You are just initiating these delete operations.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
```diff
@@ -455,6 +461,28 @@ void preProcessAsSourceSegment(MergeSegmentOperation operation) throws StreamSeg
...
+    /**
+     * Returns the total number of unique ExtendedAttributes that would exist for the segment assuming the
+     * {@link AttributeUpdateCollection} would be applied.
+     *
+     * @param attributeUpdates The {@link AttributeUpdateCollection} to apply.
+     * @return The number of unique extended attributes.
+     */
+    private int getUniqueExtendedAttributesCount(AttributeUpdateCollection attributeUpdates) {
+        int count = 0;
+        for (Map.Entry<AttributeId, Long> attribute : baseAttributeValues.entrySet()) {
```
andreipaduroiu, line 473: To speed things up, I suggest we just get the size of the `baseAttributeValues`. Even though that's going to contain core attributes as well, this code is on the hot path for every single append in the segment store so we must do everything we can to make it as fast as possible. We can always increase our attribute limit to 32 and use `.size()` here to approximate the number of attributes.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
```diff
@@ -455,6 +461,28 @@ void preProcessAsSourceSegment(MergeSegmentOperation operation) throws StreamSeg
...
+            count += !Attributes.isCoreAttribute(attribute.getKey()) ? 1 : 0;
+        }
+        Collection<AttributeId> unique = new HashSet<>();
+        for (AttributeUpdate attribute : attributeUpdates) {
```
andreipaduroiu, line 477: AttributeUpdates are not duplicated. There's no reason to check for uniqueness. I think we can get rid of the `unique` variable.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/SegmentMetadataUpdateTransaction.java
```diff
@@ -474,15 +502,18 @@ private void preProcessAttributes(AttributeUpdateCollection attributeUpdates)
             return;
         }
 
+        // Make sure that the number of existing ExtendedAttributes + incoming ExtendedAttributes does not exceed the limit.
+        if (NameUtils.isTransientSegment(this.name) && getUniqueExtendedAttributesCount(attributeUpdates) > TRANSIENT_ATTRIBUTE_LIMIT) {
```
andreipaduroiu, line 506: `NameUtils.isTransientSegment` is  very expensive method to run here. Use `metadata.getType().isTransientSegment`

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/MetadataStoreTestBase.java
```diff
@@ -741,6 +741,11 @@ public void testTransientSegmentFormat() throws ExecutionException, InterruptedE
         );
     }
 
+    @Test
+    public void testTransientSegmentCleanup() {
+
```
andreipaduroiu, line 746: Please implement this test.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/StorageWriterTests.java
```diff
@@ -747,6 +746,14 @@ public void testForceFlush() throws Exception {
         Assert.assertFalse("Not expected anything to be flushed the second time.", result2);
     }
 
+    /**
+     * Verifies that the StorageWriter does not create an Attribute Aggregator for Transient Segments.
+     */
+    @Test
+    public void testNoAttributeAggregatorTransientSegments() {
+
```
co-jo, line 754: Todo...

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -110,6 +120,7 @@
         this.tokenVerifier = tokenVerifier;
         this.replyWithStackTraceOnError = replyWithStackTraceOnError;
         this.tokenExpiryHandlerExecutor = tokenExpiryHandlerExecutor;
+        this.transientSegmentNames = new HashSet<>();
```
andreipaduroiu, line 123: This is not thread safe. Wrap it in `Collections.synchronizedSet`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
andreipaduroiu: @tkaitchuck This work should be done now. We need to validate that upon a connection close, the `AppendProcessor.close` method is also invoked.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AppendProcessor.java
```diff
@@ -408,5 +445,11 @@ private void logError(String segment, Throwable u) {
         }
     }
 
+    @Override
+    public void close() {
+        connection.close();
+        transientSegmentNames.forEach(name -> store.deleteStreamSegment(name, TIMEOUT));
```
andreipaduroiu, line 451: Let's put a comment next to this line explaining that we are purposefully not awaiting the completion of this and that the Segment store will eventually delete unused transient segments after some time (at its discretion).

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/ContainerConfig.java
```diff
@@ -40,6 +40,7 @@
     public static final Property<Integer> MAX_CACHED_EXTENDED_ATTRIBUTE_COUNT = Property.named("extended.attribute.cached.count.max", 4096, "maxCachedExtendedAttributeCount");
     public static final Property<Integer> EVENT_PROCESSOR_ITERATION_DELAY_MS = Property.named("eventprocessor.iteration.delay.ms", 100);
     public static final Property<Integer> EVENT_PROCESSOR_OPERATION_TIMEOUT_MS = Property.named("eventprocessor.operation.timeout.ms", 5000);
+    public static final Property<Integer> TRANSIENT_SEGMENT_DELETE_TIMEOUT_MS = Property.named("segment.transient.delete.timeout.ms", 1000);
```
andreipaduroiu, line 43: I believe this is way too aggressive. If the segment store is busy, it's quite likely that requests may take longer than 1s to complete. I think 10 or 30 seconds would be a reasonable default.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataCleaner.java
```diff
@@ -179,6 +177,27 @@ protected void doStop() {
...
+        // Serialize only those segments that are still alive (not deleted or merged - those will get removed anyway).
+        val deletionTasks = this.metadata.getEvictionCandidates(lastSeqNo, this.config.getMaxConcurrentSegmentEvictionCount())
+                .stream()
+                .filter(sm -> sm.getType().isTransientSegment())
+                .filter(sm -> sm.getAttributes().get(Attributes.CREATION_EPOCH) < metadata.getContainerEpoch())
+                .map(sm -> metadataStore.deleteSegment(sm.getName(), this.config.getTransientSegmentDeleteTimeout()))
+                .collect(Collectors.toList());
+
+        return Futures.allOf(deletionTasks)
+                .thenRunAsync(() -> {
```
andreipaduroiu, line 191: Performance tip: `TRACE` is unlikely to be used at all, and guaranteed never to be used in production. This means that you'll be essentially creating a new callback and queuing it onto the executor every time you invoke this method even if it won't do anything. This has the potential of slowing down the system.

Consider doing this:
1. create a local variable `result` to hold the result of `Futures.allOf`
2. `if(log.isTraceEnabled())` then attach this callback to result (but do not include it in the return value. I.e., just do `result.thenRun`
3. Notice how I didn't even do `thenRunAsync`. Since it's such a small callback, no need to re-queue it on the executor; just run it inline. 
4. Return `result`.

For examples, search for `.isTraceEnabled` or `.isDebugEnabled` in the codebase; we've used this pattern in other places too.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
```diff
@@ -141,6 +144,10 @@ public void close() {
      * </ul>
      */
     CompletableFuture<Void> createSegment(String segmentName, SegmentType segmentType, Collection<AttributeUpdate> attributes, Duration timeout) {
+        if (segmentType.isTransientSegment() && !NameUtils.isTransientSegment(segmentName)) {
+            return Futures.failedFuture(new IllegalArgumentException("Invalid name '" + segmentName + "' for SegmentType: Transient"));
```
andreipaduroiu, line 148: tip 1: Use `String.format`
tip 2: don't hardcode `Transient`; use `segmentType.toString`(or if you use `String.format`, omit the `toString` as it will automatically invoke it.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
```diff
@@ -167,6 +177,20 @@ public void close() {
...
+    /**
+     * Creates a new Transient Segment with given name.
+     *
+     * @param segmentName The case-sensitive Segment Name.
+     * @param attributes  The initial attributes for the StreamSegment, if any.
+     * @param timeout     Timeout for the operation.
+     * @return A CompletableFuture that, when completed normally, will indicate the TransientSegment has been created.
+     */
+    private CompletableFuture<Void> createTransientSegment(String segmentName, Collection<AttributeUpdate> attributes, Duration timeout) {
+        Collection<AttributeUpdate> attrs = new ArrayList<>(attributes);
```
andreipaduroiu, line 189: Tip: Consider using `AttributeUpdateCollection`; it does the same thing (you can create one from a list and then you can add arbitrary ones to it). Plus it also implements `Collection<AttributeUpdate`

segmentstore/server/src/main/java/io/pravega/segmentstore/server/containers/MetadataStore.java
```diff
@@ -167,6 +177,20 @@ public void close() {
...
+        attrs.add(new AttributeUpdate(Attributes.CREATION_EPOCH, AttributeUpdateType.None, this.connector.containerMetadata.getContainerEpoch()));
+        return Futures.toVoid(submitAssignment(newSegment(segmentName, SegmentType.TRANSIENT_SEGMENT, attrs), false, timeout));
```
andreipaduroiu, line 191: I recommend `submitAssignmentWithRetry`. We may have reached the maximum number of active segments (there is such a limit ~25,000 per container) and that method will attempt to evict some if it needs to make room. If you don't invoke that, your method will immediately fail if the container is full.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/containers/MetadataStoreTestBase.java
```diff
@@ -718,10 +743,42 @@ public void testRegisterPinnedSegment() {
...
+        final String validTransientSegment = "scope/stream/transient#transient.00000000000000000000000000000000";
+        final String invalidTransientSegment = "scope/stream/transient";
+        @Cleanup
+        TestContext context = createTestContext();
+        // Should complete successfully.
+        context.getMetadataStore().createSegment(validTransientSegment, SegmentType.TRANSIENT_SEGMENT, new ArrayList<>(), TIMEOUT).get();
+        // Make sure our Transient Segment name maps to a valid Segment ID.
+        TestUtils.await(() -> {
+            return context.getMetadata().getStreamSegmentId(validTransientSegment, false) != ContainerMetadata.NO_STREAM_SEGMENT_ID;
+        }, 100, 1000);
```
andreipaduroiu, line 765: Give this thing a bit more timeout. In GitHubActions, this may take more than 1s due to other things running on the machine. 10-30 seconds timeout should be good and will ensure that your test doesn't fail sporadically.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/StorageWriterTests.java
andreipaduroiu, line 754: Were you planning to write a test here?

shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
```diff
@@ -185,20 +200,60 @@ public static String getTransactionNameFromId(String parentStreamSegmentName, UU
...
-     * valid value only if the Transaction StreamSegmentName was generated using the generateTransactionStreamSegmentName method.
+     * Returns the transient name for a TransientSegment based on the name of the current Parent StreamSegment, and the transientId.
      *
-     * @param transactionName The name of the Transaction StreamSegment to extract the name of the Parent StreamSegment.
+     * @param parentStreamSegmentName The name of the Parent StreamSegment for this transient segment.
+     * @param writerId The Writer Id used to create the transient segment.
+     * @return The name of the Transient StreamSegmentId.
+     */
+    public static String getTransientNameFromId(String parentStreamSegmentName, UUID writerId) {
+        UUID random = UUID.randomUUID();
```
andreipaduroiu, line 210: Let's also write a test to make sure that transient segment names and transaction segment names are not confused by the new code or by the existing code (the one for transactions).

andreipaduroiu (changes requested):
Looking better. I left a few more comments.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/StorageWriterTests.java
co-jo: I did attempt to implement this test but wasn't really happy with the approaches I tried.. 

1. This first approach was to validate this by checking for side effects. I.E. add some Attributes during segment creation and check that core attributes are added to the attribute segment but extended attributes are not. The issue with this was it would take (potentially?) significant modifications to the `TestWriterDataSource` as the attribute segments are not created at the moment.
2. The second approach was to just directly check that there is no object of class `AttributeAggregator` in the `ProcessorCollection`. The problem with this was it requires changing the visibility of certain fields. Wasn't sure if its okay just to add a `VisibileForTesting` annotation.

So I just left it empty to cycle back to it if I thought of something better.

shared/protocol/src/main/java/io/pravega/shared/NameUtils.java
```diff
@@ -227,6 +228,7 @@ public static String getTransientNameFromId(String parentStreamSegmentName, UUID
      *
      * @return The start position of the delimiter contained within streamSegmentName.
      */
+    @NonNull
```
andreipaduroiu, line 231: I'm not sure what this annotation would do here. You're annotating the method, but the method can't possibly return null.

tkaitchuck (approved):
Looks good. Assuming missing test is added.

segmentstore/server/src/test/java/io/pravega/segmentstore/server/writer/StorageWriterTests.java
tkaitchuck, line 754: This is the test I am referring to.

PR #6172, opened 2021-07-22
Issue 5560: Read the contents of a segment from the storage

### LINKED ISSUE #5560: Read the contents of a segment from the storage
**Problem description**
Sometimes, it is required to read the contents of a segment from the LTS. The is useful in debugging especially.  

**Problem location**
CLI command.

Develop a CLI command to read the contents of a segment.

anirudhkovuru:
**Change log description**  
The CLI command `segmentstore read-segment` has been tweaked to write the obtained segment data into a specified file. The file will contain a byte array. 

**Purpose of the change**  
Fixes #5560 

**What the code does**  
- `ReadSegmentRangeCommand` now has a new argument pertaining to the filename where the segment byte array will be written.
- If the file exists then a `FileAlreadyExistsException` is thrown to ensure that files are not accidentally overwritten by the user using the command.
- The contents are written into the file using the `java.io.FileOutputStream` class.

**How to verify it**  
All tests must pass.

### DIFF
--- a/cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
+++ b/cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
@@ -26,6 +26,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileAlreadyExistsException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -62,13 +63,15 @@ public class ReadSegmentRangeCommand extends SegmentStoreCommand {
         WireCommands.SegmentRead segmentRead = reply.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
 
         File f = new File(fileName);
-        // Creating the file if it doesn't exist.
-        if (!f.exists()) {
-            if (!f.getParentFile().exists()) {
-                f.getParentFile().mkdirs();
-            }
-            f.createNewFile();
+        // If file exists throw FileAlreadyExistsException, an existing file should not be overwritten with new data.
+        if (f.exists()) {
+            throw new FileAlreadyExistsException("Cannot write segment data into a file that already exists.");
         }
+
+        if (!f.getParentFile().exists()) {
+            f.getParentFile().mkdirs();
+        }
+        f.createNewFile();
 
         // Write data into the file.
         @Cleanup
--- a/cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
+++ b/cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
@@ -26,14 +26,11 @@ import org.apache.curator.framework.CuratorFramework;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.nio.file.FileAlreadyExistsException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-
-import static java.lang.Math.min;
 
 public class ReadSegmentRangeCommand extends SegmentStoreCommand {
 
@@ -66,9 +63,8 @@ public class ReadSegmentRangeCommand extends SegmentStoreCommand {
         CuratorFramework zkClient = createZKClient();
         @Cleanup
         SegmentHelper segmentHelper = instantiateSegmentHelper(zkClient);
-        File file = createFileAndDirectory(fileName);
-        readAndWriteSegmentToFile(segmentHelper, segmentStoreHost, fullyQualifiedSegmentName, offset, length, file);
-        output("The segment data has been successfully written into %s", fileName);
+        readAndWriteSegmentToFile(segmentHelper, segmentStoreHost, fullyQualifiedSegmentName, offset, length, fileName);
+        output("\nThe segment data has been successfully written into %s", fileName);
     }
 
     /**
@@ -100,30 +96,40 @@ public class ReadSegmentRangeCommand extends SegmentStoreCommand {
      * @param fullyQualifiedSegmentName The name of the segment.
      * @param offset                    The starting point from where the segment is to be read.
      * @param length                    The number of bytes to read.
-     * @param file                      A {@link File} object referring to the file in which the data will be written.
-     * @throws IOException if the file write fails.
+     * @param fileName                  A name of the file to which the data will be written.
+     * @throws IOException if the file create/write fails.
      * @throws InterruptedException if the request fails.
      * @throws ExecutionException if the request fails.
      * @throws TimeoutException if the request fails.
      */
     private void readAndWriteSegmentToFile(SegmentHelper segmentHelper, String segmentStoreHost, String fullyQualifiedSegmentName,
-                                           int offset, int length, File file) throws IOException, InterruptedException, ExecutionException, TimeoutException {
+                                           int offset, int length, String fileName) throws IOException, InterruptedException, ExecutionException, TimeoutException {
+        File file = createFileAndDirectory(fileName);
+
+        output("Downloading %d bytes from offset %d into %s.", length, offset, fileName);
         int currentOffset = offset;
         int bytesToRead = length;
+        int progress = 0;
+
         while (bytesToRead > 0) {
-            int bufferLength = min(READ_WRITE_BUFFER_SIZE, bytesToRead);
+            int bufferLength = Math.min(READ_WRITE_BUFFER_SIZE, bytesToRead);
             CompletableFuture<WireCommands.SegmentRead> reply = segmentHelper.readSegment(fullyQualifiedSegmentName,
                     currentOffset, bufferLength, new PravegaNodeUri(segmentStoreHost, getServiceConfig().getAdminGatewayPort()), "");
             WireCommands.SegmentRead bufferRead = reply.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
-
             // Write the buffer into the file.
             try (FileOutputStream fileOutputStream = new FileOutputStream(file, true)) {
                 bufferRead.getData().readBytes(fileOutputStream, bufferLength);
             }
 
-            currentOffset += READ_WRITE_BUFFER_SIZE;
-            bytesToRead -= READ_WRITE_BUFFER_SIZE;
+            currentOffset += bufferLength;
+            bytesToRead -= bufferLength;
+            showProgress(progress++, String.format("Written %d/%d bytes.", length - bytesToRead, length));
         }
+    }
+
+    private void showProgress(int progress, String message) {
+        String status = "|/-\\";
+        System.out.print("\r Processing " + status.charAt(progress % status.length()) + " : " + message);
     }
 
     public static CommandDescriptor descriptor() {
--- a/cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
+++ b/cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
@@ -28,14 +28,13 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.FileAlreadyExistsException;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 public class ReadSegmentRangeCommand extends SegmentStoreCommand {
 
     private static final int REQUEST_TIMEOUT_SECONDS = 10;
     private static final int READ_WRITE_BUFFER_SIZE = 2 * 1024 * 1024;
+    private static final String PROGRESS_BAR = "|/-\\";
 
     /**
      * Creates a new instance of the ReadSegmentRangeCommand.
@@ -47,12 +46,12 @@ public class ReadSegmentRangeCommand extends SegmentStoreCommand {
     }
 
     @Override
-    public void execute() throws ExecutionException, InterruptedException, TimeoutException, IOException {
+    public void execute() throws Exception {
         ensureArgCount(5);
 
         final String fullyQualifiedSegmentName = getArg(0);
-        final int offset = getIntArg(1);
-        final int length = getIntArg(2);
+        final long offset = getLongArg(1);
+        final long length = getLongArg(2);
         final String segmentStoreHost = getArg(3);
         final String fileName = getArg(4);
 
@@ -70,7 +69,7 @@ public class ReadSegmentRangeCommand extends SegmentStoreCommand {
     /**
      * Creates the file (and parent directory if required) into which thee segment data is written.
      *
-     * @param fileName
+     * @param fileName The name of the file to create.
      * @return A {@link File} object representing the filename provided.
      * @throws FileAlreadyExistsException if the file already exists, to avoid any accidental overwrites.
      * @throws IOException if the file/directory creation fails.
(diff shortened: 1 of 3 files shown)


andreipaduroiu:
@anirudhkovuru Let's add some unit tests and figure out why the build is broken.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +45,43 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
+            throw new FileAlreadyExistsException("Cannot write segment data into a file that already exists.");
+        }
+        if (!f.getParentFile().exists()) {
+            f.getParentFile().mkdirs();
+        }
+        f.createNewFile();
+        // Write data into the file.
+        @Cleanup
+        FileOutputStream fileStream = new FileOutputStream(f);
+        fileStream.write(segmentRead.getData().array());
```
andreipaduroiu, line 75: `.array()` will give you the array over which this `ArrayView/ByteArraySegment` is built upon, however that may be more than needed. An `ArrayView` is a slice of a `byte[]`, so it has a start offset and length. You are therefore possibly writing more than requested (you may include other unrelated stuff that's on that array).

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +45,43 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
         final int length = getIntArg(2);
         final String segmentStoreHost = getArg(3);
+        final String fileName = getArg(4);
         @Cleanup
         CuratorFramework zkClient = createZKClient();
         @Cleanup
         SegmentHelper segmentHelper = instantiateSegmentHelper(zkClient);
         CompletableFuture<WireCommands.SegmentRead> reply = segmentHelper.readSegment(fullyQualifiedSegmentName,
                 offset, length, new PravegaNodeUri(segmentStoreHost, getServiceConfig().getAdminGatewayPort()), "");
         WireCommands.SegmentRead segmentRead = reply.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
andreipaduroiu, line 62: This is a good start, however it will not work for more than 16MB at a time. There is a limit of 16MB that we can send over the wire at any given time (Netty frame size), so we'll need to make this handle larger sizes as well.

Consider doing this:
1. Open/Create the file
2. Enter a loop; at each iteration request a maximum of 8MB (half of 16MB). Each buffer you receive back will be written at the end of that file.
3. Loop until you have read all the range that you were asked for in the arguments.

This approach should also enable us to download more than 2GB at once (MaxInt). 

Tip: For a good user experience, consider outputting some information to the console every time you downloaded 8MB (every iteration). For example (I take this from `jhat`), output a dot (at once; do not `println`) for every 8MB. 

Example:
```
Downloading 16000001 bytes from offset 0. Expect 3 dots:
...
Wrote 16000001 bytes to /tmp/myfile.
```

In the above example, you have 3 dots because you break 16000001 into 3 parts: [0, 8000000), [8000000, 16000000) and [16000000, 16000001) (8MB+8MB+1B).

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +45,43 @@ public ReadSegmentRangeCommand(CommandArgs args) {
     }
 
     @Override
-    public void execute() throws ExecutionException, InterruptedException, TimeoutException {
-        ensureArgCount(4);
+    public void execute() throws ExecutionException, InterruptedException, TimeoutException, IOException {
+        ensureArgCount(5);
 
         final String fullyQualifiedSegmentName = getArg(0);
         final int offset = getIntArg(1);
```
andreipaduroiu, line 52: Please convert both of these into `long`. We definitely have offsets greater than 2GB and (see my comment below), the user may want to download more than 2GB at once.

andreipaduroiu (changes requested):

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +45,43 @@ public ReadSegmentRangeCommand(CommandArgs args) {
     }
 
     @Override
-    public void execute() throws ExecutionException, InterruptedException, TimeoutException {
-        ensureArgCount(4);
+    public void execute() throws ExecutionException, InterruptedException, TimeoutException, IOException {
+        ensureArgCount(5);
```
RaulGracia, line 49: Please, add some precondition checks on the offset/length arguments (e.g., offset >= 0, length >=0)

RaulGracia (changes requested):

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: The `SegmentHelper.readSegment` call takes the offset and length as integers. While it makes sense to take them as `long` since as described below the read will be as follows (offset, length).
```
[offset, 8000000] -> [offset + 8000000, 8000000] -> [offset + 1600000, 8000000] -> ........
```

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: `readSegment` has a `MAX_READ_SIZE` of 2MB. As we see here:
https://github.com/pravega/pravega/blob/f969bc907ad59bcef3387ae27021a08c57833502/segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java#L210e
So we could reduce the buffer size.

anirudhkovuru: Also to add, `readSegment` call takes the offset and length as integers. While it makes sense to take them as `long` since as described the read will be as follows (offset, length).
```
(offset, 2000000) -> (offset + 2000000, 2000000) -> (offset + 400000, 2000000) -> ........
```
The length is fixed and shouldn't be an issue but since the offset can keep growing, should we change the offset argument in `ReadSegment` to a `long` type?

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Added the precondition checks.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: I can add a TODO referencing this: https://github.com/pravega/pravega/issues/6183

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: I've implemented this approach and for the user info to be outputted I've added a byte counter of sorts which shows the amount data that has been processed and the total that's requested in this format:
```
Processing \ : Written 8000/160001 bytes.
```

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Instead of `.array()` I use `.readBytes()` to directly write it into the `FileOutputStream`.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
andreipaduroiu: OK. Let's do the following:
1. Please add a TODO and reference that other ticket. And please make a note in 6183 to fix this usage as well - actually add a note to check all usages and fix them as needed.
2. Do change these (both) to `long` (including `length` as we may want to download more than 2GB at once). In your API call to `SegmentHelper` add a cast to int (`(int)`) and write your TODO there. That way when 6183 is actually fixed all we need to do is remove the cast and be done.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +47,97 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
-        output("ReadSegment: %s", segmentRead.toString());
-        output("SegmentRead content: %s", segmentRead.getData().toString(StandardCharsets.UTF_8));
+        readAndWriteSegmentToFile(segmentHelper, segmentStoreHost, fullyQualifiedSegmentName, offset, length, fileName);
+        output("\nThe segment data has been successfully written into %s", fileName);
+    }
+
+    /**
+     * Creates the file (and parent directory if required) into which thee segment data is written.
+     *
+     * @param fileName
```
andreipaduroiu, line 73: Please add Javadod for this param.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +47,97 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
+    /**
+     * Reads the contents of the segment starting from the given offset and writes into the provided file.
+     *
+     * @param segmentHelper             A {@link SegmentHelper} instance to read the segment.
+     * @param segmentStoreHost          Address of the segment-store to read from.
+     * @param fullyQualifiedSegmentName The name of the segment.
+     * @param offset                    The starting point from where the segment is to be read.
+     * @param length                    The number of bytes to read.
+     * @param fileName                  A name of the file to which the data will be written.
+     * @throws IOException if the file create/write fails.
```
andreipaduroiu, line 100: If you have this many exceptions, I may recommend you just say `throws Exception` and then in your Javadoc you only talk about that one.  There is no useful information that someone might get from this Javadoc (as currently written) if all the exceptions have "if the request fails". The point of separating exceptions out is for the caller to write handling code (i.e., `try-catch`) specifically for those exceptions and figure out what to do in each case.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +47,97 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
+     * @throws InterruptedException if the request fails.
+     * @throws ExecutionException if the request fails.
+     * @throws TimeoutException if the request fails.
+     */
+    private void readAndWriteSegmentToFile(SegmentHelper segmentHelper, String segmentStoreHost, String fullyQualifiedSegmentName,
+                                           int offset, int length, String fileName) throws IOException, InterruptedException, ExecutionException, TimeoutException {
```
andreipaduroiu, line 106: `long` for `offset` and `length`

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +47,97 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
+        int bytesToRead = length;
+        int progress = 0;
+
+        while (bytesToRead > 0) {
+            int bufferLength = Math.min(READ_WRITE_BUFFER_SIZE, bytesToRead);
+            CompletableFuture<WireCommands.SegmentRead> reply = segmentHelper.readSegment(fullyQualifiedSegmentName,
+                    currentOffset, bufferLength, new PravegaNodeUri(segmentStoreHost, getServiceConfig().getAdminGatewayPort()), "");
+            WireCommands.SegmentRead bufferRead = reply.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            // Write the buffer into the file.
+            try (FileOutputStream fileOutputStream = new FileOutputStream(file, true)) {
```
andreipaduroiu, line 120: You don't need to open the `FileOutputStream` in every iteration; open it outside of the `while` loop; you can always reuse it - and make your code more efficient at the same time.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +47,97 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
+                bufferRead.getData().readBytes(fileOutputStream, bufferLength);
+            }
+
+            currentOffset += bufferLength;
```
andreipaduroiu, line 124: You should only subtract the amount of data that you read. The Contract for `SegmentRead` is that it will read AT MOST the number of bytes you asked it to. So you need to get the number of bytes that came back and only account for those, not what you actually requested.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -42,29 +47,97 @@ public ReadSegmentRangeCommand(CommandArgs args) {
...
+            bytesToRead -= bufferLength;
+            showProgress(progress++, String.format("Written %d/%d bytes.", length - bytesToRead, length));
+        }
+    }
+
+    private void showProgress(int progress, String message) {
+        String status = "|/-\\";
```
andreipaduroiu, line 131: This can be a class-level constant.

cli/admin/src/test/java/io/pravega/cli/admin/segmentstore/SegmentStoreCommandsTest.java
```diff
@@ -60,6 +64,7 @@ public void testGetSegmentInfoCommand() throws Exception {
 
     @Test
     public void testReadSegmentRangeCommand() throws Exception {
+        String filename = "./tmp/readSegmentTest.txt";
```
andreipaduroiu, line 67: Use `File.createTempFile`. You are creating a file in a `tmp` directory in the current path, which may not be writeable at all times (depending on where you run this).

andreipaduroiu (changes requested):

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Added the TODO referencing #6183 . Made them `long` and typecast to `int`. We only need to remove the cast when that fix goes in.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Have added javadoc.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Shortened all the exceptions in regards to the failure of the request into a single `throws Exception`. This leaves the javadoc going into two possible exceptions here:

1. `IOException` when anything happens in regards to file creation and handling.
2. `Exception` for any other failures that can be thrown while making the request.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Changed to `long`

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Opened the `FileOutputStream` outside the loop with an `@Cleanup` annotation.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Getting the length **actually** read from the segment using `ByteBuf.readablebytes()` which is used to update the offset and bytes to be read.

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
anirudhkovuru: Made it a class level constant

cli/admin/src/test/java/io/pravega/cli/admin/segmentstore/SegmentStoreCommandsTest.java
anirudhkovuru: Using `File.createTempDirectory()` to create a temporary directory where the test can create, write and delete files.

andreipaduroiu (approved):

RaulGracia (approved):

PR #6185, opened 2021-07-28
Issue 6184: SLTS - Fix generation of inconsistent snapshots

### LINKED ISSUE #6184: SLTS - SystemJournal - Subtle concurrency issue leads to inconsistent metadata during snapshot creation.
**Describe the bug**

- _Background_ - 
  - SLTS has built-in transaction system that is lock-free and optimized for specific SLTS scenarios and takes advantage of very specific SLTS behavioral invariants. 
  - One of such invariants is that operations on the same segment are always linearized. In other words, in SLTS we schedule  operations in such a way that it avoids concurrent metadata modifications for any segment and as such for a given segment we'll always have consistent metadata view for any properly scheduled operation. This allows lock-free implementation.
 
- Specific issue 
  - However when system snapshot is created we read metadata about 4 specific storage/container metadata segments and persist it in a snapshot. 
  - While both writes to journals are linearized and also operations on individual storage system segments are linearized , they are NOT linearized _as a group_. 
  - This opens up small possibility that inconsistent/different versions of metadata are seen by two different operations on same set of system segments. Eg. reading metadata for creating snapshot while metadata commit for write operation is in progress for one of the system segment
  
- Suggested fix - Always linearize on system segments as a group.
 
**To Reproduce**

1. Run any tests derived from BookKeeperIntegrationTestBase in a loop. 
2. Either set a breakpoint or observe preconditions fails in  log
3. Eventually precondition fails with either datasize or chunk count does not match.

**Screenshots**
A few error logs 

```
WARN i.p.s.s.h.ZKSegmentContainerMonitor - Starting container 4 failed: {}
java.util.concurrent.CompletionException: java.lang.IllegalStateException: chunkCount should be 1. SegmentMetadata(name=_system/containers/metadata_4, length=0, chunkCount=156, startOffset=0, status=25, maxRollinglength=134217728, firstChunk=_system/containers/metadata_4.E-2-O-0.c2328e78-0205-4416-b958-2eee563b01d6, lastChunk=_system/containers/metadata_4.E-2-O-0.c2328e78-0205-4416-b958-2eee563b01d6, lastModified=0, firstChunkStartOffset=0, lastChunkStartOffset=0, ownerEpoch=16825)
 at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source)
 at java.base/java.util.concurrent.CompletableFuture.completeThrowable(Unknown Source)
 at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(Unknown Source)
```

**Additional information**
Add any other information about the problem here, including the root cause if known.

sachin-j-joshi:

**Change log description**  
SLTS - SystemJournal - Fix concurrency issue with inconsistent metadata during snapshot creation.

**Purpose of the change**  
Fix #6184 

**What the code does**  
Linearizes operations on storage system segments as a group , instead of linearizing them individually.

**How to verify it**  
All tests must pass.
No presence of any assert/precondition failures related with snapshots


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
```diff
@@ -754,7 +754,12 @@ void reportMetricsForSystemSegment(SegmentMetadata segmentMetadata) {
      * */
     private <R> CompletableFuture<R> executeSerialized(Callable<CompletableFuture<R>> operation, String... segmentNames) {
         Exceptions.checkNotClosed(this.closed.get(), this);
-        return this.taskProcessor.add(Arrays.asList(segmentNames), () -> executeExclusive(operation, segmentNames));
+        if (segmentNames.length == 1 && this.systemJournal.isStorageSystemSegment(segmentNames[0])) {
+            val segments = this.systemJournal.getSystemSegments();
```
andreipaduroiu, line 758: To make sure I understand this; if the only segment we have is a system segment, then we linearize on ALL the system segments? This looks a bit counterintuitive - would you mind adding a comment here explaining why so that don't accidentally undo this at some time in the future?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -228,7 +228,7 @@ public SystemJournal(int containerId, ChunkStorage chunkStorage, ChunkMetadataSt
         this.garbageCollector = Preconditions.checkNotNull(garbageCollector, "garbageCollector");
         this.containerId = containerId;
         this.systemSegments = getChunkStorageSystemSegments(containerId);
-        this.systemSegmentsPrefix = NameUtils.INTERNAL_SCOPE_NAME;
+        this.systemSegmentsPrefix = NameUtils.INTERNAL_CONTAINER_PREFIX;
```
andreipaduroiu, line 231: Is this a backward compatible change? We're changing this from `_system` to `_system/containers`

andreipaduroiu: I see below that this is mostly an idempotent change; how is `this.systemSegmentsPrefix` used though?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorage.java
andreipaduroiu: Any chance we could write a unit test to verify this? Even with mocking, using `ReusableLatch`, etc.?

andreipaduroiu (changes requested):

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
sachin-j-joshi: This just replaces "_system" + /containers/metadata...." with "_system/containers/" + "metadata....".
No functional change.

This method was actually not used anywhere in the production code until now. (i don't remember what was it used for when I first wrote that :) 
Not yet sure if this is good enough or we need to do something more fancy like regex etc. There are 4 fixed names per containers if segment is one of those it's a system segment.
```
    /**
     * Indicates whether given segment is a system segment.
     *
     * @param segmentName Name of the segment to check.
     * @return True if given segment is a system segment.
     */
    public boolean isStorageSystemSegment(String segmentName) {
        if (segmentName.startsWith(systemSegmentsPrefix)) {
            for (String systemSegment : systemSegments) {
                if (segmentName.equals(systemSegment)) {
                    return true;
                }
            }
        }
        return false;
    }
```

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
```diff
@@ -2561,6 +2561,53 @@ private void testParallelReadRequestsOnSingleSegmentWithReentry(int numberOfRequ
...
+     *
+     * @throws Exception Throws exception in case of any error.
+     */
+    @Test
+    public void testSystemSegmentConcurrency() throws Exception {
+
+        SegmentRollingPolicy policy = new SegmentRollingPolicy(2); // Force rollover after every 2 byte.
+        @Cleanup
+        TestContext testContext = getTestContext(ChunkedSegmentStorageConfig.DEFAULT_CONFIG.toBuilder().indexBlockSize(3).build());
+        if (!(testContext.metadataStore instanceof InMemoryMetadataStore)) {
```
andreipaduroiu, line 2575: Why are we doing this? This will silently not run the test at all for pretty much non-InMemoryMetadataStore.

Is there a specific unit test class/suite that uses the InMemory one? If so, let's add this test there.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
```diff
@@ -2561,6 +2561,53 @@ private void testParallelReadRequestsOnSingleSegmentWithReentry(int numberOfRequ
...
+            return;
+        }
+
+        // Simulate some writes to system segment, this should cause some new chunks being added.
+        val n = 4;
```
andreipaduroiu, line 2580: I suppose `n` is number of writes? Let's name the variables with descriptive names so we don't guess this later on.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
```diff
@@ -2561,6 +2561,53 @@ private void testParallelReadRequestsOnSingleSegmentWithReentry(int numberOfRequ
...
+        val data = new byte[n][100];
+
+        var futures = new ArrayList<CompletableFuture<Void>>();
+        @Cleanup("shutdownNow")
+        ExecutorService executor = Executors.newFixedThreadPool(n);
```
andreipaduroiu, line 2585: Just double-checking: is this executor actually needed because we need to have a thread blocked for each write (for testing purposes)? I'm trying to see if the `executorService()` that's already available in this class would be adequate for this test or not.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -1191,6 +1196,99 @@ public void testSimpleTruncation() throws Exception {
...
+
+        // Bootstrap
+        segmentStorage1.bootstrap(snapshotInfoStore).join();
+        segmentStorage1.getGarbageCollector().deleteGarbage(false, 1000).get();
+
+        checkSystemSegmentsLayout(segmentStorage1);
+
+        // Simulate some writes to system segment, this should cause some new chunks being added.
+        // Simulate some writes to system segment, this should cause some new chunks being added.
+        val n = 4;
```
andreipaduroiu, line 1238: same here

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalTests.java
```diff
@@ -1191,6 +1196,99 @@ public void testSimpleTruncation() throws Exception {
...
+        val data = new byte[n][100];
+
+        Random rnd = new Random();
```
andreipaduroiu, line 1241: In unit tests, always initialize `new Random` with a constant - that will make your test predictable and reproducible. For example, if you get a test failure due to some specific input, using a Random without constant seed will guarantee that you will never see that input ever again, so reproducing a (sporadic) issue becomes much, much harder in that case.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/ChunkedSegmentStorageTests.java
sachin-j-joshi: removed this check. This test does not have any special test hooks that would fail in non-InMemoryMetadataStore-.

andreipaduroiu (approved):

PR #6188, opened 2021-07-29
Issue 6187: Disabling authentication for the health check apis

### LINKED ISSUE #6187: TLS/auth should be disabled at the REST endpoint for the Healthcheck APIs
Discovery/health checks APIs could be left unauthenticated because of the nature of the information that they would provide i.e. information about whether a particular component is up or not (or fully ready or not) which could be useful to the authenticated and unauthenticated clients alike.

For now, we could simply disable TLS and auth for the health check APIs, so that no credentials would be required in order to access them.

Would it be possible for a caller of the health check API to abuse it in a way which could lead to DoS? The answer is yes. But there would be no less chances of an unauthenticated user overwhelming the server using invalid credentials too, if the responses are not cached at the server side and we do not allow more than a certain number of requests from the same client within a given period of time. Hence better handling would be required at the server side to protect against such abuse. Till the time being, we could however safely proceed with the approach of leaving this particular endpoint unauthenticated.

**Additional context**
Any call to the health check API should not be challenged with auth or with incorrect credentials to test it out etc. It should be a holistic API that just tells:
- whether the component is up or not?  (and fully ready)
- provides other details that we want to expose as part of the this health check response

SrishT:

**Change log description**  
Disables authentication for the health check API endpoints, so that no credentials would be required to invoke them.

**Purpose of the change**  
Fixes #6187 

**What the code does**  
Does not register the AuthHandlerManager while instantiating the HealthImpl class inside the ControllerServiceStarter and ServiceStarter classes, so that TLS and Authentication is not enabled for the Health Check API endpoints.

**How to verify it**  
Deploying pravega with TLS/Auth enabled and invoking the healthcheck api endpoints from controller and segmentstore pods should return a response from the healthcheck api (which defaults to `true` at present)

For Controller
```
curl -s -X GET 'http://localhost:10080/v1/health/liveness'
curl -s -X GET 'http://localhost:10080/v1/health/readiness'
```
For Segmentstore
```
curl -s -X GET 'http://localhost:6061/v1/health/liveness'
curl -s -X GET 'http://localhost:6061/v1/health/readiness'
```

### DIFF
--- a/shared/health-bindings/src/main/java/io/pravega/shared/health/bindings/resources/HealthImpl.java
+++ b/shared/health-bindings/src/main/java/io/pravega/shared/health/bindings/resources/HealthImpl.java
@@ -53,12 +53,10 @@ public class HealthImpl implements ApiV1.HealthApi {
 
     private final HealthEndpoint endpoint;
 
-    private final RESTAuthHelper restAuthHelper;
     private final AuthorizationResource authorizationResource = new AuthorizationResourceImpl();
 
-    public HealthImpl(AuthHandlerManager pravegaAuthManager, HealthEndpoint endpoint) {
+    public HealthImpl(HealthEndpoint endpoint) {
         this.endpoint = endpoint;
-        this.restAuthHelper = new RESTAuthHelper(pravegaAuthManager);
     }
 
     @Override
@@ -74,7 +72,6 @@ public class HealthImpl implements ApiV1.HealthApi {
     private void getHealth(String id, SecurityContext securityContext, AsyncResponse asyncResponse, String method) {
         long traceId = LoggerHelpers.traceEnter(log, method);
         processRequest(() -> {
-            restAuthHelper.authenticateAuthorize(getAuthorizationHeader(), authorizationResource.ofScopes(), READ_UPDATE);
             Health health = endpoint.getHealth(id);
             Response response = Response.status(Response.Status.OK)
                     .entity(adapter(health))
@@ -96,7 +93,6 @@ public class HealthImpl implements ApiV1.HealthApi {
     private void getLiveness(String id, SecurityContext securityContext, AsyncResponse asyncResponse, String method) {
         long traceId = LoggerHelpers.traceEnter(log, method);
         processRequest(() -> {
-            restAuthHelper.authenticateAuthorize(getAuthorizationHeader(), authorizationResource.ofScopes(), READ_UPDATE);
             boolean alive = endpoint.isAlive(id);
             asyncResponse.resume(Response.status(Response.Status.OK)
                     .entity(alive)
@@ -117,7 +113,6 @@ public class HealthImpl implements ApiV1.HealthApi {
     private void getDetails(String id, SecurityContext securityContext, AsyncResponse asyncResponse, String method) {
         long traceId = LoggerHelpers.traceEnter(log, method);
         processRequest(() -> {
-            restAuthHelper.authenticateAuthorize(getAuthorizationHeader(), authorizationResource.ofScopes(), READ_UPDATE);
             Map<String, Object> details = endpoint.getDetails(id);
             asyncResponse.resume(Response.status(Response.Status.OK)
                     .entity(adapter(details))
@@ -138,7 +133,6 @@ public class HealthImpl implements ApiV1.HealthApi {
     private void getReadiness(String id, SecurityContext securityContext, AsyncResponse asyncResponse, String method) {
         long traceId = LoggerHelpers.traceEnter(log, method);
         processRequest(() -> {
-            restAuthHelper.authenticateAuthorize(getAuthorizationHeader(), authorizationResource.ofScopes(), READ_UPDATE);
             boolean ready = endpoint.isReady(id);
             asyncResponse.resume(Response.status(Response.Status.OK)
                     .entity(ready)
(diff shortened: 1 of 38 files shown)


andreipaduroiu (approved):

shared/health-bindings/src/main/java/io/pravega/shared/health/bindings/resources/HealthImpl.java
```diff
@@ -53,12 +49,10 @@
 
     private final HealthEndpoint endpoint;
 
-    private final RESTAuthHelper restAuthHelper;
     private final AuthorizationResource authorizationResource = new AuthorizationResourceImpl();
 
-    public HealthImpl(AuthHandlerManager pravegaAuthManager, HealthEndpoint endpoint) {
```
co-jo, line 59: Do we really want to get rid of _all_ authentication, or just authentication for the readiness/liveness routes so the operator(s) can use them?

co-jo (approved):
LGTM.

shared/health-bindings/src/main/java/io/pravega/shared/health/bindings/resources/HealthImpl.java
sarlaccpit: I think that was the idea yes. BUT, to add more nuance, that idea came with the caveat of "are we sure _all_ of these are ok to expose unauthenticated"?  (maybe the answer yes, it just needs to be explicitly decided)

I see in the API spec some liveness/health, then some more detailed calls.    Depending what is returned from these more detailed ones, you may or may not want to leave it _all_ open, I'm not sure.

For sure the get /liveness , /readiness, /status, look very safe.    I don't know enough about the rest of them though.

It IS possible to secure some and not others (by passing the authManager still, but removing its use only where you want).

@SrishT , just for playing devil's advocate, what does some of these "detail" calls return in terms of payload?

andreipaduroiu: We don't have any of them yet. We will be building them gradually both in the Controller and Segment Store. 

In the Segment Store, for example, details may include:
- Each assigned segment container and its status (running, starting, etc.)
- For each container:
    - Each segment active on a segment container (this might be a lot - but we'll see what we can do). We will include segment name, length sealed/etc.
    - Status of each internal Segment Container component (storage writer, durable log, metadata cleaner, event processor, etc.) 
- Quick stats about the cache.

In essence, the only customer information that will be exposed is the segment names. From those you can derive stream and scope names (but remember that segments are internally generated). The only information that this could provide to an "attacker" would be that "segment X exists and is part of stream Y". But then such as hypothetical attacker would still need to go through our regular gateway to access more information about such segment or read its contents, etc. 

The goal of this is not to expose too many details or pry-open the segments themselves (we have the Admin CLI under works for that goal), but to get a quick glance as to what's running within normal parameters, what's running hot and what's not running. To give you an idea of how this might be used, we may decide to do a sys admin dashboard that periodically pings all our services with "details == true" and present a one-page visual about the health of the cluster; that page would allow the admin to immediately see if something is requiring immediate attention or not. 

This should not really contain any more information than we already publish in our metrics - we already publish stream and segment tags there.

andreipaduroiu:
@sarlaccpit @SrishT Even though I approved this, I want to hear your feedback to make sure that we are all on the same page before we hit the "merge" button.

shared/health-bindings/src/main/java/io/pravega/shared/health/bindings/resources/HealthImpl.java
SrishT: @sarlaccpit currently the `/details` endpoint does not return anything.
But if there is any plan of exposing any sensitive information from this endpoint which unauthenticated users should not be given access to, then in that case it will be worthwhile to expose this particular endpoint only to authenticated users.
But since I am not sure about how sensitive these details would be, I would like to hear from @andreipaduroiu  and @pbelgundi about your opinion on this.

sarlaccpit:
> @sarlaccpit @SrishT Even though I approved this, I want to hear your feedback to make sure that we are all on the same page before we hit the "merge" button.

Thanks @andreipaduroiu , @SrishT .   I read your updates and the only thing that could eventually be a problem is when it comes to revealing, even indirectly through container information, the existence of scopes/streams to the unauthenticated subject.  If you think of it, unauthenticated users that do a list scopes or list streams, will not get this information.(by design)

So even though you're right, they would need to go through the front gate and get authenticated/authorized in order to DO anything with that info, it's still some kind of info leakage that would help the attacker peel the layers.

That said, it seems that this detailed info is not yet part of the API and what is there NOW is safe.

So perhaps the right thing to do is:
- leave the auth handler being passed to the HealthImpl
- but only use it in the detail apis (preventively, based on the future plans you explained).
- leave all the basic pings/status up/are you ready type calls 100% open, to facilitate things like readiness probes etc.

I think we're on the right track and all the bits are there; IMHO, just a small adjustment to protect the more detailed calls might be best.

shared/health-bindings/src/main/java/io/pravega/shared/health/bindings/resources/HealthImpl.java
pbelgundi: @SrishT I think @andreipaduroiu has tried to answer your question about what the "details" API could contain. Itoo  don't have complete clarity on what this API would return for Controller, but to play the devils advocate I would not recommend returning this to an unauthenticated user. As suggested by @sarlaccpit "/liveness, /readiness and & status" should be fine being unauthenticated.

SrishT:
> I think we're on the right track and all the bits are there; IMHO, just a small adjustment to protect the more detailed calls might be best.
@sarlaccpit the review comments have been addressed and the changes have been tested.

sarlaccpit (approved):

PR #6197, opened 2021-08-03
Issue 6219: Feature HealthCheck for SegmentStore

dellThejas:
Add healthcheck features for segment store.

Signed-off-by: dellThejas <Thejas.Vidyasagar@dell.com>

**Change log description**  
Added healtcheck for the following features in segmentstore:

- zookeeper connection

**Purpose of the change**  
#6219 

**What the code does**  
Added healthContributors for features in segmentstore:

- zookeeper connection

**How to verify it**  
Run pravega and run `curl -v http:localhost:controllerRestPort` and `curl -v http:localhost:segmentStoreRestPort`


dellThejas:
@medvedevigorek, You recommended to pass in `HealthServiceManager` to `AbstractConnectionListener`'s constructor, but this requires changes in over 20 files. Do you want me to to add a HealthServiceManager in all of those locations? It is mainly tests.

medvedevigorek:
> @medvedevigorek, You recommended to pass in `HealthServiceManager` to `AbstractConnectionListener`'s constructor. If I do this, I wont be able to tell if its a `PravegaConnectionListener` or `AdminConnectionListener` in the health details. Also, it is being called in several other locations, so should I add an instance of `HealthServiceManager` everywhere?

@dellThejas , adding HSM to `AbstractConnectionListener` will also require changing `PravegaConnectionListener` and `AdminConnectionListener` and this is when you can distinguish between the two. 

As for calls in other locations, if it is a production code then you need to modify it accordingly, if it is a test code then you can pass HSM as null as long as the test logic doesn't depend on the health checks. Of course you would need to ensure that HSM being null is handled correctly in `PravegaConnectionListener`/`AdminConnectionListener`/`AbstractConnectionListener`

andreipaduroiu (changes requested):
@dellThejas Please fix this PR to follow our contribution guidelines (you can find them on our wiki). At the least , let's create an issue and reference it from here (see other PRs).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
```diff
@@ -62,6 +63,9 @@
 public final class ServiceStarter {
     //region Members
 
+    @VisibleForTesting
+    protected HealthServiceManager healthServiceManager;
```
andreipaduroiu, line 67: It's not a good practice to make your fields non-private (protected implies non-private). Some external actor can just modify this. 

I suggest you change it back to private and add `@Getter` to it - you can then expose it via the getter without risking that someone can modify it (i.e., replace it with some other instance).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
```diff
@@ -150,19 +143,34 @@ public void start() throws Exception {
...
                     this.serviceConfig.getListeningIPAddress(), this.serviceConfig.getAdminGatewayPort(), service, tableStoreService,
-                    tokenVerifier, this.serviceConfig.getCertFile(), this.serviceConfig.getKeyFile(), this.serviceConfig.getTlsProtocolVersion());
+                    tokenVerifier, this.serviceConfig.getCertFile(), this.serviceConfig.getKeyFile(), this.serviceConfig.getTlsProtocolVersion(),
+                    healthServiceManager);
             this.adminListener.startListening();
             log.info("AdminConnectionListener started successfully.");
         }
         log.info("StreamSegmentService started.");
+
+        log.info("Initializing HealthService ...");
```
andreipaduroiu, line 162: Just out of curiosity - why did you move this to the end? Was it starting prematurely?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -59,9 +65,12 @@
 
     private final String host;
     private final int port;
-    private Channel serverChannel;
+
+    private Channel serverChannel; // tracks the status of the connection
+
     private EventLoopGroup bossGroup;
     private EventLoopGroup workerGroup;
+    private HealthServiceManager healthServiceManager;
```
andreipaduroiu, line 73: `final` as you only assign it in the constructor and never reassign it again.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -72,14 +81,45 @@
...
+     * Creates a new instance of the AdminConnectionListener class with HealthServiceManager.
+     *
+     * @param enableTls          Whether to enable SSL/TLS.
+     * @param enableTlsReload    Whether to reload TLS when the X.509 certificate file is replaced.
+     * @param host               The name of the host to listen to.
+     * @param port               The port to listen on.
+     * @param certFile           Path to the certificate file to be used for TLS.
+     * @param keyFile            Path to be key file to be used for TLS.
+     * @param tlsProtocolVersion The version of the TLS protocol
+     * @param healthServiceManager The healService to register new health contributors related to the listeners.
```
andreipaduroiu, line 117: typo

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -72,14 +81,45 @@
...
+     * @param type               The type of listener.
```
andreipaduroiu, line 118: Who sets this type? Is it someone external? Or is it just the derived classes?

If the latter, then I suggest defining an abstract method that returns the type and each implementation will return its appropriate type.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -224,4 +271,37 @@ public void close() {
             tlsCertFileModificationMonitor.stopMonitoring();
         }
     }
+
+    public static class ConnectionListenerHealthContributor extends AbstractHealthContributor {
+        private AbstractConnectionListener listener;
```
andreipaduroiu, line 276: final for all these fields.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -224,4 +271,37 @@ public void close() {
             tlsCertFileModificationMonitor.stopMonitoring();
         }
     }
+
+    public static class ConnectionListenerHealthContributor extends AbstractHealthContributor {
```
andreipaduroiu, line 275: Does this class need to be public? What is the minimum possible visibility it may have which still allows it to serve its purpose?

Same question applies to its constructor below (and in general to any method/class that you have - give them the minimum needed visibility).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -224,4 +271,37 @@ public void close() {
...
+            if (running) {
+                status = Status.NEW;
+            }
+
+            boolean ready = listener.serverChannel.isActive();
+            if (ready) {
+                status = Status.UP;
+            }
+
+            Map<String, Object> details = new HashMap<String, Object>();
```
andreipaduroiu, line 300: Tip: use `ImmutableMap.of` or `ImmutableMap.builder()....build()`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
andreipaduroiu: Or better yet, just use `this.getClass().getSimpleName()`. This will give you the name of the class you are instantiating (not the base class, but the derived class). so it should achieve the same purpose.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
```diff
@@ -0,0 +1,49 @@
...
+
+package io.pravega.segmentstore.server.host.health;
+
+import io.pravega.shared.health.Health;
+import io.pravega.shared.health.Status;
+import io.pravega.shared.health.impl.AbstractHealthContributor;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.imps.CuratorFrameworkState;
+
+public class ZKHealthContributor  extends AbstractHealthContributor {
```
andreipaduroiu, line 25: Javadoc for this class.

Nit: extra space before `extends`

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
```diff
@@ -0,0 +1,49 @@
...
+    private CuratorFramework zk;
```
andreipaduroiu, line 26: final

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public ZKHealthContributor(CuratorFramework zk) {
```
andreipaduroiu, line 28: Add `@NotNull` to the argument - that will enforce that you aren't given a `null` value.

Tip: there's `@Nonnull`, `@Notnull` and `@NotNull` (at least in my classpath). Make sure you use the one from the `lombok` package. If in doubt, search for it elsewhere in the codebase and see which one we use.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
```diff
@@ -0,0 +1,49 @@
...
+        if (running) {
+            status = Status.NEW;
+        }
+
+        boolean ready = zk.getZookeeperClient().isConnected();
+        if (ready) {
+            status = Status.UP;
+        }
+
+        return status;
```
andreipaduroiu, line 46: In the details, is there a way you can put the ZK endpoint it's connected to? Can we get that from anywhere? 

Also, anything else interesting we can get from the `CuratorFramework` interface?

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/ServiceStarterTest.java
```diff
@@ -15,35 +15,48 @@
...
+    private ServiceStarter serviceStarter;
+    private ZooKeeperServiceRunner zkService;
 
     @Before
-    public void startZookeeper() throws Exception {
-        zkTestServer = new TestingServerStarter().start();
-        zkUrl = zkTestServer.getConnectString();
+    public void setup() throws Exception {
+        int port = TestUtils.getAvailableListenPort();
+        zkService = new ZooKeeperServiceRunner(port, false, null, null, null);
```
andreipaduroiu, line 42: Why not using `TestingServerStarter` anymore?

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/ServiceStarterTest.java
```diff
@@ -15,35 +15,48 @@
...
+        zkService.initialize();
+        zkService.start();
+        ServiceBuilderConfig.Builder configBuilder = ServiceBuilderConfig
+                .builder()
+                .include(ServiceConfig.builder()
+                        .with(ServiceConfig.CONTAINER_COUNT, 1)
+                        .with(ServiceConfig.ZK_URL, "localhost:" + port)
+                        .with(ServiceConfig.HEALTH_CHECK_INTERVAL_SECONDS, 10)
```
andreipaduroiu, line 50: Do you need to wait anywhere in your tests for 10 seconds? That would be extremely long - if you use this anywhere, I suggest you dial it down so we don't make the tests take too long.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/ServiceStarterTest.java
```diff
@@ -54,16 +67,25 @@ public void stopZookeeper() throws IOException {
...
     public void testCuratorClientCreation() throws Exception {
-        ServiceBuilderConfig.Builder configBuilder = ServiceBuilderConfig
-                .builder()
-                .include(ServiceConfig.builder()
-                        .with(ServiceConfig.CONTAINER_COUNT, 1)
-                        .with(ServiceConfig.ZK_URL, zkUrl));
-        @Cleanup("shutdown")
-        ServiceStarter serviceStarter = new ServiceStarter(configBuilder.build());
-        @Cleanup
         CuratorFramework zkClient = serviceStarter.createZKClient();
```
andreipaduroiu, line 70: Do not remove `@Cleanup` from zkClient. That ensures it closes it regardless of whether your test succeeds or fails.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListenerTest.java
```diff
@@ -195,4 +198,29 @@ public void testCreateRequestProcessor() {
                 mock(StreamSegmentStore.class), mock(TableStore.class), NoOpScheduledExecutor.get());
         Assert.assertTrue(listener.createRequestProcessor(new TrackedConnection(new ServerConnectionInboundHandler())) instanceof AppendProcessor);
     }
+
+    // Test the health status created with pravega listener.
+    @Test
+    public void testHealth() {
+        int port = TestUtils.getAvailableListenPort();
+        PravegaConnectionListener listener = new PravegaConnectionListener(false, port,
```
andreipaduroiu, line 206: Does this class implement `AutoCloseable` or does it have a `close` or `shutdown` method? If so, add the `@Cleanup` to it to ensure that any resources it acquired during its lifetime are properly cleaned up when we get rid of it.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListenerTest.java
```diff
@@ -195,4 +198,29 @@ public void testCreateRequestProcessor() {
...
+                mock(StreamSegmentStore.class), mock(TableStore.class), NoOpScheduledExecutor.get());
+        listener.startListening();
+        try {
```
andreipaduroiu, line 209: How is this relevant to the test?

andreipaduroiu (changes requested):

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/ServiceStarter.java
dellThejas: My recent change moved the HealthService back to its original spot. I forgot to move the log statement with it. Sorry about that.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/ServiceStarterTest.java
dellThejas: I can revert it, I was trying to match how it was done in the standalone for starting zkService.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListenerTest.java
dellThejas: I guess we dont need that portion to test if port is available, since its being done earlier. Also, I did modify the test since now the contributor is private.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -72,14 +82,45 @@
     @Getter
     private final boolean enableTlsReload; // whether to reload TLS certificate when the certificate changes
 
+    private final String type;
```
RaulGracia, line 85: Is this necessary? I think that you can use `this.getClass().getSimpleName()` as input for `healthServiceManager.register` and you can get rid of this field.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -89,6 +130,8 @@ public AbstractConnectionListener(boolean enableTls, boolean enableTlsReload, St
         this.tlsProtocolVersion = Arrays.copyOf(tlsProtocolVersion, tlsProtocolVersion.length);
         InternalLoggerFactory.setDefaultFactory(Slf4JLoggerFactory.INSTANCE);
         this.connectionTracker = new ConnectionTracker();
+        this.healthServiceManager = healthServiceManager;
+        this.type = this.getClass().getSimpleName();
```
RaulGracia, line 134: Please, remove this if it is used only once in `startListening()`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -159,6 +202,11 @@ public void initChannel(SocketChannel ch) {
 
         // Start the server.
         serverChannel = b.bind(host, port).awaitUninterruptibly().channel();
+
+        if (healthServiceManager != null) {
+            healthServiceManager.register(new ConnectionListenerHealthContributor(type,
```
RaulGracia, line 207: This constructor looks super redundant. The `listener` (i.e., the second argument that you pass `this`) object already contains the `host`, `port`, and not to mention the class. So, just writing `new ConnectionListenerHealthContributor(this)` may be enough, there is no need for all the other arguments, right?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -224,4 +272,35 @@ public void close() {
...
     }
+
+    // A contributor for managing health of a connection listener.
+    private static class ConnectionListenerHealthContributor extends AbstractHealthContributor {
+        private final AbstractConnectionListener listener;
+        private final String host;
+        private final int port;
+
+        private ConnectionListenerHealthContributor(String connectionType, AbstractConnectionListener listener, String host, int port) {
+            super(connectionType);
```
RaulGracia, line 283: If you leave only the `listener` parameter for the constructor, you can just write:
```
super(listener.getClass().getSimpleName());
this.listener = listener;
```
That's it, you can use the host and port with `listerner.host` and `listener.port`, and you can get rid of these fields. Also, I think that you can add the `@NonNull` annotation to the `listener` parameter in the constructor, as you do not want to have a `null` listener.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListener.java
```diff
@@ -112,12 +112,14 @@ public PravegaConnectionListener(boolean enableTls, int port, StreamSegmentStore
      * @param replyWithStackTraceOnError Whether to send a server-side exceptions to the client in error messages.
      * @param executor           The executor to be used for running token expiration handling tasks.
      * @param tlsProtocolVersion the version of the TLS protocol
+     * @param healthServiceManager The healService to register new health contributors related to the listeners.
```
RaulGracia, line 115: Typo: `healService`. Also, if you are referring to a class, use caps.

RaulGracia (changes requested):

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
dellThejas: This definitely makes it simpler, thank you.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import io.pravega.shared.health.Health;
+import io.pravega.shared.health.Status;
+import io.pravega.shared.health.impl.AbstractHealthContributor;
+import lombok.NonNull;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.imps.CuratorFrameworkState;
+
+// A contributor to manage the health of zookeeper client connection.
```
RaulGracia, line 27: Comments in headers have the `/**/` format, not `//`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/AbstractConnectionListener.java
```diff
@@ -224,4 +270,32 @@ public void close() {
             tlsCertFileModificationMonitor.stopMonitoring();
         }
     }
+
+    // A contributor for managing health of a connection listener.
```
RaulGracia, line 274: Nit: comments for this class may be with `/**/` format too.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
andreipaduroiu: @dellThejas I asked that you add `@NonNull` to the argument to the constructor, not to the class-level variable.

What you did is OK if you also use `@RequiredArgsConstructor` on the class which will auto-generate the constructor for you. But if you have a method/constructor, you need to set that annotation on the argument to the method itself.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
andreipaduroiu, line 27: Please use proper javadoc.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListenerTest.java
```diff
@@ -195,4 +200,21 @@ public void testCreateRequestProcessor() {
                 mock(StreamSegmentStore.class), mock(TableStore.class), NoOpScheduledExecutor.get());
         Assert.assertTrue(listener.createRequestProcessor(new TrackedConnection(new ServerConnectionInboundHandler())) instanceof AppendProcessor);
     }
+
+    // Test the health status created with pravega listener.
+    @Test
+    public void testHealth() {
+        HealthServiceManager healthServiceManager = new HealthServiceManager(Duration.ofSeconds(2));
```
andreipaduroiu, line 207: Are you shutting down and cleaning up this instance? `HealthServiceManager` implements `AutoCloseable` so you must close it.

Add `@Cleanup` to it.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/health/ZKHealthContributor.java
RaulGracia: ^^ @dellThejas

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaConnectionListenerTest.java
```diff
@@ -195,4 +200,21 @@ public void testCreateRequestProcessor() {
...
+        int port = TestUtils.getAvailableListenPort();
+        @Cleanup
+        PravegaConnectionListener listener = new PravegaConnectionListener(false, false, "localhost",
+                port, mock(StreamSegmentStore.class), mock(TableStore.class), SegmentStatsRecorder.noOp(),
+                TableSegmentStatsRecorder.noOp(), new PassingTokenVerifier(), null, null, true,
+                NoOpScheduledExecutor.get(), TLS_PROTOCOL_VERSION.getDefaultValue().split(","), healthServiceManager);
+
+        listener.startListening();
+        Health health = listener.getHealthServiceManager().getHealthSnapshot();
+        Assert.assertEquals("HealthContributor should report an 'UP' Status.", Status.UP, health.getStatus());
```
RaulGracia, line 218: Check other cases, like when you close the listener this should report `DOWN`, right? Please, think carefully what are the cases you can find here.

RaulGracia (changes requested):
Looks good, but please address my last comments and the ones from previous reviews that remain unaddressed.

RaulGracia (approved):

andreipaduroiu (approved):

PR #6201, opened 2021-08-03
Issue 6200: SLTS - Immediately after bootstrap, persist extra snapshot to journals and use it for identification/elimination of zombie journal records.

### LINKED ISSUE #6200: SLTS - Journal reconciliation is not triggered if zombie truncates and fails before any new chunk addition
**Describe the bug**

_Background_ 

- System journals are used to replay and recreate state of storage metadata segments.
- The journals must be applied in sequential order even if they contain entries created when instance becomes zombie. 
- The zombie records are detected by detecting inconsistency and triggers reconciliation that removes zombie entries

This is an edge case scenario which can lead to metadata corruption in case of multiple quick failures and restarts of container and zombie instances are created one after another leading up to the race between zombies and valid instance. 

_Scenario in which it can happen._
- Information about new snapshot by new instance is not saved yet to core attributes. 
- While this instance is still running another newer container instance is created on different pod.
- Now one of the following happens 
    - Either Zombie instance has few truncates without any addition of new chunks 
    - or non-zombie instance has no truncates to trigger consistency check/reconciliation.
- Therefore reconciliation logic for that entry is **not triggered** and the journal entry for truncate by zombie instance is assumed to correct and incorrectly applied.

**To Reproduce**
Hard to reproduce, may be possible by running testEndToEndWithFencingWithChunkedStorage in loop

**Screenshots**

**Additional information**
Note that - Current implementation detects the system journal entries written by zombie instance. 
Therefore as long as new valid instance after zombie either truncates or adds new chunks the reconciliation works.
However the current logic requires at least one chunk addition by zombie instance to create inconsistency that will now trigger reconciliation logic.

It is easier and more robust to use state in snapshot for correct reconciliation. 
The information in snapshot always wins.

sachin-j-joshi:
**Change log description**  
Issue 6200: SLTS - Immediately after bootstrap, persist extra snapshot to journals and use it for identification/elimination of zombie journal records.

**Purpose of the change**  
Fixes #6200 

**What the code does**  

- Immediately after bootstrap generate and save full snapshot to journals. Note that info about this snapshot can not be saved to core attributes yet as full segment store is not yet active at this point.
- However this snapshot is extremely useful during proper reconciliation of zombie entries. 
- During reconciliation, as soon as first snapshot entry of new epoch is encountered, 
   - info in that snapshot can be used as is and reconciliation process can be restarted from that point onwards. 
   - This way any zombie records created by previous instance after this snapshot are automatically ignored.
   - In addition with this logic the snapshot of most recent epoch always wins.
- Adding additional validation checks for write, truncate and defrag operations

**How to verify it**  
All tests must pass.;


segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -759,15 +763,20 @@ private SystemSnapshotRecord readSnapshotRecord(SnapshotInfo snapshotInfo, byte[
                     Preconditions.checkState(!segmentSnapshot.segmentMetadata.getFirstChunk().equals(segmentSnapshot.segmentMetadata.getLastChunk()),
                             "First chunk and last chunk should not match. Segment snapshot= %s", segmentSnapshot);
                 }
+                int dataSize = 0;
                 ChunkMetadata previous = null;
                 for (val metadata : segmentSnapshot.getChunkMetadataCollection()) {
+                    dataSize += metadata.getLength();
```
andreipaduroiu, line 769: Any chance this might exceed the bounds for Integer?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
```diff
@@ -1370,11 +1384,14 @@ private String getSystemJournalChunkName(int containerId, long epoch, long curre
     private <R> CompletableFuture<R> executeExclusive(Callable<CompletableFuture<R>> operation) {
         return CompletableFuture.completedFuture(null).thenComposeAsync(v -> {
             try {
+                Preconditions.checkState(exclusiveExecutionGuard.compareAndSet(false, true), "PANIC: Non exclusive execution");
```
andreipaduroiu, line 1387: This looks similar to a built-in mutex (`synchronized` block); I suppose the reason you did it this way was to reject, rather than block the execution of the call until the previous call is done. How would this behave (differently) if we used a mutex here?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/TruncateOperation.java
```diff
@@ -85,17 +85,27 @@
...
+                                                segmentMetadata.checkInvariants();
                                                 Preconditions.checkState(segmentMetadata.getLength() == oldLength,
                                                         "truncate should not change segment length. oldLength=%s Segment=%s", oldLength, segmentMetadata);
-                                                segmentMetadata.checkInvariants();
+                                                Preconditions.checkState(oldChunkCount - chunksToDelete.size() == segmentMetadata.getChunkCount(),
+                                                        "Number of chunks do not match. old value (%s) - number of chunks deleted (%s) must match current chunk count(%s)",
+                                                        oldChunkCount, chunksToDelete.size(), segmentMetadata.getChunkCount());
+                                                if (null != currentMetadata && null != segmentMetadata.getFirstChunk()) {
+                                                    Preconditions.checkState(segmentMetadata.getFirstChunk().equals(currentMetadata.getName()));
```
andreipaduroiu, line 103: Should we put some messages here along with some information useful for debugging?

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
sachin-j-joshi: Thanks. Good catch ! 
metadata.length is long so I don't understand why there was no compiler warning in the first place.

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/SystemJournal.java
sachin-j-joshi: unnecessary check. removing

sachin-j-joshi:
Added unit test that recreates failing scenario.
More tests coming.

segmentstore/storage/src/test/java/io/pravega/segmentstore/storage/chunklayer/SystemJournalOperationsTests.java
```diff
@@ -744,6 +743,226 @@ void testScenarioWithFlakySnapshotInfoStore(TestMethod test, TestScenarioProvide
...
+        Assert.assertEquals(0, segmentMetadata4.getLastChunkStartOffset());
+    }
+
+    /**
+     * Test zombie scenario with multiple chunks.
+     * @throws Exception Exception if any.
+     */
+    @Test
+    public void testZombieScenarioMultipleChunks() throws Exception {
+        val testContext = new TestContext(CONTAINER_ID);
```
andreipaduroiu, line 887: `@Cleanup` for all these `TestContext` instances. They have a `close` method so they should be properly cleaned up.

andreipaduroiu (changes requested):

andreipaduroiu (approved):

PR #6203, opened 2021-08-05
Issue 6142: Expose segment rolloverSizeBytes to client.

### LINKED ISSUE #6142: Expose Segment Rollover Size via Wire Protocol and Client
Each Segment inside the Segment store has a `RolloverSize` property which indicates how big the LTS Chunks can get before they get rolled over. Currently this is set via a system-wide setting (in the Segment Store config) and applies universally to every segment.

However this one-size-fits-all does not scale well. We have a wide variety of stream segments (non-table-segments) with different purposes and a single value does not necessarily apply to all. For example, while a 4TB rollover size may make sense for user segments, Reader Group and Controller Event Processor segments (with small updates and overall small states) will waste a lot of space if using that large value.

We would like to expose the `RolloverSize` property externally and allow an application to define it when they create a Segment (both stream segments and table segments).

We should do the following:
1. Add a new field `rolloverSizeBytes` to the `CreateSegment` and `CreateTableSegment` WireCommands. Default to 0 (0 means the application does not care, leave it to the system default).
2. Update `PravegaRequestProcessor` to extract this value and pass it down to the appropriate segment store API when creating segments (these APIs should already accept it; if not, we should plumb it through). Once given to the Segment Store, it already knows what to do with it, so no more work should be needed here.
3. In the Client, when the Stream is created, add a new property called `desiredRolloverSize` (or `desiredRolloverSizeBytes` - TBD). Do the same for Key-Value Tables.
4. Plumb this value through the Controller's Stream configuration, and make sure it is passed (via the above-updated Wire Commands) to the Segment Store.

### LINKED ISSUE #6246: TimestampAggregationTimeout should be passed from client to controller.
**Describe the bug**
The `timestampAggregationTimeout` defined in the `StreamConfiguration` should be passed down from client to controller. However, in the [controller.proto](https://github.com/pravega/pravega/blob/605106e7bab27da3e12297c1cf09161e12b56444/shared/controller-api/src/main/proto/Controller.proto#L395) this field is omitted. 

Should enable the grpc to pass this parameter. 
Perhaps also in the REST API

Tristan1900:
**Change log description**  
Expose segment `rolloverSizeBytes` to client.
User can specify `rolloverSizeBytes` in `StreamConfiguration` for stream segment.
User can specify `rolloverSizeBytes` in `KeyValueTableConfiguration` for KVTable segment.
`rolloverSizeBytes` for reader group segment uses a default value 4MB.
If not specified, 0 will passed to server, which means a default value in the server side will be used for `rolloverSizeBytes`.

Fix `timestampAggregationTimeout ` parameter not passed to controller issue.
**Purpose of the change**  
Fix #6142 
Fix #6246

**What the code does**  
Add `rolloverSizeBytes` field in `StreamConfiguration` and `KeyValueTableConfiguration` for stream segment and KVTable segment. 
Add `rolloverSizeBytes` in coresponding wirecommands
Add logic in `pravegaRequestProcessor` to handle this parameter.

**How to verify it**  
New tests should pass


segmentstore/server/src/main/java/io/pravega/segmentstore/server/writer/SegmentAggregator.java
```diff
@@ -1176,6 +1176,9 @@ private long getRolloverSize() {
         // Configured value.
         long rolloverSize = this.metadata.getAttributes().getOrDefault(Attributes.ROLLOVER_SIZE, SegmentRollingPolicy.NO_ROLLING.getMaxLength());
 
+        // rolloverSize being zero means the default value should be used.
+        rolloverSize = rolloverSize == 0 ? SegmentRollingPolicy.NO_ROLLING.getMaxLength() : rolloverSize;
```
andreipaduroiu, line 1180: Why would this ever be 0? Is it because it's now user-supplied? If so, what if it's negative?

Tristan1900: Thanks for the review. 
Yes, I remember you mentioned in the issue that if this field is zero then it means application doesn't care and a default value should be used. So it is possible that a zero is passed down to here. 
I have set validation logic in the `PravegaRequestProcessor` to ensure that only non negative value can be used.

client/src/main/java/io/pravega/client/stream/StreamConfiguration.java
```diff
@@ -83,12 +83,21 @@
     @EqualsAndHashCode.Exclude
     private final Set<String> tags;
 
+    /**
+     * API to return segment rollover size.
```
andreipaduroiu, line 87: What if the user is not interested in this? It will likely be defaulted to 0. In that case, I think it makes sense that `0` should be interpreted as "whatever is default on the server".

Let's update the javadoc to explain this. 

Please do this for KVT Config below too.

client/src/main/java/io/pravega/client/tables/KeyValueTableConfiguration.java
```diff
@@ -73,7 +81,8 @@ public KeyValueTableConfiguration build() {
             Preconditions.checkArgument(this.partitionCount > 0, "partitionCount must be a positive integer. Given %s.", this.partitionCount);
             Preconditions.checkArgument(this.primaryKeyLength > 0, "primaryKeyLength must be a positive integer. Given %s.", this.primaryKeyLength);
             Preconditions.checkArgument(this.secondaryKeyLength >= 0, "secondaryKeyLength must be a non-negative integer. Given %s.", this.secondaryKeyLength);
-            return new KeyValueTableConfiguration(this.partitionCount, this.primaryKeyLength, this.secondaryKeyLength);
+            Preconditions.checkArgument(this.rolloverSizeBytes >= 0, "rollover size must be non-negative. Given %s.", this.rolloverSizeBytes);
```
andreipaduroiu, line 84: For consistency, let's use the exact same error message as in StreamConfiguration.

client/src/test/java/io/pravega/client/connection/impl/CommandEncoderTest.java
```diff
@@ -94,7 +94,7 @@ public void testRoundTrip() throws IOException {
         WireCommand command = new WireCommands.Hello(0, 1);
         commandEncoder.write(command);
         assertEquals(output.decoded.remove(0), command);
-        command = new WireCommands.CreateTableSegment(0, "segment", false, 16, "");
+        command = new WireCommands.CreateTableSegment(0, "segment", false, 16, "", 0);
```
andreipaduroiu, line 97: Use something that is not 0?

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -247,7 +248,7 @@ void setTimeout(Duration duration) {
         final long requestId = connection.getFlow().asLong();
 
         WireCommands.CreateSegment request = new WireCommands.CreateSegment(requestId, transactionName,
-                WireCommands.CreateSegment.NO_SCALE, 0, delegationToken);
+                WireCommands.CreateSegment.NO_SCALE, 0, delegationToken, 0);
```
andreipaduroiu, line 251: Transaction segments should have this too, shouldn't they?

controller/src/main/java/io/pravega/controller/server/eventProcessor/requesthandlers/CommitRequestHandler.java
```diff
@@ -310,7 +310,7 @@ public CommitRequestHandler(final StreamMetadataStore streamMetadataStore,
             // Use fixed scaling policy for these segments as they are created, merged into and sealed and are not
             // supposed to auto scale.
             return streamMetadataTasks.notifyNewSegment(scope, stream, segment, ScalingPolicy.fixed(1), delegationToken, 
-                    context.getRequestId());
+                    context.getRequestId(), 0);
```
andreipaduroiu, line 313: What segments are these? I suspect these are the internal controller segments. If so, then let's make a GitHub issue to have the Controller set proper value(s) here and reference that issue with a TODO (here and anywhere else that we might need).

controller/src/test/java/io/pravega/controller/server/SegmentHelperTest.java
```diff
@@ -96,7 +96,7 @@ public void createSegment() {
         @Cleanup
         SegmentHelper helper = new SegmentHelper(factory, new MockHostControllerStore(), executorService());
         CompletableFuture<Void> retVal = helper.createSegment("", "",
-                0, ScalingPolicy.fixed(2), "", Long.MIN_VALUE);
+                0, ScalingPolicy.fixed(2), "", Long.MIN_VALUE, 0L);
```
andreipaduroiu, line 99: Please put anything but `0L` in these tests and then verify that the value "went through" (i.e, you need to verify it).

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -438,20 +439,24 @@ public void createSegment(CreateSegment createStreamSegment) {
         Timer timer = new Timer();
         final String operation = "createSegment";
 
+        Preconditions.checkArgument(createStreamSegment.getRolloverSizeBytes() >= 0,
```
andreipaduroiu, line 442: I don't recommend we do it this way. At this point there is absolutely nothing that anyone can do to fix this. The Client has already sent its request to the Controller and got an ack back, and at this point the Controller is just going to retry this indefinitely, which will obviously lead nowhere if it keeps passing a bad value.

Let's log this as a `WARN` and interpret any invalid values as `0`.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -438,20 +439,24 @@ public void createSegment(CreateSegment createStreamSegment) {
...
+                "Segment rollover size must not be negative; actual %s.", createStreamSegment.getRolloverSizeBytes());
+
         Collection<AttributeUpdate> attributes = Arrays.asList(
                 new AttributeUpdate(SCALE_POLICY_TYPE, AttributeUpdateType.Replace, ((Byte) createStreamSegment.getScaleType()).longValue()),
                 new AttributeUpdate(SCALE_POLICY_RATE, AttributeUpdateType.Replace, ((Integer) createStreamSegment.getTargetRate()).longValue()),
+                new AttributeUpdate(ROLLOVER_SIZE, AttributeUpdateType.Replace, createStreamSegment.getRolloverSizeBytes()),
```
andreipaduroiu, line 448: If `0`, leave it as "default" - that is, do not add it.

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
```diff
@@ -629,6 +634,9 @@ public void createTableSegment(final CreateTableSegment createTableSegment) {
             typeBuilder.fixedKeyLengthTableSegment();
             configBuilder.keyLength(createTableSegment.getKeyLength());
         }
+        Preconditions.checkArgument(createTableSegment.getRolloverSizeBytes() >= 0,
```
andreipaduroiu, line 637: Same here.

segmentstore/server/host/src/test/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessorTest.java
```diff
@@ -605,10 +605,10 @@ public void testConditionalSegmentMergeReplace() throws Exception {
         PravegaRequestProcessor processor = new PravegaRequestProcessor(store, mock(TableStore.class), new TrackedConnection(connection),
                 recorderMock, TableSegmentStatsRecorder.noOp(), new PassingTokenVerifier(), false);
 
-        processor.createSegment(new WireCommands.CreateSegment(0, streamSegmentName, WireCommands.CreateSegment.NO_SCALE, 0, ""));
+        processor.createSegment(new WireCommands.CreateSegment(0, streamSegmentName, WireCommands.CreateSegment.NO_SCALE, 0, "", 0L));
```
andreipaduroiu, line 608: In some places you use `0` and in others you use `0L`. Let's be consistent.

Also, please pass something different than 0 and verify that it it is actually gone through. Also test 0 and negative values.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/FixedKeyLengthTableSegmentLayout.java
```diff
@@ -125,7 +125,7 @@ public void close() {
         ensureValidKeyLength("", config.getKeyLength());
         val result = new HashMap<AttributeId, Long>();
         result.put(Attributes.ATTRIBUTE_ID_LENGTH, (long) config.getKeyLength());
-        result.putAll(this.config.getDefaultCompactionAttributes());
+        result.putAll(this.config.getOrDefaultCompactionAttributes(config));
```
andreipaduroiu, line 128: I am not sure how I feel about this pattern. The point of the `getDefaultCompactionAttributes` was to fill in default values which would later be overridden. I would prefer if we kept the old mechanism, and then we populate specific attributes (essentially override) based on `config`.

When doing so, remember to treat `0L` as "default" - so skip that value.

segmentstore/server/src/main/java/io/pravega/segmentstore/server/tables/HashTableSegmentLayout.java
```diff
@@ -104,7 +104,7 @@ public void close() {
     @Override
     Map<AttributeId, Long> getNewSegmentAttributes(@NonNull TableSegmentConfig config) {
         Preconditions.checkArgument(config.getKeyLength() == 0, "Segment KeyLength must be 0 for HashTableSegments; actual %s.", config.getKeyLength());
-        return this.config.getDefaultCompactionAttributes();
+        return this.config.getOrDefaultCompactionAttributes(config);
```
andreipaduroiu, line 107: Here too

shared/controller-api/src/main/java/io/pravega/shared/controller/event/kvtable/CreateTableEvent.java
```diff
@@ -83,6 +84,7 @@ private void write00(CreateTableEvent e, RevisionDataOutput target) throws IOExc
             target.writeUUID(e.tableId);
             target.writeInt(e.primaryKeyLength);
             target.writeInt(e.secondaryKeyLength);
+            target.writeLong(e.rolloverSizeBytes);
```
andreipaduroiu, line 87: This isn't backwards compatible. You need to create Revision 1 where you (only) write the rollover size (and of course, where you read it from).

You'll need to create `write01` and `read01` and then in `declareVersions` just tag another `.revision(1, ...)` to that statement.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1066,11 +1066,13 @@ public void writeFields(DataOutput out) throws IOException {
             long lastModified = in.readLong();
             long segmentLength = in.readLong();
             long startOffset = 0;
+            long rolloverSizeBytes = 0;
```
andreipaduroiu, line 1069: What exactly did you modify in this command?

FYI I don't think we need to return the rollover size in GetInfo.

shared/protocol/src/main/java/io/pravega/shared/protocol/netty/WireCommands.java
```diff
@@ -1100,16 +1103,21 @@ public void writeFields(DataOutput out) throws IOException {
             out.writeInt(targetRate);
             out.writeByte(scaleType);
             out.writeUTF(delegationToken == null ? "" : delegationToken);
+            out.writeLong(rolloverSizeBytes);
```
andreipaduroiu, line 1106: Don't forget to bump the wire format version (top of the `WireCommands` class).

shared/protocol/src/test/java/io/pravega/shared/protocol/netty/WireCommandsTest.java
```diff
@@ -601,7 +601,7 @@ public void testStreamSegmentInfo() throws IOException {
 
     @Test
     public void testCreateSegment() throws IOException {
-        testCommand(new WireCommands.CreateSegment(l, testString1, b, i, ""));
+        testCommand(new WireCommands.CreateSegment(l, testString1, b, i, "", 0));
```
andreipaduroiu, line 604: Test it with non-zero.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -102,41 +120,85 @@ public void setup() throws Exception {
...
+        this.segmentHelper.close();
     }
 
     /**
      * Smoke Test. Verify that the KeyValueTable can be created and listed.
      */
     @Test
     public void testCreateListKeyValueTable() {
+        // Verify default rolloverSizeBytes config.
+        val kvt0 = newKeyValueTableName();
```
andreipaduroiu, line 161: I don't understand the purpose of this test but I highly doubt it belongs here. This looks a lot like a controller unit test. Please write this unit test somewhere in the controller project. 

This class should have little to no changes as a result of your change. It's an integration test, not a unit test. Its purpose is to test the End-to-end functionality of KVTs without regards to the implementation details. The rollover size is an implementation detail. 

To test the client-side rolloversize, I suggest you look at `KeyValueTableImplTests` (in the client). You can then use the `MockController` there to verify that the correct rollover size is passed to the Controller.

test/integration/src/test/java/io/pravega/test/integration/controller/server/StreamMetadataTest.java
```diff
@@ -83,6 +103,35 @@ public void testMetadataOperations() throws Exception {
                                                                .scalingPolicy(scalingPolicy)
                                                                .build();
 
+        // Create stream and check rollover size
+        // Create segment helper
+        HostMonitorConfig hostMonitorConfig = HostMonitorConfigImpl.builder()
```
andreipaduroiu, line 108: Same comment here as in the above class.

test/integration/src/test/java/io/pravega/test/integration/KeyValueTableTest.java
```diff
@@ -102,41 +120,85 @@ public void setup() throws Exception {
...
+
+        // 5. Create segment helper
+        HostMonitorConfig hostMonitorConfig = HostMonitorConfigImpl.builder()
+                .hostMonitorEnabled(false)
+                .hostMonitorMinRebalanceInterval(10)
+                .containerCount(containerCount)
+                .hostContainerMap(HostMonitorConfigImpl.getHostContainerMap(ENDPOINT, servicePort, containerCount))
+                .build();
+        HostControllerStore hostStore = HostStoreFactory.createInMemoryStore(hostMonitorConfig);
+        executor = ExecutorServiceHelpers.newScheduledThreadPool(1, "collector");
```
andreipaduroiu, line 138: FYI you almost never need to create an Executor in your unit tests. Most unit tests extend from `ThreadPooledTestSuite` which provide you with an executor.

test/integration/src/test/java/io/pravega/test/integration/controller/server/StreamMetadataTest.java
```diff
@@ -83,6 +103,35 @@ public void testMetadataOperations() throws Exception {
...
+                .hostMonitorEnabled(false)
+                .hostMonitorMinRebalanceInterval(10)
+                .containerCount(4)
+                .hostContainerMap(HostMonitorConfigImpl.getHostContainerMap(ENDPOINT, servicePort, 4))
+                .build();
+        HostControllerStore hostStore = HostStoreFactory.createInMemoryStore(hostMonitorConfig);
```
andreipaduroiu, line 114: None of these objects are cleaned up. You need to close them (use `@Cleanup`) to release any resources they acquired during their lifetime.

andreipaduroiu (changes requested):

shrids (changes requested):
- REST endpoints for Controller create/update streams should be updated.
-

shrids (changes requested):
- REST endpoints for Controller create/update streams should be updated.
- Should the _RG streams have a different rollover size ?

segmentstore/server/host/src/main/java/io/pravega/segmentstore/server/host/handler/PravegaRequestProcessor.java
Tristan1900: I'm not sure about this one. `Arrays.asList` is unmodifiable so if we keep it as is when 0 then the code would be ugly. Do you think we can make `attributes` modifiable like using `ArrayList`?

controller/src/test/java/io/pravega/controller/server/SegmentHelperTest.java
Tristan1900: yeah I think these tests and following tests are not really testing the rolloversize feature because the connection is mocked, rollover size is not passed to the segmentstore. 0 here is just a placeholder to make the exsiting test to pass.

I have added end to end (integration) test to test this feature. In the integration test this parameter is passed down to the segmentstore.

pbelgundi:
@andreipaduroiu I agree that we may require to have different _segment rollover sizes_ for different kinds of segments, based on their use, but how does the user arrive at the right value for a `Rollover Size` ? 
Would it be reasonable for the Controller to choose some defaults for rollover size depending on weather the Segment being created is for a _Reader Group SS Stream_ or a  _User Stream_ or a _Table Segment_...... 
How often would users need to override this value for a Stream/Table Segment?

controller/src/main/java/io/pravega/controller/server/rest/generated/model/CreateStreamRequest.java
```diff
@@ -40,6 +40,9 @@
   @JsonProperty("streamTags")
   private TagsList streamTags = null;
 
+  @JsonProperty("rolloverSizeBytes")
+  private Long rolloverSizeBytes = null;
```
pbelgundi, line 44: Shouldn't we call this `segmentRolloverSizeBytes` ?

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -258,6 +260,17 @@ public ReaderGroupConfigBuilder retentionType(StreamDataRetention type) {
...
        }
 
+       /**
+        * Set the rollover size for the {@link ReaderGroup} internal stream segment.
+        *
+        * @param rolloverSizeBytes rollover size for the internal stream segment.
+        * @return Reader group config builder.
+        */
+       public ReaderGroupConfigBuilder rolloverSizeBytes(final long rolloverSizeBytes) {
+           this.rolloverSizeBytes = rolloverSizeBytes;
```
pbelgundi, line 270: Unless we require setting a different _`rolloverSizeBytes`_ for each Reader Group Stream, this value can be chosen by the Controller without requiring the user to set it.

pbelgundi (changes requested):

client/src/main/java/io/pravega/client/tables/KeyValueTableConfiguration.java
```diff
@@ -50,10 +50,22 @@
...
+     * The rollover size for table segment in LTS.
+     *
+     * The default value for this field is 0.
+     * If default value is passed down to the server, a non-zero value defined in the server
+     * will be used for the actual rollover size.
+     *
+     * @param rolloverSizeBytes The rollover size for the table segment in LTS.
+     * @return The rollover size for the table segment in LTS.
+     */
+    private final long rolloverSizeBytes;
```
pbelgundi, line 67: Could this value be chosen by the Controller? 
How often would the user need to set a different rollover size for segments of different Key Value Tables?

controller/src/main/java/io/pravega/controller/server/rest/generated/model/CreateStreamRequest.java
```diff
@@ -40,6 +40,9 @@
   @JsonProperty("streamTags")
   private TagsList streamTags = null;
 
+  @JsonProperty("rolloverSizeBytes")
```
shrids, line 43: NOTE: This class is auto generated by the swagger code generator program

controller/src/main/java/io/pravega/controller/store/stream/records/StreamConfigurationRecord.java
```diff
@@ -301,6 +302,23 @@ private void write02(StreamConfigurationRecord streamConfigurationRecord, Revisi
...
 
+        private void read03(RevisionDataInput revisionDataInput,
+                            StreamConfigurationRecordBuilder configurationRecordBuilder)
+                throws IOException {
+            StreamConfiguration.StreamConfigurationBuilder streamConfigurationBuilder = StreamConfiguration.builder();
+            streamConfigurationBuilder.scalingPolicy(configurationRecordBuilder.streamConfiguration.getScalingPolicy())
+                                      .retentionPolicy(configurationRecordBuilder.streamConfiguration.getRetentionPolicy())
+                                      .timestampAggregationTimeout(configurationRecordBuilder.streamConfiguration.getTimestampAggregationTimeout())
+                                      .tags(configurationRecordBuilder.streamConfiguration.getTags())
+                                      .rolloverSizeBytes(revisionDataInput.readLong());
```
shrids, line 313: This would read corrupted data, right? 
`write02(..)` writes other fields.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
```diff
@@ -280,7 +293,8 @@ public ReaderGroupConfig build() {
                    "Outstanding checkpoint request should be greater than zero");
 
            return new ReaderGroupConfig(groupRefreshTimeMillis, automaticCheckpointIntervalMillis,
-                   startingStreamCuts, endingStreamCuts, maxOutstandingCheckpointRequest, retentionType, generation, readerGroupId);
+                   startingStreamCuts, endingStreamCuts, maxOutstandingCheckpointRequest, retentionType,
+                   generation, readerGroupId, rolloverSizeBytes);
```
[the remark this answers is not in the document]
Tristan1900: Yes we could do that in the client, do you have a rough idea how large would that value be? 
Thanks

controller/src/main/java/io/pravega/controller/server/rest/generated/model/CreateStreamRequest.java
Tristan1900: Thanks for pointing out. Yes this is generated code, I follow the guide here to generate it https://github.com/pravega/pravega/tree/master/shared/controller-api/src/main/swagger

controller/src/main/java/io/pravega/controller/store/stream/records/StreamConfigurationRecord.java
Tristan1900: Sorry I don't quite get it. My understanding is that what's written by `write02` should be read up by the `read02` and `read03` here should only be reading what's written by `write03`. Is that the case?

andreipaduroiu: Yes, the way it's declared above, `read02` is paired with `write02` and mapped to revision 2; while `read03` is paired with `write03` and revision 3. The `VersionedSerializer` will make sure to invoke the appropriate read method for each revision.

andreipaduroiu: As far as I can see, we're now writing a single long in 03 and reading a single long in 03. This is fine. Old code will not have the 03 methods and will thus ignore this value, and the new code, when reading something serialized with the old version, will not be invoking the 03 methods at all since there's no revision 3 encoded in it (old code doesn't know about  it).

shrids: Thanks.

client/src/main/java/io/pravega/client/stream/ReaderGroupConfig.java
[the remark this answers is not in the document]
shrids: 4MB is the size that was discussed, given that updates are small and truncation happens regularly.

Tristan1900:
Please don't merge this yet as there is still a comment needs to be addressed and also I will trigger a system test for this PR

Tristan1900:
System tests have passed.

Issue #4923, opened 2020-07-06
Remove client dependency on Netty


tkaitchuck:
**Problem description**
Netty is a common library which many client applications likely have on their classpath. We should do what we can to avoid problems caused by conflicting versions.

**Problem location**
Multiple locations, both related to handling of connections as well as ByteBuf, which is used in several places in the client

**Suggestions for an improvement**
Move onto standard java sockets and BufferView similar to the server.

PR #6206, opened 2021-08-05
Issue 4923: Enable okHttp based grpc client.

shrids:
!! Testing is in Progress, DO NOT MERGE !!
**Change log description**  
Enable okHttp based Grpc client .

**Purpose of the change**  
Related to #4923

**What the code does**  
- Ensure netty is not used by the controller Grpc client. Switch to okHttp.

**How to verify it**  
All the existing tests should continue to pass. Ensure there are no performance regressions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -289,7 +289,7 @@ project('shared:protocol') {
     dependencies {
         compile project(':common')
         compile group: 'io.netty', name: 'netty-transport', version: nettyVersion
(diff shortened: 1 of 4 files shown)


shrids:
I just realized that `grpc-java` based on `okhttp `does not support TLSv1.3. This is because grpc-java depends on okhttp version 2.7.4 and not the later version. Corresponding grpc-java issue : https://github.com/grpc/grpc-java/issues/7431

PR #6220, opened 2021-08-11
Issue 6021: Make sure that commitProcessor can be unblocked if queueProcessor terminates exceptionally

### LINKED ISSUE #6021: OperationProcessor queue may be left closed while the container is in RUNNING state

RaulGracia:
**Change log description**  
Make sure that commitProcessor can be unblocked if queueProcessor terminates exceptionally in OperationProcessor class.

**Purpose of the change**  
Fixes #6021.

**What the code does**  
In `OperationProcessor`, the interplay between `commitProcessor` and `queueProcessor` may be responsible for blocking the shutdown process. That is, in some specific situations, it is possible that: i) `queueProcessor` fails and terminates exceptionally, ii) the `operationQueue` is left closed as part of the partial shutdown process executed, iii) `commitProcessor` is left blocked waiting in the `take()` method of the `commitQueue`. The service will shut down completely when the `CompletableFuture.allOf(queueProcessor, commitProcessor)` statement in `doRun()` completes, but this cannot happen until both futures are completed either normally or exceptionally. Given that `commitProcessor` is waiting to `take()` elements from `commitQueue`, the shutdown process never terminates and any incoming operation fails with an `ObjectClosedException`, given that the `operationQueue` has been closed already.

To fix this problem, we need `commitProcessor` to terminate in the previous scenario. To this end, we have done the following changes:
- Now, `commitProcessor` will use the `take()` method with timeout. That is, if no elements are available in the queue for some time (e.g., 1 second defined in `COMMIT_PROCESSOR_TIMEOUT`), the method will throw a `TimeoutException`. This allows the `commitProcessor` to start a new processing iteration instead of being stuck in the `take()` method indefinitely.
- The `loop` condition of `commitProcessor` now also considers whether `queueProcessor` has terminated exceptionally or not. In the case that `queueProcessor` terminates exceptionally, `commitProcessor` will consume the remaining elements in `commitQueue` and then terminate. This will allow `CompletableFuture.allOf(queueProcessor, commitProcessor)` to complete and run the `doStop()` method, which shuts down the `OperationProcessor` service.

**How to verify it**  
Added a new test `testStartWithFailedFuturesInDurableLog()` that resembles the situation in which the original problem occurs. Without this change, we have been able to reproduce the original problem with the above test (i.e., `OperationProcessor` was stuck and could not shut down). With the change in this PR, we have not been able to reproduce the original issue after hundreds of executions of the test.


segmentstore/server/src/main/java/io/pravega/segmentstore/server/logs/OperationProcessor.java
```diff
@@ -147,8 +149,9 @@ protected Duration getShutdownTimeout() {
         // As opposed from the QueueProcessor, this needs to process all pending commits and not discard them, even when
         // we receive a stop signal (from doStop()), otherwise we could be left with an inconsistent in-memory state.
         val commitProcessor = Futures
-                .loop(() -> isRunning() || this.commitQueue.size() > 0,
-                        () -> this.commitQueue.take(MAX_COMMIT_QUEUE_SIZE)
+                .loop(() -> (isRunning() && !queueProcessor.isCompletedExceptionally()) || this.commitQueue.size() > 0,
+                        () -> Futures.exceptionallyExpecting(this.commitQueue.take(MAX_COMMIT_QUEUE_SIZE, COMMIT_PROCESSOR_TIMEOUT, this.executor),
```
andreipaduroiu, line 153: I think I see what you're doing here. A pending `take` would have thrown once `commitQueue.close()` is invoked. If you have to have a timeout here then it means that `commitQueue` somehow isn't closed yet. Did you notice that this is the case when the `OperationProcessor` enters the partially shutdown state?

I also wonder if we could rewrite this in a nicer way, without the use of `exceptionallyExpecting`.
Maybe revert back to the old call, but replace `thenAcceptAsync` with `handleAsync((items, ex) -> ...)` and if `ex` (unwrapped) is a `TimeoutException` then we just ignore it and move on. We'd rethrow anything else and of course if no exception, just invoke `processCommits`. 

Do you think this would look more elegant? (I really can't tell unless I see it in writing :) )

RaulGracia: Yes, is the `commitProcessor` the one that closes the `commitQueue` once all the elements are processed to do not lose any. Therefore, the problem is that this piece of code may sporadically fall in a situation in which the `commitProcessor` is already blocked on `take()` and the `queueProcessor` has crashed and does not queue any further element to the `commitQueue`. In that case, things get stuck because the `CompletableFuture.allOf()` that is at the end of the `doRun()` will never complete. By adding more logs, I have observed that the cases in which the `OperationProcessor` gets stuck, the `whenComplete` stage of `commitProcessor` is never executed, so it is waiting on the `take()`. The current changes basically adds the possibility to unblock the `commitProcessor` and therefore be able to shut down the `OperationProcessor` by i) adding a timeout to the `take()` method in `commitProcessor` and, ii) adding as a necessary condition for the `commitProcessor` to run that the `queueProcessor` has not completed exceptionally (i.e., crashed). If the `queueProcessor` has crashed, we are sure that no new elements will be added, so once the remaining elements are processed by `commitProcessor`, we can safely shut down.

Besides, I have done the change according to your suggestion. Let me know if it looks more elegant now.

RaulGracia:
@andreipaduroiu system tests are green with this branch (`test-pravega-continuous/4041`).

andreipaduroiu (approved):

RaulGracia:
3/3 system tests are green with this change.

PR #6222, opened 2021-08-12
Issue 6221: Setting BookieID for every bookie instance

### LINKED ISSUE #6221: Setting a unique BookieId for every newly started bookkeeper instance
**Describe the bug**
As described in [this](https://github.com/apache/bookkeeper/issues/2612) issue, there is a possibility of a potential data loss scenario which we can end up in because of the way in which the BookieId is set at present. Say we induce a destructive restart on a given pod in the ensemble (meaning that its data is wiped out, but it's identity is still the same). In this case instead of waiting for the responses from all the Bookies in the ensemble, the Bookkeeper client might return considering only the response of the restarted Bookie, which would lead to a sporadic data loss scenario.

**To Reproduce**
Can be reproduced by running the [BookieFailoverTest](https://github.com/pravega/pravega/blob/master/test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java) multiple times

**Screenshots**
n/a

**Additional information**
We could leverage [this](http://bookkeeper.apache.org/bps/BP-41-bookieid/) feature which has been introduced in bookkeeper 4.12 onwards, wherein the BookieId has been decoupled from its underlying network address, and could help in solving the data loss issue we are currently facing intermittently.

SrishT:

**Change log description**  
Sets a unique `bookieId` value for every newly created bookkeeper instance, and reads the persisted `bookieId` value for every other bookkeeper instance, so that the client can easily distinguish between an instance which is being newly created vs an instance which might have been restarted for some reason.

**Purpose of the change**  
Fixes #6221 

**What the code does**  
Sets a unique `bookieId` value for every newly created bookkeeper instance and mounts its value for every newly added bookkeeper instance to a persistent volume, so that it can be easily accessed even if this instance happens to get restarted at some point. 
However, in case this is not a fresh bookkeeper cluster installation, and we are upgrading from a previous bookkeeper version (which does not provide users the support to explicitly configure the BookieID) to the bookkeeper version which provides this support, the code reuses the older nomenclature for the BookieID, so that we do not face an `InvalidCookieException` due to a mismatch in the BookieID of any bookeeper instance post upgrade. In addition to this, a teardown script is added which would be invoked (only if  we are on the latest bookkeeper operator), and would delete the cookies for the bookie that is being scaled down at the time of pod deletion.

**How to verify it**  
Following are the various use cases that have been verified with I/O running in the background
- old version of bookkeeper operator
  old version of bookkeeper
  upgraded bookkeeper to latest version (computes BookieID from hostname and persists it in a file within a journal subdirectory)
  upgraded bookkeeper operator to latest version
  (this approach uses older nomenclature for BookieID as expected, and uses new nomenclature for any new bookkeeper pod that comes up as a result of any scale up)

- old version of bookkeeper operator
  old version of bookkeeper
  upgraded bookkeeper operator to latest version
  upgraded bookkeeper to latest version
  (this approach uses older nomenclature for BookieID as expected, and uses new nomenclature for any new bookkeeper pod that comes up as a result of any scale up)

- latest version of bookkeeper operator (i.e. one which contains [this](https://github.com/pravega/bookkeeper-operator/pull/166) fix)
  old version of bookkeeper
  upgraded bookkeeper to latest version
  (this approach uses older nomenclature for BookieID as expected, and uses new nomenclature for any new bookkeeper pod that comes up as a result of any scale up)

- latest version of bookkeeper operator (i.e. one which contains [this](https://github.com/pravega/bookkeeper-operator/pull/166) fix)
  latest version of bookkeeper
  deletion of pod should cause it to read the BookieID from the persisted file
  scale down followed by a scale up will not find any BookieID stored and will therefore generate a new BookieID and persist it
  (this approach uses new nomenclature for BookieID as expected) 

Also verified that we are getting a green system test build over the last multiple runs.


docker/bookkeeper/entrypoint.sh
```diff
@@ -74,6 +77,24 @@ create_bookie_dirs() {
...
+# or read the Bookie ID if it already exists
+set_bookieid() {
+  if [ `find ${BK_ID_FILE} | wc -l` -gt 0 ]; then
+    echo "Found Bookie ID"
+    BK_bookieId=`cat ${BK_ID_FILE}`
+    export BK_bookieId
+  else
+    echo "Bookie ID not found"
+    TIMESTAMP="`date "+%Y%m%d-%H%M%S"`"
+    BK_bookieId="${HOST}-${TIMESTAMP}"
```
eolivelli, line 90: What about using some random UUID instead of a timestamp?
This ID will stay there forever for the whole life of the bookie and you will see it in every log file.
The timestamp may be confusing for people who read the logs

eolivelli (changes requested):

eolivelli:
I left some comment.i would like to add that the bookie I'd is strictly related to the data and mot with the host name or the configuration.
So I would keep that file inside some disk that contains the data (journal or ledgers) and not next to the configuration files

SrishT:
> I left some comment.i would like to add that the bookie I'd is strictly related to the data and mot with the host name or the configuration.
> So I would keep that file inside some disk that contains the data (journal or ledgers) and not next to the configuration files

@eolivelli have incorporated the feedback to store the bookieId inside the journal disk

docker/bookkeeper/entrypoint.sh
```diff
@@ -74,6 +75,27 @@ create_bookie_dirs() {
   done
 }
 
+# Create a Bookie ID if this is a newly added bookkeeper pod
+# or read the Bookie ID if it already exists
+set_bookieid() {
+  IFS=',' read -ra directories <<< $1
+  BK_ID_DIR="${directories[0]}"
```
RaulGracia, line 82: Here you are relying in the fact that the `directories` list will always have the list of Bookkeeper directories sorted in the same order. Do you know if that assumption will hold irrespective of the environment?

docker/bookkeeper/entrypoint.sh
```diff
@@ -74,6 +75,27 @@ create_bookie_dirs() {
...
+  BK_ID_FILE="${BK_ID_DIR}/id"
+  if [ `find ${BK_ID_FILE} | wc -l` -gt 0 ]; then
+    echo "Found BookieID"
+    BK_bookieId=`cat ${BK_ID_FILE}`
+  else
+    echo "BookieID not found"
+    if [[ $BK_setNewBookieID == "true" ]];then
+      BK_bookieId="`hostname -s`-`date "+%Y%m%d-%H%M%S"`"
+    else
+      HOST="$(echo -e `hostname -A` | sed -e 's/[[:space:]]*$//')"
```
RaulGracia, line 92: If the Operator permitted to change the option `useHostNameAsBookieID`, then this would be problematic, as you wouldn't know if it is needed to set the hostname or the IP. I think that the best way to set the previous ID of the Bookie is just to get it from the Cookie, if there is one. I know that we have not seen any command like "get me the ID from the Cookie for this Bookie", but that would be the most reliable approach I guess.

docker/bookkeeper/entrypoint.sh
```diff
@@ -74,6 +75,27 @@ create_bookie_dirs() {
...
+  IFS=',' read -ra directories <<< $1
+  BK_ID_DIR="${directories[0]}"
+  BK_ID_FILE="${BK_ID_DIR}/id"
+  if [ `find ${BK_ID_FILE} | wc -l` -gt 0 ]; then
+    echo "Found BookieID"
+    BK_bookieId=`cat ${BK_ID_FILE}`
+  else
+    echo "BookieID not found"
+    if [[ $BK_setNewBookieID == "true" ]];then
+      BK_bookieId="`hostname -s`-`date "+%Y%m%d-%H%M%S"`"
```
RaulGracia, line 90: I remember that it was suggested to use a UUID instead of timestamp here, right? It should not change much the purpose of the change, though.

docker/bookkeeper/entrypoint.sh
```diff
@@ -141,6 +163,8 @@ initialize_cluster() {
             fi
         fi
     fi
+    echo "Writing BookieID ${BK_bookieId} in the persistant file ${BK_ID_FILE}"
```
RaulGracia, line 166: Typo: `persistant` -> `persistent`

RaulGracia (changes requested):

docker/bookkeeper/entrypoint.sh
```diff
@@ -74,6 +75,27 @@ create_bookie_dirs() {
...
+# Create a Bookie ID if this is a newly added bookkeeper pod
+# or read the Bookie ID if it already exists
+set_bookieid() {
+  IFS=',' read -ra directories <<< $1
+  BK_ID_DIR="${directories[0]}"
+  BK_ID_FILE="${BK_ID_DIR}/id"
+  if [ `find ${BK_ID_FILE} | wc -l` -gt 0 ]; then
+    echo "Found BookieID"
+    BK_bookieId=`cat ${BK_ID_FILE}`
+  else
```
anishakj, line 87: Instead of looking for the options, cant we look if there are legers/journals metadata, if they are present then it is an upgrade case. If not it is the new pod coming up.

SrishT: Have incorporated this change

docker/bookkeeper/entrypoint.sh
SrishT: fixed

docker/bookkeeper/entrypoint.sh
SrishT: The webhook in the bookkeeper operator code ensures that the contents of the `directories` is not modifiable, hence it is guaranteed to have the directories sorted in the same order at all times.

docker/bookkeeper/entrypoint.sh
SrishT: Have incorporated this change

docker/bookkeeper/entrypoint.sh
SrishT: Have incorporated this change

docker/bookkeeper/bookkeeperTeardown.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
RaulGracia, line 1: Please, add copyright header here.

docker/bookkeeper/bookkeeperTeardown.sh
```diff
@@ -0,0 +1,24 @@
...
+set -ex
+
+HOST=`hostname -s`
+PRAVEGA_PATH=${PRAVEGA_PATH:-"pravega"}
+PRAVEGA_CLUSTER_NAME=${PRAVEGA_CLUSTER_NAME:-"pravega-cluster"}
+BK_CLUSTER_NAME=${BK_CLUSTER_NAME:-"bookkeeper"}
+BK_CLUSTER_ROOT_PATH=/${PRAVEGA_PATH}/${PRAVEGA_CLUSTER_NAME}/${BK_CLUSTER_NAME}
+BK_zkServers=$(echo "${ZK_URL:-127.0.0.1:2181}" | sed -r 's/;/,/g')
+
+if [[ $HOST =~ (.*)-([0-9]+)$ ]]; then
```
RaulGracia, line 12: Please, add comments about what this script is doing and why it is needed.

docker/bookkeeper/bookkeeperTeardown.sh
```diff
@@ -0,0 +1,24 @@
...
+  NAME=${BASH_REMATCH[1]}
+  ORD=${BASH_REMATCH[2]}
+else
+  echo Failed to parse name and ordinal of Pod
+  exit 1
+fi
+
+REPLICAS=`zk-shell --run-once "get ${BK_CLUSTER_ROOT_PATH}/conf" ${BK_zkServers} | cut -d'=' -f 2`
```
RaulGracia, line 20: What happens if this script is executed during a network glitch? Would it be critical that the `bookieformat` command is not executed? Would that require manual intervention when trying to scale up the same Bookie to delete the cookie?

RaulGracia (changes requested):

docker/bookkeeper/bookkeeperTeardown.sh
SrishT: done

docker/bookkeeper/bookkeeperTeardown.sh
SrishT: done

docker/bookkeeper/entrypoint.sh
SrishT: @RaulGracia [this](https://github.com/pravega/bookkeeper-operator/pull/166) PR also includes the change to disallow users from modifying the value of the option `useHostNameAsBookieID`. This would be the only way to set the previous ID since getting it from the cookie would not be possible till the connection with zookeeper cluster has been established, whereas the environment variable `bookieId` would need to be set and exposed before doing that.

docker/bookkeeper/bookkeeperTeardown.sh
SrishT: It would be critical to execute the `bookieformat` command only when we have upgraded from the older bookkeeper version. 
In case of a fresh installation with the latest bookkeeper and bookkeeper operator, this manual invention to delete old cookies before scaling down and subsequently scaling up the bookkeeper cluster would not be required.

gaddamas (changes requested):
please set JAVA_HOME in Dockerfile and remove it from entrypoint.sh line 45, as we are using same entrypoint.sh script for pravega with release build of bookkeeper and also with branch build of bookkeeper (for SDP-1.3)

RaulGracia (approved):
LGTM, assuming that all the upgrade cases have passed the tests.

docker/bookkeeper/bookkeeperTeardown.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Extract pod ordinal value from its hostname
+if [[ $HOST =~ (.*)-([0-9]+)$ ]]; then
+  ORD=${BASH_REMATCH[2]}
+else
+  echo "Failed to parse ordinal value of the pod"
+  exit 1
+fi
+
+CLUSTERSIZE=`zk-shell --run-once "get ${BK_CLUSTER_ROOT_PATH}/conf" ${BK_zkServers} | cut -d'=' -f 2`
```
shrids, line 35: nit: Can we have a comment indicating that the value is set by the bookeeper operator?

shrids (approved):
+1

anishakj (approved):
LGTM

docker/bookkeeper/bookkeeperTeardown.sh
SrishT: done

docker/bookkeeper/entrypoint.sh
fpj: This is worth a comment saying that there is an assumption on the order of directories. Just so that I understand, where is it getting `directories`?

docker/bookkeeper/entrypoint.sh
```diff
@@ -74,6 +73,28 @@ create_bookie_dirs() {
   done
 }
 
+# Create a Bookie ID if this is a newly added bookkeeper pod
+# or read the Bookie ID if it already exists
+set_bookieid() {
+  IFS=',' read -ra directories <<< $1
+  BK_ID_DIR="${directories[0]}"
+  BK_ID_FILE="${BK_ID_DIR}/id"
```
fpj, line 81: I'm not sure why we are reading this `id` file. From the BP, it is present only in the configuration file:

> This BookieId will be present only in the configuration file and it is the key to lookup the Cookie on the MetadataService. Inadvertently changing the BookieId will prevent the Bookie to boot as it won’t find a matching Cookie. There is no need to store the BookieId on the cookie or persist it on the local storage (ledgers, indexes or journal directories).

https://bookkeeper.apache.org/bps/BP-41-bookieid/

docker/bookkeeper/entrypoint.sh
```diff
@@ -74,6 +73,28 @@ create_bookie_dirs() {
...
+    if [ `find $BK_journalDirectory $BK_ledgerDirectories $BK_indexDirectories -type f 2> /dev/null | wc -l` -gt 0 ]; then
+      HOST="$(echo -e `hostname -A` | sed -e 's/[[:space:]]*$//')"
+      BK_bookieId="${HOST}:${BOOKIE_PORT}"
+    else
+      HOST="`hostname -s`"
+      BK_bookieId="${HOST}:${RANDOM}"
+    fi
+  fi
+  echo "BookieID = $BK_bookieId"
+  sed -i "s|.*bookieId=.*\$|bookieId=${BK_bookieId}|" ${BK_HOME}/conf/bk_server.conf
```
fpj, line 95: I suppose this works with the base image we are using? I'm asking because this command does not work on OS X, but it doesn't necessarily have to.

fpj (changes requested):
Thanks for the PR, @SrishT . I have a couple of questions.

docker/bookkeeper/entrypoint.sh
SrishT: @fpj the reason for persisting the bookieId within the `id` file was because in case of a bookie getting restarted, we would not want to set a new value for the bookieId, instead we would want to use the same value of bookieId we were using before the pod got restarted and expose this value as an environment variable within the pod, and the only way to read the older value would be by persisting this value.

docker/bookkeeper/entrypoint.sh
SrishT: Yes @fpj , I have verified that it works through my testing. 
Also we were already using something similar in the same file itself https://github.com/pravega/pravega/blob/ca2ba631c74122d38b894d084523f6c3d379f05c/docker/bookkeeper/entrypoint.sh#L113-L115

docker/bookkeeper/entrypoint.sh
eolivelli: @fpj in order projects, like HerdDB, we had the same approach, to store the bookieId in a file on the same directory.
assuming that the bookieid is strictly bound to the disks (contents) and not to the network address

eolivelli (approved):
LGTM
great work

docker/bookkeeper/entrypoint.sh
SrishT: @fpj it gets this value from the `options` field in the bookkeeper manifest

docker/bookkeeper/entrypoint.sh
fpj: This code is only writing the bookie id to a single directory (`BK_ID_DIR="${journal_directories[0]}"
  BK_ID_FILE="${BK_ID_DIR}/id"`). If we are trying to tie the bookie id to the disks (content), then do we have to write it to all directories and check accordingly?

gaddamas: What would startup (entrypoint) do if bk id is different across directories, which could be possible if restarts are frequent.

eolivelli: > then do we have to write it to all directories and check accordingly?
This is a good point. I agree

> What would startup (entrypoint) do if bk id is different across directories, which could be possible if restarts are frequent.
IMHO we should fail the startup and let some sysadmin to deal with the problem. Probably we mounted wrong disks

docker/bookkeeper/entrypoint.sh
```diff
@@ -73,6 +73,24 @@ create_bookie_dirs() {
...
 }
 
+# Create a Bookie ID if this is a newly added bookkeeper pod
+# or read the Bookie ID if a cookie containing this value already exists
+set_bookieid() {
+  IFS=',' read -ra journal_directories <<< $BK_journalDirectories
+  COOKIE="${journal_directories[0]}/current/VERSION"
+  if [ `find ${COOKIE} | wc -l` -gt 0 ]; then
+    # Reading the Bookie ID value from the existing cookie
+    bkHost=`cat ${COOKIE} | grep bookieHost`
```
fpj, line 83: I think this is assuming a text format, but checking the BK code, it looks like more recent versions format it using protobuf:

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java#L195
https://github.com/apache/bookkeeper/blob/master/bookkeeper-proto/src/main/proto/DataFormats.proto#L80

Are we sure this is going to work for more recent versions?

fpj (changes requested):

docker/bookkeeper/entrypoint.sh
eolivelli: IIRC the cookie is still in plain text

RaulGracia: Good point @fpj. As far as I know, this PR works with the defined version of Bookkeeper 4.14, which is the one we support today. In the case that the format changes in the future, we can address that in the future Bookkeeper upgrade and we can try to keep the change backwards compatible. Does it sound reasonable?

fpj: I'll try to confirm in the code that it is writing text, but if anyone knows and wants to share in the meanwhile, then please do it.

The cookie layout version is 4 already, so it is an indication that it must be writing in plain text if we are able to read it now:

https://github.com/apache/bookkeeper/blob/branch-4.14/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java#L69

fpj: It is fine from the write code path in Cookie.java:

https://github.com/apache/bookkeeper/blob/branch-4.14/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java#L223

fpj (approved):

Issue #4993, opened 2020-07-29
PDP-34 Add a way to register a custom ChunkStorage implementation


eolivelli:
**Problem description**
With PDP-34 is it easier to implement the LTS storage.
But it not actually possible to plug a custom implementation.

This issue tracks there request to be able to run a custom LTS storage, implemented as ChunkStorage.
We also need to have a way to start LocalPravegaEmulator with such a custom configuration, in order to allow simple integration tests. 

**Problem location**
SegmentStore

**Suggestions for an improvement**
After PDP-34 lands, or during the implementation of PDP34, remove StorageType enum, and create a way to configure a custom tier2.

Check https://github.com/eolivelli/pravega-blobit-chunkkmanager as an example of possible implementation of ChunkStorage using https://github.com/diennea/blobit a Storage Engine that stores data on Bookies.

derekm:
PDP-42 as in #4907 deals with this.

sachin-j-joshi:
Fixing it here https://github.com/pravega/pravega/pull/6231/

sachin-j-joshi:
closed via following PR- 
Issue 6223: SLTS - Enable loading third party storages. #6231

PR #6231, opened 2021-08-17
Issue 6223: SLTS - Enable loading third party storages.

### LINKED ISSUE #6230: SLTS - Cannot load 3rd party storages
**Describe the bug**
The storageImplementation field in ServiceConfig is a strongly typed enum.
This is unintentional and prevents loading of third party storages.
The field should be a string field/property

**To Reproduce**
Try specifying and loading any storage outside of Pravega tree

**Screenshots**
None

**Additional information**
None

sachin-j-joshi:

**Change log description**  
SLTS - Enable loading third party storages.

**Purpose of the change**  
Fixes #6230 #4993

**What the code does**  
- The storageImplementation field in ServiceConfig is a strongly typed enum.
- This is unintentional and prevents loading of third party storages.
- The field should be a string field/property

**How to verify it**  
All tests should pass


eolivelli (changes requested):
The patch makes sense to me.

We will have to add some minimal test in order to see it bootstrapping some non system LTS implementation.

I will try to run my custom implementation

eolivelli:
One question:
where do I put the jars for the LTS implementation ?

sachin-j-joshi:
The storage implementations are loaded using `ServiceLoader` (https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html)

As long as implementation jar is on the class path it should work.

In fact, there is nothing special about existing bindings other than they are part of the same repo and when we create docker image it puts bindings jar in same directory. But otherwise they use the same mechanism to load.

(The mechanism to load storage factory was already implemented in 0.8 timeframe however this specific bug in config parsing prevented passing arbitrary value to loader.)

Here are the steps to follow. 
**Step 1:**  Add resource file to indicate the class that implements Storage Factory creator
Example - https://github.com/pravega/pravega/blob/master/bindings/src/main/resources/META-INF/services/io.pravega.segmentstore.storage.StorageFactoryCreator

**Step 2:** Implement the StorageFactoryCreator that creates the factory for storage instances
Example  -https://github.com/pravega/pravega/blob/master/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorageFactoryCreator.java

**Step 3:** Implement factory 
https://github.com/pravega/pravega/blob/master/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemSimpleStorageFactory.java
Add custom configuration.
https://github.com/pravega/pravega/blob/master/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemStorageConfig.java


(https://github.com/pravega/pravega/blob/master/bindings/src/main/java/io/pravega/storage/filesystem/FileSystemChunkStorage.java is where ChunkStorage is implemented.)

eolivelli (changes requested):
Lgtm

So there is no need for additional testing.

eolivelli (approved):

andreipaduroiu (commented):
This looks OK to me, but how is a user to know how to load an external implementation.

At least one example, even if in a comment, on how to do that and what to pass into this config value would be great. I suggest we write this in the `StorageConfig` class (where we defined this property) and in `config.properties` where we list this property.

andreipaduroiu (changes requested):
.

eolivelli:
> At least one example, even if in a comment, on how to do that and what to pass into this config value would be great. I suggest we write this in the `StorageConfig` class (where we defined this property) and in `config.properties` where we list this property.

I second this. good idea @andreipaduroiu

sachin-j-joshi:
Updated comments - https://github.com/pravega/pravega/pull/6231/commits/9f0d53b6a430d24cdc1d3fa943eea9a1bbdd971b

segmentstore/storage/src/main/java/io/pravega/segmentstore/storage/chunklayer/ChunkStorage.java
```diff
@@ -49,8 +49,22 @@
...
+ *     <li> Implement {@link ChunkStorage}. It is recommended that the implementations should extend {@link AsyncBaseChunkStorage}
+ *     or {@link BaseChunkStorage}.</li>
+ *     <li> Implement {@link io.pravega.segmentstore.storage.SimpleStorageFactory}.  </li>
+ *     <li> Implement {@link io.pravega.segmentstore.storage.StorageFactoryCreator}.
+ *     <ul>
+ *          <li>Return {@link io.pravega.segmentstore.storage.StorageFactoryInfo} object containing name to use to identify.
+ *          This identifier is used in config.properties file to identify implementation to use.</li>
+ *          <li> The bindings are loaded using ServiceLoader (https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html)</li>
+ *          <li> Add resource file named "io.pravega.segmentstore.storage.StorageFactoryCreator" to the implementation jar.
+ *          File contains name of the class implementing {@link io.pravega.segmentstore.storage.StorageFactoryCreator}.</li>
```
eolivelli, line 64: typo: "File contains name" -> "That file contains the name"

eolivelli:
@andreipaduroiu can you please take another look ?

cc @diegosalvi you are interested in this patch as well

andreipaduroiu (approved):

PR #6235, opened 2021-08-18
Issue 6229:  Setting tlsProtocolVersion defaults in InProcPravegaCluster

### LINKED ISSUE #6229: LocalPravegaEmulator requires setting tlsProtocolVersion in all cases
**Describe the bug**
PR #6114 introduced a regression whereby `tlsProtocolVersion(String[] versions)` is now a required field of `LocalPravegaEmulator.builder()`.

**To Reproduce**
1. Find an older usage of LocalPravegaEmulator, such as: [PravegaResource.java#L36](https://github.com/pravega/snmp-trap-handler/blob/master/snmp-trap-handler/src/test/java/io/pravega/snmp/PravegaResource.java#L36)
2. Notice its use fails with NPE here: [InProcPravegaCluster.java#L301](https://github.com/pravega/pravega/blob/master/standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java#L301) unless `.tlsProtocolVersion(new String[] {"TLSv1.2","TLSv1.3"})` (or equivalent) is specified in the `LocalPravegaEmulator.builder()` construction, even with `.enableTls(false)`

**Additional information**
Sane defaults should probably be in `InProcPravegaCluster` rather than in [PravegaEmulatorResource.java#L66](https://github.com/kotlasaicharanreddy/pravega/blob/master/standalone/src/test/java/io/pravega/local/PravegaEmulatorResource.java#L66) and elsewhere.

kotlasaicharanreddy:
**Work in Progress**
Signed-off-by: SaiCharan <kotlasaicharan@yahoo.com>

**Change log description**  
- TLSv1.3 feature PR introduced a variable `tlsProtocolVersion` in `InProcPravegaCluster` and `LocalPravegaEmulator` classes
- When `tlsProtocolVersion` is not set in these classes, throws an NPE 
- Added default values of `tlsProtocolVersion` in these classes. 

**Purpose of the change**  
Fixes #6229 

**What the code does**  
Added default values of  `tlsProtocolVersion` in `InProcPravegaCluster`  and `LocalPravegaEmulator` classes.

**How to verify it** 
Old usage of [InProcPravegaCluster](https://github.com/pravega/flink-connectors/blob/master/src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java#L321)  and [LocalPravegaEmulator](https://github.com/pravega/snmp-trap-handler/blob/master/snmp-trap-handler/src/test/java/io/pravega/snmp/PravegaResource.java#L36)  should continue to work.


standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
```diff
@@ -74,7 +74,7 @@ public LocalPravegaEmulator build() {
                     .enableRestServer(enableRestServer)
                     .enableAuth(enableAuth)
                     .enableTls(enableTls)
-                    .tlsProtocolVersion(tlsProtocolVersion)
+                    //.tlsProtocolVersion(tlsProtocolVersion)
```
derekm, line 77: Shouldn't both of these facilities still accept overrides instead of only being hardcoded? In other words, sane defaults should be used unless `.tlsProtocolVersion()` is called to override the defaults.

derekm (changes requested):

standalone/src/main/java/io/pravega/local/LocalPravegaEmulator.java
kotlasaicharanreddy: Yes. Added default values in both LocalPravegaEmulator and InProcPravegaCluster classes. They are used unless `.tlsProtocolVersion()` is called to override them.

andreipaduroiu:
@derekm Can you please take another look at this PR and see if your comment was addressed?

derekm (changes requested):
Should we use constants here instead of copy/pasting literals across the codebase?

`SecurityConfigDefaults`, `ServiceConfig`, `SingleNodeConfig` and the controller's `Config` all have this value as public members. `RESTServerConfigImpl` has the value as private.

derekm (approved):

PR #6237, opened 2021-08-19
Issue 6238: Security related fixes 

### LINKED ISSUE #6238: Update 3rd party libraries to address their security vulnerabilities
**Describe the bug**
A security scan of Pravega docker images flags multiple libraries as their current versions have known CVEs 

List of flagged components: 
1. Apache Commons 1.20 -> CVE-2021-35516, CVE-2021-36090, CVE-2021-35517 
2. Apache Log4j 1.2.17 -> CVE-2019-17571, CVE-2017-5645
3. Jackson Mapper 1.9.13 -> CVE-2019-10172
4. Jackson Databind 2.9.10.6 -> CVE-2021-20190, CVE-2020-36182, CVE-2020-36179
5. JDOM 2.0.6 -> CVE-2021-33813
6. Jetty JNDI 9.4.20 -> CVE-2021-28165, CVE-2020-27216
7. JSON Smart 2.3 -> CVE-2021-27568
8. SnakeYAML 1.23 -> CVE-2017-18640
9. Log4j 1.2.17 -> CVE-2019-17571, CVE-2017-5645
10. Bouncy Castle 1.6 -> CVE-2020-15522, CVE-2020-26939
11. Commons IO  2.29 -> CVE-2021-29425

**To Reproduce**
Run a security scan like BlackDuck or similar 

**Screenshots**
None

**Additional information**
Not all of flagged libraries are direct dependencies, some of them come as transient ones. And in addition to java libraries the base docker image contains a handful number of vulnerabilities.

medvedevigorek:
**Change log description**  
1. Dockerfile updated to use Alpine+AdoptOpenJDK11 as the base container image 
2. Changed invocation of keytool to make it more flexible as previously it hardcoded location of the system cacerts which actually differs from deployment to deployment
3. Bunch of updates in dependency versions to pull newer versions 

**Purpose of the change**  
Fixes #6238 

**What the code does**  
1. Updated Swagger to pull newer version of SnakeYAML
2. Updated Hadoop to pull newer version of Jetty, JSON Smart and others 
3. Removed unused library definitions from gradle.properties file 
4. Updated commons-io and bouncycastle versions
5. Exclude log4j jar file from Pravega docker image 

**How to verify it**


docker/pravega/Dockerfile
```diff
@@ -13,27 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-FROM openjdk:11.0.8-jre-slim
+FROM adoptopenjdk/openjdk11:jre-11.0.11_9-alpine
 
-RUN apt-get update && apt-get install -y -q \
```
eolivelli, line 18: why are you removing this useful tools?

gradle.properties
```diff
@@ -21,34 +21,32 @@ org.gradle.jvmargs=-Xms1g
 
 #3rd party Versions
 apacheCommonsCsvVersion=1.5
-apacheCommonsCompressVersion=1.20
+apacheCommonsCompressVersion=1.21
```
eolivelli, line 24: can we have separate PRs for Java dependencies upgrades and the Docker image changes ?

eolivelli (changes requested):
good work.

I believe it is better to separate the Docker changes from the Java dependency changes, otherwise in case of problems it is difficult to track down to a specific trigger

docker/pravega/Dockerfile
medvedevigorek: I don't see us using requiring them (yet to be proved by testing) hence no reason to keep them in the image. The lesser components installed the smaller final docker image and lesser chances of getting CVEs found

gradle.properties
medvedevigorek: We are testing them together by running security scans on the final image so having them in a single branch makes the testing easier.  We may consider splitting them once all the testing is complete

docker/pravega/Dockerfile
eolivelli: makes sense to me.

we are running the pods as root, correct ? so the user is able to add debugging tools easily
 
if we switch to root-less docker images then we will have to provide debugging tools out of the box probably, otherwise it will be very difficult to do troubleshooting of problems

gradle.properties
eolivelli: I understand

docker/pravega/Dockerfile
```diff
@@ -13,27 +13,9 @@
...
-    rpcbind \
-    nfs-common \
-    python \
```
shrids, line 21: This would affect the entry point script https://github.com/pravega/pravega/blob/56127851f076ba96938a1aae25294ac28d17abb9/docker/pravega/scripts/entrypoint.sh#L31

I do not see this being used when Pravega is being deployed on K8s. 
It does seem to be used on docker-compose. https://github.com/pravega/pravega/blob/a5f5d670b7f2340b8388a9a2b216aa7e5b4d7151/docker/compose/swarm/pravega.yml#L32

docker/pravega/Dockerfile
medvedevigorek: wouldn't be two different cases: running a pod as root and be logged in as root inside a pod?

build.gradle
```diff
@@ -1448,7 +1448,7 @@ task preparePravegaImage(type: Copy) {
     from "docker/pravega"
     from (installDist) {
         into "pravega"
-        exclude "**/*.bat"
+        exclude "**/*.bat", "**/log4j*.jar"
```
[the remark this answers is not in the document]
medvedevigorek: yes. we need to exclude log4j 1.2.x. as this is the one with reported critical CVE

docker/pravega/Dockerfile
medvedevigorek: if you refer to python then it's installed in the new image too. I had to add it exactly for `wait_for` logic. Actually we need to remove `wait_for` as it's a simple one but requires entire python installation.

shrids: ok.

shrids (approved):
+1.
The docker image size of Pravega reduces to around `382MB` from the prev `644MB`.

docker/pravega/Dockerfile
medvedevigorek: Btw, `WAIT_FOR` is also used in K8s, Pravega Operator sets it for both Controller and Segment Store 
https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravega/pravega_segmentstore.go#L299

co-jo:
LGTM

co-jo (approved):
LGTM.

RaulGracia:
@eolivelli please, let us know if there is anything to address in this PR or if it is good to go.

docker/pravega/Dockerfile
eolivelli: you cannot login as root in a rootless container, so you cannot install additional tools (there are some tricks but they are hard and not for the regular sys admins)

btw we are not still targeting rootless containers so it is fine to have the smallest image as possible

eolivelli (approved):
@RaulGracia I thought that this was a proof of concept PR and it was not meant to be committed.
I suggested to spit the patch at least in:
- java dependencies updates
- docker image updates
this way it is easier to track down to possible problems

btw I am not going to block this patch if you are comfortable with it

RaulGracia:
Thanks @eolivelli, I think that given that the purpose of this PR is fixing few vulnerabilities in the Pravega image, both the base image and several dependencies have been considered in the same PR (separating them may be good as well). I'm aware that we have executed several rounds of tests for this PR, but will let @medvedevigorek comment if we are confident enough on the testing of this change or not.

Issue #6183, opened 2021-07-28
readSegment in SegmentHelper takes the offset parameter as an integer instead of long


pbelgundi:
**Describe the bug**
SegmentHelper.readSegment() takes an integer segment offset as where as the underlying WireCommand it invokes can take a long offset. Using an Integer offset can limit us to reading at max first 2GB of data in a Segment and hence it would help to change this API to take a long offset instead of int.

**Additional information**
Fix method signature for SegmentHelper.readSegment() method: https://github.com/pravega/pravega/blob/master/controller/src/main/java/io/pravega/controller/server/SegmentHelper.java#L638

anirudhkovuru:
As part of fixing this bug, every all usages of `SegmentHelper.readSegment()` need to be checked and fixed accordingly.

PR #6239, opened 2021-08-19
Issue 6183: Change readSegment method Parameter from int to long in SegmentHelper

shshashwat:
Signed-off-by: Shashwat Sharma <shashwat_sharma@dell.com>

**Change log description**  
SegmentHelper.readSegment() takes an integer segment offset where as the underlying WireCommand it invokes takes a long offset. Using an Integer offset can limit the reading at max first 2GB of data in a Segment and so changing this API to take a long offset instead of int.

**Purpose of the change**  
#6183 

**What the code does**  
Update SegmentHelper.readSegment() method to accept long offset instead of int

**How to verify it**  
All existing integration and system tests should pass


controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
```diff
@@ -635,7 +635,7 @@ private String getTransactionName(String scope, String stream, long segmentId, U
                 });
     }
 
-    public CompletableFuture<WireCommands.SegmentRead> readSegment(String qualifiedName, int offset, int length,
+    public CompletableFuture<WireCommands.SegmentRead> readSegment(String qualifiedName, long offset, int length,
```
pbelgundi, line 638: Please change `SegmentHelperTest.testReadSegment()` to pass long offset values instead of integers to this method.

pbelgundi: Also in class [ReadSegmentRangeCommand](https://github.com/pravega/pravega/blob/3a11b6d5496bc78c93b0ccaac14e9f6327c9e5bb/cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java#L117) we seem to be type casting the `currentOffset` to "int", which can now be removed.

pbelgundi (changes requested):

controller/src/main/java/io/pravega/controller/server/SegmentHelper.java
shshashwat: Changes done

pbelgundi (approved):
LGTM

cli/admin/src/main/java/io/pravega/cli/admin/segmentstore/ReadSegmentRangeCommand.java
```diff
@@ -114,7 +114,7 @@ private void readAndWriteSegmentToFile(SegmentHelper segmentHelper, String segme
             // TODO: readSegment should take in longs as described here: https://github.com/pravega/pravega/issues/6183
             // TODO: Remove casting to int once 6183 is resolved.
```
shrids, line 115: Please remove the TODO.

shshashwat: Yes, that makes sense. Thank you Sandeep. It's removed :)

andreipaduroiu (approved):

PR #6274, opened 2021-09-08
Issue 6204: make InProcPravegaCluster listen on all IPv4 interfaces

### LINKED ISSUE #6204: Change segment store service host address in a Pravega standalone cluster in docker
# Ask your question away
We are running a Pravega standalone and a schema registry service inside a docker compose environment. When the schema registry tried to communicate with segment store, it always got `localhost` as the service host address. 
Found the config of standalone cluster:
https://github.com/pravega/pravega/blob/39a39871e01eff63f8b87a2858c495673d77ba05/standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java#L302
https://github.com/pravega/pravega/blob/39a39871e01eff63f8b87a2858c495673d77ba05/standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java#L303

Is there any approach to change the segment store host address prior to the Pravega container start?

derekm:
**Change log description**  
Make InProcPravegaCluster listen on all IPv4 interfaces instead of just localhost, so that Docker can redirect traffic to segment store.

**Purpose of the change**  
Fixes #6204

**What the code does**  
Allows segment store to listen on all IPv4 interfaces.

**How to verify it**  
Start standalone via docker as according to: https://github.com/pravega/pravega/blob/master/documentation/src/docs/deployment/run-local.md#from-docker-image
and confirm that you are able to write and read to/from a stream.

### DIFF
--- a/standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
+++ b/standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
@@ -306,7 +306,7 @@ public class InProcPravegaCluster implements AutoCloseable {
                         .with(ServiceConfig.REST_KEYSTORE_PASSWORD_FILE, this.keyPasswordFile)
                         .with(ServiceConfig.CERT_FILE, this.certFile)
                         .with(ServiceConfig.ENABLE_TLS_RELOAD, this.enableTlsReload)
-                        .with(ServiceConfig.LISTENING_IP_ADDRESS, LOCALHOST)
+                        .with(ServiceConfig.LISTENING_IP_ADDRESS, "0.0.0.0")
                         .with(ServiceConfig.PUBLISHED_IP_ADDRESS, LOCALHOST)
                         .with(ServiceConfig.CACHE_POLICY_MAX_TIME, 60)
                         .with(ServiceConfig.CACHE_POLICY_MAX_SIZE, 128 * 1024 * 1024L)


derekm:
This PR introduces a quick-fix for allowing Docker to communicate with the Segment Store, by having the segment store listen on "0.0.0.0" while still handing out "localhost" as the segment store address. Docker uses an internal subnet for exposed port proxies, so services must listen on all interfaces. From outside of Docker, the Segment Store port will be mapped on localhost, so apps external to Docker can communicate with segment store.

Having Segment Store listen on "0.0.0.0" forces IPv4. Historically, we have had issues getting Pravega to listen on all interfaces across both IPv4 and IPv6.

If we drilled down further in `ServiceStarter`, `PravegaConnectionListener`, & `AbstractConnectionListener`, then we should be able to call Netty's `AbstractBootstrap.bind()` multiple times with IPv4 and IPv6 addresses. This may either require changes to the higher level APIs or the introduction of magic string values for the "listening IP address," such that `AbstractConnectionListener` can call v4 and v6 versions of `bind()` for either "all interfaces" or "localhost."

shrids (approved):
LGTM. Started standalone on docker and wrote events to a stream using the HOST_IP to verify it.

andreipaduroiu (changes requested):
@derekm Have you confirmed that local access to standalone still works? You can use the User CLI or one of our samples running locally to verify this.

crazyzhou (changes requested):
This change can fix some of the issues, but still having some problems. I have tested the following things.

1. Local client and standalone on the same machine. This could work previously, and can work after the change.
2. On a local docker environment. Start a docker container running Pravega standalone with

```
docker run -it -p 9090:9090 -p 12345:12345 brianzhou/pravega:derek-test standalone
```

and access with client connecting `tcp://localhost:9090`, the client can access segment store and perform reads and writes.
This did not work previously, but now it works with this fix.

3. On a remote machine/VM, start the standalone and start a local client connecting `tcp://<host-ip>:9090`, the client cannot access the segment store to perform read/writes because the publish IP is set to `localhost`.
This used to work on release 0.9.1, but does not work on 0.10-SNAPSHOT, even with this fix.

I would suggest changing the publish IP as well in this PR, to fix the scenario.

eolivelli (approved):
+1

derekm:
> 3. On a remote machine/VM, start the standalone and start a local client connecting `tcp://<host-ip>:9090`, the client cannot access the segment store to perform read/writes because the publish IP is set to `localhost`.
>    This used to work on release 0.9.1, but does not work on 0.10-SNAPSHOT, even with this fix.
> 
> I would suggest changing the publish IP as well in this PR, to fix the scenario 3.

@crazyzhou This might be hard problem when listening on "all interfaces". Since clients ask controllers for segment store address, the controller may need to know what interface the client is using to talk to the controller in order to deliver a compatible IP for a segment store.

I'll go back and compare again how 0.9.1 did this...

derekm:
> @derekm Have you confirmed that local access to standalone still works? You can use the User CLI or one of our samples running locally to verify this.

@andreipaduroiu I did confirm that for the docker image. Brian points out that remote access remains broken vs. 0.9.1. I don't think I checked `./gradlew startStandalone`, but I'll add checking that as I look into how to re-enable remote access.

standalone/src/main/java/io/pravega/local/InProcPravegaCluster.java
```diff
@@ -306,7 +306,7 @@ private void startLocalSegmentStore(int segmentStoreId) throws Exception {
                         .with(ServiceConfig.REST_KEYSTORE_PASSWORD_FILE, this.keyPasswordFile)
                         .with(ServiceConfig.CERT_FILE, this.certFile)
                         .with(ServiceConfig.ENABLE_TLS_RELOAD, this.enableTlsReload)
-                        .with(ServiceConfig.LISTENING_IP_ADDRESS, LOCALHOST)
+                        .with(ServiceConfig.LISTENING_IP_ADDRESS, "0.0.0.0")
```
tkaitchuck, line 309: Should the value of the constant be changed?

derekm: @tkaitchuck I think we need two constants, one for LOCALHOST, another for ALL_INTERFACES. I'm about to push my changes for `ServiceConfig.PUBLISHED_IP_ADDRESS`...

RaulGracia (commented):
Looks good, assuming that Standalone and integration tests work fine (I have triggered a branch builder on this branch).

PR #6291, opened 2021-09-14
Issue : EventStreamWriter.flush() should throw RetriesExhaustedException in presense of connectivity failures consistently.

shrids:
**Change log description**  
Ensure EventStreamWriter.flush() throws RetriesExhaustedException incase of all retries to establish connection fails.

**Purpose of the change**  
Fixes #

**What the code does**  
Ensure the writer checks the connectivity state even in the case of zero inflight events.

**How to verify it**  

### DIFF
--- a/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
+++ b/client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
@@ -211,7 +211,7 @@ class SegmentOutputStreamImpl implements SegmentOutputStream {
                 }
                 log.info("Handling exception {} for connection {} on writer {}. SetupCompleted: {}, Closed: {}",
                          throwable, connection, writerId, connectionSetupCompleted == null ? null : connectionSetupCompleted.isDone(), closed);
-                if (exception == null) {
+                if (exception == null || throwable instanceof RetriesExhaustedException) {
                     exception = throwable;
                 }
                 connection = null;
@@ -587,6 +587,8 @@ class SegmentOutputStreamImpl implements SegmentOutputStream {
                 throw new SegmentSealedException(segmentName + " sealed for writer " + writerId);
             }
 
+        } else if (state.exception instanceof RetriesExhaustedException) {
+            throw Exceptions.sneakyThrow(state.exception);
         }
     }
 
--- a/client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
+++ b/client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
@@ -286,7 +286,6 @@ public class SegmentOutputStreamTest extends LeakDetectorTestSuite {
         MockController controller = new MockController(uri.getEndpoint(), uri.getPort(), cf, true);
         ClientConnection connection = mock(ClientConnection.class);
         cf.provideConnection(uri, connection);
-        @Cleanup
         SegmentOutputStreamImpl output = new SegmentOutputStreamImpl(SEGMENT, true, controller, cf, cid, segmentSealedCallback,
                 retryConfig, DelegationTokenProviderFactory.createWithEmptyToken());
         output.reconnect();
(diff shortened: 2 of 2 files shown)


client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -211,7 +211,7 @@ private void failConnection(Throwable throwable) {
                 }
                 log.info("Handling exception {} for connection {} on writer {}. SetupCompleted: {}, Closed: {}",
                          throwable, connection, writerId, connectionSetupCompleted == null ? null : connectionSetupCompleted.isDone(), closed);
-                if (exception == null) {
+                if (exception == null || throwable instanceof RetriesExhaustedException) {
```
andreipaduroiu, line 214: I suggest using `Exceptions.unwrap` here in case the exception is wrapped in a `CompletionException` or `ExecutionException`.

andreipaduroiu: Same below

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
```diff
@@ -248,7 +248,6 @@ public void testConnectWithMultipleFailures() throws Exception {
         ClientConnection connection = mock(ClientConnection.class);
         InOrder verify = inOrder(connection);
         cf.provideConnection(uri, connection);
-        @Cleanup
```
andreipaduroiu, line 251: Why?

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: done. this will ensure `state.exception ` is always unwrapped.

client/src/test/java/io/pravega/client/segment/impl/SegmentOutputStreamTest.java
shrids: Valid point. I have ensured close is invoked in the finally block for all the tests.

shrids:
The system test run is successful with this changeset. `/job/test-pravega-continuous/4128/`

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
```diff
@@ -194,9 +194,10 @@ private ClientConnection getConnection() {
         }
 
         /**
-         * @param throwable Error that has occurred that needs to be handled by tearing down the connection.
+         * @param ex Error that has occurred that needs to be handled by tearing down the connection.
          */
-        private void failConnection(Throwable throwable) {
+        private void failConnection(Throwable ex) {
+            Throwable throwable = Exceptions.unwrap(ex);
```
tkaitchuck, line 200: Currently SegmentOutputStreamImpl.failConnection() is doing this unwrapping before invoking this. If you want the unwrapping here, please remove it from the callers.

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
```diff
@@ -259,12 +260,23 @@ public void flush() {
...
+                                 writer.getSegmentName(), writerId);
+                        retriesExhaustedException = e1;
                     }
                 }
             }
+            if (retriesExhaustedException != null) {
+                log.error("Flush by writer {} on Stream {} failed after all retries to connect with Pravega exhausted.",
+                          writerId, stream.getScopedName());
+                throw retriesExhaustedException;
+            }
```
tkaitchuck, line 279: This logic strikes me as wrong. It catches retriesExhausted and then doesn't break out of the loop so it keeps retrying until it succeeds and then it throws retriesExhausted anyway even though it was successfully written. 
I don't see the point in catching and rethrowing. If the exception is going to be propagated upward, there is no need to re-log it at every stack frame along the way, just let it go up.

tkaitchuck (changes requested):

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
shrids: The reason for this change is to handle a scenario where out of the many segment writers only a subset of them face a connectivity issue with a specific segment store instance. This logic would ensure we ensure flush is invoked on every segment writer before throwing a RetriesExhaustedException to the application. 
(This would ensure events from segment writers who still have connectivity with SegmentStores can flush all their inflight events before the application is intimated about the connectivity

client/src/main/java/io/pravega/client/segment/impl/SegmentOutputStreamImpl.java
shrids: Thanks, I have removed the redundant `Exceptions.unwrap(ex)`

client/src/main/java/io/pravega/client/stream/impl/EventStreamWriterImpl.java
tkaitchuck: ok. But it doesn't do that. It will keep retrying as long as any segment fails to write forever. And it may throw an error if it succeeds.

PR #6295, opened 2021-09-15
Issue 6294: Need for setting memory limits in Kubernetes system test pods

### LINKED ISSUE #6294: Need for setting memory limits in Kubernetes system test pods
**Describe the bug**
We need to set memory limits to Segment Store and Controller pods in Kubernetes system tests. Otherwise, they can sporadically throw OOM problems.

**To Reproduce**
Run system tests, OOM can be seen sporadically on Segment Stores.

**Screenshots**
n/a

**Additional information**
n/a

RaulGracia:
**Change log description**  
Added memory limits for JVM heap size and Direct Memory in Kubernetes system test pods.

**Purpose of the change**  
Fixes #6294.

**What the code does**  
Sets some reasonable memory limits to pods, specially to Segment Store pods which could go OOM in some situations as no heap or direct memory limits are set. Specifically, the Segment Store defines a 4GB cache size by default and 5GB of direct memory are allocated. The pod size is 6GB and the JVM heap size is set to 1GB. Given the load in our system tests, this should me more than sufficient.

**How to verify it**  
Have not found OOM issues with this change.

### DIFF
--- a/test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
+++ b/test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
@@ -224,7 +224,7 @@ public abstract class AbstractService implements Service {
     // Removal of the JVM option 'UseCGroupMemoryLimitForHeap' is required with JVM environments >= 10. This option
     // is supplied by default by the operators. We cannot 'deactivate' it using the XX:- counterpart as it is unrecognized.
     private String[] getSegmentStoreJVMOptions() {
-        return new String[]{"-XX:+UseContainerSupport", "-XX:+IgnoreUnrecognizedVMOptions", "-XX:MaxDirectMemorySize=5Gi", "-Xmx1024m"};
+        return new String[]{"-XX:+UseContainerSupport", "-XX:+IgnoreUnrecognizedVMOptions", "-XX:MaxDirectMemorySize=5g", "-Xmx1024m"};
     }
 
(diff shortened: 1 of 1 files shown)


test/system/src/main/java/io/pravega/test/system/framework/services/kubernetes/AbstractService.java
```diff
@@ -224,15 +224,15 @@ public String getID() {
     // Removal of the JVM option 'UseCGroupMemoryLimitForHeap' is required with JVM environments >= 10. This option
     // is supplied by default by the operators. We cannot 'deactivate' it using the XX:- counterpart as it is unrecognized.
     private String[] getSegmentStoreJVMOptions() {
-        return new String[]{"-XX:+UseContainerSupport", "-XX:+IgnoreUnrecognizedVMOptions"};
+        return new String[]{"-XX:+UseContainerSupport", "-XX:+IgnoreUnrecognizedVMOptions", "-XX:MaxDirectMemorySize=5g", "-Xmx1024m"};
```
andreipaduroiu, line 227: Max 5GB for DM; 1GB for heap. In line 134 you have an allocation of  6GB, which doesn't leave any room for the runtime.

Should we set DM to 4GB, heap to 1GB, and Cache to 2GB (I remember the cache was defaulted to 4GB - should we adjust that too?)

shrids (approved):

anishakj (approved):
LGTM

andreipaduroiu (approved):

PR #6299, opened 2021-09-16
Issue 6298: pravega version change for 0.10 branch

### LINKED ISSUE #6298: Change `pravegaVersion` in gradle.properties for r0.10 branch
**Describe the change**
Change `pravegaVersion` in `gradle.properties` for r0.10 branch 
This is required as part of the release process for Pravega 0.10.

pbelgundi:

**Change log description**  
Update pravegaVersion for 0.10 branch

**Purpose of the change**  
Fixes #6298

**How to verify it**  
All tests should pass

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -69,7 +69,7 @@ jjwtVersion=0.9.1
 bouncyCastleVersion=1.69
 
 # Version and base tags can be overridden at build time
-pravegaVersion=0.10.0-SNAPSHOT
+pravegaVersion=0.10.1-SNAPSHOT
 pravegaBaseTag=pravega/pravega
 bookkeeperBaseTag=pravega/bookkeeper
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -69,7 +69,7 @@ jjwtVersion=0.9.1
 bouncyCastleVersion=1.69
 
 # Version and base tags can be overridden at build time
-pravegaVersion=0.10.1-SNAPSHOT
+pravegaVersion=0.10.0
 pravegaBaseTag=pravega/pravega
 bookkeeperBaseTag=pravega/bookkeeper
 


shrids (approved):

RaulGracia:
@pbelgundi I think that we first need to set pravegaVersion to 0.10.0 to make the release at this point, and only after that, change again the suffix to `SNAPSHOT`.

RaulGracia (changes requested):
I think this one comes after the actual release.

pbelgundi:
> I think this one comes after the actual release.

I have done this per instructions in the release wiki: 
https://github.com/pravega/pravega/wiki/How-to-release#major-release

I also see the same kind of version change being made on earlier release branches.

RaulGracia:
@pbelgundi this it what I read:
> In r0.3, create an issue and corresponding pull request to change the pravegaVersion in gradle.properties to 0.3.0. Once that change is merged, we need to tag the commit point in the same way we described for a bug-fix release. See instructions in the previous section.

In this case, we are in `r0.10`, so we should be changing `pravegaVersion=0.10.0`, right?

gradle.properties
```diff
@@ -69,7 +69,7 @@ jjwtVersion=0.9.1
 bouncyCastleVersion=1.69
 
 # Version and base tags can be overridden at build time
-pravegaVersion=0.10.0-SNAPSHOT
+pravegaVersion=0.10.1-SNAPSHOT
```
andreipaduroiu, line 72: @pbelgundi We first need to set this to `0.10.0`, do the release, then change it to `0.10.1` after the release is done. That way the build artifacts point to a commit in GH that has the correct version set.

RaulGracia (approved):

gradle.properties
pbelgundi: Ok done

shrids (approved):
Thanks for pointing that out @RaulGracia

